PR #1, opened 2017-03-25
Mergerepos

btracey:


btracey:
This merges the Gonum repos together into `gonum/gonum. 

It merged the following repos, in this order 
````
                 "license",
		"internal",
		"blas",
		"floats",
		"lapack",
		"matrix",
		"mathext",
		"stat",
		"diff",
		"integrate",
		"optimize",
		"graph",
		"unit",
````

It did not include `plot`, `gonum.github.io`, `hdf5`, `talks`.

For each repository, the following set of commands was run (with the appropriate substitutions)
````
git remote add license_remote https://github.com/gonum/license
git fetch license_remote
git merge -s ours --no-commit license_remote/master
git read-tree --prefix=license/ -u license_remote/master
git commit -m "Imported license as a subtree"
````

I'm happy to share the script if that would be useful.

**Comment by btracey** [comment]
PTAL @kortschak @sbinet @vladimir-ch

**Comment by sbinet** [comment]
LGTM (even if actually didn't look thru all the diffs :P)

**Comment by kortschak** [comment]
What is the plan from here? Is this just the repo merge?

I've done a diff over the repos as they stand and the directories in this and I'm happy with that.

One significant question that I have is how currently extant PRs in the old repos will be merged here.

**Comment by kortschak** [comment]
I forgot, with the merging of repos it become more important that PR and commit message have an indication of packages being affected. For example here, "all: merge repos".

**Comment by btracey** [comment]
I wanted to get a confirmation I was doing this correctly.

I would suggest that the ones which are in the process of merging be completed. For example, the `dggsvd3` PR has had a lot of work go into it. I would explicitly request that https://github.com/gonum/matrix/pull/436 and companions be merged afterward so old code isn't affected by those breaking changes. For everything else, one should be able to transfer the git diff to the new repo, from what I understand (some PRs have stalled, for instance). 

We can decide on specific PRs here. For instance, maybe https://github.com/gonum/matrix/pull/404 is better left after the merge so we can have the complex matrices properly with the first commit.

**Comment by btracey** [comment]
To add: It's really easy to just close this PR and do it again once we're done with the in-progress PRs.

I can also modify the commit messages here to include the package prefixes.

**Comment by kortschak** [comment]
OK. If this is an experiment, then it's worth also making sure the .travis* files are working and the tests pass.

File: .travis.yml [hunk 1]
```diff
@@ -0,0 +1,23 @@
+sudo: false
+
+language: go
+
+# Versions of go that are explicitly supported by gonum.
+go:
+ - 1.5.4
```
kortschak: Please update these.

.travis.yml
```diff
@@ -0,0 +1,23 @@
...
+ - go get github.com/mattn/goveralls
+
+# Get deps, build, test, and ensure the code is gofmt'ed.
+# If we are building as gonum, then we have access to the coveralls api key, so we can run coverage as well.
+script:
+ - go get -d -t -v ./...
+ - go build -v ./...
+ - go test -v ./...
+ - test -z "$(gofmt -d .)"
+ - if [[ $TRAVIS_SECURE_ENV_VARS = "true" ]]; then bash ./.travis/test-coverage.sh; fi
```
kortschak: Can you add the new line here.

btracey:
I committed a travis script and greenlight the repo on travis. I'm not sure how to kick the tests off. 

The tests should probably pass, but they aren't really right, because we'll also need to adjust all of the import paths.

.travis.yml
btracey: I looked at a couple other travis scripts, and I'm not sure what you mean.

btracey:
Do you want the internal files deleted as part of this series of commits?

.travis.yml
kortschak: They are missing the final `\n` too.

kortschak:
I suspect travis is not testing this because there is no ".travis.yml" in master.

kortschak:
I think maybe the first PR should be to get travis hooked.

btracey:
Trying again with the comments and new Travis script

PR #2, opened 2017-03-27
Add Travis script

btracey:


btracey:
I'm not really sure how this will interact (or not) with BLAS and Lapack. Does the full travis script need to be the superset of the commands (for instance, have sudo be needed everywhere?)

btracey:
Okay, got it. PTAL if you would like.

PR #3, opened 2017-03-30
Mergerepos2

btracey:


btracey:
This isn't going to pass because of the internal imports. I would imagine it's best to merge the trees, and from there fix all of the imports? Or should I also try to rewrite the imports as part of this change.

kortschak:
>This isn't going to pass because of the internal imports. I would imagine it's best to merge the trees, and from there fix all of the imports? Or should I also try to rewrite the imports as part of this change.

In two parts in the one PR I think.

I'd still like to get some things merged in matrix though - the gsvd things.

btracey:
Okay, I used rogpeppe/govers  to rewrite the imports. From random sampling it looks like it did what I expected, and the tests all pass on my machine. I don't think they'll pass on travis because of the cgo issues.

Seems like this would be ready to go when the last PRs are in.

btracey:
Sorry, this should have changed to gonum.org/v1/gonum. I'll fix that in the next attempt when the PRs are in.

PR #5, opened 2017-05-23
Mergerepos3

btracey:


kortschak:
That is an interesting quandary:

```
get "gonum.org/v1/gonum/blas": found meta tag main.metaImport{Prefix:"gonum.org/v1/", VCS:"git", RepoRoot:"https://github.com/gonum/gonum"} at https://gonum.org/v1/gonum/blas?go-get=1
```

I think we just have to accept that the PR fails because there is no code to pull.

btracey:
Yea, I was just looking into this. I don't know how travis works, but can we skirt this with the fact that "go get gonum.org/v1/gonum" should get the entire repository?

btracey:
Also, the current meta tag is `<meta name="go-import" content="gonum.org/v1/ git https://github.com/gonum/gonum">`

Should this really have `content="gonum.org/v1/gonum`?

kortschak:
According to the [docs](https://golang.org/cmd/go/), yes.

```
The meta tag has the form:

<meta name="go-import" content="import-prefix vcs repo-root">

The import-prefix is the import path corresponding to the repository root. It must be a prefix or an exact match of the package being fetched with "go get". If it's not an exact match, another http request is made at the prefix to verify the <meta> tags match. 
```

kortschak:
>Yea, I was just looking into this. I don't know how travis works, but can we skirt this with the fact that "go get gonum.org/v1/gonum" should get the entire repository?

Try adding a `git checkout mergerepos3` in the local repo on the builder. This will fail later, but it will at least show that this is working. Alternatively, we can just merge as it is and then do fixups.

kortschak:
I just checked locally, and that won't work. It invokes git to only get the default branch.

kortschak:
Sorry for the late bikeshedding. Did we decide on "gonum.org/v1/gonum" over "gonum.org/v1/pkg"?

btracey:
Ah, shoot. I think I undid that travis commit, but have kept deleting all of the (non-main) travis scripts.

Should we merge as is?

btracey:
I just looked through the discussion and from what I could tell I'm the only one to use either. I guess I have a slight preference for `gonum`, since then `v1/plot` and `v1/netlib` feel more in line. They aren't gonum proper, but they are packages (or something like that). The `v1` in the URL gives enough of a sense that it is code.

kortschak:
I agree. Can you fix that so we have "gonum.org/v1/..." and then push the big red button.

btracey:
Sorry, "fix that so we have"?

btracey:
Meaning in the meta-tag?

kortschak:
In the meta-tag and in the second to last commit which does "github.com/gonum" => "gonum.org/v1/gonum". This last should be "github.com/gonum" => "gonum.org/v1".

btracey:
Oh, so you're suggesting they'll be `gonum.org/v1/matrix`? I was thinking it would be `gonum.org/v1/gonum/matrix`, better mirroring the true git structure.

btracey:
Especially since we're planning to keep the old repos. Having `gonum.org/v1/plot` be `github.com/gonum/plot`, but `gonum.org/v1/matrix` is not `github.com/gonum/matrix` is confusing

kortschak:
See discussion starting from [here](https://groups.google.com/d/msg/gonum-dev/4oLmSvtDuZY/zVY2UcZgEAAJ). There is either a `/pkg/` or a nothing.

btracey:
Sorry, I thought the `gonum.org/v1` was resolving that particular discussion to "nothing". I thought this discussion is about if the `gonum` paths should have `gonum` after the `v1`. My impression is that the `gonum` repo will have versions change at the same time, while `plot` could be on a different schedule. It seems the templating would be easier if the directories match the underlying repositories

kortschak:
I think we can merge, but there is still confusion here (at least in my mind).

At the moment, the import paths are in the form "gonum.org/v1/gonum/...". The discussion I linked to talks about them having either "gonum.org/v1/..." or "gonum.org/v1/pkg/...".

btracey:
I agree with the confusion. The linked discussion is actually about `gonum.org/pkg/v1/...` vs. `gonum.org/v1/...`. It seemed like the consensus was that the combined `pkg/v1` would be redundant. I then thought we wanted it to mirror the repo structure, so, `gonum.org/v1/netlib/...`, `gonum.org/v1/gonum/...`, `gonum.org/v1/plot...`.

kortschak:
>I then thought we wanted it to mirror the repo structure, so, `gonum.org/v1/netlib/...`, `gonum.org/v1/gonum/...`, `gonum.org/v1/plot...`. 

I missed that last bit. Can you point to that?

kortschak:
OK. Gottit. That seems fair. I'll get over the stutter in time.

btracey:
Great. 

I just noticed that https://github.com/gonum/stat/pull/185 is still open. Is it possible you could review that? Then I'll create a mergerepo4, run the commits again, and then submit?

PR #7, opened 2017-05-23
gonum: clean away old files and fix-up travis

kortschak:
Partial fix for travis failures. The next step is to move cgo-dependent code out into netlib.

@btracey Please take a look.


btracey:
LGTM

PR #9, opened 2017-05-23
Add (most) canonical import paths.

btracey:
This adds canonical import paths to most of the gonum packages. It skips the packages with autogeneration, specifically Blas, Lapack, matrix/mat64, matrix/mat128, and most of the packages in graph/formats/dot/internal.


btracey:
@kortschak  Could you handle the `graph` canonical import issue? I figure BLAS, Lapack, and matrix can wait until we've fixed those package structures.

diff/fd/diff.go
```diff
@@ -3,7 +3,7 @@
 // license that can be found in the LICENSE file.
 
 // Package fd provides functions to approximate derivatives using finite differences.
-package fd
+package fd // import "golang.org/v1/gonum/diff/fd"
```
sbinet: `s/golang.org/gonum.org/g` I presume :)

kortschak: Ha!

btracey: Oh geez.

btracey:
By entire repo, do you mean all of `graph`, or `graph/internal`? Should I remove the current ones?

kortschak:
I mean from $GOPATH/src

```
for f in $(find gonum.org/v1/gonum/ -name "*.go"); do
	if grep -q "^package main$" $f; then echo skipping $f ...; continue; fi
	sed -i "s|^\(package .*\)$|\1 // import \"$(dirname $f)\"|" $f
done
```

btracey:
Oh, I misunderstood the spec. I thought it had to be in exactly one file. Do you want to do that and I'll close this one?

kortschak:
I've just pushed it.

PR #11, opened 2017-05-24
all: add canonical imports

kortschak:
Performed by:

```
for f in $(find gonum.org/v1/gonum/ -name "*.go"); do
	if grep -q "^package main$" $f; then echo skipping $f ...; continue; fi
	sed -i "s|^\(package .*\)$|\1 // import \"$(dirname $f)\"|" $f
done
```

then some manual fixes including checks of the code generation.

Locally passes `CGO_LDFLAGS="-L/usr/local -lopenblas" go test ./...` except for expected `Dgeev` error in OpenBLAS code.


kortschak:
I have left `blas/testblas/benchautogen/autogen_bench_level1double.go` as I don't know how that is invoked.

btracey:
In the future is the convention to add the canonical URL to all new files?

kortschak:
I think so.

If we had a `doc.go` in every directory we could have done it there, but we didn't.

btracey:
LGTM

kortschak:
I can redo this that way though: iterate over all package directories adding a `doc.go` if it doesn't exist and add the canonical path there, otherwise to the existing `doc.go`. Then we can slowly go over things and move the actual docs as needed.

kortschak:
I should not do this to `*_test.go` files at least. Fixing that now.

btracey:
That seems like a better idea.

The other thing to check if there is a `pkgname.go`, for example `stat` has stat.go

kortschak:
It probably is a better idea, but it is not trivial. After much hating of bash, I think it needs a Go parser to do correctly. I can have a look at that, but not now.

kortschak:
Abandoning in favour of #13.

PR #16, opened 2017-05-25
blas/lapack/internal: Remove cgo packages

btracey:
This removes the cgo packages from the main gonum library now that they have been moved to gonum.org/v1/netlib. Specifically, this removes blas/cgo, lapack/cgo, and internal/binding


btracey:
Thanks for mentioning, I didn't think about that issue. Updating.

btracey:
Updated the imports.

I added a shell implementation of native, so now it panics if the complex routines are called.

blas/cblas128/cblas128.go
```diff
@@ -7,12 +7,12 @@ package cblas128 // import "gonum.org/v1/gonum/blas/cblas128"
 
 import (
 	"gonum.org/v1/gonum/blas"
-	"gonum.org/v1/gonum/blas/cgo"
+	"gonum.org/v1/gonum/blas/native"
 )
 
 // TODO(kortschak): Change this and the comment below to native.Implementation
```
kortschak: We need to either remove this comment or make the implementation local to this package and call is something like `no`.

PR #20, opened 2017-05-26
travis: fix build

kortschak:
* Add completish set of tests (from "github.com/gonum/matrix").
* Fix single failure.

@btracey Please take a look.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,9 +4,9 @@ language: go
 
 # Versions of go that are explicitly supported by gonum.
 go:
- - 1.6
- - 1.7
- - 1.8
+ - 1.6.x
+ - 1.7.x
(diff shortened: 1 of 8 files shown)


btracey:
LGTM

kortschak:
We get three failures here for noasm/appengine.

- stat principle component analysis — error in lowest valued column
- lapack/native Dgeev — unexpected eigenvalues
- lapack/native Dgebrd — a, d, e and tauQ mismatches

I can skip those tests for noasm/appengine in the interim.

btracey:
Okay to skip until we can improve tests.

kortschak:
There is one set of codegen that is not checked - the dot parser (#24).

PR #28, opened 2017-05-26
matrix/mat64,stat: reverse matrix extraction

kortschak:
Port of gonum/matrix#436.

@btracey Please take a look.

### DIFF
--- a/matrix/mat64/gsvd.go
+++ b/matrix/mat64/gsvd.go
@@ -205,7 +205,8 @@ func (gsvd *GSVD) ValuesB(s []float64) []float64 {
 
 // ZeroRTo extracts the matrix [ 0 R ] from the singular value decomposition, storing
 // the result in-place into dst. [ 0 R ] is size (k+l)×c.
-func (gsvd *GSVD) ZeroRTo(dst *Dense) {
+// If dst is nil, a new matrix is allocated. The resulting ZeroR matrix is returned.
+func (gsvd *GSVD) ZeroRTo(dst *Dense) *Dense {
 	if gsvd.kind == 0 {
 		panic("gsvd: no decomposition computed")
 	}
@@ -214,7 +215,11 @@ func (gsvd *GSVD) ZeroRTo(dst *Dense) {
 	k := gsvd.k
 	l := gsvd.l
 	h := min(k+l, r)
-	dst.reuseAsZeroed(k+l, c)
+	if dst == nil {
+		dst = NewDense(k+l, c, nil)
+	} else {
+		dst.reuseAsZeroed(k+l, c)
+	}
 	a := Dense{
 		mat:     gsvd.a,
 		capRows: r,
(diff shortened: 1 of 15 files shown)


matrix/mat64/lq.go
```diff
@@ -58,10 +58,15 @@ func (lq *LQ) Factorize(a Matrix) {
 // TODO(btracey): Add in the "Reduced" forms for extracting the m×m orthogonal
 // and upper triangular matrices.
 
-// LFromLQ extracts the m×n lower trapezoidal matrix from a LQ decomposition.
-func (m *Dense) LFromLQ(lq *LQ) {
+// LTo extracts the m×n lower trapezoidal matrix from a LQ decomposition.
```
btracey: Do we need the `To` on the function calls? It's pretty clear from the function signature alone that it produces the matrix `L`, and very clear with the documentation. Seeing `x := lq.L(nil)` or `lq.L(x)` both feel well in line with our other signatures.

kortschak: Thanks for reminding me. Yes, we do - see comments https://github.com/gonum/matrix/pull/436#issuecomment-287593489.

I will go through the other instances that are like this and change them to match. For example https://github.com/gonum/matrix/pull/433#discussion_r107590374.

btracey: Thanks for reminding me. I still buy the argument.

kortschak:
Added GSVD and HOGSVD equivalents. PTAL

(also some missed doc errors in previous commits)

matrix/mat64/gsvd.go
```diff
@@ -214,112 +215,147 @@ func (m *Dense) ZeroRFromGSVD(gsvd *GSVD) {
...
-		m.set(i, i+k, 1)
+		dst.set(i, i+k, 1)
 	}
+	return dst
 }
 
-// UFromGSVD extracts the matrix U from the singular value decomposition, storing
-// the result in-place into the receiver. U is size r×r.
-func (m *Dense) UFromGSVD(gsvd *GSVD) {
+// UTo extracts the matrix U from the singular value decomposition, storing
```
kortschak: > The documentation says U is rxr, but this makes it rxc.

U is (A_r,A_c), so U_r = U_c = r.

> The HOGSVD types have a "VTo will panic if the receiver does not contain a successful factorization." Should the GSVD types as well?

Will do.

matrix/mat64/gsvd.go
```diff
@@ -214,112 +215,147 @@ func (m *Dense) ZeroRFromGSVD(gsvd *GSVD) {
...
 	}
-	m.Copy(tmp)
+	dst.Copy(tmp)
+	return dst
 }
 
-// VFromGSVD extracts the matrix V from the singular value decomposition, storing
-// the result in-place into the receiver. V is size p×p.
-func (m *Dense) VFromGSVD(gsvd *GSVD) {
+// VTo extracts the matrix V from the singular value decomposition, storing
```
kortschak: See above.

matrix/mat64/gsvd.go
```diff
@@ -214,112 +215,147 @@ func (m *Dense) ZeroRFromGSVD(gsvd *GSVD) {
...
 	}
-	m.Copy(tmp)
+	dst.Copy(tmp)
+	return dst
 }
 
-// QFromGSVD extracts the matrix Q from the singular value decomposition, storing
-// the result in-place into the receiver. Q is size c×c.
-func (m *Dense) QFromGSVD(gsvd *GSVD) {
+// QTo extracts the matrix Q from the singular value decomposition, storing
```
kortschak: See above.

PR #29, opened 2017-05-27
floats: adding AreUnique

zeroviscosity:
While discussing a PR (https://github.com/gonum/floats/pull/61) on the floats package before the merger, it was suggested that an AreUnique function would be useful.

### DIFF
--- a/floats/floats.go
+++ b/floats/floats.go
@@ -69,6 +69,20 @@ func AddScaledTo(dst, y []float64, alpha float64, s []float64) []float64 {
 	return dst
 }
 
+// AreUnique returns true if all elements of the slice are unique.
+// Elements are compared using normal float64 equality, so NaN values are
+// considered unique.
+func AreUnique(s []float64) bool {
+	m := make(map[float64]bool, len(s))
+	for _, val := range s {
+		if m[val] {
+			return false
+		}
+		m[val] = true
+	}
+	return true
+}
+
 // argsort is a helper that implements sort.Interface, as used by
 // Argsort.
 type argsort struct {
--- a/floats/floats_test.go
+++ b/floats/floats_test.go
@@ -133,6 +133,49 @@ func TestAddScaledTo(t *testing.T) {
 	}
 }
 
+func TestAreUnique(t *testing.T) {
+	for _, test := range []struct {
+		s    []float64
+		want bool
+	}{
+		{
+			s:    []float64{1, 2, 3, 4},
+			want: true,
+		},
+		{
+			s:    []float64{1, 2, 3, 4, 1},
+			want: false,
+		},
+		{
+			s:    []float64{1, 1, 1, 1},
+			want: false,
+		},
+		{
+			s: []float64{
+				0.3,
+				0.1 + 0.2,
+				3 / 10,
+			},
+			want: false,
+		},
+		{
+			s: []float64{
+				0,
+				math.SmallestNonzeroFloat64,
+				2 * math.SmallestNonzeroFloat64,
+				3 * math.SmallestNonzeroFloat64,
+				4 * math.SmallestNonzeroFloat64,
+			},
+			want: true,
+		},
+	} {
+		got := AreUnique(test.s)
+		if got != test.want {
+			t.Errorf("AreUnique returned incorrect value for %v, got %v, want %v", test.s, got, test.want)
+		}
+	}
+}
+
 func TestArgsort(t *testing.T) {
 	s := []float64{3, 4, 1, 7, 5}
 	inds := make([]int, len(s))


zeroviscosity:
Good point about `math.NaN()`. Rather than just a comment would checking for that be preferable?

Regarding implementation, I went with the map based off a comment by @btracey in the PR mentioned above. I can, however, look into some benchmarking.

floats/floats.go
```diff
@@ -69,6 +69,18 @@ func AddScaledTo(dst, y []float64, alpha float64, s []float64) []float64 {
 	return dst
 }
 
+// AreUnique returns true if all elements of the slice are unique.
+func AreUnique(s []float64) bool {
+	m := make(map[float64]bool)
```
kortschak: Give this a size hint: `m := make(map[float64]bool, len(s))`.

zeroviscosity: Will do.

kortschak:
There is no good answer for what to do with NaN. Some people want it to be different, some people want it to be the same. Do the simple thing and document it.

floats/floats.go
```diff
@@ -69,6 +69,19 @@ func AddScaledTo(dst, y []float64, alpha float64, s []float64) []float64 {
 	return dst
 }
 
+// AreUnique returns true if all elements of the slice are unique.
+// Note that NaN values are treated as different values.
```
kortschak: `// Elements are compared using normal float64 equality, so NaN values are considered unique.`

kortschak:
Please wait for at least one other reviewer.

zeroviscosity:
Thanks, @kortschak! Shall I squash in the mean time?

kortschak:
Sure.

btracey:
In my case, I've pretty much only needed this function when the data was sorted (or was going to be sorted). If that argument is (mostly) true, then it seems like demanding sorting and not allocating is better. Furthermore, at that point, we shouldn't test for NaN similarity because one doesn't know where they'll be. This has the upside of being very simple to document (if s[i] == s[i+1] for any i)

kortschak:
By way of comparison, this is the benchmarking I have done now with assumption of sorted input.

```
// AreUniqueSorted returns true if all elements of the slice are unique.
func AreUniqueSorted(s []float64) bool {
        if len(s) < 2 {
                return true
        }
        for i, v := range s[1:] {
                if s[i] == v {
                        return false
                }
                if s[i] > v {
                        panic("floats: input slice not sorted")
                }
        }
        return true
```

Results:
```
BenchmarkAreUnique/1000_elements-8         	   10000	    166811 ns/op	   55517 B/op	      96 allocs/op
BenchmarkAreUnique/100000_elements-8       	     100	  15690459 ns/op	 3645913 B/op	    6224 allocs/op
BenchmarkAreUnique/10000000_elements-8     	       1	2656867203 ns/op	444970960 B/op	  551688 allocs/op
BenchmarkAreUniqueSorted/1000_elements-8   	 2000000	       758 ns/op	       0 B/op	       0 allocs/op
BenchmarkAreUniqueSorted/100000_elements-8 	   20000	     74656 ns/op	       0 B/op	       0 allocs/op
BenchmarkAreUniqueSorted/10000000_elements-8         200	   8127642 ns/op	       0 B/op	       0 allocs/op
```

zeroviscosity:
What about providing both options, either via separate functions or via an `isSorted` parameter? That way, if someone does want to check for uniqueness on a large unsorted slice, they have the option of compromising allocations for speed.

kortschak:
I don't like the idea of a parameterised function for this and having near equivalent functions seems excessive. Do you have known use cases when an unsorted slice is likely to be a target?

zeroviscosity:
Offhand I don't know of any such use cases, although I would think there must be some. However, if that seems rare enough that it doesn't justify the map-based approach, I can swap out the current implementation for the `AreUniqueSorted` implementation above. Should I go ahead with that?

kortschak:
Please do. The comment will need to be changed. I would suggest

```
// AreUnique returns true if all elements of the sorted slice s are unique.
// Uniqueness is defined as s[i] != s[i+1] for all possible values of i.
// Are Unique will panic if s is not sorted according to sort.Float64sAreSorted.
```

At the moment we are being strict on sort order, though that is not strictly speaking necessary. We could conceivably determine the sort order from the first comparison and panic if subsequent changes do not agree. This at least prevents a user with descending order needing to reverse the slice for this test. @btracey?

I'm not quite happy with the NaN caveat wording (it needs to be there in some form - many people sort NaN after for sort ascending, or always after). There must be tests for cases with NaN (my implementation above was incorrect because of the various ways NaN is sorted).

```
// AreUnique returns true if all elements of the sorted slice s are unique.
// Uniqueness is defined as s[i] != s[i+1] for all possible values of i.
// AreUnique may panic if s is not monotonically changing or NaN is sorted
// greater than a number.
func AreUnique(s []float64) bool {
        if len(s) < 2 {
                return true
        }
        if s[0] < s[1] {
                for i, v := range s[1:] {
                        if s[i] == v {
                                return false
                        }
                        if s[i] > v || (!math.IsNaN(s[i]) && math.IsNaN(v)) {
                                panic("floats: input slice not sorted")
                        }
                }
                return true
        }
        for i, v := range s[1:] {
                if s[i] == v {
                        return false
                }
                if s[i] < v || (math.IsNaN(s[i]) && !math.IsNaN(v)) {
                        panic("floats: input slice not sorted")
                }
        }
        return true
}
```

btracey:
We force the data to be ascending in other functions (`stat.Hist`). It's not so harmful to let this one function be either direction, but I don't want a convention where we have double implementations for every function that wants a sorted slice. It seems more consistent to always force sorted ascending.

@kortschak your wording and implementation don't match. For floating point numbers, `a == b` is not the same as `!(a != b)`. Personally, though slightly more confusing, I think we should use the second definition. I can't think of a use of this functionality where I want it to return true with `NaN` (unlike `Equal` where both cases are plausible).

 This gives us something like

````
// AreUnique returns true if all elements of the sorted slice s are unique.
// Non-uniqueness is defined as s[i] != s[j] for all possible values of i and j
// (so the presence of NaN values causes AreUnique to return false).
// AreUnique may panic if s is not sorted ascending.
func AreUnique(s []float64) bool {
        if len(s) < 2 {
                if len(s) == 1 && math.IsNaN(s[0]){
                         return false
                }
                return true
        }
        for i, v := range s[1:] {
               if s[i] < v {
                     panic("floats: input slice not sorted")
               }
               if math.IsNaN(v) || s[i] == v {
                        return false
               }
        }
        if math.IsNaN(s[len(s)-1]) {
                  return false
         }
        return true
}
````

kortschak:
Yes the wording is incorrect. I don't think this is the same as `stat.Histogram` or other places where it is unlikely to be in an inner loop. This is much more likely to be there. I don't like have to split paths, but without it I think I'm in the situation where I don't think it is work adding.

btracey:
Alright. I'm okay with both paths, I just don't want this to be a precedence that everything has to support both paths.

kortschak:
>Personally, though slightly more confusing, I think we should use the second definition.

I'm not convinced by this. If I have `[]float64{math.NaN(), 1,2,3}` I would expect that to return as unique. I would however expect that `[]float64{math.NaN(), math.NaN(), 1,2,3}` returns false.

btracey:
I'm okay with that behavior also. I don't think it is ever desired for `[]float64{math.NaN(), math.NaN(), 1,2,3}` to return true.

Do we know if this `NaN` sorting is guaranteed by the sort package? It's not in the documentation, and they've changed undocumented behavior in the past. One could also imagine a NaN unstable sort that just uses `<` or something. If it is guaranteed, then it's fine to enforce that behavior here.

btracey:
Opened https://github.com/golang/go/issues/20540

btracey:
Okay, now the NaN sorting order is guaranteed, so I'm okay with requiring the NaNs to be first.

btracey:
Ping?

btracey:
I'm closing this because it's been months.

PR #30, opened 2017-05-27
graph/simple: factor ID handling out of graph types

kortschak:
This is preparation for adding `ID() int` to the `graph.Edge` interface.

@vladimir-ch Please take a look.

### DIFF
--- a/graph/simple/simple.go
+++ b/graph/simple/simple.go
@@ -9,6 +9,8 @@ package simple // import "gonum.org/v1/gonum/graph/simple"
 import (
 	"math"
 
+	"golang.org/x/tools/container/intsets"
+
 	"gonum.org/v1/gonum/graph"
 )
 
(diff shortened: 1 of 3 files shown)


graph/simple/simple.go
```diff
@@ -43,3 +45,37 @@ const maxInt int = int(^uint(0) >> 1)
 func isSame(a, b float64) bool {
 	return a == b || (math.IsNaN(a) && math.IsNaN(b))
 }
+
+// idset implements available ID storage.
```
vladimir-ch: `idSet`

graph/simple/simple.go
```diff
@@ -43,3 +45,37 @@ const maxInt int = int(^uint(0) >> 1)
...
+type idSet struct {
+	used, free intsets.Sparse
+}
+
+// newID returns a new unique ID.
+func (s *idSet) newID() int {
+	var id int
+	if s.free.Len() != 0 && s.free.TakeMin(&id) {
```
vladimir-ch: This can be simplified to just `if s.free.TakeMin(&id) {` but on the other hand it is less explicit.

graph/simple/simple.go
kortschak: Done.

graph/simple/simple.go
kortschak: I think this is residual from when the behaviour was map based. I've made the change, but it looks like we may end up back at the map-based approach anyway.

Issue #32, opened 2017-05-30
graph/simple: node ID addition is slow


kortschak:
See discussion at https://groups.google.com/d/topic/gonum-dev/twUcG1rEMIU/discussion

Also note that this impacts on decisions regarding the integer type used for node IDs in #31.

kortschak:
Closed by #34.

@leoneu please reopen if you are still having problems with this.

PR #34, opened 2017-05-31
graph/simple: do not use intsets.Sparse for ID sets

kortschak:
This does not go the complete way to the kind of implementation referred to in the gonum-dev [thread](https://groups.google.com/d/topic/gonum-dev/twUcG1rEMIU/discussion), but covers a reasonable set of use cases. If we make an `IDSource` interface to use, this would be a good default.

@leoneu Would you please try this in your problematic code and see if it is satisfactory.

@leoneu @vladimir-ch Please take a look.

Updates #32.

### DIFF
--- a/graph/simple/directed.go
+++ b/graph/simple/directed.go
@@ -31,6 +31,8 @@ func NewDirectedGraph(self, absent float64) *DirectedGraph {
 
 		self:   self,
 		absent: absent,
+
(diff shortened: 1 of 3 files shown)


kortschak:
@leoneu *ping*

kortschak:
Moving ahead.

PR #35, opened 2017-05-31
all: run gofmt -s

kortschak:
Purely mechanical change.

@gonum/developers Please take a look.


sbinet:
would it make sense to add a test at the root of the `gonum/gonum` repo to make sure `gofmt/goimport` is run?
(and possibly add a `gofmt -s` pass?)

kortschak:
The repo already does the test for gofmt. The second commit in this PR adds a warning about files that would be simplified (I had it failable, but that seems harsh).

PR #41, opened 2017-06-01
matrix/mat64: remove lint found by megacheck

vladimir-ch:


matrix/mat64/shadow.go
```diff
@@ -208,7 +208,7 @@ func rectanglesOverlap(off, aCols, bCols, stride int) bool {
 
 	// Flatten the shifted matrix column positions
 	// so a starts at 0, modulo the common stride.
-	const aFrom = 0
+	// const aFrom = 0
```
sbinet: why just not remove it altogether?

kortschak: I should have left the const in the code with the hope the compiler optimised it away. It will take time to figure out where it actually sits, so maybe just removing it is OK.

matrix/mat64/triangular.go
```diff
@@ -418,8 +418,8 @@ func copySymIntoTriangle(t *TriDense, s Symmetric) {
 				for j := i; j < n; j++ {
 					t.mat.Data[i*ts+j] = sd.Data[j*ss+i]
 				}
-				return
 			}
+			return
```
kortschak: This is a semantic change. It needs a test.

matrix/mat64/shadow.go
vladimir-ch: The motivation to just comment it out was documentation. I expected that it would trigger some comments that would clarify what to do with this. I will remove it.

kortschak: The part of the comment above "a starts at 0" matches the const. I think the reason I left it out of the logic is that it would have gone in the condition below as `return aFrom <= bTo && bFrom < aTo` (or something like that) which would not have been elided by the compiler.

matrix/mat64/triangular.go
vladimir-ch: Done. The test fails on master, passes with this change. But since RawSymmetricer has data stored in the upper triangle, the bad code path would never be taken.

vladimir-ch:
Addressed both comments. PTAL

PR #43, opened 2017-06-02
stat/spatial: new package for spatial statistic measures

kortschak:
@btracey @vladimir-ch @Armadilloa16 @ReubenBuck Please take a look.

This is commit one of ~~two~~ three. The ~~second~~ final commit will add element-wise weights, after the API is otherwise settled.

Possible considerations are adding a reset method to allow data/locality reuse.

### DIFF
--- a/stat/spatial/spatial.go
+++ b/stat/spatial/spatial.go
@@ -1,0 +1,132 @@
+// Copyright ©2017 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package spatial // import "gonum.org/v1/gonum/stat/spatial"
+
+import (
+	"math"
+
+	"gonum.org/v1/gonum/mat"
+	"gonum.org/v1/gonum/stat"
+)
+
+// TODO(kortschak): Implement weighted routines.
+// TODO(kortschak): Make use of banded matrices when they exist in mat.
+
+// GetisOrdGStar returns the Local Getis-Ord G*i statistic for element of of the
+// weighted data using the provided locality matrix. The returned value is a z-score.
+//
+//  G^*_i = num_i / den_i
+//
+//  num_i = \sum_j (w_{ij} x_j) - \bar X \sum_j w_{ij}
+//  den_i = S \sqrt(((n \sum_j w_{ij}^2 - (\sum_j w_{ij})^2))/(n - 1))
+//  \bar X = (\sum_j x_j) / n
+//  S = \sqrt((\sum_j x_j^2)/n - (\bar X)^2)
+//
+// GetisOrdGStar will panic if locality is not a square matrix with dimensions the
+// same as the length of data or if i is not a valid index into data.
+//
+// See doi.org/10.1111%2Fj.1538-4632.1995.tb00912.x.
+//
+// Weighted Getis-Ord G*i is not currently implemented and GetisOrdGStar will
+// panic if weights is not nil.
+func GetisOrdGStar(i int, data, weights []float64, locality mat.Matrix) float64 {
+	if weights != nil {
+		panic("spatial: weighted data not yet implemented")
+	}
+	r, c := locality.Dims()
+	if r != len(data) || c != len(data) {
+		panic("spatial: data length mismatch")
+	}
+
+	n := float64(len(data))
+	mean, std := stat.MeanStdDev(data, weights)
+	var dwd, dww, sw float64
+	for j, v := range data {
+		w := locality.At(i, j)
+		sw += w
+		dwd += w * v
+		dww += w * w
+	}
+	s := std * math.Sqrt((n-1)/n)
+
+	return (dwd - mean*sw) / (s * math.Sqrt((n*dww-sw*sw)/(n-1)))
+}
+
+// GlobalMoransI performs Global Moran's I calculation of spatial autocorrelation
+// for the given data using the provided locality matrix. GlobalMoransI returns
+// Moran's I, Var(I) and the z-score associated with those values.
+// GlobalMoransI will panic if locality is not a square matrix with dimensions the
+// same as the length of data.
+//
+// See https://doi.org/10.1111%2Fj.1538-4632.2007.00708.x.
+//
+// Weighted Global Moran's I is not currently implemented and GlobalMoransI will
+// panic if weights is not nil.
+func GlobalMoransI(data, weights []float64, locality mat.Matrix) (i, v, z float64) {
+	if weights != nil {
+		panic("spatial: weighted data not yet implemented")
+	}
+	if r, c := locality.Dims(); r != len(data) || c != len(data) {
+		panic("spatial: data length mismatch")
+	}
+	mean := stat.Mean(data, nil)
+
+	// Calculate Moran's I for the data.
+	var num, den, sum float64
+	for i, xi := range data {
+		zi := xi - mean
+		den += zi * zi
+		for j, xj := range data {
+			w := locality.At(i, j)
+			sum += w
+			zj := xj - mean
+			num += w * zi * zj
+		}
+	}
+	i = (float64(len(data)) / sum) * (num / den)
+
+	// Calculate Moran's E(I) for the data.
+	e := -1 / float64(len(data)-1)
+
+	// Calculate Moran's Var(I) for the data.
+	//  http://pro.arcgis.com/en/pro-app/tool-reference/spatial-statistics/h-how-spatial-autocorrelation-moran-s-i-spatial-st.htm
+	//  http://pro.arcgis.com/en/pro-app/tool-reference/spatial-statistics/h-global-morans-i-additional-math.htm
+	var s0, s1, s2 float64
+	var var2, var4 float64
+	for i, v := range data {
+		v -= mean
+		v *= v
+		var2 += v
+		var4 += v * v
+
+		var p2 float64
+		for j := range data {
+			wij := locality.At(i, j)
+			wji := locality.At(j, i)
+
+			s0 += wij
+
+			v := wij + wji
+			s1 += v * v
+
+			p2 += v
+		}
+		s2 += p2 * p2
+	}
+	s1 *= 0.5
+
+	n := float64(len(data))
+	a := n * ((n*n-3*n+3)*s1 - n*s2 + 3*s0*s0)
+	c := (n - 1) * (n - 2) * (n - 3) * s0 * s0
+	d := var4 / (var2 * var2)
+	b := d * ((n*n-n)*s1 - 2*n*s2 + 6*s0*s0)
+
+	v = (a-b)/c - e*e
+
+	// Calculate z-score associated with Moran's I for the data.
+	z = (i - e) / math.Sqrt(v)
+
+	return i, v, z
+}
--- a/stat/spatial/spatial_areal_example_test.go
+++ b/stat/spatial/spatial_areal_example_test.go
@@ -1,0 +1,74 @@
+// Copyright ©2017 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package spatial_test
+
+import (
+	"fmt"
+	"math"
+
+	"gonum.org/v1/gonum/floats"
+	"gonum.org/v1/gonum/mat"
+	"gonum.org/v1/gonum/stat/spatial"
+)
+
+// Euclid is a mat.Matrix whose elements refects the Euclidean
+// distance between a series of unit-separated points strided
+// to be arranged in an x by y grid.
+type Euclid struct{ x, y int }
+
+func (e Euclid) Dims() (r, c int) { return e.x * e.y, e.x * e.y }
+func (e Euclid) At(i, j int) float64 {
+	d := e.x * e.y
+	if i < 0 || d <= i || j < 0 || d <= j {
+		panic("bounds error")
+	}
+	if i == j {
+		return 0
+	}
+	x := float64(j%e.x - i%e.x)
+	y := float64(j/e.x - i/e.x)
+	return 1 / math.Hypot(x, y)
+}
+func (e Euclid) T() mat.Matrix { return mat.Transpose{e} }
+
+func ExampleGlobalMoransI_areal() {
+	locality := Euclid{10, 10}
+
+	data1 := []float64{
+		1, 0, 0, 1, 0, 0, 1, 0, 0, 0,
+		0, 1, 1, 0, 0, 1, 0, 0, 0, 0,
+		1, 0, 0, 1, 0, 0, 0, 0, 1, 0,
+		0, 0, 1, 0, 1, 0, 1, 0, 0, 0,
+		1, 0, 0, 0, 0, 0, 0, 1, 0, 0,
+		0, 0, 0, 0, 1, 0, 0, 0, 0, 0,
+		0, 0, 1, 0, 0, 0, 1, 0, 1, 0,
+		1, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+		0, 0, 1, 0, 1, 0, 1, 0, 0, 0,
+		1, 0, 0, 0, 0, 0, 0, 0, 1, 0,
+	}
+	i1, _, z1 := spatial.GlobalMoransI(data1, nil, locality)
+
+	data2 := []float64{
+		0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+		0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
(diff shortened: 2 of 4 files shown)


btracey:
Sorry, I've been thinking about this. Should it instead be under `gonum/stat/spatial`? It seems like there are a lot of ways one could do spatial statistics, and so it might be better to group those under one heading. Similarly, if these are going to gain bells and whistles, it might make more sense to have them in their own thing rather than stat where all the functions are one-off.

kortschak:
Thanks. I was thinking the same thing - yes there are many spatial statistics. The move is trivial; it will happen shortly.

What do you think about the API? Should I make it a resetable value?

btracey:
Do you have any nice references, especially for GetisOrd? It feels like these must have different names from some other field, but I'm not sure what they are. 

I'm not really sure. In the things I've done, typically there's one field where you want a bunch of different weight measures. Also though, usually there is some very fine discretization over the field, so you wouldn't want an explicit matrix. If it's common for you to have many different fields with the same weight matrix then that suggests there should be some reset value.

btracey:
In other words, these implementations aren't focused around an interface, but it feels like there's an abstract problem with a nice interface to build around.

kortschak:
Getis-Ord Gi* is from from Getis, A., and J. K. Ord (1992). "The Analysis of Spatial Association by Use of Distance Statistics.” Geographical Analysis 24 (July), 189-206. There is a rough description [on the arcgis website](http://pro.arcgis.com/en/pro-app/tool-reference/spatial-statistics/h-how-hot-spot-analysis-getis-ord-gi-spatial-stati.htm).

I'll add the reset - two actually; one for the data and one for the locality matrix.

Armadilloa16:
Interesting! I don't think I have anything meaningful to say beyond that though...

kortschak:
@btracey PTAL

I think the API is OK - I have separated the reset methods out because with one method the semantics of `mat64.Matrix(nil)` versus `mat64.Matrix((*mat64.Dense)(nil))` become difficult. Also with the addition of weights, `nil` is a valid change-to value and explaining that in the docs would be harder than is worthwhile.

I'm still thinking about how to add in weighting and the API may have to change when I have sorted that out.

btracey:
This feels over-engineered to me. `GetisOrd` has one productive method, `Gstar`. Instead, can't we just have 
````
// GetisOrd computes the Local Getis-Ord G* statistic for the elements in idxs
func GetisOrd(idxs []int,data []float64, locality mat64.Matrix) []float64 {}
````

`Moran` is a bit more difficult, but it feels like we could take a similar approach as with `stat.StdErr`. There is an expensive function to compute `I` (and E is trivial), and then a second function to compute `Var`. Something like

````
// MoranI computes Moran 's I  measure of spatial autocorrelation
func MoranI(data []float64, locality mat64.Matrix) float64 {}

// MoranIVar computes Moran 's I  measure of spatial autocorrelation and the variance
func MoranIVar(data []float64, locality mat64.Matrix) (i, var float64){}

// ... where n is the length of data.
func MoranZ(i, var float64, n int) float64 {}
````

**Comment by kortschak** [comment]
I don't like the Getis-Ord signature (should be `GetisOrdGstari` since there is a global Getis-Ord index); I don't think we have ever taken a set of indices to calculate over.

For the Moran's global I, if it's going to be simplified, then `MoransGlobal(data []float64, locality *mat64.Dense) (i, var, z float64)`. Note that locality must be a `*mat64.Dense` since we would otherwise have to make a copy.

**Comment by btracey** [comment]
> Note that locality must be a *mat64.Dense since we would otherwise have to make a copy.

Why? The (simple version of the) function can be 
````
func MoransGlobal(data []float64, locality mat64.Matrix) (i float64) {
    mean, std := stat.Mean(data)
    ss := std * float64(len(data) - 1)  // remove population correction
    N := float64(len(data))
    r, c := locality.Dims()
    var num, W float64
    for i := 0; i < r; i++ {
       xi := data[i] - mean
       for j := 0; j < c; j++ {
           xj := data[j] - mean
            w := locality.At(i,j)
            W += w
            num += w * xi * xj
       }
   }
   return (N/W) * num
}
````

**Comment by btracey** [comment]
> I don't think we have ever taken a set of indices to calculate over.

I don't understand. The example case ranges over all of the indices and prints the answer for each one. Presumably the intended usage is generally to take a set of indices and calculate over it.

**Comment by kortschak** [comment]
> Why? The (simple version of the) function can be

Yes, sorry. That's reasonable.

> I don't understand. The example case ranges over all of the indices and prints the answer for each one. Presumably the intended usage is generally to take a set of indices and calculate over it.

There are two main uses, testing for elements that are in significant hotspots or using subregions to do the same. I don't like the set of indices being required to do the former.

If we don't keep the data and locality matrix as copies, the cost reduction that is obtained by the additional methods goes away for Getis-Ord and we just need `NewGetisOrd`, `(*GetisOrd) Reset`, `Len` and `Gstar`.

**Comment by btracey** [comment]
My main complaint is that in the current PR `GetisOrd` effectively has one productive method, but has a lot of scaffolding around it. It feels like it should be a function, and not a type.

It looks like the computation is reasonably trivial for each index. Maybe `GetisOrdGstarIdx` and  `GetisOrdGstarAll`, where the `Idx` version computes it for one index, and `All` computes the value for every Index in data? That should support both cases reasonably cleanly I would think.

**Comment by kortschak** [comment]
The justification for a type is not how many mathods, but whether there is retained state. Whether that is justified comes down to how expensive the state is to generate. Looking at the breakdown, the thin majority of the computation occurs in the construction with 2 iterations over the data slice and one copy of the locality (thus n - needed to get a vector for use as `wi`) going into the construction (out of a total of 5+n). I'll make it a function, but I think this may want to be revisited if performance becomes an issue here.

**Comment by kortschak** [comment]
What about a GetisOrd type with only `Len()` and `Gstar` methods. This gives O(n) time for the element calculate and has the minimal set of methods.

**Comment by btracey** [comment]
I think this still uses n^2 memory and time unnecessarily. 

Here would be my implementation (based on this formula http://resources.esri.com/help/9.3/arcgisengine/java/gp_toolref/spatial_statistics_tools/how_hot_spot_analysis_colon_getis_ord_gi_star_spatial_statistics_works.htm)
````
// GStarS computes the effective standard deviation for the given locality matrix
func GStarI(idx int, data []float64, locality mat.Matrix) {
	r, c := locality.Dims()
	if r != len(data) || c != len(data) {
		panic("spatial: data length mismatch")
	}

	mean, varX := stat.MeanVariance(data, nil)

	// Correct because GetisOrd uses the second moment rather than the sample
	n := float64(len(data))
	varX *= (n - 1) / n

	var sw, ss, sx float64
	for j, x := range data {
		w := locality.At(i, j)
		sw += w
		ss += w * w
		sx += w * x
	}
	varW := math.Sqrt((n*ss - sw*sw) / (n - 1))
	return (sx - mean*sw) / math.Sqrt(varX*varW)
}
````

This is a linear algorithm, unlike NewGetisOrd which needs to save the locality matrix (so is n^2). There is nothing to cache about the locality matrix, since the computation for each index only needs the elements in that column. The only work repeated by multiple calls to GetisOrdI is computing the mean and standard deviation of the `data` slice, which is maybe saving a tenth of the cost.

**Comment by kortschak** [comment]
That code disagrees with the code I have, but I will work on getting something like it that is correct.

**Comment by kortschak** [comment]
PTAL

**Comment by reconditematter** [comment]
@btracey @kortschak You may have already seen this, the original paper is available at[http://archive.nefmc.org/tech/cte_mtg_docs/130516/CATT%20Report/Getis-Ord%20statistic.pdf](url)

**Comment by btracey** [comment]
I'm happy with the API. Would you like me to check the Moran's I equations?

I'll go through the tests in more detail tomorrow.

**Comment by kortschak** [comment]
I have compared the equations to an R implementation, but it's probably worth going over. Thanks.

File: stat/spatial/spatial_test.go [hunk 1]
```diff
@@ -0,0 +1,188 @@
...
+
+		nseg := getisOrdSegments(data, nil, locality)
+		if nseg != test.wantSegs {
+			t.Errorf("unexpected number of significant segments for test %d: got:%d want:%d",
+				ti, nseg, test.wantSegs)
+		}
+	}
+}
+
+func getisOrdSegments(data, weight []float64, locality mat.Matrix) int {
```
btracey: Can you comment this? I don't really understand why this is a test of GetisOrd

stat/spatial/spatial.go
```diff
@@ -0,0 +1,131 @@
...
+//  S = \sqrt((\sum_j x_j^2)/n - (\bar X)^2)
+//
+// GetisOrdGStar will panic if locality is not a square matrix with dimensions the
+// same as the length of data or if i is not a valid index into data.
+//
+// See doi.org/10.1111%2Fj.1538-4632.1995.tb00912.x.
+//
+// Weighted Getis-Ord G*i is not currently implemented and GetisOrdGStar will
+// panic if weights is not nil.
+func GetisOrdGStar(i int, data, weights []float64, locality mat.Matrix) float64 {
```
kortschak: I want to be able to zero weight out NA data. That cannot be done via the locality matrix.

stat/spatial/spatial.go
```diff
@@ -0,0 +1,131 @@
...
+// for the given data using the provided locality matrix. GlobalMoransI returns
+// Moran's I, Var(I) and the z-score associated with those values.
+// GlobalMoransI will panic if locality is not a square matrix with dimensions the
+// same as the length of data.
+//
+// See https://doi.org/10.1111%2Fj.1538-4632.2007.00708.x.
+//
+// Weighted Global Moran's I is not currently implemented and GlobalMoransI will
+// panic if weights is not nil.
+func GlobalMoransI(data, weights []float64, locality mat.Matrix) (i, v, z float64) {
```
kortschak: As for G*i.

stat/spatial/spatial.go
```diff
@@ -0,0 +1,131 @@
...
+	var num, den float64
+	for i, xi := range data {
+		zi := xi - mean
+		den += zi * zi
+		for j, xj := range data {
+			zj := xj - mean
+			num += locality.At(i, j) * zi * zj
+		}
+	}
+	i = (float64(len(data)) * num) / (mat.Sum(locality) * den)
```
kortschak: Readability?

stat/spatial/spatial_areal_example_test.go
```diff
@@ -0,0 +1,71 @@
...
+import (
+	"fmt"
+	"math"
+
+	"gonum.org/v1/gonum/floats"
+	"gonum.org/v1/gonum/mat"
+	"gonum.org/v1/gonum/stat/spatial"
+)
+
+type Euclid struct{ x, y int }
```
kortschak: Done.

stat/spatial/spatial.go
```diff
@@ -0,0 +1,131 @@
...
+		for j := range data {
+			wij := locality.At(i, j)
+			wji := locality.At(j, i)
+
+			s0 += wij
+
+			v := wij + wji
+			s1 += v * v
+
+			p2 += wij + wji
```
kortschak: Done.

stat/spatial/spatial_test.go
kortschak: Done.

stat/spatial/spatial_areal_example_test.go
```diff
@@ -0,0 +1,71 @@
...
+		0, 0, 1, 1, 1, 1, 0, 0, 0, 0,
+		0, 0, 0, 1, 1, 1, 0, 0, 0, 0,
+		0, 0, 0, 0, 1, 0, 0, 0, 1, 0,
+		0, 0, 0, 0, 0, 0, 0, 1, 1, 1,
+		0, 0, 0, 0, 0, 0, 0, 1, 1, 1,
+		0, 0, 0, 0, 0, 0, 0, 1, 1, 1,
+	}
+	i2, _, z2 := spatial.GlobalMoransI(data2, nil, locality)
+
+	fmt.Printf("%v sparse points Moran's I=%.4v z-score=%.4v\n", floats.Sum(data1), i1, z1)
```
kortschak: Done.

stat/spatial/spatial_example_test.go
```diff
@@ -0,0 +1,69 @@
...
+import (
+	"fmt"
+
+	"gonum.org/v1/gonum/mat"
+	"gonum.org/v1/gonum/stat/spatial"
+)
+
+func ExampleGlobalMoransI_1() {
+	data := []float64{0, 0, 0, 1, 1, 1, 0, 1, 0, 0}
+	locality := mat.NewDense(10, 10, []float64{
```
kortschak: Done.

stat/spatial/spatial_test.go
```diff
@@ -0,0 +1,188 @@
...
+	wantSegs   int
+}{
+	{
+		from: 0, to: 1, n: 1000, wide: 1,
+		fn: func(_ float64, _ int, rnd *rand.Rand) float64 {
+			return rnd.Float64()
+		},
+		locality: simpleAdjacency,
+
+		wantMoranI: -0.0019631298955953233,
```
kortschak: Done.

stat/spatial/spatial.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+	i = (float64(len(data)) * num) / (mat.Sum(locality) * den)
+
+	// Calculate Moran's E(I) for the data.
+	e := -1 / float64(len(data)-1)
+
+	// Calculate Moran's Var(I) for the data.
+	//  http://pro.arcgis.com/en/pro-app/tool-reference/spatial-statistics/h-how-spatial-autocorrelation-moran-s-i-spatial-st.htm
+	//  http://pro.arcgis.com/en/pro-app/tool-reference/spatial-statistics/h-global-morans-i-additional-math.htm
+	var s0, s1, s2 float64
```
kortschak: These loops are separated from clarity.

stat/spatial/spatial.go
```diff
@@ -0,0 +1,131 @@
...
+	if weights != nil {
+		panic("spatial: weighted data not yet implemented")
+	}
+	r, c := locality.Dims()
+	if r != len(data) || c != len(data) {
+		panic("spatial: data length mismatch")
+	}
+
+	n := float64(len(data))
+	mean := stat.Mean(data, weights)
```
kortschak: I tried that (both with your code and with modifications of what I have here and they are not the same).

stat/spatial/spatial.go
```diff
@@ -0,0 +1,131 @@
...
+// TODO(kortschak): Implement weighted routines.
+// TODO(kortschak): Make use of banded matrices when they exist in mat.
+
+// GetisOrdGStar returns the Local Getis-Ord G*i statistic for element of of the
+// weighted data using the provided locality matrix. The returned value is a z-score.
+//
+//  G^*_i = num_i / den_i
+//
+//  num_i = \sum_j (w_{ij} x_j) - \bar X \sum_j w_{ij}
+//  den_i = S \sqrt((n \sum_j w_{ij}^2 - (\sum_j w_{ij})^2)/(n - 1))
```
kortschak: Thanks. Parens!

stat/spatial/spatial.go
btracey: Got it.

stat/spatial/spatial.go
btracey: But you use `mean`, so the results are already co-mingled.

stat/spatial/spatial.go
btracey: Still, it seems like it would be better to save N^2 time?

stat/spatial/spatial.go
btracey: If I do
````
mean, std := stat.MeanStdDev(data, nil)
s := std * math.Sqrt((n-1)/n)
````
delete `ss` from the loop, and remove the `s` assigment below the tests pass for me.

kortschak: OK. I don't know what I did then.

Done.

stat/spatial/spatial.go
kortschak: Sorry, the question was about "I'd do (N/W)*(num/den)". Do you want that change for readability?

btracey: Yes. I also think it'll be better numerically, since `den` is already the sum of squares, which isn't the best.

stat/spatial/spatial.go
kortschak: The mean is calculated outside the I loop, then there's the I loop and then there is the Var(I) loop.

stat/spatial/spatial.go
kortschak: Done.

btracey:
LGTM. Thanks.

PR #45, opened 2017-06-02
lapack/{native,testlapack}: remove some lint found by megacheck

vladimir-ch:
I left some warnings in.

Most importantly this one:
`native/dlasq4.go:74:6: this value of b2 is never used (SA4006)`
Looking at the code it's not immediately clear whether it is just a harmless oversight, so I submitted an issue https://github.com/Reference-LAPACK/lapack/issues/158

`native/general.go:42:2: const badLdA is unused (U1000)`: badLdA is checked in `netlib`. This can be handled together with https://github.com/gonum/lapack/issues/230

`testlapack/fortran.go:27:6: func printFortranArray is unused (U1000)` can be potentially useful in the future. Similarly for `testlapack/general.go:24:2: const dlamchP is unused (U1000)`.


lapack/native/dlasq5.go
```diff
@@ -28,7 +28,8 @@ func (impl Implementation) Dlasq5(i0, n0 int, z []float64, pp int, tau, sigma fl
 		emin = z[j4+4]
 		d := z[j4] - tau
 		dmin = d
-		dmin1 = -z[j4]
+		// In the reference there are code paths that actually return this value.
```
kortschak: What are the netlib code paths that return this?

lapack/native/general.go
```diff
@@ -57,12 +57,10 @@ const (
 	badVn2          = "lapack: vn2 has insufficient length"
 	badUplo         = "lapack: illegal triangle"
 	badWork         = "lapack: insufficient working memory"
-	badWorkStride   = "lapack: insufficient working array stride"
```
kortschak: This needs to be reflected in the netlib repository. Please make an issue if it doesn't already match.

lapack/native/dlasq5.go
vladimir-ch: In the `else` branches for non-IEEE arithmetic there is a return before `dmin1` is set again.

PR #50, opened 2017-06-05
floats: add ParseWithNA

### LINKED ISSUE #44: proposal: add ParseFloat64(s string) (v, w float64, err error)
We provide weighted value support for tests etc in stat, but there is no handling for textual representation of "NA" provided by stdlib. So I propose we add in either stat or floats a `ParseFloat` or `Parse` respectively that handles this.

Edited to incorporate suggestion below to allow a user-provided na string.

Assume in floats.
```
// Parse converts the string s to a float64 in v. If s equals na, w is returned as 0, otherwise 1.
func Parse(s, na string) (v, w float64, err error) {
    if s == na {
        return 0, 0, nil
    }
    v, err = strconv.ParseFloat(s, 64)
    if err == nil {
        w = 1
    }
    return v, w, err
}
```

This would then be used like so, e.g. single column data:
```
var vals, weights []float64
sc := bufio.NewScanner(r)
for sc.Scan() {
    v, w, err := floats.Parse(sc.Text(), "NA")
    if err != nil { ... }
    vals = append(vals, v)
    weights = append(weights, w)
}
err := sc.Err() ...
mean := stat.Mean(vals, weights)
```

kortschak:
@btracey Please take a look.

Closes #44.

### DIFF
--- a/floats/floats.go
+++ b/floats/floats.go
@@ -14,6 +14,7 @@ import (
 	"errors"
 	"math"
 	"sort"
+	"strconv"
 
 	"gonum.org/v1/gonum/internal/asm/f64"
 )
@@ -565,6 +566,19 @@ func Norm(s []float64, L float64) float64 {
 		norm += math.Pow(math.Abs(val), L)
 	}
 	return math.Pow(norm, 1/L)
+}
+
+// ParseWithNA converts the string s to a float64 in v.
+// If s equals missing, w is returned as 0, otherwise 1.
+func ParseWithNA(s, missing string) (v, w float64, err error) {
+	if s == missing {
+		return 0, 0, nil
+	}
+	v, err = strconv.ParseFloat(s, 64)
+	if err == nil {
+		w = 1
+	}
+	return v, w, err
 }
 
 // Prod returns the product of the elements of the slice.
--- a/floats/parse_example_test.go
+++ b/floats/parse_example_test.go
@@ -1,0 +1,43 @@
+// Copyright ©2017 The gonum Authors. All rights reserved.
+// Use of this code is governed by a BSD-style
+// license that can be found in the LICENSE file
+
+package floats_test
+
+import (
+	"bufio"
+	"fmt"
+	"log"
+	"strings"
+
+	"gonum.org/v1/gonum/floats"
+	"gonum.org/v1/gonum/stat"
+)
+
+func ExampleParseWithNA() {
+	// Calculate the mean of a list of numbers
+	// ignoring missing values.
+	const data = `6
+missing
+4
+`
+
+	var vals, weights []float64
+	sc := bufio.NewScanner(strings.NewReader(data))
+	for sc.Scan() {
+		v, w, err := floats.ParseWithNA(sc.Text(), "missing")
+		if err != nil {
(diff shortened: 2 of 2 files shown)


floats/floats.go
```diff
@@ -567,6 +568,19 @@ func Norm(s []float64, L float64) float64 {
 	return math.Pow(norm, 1/L)
 }
 
+// ParseWithNA converts the string s to a float64 in v.
+// If s equals na, w is returned as 0, otherwise 1.
```
btracey: Can we somehow use the phrase "not available" into the documentation? I feel the purpose isn't very clear, especially for non-native speakers.

floats/parse_example_test.go
```diff
@@ -0,0 +1,41 @@
...
+	"gonum.org/v1/gonum/floats"
+	"gonum.org/v1/gonum/stat"
+)
+
+func ExampleParseWithNA() {
+	const data = `6
+NA
+4
+`
+
```
btracey: It would be nice if this had some comment like 
"Compute the mean of a list of numbers, ignoring missing entries"

kortschak:
PTAL

floats/floats.go
```diff
@@ -567,6 +568,19 @@ func Norm(s []float64, L float64) float64 {
 	return math.Pow(norm, 1/L)
 }
 
+// ParseWithNA converts the string s to a float64 in v.
```
btracey: Should this be ParseWithMissing ? NA is shorter, but missing reads much better.

kortschak: I was thinking about that and I think the shortness should win here. It's almost never the case that the word "missing" is used, but "NA" is very common. I'll wait for other input.

btracey:
@dwhitena

floats/floats.go
vladimir-ch: Recently I read an article about NaN and NA. I had to look up what NA means (but on the other hand I never processed any external data). Now I would side with ParseWithNA.

kortschak:
Was just checking out how R handles NA and I find this:

```
typedef union
{
    double value;
    unsigned int word[2];
} ieee_double;
```

```
static double R_ValueOfNA(void)
{
    volatile ieee_double x;
    x.word[hw] = 0x7ff00000;
    x.word[lw] = 1954;
    return x.value;
}
```

So they are using a specific `NaN` ([`0x7ff80000000007a2`](https://play.golang.org/p/7ntbMNCD5v)) as NA. According to [this](http://adv-r.had.co.nz/C-interface.html), the 1954 comes from one of the original R authors' birthdays. This approach is more space efficient than what we have but would involve significant rewriting of routines for them to be NA-aware.

btracey:
It's a clever idea, but I think it'll be hard to operationalize throughout the suite. Ignoring NA values (by setting 0 weight) is one reasonable thing to do, but for many problems that's not the correct thing to do. Maybe NA values arise because the monitoring period was only so long, and so NA really means `> 5 weeks` .  I see handling NA values as part of the steps between loading values and performing computation. It seems to me to be better to force that handling step early, rather than having every `gonum` routine have a set of complicated actions and options around NA .

PR #53, opened 2017-06-06
dense: add SliceRows method

wrwr:
I've noticed that mat64.Dense doesn't have a method for slicing specified rows. In my opinion it would be very useful (at least for me ;>) so I've created a simple implementation. What do you think? Does it have any sense?

### DIFF
--- a/mat64/dense.go
+++ b/mat64/dense.go
@@ -5,6 +5,8 @@
 package mat64
 
 import (
+	"sort"
+
 	"github.com/gonum/blas"
 	"github.com/gonum/blas/blas64"
 	"github.com/gonum/matrix"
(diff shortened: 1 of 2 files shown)


btracey:
For a real slice, we have https://godoc.org/gonum.org/v1/gonum/matrix/mat64#Dense.Slice

I assume from the code you mean a permutation, along the lines of https://github.com/gonum/gonum/blob/master/matrix/mat64/lu.go#L251 ?

btracey:
Oh, no, you don't mean that.

If we were to accept something along these lines, it shouldn't sort the indices. []int{2,0} should return a new matrix with those rows as is.

Could you give an example of why you need this? I know it's used a lot in languages like matlab where vectorization is important, but I typically find in Go code that this kind of step is unnecessary.

wrwr:
> Could you give an example of why you need this? I know it's used a lot in languages like matlab where vectorization is important, but I typically find in Go code that this kind of step is unnecessary.

Generally I need it for filtering. I iterate over matrix rows, extract some 'special' rows indices, create matrix with these rows, make some operation on the new matrix, iterate over rows, extract another indices etc.

kortschak:
I don't think this meets the threshold for general utility in mat64. But if it did, `RowView`/`RawRowView` would be the appropriate ways to access the elements, or slightly less efficiently with `lapack64.Lapmt`.

btracey:
The permutation functionality in Dlapmt isn't what is being requested. The request is something like `func (m *Dense) RowsFrom(a Matrix, inds []int)` which would have implementation like

````
func (m *Dense) RowsFrom(a Matrix, inds []int){
    // preamble stuff about size and overlap and all that
    tmp := make([]float64, c)
    for i, v := range inds {
          m.SetRow(i, mat64.Row(tmp, a, v))
    }
}
````

kortschak:
Yes, that is more efficient, but `Dlapmt` followed by a slice of the matrix does the work as well.

kortschak:
What is the state of this? I will close in about a week if there is no input from @wrwr.

PR #54, opened 2017-06-07
matrix/mat64: use pools for []float64 and []int where possible

kortschak:
@btracey Please take a look.


matrix/mat64/pool.go
```diff
@@ -81,6 +87,12 @@ func init() {
 				Data: make([]float64, l),
 			}}
 		}
+		poolFloats[i].New = func() interface{} {
+			return make([]float64, l)
+		}
+		poolInts[i].New = func() interface{} {
+			return make([]int, l)
```
btracey: Not your fault, but I don't like how close "l" and "1" are in the github font.

kortschak: Yeah. I am on ubuntu which has generally good fonts for this, so I don't notice.

PR #58, opened 2017-06-08
graph/encoding/dot: allow unmarshaling of global attributes

### LINKED ISSUE #57: graph/encoding/dot: Unmarshal does not know about graph-level DOT attribute blocks
The DOT language [definition](http://www.graphviz.org/doc/info/lang.html) allows graph, node and edge attribute statements, and the `Marshal` function in dot allows these to be rendered into a .dot file from a graph that is aware of them (the `Attributers` interface provides this hook). We should be able to read what we can write.

So we need something that can be hooked (it looks like `*generator.addStmt` in the `*ast.AttrStmt` type case is the right place). ~~To simplify this, it's probably worth simplifying the `Attributers` interface - I'll look into that part now.~~ Looked into it and it makes no real difference for additional code churn.

kortschak:
Fixes #57.

@mewmew Please take a look.

Working on this has prompted some thinking. There is a discordance between the interfaces provided for marshaling versus unmarshaling. It would be nice if we could harmonise them - this would be API breaking.

### DIFF
--- a/graph/encoding/dot/decode.go
+++ b/graph/encoding/dot/decode.go
@@ -25,6 +25,13 @@ type Builder interface {
 	NewEdge(from, to graph.Node) graph.Edge
 }
 
+// UnmashalerAttrs is implemented by graph values that can unmarshal global
+// DOT attributes.
+type UnmarshalerAttrs interface {
+	// DOTUnmarshalerAttrs returns the global attribute unmarshalers.
+	DOTUnmarshalerAttrs() (graph, node, edge UnmarshalerAttr)
+}
+
 // UnmarshalerAttr is implemented by types that can unmarshal a DOT
 // attribute description of themselves.
 type UnmarshalerAttr interface {
@@ -60,6 +67,9 @@ func copyGraph(dst Builder, src *ast.Graph) (err error) {
 		directed: src.Directed,
 		ids:      make(map[string]graph.Node),
 	}
+	if a, ok := dst.(UnmarshalerAttrs); ok {
+		gen.graphAttr, gen.nodeAttr, gen.edgeAttr = a.DOTUnmarshalerAttrs()
+	}
 	for _, stmt := range src.Stmts {
 		gen.addStmt(dst, stmt)
 	}
@@ -79,6 +89,8 @@ type generator struct {
 	// Stack of start indices into the subgraph node slice. The top element
 	// corresponds to the start index of the active (or inner-most) subgraph.
 	subStart []int
+	// graphAttr, nodeAttr and edgeAttr are global graph attributes.
+	graphAttr, nodeAttr, edgeAttr UnmarshalerAttr
 }
 
 // node returns the gonum node corresponding to the given dot AST node ID,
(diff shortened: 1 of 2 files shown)


graph/encoding/dot/decode.go
```diff
@@ -118,7 +130,40 @@ func (gen *generator) addStmt(dst Builder, stmt ast.Stmt) {
 	case *ast.EdgeStmt:
 		gen.addEdgeStmt(dst, stmt)
 	case *ast.AttrStmt:
-		// ignore.
+		var n UnmarshalerAttr
+		var ok bool
+		var dst string
+		switch stmt.Kind {
+		case ast.KindGraph:
+			n, ok = gen.graphAttr.(UnmarshalerAttr)
```
mewmew: Mostly of a personal preference, but I would check for `nil` rather than type asserting on the `UnmarshalerAttr` interface. Likewise for `nodeAttr` and `edgeAttr`.

For the current code, I had to [consult the playground](https://play.golang.org/p/ChPMnJqLnh) to persuade myself that the current code was correct, since I thought that any value which may be assigned to `graphAttr, nodeAttr, edgeAttr UnmarshalerAttr` must implement the `UnmarshalerAttr` interface to be a valid assignment, and thus all type assertions would yield true. As the playground confirmed, the exception for this is of course the `nil` value.

graph/encoding/dot/decode_test.go
```diff
@@ -105,12 +130,23 @@ func (g *dotDirectedGraph) NewEdge(from, to graph.Node) graph.Edge {
 	return e
 }
 
+// DOTAttributers implement the dot.Attributers interface.
+func (g *dotDirectedGraph) DOTAttributers() (graph, node, edge Attributer) {
+	return g.graph, g.node, g.edge
+}
+
+// DOTAttributers implement the dot.UnmarshalerAttrs interface.
```
mewmew:
```diff
-// DOTAttributers implement the dot.UnmarshalerAttrs interface.
+-// DOTUnmarshalerAttrs implement the dot.UnmarshalerAttrs interface.
```

graph/encoding/dot/decode_test.go
```diff
@@ -137,6 +173,16 @@ func (g *dotUndirectedGraph) NewEdge(from, to graph.Node) graph.Edge {
 	return e
 }
 
+// DOTAttributers implement the dot.Attributers interface.
+func (g *dotUndirectedGraph) DOTAttributers() (graph, node, edge Attributer) {
+	return g.graph, g.node, g.edge
+}
+
+// DOTAttributers implement the dot.UnmarshalerAttrs interface.
```
mewmew:
```diff
-// DOTAttributers implement the dot.UnmarshalerAttrs interface.
+-// DOTUnmarshalerAttrs implement the dot.UnmarshalerAttrs interface.
```

mewmew:
LGTM after the feedback.

Good job!

mewmew:
> Working on this has prompted some thinking. There is a discordance between the interfaces provided for marshaling versus unmarshaling. It would be nice if we could harmonise them - this would be API breaking.

I also felt that the naming conventions used for marshalling and unmarshalling could be made more consistent. Do you have any drafts for how you would wish to change the API to harmonise them?
Feel free to open a dedicated issue for this so we may track the discussion of such API changes.

graph/encoding/dot/decode.go
kortschak: TBH I have no idea what I was thinking here.

Done.

graph/encoding/dot/decode_test.go
kortschak: Done.

graph/encoding/dot/decode_test.go
kortschak: Done.

PR #65, opened 2017-06-12
matrix, all: combine matrix packages, change matrix to mat

btracey:
This changes the matrix packages to the new format, and makes the necessary changes to other packages so the test suite passes. Specifically it:
- Deletes matrix/cov/conv.go, matrix/gendoc.go, and matrix/generate.go
- Renames gonum/matrix to gonum/mat
- Moves all of the files in gonum/mat/mat64 into gonum/mat
- Updates the documentation in doc.go to reflect the combined package
- Fixes all of the package names and imports in the new gonum/mat
- Updates the Formatting tests in mat that required a mat64 literal to now require a mat literal
- Moves mat/cmat128/matrix.go to mat/cmatrix.go and updates the type names
- Fixes all of the other gonum package names to import mat instead of mat64
- Fixes a view name clashes between the package mat and the variable name mat

Updates #4.


kortschak:
I'd like this to be in a number of commits, at least for the PR review.

btracey:
How would you like them? It doesn't pass at any stage in the commit, so there's not a lot to check in the interim unfortunately. 

Also unfortunately, it's not particularly easy to automate most of the steps. You can't just find and replace `matrix.` because there are a lot of comments that end in that string, meaning matrix. Similarly, you can't just skip the comments, because there are a bunch of places where `matrix.` refers to the package. It's thus context sensitive, and so quite hard to automate. I did most of the change by hand.

btracey:
An easy way to partition is to have 1) All of the matrix changes except doc.go 2) The doc.go changes 3) All of the package fixes.

btracey:
Faliure is because I missed the -noasm file. I will fix after I understand how you would like me to partition the changes so I can still "revert last commit".

kortschak:
The bullet points you have in the PR description look pretty good as a set of commits.

btracey:
Okay, re-organized into 4 commits. Delete the duplicate files, merge all into matrix and rename, fix all of the other packages, update documentation file.

matrix/conv/conv.go
```diff
@@ -1,142 +0,0 @@
-// Copyright ©2015 The gonum Authors. All rights reserved.
```
btracey: It could be. I was thinking at least we would have `CDense.FromReal(m Matrix)`

btracey:
I don't like the extra naming at all. I'm just one data point, but real matrices are a huge fraction (probably 90+%) of the matrices used. We're at something like 80% of the numpy functionality, and the only place we've needed them is eigenvalues. It's justified to have nice names for the common case, and slightly less nice names for the uncommon case.

btracey:
Should I merge the commits?

kortschak:
I'm happy for you to do that, though it's probably worth waiting for at least one other.

btracey:
@vladimir-ch ?  it's a doozy, but also the big one

vladimir-ch:
Give me a moment.

mat/doc.go
```diff
@@ -42,30 +38,35 @@
...
 //  func (m *Dense) Mul(a, b Matrix)
 // constructs a *Dense from the result of a multiplication with any Matrix types,
 // not just *Dense. Where more restrictive requirements must be met, there are also the
 // Symmetric and Triangular interfaces. For example, in
 //  func (s *SymDense) AddSym(a, b Symmetric)
 // the Symmetric interface guarantees a symmetric result.
 //
-// Transposes
+// The CMatrix interface plays the same role for complex matrices. The difference
+// is that the CMatrix type has the H method instead T, for returning the Hermitian
```
vladimir-ch: instead of

Also, since the related type is called Conjugate, what about "for return the conjugate (or Hermitian) transpose"?

mat/doc.go
```diff
@@ -42,30 +38,35 @@
...
+// instead of the raw transpose.
```
vladimir-ch: I would avoid using "raw" here. What about "ordinary transpose"?

mat/doc.go
```diff
@@ -42,30 +38,35 @@
...
+//
+// Transposes / Hermitians
```
vladimir-ch: Hermitians looks strange here. I would use the full "Transposes and conjugate transposes". We should probably settle either on conjugate or Hermitian in the documentation.

mat/doc.go
```diff
@@ -42,30 +38,35 @@
...
 //
-// The T method is used for transposition. For example, c.Mul(a.T(), b) computes
-// c = a^T * b. The mat64 types implement this method using an implicit transpose —
-// see the Transpose type for more details. Note that some operations have a
-// transpose as part of their definition, as in *SymDense.SymOuterK.
+// The T method is used for transposition on real matrices, and H is used for the
+// Hermitian on complex matrices. For example, c.Mul(a.T(), b) computes
```
vladimir-ch: missing "transposition"

mat/doc.go
btracey: Done.

mat/doc.go
btracey: I just cut out that phrase

mat/doc.go
btracey: switched to conjugate.

mat/doc.go
btracey: Done.

mat/doc.go
```diff
@@ -42,30 +38,35 @@
...
 // Symmetric and Triangular interfaces. For example, in
 //  func (s *SymDense) AddSym(a, b Symmetric)
 // the Symmetric interface guarantees a symmetric result.
 //
-// Transposes
+// The CMatrix interface plays the same role for complex matrices. The difference
+// is that the CMatrix type has the H method instead T, for returning the conjugate
+// transpose.
+//
+//  (Conjugate) Transposes
```
vladimir-ch: It seems there are two leading spaces.

mat/symmetric.go
```diff
@@ -483,12 +482,12 @@ func (s *SymDense) PowPSD(a Symmetric, pow float64) error {
...
 	ok := eigen.Factorize(a, true)
 	if !ok {
-		return matrix.ErrFailedEigen
+		return ErrFailedEigen
 	}
 	values := eigen.Values(nil)
 	for i, v := range values {
 		if v <= 0 {
-			return matrix.ErrNotPSD
+			return ErrNotPSD
```
vladimir-ch: Unrelated to this PR but now that I see it: positive symmetric definite looks weird, positive belongs to definite. I would put symmetric first: symmetric positive definite (or SPD).

Issue #31, opened 2017-05-29
graph: [proposal] use node ID for retrieval operations and graph.Node for creation


mewmew:
Background discussion: https://github.com/gonum/graph/issues/87#issuecomment-299039592

The intention of this proposal is to refine the graph API to mitigate subtle issues related to node retrieval from edges (original issue reported in gonum/graph#87), and to clearly communicate intent through the API; specifically that retrieval operations should only require access to a node ID, while creation operations require the entire data structure implementing the `graph.Node` interface.

Since this is an extensive update of the API, we should take the opportunity to simultaneously update the return type of `NodeID` from `int` to `int64`, to allow for future uses in huge graphs.

A rough sketch of the intended API changes is presented below. Please feel free to discuss and further improve upon it.

```diff
 // Node is a graph node.
 type Node interface {
 	// ID returns a graph-unique integer ID of the graph node.
-	ID() int
+	ID() int64
 }

 // Graph is a generalized graph.
 type Graph interface {
 	// Has reports whether the given node ID exists within the graph.
-	Has(Node) bool
+	Has(id int64) bool
 
 	// Nodes returns all the nodes in the graph.
 	Nodes() []Node
 
 	// From returns all nodes that can be reached directly
 	// from the given node ID.
-	From(Node) []Node
+	From(id int64) []Node
 
 	// HasEdgeBeteen reports whether an edge exists between the
 	// node IDs xid and yid without considering direction.
-	HasEdgeBetween(x, y Node) bool
+	HasEdgeBetween(xid, yid int64) bool
 
 	// Edge returns the edge from the node ID uid to vid if such an
 	// edge exists and nil otherwise. The node ID vid must be directly
 	// reachable from node ID uid as defined by the From method.
-	Edge(u, v Node) Edge
+	Edge(uid, vid int64) Edge
 }

 // Undirected is an undirected graph.
 type Undirected interface {
 	Graph
 
 	// EdgeBetween returns the edge between the node IDs xid and yid.
-	EdgeBetween(x, y Node) Edge
+	EdgeBetween(xid, yid int64) Edge
 }
 
 // Directed is a directed graph.
 type Directed interface {
 	Graph
 
 	// HasEdgeFromTo reports whether an edge exists
 	// in the graph from node ID uid to vid.
-	HasEdgeFromTo(u, v Node) bool
+	HasEdgeFromTo(uid, vid int64) bool
 
 	// To returns all nodes that can reach directly
 	// to the given node ID.
[...]

PR #70, opened 2017-06-16
graph/...: use int64 IDs

### LINKED ISSUE #69: graph/community: reduced graph Has methods are incorrect
They use the disjunction instead of the conjunction.

This will be fixed when I go through and do the change from `int` to `int64` IDs.

kortschak:
Two commits that deserve careful review are "graph/topo: update for int64 IDs" (the Bron-Kerbosch routine) and "graph/community: update for int64 IDs" which is a little subtle.

After #68.

Updates #31.
Fixes #69.
Updates gonum/graph#197.
Closes gonum/graph#114.

### DIFF
--- a/graph/community/louvain_directed.go
+++ b/graph/community/louvain_directed.go
@@ -29,7 +29,7 @@ func qDirected(g graph.Directed, communities [][]graph.Node, resolution float64)
 	// Calculate the total edge weight of the graph
 	// and the table of penetrating edge weight sums.
 	var m float64
-	k := make(map[int]directedWeights, len(nodes))
+	k := make(map[int64]directedWeights, len(nodes))
 	for _, n := range nodes {
 		var wOut float64
 		u := n
@@ -191,7 +191,7 @@ func reduceDirected(g graph.Directed, communities [][]graph.Node) *ReducedDirect
 			},
 			communities: communities,
 		}
-		communityOf := make(map[int]int, len(nodes))
+		communityOf := make(map[int64]int, len(nodes))
 		for i, n := range nodes {
 			r.nodes[i] = community{id: i, nodes: []graph.Node{n}}
 			communityOf[n.ID()] = i
@@ -257,7 +257,7 @@ func reduceDirected(g graph.Directed, communities [][]graph.Node) *ReducedDirect
 		r.parent = g
 	}
 	weight := positiveWeightFuncFor(g)
-	communityOf := make(map[int]int, commNodes)
+	communityOf := make(map[int64]int, commNodes)
 	for i, comm := range communities {
 		r.nodes[i] = community{id: i, nodes: comm}
 		for _, n := range comm {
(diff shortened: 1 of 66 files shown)


graph/community/louvain_directed.go
```diff
@@ -353,25 +352,25 @@ func (g *ReducedDirected) To(v graph.Node) []graph.Node {
 func (g *ReducedDirected) HasEdgeBetween(x, y graph.Node) bool {
 	xid := x.ID()
 	yid := y.ID()
-	if xid == yid {
+	if xid == yid || xid != int64(int(xid)) || yid != int64(int(yid)) {
```
mewmew: It seems to be rather involved to keep the community node as an `int` rather than an `int64`. May I ask, what is the main rationale for keeping it `int` than updating it to be consistent with node ID as `int64`, as is done elsewhere. This change will inevitably introduce API breakage. Might as well do one swooping change to `int64`, and keep the code simple and consistent?

kortschak: The communities are internal representations with dense IDs stored in memory; this means that an `int` is large enough. It has additional benefits performance-wise in that we are dealing with machine word-sized values on all archs instead of pushing `int64`s through 386 and arm. These are the people we are making the `int`->`int64` ID change for. In addition, the code changes required for this approach are far less wide-reaching than the straight out mechanical change.

mewmew: I agree on an intuitive level that keeping the IDs in a machine word-sized value may have performance benefits for 32-bit platforms.

At the same time, without doing a benchmark on code using communities, we won't know for sure if keeping the node ID an `int` or changing it to `int64` actually has a huge impact. There may be other parts of the code which dominate the CPU profiles, such that the ID being in `int` or `int64` may not make a difference performance wise.

From my point of view, if the performance difference as measured by benchmarks and profiling is negligible, then focusing on readability and maintainability is more important.

Still, if there is an actually difference in measured performance when changing from `int` to `int64`, then I agree that it may be worth while to keep the internal community node IDs as `int`s.

kortschak: There are three things here. One is that the changes associated with moving from `int` to `int64` are not at all clearly cleaner or more readable (this was my first motivation for limiting the change - I did actually go through and do the complete transformation first and decided it was not the way to go) and the second is that there is an immediate ~10% improvement in performance available as I can partially revert the removal of intsets, which clearly did show an improvement in this context. The last is safety, community graphs are held in slices, so the appropriate ID type is an `int` internally.

I will, however, add an `isValidCommunityID` to replace the `x = int64(int(x))` checks - adding sign checks which I don't yet have.

kortschak:
@mewmew *ping*

mewmew:
> The last is safety, community graphs are held in slices, so the appropriate ID type is an int internally.

I didn't think about this, but I guess you are right. Since slices are using int `len` and `cap` fields, it would be safer not to convert back and forth between `int64` to `int` for these.

I have looked through the code roughly, and from my perspective it looks good. I also trust your judgement on making these decisions as you have more experience with working with these graphs, and the environments that may have specific requirements or constraints (e.g. 32-bit environments). My main motivation was to make the graph API more future proof, before it would be finalized. Didn't really think about the more tricky aspects of the change, as you definitely seem to have done.

I wish I could give it a more thorough review of the change, but I believe someone with experience of using the topo and community packages would be better suited for finding errors (other than trivial ones that I may locate).

That being said, I know I'll play around a lot with topo this autumn, especially when experimenting with connected components in graphs, so at that point, I'll make sure to report anything I may bump into that could be strange.

So, at least from me you have a LGTM to merge.

@vladimir-ch perhaps you may wish to take a quick look before the merge?

kortschak:
*ping*

kortschak:
Thanks.

PR #71, opened 2017-06-16
mat: s/mat64/mat/g

kortschak:
Noticed some missed changes in strings and comment when I updated the godoc registry.

@btracey Please take a look.

### DIFF
--- a/mat/vector.go
+++ b/mat/vector.go
@@ -23,7 +23,7 @@ type Vector struct {
 	mat blas64.Vector
 	n   int
 	// A BLAS vector can have a negative increment, but allowing this
-	// in the mat64 type complicates a lot of code, and doesn't gain anything.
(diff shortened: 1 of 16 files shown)


mat/vector.go
```diff
@@ -23,7 +23,7 @@ type Vector struct {
 	mat blas64.Vector
 	n   int
 	// A BLAS vector can have a negative increment, but allowing this
-	// in the mat64 type complicates a lot of code, and doesn't gain anything.
+	// in the mat type complicates a lot of code, and doesn't gain anything.
```
sbinet: perhaps name it `mat.Vector` otherwise it could be (a bit) confusing with the `mat` field, above ?

kortschak: It says "mat type", so I think it's clear.

btracey:
Thanks

PR #80, opened 2017-06-19
graph/topo: add undirected cycles routine

kortschak:
@mewmew @vladimir-ch Please take a look.

Depends on #70.

Fixes gonum/graph#197.

### DIFF
--- a/graph/topo/paton_cycles.go
+++ b/graph/topo/paton_cycles.go
@@ -1,0 +1,86 @@
+// Copyright ©2017 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package topo
+
+import (
+	"gonum.org/v1/gonum/graph"
+	"gonum.org/v1/gonum/graph/internal/linear"
+	"gonum.org/v1/gonum/graph/internal/set"
+)
+
+// UndirectedCyclesIn returns the set of cycle bases in the graph g.
+// All cycles in g can be constructed from the returned cycle bases
+// by combining cycle bases such that presence of edges shared by the
+// bases are xored together.
+func UndirectedCyclesIn(g graph.Undirected) [][]graph.Node {
+	// From "An algorithm for finding a fundamental set of cycles of a graph"
+	// https://doi.org/10.1145/363219.363232
+
+	var cycles [][]graph.Node
+	done := make(set.Int64s)
+	var tree linear.NodeStack
+	for _, n := range g.Nodes() {
+		id := n.ID()
+		if done.Has(id) {
+			continue
+		}
+		done.Add(id)
+
+		tree = tree[:0]
+		tree.Push(n)
+		from := sets{id: set.Int64s{}}
+		to := map[int64]graph.Node{id: n}
+
+		for tree.Len() != 0 {
+			u := tree.Pop()
+			uid := u.ID()
+			adj := from[uid]
+			for _, v := range g.From(u) {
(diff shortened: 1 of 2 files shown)


mewmew:
Glad to see you are implementing these kind of algorithms. Truthfully, I'm still in the learning process of graph theory, and feel I have yet some way to go before being able to review this with confidence. Thus leaving this one for @vladimir-ch.

This autumn I'll attend a course in graph theory at Uni, which should hopefully put me in a better position for the future regarding these kind of algorithms.

kortschak:
@vladimir-ch *ping* when you have time.

graph/topo/paton_cycles.go
```diff
@@ -0,0 +1,86 @@
...
+
+package topo
+
+import (
+	"gonum.org/v1/gonum/graph"
+	"gonum.org/v1/gonum/graph/internal/linear"
+	"gonum.org/v1/gonum/graph/internal/set"
+)
+
+// UndirectedCyclesIn returns the set of cycle bases in the graph g.
```
vladimir-ch: For me the basis would be the set itself, and the cycles would then be called basis cycles. I would maybe change the comment to
```
UndirectedCyclesIn returns a set of cycles that forms a cycle basis in the graph g.
Any cycle in g can be constructed as a symmetric difference of its elements.
```

graph/topo/paton_cycles_test.go
```diff
@@ -0,0 +1,131 @@
...
+	"gonum.org/v1/gonum/graph/internal/ordered"
+	"gonum.org/v1/gonum/graph/simple"
+)
+
+var undirectedCyclesInTests = []struct {
+	g    []intset
+	want [][][]int64
+}{
+	{
+		g: []intset{
```
vladimir-ch: Would it be worth adding to this graph another component that has a cycle? Like:
```
10: linksTo(11, 12),
11: linksTo(12),
12: nil,
```

graph/topo/paton_cycles_test.go
```diff
@@ -0,0 +1,131 @@
...
+				break
+			}
+		}
+		if !matched {
+			t.Errorf("unexpected paton result for %d:\n\tgot:%#v\n\twant from:%#v", i, got, test.want)
+		}
+	}
+}
+
+func canonicalise(c []graph.Node) []graph.Node {
```
kortschak: Done.

graph/topo/paton_cycles.go
```diff
@@ -0,0 +1,86 @@
...
+
+		for tree.Len() != 0 {
+			u := tree.Pop()
+			uid := u.ID()
+			adj := from[uid]
+			for _, v := range g.From(u) {
+				vid := v.ID()
+				switch {
+				case uid == vid:
+					panic("topo: found self loop")
```
kortschak: This is removed since we now allow user graphs to contain self-loops.

graph/topo/paton_cycles.go
kortschak: Done.

graph/topo/paton_cycles_test.go
kortschak: Done.

kortschak:
PTAL to re-approve.

kortschak:
Probably I should add a `Xor(cycles ...[]graph.Node) [][]graph.Node`/`SymDiff(cycles ...[]graph.Node) [][]graph.Node` function to allow people to reconstruct any given set of cycles. This would be a new PR.

PR #88, opened 2017-06-20
mat: extend Vector slicing sematics to allow slicing up to capacity

### LINKED ISSUE #76: mat: extend Vector slicing sematics to allow slicing up to capacity
This follows from changes to `*Dense` and `*SymDense` in #73.

kortschak:
Please take a look.

Fixes #76.

### DIFF
--- a/mat/vector.go
+++ b/mat/vector.go
@@ -49,10 +49,10 @@ func NewVector(n int, data []float64) *Vector {
 
 // SliceVec returns a new Vector that shares backing data with the receiver.
 // The returned matrix starts at i of the receiver and extends k-i elements.
-// SliceVec panics with ErrIndexOutOfRange if the slice is outside the bounds
+// SliceVec panics with ErrIndexOutOfRange if the slice is outside the capacity
 // of the receiver.
 func (v *Vector) SliceVec(i, k int) *Vector {
-	if i < 0 || k <= i || v.n < k {
+	if i < 0 || k <= i || v.Cap() < k {
 		panic(ErrIndexOutOfRange)
 	}
 	return &Vector{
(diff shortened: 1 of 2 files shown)


mat/vector_test.go
```diff
@@ -51,6 +51,40 @@ func TestNewVector(t *testing.T) {
 	}
 }
 
+func TestCap(t *testing.T) {
+	for i, test := range []struct {
+		vector *Vector
+		want   int
+	}{
+		{vector: NewVector(3, nil), want: 3},
+		{
```
btracey: I forget, do we guarantee that the last element in the vector is valid? If not, could you add a test for all of the remainders? That is, check that 10, 11 and 12 all give a capacity of 4?

kortschak: We don't guarantee that, but that is our behaviour.

Done.

mat/vector.go
```diff
@@ -72,11 +72,24 @@ func (v *Vector) Dims() (r, c int) {
 	return v.n, 1
 }
 
+// Caps returns the number of rows and columns in the backing matrix. Columns is always 1.
+func (v *Vector) Caps() (r, c int) {
+	if v.isZero() {
```
kortschak: Oops. This should not be like this. Removing.

mat/vector.go
```diff
@@ -72,11 +72,21 @@ func (v *Vector) Dims() (r, c int) {
...
+}
+
 // Len returns the length of the vector.
 func (v *Vector) Len() int {
 	return v.n
 }
 
+// Cap returns the capacity of the vector.
+func (v *Vector) Cap() int {
+	return (cap(v.mat.Data)-1)/v.mat.Inc + 1
```
btracey: Won't this panic if vector is zero? Integer divide by zero?

kortschak: Good catch. And we can't know what a future inc will be, so returning 0 is appropriate.

The documentation is tricky. I'm really leaning towards the inclusion of IsZero methods as per gonum/matrix#396 even if only because it allows us to then define a usable zero state formally.

btracey: Yea, I think that's reasonable.

Issue #90, opened 2017-06-20
unit: clean up unit package


btracey:
The unit package has a lot of problems. These should be cleaned up. Here are my suggestions

- [x] Clean up basic documentation so that it actually matches the code
- [x] Remove the (incorrect) comment examples and make them test Examples
- [x] Do not provide so many constants for the units. Not `unit.Millimeter`, instead just supply a `unit.Milli` constant, so it can be `unit.Milli * unit.Meter`. Slightly less pretty, but drastically reduces the API surface and makes it more composable with outside code
- [ ] ~~Redefine the `Unit` combination methods to match the `mat` types. For example, `Add` should take in two arguments and store to the receiver~~
- [x] Remove the `formatted` field from `Unit`. It introduces unnecessary state
- [x] Update the autogeneration to use `go generate`. 
- [x] It seems like the formatting code for unit types needs fixing.

kortschak:
I'm not sure about the 4 point here. There is only a concrete `Unit` type, so it's reasonable to be able to chain things.

btracey:
Will ponder.

kortschak:
What is the plan for the `formatted` field?

kortschak:
I've been thinking more about the `formatted` field and I think it should stay. It saves potentially large numbers of allocations (one per print). The race that can exist can be fixed by addition of a `sync.RWMutex` protecting it.

Related to the combination behaviour point. The zero value should be usable for these; if the `dimensions` field is nil (as opposed to empty), they should be set to the appropriate state.

btracey:
I think all that's left of this is point 3

kortschak:
Agreed

PR #91, opened 2017-06-20
unit: Update documentation, create examples, and comment types

btracey:
Updates #90.


unit/doc.go
```diff
@@ -72,18 +72,6 @@
 // used, however, to create the unit of 'Slide', because in this case slide
 // is just a measurement of area. Instead, a constant could be defined.
 //		const Slide unit.Area =  0.001875 // m^2
-//		var WhiteBloodCellDim unit.Dimension
```
btracey: The plan is to make this a real example eventually, but this will come after the fix to the `Mul` method.

PR #92, opened 2017-06-21
mat: add IsZero method to resetable types

kortschak:
@btracey Please take a look.

### DIFF
--- a/mat/doc.go
+++ b/mat/doc.go
@@ -36,10 +36,23 @@
 //  zero.Copy(a)
 //
 // Receivers must be the correct size for the matrix operations, otherwise the
-// operation will panic. As a special case for convenience, a zero-sized matrix
+// operation will panic. As a special case for convenience, a zero-value matrix
 // will be modified to have the correct size, allocating data if necessary.
 //  var c mat.Dense // construct a new zero-sized matrix
-//  c.Mul(a, a)       // c is automatically adjusted to be 6×6
+//  c.Mul(a, a)     // c is automatically adjusted to be 6×6
+//
+// Zero-value of a matrix
+//
+// A zero-value matrix is either the Go language definition of a zero-value or
+// is a zero-sized matrix with zero-length stride. Matrix implementations may have
+// a Reset method to revert the receiver into a zero-valued matrix and an IsZero
+// method that returns whether the matrix is zero-valued.
+// So the following will all result in a zero-value matrix.
+//  - var a mat.Dense
+//  - a := NewDense(0, 0, make([]float64, 0, 100))
+//  - a.Reset()
+// A zero-value matrix can not be sliced even if it does have an adequately sized
+// backing data slice, but can be expanded using its Grow method if it exists.
 //
 // The Matrix Interfaces
 //
--- a/mat/cholesky.go
+++ b/mat/cholesky.go
@@ -457,5 +457,5 @@ func (c *Cholesky) isZero() bool {
 }
 
 func (c *Cholesky) valid() bool {
-	return !c.isZero() && !c.chol.isZero()
+	return !c.isZero() && !c.chol.IsZero()
 }
--- a/mat/dense.go
+++ b/mat/dense.go
@@ -69,7 +69,7 @@ func (m *Dense) reuseAs(r, c int) {
 		// Panic as a string, not a mat.Error.
 		panic("mat: caps not correctly set")
 	}
-	if m.isZero() {
+	if m.IsZero() {
 		m.mat = blas64.General{
 			Rows:   r,
 			Cols:   c,
@@ -95,7 +95,7 @@ func (m *Dense) reuseAsZeroed(r, c int) {
 		// Panic as a string, not a mat.Error.
 		panic("mat: caps not correctly set")
 	}
-	if m.isZero() {
+	if m.IsZero() {
 		m.mat = blas64.General{
 			Rows:   r,
 			Cols:   c,
@@ -147,7 +147,9 @@ func (m *Dense) Reset() {
 	m.mat.Data = m.mat.Data[:0]
 }
 
-func (m *Dense) isZero() bool {
+// IsZero returns whether the receiver is zero-sized. Zero-sized matrices can be the
+// receiver for size-restricted operations. Dense matrices can be zeroed using Reset.
+func (m *Dense) IsZero() bool {
 	// It must be the case that m.Dims() returns
 	// zeros in this case. See comment in Reset().
 	return m.mat.Stride == 0
--- a/mat/dense_arithmetic.go
+++ b/mat/dense_arithmetic.go
@@ -467,7 +467,7 @@ func (m *Dense) Exp(a Matrix) {
 	}
 
 	var w *Dense
-	if m.isZero() {
+	if m.IsZero() {
 		m.reuseAsZeroed(r, r)
 		w = m
 	} else {
@@ -696,7 +696,7 @@ func (m *Dense) Outer(alpha float64, x, y *Vector) {
 		// Panic as a string, not a mat.Error.
 		panic("mat: caps not correctly set")
 	}
-	if m.isZero() {
+	if m.IsZero() {
 		m.mat = blas64.General{
 			Rows:   r,
 			Cols:   c,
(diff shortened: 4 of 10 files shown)


btracey:
I think the package documentation needs to be modified to talk about this. We talk about zero-sized matrices being flexible, but now that we're exposing `IsZero` I think we should talk about `IsZero` and `Reset`. I'm not sure, but I think we also need to commit to a decision on allowing `0x6` matrices (for instance). We currently allow creating a matrix of that size. I don't know if we want to say that 0x0 matrices are Zero, or any zero size is. We could also forbid creating a matrix like that, and thus zeros can only come when one does `var m mat.Dense` or uses `Reset`.

kortschak:
Yes, I agree the docs need fixing for this.

btracey:
We also have to remember this interacts with `Slice`

kortschak:
Now that that's in, this can be fixed to reflect that.

kortschak:
It's not clear to me how to deal with that. The options I see are

1. In the case the matrix is zero, it cannot be sliced. This is how it would work now. Grow can be called on zero matrices (not implemented for all types yet).
2. If it's zero, a minimum inc/stride is chosen (1 for vectors, c for matrices...) and the operation proceeds. I'm not happy with this since there is not an easy way to know in advance whether this will panic.

btracey:
Here is an idea:

Matrices that are 0x0 (or 0x1 for vectors) are zero matrices. When created, everything is 0 (lens, caps, stride, etc.). Similarly, when a matrix is reset, everything gets set to zero. 

Slice and Grow can be called on zero matrices. With slice, it keeps the stride and capacity of the original matrix. I've needed to add single rows to matrices. It's nice to be able to use slice, and pair it down to zero.

The first question is if you do `mat.Slice(0, 0, 0, 0)`. Is that a zero matrix? I'm okay with saying no, it's not, and panicking. If you want to have a flexible size matrix, call `Reset`, don't play tricky games. The second question is if we allow mat.NewDense(0,0,data), so the underlying data can be supplied (say with 0 length and a bunch of capacity). I think it's simpler to allow that than to forbid it.

kortschak:
>Matrices that are 0x0 (or 0x1 for vectors) are zero matrices. When created, everything is 0 (lens, caps, stride, etc.). Similarly, when a matrix is reset, everything gets set to zero.

>Slice and Grow can be called on zero matrices. With slice, it keeps the stride and capacity of the original matrix. I've needed to add single rows to matrices. It's nice to be able to use slice, and pair it down to zero.

I don't see how these two can work together. A zero matrix produced from `var m mat.Dense` cannot be sliced since the stride is 0. It can be grown since `Grow` makes up a sensible stride when operating on a zero (or sub-capacity) matrix.

Because of this, I don't think defining the zero in terms of size is sensible. It should be defined in terms of the behaviour of `IsZero`. `time.Time` has a similar arrangement.

>The first question is if you do `mat.Slice(0, 0, 0, 0)`. Is that a zero matrix? I'm okay with saying no, it's not, and panicking. If you want to have a flexible size matrix, call `Reset`, don't play tricky games. The second question is if we allow `mat.NewDense(0,0,data)`, so the underlying data can be supplied (say with 0 length and a bunch of capacity). I think it's simpler to allow that than to forbid it.

Under the approach of using `IsZero`'s opinion as the definition, then a 0x0 sliced matrix is not a zero matrix since the stride will remain whatever it was prior to the slice call (unless it was already zero, in which case `Slice` would have panicked. Similarly, `mat.NewDense(0, 0, data)` should be illegal, unless data is sliced to zero length. That is already the existing behaviour.

btracey:
Sorry, I misspoke. I agree that `IsZero` should and will provide the canonical definition of what a zero matrix means. I just want to make sure we have that definition fixed before we expose the method. Right now a matrix is zero if the stride is zero, and I just want to confirm that we want it to be an iff relationship.

If it is an iff relationship, then `mat.NewDense(0, 6, nil)` is a valid statement, as is `mat.Slice(0, 0, 0, 0)`, though neither of these yield `IsZero` matrices. Zero matrices themselves can arise from `var m mat.Dense`, `m.Reset()`, and `mat.NewDense(0, 0, data[:0])`. 

I want to confirm that we're agreed the internals are based on the stride and not size. I think the behavior above is simple and comprehensible, and so I'm happy with it. It's maybe a tad opaque, but it's obvious if one understands the internals at all, and I think it's not so hard to interface with if one does not.

kortschak:
I think the current de facto definition is the simplest to understand and easiest to work with. It should be an iff relationship. I think the internals are right to move ahead with.

btracey:
Great. Agreed.

kortschak:
So what is the plan from here? Go through and fix up documentation?

btracey:
That was my thought. At the very least, adding to the package documentation
to make sure we have the behavior contact

kortschak:
PTAL

kortschak:
*ping*

mat/doc.go
```diff
@@ -36,10 +36,25 @@
...
+// operation will panic. As a special case for convenience, a zero-value matrix
 // will be modified to have the correct size, allocating data if necessary.
 //  var c mat.Dense // construct a new zero-sized matrix
-//  c.Mul(a, a)       // c is automatically adjusted to be 6×6
+//  c.Mul(a, a)     // c is automatically adjusted to be 6×6
+//
+// Zero-value of a matrix
+//
+// A zero-value matrix is either the Go language definition of a zero-value or
+// a matrix that has had its Reset method called. Matrix implementations provided
```
btracey: The first sentence is great. The next couple feel a bit wordy. I would recommend the simplification:

Matrix implementations have an IsZero method that returns whether the matrix is zero-valued, and a Reset method to revert the receiver into a zero-valued matrix. The following will all result in a zero-valued matrix:

mat/doc.go
```diff
@@ -36,10 +36,25 @@
...
+// by the mat package that can be reset have an IsZero method that returns whether
+// the matrix is a zero-value matrix. Internally, a zero-value is represented by a
+// zero-sized matrix with a zero-length stride. So the following will all result
+// in a zero-value matrix.
+//  var a mat.Dense
+//  a := NewDense(0, 0, make([]float64, 0, 100))
+//  a.Reset()
+//
```
btracey: No newline.

mat/doc.go
```diff
@@ -36,10 +36,25 @@
...
+// Zero-value of a matrix
+//
+// A zero-value matrix is either the Go language definition of a zero-value or
+// a matrix that has had its Reset method called. Matrix implementations provided
+// by the mat package that can be reset have an IsZero method that returns whether
+// the matrix is a zero-value matrix. Internally, a zero-value is represented by a
+// zero-sized matrix with a zero-length stride. So the following will all result
+// in a zero-value matrix.
+//  var a mat.Dense
```
btracey: I would number the examples or bullet the examples so it's clear lines 2 and 3 are separate examples.

mat/doc.go
kortschak: That doesn't quite say what I want though; only resetable matrices have `IsZero`. At the moment, this is all of them, but `Band` won't in the short term, and it may be possible that we will provide non-resetable types in the future.

kortschak: Also, I think it is worthwhile explaining why `NewDense(0, 0, make([]float64, 0, 100)).IsZero()` is true.

kortschak: Also, the first sentence needs some change because of that `NewDense` pathway to a zero-matrix. I've moved the internal definition (shorted into the first sentence and reordered your first sentence to bring `Reset` to the front.

mat/doc.go
kortschak: Done.

mat/doc.go
kortschak: Done.

mat/doc.go
```diff
@@ -36,10 +36,23 @@
...
 //  var c mat.Dense // construct a new zero-sized matrix
-//  c.Mul(a, a)       // c is automatically adjusted to be 6×6
+//  c.Mul(a, a)     // c is automatically adjusted to be 6×6
+//
+// Zero-value of a matrix
+//
+// A zero-value matrix is either the Go language definition of a zero-value or
+// is a zero-sized matrix with zero-length stride. Matrix implementations may have
+// a Reset method to revert the receiver into a zero-valued matrix and an IsZero
+// method that returns whether the matrix is zero-valued.
```
btracey: Sounds great.

kortschak:
Sorry, with the experiment of using up-to-date checking, I have blown away your accept. Can you please redo that.

btracey:
Sorry, missed your message.

kortschak:
No worries.

PR #95, opened 2017-06-21
mat: add basis for band matrix type

kortschak:
This adds a starting point for building. I'm not currently happy with matrix construction since it's possible as the code stands to change the test case from 6x6 to 6x4 and no other changes without failure. I don't see a way to address this.

@btracey Please take a look.


mat/band.go
```diff
@@ -0,0 +1,147 @@
...
+func (t TransposeBand) UntransposeBand() Band {
+	return t.Band
+}
+
+// NewBandDense creates a new Band matrix with r rows and c columns. If data == nil,
+// a new slice is allocated for the backing slice. If len(data) == r*(kl+ku+1),
+// data is used as the backing slice, and changes to the elements of the returned BandDense
+// will be reflected in data. If neither of these is true, NewBandDense will panic.
+//
+// The data must be arranged in row-major order. Only the values in the band portion
```
btracey: I think this needs a definition of "row-major order", as if you haven't seen banded matrices it's not obvious what the layout is. Something like a copy and paste from the BLAS documentation.

mat/band.go
```diff
@@ -0,0 +1,147 @@
...
+	_         Band    = bandDense
+	_         RawBand = bandDense
+)
+
+// BandDense represents a band matrix in dense storage format.
+type BandDense struct {
+	mat blas64.Band
+}
+
+type Band interface {
```
btracey: I think we want the interface to be `Banded` rather than `Band` to match `Triangular` and `Symmetric`. I'm not sure if we want the type to be `BandedDense` instead of `BandDense`

mat/band_test.go
```diff
@@ -0,0 +1,144 @@
...
+			t.Errorf("unexpected number of cols for test %d: got: %d want: %d", i, cols, test.c)
+		}
+		if !reflect.DeepEqual(band, test.mat) {
+			t.Errorf("unexpected data slice for test %d: got: %v want: %v", i, band, test.mat)
+		}
+	}
+}
+
+func TestBandAtSet(t *testing.T) {
+	band := NewBandDense(6, 6, 1, 2, []float64{
```
btracey: It would be nice to have a test that just accesses all of the matrix elements to make sure the index tests are correct.

btracey:
I also do not see a way to address the allocation problem. 

Two general concerns moving forward (not a problem with this PR).

1) Do we have / could we have a way of maintaining a list of the special cases that we have implemented? It would be nice to have some big table where we could see gaps that need to be filled.

2) Here is a nice property of banded matrices. The Cholesky decomposition of a (symmetric) banded matrix is also banded. It would be great if the Cholesky type could deal with that and "just work". In particular, somehow making, say, `distmv.Normal` agnostic to the representation. I believe the constraint on that particular issue is having the "Lower" extraction be agnostic.

mat/band.go
kortschak: Done.

mat/band.go
kortschak: Done.

mat/band_test.go
kortschak: Done.

kortschak:
PTAL

mat/band.go
```diff
@@ -0,0 +1,163 @@
...
+var (
+	_ Matrix            = TransposeBand{}
+	_ Banded            = TransposeBand{}
+	_ UntransposeBander = TransposeBand{}
+)
+
+// TransposeBand is a type for performing an implicit transpose of a band
+// matrix. It implements the Banded interface, returning values from the
+// transpose of the matrix within.
+type TransposeBand struct {
```
vladimir-ch: Should this be `TransposeBanded`? In general, from a quick look, I don't understand the rules when Band or Banded is used.

kortschak: I did write it like that, but reverted because of length. Various people call these matrices band or banded. Here Banded is the interface and Band is the theoretical notion. I don't really mind either way.

mat/band.go
```diff
@@ -0,0 +1,163 @@
...
+// BandDense represents a band matrix in dense storage format.
+type BandDense struct {
+	mat blas64.Band
+}
+
+type Banded interface {
+	Matrix
+	// Triangular returns the number of rows/columns in the matrix and its
+	// orientation.
+	Bandwidth() (kl, ku int)
```
btracey: Should this be plural?

mat/band.go
```diff
@@ -0,0 +1,163 @@
...
+
+	// TBanded is the equivalent of the T() method in the Matrix interface but
+	// guarantees the transpose is of banded type.
+	TBanded() Banded
```
btracey: This should be `TBand` to match `TTri` and the rest of the "abbreviation" conventions.

mat/band_test.go
```diff
@@ -0,0 +1,180 @@
...
+		band := NewBandDense(test.r, test.c, test.kl, test.ku, test.data)
+		rows, cols := band.Dims()
+
+		if rows != test.r {
+			t.Errorf("unexpected number of rows for test %d: got: %d want: %d", i, rows, test.r)
+		}
+		if cols != test.c {
+			t.Errorf("unexpected number of cols for test %d: got: %d want: %d", i, cols, test.c)
+		}
+		if !reflect.DeepEqual(band, test.mat) {
```
btracey: Add a test here with the *Dense matrix and also use mat.Equal

mat/band_test.go
```diff
@@ -0,0 +1,180 @@
...
+			t.Errorf("unexpected value for At(%d, %d): got: %v want: %v", st.row, st.col, e, st.orig)
+		}
+		band.SetBand(st.row, st.col, st.new)
+		if e := band.At(st.row, st.col); e != st.new {
+			t.Errorf("unexpected value for At(%d, %d) after SetTri(%[1]d, %d, %v): got: %v want: %[3]v", st.row, st.col, st.new, e)
+		}
+	}
+}
+
+// bandImplicit is an implicit band matrix returning val(i, j)
```
btracey: Could you add these types to list_test as well?

mat/band.go
kortschak: The band's width is kl+ku+1, so no, I don't think so.

btracey: Ok.

mat/band.go
kortschak: Done.

kortschak:
I will lose approval with these changes, so I'll rebase on to master while I'm here.

mat/band_test.go
kortschak: Which `*Dense` matrix?

mat/band_test.go
kortschak: Can I leave that for you? That system scares me a little.

mat/band_test.go
kortschak: Done, I think.

kortschak:
PTAL

mat/band.go
kortschak: But the doc comment should not say Triangular!

btracey:
It should also probably actually say what the function does.

Sorry for missing that.

kortschak:
OK, that and another missed typo are fixed. PTAL

mat/band_test.go
btracey: I was hoping for 
```
matDense: NewDense(6, 6, []float64{
1,  2,  3,  0,  0,  0,
4,  5,  6,  7,  0,  0,
0,  8,  9, 10, 11,  0,
0,  0, 12, 13, 14, 15,
0,  0,  0, 16, 17, 18,
0,  0,  0,  0, 19, 20,
})
```

and then comparing that with test.band

kortschak:
I'll make those changes later today. This will invalidate the approval.

mat/band_test.go
kortschak: Ah, OK. Done.

mat/band_test.go
kortschak: Thanks.

kortschak:
PTAL

mat/band.go
```diff
@@ -0,0 +1,163 @@
...
+//    0  0  0 16 17 18
+//    0  0  0  0 19 20
+// becomes (* entries are never accessed)
+//     *  1  2  3
+//     4  5  6  7
+//     8  9 10 11
+//    12 13 14 15
+//    16 17 18  *
+//    19 20  *  *
+// which is given to the passed to NewBandDense as []float64{*, 1, 2, 3, 4, ...}.
```
vladimir-ch: `which is passed to NewBandDense ...` ?
Also, wouldn't it be worth to write that in this example `kl == 1` and `ku == 2`?

mat/band.go
```diff
@@ -0,0 +1,163 @@
...
+func (b *BandDense) T() Matrix {
+	return Transpose{b}
+}
+
+// TBand performs an implicit transpose by returning the receiver inside a TransposeBand.
+func (b *BandDense) TBand() Banded {
+	return TransposeBand{b}
+}
+
+// RawMatrix returns the underlying blas64.Band used by the receiver.
```
vladimir-ch: RawBand

mat/band.go
```diff
@@ -0,0 +1,163 @@
...
+		},
+	}
+}
+
+// Dims returns the number of rows and columns in the matrix.
+func (b *BandDense) Dims() (r, c int) {
+	return b.mat.Rows, b.mat.Cols
+}
+
+// Dims returns the upper and lower bandwidths of the matrix.
```
vladimir-ch: `// Bandwidth returns the lower and upper bandwidths ...`

mat/band_test.go
```diff
@@ -0,0 +1,207 @@
...
+		return float64(i*(kl+ku) + j + kl + 1)
+	}}
+	for i := 0; i < 6; i++ {
+		for j := 0; j < 6; j++ {
+			if band.At(i, j) != want.At(i, j) {
+				t.Errorf("unexpected value for band.At(%d, %d): got:%v want:%v", band.At(i, j), want.At(i, j))
+			}
+		}
+	}
+	// Do that same thing via a vall to Equal.
```
vladimir-ch: via a call

mat/band.go
kortschak: Done.

mat/band.go
kortschak: Done.

mat/band.go
kortschak: Done

mat/band_test.go
kortschak: Done.

kortschak:
PTAL

kortschak:
*ping* @btracey since you are up.

mat/band.go
```diff
@@ -0,0 +1,170 @@
...
+// UntransposeBand returns the Banded field.
+func (t TransposeBand) UntransposeBand() Banded {
+	return t.Banded
+}
+
+// NewBandDense creates a new Band matrix with r rows and c columns. If data == nil,
+// a new slice is allocated for the backing slice. If len(data) == min(r, c+kl)*(kl+ku+1),
+// data is used as the backing slice, and changes to the elements of the returned
+// BandDense will be reflected in data. If neither of these is true, NewBandDense
+// will panic. kl must be at least zero and less r, and ku must be at least zero and
```
kortschak: Note comments on kl and ku now.

btracey: Does this behavior match Lapack? I think we want to be compliant here. I've only started looking at lapack banded implementations, but it would be nice to not have to play too many games (that said, we are stricter for Vector and it's worked out fine)

mat/band.go
```diff
@@ -136,6 +136,12 @@ func NewBandDense(r, c, kl, ku int, data []float64) *BandDense {
 	}
 }
 
+// NewDiagonal is a convenience function that returns a diagonal matrix represented by a
+// BandDense. The length of data must be min(r, c) otherwise NewDiagonal will panic.
+func NewDiagonal(r, c int, data []float64) *BandDense {
```
btracey: Please leave this out of this pull request.

I think we want a `SymBanded` interface / type. It makes a huge difference in `distmv.NewNormal` (O(n^3) to O(n) for computing the Cholesky), and is probably a common case in the spatial statistics world also. Typically, diagonal matrices are also square, so I think we would want this to return a SymBandDense by default.

mat/band.go
kortschak: The documentation says that the `*` elements are not referenced. If this causes a failure, I think then that the LAPACK implementation is buggy.

Looking at how we do this, I suspect it will fail for the gonum implementation, for example [here](https://github.com/gonum/gonum/blob/master/blas/gonum/level2double.go#L272-L283). However, I think that is a result of our slice optimisations; the [Fortran should not fail](http://www.netlib.org/lapack/explore-html/d2/d3f/dgbmv_8f_source.html).

btracey: Thanks. I agree. We/I should probably revisit the banded BLAS implementations before we're 6 layers deep in Lapack trying to figure out what's going on.

mat/band.go
kortschak: The sigma in SVD is potentially non-square. Similarly in GSVD. `NewDiagonalBand` and `NewDiagonalSym`? The reason this is here is purely for legibility to draw attention to the fact that the result is a diagonal.

mat/band.go
kortschak: There are two ways to deal with those, either `continue` if the proposed length is 0, or probably preferably, make the same assessment that I am doing here.

mat/band.go
btracey: Yes, I agree with the intent. Wikipedia suggests that the term Diagonal "usually refers to square matrices", and calls non-square ones "rectangular diagonal matrices". 

I would go with `NewDiagonal` for the symmetric case, and `NewDiagonalBand` or `NewDiagonalRect` for the rectangular case.

`data` should be allowed to be nil, just like the other `NewXxx` functions. I commonly create a new matrix then use a loop to set the diagonal elements (and fortunately these will be much more efficient soon).

btracey: Sorry, I'm an idiot. It can already be nil.

kortschak: An alternative is to have `NewDiagonal(r, c int, data []float64) Matrix` and return either a `*BandDense` or a `*SymBandDense` depending on whether r == c.

btracey: I'd rather have the two functions that return their specific types and not force type assertions.

btracey: We could also have `NewDiagSq` and `NewDiagRect`

kortschak: I'll rename this to `NewDiagonalRect`.

PR #96, opened 2017-06-21
stat/distuv: move destination to first position

kortschak:
@btracey Please take a look.

Updates gonum/stat#179.
Updates #4.


btracey:
LGTM

PR #97, opened 2017-06-21
stat: append To suffix to extraction methods

kortschak:
@btracey Please take a look.

This is a replay of gonum/stat#170
Updates #4.


btracey:
LGTM

PR #103, opened 2017-06-23
stat/distmv: Add ScoreInput function for the Normal distribution

btracey:


btracey:
Fixed a bug in the implementation. Very tricksy.

stat/distmv/normal.go
```diff
@@ -277,6 +277,38 @@ func (n *Normal) Rand(x []float64) []float64 {
 	return x
 }
 
+// ScoreInput returns the score function with respect to the input of the
```
vladimir-ch: I would rewrite these two sentences as: ScoreInput returns the value at x of the score function s which is defined as the gradient of the log-probability:  s(x) = \nabla log(p(x)).
If gradient is not used in statistics, then keep derivative.
Also, why not call the method just Score?

stat/distmv/normal.go
```diff
@@ -277,6 +277,38 @@ func (n *Normal) Rand(x []float64) []float64 {
...
+// distribution at the location specified by x. The score function is the
+// derivative of the log-probability at the input.
+//  ∂/∂x log(p(x))
+// If deriv is nil, a new slice will be allocated and returned. If deriv is of
+// length the dimension of Normal, then the result will be put in-place into deriv.
+// If neither of these is true, ScoreInput will panic.
+func (n *Normal) ScoreInput(deriv, x []float64) []float64 {
```
vladimir-ch: I would call the first parameter `score`.

stat/distmv/normal.go
btracey: Because the score function https://en.wikipedia.org/wiki/Score_(statistics) is the derivative with respect to the parameters, and this is with respect to the inputs.

btracey: Done.

stat/distmv/normal.go
btracey: Done.

stat/distmv/normal.go
```diff
@@ -277,6 +277,37 @@ func (n *Normal) Rand(x []float64) []float64 {
 	return x
 }
 
+// ScoreInput returns the gradient of the log-probability with respect to the
+// input x. That is, ScoreInput computes
+//  ∇_x log(p(x))
+// If score is nil, a new slice will be allocated and returned. If score is of
+// length the dimension of Normal, then the result will be put in-place into deriv.
```
kortschak: `deriv` does not exist. Do you mean `score`?

btracey: Yes, sorry, missed that one.

btracey: Done. (sorry for squashing, made some silly git decisions)

kortschak:
This has gone semantically stale. It uses the solve method on `mat.Vector` which no longer exists. You should probably rebase onto master and fix that.

stat/distmv/normal.go
```diff
@@ -277,6 +277,37 @@ func (n *Normal) Rand(x []float64) []float64 {
...
+	if len(score) != len(x) {
+		panic(badSizeMismatch)
+	}
+	tmp := make([]float64, len(x))
+	copy(tmp, x)
+	floats.Sub(tmp, n.mu)
+
+	dv := mat.NewVector(len(score), score)
+	dt := mat.NewVector(len(tmp), tmp)
+	dv.SolveCholeskyVec(&n.chol, dt)
```
kortschak: `n.chol.SolveVec(mat.NewVector(len(score), score), mat.NewVector(len(tmp), tmp))`

btracey: Thanks

btracey:
Rebased.

PR #107, opened 2017-06-26
travis,unit: fix and check unit code generation

kortschak:
Please take a look.


btracey:
This fixes the last bullet of #90 , right?

kortschak:
And the second last (I fixed the markdown on that issue).

btracey:
Ah yes, the more obvious of the two :-)

PR #109, opened 2017-06-27
diff/fd: implement Hessian finite difference, and code cleanups.

btracey:
This commit primarily adds the Hessian function for finding a finite difference approximation to the Hessian. At the same time, it combines duplicated functionality across the difference routines so that the preludes to all the difference routines look similar


btracey:
PTAL @vladimir-ch 

This doesn't do anything tricky with trying to prevent repeat evaluations, as you said, quite complicated. Saving O(n) evaluations isn't nothing, but out of O(n^2) computations, maybe it isn't worth the additional complexity. 

I thought about the two-formula approach, but I think in the general case this is tricky with also forcing the matrix to be symmetric. If you want distinct formulas in the x and y direction, this means that you should also get different derivative estimates in the different directions. It's somewhat annoying this effectively uses `Forward2nd` on the diagonal entries, but the differences aren't so huge and this gives the user the option of using `Central` to get a better estimate. I think the only other reasonable option is to forbid setting the formula.

Fixes https://github.com/gonum/diff/issues/23

diff/fd/derivative.go
```diff
@@ -0,0 +1,71 @@
...
+			step = formula.Step
+			checkFormula(formula)
+		}
+		if settings.Step != 0 {
+			step = settings.Step
+		}
+		originKnown = settings.OriginKnown
+		originValue = settings.OriginValue
+		concurrent = settings.Concurrent
+	}
```
btracey: Sorry, I didn't think how the diffs would look. I didn't change anything below here except removing the `settings.` uses

diff/fd/gradient.go
```diff
@@ -0,0 +1,138 @@
...
+				panic(badDerivOrder)
+			}
+		}
+		if settings.Step != 0 {
+			step = settings.Step
+		}
+		originKnown = settings.OriginKnown
+		originValue = settings.OriginValue
+		concurrent = settings.Concurrent
+	}
```
btracey: Sorry, I didn't think how the diffs would look. I didn't change anything below here except removing the `settings.` uses

diff/fd/diff.go
```diff
@@ -274,3 +94,53 @@ var Central2nd = Formula{
...
+		return 1
+	}
+	nWorkers := runtime.GOMAXPROCS(0)
+	if nWorkers > evals {
+		nWorkers = evals
+	}
+	return nWorkers
+}
+
+// hasOrigin returns whether the stencil uses the origin, which is true iff
```
sbinet: `s/hasOrigin/usesOrigin/`

diff/fd/hessian.go
```diff
@@ -0,0 +1,181 @@
...
+					var v float64
+					if pti.Loc == 0 && ptj.Loc == 0 {
+						v = origin
+					} else {
+						copy(xCopy, x) // perform copy anew because of floating point issues
+						xCopy[i] += pti.Loc * step
+						xCopy[j] += ptj.Loc * step
+						v = f(xCopy)
+					}
+					hess += v * pti.Coeff * ptj.Coeff / (step * step)
```
sbinet: perhaps hoist the `/ (step*step)` into an `invS2 := 1 / (step*step)` variable ?

diff/fd/diff.go
btracey: Done.

diff/fd/hessian.go
btracey: Done.

diff/fd/diff.go
```diff
@@ -274,3 +94,53 @@ var Central2nd = Formula{
 	Derivative: 2,
 	Step:       1e-4,
 }
+
+// checkFormula checks if the formula is valid, and panics otherwise.
+func checkFormula(formula Formula) {
+	if formula.Derivative == 0 || formula.Stencil == nil || formula.Step == 0 {
```
btracey: Done.

diff/fd/hessian.go
```diff
@@ -0,0 +1,183 @@
...
+
+import (
+	"math"
+	"sync"
+
+	"gonum.org/v1/gonum/mat"
+)
+
+// Hessian approximates the Hessian matrix of the multivariate function f
+// at the location. That is
```
btracey: Done.

diff/fd/hessian.go
```diff
@@ -0,0 +1,183 @@
...
+//  H_{i,j} = ∂^2 f/∂x_i ∂x_j
```
btracey: Done.

diff/fd/hessian.go
```diff
@@ -0,0 +1,183 @@
...
+// If dst is not nil, the result will be stored in-place into dst
```
btracey: Done.

diff/fd/hessian.go
```diff
@@ -0,0 +1,183 @@
...
+// and returned, otherwise a new matrix will be allocated first. Finite difference
+// kernel and other options are specified by settings. If settings is nil,
```
btracey: Done.

diff/fd/hessian.go
```diff
@@ -0,0 +1,183 @@
...
+		if !settings.Formula.isZero() {
+			formula = settings.Formula
+			step = math.Sqrt(formula.Step)
+			checkFormula(formula)
+			if formula.Derivative != 1 {
+				panic(badDerivOrder)
+			}
+		}
+		if settings.Step != 0 {
+			step = settings.Step
```
btracey: Done.

diff/fd/jacobian.go
```diff
@@ -79,18 +81,14 @@ func Jacobian(dst *mat.Dense, f func(y, x []float64), x []float64, settings *Jac
...
+	nWorkers := computeWorkers(concurrent, evals)
 	if nWorkers == 1 {
-		jacobianSerial(dst, f, x, settings.OriginValue, formula, step)
-	} else {
-		jacobianConcurrent(dst, f, x, settings.OriginValue, formula, step, nWorkers)
+		jacobianSerial(dst, f, x, originValue, formula, step)
+		return
 	}
+	jacobianConcurrent(dst, f, x, originValue, formula, step, nWorkers)
+	return
```
btracey: Done.

diff/fd/gradient.go
```diff
@@ -0,0 +1,138 @@
...
+// license that can be found in the LICENSE file.
+
+package fd
+
+import "gonum.org/v1/gonum/floats"
+
+// Gradient estimates the gradient of the multivariate function f at the
+// location x. If dst is not nil, the result will be stored in-place into dst
+// and returned, otherwise a new slice will be allocated first. Finite
+// difference kernel and other options are specified by settings. If settings is
```
btracey: Done.

diff/fd/hessian_test.go
```diff
@@ -0,0 +1,69 @@
...
+
+func TestHessian(t *testing.T) {
+	for cas, test := range []struct {
+		h        HessianTester
+		x        []float64
+		settings *Settings
+		tol      float64
+	}{
+		{
+			h:   Watson{},
```
btracey: Done.

btracey:
PTAL.

This needs another LGTM because it was opened before the change (I believe).

btracey:
Ping @vladimir-ch

diff/fd/hessian.go
```diff
@@ -0,0 +1,186 @@
...
+	for i := 0; i < n; i++ {
+		for j := i; j < n; j++ {
+			var hess float64
+			for _, pti := range stencil {
+				for _, ptj := range stencil {
+					var v float64
+					if pti.Loc == 0 && ptj.Loc == 0 {
+						v = origin
+					} else {
+						copy(xCopy, x) // perform copy anew because of floating point issues
```
vladimir-ch: Isn't this done rather because `f` could modify `x`? Also, capital P and final dot.

diff/fd/hessian_test.go
vladimir-ch: Nice, thanks.

diff/fd/hessian.go
btracey: No, it's because with floating point numbers,  `x += pti.Loc*step; x -= pti.Loc*step` doesn't get you back to `x`. It seemed like that kind of "origin creep" is not a good idea. Being robust to the value being modified is a side benefit.

btracey:
Fixed comment punctuation. You need to approve this again though now.

diff/fd/hessian.go
vladimir-ch: Ah, I thought the primary motivation was the protection against modification because otherwise the modified elements could be stored into local variables and we could avoid the allocations and copying. Also, a comment in `Gradient` uses this motivation, so these two should be unified.

btracey: Would you like me to do that here?

vladimir-ch: I think so. Thanks.

btracey:
Done, with both reasons given. The original comment in Gradient only serves the one function since it's at the origin.

diff/fd/hessian.go
btracey: Done. PTAL

PR #113, opened 2017-06-30
mat: rewrite extractions and solves as methods on factorising types

### LINKED ISSUE #101: mat: Move Solve routines into their particular type
We moved the extraction types (i.e. `LTo`) into their respective factorized types. We left the Solve routines in their relative types. This still makes certain factorizations privileged.

### LINKED ISSUE #102: mat: Move LFromCholesky into Cholesky
We forgot this one when we moved the other types.

kortschak:
@btracey Please take a look.

Fixes #101.
Fixes #102.

### DIFF
--- a/mat/cholesky.go
+++ b/mat/cholesky.go
@@ -267,25 +267,25 @@ func (c *Cholesky) To(s *SymDense) {
 	s.SymOuterK(1, c.chol.T())
 }
 
-// InverseCholesky computes the inverse of the matrix represented by its Cholesky
-// factorization and stores the result into the receiver. If the factorized
+// InverseTo computes the inverse of the matrix represented by its Cholesky
+// factorization and stores the result into s. If the factorized
 // matrix is ill-conditioned, a Condition error will be returned.
 // Note that matrix inversion is numerically unstable, and should generally be
 // avoided where possible, for example by using the Solve routines.
-func (s *SymDense) InverseCholesky(chol *Cholesky) error {
-	if !chol.valid() {
+func (c *Cholesky) InverseTo(s *SymDense) error {
+	if !c.valid() {
 		panic(badCholesky)
 	}
 	// TODO(btracey): Replace this code with a direct call to Dpotri when it
 	// is available.
-	s.reuseAs(chol.chol.mat.N)
+	s.reuseAs(c.chol.mat.N)
 	// If:
 	//  chol(A) = U^T * U
 	// Then:
 	//  chol(A^-1) = S * S^T
 	// where S = U^-1
 	var t TriDense
-	err := t.InverseTri(chol.chol)
+	err := t.InverseTri(c.chol)
 	s.SymOuterK(1, &t)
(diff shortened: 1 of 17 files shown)


btracey:
Here's a meta comment, just to make sure I'm thinking about this correctly in the code review.

I would love to figure out how to make higher-level libraries be efficient with their use of matrix types. Let's say (in a future gonum), we have `var a mat.Diag`, and we have filled in `a`. I would love if `cholesky.Factorize` just takes the square root of the matrix entries and stores a diagonal matrix. This could happen, for instance, in `distmv.Normal`, passing in a diagonal covariance matrix (very common). 

I believe moving the Solve routines into `Cholesky` allows this. `Cholesky` can somehow maintain the inner matrix type, and so when we call `chol.SolveVec`, we can call the diagonal SolveVec routine (which would be trivial). Similarly, by providing a `chol.MulVec`, we have now reduced all operations on Normal from `N^3` to just `O(N)`, a huge speedup. 

I also think if we do this properly, this opens the door for really nice interoperability with sparse matrices. For instance, our current linear programming implementation forces using dense matrices, which means it'll never compete with "real" lp solvers. But, we could create some interface that's like

```
type LPMatrix struct {
      SolveVec(v, b *Vector) error
      MulVec(a,b *Vector)
}
```

This interface can now be implemented by Cholesky (for Dense LPs, or structured sparse LPs), and then by some outside/future package that provides those routines for a sparse matrix and Cholesky decomposition. 

This is the opportunity I think we can provide with a potentially huge benefit. I'd like to see if we're on the same page.

kortschak:
Moving the methods onto the factorising types I don't think changes this (first and second part), but can be used for the lp part.

My main motivation is just ease of legibility.

Note that there's another sweep that needs to go through that allows many of these methods to take nil and allocate returning the result. Also, I think Cholesky.To may want to be called ToSymmetric. What do you think?

Finally, with the addition of the band types, we get diagonals for free, so that can just be type switched in the appropriate places.

mat/cholesky.go
```diff
@@ -223,69 +223,84 @@ func (v *Vector) SolveCholeskyVec(chol *Cholesky, b *Vector) error {
...
 }
 
-// UFromCholesky extracts the n×n upper triangular matrix U from a Cholesky
-// decomposition
+// UTo extracts the n×n upper triangular matrix U from a Cholesky
+// decomposition into dst and returns the result. If dst is nil a new
+// TriDense is allocated.
 //  A = U^T * U.
-func (t *TriDense) UFromCholesky(chol *Cholesky) {
+func (chol *Cholesky) UTo(dst *TriDense) *TriDense {
```
vladimir-ch: Please unify the name of the receiver.

kortschak: Thanks. Done.

kortschak:
I have not made errorable methods (inverse and solve) able to allocate a destination and return it. This can be done later if people want it.

PR #125, opened 2017-07-07
Update AUTHORS and CONTRIBUTORS

bnoon:
Made small fix: 
[gonum/plot] Correction for Luminance.At() when range min < 0 (#369)


kortschak:
https://github.com/gonum/plot/pull/369

PR #126, opened 2017-07-08
mat: add Doer interface and implementations

kortschak:
This is a proposal for an approach to optimising sparse or sparsish matrix accesses raised in https://github.com/gonum/gonum/pull/124#discussion_r126201816

~~Only https://github.com/gonum/gonum/commit/2eb65a280581fdb47fa2714f84b793f652ccbb58 is relevant to this PR and I'll revise the history of the PR when #95 is merged.~~

Depends #95.

### DIFF
--- a/mat/matrix_test.go
+++ b/mat/matrix_test.go
@@ -521,3 +521,94 @@ func TestTrace(t *testing.T) {
 	}
 	testOneInputFunc(t, "Trace", f, denseComparison, sameAnswerFloat, isAnyType, isSquare)
 }
+
+func TestDoer(t *testing.T) {
+	type MatrixDoer interface {
+		Matrix
+		NonZeroDoer
+		RowNonZeroDoer
+		ColNonZeroDoer
+	}
+	ones := func(n int) []float64 {
+		data := make([]float64, n)
+		for i := range data {
+			data[i] = 1
+		}
+		return data
+	}
+	for i, m := range []MatrixDoer{
+		NewTriDense(3, Lower, ones(3*3)),
+		NewTriDense(3, Upper, ones(3*3)),
+		NewBandDense(6, 6, 1, 1, ones(3*6)),
+		NewBandDense(6, 10, 1, 1, ones(3*6)),
+		NewBandDense(10, 6, 1, 1, ones(7*3)),
+		NewSymBandDense(3, 0, ones(3)),
+		NewSymBandDense(3, 1, ones(3*(1+1))),
+		NewSymBandDense(6, 1, ones(6*(1+1))),
+		NewSymBandDense(6, 2, ones(6*(2+1))),
+	} {
+		r, c := m.Dims()
+
+		want := Sum(m)
+
+		// got and fn sum the accessed elements in
+		// the Doer that is being operated on.
+		// fn also tests that the accessed elements
+		// are within the writable areas of the
+		// matrix to check that only valid elements
+		// are operated on.
+		var got float64
+		fn := func(i, j int, v float64) {
+			got += v
+			switch m := m.(type) {
+			case MutableTriangular:
+				m.SetTri(i, j, v)
+			case MutableBanded:
+				m.SetBand(i, j, v)
+			case MutableSymBanded:
+				m.SetSymBand(i, j, v)
+			default:
+				panic("bad test: need mutable type")
+			}
+		}
+
+		panicked, message := panics(func() { m.DoNonZero(fn) })
+		if panicked {
+			t.Errorf("unexpected panic for Doer test %d: %q", i, message)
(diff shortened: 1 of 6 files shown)


kortschak:
Cleaned up and added `SymBandDense` implementation.

PTAL

btracey:
Can we find a better name than `Do`? In particular, the key feature of this is that it only operates on the "non-provably zero" region. The name should reflect that somehow.

kortschak:
Please suggest a name. I have nothing better.

btracey:
How about `WalkNZ`? It iterates over the (provably) non-zero entries. Or, `ReduceNZ`, because reduce will frequently be the intended purpose?

kortschak:
WalkNZ sounds like an adventure tourism campaign. I don't like either of walk or reduce because of loading from other domains (walk is a graph theoretic thing and reduce sounds like it comes from mapreduce). How about DoNonZero, DoRowNonZero and DoColNonZero?

btracey:
The problem with `Do` is it means nothing. The method iterates over the non-zero entries and calls the function. `CallNonZero`? `IterNonZero`?

kortschak:
I'd argue that Do and Call are synonyms when the parameter is a func.

btracey:
Okay. SGTM

kortschak:
PTAL

kortschak:
I've used "provably non-zero" instead of "filled". I don't like it much, so if you have a better suggestion, I'd happily change it.

btracey:
It should be "not provably zero". We don't have any elements that are provably non-zero

btracey:
I wonder if it's better that this call the function at the non-zero elements, rather than at the elements which are provably non-zero. First of all, the behavior is easier to understand. Second of all, it means this behavior works correctly for sparse matrices where often no elements are provably non-zero, but many elements will actually be non-zero.

btracey:
(and the change improves the important sparsity properties where a bunch of elements are known to be zero, so they don't have to be checked)

btracey:
(also the method name doesn't have to be changed :) )

kortschak:
I'm happy to call it "non-zero element", except that a band created by `mat.NewBand(3, 3, 0, 0, make([]float64, 3))` is all zero, but the function will be called with {0, 0, 0}, {1, 1, 0} and {2, 2, 0}. This is why I used "filled" before.

btracey:
I understand that is the behavior now, but I'm suggesting the function is more generally useful if it doesn't call the function at all for zero elements. Meaning, for `mat.NewBand(3, 3, 0, 0, make([]float64, 3))`, the function isn't called at all. This behavior is much more understandable from an interface perspective. 

Code that relies on the "provably" part will have to do a type switch anyway, because you can't tell what "provably" means from the interface. In that light, I think it's better to not call at all for non-zero entries.

kortschak:
>Code that relies on the "provably" part will have to do a type switch anyway, because you can't tell what "provably" means from the interface. In that light, I think it's better to not call at all for non-zero entries.

OK. I can't do that today. I'll get back to this later.

kortschak:
Was less than I though - the tests were already written with only non-zero elements in the bands.

mat/matrix_test.go
```diff
@@ -521,3 +521,86 @@ func TestTrace(t *testing.T) {
...
+		NewSymBandDense(6, 1, ones(6*(1+1))),
+		NewSymBandDense(6, 2, ones(6*(2+1))),
+	} {
+		r, c := m.Dims()
+
+		want := Sum(m)
+
+		var got float64
+		fn := func(i, j int, v float64) {
+			got += v
```
btracey: Could you add a comment to the effect of "this sets to the same value to test accessing"? I originally was thinking this was mutating the matrix to have new values for the subsequent tests.

mat/matrix_test.go
```diff
@@ -521,3 +521,86 @@ func TestTrace(t *testing.T) {
...
+		panicked, message := panics(func() { m.DoNonZero(fn) })
+		if panicked {
+			t.Errorf("unexpected panic for Doer test %d: %q", i, message)
+			continue
+		}
+		if got != want {
+			t.Errorf("unexpected Doer sum: got:%f want:%f", got, want)
+		}
+
+		got = 0
```
btracey: Could you also add a comment like "test that the sum over all the rows equals the sum over the matrix"? I was caught up a while on the statefulness of `got`.

mat/matrix_test.go
kortschak: Done

mat/matrix_test.go
kortschak: Done

kortschak:
PTAL

PR #130, opened 2017-07-09
mat: add symmetric band types

kortschak:
@btracey Please take a look.

### DIFF
--- a/mat/index_bound_checks.go
+++ b/mat/index_bound_checks.go
@@ -197,7 +197,7 @@ func (s *SymBandDense) at(i, j int) float64 {
 		i, j = j, i
 	}
 	pj := j - i
-	if pj < 0 || s.mat.K+1 <= pj {
+	if s.mat.K+1 <= pj {
 		return 0
 	}
 	return s.mat.Data[i*s.mat.Stride+pj]
@@ -220,7 +220,7 @@ func (s *SymBandDense) set(i, j int, v float64) {
 		i, j = j, i
 	}
 	pj := j - i
-	if pj < 0 || s.mat.K+1 <= pj {
+	if s.mat.K+1 <= pj {
 		panic(ErrBandSet)
 	}
 	s.mat.Data[i*s.mat.Stride+pj] = v
--- a/mat/index_bound_checks.go
+++ b/mat/index_bound_checks.go
@@ -180,3 +180,48 @@ func (b *BandDense) set(i, j int, v float64) {
 	}
 	b.mat.Data[i*b.mat.Stride+pj] = v
 }
+
+// At returns the element at row i, column j.
+func (s *SymBandDense) At(i, j int) float64 {
+	return s.at(i, j)
+}
+
+func (s *SymBandDense) at(i, j int) float64 {
+	if uint(i) >= uint(s.mat.N) {
+		panic(ErrRowAccess)
+	}
+	if uint(j) >= uint(s.mat.N) {
+		panic(ErrColAccess)
+	}
+	if i > j {
+		i, j = j, i
+	}
+	pj := j - i
+	if pj < 0 || s.mat.K+1 <= pj {
+		return 0
+	}
+	return s.mat.Data[i*s.mat.Stride+pj]
+}
+
+// SetSymBand sets the element at row i, column j to the value v.
+// It panics if the location is outside the appropriate region of the matrix.
+func (s *SymBandDense) SetSymBand(i, j int, v float64) {
+	s.set(i, j, v)
+}
+
+func (s *SymBandDense) set(i, j int, v float64) {
+	if uint(i) >= uint(s.mat.N) {
+		panic(ErrRowAccess)
+	}
+	if uint(j) >= uint(s.mat.N) {
+		panic(ErrColAccess)
+	}
+	if i > j {
+		i, j = j, i
+	}
+	pj := j - i
+	if pj < 0 || s.mat.K+1 <= pj {
+		panic(ErrBandSet)
+	}
+	s.mat.Data[i*s.mat.Stride+pj] = v
+}
--- a/mat/index_no_bound_checks.go
+++ b/mat/index_no_bound_checks.go
@@ -198,7 +198,7 @@ func (s *SymBandDense) at(i, j int) float64 {
 		i, j = j, i
 	}
 	pj := j - i
-	if pj < 0 || s.mat.K+1 <= pj {
+	if s.mat.K+1 <= pj {
 		return 0
 	}
 	return s.mat.Data[i*s.mat.Stride+pj]
@@ -221,7 +221,7 @@ func (s *SymBandDense) set(i, j int, v float64) {
 		i, j = j, i
 	}
 	pj := j - i
-	if pj < 0 || s.mat.K+1 <= pj {
+	if s.mat.K+1 <= pj {
 		panic(ErrBandSet)
 	}
 	s.mat.Data[i*s.mat.Stride+pj] = v
--- a/mat/index_no_bound_checks.go
+++ b/mat/index_no_bound_checks.go
@@ -181,3 +181,48 @@ func (b *BandDense) set(i, j int, v float64) {
 	pj := j + b.mat.KL - i
 	b.mat.Data[i*b.mat.Stride+pj] = v
 }
+
+// At returns the element at row i, column j.
+func (s *SymBandDense) At(i, j int) float64 {
+	if uint(i) >= uint(s.mat.N) {
+		panic(ErrRowAccess)
+	}
+	if uint(j) >= uint(s.mat.N) {
+		panic(ErrColAccess)
+	}
+	return s.at(i, j)
+}
+
+func (s *SymBandDense) at(i, j int) float64 {
+	if i > j {
+		i, j = j, i
+	}
+	pj := j - i
+	if pj < 0 || s.mat.K+1 <= pj {
+		return 0
+	}
+	return s.mat.Data[i*s.mat.Stride+pj]
+}
+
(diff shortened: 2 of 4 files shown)


mat/index_bound_checks.go
```diff
@@ -180,3 +180,48 @@ func (b *BandDense) set(i, j int, v float64) {
...
+		panic(ErrRowAccess)
+	}
+	if uint(j) >= uint(s.mat.N) {
+		panic(ErrColAccess)
+	}
+	if i > j {
+		i, j = j, i
+	}
+	pj := j - i
+	if pj < 0 || s.mat.K+1 <= pj {
```
btracey: Is it possible pj < 0? this seems impossible given the above lines

mat/index_bound_checks.go
```diff
@@ -180,3 +180,48 @@ func (b *BandDense) set(i, j int, v float64) {
...
```
btracey: Same here

mat/index_no_bound_checks.go
```diff
@@ -181,3 +181,48 @@ func (b *BandDense) set(i, j int, v float64) {
...
+	}
+	return s.at(i, j)
+}
+
+func (s *SymBandDense) at(i, j int) float64 {
+	if i > j {
+		i, j = j, i
+	}
+	pj := j - i
+	if pj < 0 || s.mat.K+1 <= pj {
```
btracey: and here

mat/index_no_bound_checks.go
```diff
@@ -181,3 +181,48 @@ func (b *BandDense) set(i, j int, v float64) {
...
+	}
+	s.set(i, j, v)
+}
+
+func (s *SymBandDense) set(i, j int, v float64) {
+	if i > j {
+		i, j = j, i
+	}
+	pj := j - i
+	if pj < 0 || s.mat.K+1 <= pj {
```
btracey: and here.

mat/symband.go
```diff
@@ -0,0 +1,127 @@
...
+	_            RawSymBander     = symBandDense
+	_            MutableSymBanded = symBandDense
+)
+
+// SymBandDense represents a symmetric band matrix in dense storage format.
+type SymBandDense struct {
+	mat blas64.SymmetricBand
+}
+
+// MutableSymBanded is a symmetric band matrix interface type that allows elements
```
btracey: symmetric banded matrix

mat/symband.go
```diff
@@ -0,0 +1,127 @@
...
+var (
+	symBandDense *SymBandDense
+	_            Matrix           = symBandDense
+	_            Symmetric        = symBandDense
+	_            Banded           = symBandDense
+	_            RawSymBander     = symBandDense
+	_            MutableSymBanded = symBandDense
+)
+
+// SymBandDense represents a symmetric band matrix in dense storage format.
```
btracey: This doesn't seem to have a `SymBanded` interface. I would recommend a `SymBand() (ku int)` or variation in order to mimic the `Symmetric` style

mat/symband.go
```diff
@@ -0,0 +1,127 @@
...
+type SymBandDense struct {
+	mat blas64.SymmetricBand
+}
+
+// MutableSymBanded is a symmetric band matrix interface type that allows elements
+// to be altered.
+type MutableSymBanded interface {
+	Symmetric
+	Bandwidth() (kl, ku int)
```
btracey: Then this can be `SymBanded` with `SetSymBand`

mat/symband.go
```diff
@@ -0,0 +1,127 @@
...
+// The data must be arranged in row-major order constructed by removing the zeros
+// from the rows outside the band and aligning the diagonals. SymBandDense matrices
+// are stored in the upper triangle. For example, the matrix
+//    1  2  3  0  0  0
+//    2  4  5  6  0  0
+//    3  5  7  8  9  0
+//    0  6  8 10 11 12
+//    0  0  9 11 13 14
+//    0  0  0 12 14 15
+// becomes (* entries are never accessed)
```
btracey: This storage isn't correct. The "used' elements slide down.
````
*  *   3 
*  2   5
1  4   7
```

etc. (See http://www.netlib.org/lapack/explore-html/d5/d3d/dpbtf2_8f_source.html)

mat/index_no_bound_checks.go
```diff
@@ -181,3 +181,48 @@ func (b *BandDense) set(i, j int, v float64) {
...
+
+func (s *SymBandDense) at(i, j int) float64 {
+	if i > j {
+		i, j = j, i
+	}
+	pj := j - i
+	if pj < 0 || s.mat.K+1 <= pj {
+		return 0
+	}
+	return s.mat.Data[i*s.mat.Stride+pj]
```
btracey: This layout is not correct.

mat/symband.go
```diff
@@ -0,0 +1,127 @@
...
+			K:      k,
+			Stride: bc,
+			Uplo:   blas.Upper,
+			Data:   data,
+		},
+	}
+}
+
+// NewDiagonal is a convenience function that returns a diagonal matrix represented by a
+// SymBandDense. The length of data must be n otherwise NewDiagonal will panic.
```
btracey: or data can be nil.

mat/index_bound_checks.go
kortschak: Yes, this should be `pj := j + b.mat.K - i` throughout.

mat/index_no_bound_checks.go
kortschak: I think it is.

mat/symband.go
kortschak: I was thinking about that and you get the same behaviour from composition.

mat/symband.go
kortschak: The normal way that I have seen it written is "band".

mat/symband.go
kortschak: See comment above about composition.

mat/symband.go
kortschak: Those are column major. We are row major. It should be the case that you can convert a BLAS band matrix to a upper representation BLAS symband matrix by keeping the stride the same, and offsetting by kL. Currently, that is possible, the storage you have would not allow that.

mat/index_bound_checks.go
kortschak: Done.

mat/index_bound_checks.go
kortschak: Done.

mat/index_no_bound_checks.go
kortschak: Done.

mat/index_no_bound_checks.go
kortschak: Done.

mat/symband.go
kortschak: Done.

mat/index_bound_checks.go
kortschak: Grumble, I should read properly.

mat/symband.go
btracey: I see your point, but ... 

We have to do this because this is the C interface definition? For the other types, the place of the elements in the matrix is the same, they're just in a different order in the linear array because of row major. I think it's going to be much easier to have our layout match Lapack, rather than having to understand the effects of these new quasi-transposes for every new function. I would happily re-write our BLAS implementation to make this the case (so bands are still down the row), but I don't know how this would interact with the C wrappers.

kortschak:
Have a look at chapter 2 of the blas book, it describes row major storage, and so does the band documentation in the blas/gonum package, which states that the bands are stored in row major.

btracey:
What is the "blas book"?

I understand that this PR matches our current layout. I'm just really dreading trying to translate Lapack functions when the layouts are so different, and wondering if there's any way to save that pain. I remember the banded BLAS functions being more difficult to translate for this reason, and I can only imagine it'll get worse from there. 

If the storage has to be like this, then I guess we'll deal with it.

kortschak:
http<http://www.netlib.org/blas/blast-forum/>://www.netlib.org/blas/<http://www.netlib.org/blas/blast-forum/>blast<http://www.netlib.org/blas/blast-forum/>-<http://www.netlib.org/blas/blast-forum/>forum<http://www.netlib.org/blas/blast-forum/>/<http://www.netlib.org/blas/blast-forum/>

mat/symband.go
btracey: Okay.

btracey:
So I guess we're eating the cost later then.

mat/symband.go
btracey: How so? Let's say we wanted

```
func (sb *SymBandDense) Add(a,b SymBanded){}
```

How would this method signature look otherwise?

mat/symband.go
btracey: Okay

kortschak:
That horse has well and truly bolted.

btracey:
Yea, okay, I guess it has to be like this. Not going to be fun though.

kortschak:
Sure, I think I should revise introduction of the fortran package, and add the band a packed formats. That was the reason I wrote it.

btracey:
Well, there exists another way, which is represent it in the column-major version, and then hack the netlib implementation. This is what I was originally thinking we could do.

This way does not work though, there's a good reason why the storage is the way it is. Worked partially through Dpbtf2, and now I understand why this is the correct way.

kortschak:
We need the band representations to agree, at the moment they do, if they don't we are opening our selves up to a world of pain on gonum-dev and issues here.

btracey:
I agree. I was previously suggesting we could change the other representation. This way is better.

PR #135, opened 2017-07-15
blas/gonum: add Zaxpy

vladimir-ch:
Starting to slowly add complex BLAS routines.

### DIFF
--- a/blas/testblas/zaxpy.go
+++ b/blas/testblas/zaxpy.go
@@ -1,0 +1,177 @@
+// Copyright ©2017 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package testblas
+
+import (
+	"fmt"
+	"math"
+	"testing"
+)
+
+type Zaxpyer interface {
+	Zaxpy(n int, alpha complex128, x []complex128, incX int, y []complex128, incY int)
+}
+
+func ZaxpyTest(t *testing.T, impl Zaxpyer) {
+	for tc, test := range []struct {
+		alpha complex128
+		x, y  []complex128
+
+		want    []complex128 // Result when both increments have the same sign.
+		wantRev []complex128 // Result when the increments have opposite sign.
+	}{
+		{
+			alpha:   0,
+			x:       []complex128{1 + 2i, 3 + 4i, 5 + 6i, 7 + 8i, 9 + 10i, 11 + 12i, 13 + 14i, 15 + 16i, 17 + 18i, 19 + 20i, 21 + 22i, 23 + 24i},
+			y:       []complex128{30 + 31i, 33 + 34i, 36 + 37i, 39 + 40i, 42 + 43i, 45 + 46i, 48 + 49i, 51 + 52i, 54 + 55i, 57 + 58i, 60 + 61i, 63 + 64i},
+			want:    []complex128{30 + 31i, 33 + 34i, 36 + 37i, 39 + 40i, 42 + 43i, 45 + 46i, 48 + 49i, 51 + 52i, 54 + 55i, 57 + 58i, 60 + 61i, 63 + 64i},
+			wantRev: []complex128{30 + 31i, 33 + 34i, 36 + 37i, 39 + 40i, 42 + 43i, 45 + 46i, 48 + 49i, 51 + 52i, 54 + 55i, 57 + 58i, 60 + 61i, 63 + 64i},
+		},
+		{
+			alpha:   1,
+			x:       []complex128{1 + 2i, 3 + 4i, 5 + 6i, 7 + 8i, 9 + 10i, 11 + 12i, 13 + 14i, 15 + 16i, 17 + 18i, 19 + 20i, 21 + 22i, 23 + 24i},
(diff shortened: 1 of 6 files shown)


blas/testblas/zaxpy.go
```diff
@@ -0,0 +1,119 @@
...
+			alpha:   3 + 7i,
+			x:       []complex128{2 + 3i, 3 + 4i, 4 + 5i},
+			y:       []complex128{20 + 30i, 30 + 40i, 40 + 50i},
+			want:    []complex128{5 + 53i, 11 + 73i, 17 + 93i},
+			wantRev: []complex128{-3 + 73i, 11 + 73i, 25 + 73i},
+		},
+	} {
+		n := len(test.x)
+		for _, incX := range []int{-10, -1, 1, 10} {
+			for _, incY := range []int{-10, -1, 1, 10} {
```
btracey: It's probably a good idea to have these be non-equal between x and y, just in case.

blas/testblas/zaxpy.go
```diff
@@ -0,0 +1,119 @@
...
+					for i, v := range test.wantRev {
+						want[i*aincY] = v
+					}
+				}
+
+				impl.Zaxpy(n, test.alpha, x, incX, y, incY)
+
+				prefix := fmt.Sprintf("Case %v (incX=%v,incY=%v):", tc, incX, incY)
+
+				equal := true
```
btracey: We're going to have this kind of equality check a lot. Should we start a `complexes` package to correspond to `floats`?

blas/testblas/zaxpy.go
```diff
@@ -0,0 +1,119 @@
...
+
+func ZaxpyTest(t *testing.T, impl Zaxpyer) {
+	rnd := rand.New(rand.NewSource(1))
+
+	for tc, test := range []struct {
+		alpha complex128
+		x, y  []complex128
+
+		want    []complex128
+		wantRev []complex128
```
btracey: Could you comment this field?

blas/testblas/zaxpy.go
```diff
@@ -0,0 +1,119 @@
...
+					aincX = -aincX
+				}
+				aincY := incY
+				if aincY < 0 {
+					aincY = -aincY
+				}
+
+				x := make([]complex128, (n-1)*aincX+1)
+				for i := range x {
+					x[i] = complex(rnd.NormFloat64(), rnd.NormFloat64())
```
btracey: Populate with NaN instead?

btracey:
It looks like this game is starting again

blas/testblas/zaxpy.go
```diff
@@ -0,0 +1,119 @@
...
+			want:    []complex128{22 + 33i, 33 + 44i, 44 + 55i},
+			wantRev: []complex128{24 + 35i, 33 + 44i, 42 + 53i},
+		},
+		{
+			alpha:   3 + 7i,
+			x:       []complex128{2 + 3i, 3 + 4i, 4 + 5i},
+			y:       []complex128{20 + 30i, 30 + 40i, 40 + 50i},
+			want:    []complex128{5 + 53i, 11 + 73i, 17 + 93i},
+			wantRev: []complex128{-3 + 73i, 11 + 73i, 25 + 73i},
+		},
```
Kunde21: The main loop in `asm/c128` is 4-wide and the future AVX version will be 8-wide, so there should be test entries of length 10+ in here.

blas/testblas/zaxpy.go
Kunde21: We had discussed this in the [major breaking changes thread](https://groups.google.com/forum/#!topic/gonum-dev/4oLmSvtDuZY) but didn't come to an agreement on a single package (`Vector/<type>`?) or mirrored packages (`floats` and `complexes`).  If we're going to start building out the complex blas routines, we should get the vector routines organized, too.

blas/testblas/zaxpy.go
vladimir-ch: Done.

blas/testblas/zaxpy.go
vladimir-ch: Done.

blas/testblas/zaxpy.go
vladimir-ch: Done.

blas/testblas/zaxpy.go
vladimir-ch: At the moment I will do it manually. If we pre-fill the slices with NaNs, `complexes.Equal` would not work here anyway.

vladimir-ch:
PTAL

blas/testblas/zaxpy.go
btracey: Could you at least make it a function in here?

btracey: And `complexes.Same` presumably would work.

vladimir-ch: Sorry, I don't know our own packages, that would of course work.
Done.

blas/testblas/zaxpy.go
```diff
@@ -0,0 +1,178 @@
...
+			prefix := fmt.Sprintf("Case %v (incX=%v,incY=%v):", tc, incX, incY)
+
+			if !zsame(x, xCopy) {
+				t.Errorf("%v: unexpected modification of x", prefix)
+			}
+
+			if !zsame(y, want) {
+				t.Errorf("%v: unexpected y:\nwant %v\ngot %v", prefix, want, y)
+			}
+			t.Log(prefix)
```
kortschak: What is this for?

vladimir-ch: Sorry, removed now.

Issue #138, opened 2017-07-17
blas,mat: consider changes necessary to allow CUDA/OpenGL backend


kortschak:
This is something that has been discussed on an off over the past few years, but we probably should devise an approach before we get too far along.

There are CUDA/OpenGL BLAS implementations that we could use, but we likely lose much of the benefit of the GPU if we hold the matrix in system memory, transfer to the card, perform the operation and return it to the system for each matrix operation. Ideally, we would want to move the data across, perform multiple operations and only return the results to the main system when completed.

The upshot of that is that we want a structure that is similar to the blas handle-like `mat` fields in the mat package, but with knowledge of where the data is actually held, and whether it is dirty. User access to the data would wait on GPU completion if the data were being operated on by the GPU and non-GPU writes (including `Set*` method calls) would mark the data as dirty (after having waited on any GPU operations and returned the data from the card).

This set of requirements is inconsistent with the current dumb types in blas64 which make up the `mat` fields in the mat package. What seems to be needed is a change from `mat blas64.T` to `mat blas64.I` and addition of necessary methods. With this, a GPU-aware blas package could be built. I'm concerned though about the depth of interface layering that we are building up.

btracey:
> What seems to be needed is a change from mat blas64.T to mat blas64.I and addition of necessary methods.

I do not understand

kortschak:
We make the field an interface value rather than a concrete value. Off the top of my head, the simples way to do it is to make each of the `mat` fields the appropriate `Raw*` interface and make the data types implement those interfaces. That way, a CUDA/OpenGL implementation can wrap the blas64 data type and hold the relevant handles and dirty flags.

kortschak:
Having slept on this, I don't think this should happen in the blas64 interface layer, but rather at the level the mat package operates, but in another package. The logic required to keep consistency between the implementation and the blas-CUDA/OpenGL handle would add too much weight both from a maintenance perspective and to performance of non-GPU users.

So the new suggestion is to add a cuda and/or opengl package that implements (in the first instance) a `mat.Matrix based` on `blas64.General`. This implementation would hold the data location and dirtiness status, and be a `mat.RawMatrixer` to facilitate performant interop with mat. Initially, as a PoC, it would implement `Add`, `Mul`, `MulElem` and `Sub`.

One issue that needs addressing for this work is that CUDA (at least) takes column-major matrices, so the work in https://github.com/gonum/lapack/pull/69, or something like it would need to be merged first.

[...]

PR #141, opened 2017-07-20
blas64: add row/col major conversion methods

kortschak:
Reference #138.

This covers the types that are supported in mat and makes no change to the API other than adding the conversion tools. My sense of symmetry says that the row major types should be labelled as such, but I could get over on the basis that we are natively row major and the column major types are for external use.

asm-dependent code can come later as can generation of `float32` and complex equivalents.

### DIFF
--- a/blas/blas64/conv.go
+++ b/blas/blas64/conv.go
@@ -1,0 +1,312 @@
+// Copyright ©2015 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package blas64
+
+import "gonum.org/v1/gonum/blas"
+
+// GeneralCols represents a matrix using the conventional column-major storage scheme.
+type GeneralCols General
+
+// From fills the receiver with elements from a. The receiver
+// must have the same dimensions as a and have adequate backing
+// data storage.
+func (t GeneralCols) From(a General) {
+	if t.Rows != a.Rows || t.Cols != a.Cols {
+		panic("blas64: mismatched dimension")
+	}
+	if len(t.Data) < (t.Cols-1)*t.Stride+t.Rows {
+		panic("blas64: short data slice")
+	}
+	for i := 0; i < a.Rows; i++ {
+		for j, v := range a.Data[i*a.Stride : i*a.Stride+a.Cols] {
+			t.Data[i+j*t.Stride] = v
+		}
+	}
+}
+
+// From fills the receiver with elements from a. The receiver
+// must have the same dimensions as a and have adequate backing
+// data storage.
+func (t General) From(a GeneralCols) {
+	if t.Rows != a.Rows || t.Cols != a.Cols {
+		panic("blas64: mismatched dimension")
+	}
+	if len(t.Data) < (t.Rows-1)*t.Stride+t.Cols {
+		panic("blas64: short data slice")
+	}
+	for i := 0; i < a.Cols; i++ {
+		for j, v := range a.Data[i*a.Stride : i*a.Stride+a.Rows] {
+			t.Data[i+j*t.Stride] = v
+		}
+	}
+}
+
+// SymmetricCols represents a matrix using the conventional column-major storage scheme.
+type SymmetricCols Symmetric
+
+// From fills the receiver with elements from a. The receiver
+// must have the same dimensions and uplo as a and have adequate
+// backing data storage.
+func (t SymmetricCols) From(a Symmetric) {
+	if t.N != a.N {
+		panic("blas64: mismatched dimension")
+	}
+	if t.Uplo != a.Uplo {
+		panic("blas64: mismatched BLAS uplo")
+	}
+	switch a.Uplo {
+	default:
+		panic("blas64: bad BLAS uplo")
+	case blas.Upper:
+		for i := 0; i < a.N; i++ {
+			for j := i; j < a.N; j++ {
+				t.Data[i+j*t.Stride] = a.Data[i*a.Stride+j]
+			}
+		}
+	case blas.Lower:
+		for i := 0; i < a.N; i++ {
+			for j := 0; j <= i; j++ {
+				t.Data[i+j*t.Stride] = a.Data[i*a.Stride+j]
+			}
+		}
+	}
+}
+
+// From fills the receiver with elements from a. The receiver
+// must have the same dimensions and uplo as a and have adequate
+// backing data storage.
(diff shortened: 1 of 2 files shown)


btracey:
I've been seeing performance issues with the lapack row major implementation (a code of mine is 3x slower using cgo.Lapack).  A lot of the BLAS functions (at least) can be performed calling into the column-major implementation with row-major data using appropriate tweaks to the transpose calls (etc.). I wonder if it's worth trying to make these transformations for the cgo.Implementation in both BLAS and Lapack (that is, just the xxx64 routines), having the implementations call the column-major version,  and then only using the transformation types when we can't get around it otherwise. I think this would fix a bunch of the performance issues, and issues with the row-major lapack implementation.

kortschak:
>I've been seeing performance issues with the lapack row major implementation (a code of mine is 3x slower using cgo.Lapack).

Do you mean with the lapacke implementation?

OpenBLAS (and others already do argument swapping and intelligent transpose) so it would be interesting to see which routines you are using that are causing the slow down. However, I don't think it's our job to fix that in gonum, though maybe filing issues/sending PRs to OpenBLAS/Netlib may be.

kortschak:
A question about implementation here before I go ahead and add `TriangularBand` routines. The symmetric band and trangular band are special cases of band and can be implemented as such with a conversion from the actual to `Band`/`BandCols` and then back again after the operation. This reduces the code, but may be a little less clear and possible very marginally slower. How do you feel about that?

btracey:
> However, I don't think it's our job to fix that in gonum

I understand the position, but the point of these changes is to enable efficient use of GPU/future tech that is only ColMajor. If we have to copy the data from RowMajor to ColMajor, a lot of that potential benefit goes away. An alternate strategy is to provide that interface layer ourself, at least for the blas64, lapack64 routines.

kortschak:
Because of the differences between the GPU BLAS implementation APIs and the CPU BLAS API, I am not convinced that the GPU implementations will be *directly* interoperable with what we have (i.e. as a drop in replacement), so these issues are essentially orthogonal. It is true that we will need to do the kind of work that LAPACKE does before handing off to LAPACK in our GPU client code, but we don't need to do it in our LAPACKE-calling code, and we will not (in the long term) do a better job than LAPACKE does there (i.e. after the transpose bugs are fixed), for a significant maintenance burden.

The plan for the gonum GPU representations is that they will be stored internally as col-major, and only return row-major when calling `RawMatrix` or setting a raw matrix with `SetRawMAtrix` - this adds a small cost to the already existing cost of returning the data from the card (which will be minimised in GPU-only computations when the implementation knows that the data is in the needed location).

All up, there are issues that need to be addressed, but not here.

kortschak:
One thing that may need to happen though, since `mat.Vector` is used as a concrete type, we will probably need to make it an interface and rename the concrete type to `mat.DenseVector`.

blas/blas64/conv.go
```diff
@@ -0,0 +1,312 @@
...
+// must have the same dimensions as a and have adequate backing
+// data storage.
+func (t General) From(a GeneralCols) {
+	if t.Rows != a.Rows || t.Cols != a.Cols {
+		panic("blas64: mismatched dimension")
+	}
+	if len(t.Data) < (t.Rows-1)*t.Stride+t.Cols {
+		panic("blas64: short data slice")
+	}
+	for i := 0; i < a.Cols; i++ {
```
btracey: Swap `i` with `j` and make it `t.Data[i*t.Stride +j]`. We should keep it that `i` is the row index and `j` is the column index.

blas/blas64/conv.go
```diff
@@ -0,0 +1,312 @@
...
+	}
+	if t.Uplo != a.Uplo {
+		panic("blas64: mismatched BLAS uplo")
+	}
+	switch a.Uplo {
+	default:
+		panic("blas64: bad BLAS uplo")
+	case blas.Upper:
+		for i := 0; i < a.N; i++ {
+			for j := i; j < a.N; j++ {
```
btracey: Is there a reason to drop the `range` syntax?

btracey:
Changing `Vector` is a good idea, the same issues will come up when dealing with sparse matrices.

No comment on the `RawMatrix` and other storage issues ... there isn't a plan yet.

blas/blas64/conv.go
kortschak: Will do.

blas/blas64/conv.go
kortschak: The initial implementation is intended to be simple. I will replace the inner loops with calls the asm equivalents at a later stage.

kortschak:
The RawMatrix/SetRawMatrix will happen in the opencl matrix code. What about code reuse for band types?

btracey:
I think the code reuse is a good idea for Banded. The overhead should pale in comparison to the copy, and the band conversions are tricky, so one implementation is best.

Do you think there's merit in combining the triangular / symmetric for similar reasons?

blas/blas64/conv.go
kortschak: Done

kortschak:
The reuse made the tests easier to write as well.

kortschak:
I don't think the symmetric/triangular merge is worth it though.

kortschak:
PTAL

blas/blas64/conv.go
```diff
@@ -0,0 +1,423 @@
...
+		}
+	}
+}
+
+// BandCols represents a matrix using the band column-major storage scheme.
+type BandCols Band
+
+// From fills the receiver with elements from a. The receiver
+// must have the same dimensions and bandwidth as a and have
+// adequate backing data storage.
```
btracey: Is it worth saying here that the "stored matrix size" will be different?  The matrix sizes are the same, but the storage sizes are not.

blas/blas64/conv.go
```diff
@@ -0,0 +1,423 @@
+// Copyright ©2015 The gonum Authors. All rights reserved.
```
btracey: Just confirming this is the right year.

blas/blas64/conv.go
kortschak: I don't think so. Maybe if we get queries. In the meantime, I'd like to keep this very terse - people should probably not be using this much.

kortschak:
So just checking that you're happier with type X and type XCols rather than XRows and XCols?

btracey:
Yes. We're row major, and it's good to be biased in that direction I think.

PR #144, opened 2017-07-21
add triangle distribution and associated tests

bpgray:
This adds a triangle distribution to the package, which is useful for business simulations when there is limited sample data available on input variables.

### DIFF
--- a/stat/distuv/triangle.go
+++ b/stat/distuv/triangle.go
@@ -136,7 +136,7 @@ func (t Triangle) Rand() float64 {
 // Skewness returns the skewness of the distribution.
 func (t Triangle) Skewness() float64 {
 	n := math.Sqrt2 * (t.a + t.b - 2*t.c) * (2*t.a - t.b - t.c) * (t.a - 2*t.b + t.c)
-	d := 5 * math.Pow(t.a*t.a+t.b*t.b+t.c*t.c-t.a*t.b-t.a*t.c-t.b*t.c, 3/2)
+	d := 5 * math.Pow(t.a*t.a+t.b*t.b+t.c*t.c-t.a*t.b-t.a*t.c-t.b*t.c, 3.0/2.0)
 
 	return n / d
 }
--- a/stat/distuv/triangle.go
+++ b/stat/distuv/triangle.go
@@ -1,0 +1,177 @@
+// Copyright ©2017 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package distuv
+
+import (
+	"math"
+	"math/rand"
+)
+
+// UnitTriangle is an instantiation of the triangle distribution with A=0, C=0.5, and B=1
+var UnitTriangle = Triangle{A: 0, C: 0.5, B: 1}
+
+// Triangle represents a triangle distribution (https://en.wikipedia.org/wiki/Triangular_distribution).
+type Triangle struct {
+	A      float64
+	C      float64
+	B      float64
+	Source *rand.Rand
+}
+
+// CDF computes the value of the cumulative density function at x.
+func (t Triangle) CDF(x float64) float64 {
+	if x <= t.A {
+		return 0
+	}
+	if x <= t.C {
+		return math.Pow(x-t.A, 2) / ((t.B - t.A) * (t.C - t.A))
+	}
+	if x < t.B {
+		return 1 - math.Pow(t.B-x, 2)/((t.B-t.A)*(t.B-t.C))
+	}
+	return 1
+}
+
+// Entropy returns the entropy of the distribution.
+func (t Triangle) Entropy() float64 {
+	return 1.0/2.0 + math.Log((t.B-t.A)/2)
+}
+
+// ExKurtosis returns the excess kurtosis of the distribution.
+func (Triangle) ExKurtosis() float64 {
+	return -3.0 / 5.0
+}
+
+// Fit is not appropriate for Triangle, because the distribution is generally used when there is little data.
+
+// LogProb computes the natural logarithm of the value of the probability density function at x.
+func (t Triangle) LogProb(x float64) float64 {
+	return math.Log(t.Prob(x))
+}
+
+// MarshalParameters implements the ParameterMarshaler interface
+func (t Triangle) MarshalParameters(p []Parameter) {
+	if len(p) != t.NumParameters() {
+		panic("triangle: improper parameter length")
+	}
+	p[0].Name = "A"
+	p[0].Value = t.A
+	p[1].Name = "C"
+	p[1].Value = t.C
+	p[2].Name = "B"
+	p[2].Value = t.B
+	return
+}
+
+// Mean returns the mean of the probability distribution.
+func (t Triangle) Mean() float64 {
+	return (t.A + t.B + t.C) / 3
+}
+
+// Median returns the median of the probability distribution.
+func (t Triangle) Median() float64 {
+	if t.C >= (t.A+t.B)/2 {
+		return t.A + math.Sqrt((t.B-t.A)*(t.C-t.A)/2)
+	} else {
+		return t.B - math.Sqrt((t.B-t.A)*(t.B-t.C)/2)
+	}
+}
+
+// Mode returns the mode of the probability distribution.
+func (t Triangle) Mode() float64 {
+	return t.C
+}
+
+// NumParameters returns the number of parameters in the distribution.
+func (Triangle) NumParameters() int {
+	return 3
+}
+
+// Prob computes the value of the probability density function at x.
+func (t Triangle) Prob(x float64) float64 {
+	if x < t.A {
+		return 0
+	}
+	if x < t.C {
+		return 2 * (x - t.A) / ((t.B - t.A) * (t.C - t.A))
+	}
+	if x == t.C {
+		return 2 / (t.B - t.A)
+	}
+	if x <= t.B {
+		return 2 * (t.B - x) / ((t.B - t.A) * (t.B - t.C))
+	}
+	return 0
+}
+
+// Quantile returns the inverse of the cumulative probability distribution.
+func (t Triangle) Quantile(p float64) float64 {
+	if p < 0 || p > 1 {
+		panic(badPercentile)
+	}
+
+	f := (t.C - t.A) / (t.B - t.A)
+
+	if p < f {
+		return t.A + math.Sqrt(p*(t.B-t.A)*(t.C-t.A))
+	} else {
+		return t.B - math.Sqrt((1-p)*(t.B-t.A)*(t.B-t.C))
+	}
+}
+
+// Rand returns a random sample drawn from the distribution.
+func (t Triangle) Rand() float64 {
+	var rnd float64
+	if t.Source == nil {
+		rnd = rand.Float64()
+	} else {
+		rnd = t.Source.Float64()
+	}
+
+	return t.Quantile(rnd)
+}
+
+// Skewness returns the skewness of the distribution.
+func (t Triangle) Skewness() float64 {
+	n := math.Sqrt2 * (t.A + t.B - 2*t.C) * (2*t.A - t.B - t.C) * (t.A - 2*t.B + t.C)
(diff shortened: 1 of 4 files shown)


stat/distuv/triangle.go
```diff
@@ -0,0 +1,177 @@
...
+// license that can be found in the LICENSE file.
+
+package distuv
+
+import (
+	"math"
+	"math/rand"
+)
+
+// UnitTriangle is an instantiation of the triangle distribution with A=0, C=0.5, and B=1
```
kortschak: Is there a reason that we need to provide this?

stat/distuv/triangle.go
```diff
@@ -0,0 +1,177 @@
...
+var UnitTriangle = Triangle{A: 0, C: 0.5, B: 1}
+
+// Triangle represents a triangle distribution (https://en.wikipedia.org/wiki/Triangular_distribution).
+type Triangle struct {
+	A      float64
```
kortschak: Please write this as `A, B, C float64`.

stat/distuv/triangle.go
```diff
@@ -0,0 +1,177 @@
...
+	C      float64
+	B      float64
+	Source *rand.Rand
+}
+
+// CDF computes the value of the cumulative density function at x.
+func (t Triangle) CDF(x float64) float64 {
+	if x <= t.A {
```
kortschak: This would be clearer as a switch, also, for n^2, please just multiply.

```
switch {
case x <= t.A:
    return 0
case x <= t.C:
    d := x-t.A
    return (d*d) / ((t.B - t.A) * (t.C - t.A))
case x < t.B:
    d := x-t.B
    return 1 - (d*d)/((t.B-t.A)*(t.B-t.C))
default:
    return 1
}
```

stat/distuv/triangle.go
```diff
@@ -0,0 +1,177 @@
...
+	}
+	if x < t.B {
+		return 1 - math.Pow(t.B-x, 2)/((t.B-t.A)*(t.B-t.C))
+	}
+	return 1
+}
+
+// Entropy returns the entropy of the distribution.
+func (t Triangle) Entropy() float64 {
+	return 1.0/2.0 + math.Log((t.B-t.A)/2)
```
kortschak: `return math.Log(t.B-t.A) - 0.19314718055994530941723212145817656807550013436025525412 // ln(2) - 0.5`

stat/distuv/triangle.go
```diff
@@ -0,0 +1,177 @@
...
+	if len(p) != t.NumParameters() {
+		panic("triangle: improper parameter length")
+	}
+	p[0].Name = "A"
+	p[0].Value = t.A
+	p[1].Name = "C"
+	p[1].Value = t.C
+	p[2].Name = "B"
+	p[2].Value = t.B
+	return
```
kortschak: Delete this.

stat/distuv/triangle.go
```diff
@@ -0,0 +1,177 @@
...
+// Mean returns the mean of the probability distribution.
+func (t Triangle) Mean() float64 {
+	return (t.A + t.B + t.C) / 3
+}
+
+// Median returns the median of the probability distribution.
+func (t Triangle) Median() float64 {
+	if t.C >= (t.A+t.B)/2 {
+		return t.A + math.Sqrt((t.B-t.A)*(t.C-t.A)/2)
+	} else {
```
kortschak: Remove the else and de-indent.

stat/distuv/triangle.go
```diff
@@ -0,0 +1,177 @@
...
+}
+
+// NumParameters returns the number of parameters in the distribution.
+func (Triangle) NumParameters() int {
+	return 3
+}
+
+// Prob computes the value of the probability density function at x.
+func (t Triangle) Prob(x float64) float64 {
+	if x < t.A {
```
kortschak: A switch here please.

stat/distuv/triangle.go
```diff
@@ -0,0 +1,177 @@
...
+func (t Triangle) Quantile(p float64) float64 {
+	if p < 0 || p > 1 {
+		panic(badPercentile)
+	}
+
+	f := (t.C - t.A) / (t.B - t.A)
+
+	if p < f {
+		return t.A + math.Sqrt(p*(t.B-t.A)*(t.C-t.A))
+	} else {
```
kortschak: Remove the else and de-indent.

stat/distuv/triangle.go
```diff
@@ -0,0 +1,177 @@
...
+		rnd = t.Source.Float64()
+	}
+
+	return t.Quantile(rnd)
+}
+
+// Skewness returns the skewness of the distribution.
+func (t Triangle) Skewness() float64 {
+	n := math.Sqrt2 * (t.A + t.B - 2*t.C) * (2*t.A - t.B - t.C) * (t.A - 2*t.B + t.C)
+	d := 5 * math.Pow(math.Pow(t.A, 2)+math.Pow(t.B, 2)+math.Pow(t.C, 2)-t.A*t.B-t.A*t.C-t.B*t.C, 3/2)
```
kortschak: `5 * math.Pow(t.A*t.A+t.B*t.B+t.C*t.C-t.A*t.B-t.A*t.C-t.B*t.C, 3/2)`

stat/distuv/triangle.go
```diff
@@ -0,0 +1,177 @@
...
+	}
+
+	t.A = p[0].Value
+	t.C = p[1].Value
+	t.B = p[2].Value
+}
+
+// Variance returns the variance of the probability distribution.
+func (t Triangle) Variance() float64 {
+	return (math.Pow(t.A, 2) + math.Pow(t.B, 2) + math.Pow(t.C, 2) - t.A*t.B - t.A*t.C - t.B*t.C) / 18
```
kortschak: `(t.A*T.A + t.B*t.B + t.C*t.C - t.A*t.B - t.A*t.C - t.B*t.C) / 18`

stat/distuv/triangle.go
```diff
@@ -0,0 +1,177 @@
...
+
+import (
+	"math"
+	"math/rand"
+)
+
+// UnitTriangle is an instantiation of the triangle distribution with A=0, C=0.5, and B=1
+var UnitTriangle = Triangle{A: 0, C: 0.5, B: 1}
+
+// Triangle represents a triangle distribution (https://en.wikipedia.org/wiki/Triangular_distribution).
```
kortschak: There are constraints on a, b and c (-inf < a < inf, a < b; a <= c <= b) that are not mentioned here and are never checked for. It seems to me that this should have a `NewTriangle(a, b, c float64, rnd *rand.Rand) Triangle` that does this check (and also in the parameter marshaling), and the fields should be unexported.

stat/distuv/triangle.go
```diff
@@ -9,34 +9,47 @@ import (
...
-var UnitTriangle = Triangle{A: 0, C: 0.5, B: 1}
-
 // Triangle represents a triangle distribution (https://en.wikipedia.org/wiki/Triangular_distribution).
 type Triangle struct {
-	A      float64
-	C      float64
-	B      float64
-	Source *rand.Rand
+	a, b, c float64
+	Source  *rand.Rand
```
kortschak: If this is exported, you can leave it out of the constructor signature.

stat/distuv/triangle.go
```diff
@@ -9,34 +9,47 @@ import (
...
+}
+
+// Triangle constructs a new triangle distribution with lower limit a, upper limit b, and mode c.
+// Constraints are a < b and a ≤ c ≤ b.
+// This distribution is uncommon in nature, but may be useful for simulation
```
kortschak: Final period please.

stat/distuv/triangle.go
```diff
@@ -51,37 +64,22 @@ func (t Triangle) LogProb(x float64) float64 {
 	return math.Log(t.Prob(x))
 }
 
-// MarshalParameters implements the ParameterMarshaler interface
-func (t Triangle) MarshalParameters(p []Parameter) {
```
kortschak: Is this not useful? It can remain, just make sure the bounds are checked.

stat/distuv/triangle.go
```diff
@@ -12,13 +12,13 @@ import (
 // Triangle represents a triangle distribution (https://en.wikipedia.org/wiki/Triangular_distribution).
 type Triangle struct {
 	a, b, c float64
-	Source  *rand.Rand
+	src     *rand.Rand
```
kortschak: Sorry, I should have been more clear; make it public and leave it out of the signature.

stat/distuv/triangle.go
```diff
@@ -162,6 +175,9 @@ func (t *Triangle) UnmarshalParameters(p []Parameter) {
 		panic("triangle: " + panicNameMismatch)
 	}
 
+	// verify parameters are valid before directly setting
+	NewTriangle(a, b, c, src)
```
kortschak: `a`, `b` and `c` don't exist, but please don't do it this way; factor the check out to a helper, `checkTriangleParameters` and call it from both places. Then the comment can go away (note that we enforce capitalisation and punctuation on internal comments also).

stat/distuv/triangle.go
```diff
@@ -0,0 +1,192 @@
...
+	"math/rand"
+)
+
+// Triangle represents a triangle distribution (https://en.wikipedia.org/wiki/Triangular_distribution).
+type Triangle struct {
+	a, b, c float64
+	Source  *rand.Rand
+}
+
+// Triangle constructs a new triangle distribution with lower limit a, upper limit b, and mode c.
```
btracey: s/Triangle/NewTriangle

stat/distuv/triangle.go
```diff
@@ -0,0 +1,192 @@
...
+// Constraints are a < b and a ≤ c ≤ b.
+// This distribution is uncommon in nature, but may be useful for simulation.
+func NewTriangle(a, b, c float64) Triangle {
```
btracey: The documentation says that it is acceptable for `c` to equal `a` or `b`. However, the coding of the functions returns non-sensical answers (divide by `a-c`) for instance. One of these needs to be fixed.

stat/distuv/triangle.go
```diff
@@ -0,0 +1,192 @@
...
+		d := t.b - x
+		return 1 - (d*d)/((t.b-t.a)*(t.b-t.c))
+	default:
+		return 1
+	}
+}
+
+// Entropy returns the entropy of the distribution.
+func (t Triangle) Entropy() float64 {
+	return 0.5 + math.Log(t.b-t.a) - ln2
```
btracey: use math.Ln2 instead, or define a constant for `math.Ln2-0.5`

stat/distuv/triangle.go
bpgray: I reviewed the code and all the non-sensical instances of divide by a-c (or similar) are unreachable in the case of c=a. Since it's a bit tedious to trace through, I've added a test case of c=a. It's also OK to set b=c; however, that test case would need a slightly looser tolerance on checkProbContinuous to pass, and I didn't think it was worth the code changes to prove that b=c is OK on every build.

btracey:
I think it's a good idea to have this. I've used it as a couple times because of nice mathematical properties (even though it's typically a bad modeling choice). Also, hopefully this being next to different distributions means people will start searching.

The article you link is an interesting read, but the conclusion is bad.
```
So, if you are certain that your random variable has hard upper and lower bounds, you can choose between the

- uniform,
- truncated exponential and
- median step function
```

No. If you know your distribution is bounded, and has a certain mean, a (stretched) `Beta` distribution is often a much better choice, especially over a step median. It has different constraints for it to be the maxent, but it more matches properties modelers typically want when they choose triangular (notably, something that is like a mode with the edges of the distribution looking very unlikely).

bpgray:
The triangle distribution has some pretty useful features for Monte Carlo simulations, which aren't discussed in the article linked above. Also, a triangle distribution can be a good way to generate a distribution of inputs for trying to understand a models sensitivities to various inputs. Agreed that it doesn't align with natural processes, but that's not how I use this distribution.

kortschak:
No worries. If @btracey is happy to have it here, that's good enough for me.

PR #154, opened 2017-07-28
Add Laplacian and CrossLaplacian difference functions

btracey:


btracey:
I'll fix the commit message when responding to PRs (assuming there will be one).

Fixes https://github.com/gonum/diff/issues/48

btracey:
PTAL @vladimir-ch

diff/fd/crosslaplacian.go
```diff
@@ -0,0 +1,190 @@
...
+)
+
+// CrossLaplacian computes a Laplacian-like quantity for a function of two vectors
+// at the locations x and y.
+// It computes
+//  ∇_y · ∇_x f(x,y) = \sum_i ∂^2 f(x,y)/∂x_i ∂y_i
+// The two input vector lengths must be the same.
+//
+// Finite difference formula and other options are specified by settings. If
+// settings is nil, LaplacianTwo will be estimated using the Forward formula and
```
vladimir-ch: What is `LaplacianTwo`? Also at several other places.

diff/fd/laplacian.go
```diff
@@ -0,0 +1,160 @@
...
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package fd
+
+import "sync"
+
+// Laplacian computes the Laplacian of the multivariate function f at the location
+// x. That is, Laplacian returns
+//  ∇_x · ∇_x f(x,y) = \sum_i ∂^2 f(x)/∂x_i ∂x_i
```
vladimir-ch: `∆ f(x) = ∇ · ∇ f(x) = \sum_i ∂^2 f(x)/∂x_i^2`

diff/fd/crosslaplacian.go
btracey: Sorry, this operator doesn't actually have a name as best I can tell (but it does come up). I originally called it LaplacianTwo, but decided CrossLaplacian made more sense. 

Fixed.

diff/fd/laplacian.go
btracey: Done.

diff/fd/crosslaplacian_test.go
```diff
@@ -0,0 +1,111 @@
...
+		},
+		{
+			l: WrapperCL{QuadFunc{
+				a: mat.NewSymDense(4, []float64{
+					10, 2, 1, 9,
+					2, 5, -3, 4,
+					1, -3, 6, 2,
+					9, 4, 2, -14,
+				}),
+				b: mat.NewVector(4, []float64{3, -2, -1, 4}),
```
kortschak: This will need to change now.

diff/fd/hessian_test.go
```diff
@@ -16,60 +16,62 @@ type HessianTester interface {
...
+		tol: 1e-6,
+	},
+	{
+		h: QuadFunc{
+			a: mat.NewSymDense(3, []float64{
+				10, 2, 1,
+				2, 5, -3,
+				1, -3, 6,
+			}),
+			b: mat.NewVector(3, []float64{3, -2, -1}),
```
kortschak: And here.

diff/fd/crosslaplacian_test.go
btracey: Done.

diff/fd/hessian_test.go
btracey: Done.

diff/fd/crosslaplacian.go
```diff
@@ -0,0 +1,190 @@
...
+
+import (
+	"math"
+	"sync"
+)
+
+// CrossLaplacian computes a Laplacian-like quantity for a function of two vectors
+// at the locations x and y.
+// It computes
+//  ∇_y · ∇_x f(x,y) = \sum_i ∂^2 f(x,y)/∂x_i ∂y_i
```
vladimir-ch: Do you have a reference for this, where this comes up, what is it used for? For example, I think that the restriction len(x)=len(y) could be dropped if
`∇_y · ∇_x f(x,y) = \sum_i \sum_j ∂^2 f(x,y)/∂x_i ∂y_j`

diff/fd/crosslaplacian.go
```diff
@@ -0,0 +1,190 @@
...
+		}
+	}
+
+	nWorkers := computeWorkers(concurrent, evals)
+	if nWorkers == 1 {
+		return crossLaplacianSerial(f, x, y, formula.Stencil, step, originKnown, originValue)
+	} else {
+		return crossLaplacianConcurrent(nWorkers, evals, f, x, y, formula.Stencil, step, originKnown, originValue)
+	}
+	panic("fix")
```
vladimir-ch: What is this? A reminder to remove else and deindent?

diff/fd/crosslaplacian.go
btracey: I am using it to implement https://arxiv.org/pdf/1410.2392.pdf (Theorem 1). You're right, the length restriction can be dropped.

diff/fd/crosslaplacian.go
btracey: Yep, that's what it was. Gone.

diff/fd/crosslaplacian.go
btracey: I don't have a reference for this. I asked on math overflow, but got no reply. It doesn't feel like a totally crazy operator though.

btracey: Sorry, I take it back, the length restriction can't be dropped.

If we take the hessian as

```
[  ∂^2 f(x,y)/dx^2    ∂^2 f(x,y)/dxdy
   ∂^2 f(x,y)/dxdx    ∂^2 f(x,y)/dy^2 ]
```
which would be the blocks of the Hessian if we concatenate `x` and `y` together. This computes the diagonal of `∂^2 f(x,y) /dxdx`, while your formula computes the sum of all of the elements in that submatrix. I could imagine we would also want a `CrossHessian` which returns the full matrix instead of its Laplacian. For the Laplacian though, the vector lengths have to be equal.

btracey:
Do I need to do something with rebasing this on to master? The tests are failing because it doesn't know `NewVecDense`, but I thought the tests were always run by adding this to master.

diff/fd/crosslaplacian.go
vladimir-ch: Ok, you're right, the lengths must be equal here.

btracey: Ha ha! Right twice in one day!

vladimir-ch:
I think the tests are run with what is in the branch. If this uses NewVecDense, I would indeed rebase on to master:
```
git checkout master
git pull
git checkout difflaplace
git rebase master
git push -f
```

btracey:
Okay, done successfully I believe

btracey:
ping @vladimir-ch

diff/fd/laplacian.go
```diff
@@ -0,0 +1,160 @@
...
+			}
+			step = settings.Step
+		}
+		originKnown = settings.OriginKnown
+		originValue = settings.OriginValue
+		concurrent = settings.Concurrent
+	}
+
+	evals := n * len(formula.Stencil)
+	for _, pt := range formula.Stencil {
```
vladimir-ch: Use `usesOrigin` here.

diff/fd/crosslaplacian.go
```diff
@@ -0,0 +1,188 @@
...
+			}
+			step = settings.Step
+		}
+		originKnown = settings.OriginKnown
+		originValue = settings.OriginValue
+		concurrent = settings.Concurrent
+	}
+
+	evals := n * len(formula.Stencil) * len(formula.Stencil)
+	for _, pt := range formula.Stencil {
```
vladimir-ch: Use `usesOrigin` here.

btracey: Done. Good catch.

diff/fd/laplacian.go
btracey: Done.

btracey:
Thanks for the detailed review.

PR #156, opened 2017-07-31
mat: export Dense.solveTwoChol

kortschak:
@btracey Please take a look.

### DIFF
--- a/mat/cholesky.go
+++ b/mat/cholesky.go
@@ -182,10 +182,9 @@ func (c *Cholesky) Solve(m *Dense, b Matrix) error {
 	return nil
 }
 
-// TODO(kortschak): Export this as SolveTwoChol.
-// solveTwoChol finds the matrix m that solves A * m = B where A and B are represented
+// SolveChol finds the matrix m that solves A * m = B where A and B are represented
 // by their Cholesky decompositions a and b, placing the result in the receiver.
-func (m *Dense) solveTwoChol(a, b *Cholesky) error {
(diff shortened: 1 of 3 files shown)


btracey:
Shouldn't this be a method on Cholesky instead? (Probably the one that is `A`).

kortschak:
`func (a *Cholesky) SolveTwo(m *Dense, b *Cholesky) error`? I don't know. It's sort of troubling since the destination is nestled in the middle.

btracey:
Is it that different from `func (c *Cholesky) Solve(m *Dense, b Matrix) error`?

kortschak:
Not really. I'll fix it.

kortschak:
PTAL

mat/cholesky.go
```diff
@@ -182,10 +182,9 @@ func (c *Cholesky) Solve(m *Dense, b Matrix) error {
 	return nil
 }
 
-// TODO(kortschak): Export this as SolveTwoChol.
-// solveTwoChol finds the matrix m that solves A * m = B where A and B are represented
+// SolveTwo finds the matrix m that solves A * m = B where A and B are represented
 // by their Cholesky decompositions a and b, placing the result in the receiver.
-func (m *Dense) solveTwoChol(a, b *Cholesky) error {
+func (a *Cholesky) SolveTwo(m *Dense, b *Cholesky) error {
```
btracey: Should this be `SolveChol` instead? Seems more evocative than `SolveTwo`, which means less now.

kortschak: Done

PR #157, opened 2017-07-31
blas/gonum: add Zdotc

vladimir-ch:


blas/gonum/level1cmplx128.go
```diff
@@ -83,3 +83,44 @@ func (Implementation) Zcopy(n int, x []complex128, incX int, y []complex128, inc
 		iy += incY
 	}
 }
+
+// Zdotc computes the dot product
+//  x^H · y
```
vladimir-ch: I will change this to `x · y^H` since for us x,y are row vectors.

btracey: I don't think that's right. Our "vector" in `mat` is a column vector, and most of the inputs to lapack functions etc. are also column vectors. We're in row-major layout, but our interpretation of the data is the same beyond that.

vladimir-ch: Of course. Sometimes you have to disregard what I say.

PR #159, opened 2017-08-01
blas: add generated conversion code

kortschak:
Complex versions probably could be generated with reasonable ease.


blas/blas32/conv_test.go
```diff
@@ -0,0 +1,620 @@
...
+	diag() blas.Diag
+}
+
+func equalTriangular(a, b triangular) bool {
+	an := a.n()
+	bn := b.n()
+	if an != bn {
+		return false
+	}
+	for i := 0; i < an; i++ {
```
vladimir-ch: For other types you also compare uplo and diag. Not sure if it matters.

blas/blas32/conv.go
```diff
@@ -0,0 +1,279 @@
...
+				t.Data[i+j*t.Stride] = a.Data[i*a.Stride+j]
+			}
+		}
+	case blas.Lower:
+		for i := 0; i < a.N; i++ {
+			for j := 0; j <= i; j++ {
+				t.Data[i+j*t.Stride] = a.Data[i*a.Stride+j]
+			}
+		}
+	case blas.All:
```
kortschak: It exists, so I figured we should handle it.

blas/blas32/conv_symmetric_test.go
```diff
@@ -0,0 +1,307 @@
...
+		Data:   make([]float32, a.N*a.N),
+		Uplo:   a.Uplo,
+	}
+	t.From(a)
+	return t
+}
+
+func (m Symmetric) n() int { return m.N }
+func (m Symmetric) at(i, j int) float32 {
+	if m.Uplo == blas.Lower && i < j && j < m.N {
```
kortschak: Symmetric upper truncate the leading column for col major and the trailing row for row major, and the opposite for symmetric lower.

blas/blas32/conv_symmetric_test.go
```diff
@@ -0,0 +1,307 @@
...
+}
+func (m SymmetricCols) uplo() blas.Uplo { return m.Uplo }
+
+type symmetric interface {
+	n() int
+	at(i, j int) float32
+	uplo() blas.Uplo
+}
+
+func equalSymmetric(a, b symmetric) bool {
```
kortschak: Yes, it should be.

blas/blas32/conv_test.go
kortschak: It doesn't really matter since we are only dealing with matrix elements, not matrix metadata.

blas/blas32/conv_symmetric_test.go
kortschak: Issue #173.

PR #162, opened 2017-08-01
stat/sampleuv: set src appropriately in NewWeighted.

### LINKED ISSUE #161: src member of Weighted struct does not appear to be settable via the NewWeighted function call
I am probably missing something, but thought it is worth logging this, just in case. I don't see where the NewWeighted function in `https://github.com/gonum/gonum/blob/master/stat/sampleuv/weighted.go` uses the src parameter. It looks to me like src is not even mentioned in the function block. Is this intentional, or is this actually a bug?

```
func NewWeighted(w []float64, src *rand.Rand) Weighted {
	s := Weighted{
		weights: make([]float64, len(w)),
		heap:    make([]float64, len(w)),
	}
	s.ReweightAll(w)
	return s
}
```
Thanks!

btracey:
Fixes #161.

### DIFF
--- a/stat/sampleuv/weighted.go
+++ b/stat/sampleuv/weighted.go
@@ -42,6 +42,7 @@ func NewWeighted(w []float64, src *rand.Rand) Weighted {
 	s := Weighted{
 		weights: make([]float64, len(w)),
 		heap:    make([]float64, len(w)),
+		src:     src,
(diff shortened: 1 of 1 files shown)


btracey:
PTAL @szaydel @kortschak

PR #163, opened 2017-08-02
coveralls: exclude testlapack and testblas from coverage testing

vladimir-ch:


kortschak:
That's a very nice win!

Issue #61, opened 2017-06-08
graph/encoding/dot: harmonise encoding/decoding API


kortschak:
See comment at #58.

/cc @mewmew

mewmew:
Good.

Do you wish to propose how we may change the API, to get a starting point for discussion?

kortschak:
There are two ways to go. Either towards the `UnmarshalX`/`MarshalX` approach that you have, or to the `X() X`/`SetX(X)` that I have. I don't know which is better at this stage. get/set is easier to deal with, but is less flexible.

kortschak:
Having spent some more time on this (while writing some teaching code that depends on encoding/dot), I think the `X() X`/`SetX(X)` approach is going to be better from an API usability perspective. The main issue is that attributes are a list of arbitrary key/value pairs, so it is not possible for the marshaler to call a `n.MarshalAttr(key)` method without knowing the value of `key` ahead of time - by which time, it may as well have just got a `[]dot.Attribute`.

mewmew:
> Having spent some more time on this (while writing some teaching code that depends on encoding/dot),

If you feel like sharing, I'd be curious to take a look at the code used for teaching. It would be a great basis for discussion, if we could have that code implemented using both the current (marshal/unmarshal) API and the proposed (get/set) API, so we may use it for comparison on readability, extensibility and other aspects of the API.

kortschak:
https://github.com/kortschak/graphprac

mewmew:
> https://github.com/kortschak/graphprac

Thanks for providing a use case that we can use for comparison.

How would the code of `graphprac` differ, if `encoding/dot` used the X/SetX approach instead?

With this diff, we could discuss the benefits and drawbacks for a real use case, and could also see if there are specific limitations to either one of the approaches.

I've always been taught that you should not define an API without a clear use case in mind, so lets use this for discussion.

kortschak:
[...]

PR #164, opened 2017-08-03
graph/encoding: move basic encoding API into new encoding package

kortschak:
This is the first of two PRs to harmonise the graph encoding API. The next will change the `Unmarshal` methods to be `SetAttribute`. The API is still in flux.


kortschak:
Updates #61.

mewmew:
LGTM. Great that you are putting the effort into harmonising the API.

PR #169, opened 2017-08-03
graph/encoding/graphql: add new package for decoding GraphQL output

### LINKED ISSUE #72: graph/encoding: add support for parsing JSON-represented graphs
GraphQL outputs JSON in response to queries. We should be able to read these results in to allow operation on those graphs.

This is one of the use cases prompting the change from `int` to `int64` IDs.

There are two cases for the GraphQL output as far as we are concerned, with uid in which case we respect the provide value, and without when we need to make it up.

kortschak:
Please take a look.

Depends on #168.
Closes #72.

### DIFF
--- a/graph/encoding/graphql/decode.go
+++ b/graph/encoding/graphql/decode.go
@@ -1,0 +1,157 @@
+// Copyright ©2017 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package graphql
+
+import (
+	"bytes"
+	"encoding/json"
+	"errors"
+	"fmt"
+
+	"gonum.org/v1/gonum/graph"
+	"gonum.org/v1/gonum/graph/encoding"
+)
+
+// Unmarshal parses the the JSON-encoded data and stores the result in dst.
+// Node IDs are obtained from the JSON fields identified by the uid parameter.
+// UIDs obtained from the JSON encoding must map to unique node ID values
+// consistently across the JSON-encoded spanning tree.
+func Unmarshal(data []byte, uid string, dst encoding.Builder) error {
+	if uid == "" {
+		return errors.New("graphql: invalid UID field name")
+	}
+	var src json.RawMessage
+	err := json.Unmarshal(data, &src)
+	if err != nil {
+		return err
+	}
+	gen := generator{dst: dst, uidName: uid, nodes: make(map[string]graph.Node)}
+	return gen.walk(src, nil, "")
+}
+
+// StringIDSetter is a graph node that can set its ID based on the given uid string.
+type StringIDSetter interface {
+	SetIDFromString(uid string) error
+}
+
+// LabelSetter is a graph edge that can set its label.
+type LabelSetter interface {
+	SetLabel(string)
+}
+
+type generator struct {
+	dst encoding.Builder
+
+	// uidName is the name of the UID field in the source JSON.
+	uidName string
+	// nodes maps from GraphQL UID string to graph.Node.
+	nodes map[string]graph.Node
+}
+
+func (g *generator) walk(src json.RawMessage, node graph.Node, attr string) error {
+	switch src[0] {
+	case '{':
+		var val map[string]json.RawMessage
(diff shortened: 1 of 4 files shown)


graph/encoding/graphql/decode.go
```diff
@@ -0,0 +1,157 @@
...
+	SetLabel(string)
+}
+
+type generator struct {
+	dst encoding.Builder
+
+	// uidName is the name of the UID field in the source JSON.
+	uidName string
+
+	nodes map[string]graph.Node
```
mewmew: Add a comment stating the key of the map; e.g.

```go
// nodes maps from node value to graph node.
```

Is this comment correct? Are node values unique?

graph/encoding/graphql/decode.go
```diff
@@ -0,0 +1,157 @@
...
+				}
+				return fmt.Errorf("graphql: no UID for node: `%s`", &buf)
+			}
+		} else {
+			var v interface{}
+			err = json.Unmarshal(next, &v)
+			if err != nil {
+				return err
+			}
+			value := fmt.Sprint(v)
```
mewmew: What does the call to `fmt.Sprint` do? Can it be replaced with a type assertion for `fmt.Stringer`?

graph/encoding/graphql/decode.go
```diff
@@ -0,0 +1,157 @@
...
+		}
+
+	default:
+		var v interface{}
+		err := json.Unmarshal(src, &v)
+		if err != nil {
+			return err
+		}
+		if attr == g.uidName {
+			value := fmt.Sprint(v)
```
mewmew: Same as comment above.

graph/encoding/graphql/decode.go
kortschak: Done.

UIDs are unique in the input text.

graph/encoding/graphql/decode.go
kortschak: Everything can be represented as a string, `fmt.Sprint(v)` does this conversion. The two things we are likely to get here are a `string` or an `int`. Neither of these implement `fmt.Stringer`.

graph/encoding/graphql/decode.go
kortschak: Same answer.

kortschak:
PTAL

graph/encoding/graphql/decode.go
mewmew: Ok.

mewmew:
LGTM

kortschak:
Can you re-approve please.

mewmew:
I tried to have a look around, but can't find the option to do so in the web UI.

Where to re-approve?

mewmew:
It says

>  kortschak dismissed mewmew’s stale review via 1b863d0 7 hours ago

and now, the review button is gone for what it looks like.

mewmew:
I think the reason may be the force push,

> We went looking everywhere, but couldn’t find those commits.

kortschak:
You need to be in the [files changed tab](https://github.com/gonum/gonum/pull/169/files), where you'll see green "Review changes" button/menu/dialogue. The review is dismissed by any additional push.

mewmew:
> You need to be in the files changed tab, where you'll see green "Review changes" button/menu/dialogue. The review is dismissed by any additional push.

Ok, I'll take a look.

mewmew:
For future notes,

@kortschak Did you do a force push, or how did you push the latest incremental commit?

The benefit of uploading incremental commits, is that it is easier to review the changes, then you see the exact lines which change. Also, when the PR is complete, you may hit [squash and merge](https://github.com/blog/2141-squash-your-commits) to turn those several commits into a single one.

kortschak:
It was forced. In this PR there was very minimal change (the changes are visible from the GUI - it's unfortunate githubs git GC is so aggressive in removing those commits - it did not use to be) in addressing the comments - and significantly, the PR would not pass unless rebased onto the changes in #168 merged into master. Normally I do do incremental and retain the history (at least since the addition of the squash option on the GUI).

mewmew:
I see. Thanks for the background.

PR #170, opened 2017-08-04
mat: add Cond method to factorising types

### LINKED ISSUE #165: mat: Export Cond() methods for factorized types, and possibly rectify Cond() function implementation
`SVD` has a Cond method for returning the condition number. `LU`, `LQ`, and `QR` should also have the same.

There is also a lot of repeated code between the `mat.Cond` function and the factorized types. The reason for this repeat is allowing the norm of the condition number to adjust. I think we can have an unexported `factorize` function that inputs both the matrix and the condition norm. Then we can delete most of the code in `Cond`, and instead have

```
if m == n {
    var lu LU
    lu.factorize(m, lnorm)
    return lu.Cond()
}
```

and similar substitutions for the other cases.

kortschak:
@btracey Please take a look.

Fixes #165.

### DIFF
--- a/mat/lq.go
+++ b/mat/lq.go
@@ -9,6 +9,7 @@ import (
 
 	"gonum.org/v1/gonum/blas"
 	"gonum.org/v1/gonum/blas/blas64"
+	"gonum.org/v1/gonum/lapack"
 	"gonum.org/v1/gonum/lapack/lapack64"
 )
 
@@ -19,14 +20,14 @@ type LQ struct {
 	cond float64
 }
 
-func (lq *LQ) updateCond() {
+func (lq *LQ) updateCond(norm lapack.MatrixNorm) {
 	// A = LQ, where Q is orthonormal. Orthonormal multiplications do not change
 	// the condition number. Thus, ||A|| = ||L|| ||Q|| = ||Q||.
 	m := lq.lq.mat.Rows
 	work := getFloats(3*m, false)
 	iwork := getInts(m, false)
 	l := lq.lq.asTriDense(m, blas.NonUnit, blas.Lower)
-	v := lapack64.Trcon(CondNorm, l.mat, work, iwork)
+	v := lapack64.Trcon(norm, l.mat, work, iwork)
 	lq.cond = 1 / v
 	putFloats(work)
 	putInts(iwork)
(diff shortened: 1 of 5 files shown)


mat/matrix.go
```diff
@@ -302,63 +302,26 @@ func Cond(a Matrix, norm float64) float64 {
 	case math.Inf(1):
 		lnorm = lapack.MaxRowSum
 	}
+
+	tmp := getWorkspace(m, n, false)
```
btracey: I don't think we need this now. The factorization types already guard against modifying the matrix.

kortschak:
PTAL

mat/lq.go
```diff
@@ -19,14 +20,14 @@ type LQ struct {
 	cond float64
 }
 
-func (lq *LQ) updateCond() {
+func (lq *LQ) updateCond(norm lapack.MatrixNorm) {
 	// A = LQ, where Q is orthonormal. Orthonormal multiplications do not change
 	// the condition number. Thus, ||A|| = ||L|| ||Q|| = ||Q||.
```
vladimir-ch: Unrelated to this PR but I think that the formula misuses the norm notation for the condition number. Denoting it for example by `κ`, it should probably read:
`Thus, κ(A) = κ(L*Q) = κ(L)`.
This would be also in line with what the code does. Also, I'm not sure if the statement "Orthogonal multiplications do not change the condition number" is true for the max row/col sum norms used here, but since the exact value of the condition number is not important, we can probably leave the comment.

mat/lu.go
```diff
@@ -22,27 +23,27 @@ type LU struct {
...
+func (lu *LU) updateCond(anorm float64, norm lapack.MatrixNorm) {
 	n := lu.lu.mat.Cols
 	work := getFloats(4*n, false)
 	defer putFloats(work)
 	iwork := getInts(n, false)
 	defer putInts(iwork)
-	if norm < 0 {
+	if anorm < 0 {
 		// This is an approximation. By the definition of a norm, ||AB|| <= ||A|| ||B||.
 		// The condition number is ||A|| || A^-1||, so this will underestimate
```
vladimir-ch: Again unrelated, but isn't it `overestimate`? `κ(A) := |A| |A^-1| = |LU| |A^-1| <= |L| |U| |A^-1|`

mat/qr.go
```diff
@@ -19,14 +20,14 @@ type QR struct {
 	cond float64
 }
 
-func (qr *QR) updateCond() {
+func (qr *QR) updateCond(norm lapack.MatrixNorm) {
 	// A = QR, where Q is orthonormal. Orthonormal multiplications do not change
 	// the condition number. Thus, ||A|| = ||Q|| ||R|| = ||R||.
```
vladimir-ch: Also here.

mat/lq.go
kortschak: It's worth filing these as issues. They'll get forgotten here.

PR #172, opened 2017-08-04
mat: add Fill function for populating a Dense matrix

btracey:


btracey:
This is definitely only a convenience function, but would address https://github.com/gonum/matrix/issues/450, allowing code like

```
a := mat.NewDense(m, n, nil)
a.Fill(rand.NormFloat64)
```

sbinet:
what's wrong with:

```go
a := mat.NewDense(m, n, nil)
a.Apply(func(i, j int, v float64) float64 { return rand.NormFloat64() }, a)
```

except, perhaps, performances ?

btracey:
That works, but it doesn't really match the semantics. `Apply` is about transforming the elements of one matrix into another. `Fill` is just about populating a matrix with values. Also, `Fill` is easier to use, and given it's a convenience method, that's worth consideration.

sbinet:
ok :)

kortschak:
I am reluctant to add this, but if it is added, it should be a package functions `Fill(m Mutatable)`.

Note that the name apply comes from R, where it merely means apply the function `fn` to all elements of a list, vector, etc.

mat/dense_arithmetic.go
```diff
@@ -650,6 +650,22 @@ func (m *Dense) Apply(fn func(i, j int, v float64) float64, a Matrix) {
 	}
 }
 
+// Fill calls the input function for every element in the receiver and stores
+// the result.
+//
+// Fill panics if the receiver is zero.
+func (m *Dense) Fill(f func() float64) {
```
kortschak: s/f/fn/g

I there a reason that we need to panic for a zero value?

btracey:
Is there a reason it should be a package level function? It effectively constructs a matrix (hence the receiver seems appropriate), and `Mutable` is the least useful interface we have (any matrix that can be set at all possible elements is effectively a Dense matrix)

kortschak:
There will be at least one more general matrix type provided by us (the opencl implementation, and possibly a sparse type which is also general), so a package level function makes sense (I'm actually thinking `Apply` should have the same argument applied to it).

If, however, it gets implemented (I'm resisting here), it should be implemented as a wrapper around `Apply`; if this is being used as a constructor, the cost of the unused code in `Apply` should not be burdensome.

```
func (m *Dense) Fill(fn func() float64) {
    m.Apply(func(_, _ int, v float64) float64 { return fn() }, m)
}
```

btracey:
It's fine to be a wrapper around `Apply`. The reason I mention adding it is there are the threads here, plus I got a couple of questions to that effect at GopherCon. 

I don't agree with your logic on a package level function for the time being. It's a terrible idea to call Fill on a sparse matrix. The matrix will no longer be sparse. The OpenCL-based implementation doesn't exist, and so it's not clear what the impacts of the interop will be. If this should be a function rather than a method for OpenCL, then it's likely the entire package will need to change around (so that, for example, Mul can have a receiver that is a `Mutable` and not a `*Dense`). If we decide that kind of large scale change is necessary, we can fix this then too.

kortschak:
>It's a terrible idea to call Fill on a sparse matrix. The matrix will no longer be sparse.

This is true under two conditions.

1. The user doesn't understand the use cases of a sparse matrix.

2. The implementation is not intelligent in assignment of zeros.

In the first case, there is nothing we can do about that (saying that we can argues that a sparse type should not be mutable), in the second case we should be embarrassed.

A use of a `Fill(Mutable, func() float64)` is entirely valid on a mutable sparse destination if the function only sparsely returns non-zero values and the sparse destination is intelligent enough that a `Set(i, j, 0)` results in either removal of a non-zero element or a no-op in the case that the element is already zero.

The difference between `Mul` and `Fill`/`Apply` is that there is a lot more logic going on in `Mul`, and placing the logic in the destination type reduces the nxm code complexity for any given function.

btracey:
I don't think this is a good idea. It's just saving loops.

Issue #176, opened 2017-08-05
Possible function name confusion in gonum/stat/roc_example_test.go


szaydel:
It looks to me like function `func ExampleROC_unweighted()` while called unweighted is using and passing weights to stat.ROC,
```
tpr, fpr := stat.ROC(0, y, classes, weights)
```
while the function `ExampleROC_weighted()` does not pass any weights.
```
tpr, fpr := stat.ROC(0, y, classes, nil)
```
This seems backwards to me. I am happy to correct and do a PR if confirmed that this is indeed a flaw.

sbinet:
yep. thanks for spotting it!

szaydel:
np. Let me if I should do a PR with change, or it is too small and easy enough for someone already looking at it. Cheers.

sbinet:
I am not in front of a real keyboard at the moment, so feel free to send a PR.

kortschak:
Closed by #179.

PR #179, opened 2017-08-06
stat: Correct names of two functions to reflect proper meaning, which is currently reversed

szaydel:
A hopefully cleaner version of fix for #176 with a single commit. Please, let me know if this is not exactly what you were hoping to see. If not, a tiny bit of guidance will be kindly appreciated.


szaydel:
I will be sure to read over and follow more closely next time I commit something. I appreciate letting this one slide. Thanks a lot.

PR #180, opened 2017-08-06
mat: clarify implementation comments about condition number calculation

vladimir-ch:
Are we ok with the utf8 symbol usage?

### DIFF
--- a/mat/cholesky.go
+++ b/mat/cholesky.go
@@ -40,12 +40,13 @@ func (c *Cholesky) updateCond(norm float64) {
 	work := getFloats(3*n, false)
 	defer putFloats(work)
 	if norm < 0 {
-		// This is an approximation. By the definition of a norm, ||AB|| <= ||A|| ||B||.
-		// Here, A = U^T * U.
-		// The condition number is ||A|| || A^-1||, so this will underestimate
-		// the condition number somewhat.
-		// The norm of the original factorized matrix cannot be stored because of
-		// update possibilities.
+		// This is an approximation. By the definition of a norm,
+		//  |AB| <= |A| |B|.
+		// Since A = U^T*U, we get for the condition number κ that
+		//  κ(A) := |A| |A^-1| = |U^T*U| |A^-1| <= |U^T| |U| |A^-1|,
+		// so this will overestimate the condition number somewhat.
+		// The norm of the original factorized matrix cannot be stored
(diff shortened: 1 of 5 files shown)


vladimir-ch:
Also, looking at this, do we really have to export CondNormTrans? It's seems not. It is not a very useful piece of information, and it is used internally only in one place.

mat/cholesky.go
```diff
@@ -40,12 +40,13 @@ func (c *Cholesky) updateCond(norm float64) {
...
 	defer putFloats(work)
 	if norm < 0 {
-		// This is an approximation. By the definition of a norm, ||AB|| <= ||A|| ||B||.
-		// Here, A = U^T * U.
-		// The condition number is ||A|| || A^-1||, so this will underestimate
-		// the condition number somewhat.
-		// The norm of the original factorized matrix cannot be stored because of
-		// update possibilities.
+		// This is an approximation. By the definition of a norm,
+		//  ‖AB‖ <= ‖A‖ ‖B‖.
```
kortschak: `‖` renders unfortunately poorly in the font chosen by github (though this is not surprising given their other choices in the default style - pastels and low contrast text).

vladimir-ch: Ok, I will change it to just `|`.

vladimir-ch: Done.

vladimir-ch:
PTAL

mat/lq.go
```diff
@@ -21,8 +21,13 @@ type LQ struct {
...
 
 func (lq *LQ) updateCond(norm lapack.MatrixNorm) {
-	// A = LQ, where Q is orthonormal. Orthonormal multiplications do not change
-	// the condition number. Thus, ||A|| = ||L|| ||Q|| = ||Q||.
+	// Since A = L*Q, and Q is orthogonal, we get for the condition number κ
+	//  κ(A) := |A| |A^-1| = |L*Q| |(L*Q)^-1| = |L| |Q^T * L^-1|
+	//        = |L| |L^-1| = κ(L),
+	// where we used that fact that Q^-1 = Q^T. However, this assumes that
+	// the matrix norm is invariant under orthogonal transformations which
+	// is not the case for CondNorm. Hopefully the error is negligible, and κ
```
vladimir-ch: Done. Could you please reapprove?

PR #181, opened 2017-08-06
stat/all: reduce random size and test tolerance to decrease testing t…

btracey:
…ime.

We were generating a lot of random numbers, which is slow. Decrease the size of those random numbers, and in some cases increase the tolerance to compensate. In a couple cases, pull out code from testFullDist to allow for more fine-grained testing. This decrases:
distmat from 4.5s to 0.5s
distmv from 24.8s to 9s
distuv from 65.2s to 13s
samplemv from 2.8s to 1.2s
sampleuv from 3.5s to 2.1s

### DIFF
--- a/stat/distuv/triangle.go
+++ b/stat/distuv/triangle.go
@@ -18,9 +18,9 @@ type Triangle struct {
 // NewTriangle constructs a new triangle distribution with lower limit a, upper limit b, and mode c.
 // Constraints are a < b and a ≤ c ≤ b.
 // This distribution is uncommon in nature, but may be useful for simulation.
-func NewTriangle(a, b, c float64) Triangle {
+func NewTriangle(a, b, c float64, src *rand.Rand) Triangle {
 	checkTriangleParameters(a, b, c)
-	return Triangle{a, b, c, nil}
+	return Triangle{a, b, c, src}
 }
 
 func checkTriangleParameters(a, b, c float64) {
--- a/stat/distuv/triangle.go
+++ b/stat/distuv/triangle.go
@@ -18,9 +18,9 @@ type Triangle struct {
 // NewTriangle constructs a new triangle distribution with lower limit a, upper limit b, and mode c.
 // Constraints are a < b and a ≤ c ≤ b.
 // This distribution is uncommon in nature, but may be useful for simulation.
-func NewTriangle(a, b, c float64, src *rand.Rand) Triangle {
+func NewTriangle(a, b, c float64) Triangle {
 	checkTriangleParameters(a, b, c)
-	return Triangle{a, b, c, src}
+	return Triangle{a, b, c, nil}
 }
 
 func checkTriangleParameters(a, b, c float64) {
--- a/stat/distmat/wishart_test.go
+++ b/stat/distmat/wishart_test.go
@@ -94,7 +94,7 @@ func TestWishartRand(t *testing.T) {
 		{
 			v:       mat.NewSymDense(3, []float64{0.8, 0.3, 0.1, 0.3, 0.7, -0.1, 0.1, -0.1, 7}),
 			nu:      5,
-			samples: 300000,
+			samples: 30000,
 			tol:     3e-2,
 		},
 		{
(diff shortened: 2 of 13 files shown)


stat/distuv/triangle.go
```diff
@@ -18,9 +18,9 @@ type Triangle struct {
 // NewTriangle constructs a new triangle distribution with lower limit a, upper limit b, and mode c.
 // Constraints are a < b and a ≤ c ≤ b.
 // This distribution is uncommon in nature, but may be useful for simulation.
-func NewTriangle(a, b, c float64) Triangle {
+func NewTriangle(a, b, c float64, src *rand.Rand) Triangle {
```
kortschak: This deserves its own PR.

btracey: I did it because I was having issues with these tests being flaky as I was working, but I'll move it out of this PR.

btracey:
PTAL @sbinet

stat/distuv/triangle.go
kortschak: The original API here was to do `t := NewTriangle(a, b, c); t.Source = src`. This was my fault; few of the types here have constructors and depend on direct setting of the source, but `Triangle` needs one because of the constraints between the parameters. I left the normal API approach of direct setting of the source. I'm not convinced that it needs to change, but I'm also not convinced that it shouldn't.

sbinet:
LGTM.

Just for reference, I thought we could keep the number of samples and the tolerances as is, but use a pre-generated stream of pseudo-random values (and share them among the tests).
The gains weren't impressive:

```go
package srcrnd

import (
	"bufio"
	"encoding/binary"
	"io"
	"math/rand"
	"os"
)

const (
	Max = 300000
)

type fileSource struct {
	data []int64
	cur  int
}

func (src *fileSource) Seed(seed int64) {
	panic("not implemented")
}

func (src *fileSource) Int63() int64 {
	if src.cur == len(src.data) {
		src.cur = 0
	}
	v := src.data[src.cur]
	src.cur++
	return v
}

func NewSource() rand.Source {
	var src fileSource
	f, err := os.Open("rnd.bin")
	if err != nil {
		panic(err)
	}
	defer f.Close()

	r := bufio.NewReader(f)
loop:
	for {

		var buf [8]byte
		_, err := io.ReadFull(r, buf[:])
		if err == io.EOF {
			break loop
		}
		if err != nil {
			panic(err)
		}
		v := binary.LittleEndian.Uint64(buf[:])
		src.data = append(src.data, int64(v))
	}
	return &src
}
```

the `"rnd.bin"` would have been generated with:
```go
// +build ignore

package main

import (
	"encoding/binary"
	"log"
	"math/rand"
	"os"

	"gonum.org/v1/gonum/srcrnd"
)

func main() {
	f, err := os.Create("rnd.bin")
	if err != nil {
		log.Fatal(err)
	}
	defer f.Close()

	buf := [8]byte{}
	for i := 0; i < srcrnd.Max*100; i++ {
		v := rnd.Int63()
		binary.LittleEndian.PutUint64(buf[:], uint64(v))
		_, err = f.Write(buf[:])
		if err != nil {
			log.Fatal(err)
		}
	}

	err = f.Close()
	if err != nil {
		log.Fatal(err)
	}
}
```

I benched it:
```go
package srcrnd

import (
	"math/rand"
	"testing"
)

func TestNewSource(t *testing.T) {
	rrnd := rand.New(rand.NewSource(1))
	fsrc := rand.New(NewSource())
	for i := 0; i < Max; i++ {
		want := rrnd.Int63()
		got := fsrc.Int63()
		if want != got {
			t.Fatalf("iteration #%d: got=%v, want=%v", i, got, want)
		}
	}
}

func BenchmarkRandSource(b *testing.B) {
	sum := int64(0)
	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		sum += rnd.Int63()
	}
}

func BenchmarkFileSource(b *testing.B) {
	rnd := rand.New(NewSource())
	sum := int64(0)
	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		sum += rnd.Int63()
	}
}
```

```
$> go test -bench=.Source -v .
=== RUN   TestNewSource
--- PASS: TestNewSource (3.87s)
BenchmarkRandSource-4   	100000000	        11.1 ns/op
BenchmarkFileSource-4   	200000000	         9.02 ns/op
PASS
ok  	gonum.org/v1/gonum/srcrnd	23.658s
```

btracey:
I undid the triangle refactor and opened https://github.com/gonum/gonum/issues/183

PTAL @sbinet @kortschak

PR #186, opened 2017-08-08
lapack/testlapack: decrease testing time for Dbdsqr

vladimir-ch:


vladimir-ch:
On my machine from 20s to 2s.

PR #192, opened 2017-08-14
blas/gonum: add Zher

vladimir-ch:


blas/testblas/zher.go
```diff
@@ -0,0 +1,152 @@
...
+					} else {
+						for i := 0; i < n; i++ {
+							for j := i + 1; j < n; j++ {
+								a[i*lda+j] = cmplx.NaN()
+								want[i*lda+j] = cmplx.NaN()
+							}
+						}
+					}
+
+					impl.Zher(uplo, n, test.alpha, x, incX, a, lda)
```
kortschak: Can you do this for a complex diagonal and the same diaginal where the imaginary part has been zeroed before the call to confirm the same result.

blas/gonum/level2cmplx128.go
```diff
@@ -78,3 +79,102 @@ func (Implementation) Zgeru(m, n int, alpha complex128, x []complex128, incX int
...
+
+	var kx int
+	if incX < 0 {
+		kx = (1 - n) * incX
+	}
+	if uplo == blas.Upper {
+		if incX == 1 {
+			for i := 0; i < n; i++ {
+				if x[i] != 0 {
+					tmp := complex(alpha*real(x[i]), alpha*imag(x[i]))
```
kortschak: The Fortran write this as `alpha*dconjg(x(j))`, which should be `alpha*cmplx.Conj(x[i])`, no? Doing the conjugate in the `tmp` line means the two calls to `cmplx.Conj`, below can go away. Or is it this way because you are operating on rows?

blas/gonum/level2cmplx128.go
```diff
@@ -78,3 +79,102 @@ func (Implementation) Zgeru(m, n int, alpha complex128, x []complex128, incX int
...
+					a[i*lda+i] = complex(aii, 0)
+				}
+			}
+			return
+		}
+
+		ix := kx
+		for i := 0; i < n; i++ {
+			if x[ix] != 0 {
+				tmp := complex(alpha*real(x[ix]), alpha*imag(x[ix]))
```
kortschak: Same here.

blas/gonum/level2cmplx128.go
```diff
@@ -78,3 +79,102 @@ func (Implementation) Zgeru(m, n int, alpha complex128, x []complex128, incX int
...
+			}
+			ix += incX
+		}
+		return
+	}
+
+	if incX == 1 {
+		for i := 0; i < n; i++ {
+			if x[i] != 0 {
+				tmp := complex(alpha*real(x[i]), alpha*imag(x[i]))
```
kortschak: Same here.

blas/gonum/level2cmplx128.go
```diff
@@ -78,3 +79,102 @@ func (Implementation) Zgeru(m, n int, alpha complex128, x []complex128, incX int
...
+				a[i*lda+i] = complex(aii, 0)
+			}
+		}
+		return
+	}
+
+	ix := kx
+	for i := 0; i < n; i++ {
+		if x[ix] != 0 {
+			tmp := complex(alpha*real(x[ix]), alpha*imag(x[ix]))
```
kortschak: Same here.

blas/testblas/zher.go
vladimir-ch: Will do.

blas/gonum/level2cmplx128.go
vladimir-ch: To do `alpha*cmplx.Conj(x[i])` I would have to convert `alpha` to `complex128` which will be slower. Correct, I cannot do the conjugation in `tmp` because I'm operating on rows. However, now I'm thinking that simply operating on columns like in `Zgerc` would be simpler and probably faster: conjugation is done in tmp, and then asm Zaxpy updates the column. Do you agree?

kortschak: That seems fair, but it deserves measurement.

blas/gonum/level2cmplx128_bench_test.go
```diff
@@ -0,0 +1,59 @@
...
+
+func BenchmarkZherLowerN10Inc10(b *testing.B)    { benchmarkZher(b, blas.Lower, 10, 10) }
+func BenchmarkZherLowerN100Inc10(b *testing.B)   { benchmarkZher(b, blas.Lower, 100, 10) }
+func BenchmarkZherLowerN1000Inc10(b *testing.B)  { benchmarkZher(b, blas.Lower, 1000, 10) }
+func BenchmarkZherLowerN10000Inc10(b *testing.B) { benchmarkZher(b, blas.Lower, 10000, 10) }
+
+func BenchmarkZherLowerN10Inc1000(b *testing.B)    { benchmarkZher(b, blas.Lower, 10, 1000) }
+func BenchmarkZherLowerN100Inc1000(b *testing.B)   { benchmarkZher(b, blas.Lower, 100, 1000) }
+func BenchmarkZherLowerN1000Inc1000(b *testing.B)  { benchmarkZher(b, blas.Lower, 1000, 1000) }
+func BenchmarkZherLowerN10000Inc1000(b *testing.B) { benchmarkZher(b, blas.Lower, 10000, 1000) }
```
Kunde21: With go1.7, we can collapse these into a single Benchmark with sub-benchmark calls.  Take a look at [f64/axpy](https://github.com/gonum/gonum/blob/master/internal/asm/f64/benchAxpy_test.go) for an example.  When go1.9 releases, we can migrate everything to the sub-benchmark style while maintaining the 2-version promise.

vladimir-ch: Ok, I will do it. I'm not keeping mental track of when the sub-benchmarks were introduced, so I keep using the old style.

blas/gonum/level2cmplx128.go
vladimir-ch: Unless the benchmark is not correct, the numbers suggest to keep the row-wise operation:
```
name                      old time/op  new time/op  delta
ZherUpperN10Inc1-4         153ns ± 6%   175ns ± 5%   +14.54%  (p=0.029 n=4+4)
ZherUpperN100Inc1-4       12.8µs ± 7%   7.2µs ± 4%   -43.37%  (p=0.029 n=4+4)
ZherUpperN1000Inc1-4      1.19ms ± 9%  1.28ms ± 7%      ~     (p=0.114 n=4+4)
ZherUpperN10000Inc1-4      118ms ± 0%  1139ms ± 5%  +863.02%  (p=0.029 n=4+4)
ZherUpperN10Inc10-4        152ns ± 3%   176ns ± 3%   +15.46%  (p=0.029 n=4+4)
ZherUpperN100Inc10-4      12.4µs ± 5%   6.6µs ± 9%   -46.41%  (p=0.029 n=4+4)
ZherUpperN1000Inc10-4     1.47ms ± 6%  1.55ms ± 7%      ~     (p=0.343 n=4+4)
ZherUpperN10000Inc10-4     236ms ± 3%  1204ms ± 4%  +409.24%  (p=0.029 n=4+4)
ZherUpperN10Inc1000-4      149ns ± 7%   180ns ± 4%   +20.77%  (p=0.029 n=4+4)
ZherUpperN100Inc1000-4    12.8µs ± 4%   8.0µs ± 2%   -37.46%  (p=0.029 n=4+4)
ZherUpperN1000Inc1000-4   1.46ms ± 1%  2.32ms ± 6%   +59.43%  (p=0.029 n=4+4)
ZherUpperN10000Inc1000-4   747ms ± 6%  1741ms ± 1%  +133.18%  (p=0.029 n=4+4)
ZherLowerN10Inc1-4         147ns ± 6%   169ns ± 8%   +15.33%  (p=0.029 n=4+4)
ZherLowerN100Inc1-4       12.3µs ± 3%   7.1µs ± 8%   -42.15%  (p=0.029 n=4+4)
ZherLowerN1000Inc1-4      1.25ms ± 7%  1.30ms ± 9%      ~     (p=0.686 n=4+4)
ZherLowerN10000Inc1-4      121ms ± 4%  1101ms ± 6%  +810.65%  (p=0.029 n=4+4)
ZherLowerN10Inc10-4        143ns ± 8%   170ns ± 4%   +18.50%  (p=0.029 n=4+4)
ZherLowerN100Inc10-4      12.5µs ± 7%   6.9µs ±10%   -44.83%  (p=0.029 n=4+4)
ZherLowerN1000Inc10-4     1.45ms ± 4%  1.53ms ± 3%      ~     (p=0.057 n=4+4)
ZherLowerN10000Inc10-4     234ms ± 4%  1129ms ± 1%  +381.79%  (p=0.029 n=4+4)
ZherLowerN10Inc1000-4      140ns ±11%   162ns ±10%      ~     (p=0.057 n=4+4)
ZherLowerN100Inc1000-4    12.7µs ± 5%   8.0µs ± 4%   -37.02%  (p=0.029 n=4+4)
ZherLowerN1000Inc1000-4   1.65ms ± 7%  2.43ms ± 5%   +47.40%  (p=0.029 n=4+4)
ZherLowerN10000Inc1000-4   813ms ± 2%  1650ms ± 1%  +103.03%  (p=0.029 n=4+4)
```
old is row-wise with for loops, new is column-wise with `c128.AxpyInc`.

blas/gonum/level2cmplx128_bench_test.go
Kunde21: That's okay, they'll still work.  I suggest because it's easier to write a loop to generate sub-benchmarks, but it's not difficult to collapse them.  

Please don't let me get in the way of the work you're doing in the complex Blas realm.

vladimir-ch:
* Removed the commit with column-wise operation and `c128.AxpyInc`
* In the test, added setting the imaginary parts of the diagonal to NaN to test the documented behavior of Zher
* Switched to sub-benchmarks.

PTAL, @kortschak @Kunde21

blas/gonum/level2cmplx128_bench_test.go
kortschak: @Kunde21 unfortunately we can't migrate to 1.7-style subbenchmarks until 1.10 since we have a 3-version support window. We can however only do benchmarks for `// +build go1.7`. This is what we have done elsewhere.

blas/gonum/level2cmplx128_bench_test.go
```diff
@@ -0,0 +1,46 @@
...
+
+import (
+	"fmt"
+	"math/rand"
+	"testing"
+
+	"gonum.org/v1/gonum/blas"
+)
+
+var za []complex128
```
kortschak: I think in other places we have called it `benchSink`.

vladimir-ch: Done.

PR #193, opened 2017-08-14
fix package name in Package doc

nightlyone:
so the package being described in prose matches the package where this description lives in.

### DIFF
--- a/blas/gonum/doc.go
+++ b/blas/gonum/doc.go
@@ -5,7 +5,7 @@
 // Ensure changes made to blas/native are reflected in blas/cgo where relevant.
 
 /*
-Package native is a Go implementation of the BLAS API. This implementation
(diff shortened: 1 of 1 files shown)


nightlyone:
This is just a drive-by PR after looking at the documentation.

nightlyone:
Will do it later tonight. Wasn't sure about the process here.

vladimir-ch:
If you grep for "native", there are a couple of other places where it also occurs. You could include them here or in a next PR.

btracey:
Ping?

btracey:
Ping to sign A+C?

nightlyone:
Please ignore or apply the change in a different way later. I saw wanted to have a quick satisfying drive-by commit. Too much process for such a small change.

PR #194, opened 2017-08-14
asm/c128: Added [d]scalinc and [d]scalunitary

Kunde21:
Combining due to code similarity.  I can add the `To` variants if they're needed.

### DIFF
--- a/internal/asm/c128/dscalinc_amd64.s
+++ b/internal/asm/c128/dscalinc_amd64.s
@@ -15,6 +15,8 @@
 #define ALPHA X0
 #define ALPHA_2 X1
 
+#define MOVDDUP_ALPHA    LONG $0x44120FF2; WORD $0x0824 // MOVDDUP 8(SP), X0
+
 // func DscalInc(alpha float64, x []complex128, n, inc int)
 TEXT ·DscalInc(SB), NOSPLIT, $0
 	MOVQ x_base+8(FP), SRC // SRC = &x
@@ -22,30 +24,30 @@ TEXT ·DscalInc(SB), NOSPLIT, $0
 	CMPQ LEN, $0           // if LEN == 0 { return }
 	JE   dscal_end
 
-	MOVSD alpha+0(FP), ALPHA // ALPHA = alpha
-	MOVQ  inc+40(FP), INC    // INC = inc
-	SHLQ  $4, INC            // INC = INC * sizeof(complex128)
-	LEAQ  (INC)(INC*2), INC3 // INC3 = 3 * INC
-	MOVSD ALPHA, ALPHA_2     // Copy ALPHA and ALPHA_2 for pipelining
-	MOVQ  LEN, TAIL          // TAIL = LEN
-	SHRQ  $2, LEN            // LEN = floor( n / 4 )
-	JZ    dscal_tail         // if LEN == 0 { goto dscal_tail }
+	MOVDDUP_ALPHA             // ALPHA = alpha
+	MOVQ   inc+40(FP), INC    // INC = inc
+	SHLQ   $4, INC            // INC = INC * sizeof(complex128)
+	LEAQ   (INC)(INC*2), INC3 // INC3 = 3 * INC
+	MOVUPS ALPHA, ALPHA_2     // Copy ALPHA and ALPHA_2 for pipelining
+	MOVQ   LEN, TAIL          // TAIL = LEN
+	SHRQ   $2, LEN            // LEN = floor( n / 4 )
+	JZ     dscal_tail         // if LEN == 0 { goto dscal_tail }
 
 dscal_loop: // do {
-	MOVSD (SRC), X2         // X_i = real(x[i])
-	MOVSD (SRC)(INC*1), X3
-	MOVSD (SRC)(INC*2), X4
-	MOVSD (SRC)(INC3*1), X5
+	MOVUPS (SRC), X2         // X_i = real(x[i])
+	MOVUPS (SRC)(INC*1), X3
+	MOVUPS (SRC)(INC*2), X4
+	MOVUPS (SRC)(INC3*1), X5
 
-	MULSD ALPHA, X2   // X_i *= alpha
-	MULSD ALPHA_2, X3
-	MULSD ALPHA, X4
-	MULSD ALPHA_2, X5
+	MULPD ALPHA, X2   // X_i *= alpha
+	MULPD ALPHA_2, X3
+	MULPD ALPHA, X4
+	MULPD ALPHA_2, X5
 
-	MOVSD X2, (DST)         // real(x[i]) = X_i
-	MOVSD X3, (DST)(INC*1)
-	MOVSD X4, (DST)(INC*2)
-	MOVSD X5, (DST)(INC3*1)
+	MOVUPS X2, (DST)         // real(x[i]) = X_i
+	MOVUPS X3, (DST)(INC*1)
+	MOVUPS X4, (DST)(INC*2)
+	MOVUPS X5, (DST)(INC3*1)
 
 	LEAQ (SRC)(INC*4), SRC // SRC += INC*4
 	DECQ LEN
@@ -56,12 +58,12 @@ dscal_tail:
 	JE   dscal_end // if TAIL == 0 { return }
 
 dscal_tail_loop: // do {
-	MOVSD (SRC), X2       // X_i = real(x[i])
-	MULSD ALPHA, X2       // X_i *= alpha
-	MOVSD X2, (DST)       // real(x[i]) = X_i
-	ADDQ  INC, SRC        // SRC += INC
-	DECQ  TAIL
-	JNZ   dscal_tail_loop // } while --TAIL > 0
+	MOVUPS (SRC), X2       // X_i = real(x[i])
+	MULPD  ALPHA, X2       // X_i *= alpha
+	MOVUPS X2, (DST)       // real(x[i]) = X_i
+	ADDQ   INC, SRC        // SRC += INC
+	DECQ   TAIL
+	JNZ    dscal_tail_loop // } while --TAIL > 0
 
 dscal_end:
 	RET
--- a/internal/asm/c128/dscalinc_amd64.s
+++ b/internal/asm/c128/dscalinc_amd64.s
@@ -1,0 +1,67 @@
+// Copyright ©2017 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+//+build !noasm,!appengine
+
+#include "textflag.h"
+
+#define SRC SI
+#define DST SI
+#define LEN CX
+#define TAIL BX
+#define INC R9
+#define INC3 R10
+#define ALPHA X0
+#define ALPHA_2 X1
+
+// func DscalInc(alpha float64, x []complex128, n, inc int)
+TEXT ·DscalInc(SB), NOSPLIT, $0
+	MOVQ x_base+8(FP), SRC // SRC = &x
+	MOVQ n+32(FP), LEN     // LEN = n
+	CMPQ LEN, $0           // if LEN == 0 { return }
+	JE   dscal_end
+
+	MOVSD alpha+0(FP), ALPHA // ALPHA = alpha
+	MOVQ  inc+40(FP), INC    // INC = inc
+	SHLQ  $4, INC            // INC = INC * sizeof(complex128)
+	LEAQ  (INC)(INC*2), INC3 // INC3 = 3 * INC
+	MOVSD ALPHA, ALPHA_2     // Copy ALPHA and ALPHA_2 for pipelining
+	MOVQ  LEN, TAIL          // TAIL = LEN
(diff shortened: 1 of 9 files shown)


internal/asm/c128/dscalinc_amd64.s
```diff
@@ -0,0 +1,67 @@
...
+#define SRC SI
+#define DST SI
+#define LEN CX
+#define TAIL BX
+#define INC R9
+#define INC3 R10
+#define ALPHA X0
+#define ALPHA_2 X1
+
+// func DscalInc(alpha float64, x []complex128, n, inc int)
```
kortschak: Integer parameters should all be `uintptr`.

internal/asm/c128/scalinc_amd64.s
```diff
@@ -0,0 +1,121 @@
...
+#define MOVDDUP_X4_X5    LONG $0xEC120FF2 // MOVDDUP X4, X5
+#define MOVDDUP_X6_X7    LONG $0xFE120FF2 // MOVDDUP X6, X7
+#define MOVDDUP_X8_X9    LONG $0x120F45F2; BYTE $0xC8 // MOVDDUP X8, X9
+
+#define ADDSUBPD_X2_X3    LONG $0xDAD00F66 // ADDSUBPD X2, X3
+#define ADDSUBPD_X4_X5    LONG $0xECD00F66 // ADDSUBPD X4, X5
+#define ADDSUBPD_X6_X7    LONG $0xFED00F66 // ADDSUBPD X6, X7
+#define ADDSUBPD_X8_X9    LONG $0xD00F4566; BYTE $0xC8 // ADDSUBPD X8, X9
+
+// func ScalInc(alpha complex128, x []complex128, n, inc int)
```
kortschak: Here too.

internal/asm/c128/stubs_amd64.go
```diff
@@ -34,3 +34,29 @@ func AxpyInc(alpha complex128, x, y []complex128, n, incX, incY, ix, iy uintptr)
...
+//  for i, v := range x {
+//  	x[i] = complex(real(v)*alpha, imag(v))
+//  }
+func DscalUnitary(alpha float64, x []complex128)
+
+// DscalInc is
+//  for i := 0; i < n; i++ {
+//  	x[i*inc] = complex(real(x[i*inc])*alpha, imag(x[i*inc]))
+//  }
+func DscalInc(alpha float64, x []complex128, n, inc int)
```
kortschak: And here.

internal/asm/c128/stubs_amd64.go
```diff
@@ -34,3 +34,29 @@ func AxpyInc(alpha complex128, x, y []complex128, n, incX, incY, ix, iy uintptr)
...
+
+// ScalInc is
+//  var ix uintptr
+//  for i := 0; i < int(n); i++ {
+//  	x[ix] *= alpha
+//  	ix += incX
+//  }
+func ScalInc(alpha complex128, x []complex128, n, inc int)
```
kortschak: And here.

internal/asm/c128/stubs_noasm.go
```diff
@@ -55,3 +55,47 @@ func AxpyIncTo(dst []complex128, incDst, idst uintptr, alpha complex128, x, y []
...
+	for i, v := range x {
+		x[i] = complex(real(v)*alpha, imag(v))
+	}
+}
+
+// DscalInc is
+//  for i := 0; i < n; i++ {
+//  	x[i*inc] = complex(real(x[i*inc])*alpha, imag(x[i*inc]))
+//  }
+func DscalInc(alpha float64, x []complex128, n, inc int) {
```
kortschak: And here.

internal/asm/c128/stubs_noasm.go
```diff
@@ -55,3 +55,47 @@ func AxpyIncTo(dst []complex128, incDst, idst uintptr, alpha complex128, x, y []
...
+	}
+}
+
+// ScalInc is
+//  var ix uintptr
+//  for i := 0; i < int(n); i++ {
+//  	x[ix] *= alpha
+//  	ix += incX
+//  }
+func ScalInc(alpha complex128, x []complex128, n, inc int) {
```
kortschak: And here.

internal/asm/c128/dscalinc_amd64.s
Kunde21: Is there a reason we're taking in `int` at the Implementation level and casting to `uintptr` for the asm?  It adds type-casting without benefiting us at all.

kortschak: The exact details will be in the initial commits to the asm package in gonum/internal, but the rough picture is that `uintptr` and `int` are not necessarily the same size (amd64p32) and it is easier to pick this failure in the Go code than in the asm. For the common case though, there is no casting since the size is the same, it's just a type annotation.

internal/asm/c128/stubs_noasm.go
```diff
@@ -55,3 +55,13 @@ func AxpyIncTo(dst []complex128, incDst, idst uintptr, alpha complex128, x, y []
...
 	}
 }
+
+// DscalUnitary is
+//  for i, v := range x {
+//  	x[i] = complex(real(v)*alpha, imag(v))
+//  }
+func DscalUnitary(alpha float64, x []complex128) {
+	for i, v := range x {
+		x[i] = complex(real(v)*alpha, imag(v))
```
vladimir-ch: This is not correct, it should be `complex(alpha*real(v), alpha*imag(v))`.

internal/asm/c128/stubs_amd64.go
```diff
@@ -34,3 +34,31 @@ func AxpyInc(alpha complex128, x, y []complex128, n, incX, incY, ix, iy uintptr)
 //  	idst += incDst
 //  }
 func AxpyIncTo(dst []complex128, incDst, idst uintptr, alpha complex128, x, y []complex128, n, incX, incY, ix, iy uintptr)
+
+// DscalUnitary is
+//  for i, v := range x {
+//  	x[i] = complex(real(v)*alpha, imag(v))
```
kortschak: `complex(real(v)*alpha, imag(v)*alpha)`

internal/asm/c128/stubs_noasm.go
```diff
@@ -55,3 +55,51 @@ func AxpyIncTo(dst []complex128, incDst, idst uintptr, alpha complex128, x, y []
 		idst += incDst
 	}
 }
+
+// DscalUnitary is
+//  for i, v := range x {
+//  	x[i] = complex(real(v)*alpha, imag(v))
```
kortschak: `complex(real(v)*alpha, imag(v)*alpha)`

internal/asm/c128/stubs_amd64.go
Kunde21: Fixed

internal/asm/c128/stubs_test.go
```diff
@@ -82,150 +85,251 @@ var tests = []struct {
...
+		x:     []complex128{1 + 2i},
+		want:  []complex128{1 + 2i},
+	},
+	{
+		alpha: 2,
+		x:     []complex128{1 + 2i},
+		want:  []complex128{2 + 4i},
+	},
+	{
+		alpha: 2,
```
vladimir-ch: This test case is identical to the previous one.

internal/asm/c128/dscalinc_amd64.s
```diff
@@ -0,0 +1,69 @@
...
+	MOVQ   inc+40(FP), INC    // INC = inc
+	SHLQ   $4, INC            // INC = INC * sizeof(complex128)
+	LEAQ   (INC)(INC*2), INC3 // INC3 = 3 * INC
+	MOVUPS ALPHA, ALPHA_2     // Copy ALPHA and ALPHA_2 for pipelining
+	MOVQ   LEN, TAIL          // TAIL = LEN
+	SHRQ   $2, LEN            // LEN = floor( n / 4 )
+	JZ     dscal_tail         // if LEN == 0 { goto dscal_tail }
+
+dscal_loop: // do {
+	MOVUPS (SRC), X2         // X_i = real(x[i])
```
vladimir-ch: The comment is not correct.

internal/asm/c128/dscalinc_amd64.s
```diff
@@ -0,0 +1,69 @@
...
+	MOVUPS (SRC)(INC*1), X3
+	MOVUPS (SRC)(INC*2), X4
+	MOVUPS (SRC)(INC3*1), X5
+
+	MULPD ALPHA, X2   // X_i *= alpha
+	MULPD ALPHA_2, X3
+	MULPD ALPHA, X4
+	MULPD ALPHA_2, X5
+
+	MOVUPS X2, (DST)         // real(x[i]) = X_i
```
vladimir-ch: Also here and throughout.

internal/asm/c128/dscalunitary_amd64.s
```diff
@@ -0,0 +1,66 @@
...
+	MOVQ x_len+16(FP), LEN // LEN = len(x)
+	CMPQ LEN, $0           // if LEN == 0 { return }
+	JE   dscal_end
+
+	MOVDDUP_ALPHA         // ALPHA = alpha
+	XORQ   IDX, IDX       // IDX = 0
+	MOVUPS ALPHA, ALPHA_2 // Copy ALPHA to ALPHA_2 for pipelining
+	MOVQ   LEN, TAIL      // TAIL = LEN
+	SHRQ   $2, LEN        // LEN = floor( n / 4 )
+	JZ     dscal_tail     // if LEN == 0 { goto caxy_tail }
```
vladimir-ch: `goto dscal_tail`

internal/asm/c128/dscalinc_amd64.s
```diff
@@ -0,0 +1,69 @@
...
+	MOVQ x_base+8(FP), SRC // SRC = &x
+	MOVQ n+32(FP), LEN     // LEN = n
+	CMPQ LEN, $0           // if LEN == 0 { return }
+	JE   dscal_end
+
+	MOVDDUP_ALPHA             // ALPHA = alpha
+	MOVQ   inc+40(FP), INC    // INC = inc
+	SHLQ   $4, INC            // INC = INC * sizeof(complex128)
+	LEAQ   (INC)(INC*2), INC3 // INC3 = 3 * INC
+	MOVUPS ALPHA, ALPHA_2     // Copy ALPHA and ALPHA_2 for pipelining
```
kortschak: I hate plan9 asm. Why do they have to overload names that have clear meaning elsewhere.

internal/asm/c128/scalUnitary_amd64.s
```diff
@@ -0,0 +1,116 @@
...
+#define ADDSUBPD_X8_X9    LONG $0xD00F4566; BYTE $0xC8 // ADDSUBPD X8, X9
+
+// func ScalUnitary(alpha complex128, x []complex128)
+TEXT ·ScalUnitary(SB), NOSPLIT, $0
+	MOVQ x_base+16(FP), SRC // SRC = &x
+	MOVQ x_len+24(FP), LEN  // LEN = len(x)
+	CMPQ LEN, $0            // if LEN == 0 { return }
+	JE   scal_end
+
+	MOVUPS alpha+0(FP), ALPHA     // ALPHA = { imag(a), real(a) }
```
kortschak: Comments here and below should use `alpha` instead of `a` (and again, much hate for plan9 asm - how does this specify both parts of the `complex128`?).

internal/asm/c128/dscalinc_amd64.s
vladimir-ch: Out of curiosity, what overload do you mean, @kortschak ?

Having my attention drawn to this, why `MOVUPS` and not `MOVUPD`, @Kunde21 ?

Kunde21: I'm not sure to what overload you're referring either, @kortschak.  

@vladimir-ch The `MOVUPS` and `MOVUPD` instructions are bitwise-equivalent operations, decoding into the same micro-ops on the chip for the past few chip architecture generations.  The difference is that `MOVUPS` instruction is one byte smaller than the `MOVUPD` instruction, giving us more room in the i-cache.  (Ref: Intel Optimization Manual, Section 3.5.1.11: Mixing SIMD Data Types)

internal/asm/c128/scalUnitary_amd64.s
Kunde21: "how does this specify both parts of the `complex128`?"

It doesn't, `MOVUPS` simply moves a 128-bit chunk of data.  The comment is there to show the conceptual view of those 128 bits, while showing how it differs from two lines below where the 2 component elements are swapped.  The hardware has no concept of built-in complex numbers.

internal/asm/c128/dscalinc_amd64.s
kortschak: Thank you for clarifying this. When I was looking through, I had the same query as @vladimir-ch and so wen through the Go source tree to see if there was a definition - `MOVUPS` is use extensively, but only in integer operations and seems to be defined in a comment as purely a bytes move - though I see now why.

Issue #195, opened 2017-08-15
meta: submit Gonum to JOSS


sbinet:
We should consider submitting a pair about Gonum to the journal of open source software:

http://joss.theoj.org/about

PR #196, opened 2017-08-15
joss-paper: first stab

sbinet:
Updates gonum/gonum#195.

### DIFF
--- a/joss-paper/paper.md
+++ b/joss-paper/paper.md
@@ -1,0 +1,34 @@
+---
+title: 'Gonum: consistent, composable and comprehensible scientifc code'
+tags:
+  - Go
+  - Linear Algebra
+  - Lapack
+  - BLAS
+  - Statistics
+authors:
+  - name: Sebastien Binet
+    orcid: 0000-0003-4913-6104
+    affiliation: 1
+affiliations:
+  - name: IN2P3
+    index: 1
+date: 15 August 2017
+bibliography: paper.bib
+---
+
+# Summary
+
+Gonum provides tools enabling scientific research through consistent,
+composable and comprehensible scientific code.
+Gonum aims to be the equivalent of NumPy/SciPy but for the Go programming
+language.
+
+Gonum strives to deliver code that run fast (as fast as C/C++ or Fortran)
+and code that is as easy to write than in Matlab, R or Python.
+Gonum is still being actively developed but already offers a solid set of
+packages, providing matrices and linear algebra; statistics functions,
+probability distributions and sampling routines; finding the extrema of functions; and tools to compute numerical derivatives.
+
+# References
+
--- a/joss-paper/paper.md
+++ b/joss-paper/paper.md
@@ -25,7 +25,7 @@ Gonum aims to be the equivalent of NumPy/SciPy but for the Go programming
 language.
 
 Gonum strives to deliver code that run fast (as fast as C/C++ or Fortran)
-and code that is as easy to write than in Matlab, R or Python.
+and code that is as easy to write as in Matlab, R or Python.
 Gonum is still being actively developed but already offers a solid set of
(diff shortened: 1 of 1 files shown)


sbinet:
it's just to get the ball rolling...

kortschak:
This is too early. Note that I've raised the prospect of submitting to JOSS a couple of times on gonum-dev, but always in th context of after achieving what we consider API stability. We are not there yet (particularly after the migration to the single repo and the opportunities that gave us to break things without causing too much distress to users).

Waiting for some semblance of API stability gives us two things, a better product to go in the paper and possibly more importanly, a driver to fix up some of the API design decisions that have been noted and not had work done on them (the graph API needs work and has done for a couple of years, but I only have so much time - with @mewmew now interested we may get more work done in this direction, but probably not before the end of this semester while I am teaching).

I think there should be more discussion on gonum-dev about process before we start putting down files.

mewmew:
> the graph API needs work and has done for a couple of years, but I only have so much time - with @mewmew now interested we may get more work done in this direction, but probably not before the end of this semester while I am teaching

I feel very enthusiastic to work on the graph API and feel we can arrive at a simple, clean, and powerful API by continued collaboration and discussions. Attending a course in graph theory at Uni this autumn, starting in a couple of weeks :)

I also feel that the merge to a single repo is a wonderful opportunity to take a careful look at the API and design of the gonum packages, before declaring them as stable.

At that stage it would be lovely to see gonum submitted to JOSS.

sbinet:
Well, it all depends on what we expect by submitting to JOSS.
I was under the assumption that it would be a great way to raise Gonum's awareness and/or increase the number of possible developers+users.

some packages are more stable than others: perhaps we should consider submitting these instead? one paper per stable package?

anyways, I won't press this issue further.
feel free to close it.

kortschak:
Sure, really all I would like is that there be some discussion first, preferably on gonum-dev rather than here.

kortschak:
Also, it may be worth reaching out to the editorial board of JOSS to see what kinds of models are appropriate. It is unclear to me whether it is possible to submit per package multiple times for a single repo, and gonom/gonum now is substantially bigger than the common size of works submitted from a recent survey I took.

joss-paper/paper.md
```diff
@@ -0,0 +1,34 @@
...
+
+# Summary
+
+Gonum provides tools enabling scientific research through consistent,
+composable and comprehensible scientific code.
+Gonum aims to be the equivalent of NumPy/SciPy but for the Go programming
+language.
+
+Gonum strives to deliver code that run fast (as fast as C/C++ or Fortran)
+and code that is as easy to write than in Matlab, R or Python.
```
mewmew:
```diff
-as easy to write than in Matlab, R or Python
+as easy to write as in Matlab, R or Python
```

sbinet: done

btracey:
I agree with Dan that jumping to PRs already is premature. There is a lot to be discussed (location, scope, etc.). My past experience with writing these kind of software papers is that it's not necessary to have perfectly stable APIs, but having a stable location (which we now have) is important. I'm unfortunately gone for a week starting now, but I'll be happy to continue the discussion then.

joss-paper/paper.md
```diff
@@ -0,0 +1,34 @@
+---
+title: 'Gonum: consistent, composable and comprehensible scientifc code'
```
Konstantin8105: misspell in word "scientifc"

kortschak:
I think we might soon be in a position to start work on this again. Where do we want discussion to happen?

soypat:
Was this ever done? I've found sbinet's Go-HEP paper but nothing else.

sbinet:
nope.
IIRC, the consensus was to wait for a `v1.0.0`.
(let me close this one.)

PR #198, opened 2017-08-16
all: use fixdocs tool to store package documentation in doc.go

### LINKED ISSUE #12: all: add doc.go file to all packages
We have a variety of approaches to providing package-level documentation. This has confounded automation of adding canonical import comments with more finesse than in #11.

1. `doc.go`
2. `pkgname.go`
3. `doc.go` and `pkgname.go`
4. `othername.go`

To address this we should have a single `doc.go` in each package to hold package-level documentations.

This is unfortunately manual.

mewmew:
Fixes #12.


mewmew:
For `fixdocs` tool see https://github.com/gonum/gonum/issues/12#issuecomment-322739733

kortschak:
This has the unfortunate attribute of stripping copyright headers. I'd prefer if we kept the C++ format comments too.

mewmew:
> This has the unfortunate attribute of stripping copyright headers. I'd prefer if we kept the C++ format comments too.

No copyright headers are stripped, however, none were added to `doc.go` and adding such should be trivial with a small update to `fixdocs`. With C++ format comments, what do you mean? Do you mean line comments rather than block comments (`// foo` rather than `/* foo */`)?

mewmew:
The latest version of `fixdocs` adds copyright headers to `doc.go`: https://play.golang.org/p/nxCJ441zpP

mewmew:
I was too quick at the buttons, forgot to include line breaks for doc comments containing multiple line comments; e.g. https://github.com/gonum/gonum/pull/198/files#diff-254092d987e3bc62d648e1d55bf3021fR5

mewmew:
With this change we could also strip vanity import paths from all files except `doc.go`; e.g. in  [floats/floats.go](https://github.com/gonum/gonum/pull/198/files#diff-254092d987e3bc62d648e1d55bf3021fR5)

```diff
-package floats // import "gonum.org/v1/gonum/floats"
+package floats
```

@kortschak  https://github.com/gonum/gonum/issues/12#issue-230920601
> This has confounded automation of adding canonical import comments with more finesse than in #11.

Should I update `fixdocs` to remove unnecessary vanity imports as well?

mewmew:
The latest version of fixdocs removes vanity imports from all source, except doc.go. If vanity imports should be kept in other files, it is easy to revert to an older version of fixdocs to update this PR.

One last consideration. Should vanity imports also be added for automatic code generation scripts? The current version of fixdocs does so, e.g.  blas/testblas/benchautogen now has a vanity import, but didn't before.

https://github.com/gonum/gonum/pull/198/files#diff-1f5e830a5be6aada9743463c1af7eb28R6

blas/testblas/benchautogen/doc.go
```diff
@@ -0,0 +1,6 @@
+// Copyright ©2017 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Script for automatic code generation of the benchmark routines
+package main // import "gonum.org/v1/gonum/blas/testblas/benchautogen"
```
kortschak: Yeah, I don't think we need import comments on main packages.

mewmew: Done. PTAL, this PR should be ready for merge unless some other issues are identified.

mewmew:
Thanks for the review @kortschak!

P.S. what timezone are you in? It's 4 am where I'm at :)

kortschak:
+9:30

Issue #167, opened 2017-08-03
all: generalise vector functions


kortschak:
Candidate functions and methods at this stage are (only exported functions are listed):

- [x] `stat/statmat.go:func Mahalanobis(x, y *mat.VecDense, chol *mat.Cholesky) float64`
- [x] `mat/cholesky.go:func (c *Cholesky) SolveVec(v, b *VecDense) error`
- [x] `mat/cholesky.go:func (c *Cholesky) SymRankOne(orig *Cholesky, alpha float64, x *VecDense)`ok bool) {
- [x] `mat/symmetric.go:func (s *SymDense) SymRankOne(a Symmetric, alpha float64, x *VecDense)`
- [x] `mat/symmetric.go:func (s *SymDense) RankTwo(a Symmetric, alpha float64, x, y *VecDense)`
- [x] `mat/inner.go:func Inner(x *VecDense, A Matrix, y *VecDense) float64`
- [x] `mat/solve.go:func (v *VecDense) SolveVec(a Matrix, b *VecDense) error`
- [x] `mat/matrix.go:func Dot(a, b *VecDense) float64`
- [x] `mat/vector.go:func (v *VecDense) SliceVec(i, k int) *VecDense`
- [x] `mat/vector.go:func (v *VecDense) CloneVec(a *VecDense)`
- [x] `mat/vector.go:func (v *VecDense) CopyVec(a *VecDense) int`
- [x] `mat/vector.go:func (v *VecDense) ScaleVec(alpha float64, a *VecDense)`
- [x] `mat/vector.go:func (v *VecDense) AddScaledVec(a *VecDense, alpha float64, b *VecDense)`
- [x] `mat/vector.go:func (v *VecDense) AddVec(a, b *VecDense)`
- [x] `mat/vector.go:func (v *VecDense) SubVec(a, b *VecDense)`
- [x] `mat/vector.go:func (v *VecDense) MulElemVec(a, b *VecDense)`
- [x] `mat/vector.go:func (v *VecDense) DivElemVec(a, b *VecDense)`
- [x] `mat/vector.go:func (v *VecDense) MulVec(a Matrix, b *VecDense)`
- [x] `mat/qr.go:func (qr *QR) SolveVec(v *VecDense, trans bool, b *VecDense) error`
- [x] `mat/dense_arithmetic.go:func (m *Dense) RankOne(a Matrix, alpha float64, x, y *VecDense)`
[...]

PR #199, opened 2017-08-17
Generalise vector methods

james-bowman:
Generalised ColView and RowView methods to return Vector rather than VecDense and changed Dot() function to use Vector.

### DIFF
--- a/mat/dense.go
+++ b/mat/dense.go
@@ -205,22 +205,9 @@ func (m *Dense) T() Matrix {
 //
 // See ColViewer for more information.
 func (m *Dense) ColView(j int) Vector {
-	return m.ColViewVecDense(j)
-}
-
-// ColViewVecDense returns a VecDense reflecting the column j, backed by
-// the matrix data.
-func (m *Dense) ColViewVecDense(j int) *VecDense {
-	if j >= m.mat.Cols || j < 0 {
-		panic(ErrColAccess)
-	}
-	return &VecDense{
-		mat: blas64.Vector{
-			Inc:  m.mat.Stride,
-			Data: m.mat.Data[j : (m.mat.Rows-1)*m.mat.Stride+j+1],
-		},
-		n: m.mat.Rows,
-	}
+	v := &VecDense{}
+	v.ColViewOf(m, j)
+	return v
 }
 
 // SetCol sets the values in the specified column of the matrix to the values
@@ -257,22 +244,9 @@ func (m *Dense) SetRow(i int, src []float64) {
 //
 // See RowViewer for more information.
 func (m *Dense) RowView(i int) Vector {
-	return m.RowViewVecDense(i)
-}
-
-// RowViewVecDense returns a VecDense reflecting the row i, backed by
-// the matrix data.
-func (m *Dense) RowViewVecDense(i int) *VecDense {
-	if i >= m.mat.Rows || i < 0 {
-		panic(ErrRowAccess)
-	}
-	return &VecDense{
-		mat: blas64.Vector{
-			Inc:  1,
-			Data: m.rawRowView(i),
-		},
-		n: m.mat.Cols,
-	}
+	v := &VecDense{}
+	v.RowViewOf(m, i)
+	return v
 }
 
 // RawRowView returns a slice backed by the same array as backing the
--- a/mat/dense.go
+++ b/mat/dense.go
@@ -204,7 +204,7 @@ func (m *Dense) T() Matrix {
 // ColView returns a VecDense reflecting the column j, backed by the matrix data.
 //
 // See ColViewer for more information.
-func (m *Dense) ColView(j int) *VecDense {
+func (m *Dense) ColView(j int) Vector {
 	if j >= m.mat.Cols || j < 0 {
 		panic(ErrColAccess)
 	}
@@ -250,7 +250,7 @@ func (m *Dense) SetRow(i int, src []float64) {
 // backed by the matrix data.
 //
 // See RowViewer for more information.
-func (m *Dense) RowView(i int) *VecDense {
+func (m *Dense) RowView(i int) Vector {
 	if i >= m.mat.Rows || i < 0 {
 		panic(ErrRowAccess)
 	}
--- a/mat/dense.go
+++ b/mat/dense.go
@@ -205,6 +205,12 @@ func (m *Dense) T() Matrix {
 //
 // See ColViewer for more information.
 func (m *Dense) ColView(j int) Vector {
+	return m.ColViewVecDense(j)
+}
+
+// ColViewVecDense returns a VecDense reflecting the column j, backed by
+// the matrix data.
+func (m *Dense) ColViewVecDense(j int) *VecDense {
 	if j >= m.mat.Cols || j < 0 {
 		panic(ErrColAccess)
 	}
@@ -251,6 +257,12 @@ func (m *Dense) SetRow(i int, src []float64) {
 //
 // See RowViewer for more information.
 func (m *Dense) RowView(i int) Vector {
+	return m.RowViewVecDense(i)
+}
+
+// RowViewVecDense returns a VecDense reflecting the row i, backed by
+// the matrix data.
+func (m *Dense) RowViewVecDense(i int) *VecDense {
 	if i >= m.mat.Rows || i < 0 {
 		panic(ErrRowAccess)
 	}
--- a/mat/dense.go
+++ b/mat/dense.go
@@ -201,7 +201,7 @@ func (m *Dense) T() Matrix {
 	return Transpose{m}
 }
 
-// ColView returns a VecDense reflecting the column j, backed by the matrix data.
+// ColView returns a Vector reflecting the column j, backed by the matrix data.
 //
 // See ColViewer for more information.
 func (m *Dense) ColView(j int) Vector {
--- a/mat/dense.go
+++ b/mat/dense.go
@@ -205,9 +205,9 @@ func (m *Dense) T() Matrix {
 //
 // See ColViewer for more information.
 func (m *Dense) ColView(j int) Vector {
-	v := &VecDense{}
+	var v VecDense
 	v.ColViewOf(m, j)
-	return v
+	return &v
 }
 
 // SetCol sets the values in the specified column of the matrix to the values
@@ -244,9 +244,9 @@ func (m *Dense) SetRow(i int, src []float64) {
 //
 // See RowViewer for more information.
 func (m *Dense) RowView(i int) Vector {
-	v := &VecDense{}
+	var v VecDense
 	v.RowViewOf(m, i)
-	return v
+	return &v
 }
 
 // RawRowView returns a slice backed by the same array as backing the
--- a/mat/hogsvd.go
+++ b/mat/hogsvd.go
@@ -102,11 +102,10 @@ func (gsvd *HOGSVD) Factorize(m ...Matrix) (ok bool) {
 		return false
 	}
 	v := eig.Vectors()
-	cv := &VecDense{}
+	var cv VecDense
 	for j := 0; j < c; j++ {
 		cv.ColViewOf(v, j)
-		//cv := v.ColViewVecDense(j)
-		cv.ScaleVec(1/blas64.Nrm2(c, cv.mat), cv)
+		cv.ScaleVec(1/blas64.Nrm2(c, cv.mat), &cv)
 	}
 
 	b := make([]Dense, len(m))
@@ -161,11 +160,10 @@ func (gsvd *HOGSVD) UTo(dst *Dense, n int) *Dense {
 		dst.reuseAs(gsvd.b[n].Dims())
 	}
 	dst.Copy(&gsvd.b[n])
-	v := &VecDense{}
+	var v VecDense
 	for j, f := range gsvd.Values(nil, n) {
 		v.ColViewOf(dst, j)
-		//v := dst.ColViewVecDense(j)
-		v.ScaleVec(1/f, v)
+		v.ScaleVec(1/f, &v)
 	}
 	return dst
 }
@@ -191,7 +189,7 @@ func (gsvd *HOGSVD) Values(s []float64, n int) []float64 {
 	} else if len(s) != c {
 		panic(ErrSliceLengthMismatch)
 	}
-	v := &VecDense{}
+	var v VecDense
 	for j := 0; j < c; j++ {
 		v.ColViewOf(&gsvd.b[n], j)
 		s[j] = blas64.Nrm2(r, v.mat)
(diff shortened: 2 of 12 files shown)


kortschak:
One thing that is worth considering here is that it might be better to approach the transition from the call graph leaves, rather than from internal nodes; it would be easier to first address methods and functions that do not currently return `*VecDense`.

james-bowman:
Thanks for your comments, I am not happy with the type assertions either.  Looking through the code, one thing that occurs to me, is that currently, the only functions or methods that return `Vector` or `VecDense` (beyond `NewVecDense`) are `ColView` and `RowView`.   Within the `mat` package, `Dense` is the only type that implements the `ColViewer` or `RowViewer` interfaces and everywhere in the `mat` package that a vector is being requested, it is explicitly from a concrete `Dense` receiver.  Do we think it reasonable to expect that when obtaining a `Vector` from a concrete `Dense` type that it is a `VecDense`?  To me, the purpose of the Vector interface is an abstraction over matrix specific vector implementations i.e. I would expect to have pairs of associated Matrix/Vector implementations (e.g. SparseMatrix/SparseVector).  Is this the direction you see it going?

The type assertions are definitely ugly.  Perhaps one way around this, assuming the above is indeed the intended direction, could be to create `Dense` specific `ColViewDense` and `RowViewDense` (or similar) methods on the concrete `Dense` type that return a `VecDense`.  This way other matrix implementations (Sparse, etc.) could implement `ColViewer` and `RowViewer` and return associated sparse vector implementations whilst `Dense` specific client code could continue to use Dense matrices and vectors without needing to type assert.  Let me know your thoughts?

kortschak:
How we do things like this when specialisation is necessary is to provide the generalised method and a specialised method with the non-interface type being returned. See `T` and `TTri` on `TriDense`. In the first instance the `ColView` and `RowView` should return `Vector`. @btracey may have things to say here as well, but he is AFK at the moment.

mat/matrix.go
```diff
@@ -330,13 +330,22 @@ func Det(a Matrix) float64 {
...
 // Dot panics if the matrix sizes are unequal.
-func Dot(a, b *VecDense) float64 {
+func Dot(a, b Vector) float64 {
 	la := a.Len()
 	lb := b.Len()
 	if la != lb {
 		panic(ErrShape)
 	}
-	return blas64.Dot(la, a.mat, b.mat)
+	if avd, ok := a.(*VecDense); ok {
```
james-bowman: Thanks - updated accordingly.

james-bowman:
Great, sounds like we are thinking along the same lines.  I have added `ColViewVecDense()` and `RowViewVecDense()` methods to `Dense`, both of which return `*VecDense`.  I have left the methods of the `ColViewer` and `RowViewer` interfaces returning `Vector`.  This has now removed the need for the majority of type assertions.

mat/dense.go
```diff
@@ -201,10 +201,16 @@ func (m *Dense) T() Matrix {
 	return Transpose{m}
 }
 
-// ColView returns a VecDense reflecting the column j, backed by the matrix data.
+// ColView returns a Vector reflecting the column j, backed by the matrix data.
```
btracey: I think the current documentation is okay, but want to check. With a `*VecDense`, really only Dense could implement the interface. With it being `Vector`, now hypothetically the other types (e.g. `SymDense`) could implement it, and just not have the returned type be mutable. I think "backed by the matrix data" is vague enough to allow this.

mat/list_test.go
```diff
@@ -148,8 +148,18 @@ func legalTypeVec(v Matrix) bool {
 	return ok
 }
 
-// legalTypesVecVec returns whether both inputs are *VecDense.
```
btracey: "... both inputs are Vector"

mat/vector_test.go
```diff
@@ -225,17 +225,17 @@ func TestVecDenseScale(t *testing.T) {
 			want:  NewVecDense(3, []float64{0, -2, -4}),
 		},
 		{
-			a:     NewDense(3, 1, []float64{0, 1, 2}).ColView(0),
+			a:     NewDense(3, 1, []float64{0, 1, 2}).ColView(0).(*VecDense),
```
btracey: ColViewVecDense (and below)

mat/vector_test.go
```diff
@@ -307,12 +307,12 @@ func TestVecDenseAdd(t *testing.T) {
 		},
 		{
 			a:    NewVecDense(3, []float64{0, 1, 2}),
-			b:    NewDense(3, 1, []float64{0, 2, 3}).ColView(0),
+			b:    NewDense(3, 1, []float64{0, 2, 3}).ColView(0).(*VecDense),
```
btracey: ColViewVecDense (and below)

btracey:
These changes seem in the right direction. I agree that things should look better when we've changed over the arithmetic methods as well.

mat/vector_test.go
kortschak: I think this depends on what is being tested here; I would prefer this to be assigned to a `Vector` field and have the assertion in one place at L320 so we remember to remove it when `AddVec` is changed to accept `Vector` arguments.

mat/dense.go
james-bowman: Agree, `*VecDense` is closely associated to `Dense` so only a `Dense` is likely to return it.  Other `Matrix` types implementing `ColViewer`/`RowViewer` may need to return matrix specific `Vector` implementations, e.g. `SymDense` and `SymDenseVector`, etc.  Let me know if you think the documentation could do with more detail and I can update.

mat/dense.go
```diff
@@ -201,10 +201,16 @@ func (m *Dense) T() Matrix {
...
 //
 // See ColViewer for more information.
-func (m *Dense) ColView(j int) *VecDense {
+func (m *Dense) ColView(j int) Vector {
+	return m.ColViewVecDense(j)
+}
+
+// ColViewVecDense returns a VecDense reflecting the column j, backed by
+// the matrix data.
+func (m *Dense) ColViewVecDense(j int) *VecDense {
```
kortschak: On reflection this signature should be intverted; `func (v *VecDense) ColViewOf(m *Dense, j int)`. The same is true for the row equivalent.

By doing this we gain many allocation reductions cases where vector views are taken by getting the destination vector prior to a loop (possible getting it from a `*VecDense` pool).

The signature that exists is a product of history and us not thinking about changes when they happened; the views originally were slices, which meant that they were not allocated on the heap when a view was taken.

mat/vector.go
```diff
@@ -482,3 +482,31 @@ func (v *VecDense) asGeneral() blas64.General {
 		Data:   v.mat.Data,
 	}
 }
+
+// ColViewOf reflects the column j of the RawMatrixer m, into the receiver
```
kortschak: This needs to have documentation explaining that the sizes of the values do not need to match. Or alliteratively make a test for `IsZero` and allow free assignment in that case and otherwise test for size.

mat/vector.go
```diff
@@ -482,3 +482,31 @@ func (v *VecDense) asGeneral() blas64.General {
...
+	if j >= rm.Cols || j < 0 {
+		panic(ErrColAccess)
+	}
+
+	v.mat.Inc = rm.Stride
+	v.mat.Data = rm.Data[j : (rm.Rows-1)*rm.Stride+j+1]
+	v.n = rm.Rows
+}
+
+// RowViewOf reflects the row i of the RawMatrixer m, into the receiver
```
kortschak: Same here.

mat/dense.go
```diff
@@ -257,22 +244,9 @@ func (m *Dense) SetRow(i int, src []float64) {
...
-		panic(ErrRowAccess)
-	}
-	return &VecDense{
-		mat: blas64.Vector{
-			Inc:  1,
-			Data: m.rawRowView(i),
-		},
-		n: m.mat.Cols,
-	}
+	v := &VecDense{}
```
kortschak: Conventionally we write this as

```
var v VecDense
v.RowViewOf(m, i)
return &v
```

Elsewhere too please.

mat/hogsvd.go
```diff
@@ -102,8 +102,10 @@ func (gsvd *HOGSVD) Factorize(m ...Matrix) (ok bool) {
 		return false
 	}
 	v := eig.Vectors()
+	cv := &VecDense{}
 	for j := 0; j < c; j++ {
-		cv := v.ColViewVecDense(j)
+		cv.ColViewOf(v, j)
+		//cv := v.ColViewVecDense(j)
```
kortschak: Please delete the commented out code.

mat/hogsvd.go
```diff
@@ -159,8 +161,10 @@ func (gsvd *HOGSVD) UTo(dst *Dense, n int) *Dense {
 		dst.reuseAs(gsvd.b[n].Dims())
 	}
 	dst.Copy(&gsvd.b[n])
+	v := &VecDense{}
 	for j, f := range gsvd.Values(nil, n) {
-		v := dst.ColViewVecDense(j)
+		v.ColViewOf(dst, j)
+		//v := dst.ColViewVecDense(j)
```
kortschak: And here.

mat/vector_test.go
kortschak: Please remove the type assertions and make the `a` field a `Vector`. Then do the assertion at the call site.

mat/vector.go
```diff
@@ -484,7 +484,8 @@ func (v *VecDense) asGeneral() blas64.General {
 }
 
 // ColViewOf reflects the column j of the RawMatrixer m, into the receiver
-// backed by the same underlying data.
+// backed by the same underlying data.  The length of the receiver does not need
```
kortschak: Single space after dot.

mat/vector.go
```diff
@@ -498,7 +499,8 @@ func (v *VecDense) ColViewOf(m RawMatrixer, j int) {
 }
 
 // RowViewOf reflects the row i of the RawMatrixer m, into the receiver
-// backed by the same underlying data.
+// backed by the same underlying data.  The length of the receiver does not need
```
kortschak: Same here.

kortschak:
I would still like to wait for @btracey since there is API here that has not been discussed. I like what is here though.

btracey:
Sorry, I don't understand / like the `RowViewOf` method. What's wrong with just calling `RowView`? For other receiver methods, the point is that the type owns the data, and the method manipulates and sets it. The `RowView` method produces a type because the generator type is where the data is from. 

There are many kinds of types that can return a `RowView`, but it's basically only `Dense` that can be fed as an input to `RowViewOf`. I think we should get rid of it.

kortschak:
You've missed what is happening. There are Vectoe interface-returning RowView and ColView that satisfy the RowViewer and ColViewee interfaces and the methods that previously returned the concrete types are now on vector. This allows a reduction in the number of Vector allocations since a new Vector is not being allocated each use in loops.

kortschak:
>There are many kinds of types that can return a `RowView`, but it's basically only `Dense` that can be fed as an input to `RowViewOf`. I think we should get rid of it.

This is an argument for replacing pretty much every `.(RawMatrixer)` assertion with a `.(*Dense)`.

btracey:
> This is an argument for replacing pretty much every `.(RawMatrixer)` assertion with a `.(*Dense)`

I think that may be a good idea.

> This allows a reduction in the number of Vector allocations since a new Vector is not being allocated each use in loops.

I see now that's the intention. I'm not sure how large the benefit is, since the lifetime of the object should be reasonably clear, and thus easy to put on the stack. I don't know if that benefit is worth the cognitive complexity of the method, especially because there aren't many types for which it'll work, while there are a lot of matrix types that can implement `RowView`. We should just be encouraging interfaces that use `RowViewer` instead, it seems to me, as its simpler without much loss.

james-bowman:
I wonder if what we are seeing could possibly be a function of using narrow interfaces, and that we are storing results of operations in the receiver.  Matrix and Vector only offer `At()`, `Dims()` and `T()`.  If one wishes to do anything more complex or an arithmetic operation, one requires a concrete type.  This means that in many examples where `ColView()` is called, the  receiver is a `*Dense` rather than a `Matrix` and then there is a `.(*VecDense)` type assertion on the returned `Vector` shortly afterwards.  Here is a (highly) contrived example:

```
var m Dense
m.Mul(a, b)
r, c := m.Dims()
for i := 0; i < c; i++ {
   v := m.ColView(i)
   v.(*VecDense).<SomeOperation>()
}
```
In this case it feels reasonable to expect the returned `Vector` is in fact a `*VecDense` because it clearly came from a `*Dense`.

We could broaden the interfaces to avoid the need for the type assertion (and raise the requirement for implementing those interfaces), accept that we will always have type assertions, or provide additional, companion methods to `ColView/RowView`specifically return the concrete Vector implementation on the associated concrete matrix type where appropriate (or invert this signature as suggested to avoid multiple allocation inside loops).

What are your thoughts?

kortschak:
@btracey 

>> This is an argument for replacing pretty much every `.(RawMatrixer)` assertion with a `.(*Dense)`

>I think that may be a good idea.

This would be IMO a massively significant functionality regression, and something I will oppose very strongly. Having a type assertion on `RawMatrixer` and the other raw matrix returning types, give clients a first class hook into signals for potentially significantly more efficient lazy evaluation of matrices. This is how I plan to use it in the opencl matrix implementation - I can stop work on that if we lose this approach.

@james-bowman 

The approach that we have conventioally taken is that if you want to know the type, you create it. This is not always what we have, but it is clearly what we have been working towards if you look at the changes that have happened since the migration to a single repo.

btracey:
Fine to keep `RawMatrixer`. I'm still not sure that `ColViewOf` provides enough savings to merit the extra complexity.

kortschak:
It's useful internally, previously a `ColViewVecDense`/`RowViewVecDense` were accepted. These have the same complexity addition, and I'd argue poorer clarity since they are on the `Dense`.

btracey:
Okay.

mat/list_test.go
```diff
@@ -148,8 +148,19 @@ func legalTypeVec(v Matrix) bool {
...
+// legalTypesVecVec returns whether both inputs are Vectot
```
btracey: s/Vectot/Vector

mat/vector.go
btracey: I think it should be that either the size matches or `IsZero`. This overwrites data and I'd like to do what we can to minimize that error. I'd lean toward forcing `IsZero`, but that's too burdensome with loops.

mat/vector.go
```diff
@@ -492,6 +492,9 @@ func (v *VecDense) ColViewOf(m RawMatrixer, j int) {
 	if j >= rm.Cols || j < 0 {
 		panic(ErrColAccess)
 	}
+	if !v.IsZero() && v.n != rm.Rows {
```
btracey: You need to change the documentation to match the behavior.

james-bowman: Ah, good catch!

btracey:
Letting @kortschak merge to make sure he agrees with the behavior change

kortschak:
Updates #167.

PR #200, opened 2017-08-18
graph: move NewEdge method from encoding.Builder to graph.EdgeAdder

### LINKED ISSUE #56: graph/encoding/dot: NewEdge method of Builder is troubling
@mewmew I'm troubled by the presence of the NewEdge method in the Builder interface. I can see why it's there, but it is almost the same as SetEdge, and that's a problem. Possibly it could go away in the changes associated with the fix to #31?

mewmew:
Also rename graph.EdgeSetter to graph.EdgeAdder.

Fixes #56.

### DIFF
--- a/graph/encoding/dot/decode.go
+++ b/graph/encoding/dot/decode.go
@@ -167,21 +167,23 @@ func (gen *generator) addStmt(dst encoding.Builder, stmt ast.Stmt) {
 }
 
 // addEdgeStmt adds the given edge statement to the graph.
-func (gen *generator) addEdgeStmt(dst encoding.Builder, e *ast.EdgeStmt) {
-	fs := gen.addVertex(dst, e.From)
-	ts := gen.addEdge(dst, e.To)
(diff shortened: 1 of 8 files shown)


kortschak:
Sorry. I am just looking at using the new API in gen, and it seems that `NewEdge` should take a weight parameter, since it is not possible to set weight otherwise (without excursion from the API).

kortschak:
What is your view on the addition of a weight parameter here? At the moment, edge weights are first class, so it should go in, but depending on the outcome of thinking in #201, we could conceivably remove it later.

mewmew:
> What is your view on the addition of a weight parameter here? At the moment, edge weights are first class, so it should go in, but depending on the outcome of thinking in #201, we could conceivably remove it later.

I think there is no rush to merge this PR. It is a disruptive change, so lets get it right from the start. Lets continue discussions in #203 until we reach consensus, and keep this PR and #201 on hold for now.

mewmew:
Ok, merging.

PR #201, opened 2017-08-18
graph/graphs/gen: use EdgeAdder for duplication graph

kortschak:
The makes use of the new `graph.EdgeAdder` interface (currently the functionality is degraded because we cannot set the weight here - though do we want to?).

In order to make similar changes to the other random graph generation functions, it looks like we need a way to obtain a node with a given ID.

@mewmew Please take a look.

Depends #200.

### DIFF
--- a/graph/graphs/gen/duplication.go
+++ b/graph/graphs/gen/duplication.go
@@ -12,7 +12,6 @@ import (
 
 	"gonum.org/v1/gonum/graph"
 	"gonum.org/v1/gonum/graph/internal/ordered"
-	"gonum.org/v1/gonum/graph/simple"
 )
 
 // UndirectedMutator is an undirected graph builder that can remove edges.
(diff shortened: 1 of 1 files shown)


mewmew:
> The makes use of the new `graph.EdgeAdder` interface (currently the functionality is degraded because we cannot set the weight here - though do we want to?).

This is something I've been meaning to bring up in the past, to discuss whether an edge should be required to have a weight, or if simple there should be edges which implement a `Weighter` interface or something along those lines.

The rationale for not including weights in the original edge representation is that there are many kinds of graphs which don't make use of weights, and thus are required to include dummy values or dummy methods for edges.

Should we open a specific issue to discuss this topic?

The PR LGTM and may be merged after #200.

kortschak:
The issue around weights is difficult. The discussion happened in PRs and issues in gonum/graph if you want to wade through that. Briefly though, in order to be able to distinguish between weighted and unweighted graphs we need to have method set differences between the graph types (not the edge types, since the caller doesn't necessarily see those).

I guess it would be possible to change this so that edges did not by default have weights, but weighted graphs return a `type WeighedEdge interface { Edge; Weight() float64 }` from a `WeightedEdgeBetween(x, y Node) WeightedEdge` method etc. This increases the burden on implementers to add those additional methods (though they are trivially just the base `EdgeBetween` method with a different return type (and the base method then just wraps them with the `Edge` return type.

I think I would be OK with this. Thoughts?

mewmew:
Created #203 as a dedicated issue for discussions related to edge representations with and without associated weights.

kortschak:
I have rebased on to master. Please re-approve.

PR #202, opened 2017-08-18
blas/gonum: add Zher2

vladimir-ch:


blas/gonum/level2cmplx128.go
```diff
@@ -178,3 +178,112 @@ func (Implementation) Zher(uplo blas.Uplo, n int, alpha float64, x []complex128,
...
+			ky = (1 - n) * incY
+		}
+		ix = kx
+		iy = ky
+	}
+	if uplo == blas.Upper {
+		if incX == 1 && incY == 1 {
+			for i := 0; i < n; i++ {
+				if x[i] != 0 || y[i] != 0 {
+					tmp1 := alpha * x[i]
```
kortschak: Any reason for flipping these (`x`/`tmp1`, `y`/`tmp2`) from the fortran?

vladimir-ch: It's the row/column-wise operation difference. `tmp1` is the constant part per row/column in the first term which is `alpha*conj(y[i])` for the reference and `alpha*x[i]` for us.

PR #206, opened 2017-08-21
graph: redesign weight handling

kortschak:
@mewmew Please take a look.

This is the first few steps of the redesign. The next is to delete the `Weight` method from `Edge`. This step requires that the simple graphs and the undirect graph need to be duplicated into weighted and unweighted type because we cannot mix edge types without doing sneaky underhanded and difficult to maintain things behind the scenes. I have had a stab at the next step, but I'll pick it up again later. I figured it would be good to get your views on the progress so far.

After that (or in conjunction) the weighted edge addition methods and interfaces will be added.

### DIFF
--- a/graph/community/bisect_test.go
+++ b/graph/community/bisect_test.go
@@ -10,8 +10,8 @@ import (
 	"sort"
 	"testing"
 
+	"gonum.org/v1/gonum/graph"
 	"gonum.org/v1/gonum/graph/internal/ordered"
-
 	"gonum.org/v1/gonum/graph/simple"
 )
 
@@ -24,10 +24,10 @@ func ExampleProfile_simple() {
 	//  |/     \|
 	//  1       5
 	//
-	g := simple.NewUndirectedGraph(0, 0)
+	g := simple.NewUndirectedGraph()
 	for u, e := range smallDumbell {
 		for v := range e {
-			g.SetEdge(simple.Edge{F: simple.Node(u), T: simple.Node(v), W: 1})
+			g.SetEdge(simple.Edge{F: simple.Node(u), T: simple.Node(v)})
 		}
 	}
 
@@ -56,27 +56,27 @@ func ExampleProfile_simple() {
 	// Low:3.5 High:10 Score:0 Communities:[[0] [1] [2] [3] [4] [5]] Q=-0.607
 }
 
-var friends, enemies *simple.UndirectedGraph
+var friends, enemies *simple.WeightedUndirectedGraph
 
 func init() {
-	friends = simple.NewUndirectedGraph(0, 0)
+	friends = simple.NewWeightedUndirectedGraph(0, 0)
 	for u, e := range middleEast.friends {
 		// Ensure unconnected nodes are included.
 		if !friends.Has(simple.Node(u)) {
 			friends.AddNode(simple.Node(u))
 		}
 		for v := range e {
-			friends.SetEdge(simple.Edge{F: simple.Node(u), T: simple.Node(v), W: 1})
+			friends.SetWeightedEdge(simple.Edge{F: simple.Node(u), T: simple.Node(v), W: 1})
 		}
 	}
-	enemies = simple.NewUndirectedGraph(0, 0)
+	enemies = simple.NewWeightedUndirectedGraph(0, 0)
 	for u, e := range middleEast.enemies {
 		// Ensure unconnected nodes are included.
 		if !enemies.Has(simple.Node(u)) {
 			enemies.AddNode(simple.Node(u))
 		}
 		for v := range e {
-			enemies.SetEdge(simple.Edge{F: simple.Node(u), T: simple.Node(v), W: -1})
+			enemies.SetWeightedEdge(simple.Edge{F: simple.Node(u), T: simple.Node(v), W: -1})
 		}
 	}
 }
@@ -112,8 +112,8 @@ func ExampleProfile_multiplex() {
 	// Output:
 	// Low:0.1 High:0.72 Score:26 Communities:[[0] [1 7 9 12] [2 8 11] [3 4 5 10] [6]] Q=[24.7 1.97]
 	// Low:0.72 High:1.1 Score:24 Communities:[[0 6] [1 7 9 12] [2 8 11] [3 4 5 10]] Q=[16.9 14.1]
-	// Low:1.1 High:1.2 Score:18 Communities:[[0 2 6 11] [1 7 9 12] [3 4 5 8 10]] Q=[9.16 25.1]
-	// Low:1.2 High:1.6 Score:10 Communities:[[0 3 4 5 6 10] [1 7 9 12] [2 8 11]] Q=[11.4 24.1]
+	// Low:1.1 High:1.1 Score:18 Communities:[[0 2 6 11] [1 7 9 12] [3 4 5 8 10]] Q=[9.16 25.1]
+	// Low:1.1 High:1.6 Score:10 Communities:[[0 3 4 5 6 10] [1 7 9 12] [2 8 11]] Q=[11.5 23.9]
 	// Low:1.6 High:1.6 Score:8 Communities:[[0 1 6 7 9 12] [2 8 11] [3 4 5 10]] Q=[5.56 39.8]
 	// Low:1.6 High:1.8 Score:2 Communities:[[0 2 3 4 5 6 10] [1 7 8 9 11 12]] Q=[-1.82 48.6]
 	// Low:1.8 High:2.3 Score:-6 Communities:[[0 2 3 4 5 6 8 10 11] [1 7 9 12]] Q=[-5 57.5]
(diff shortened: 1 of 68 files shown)


graph/graph.go
```diff
@@ -18,6 +18,11 @@ type Edge interface {
 	Weight() float64
 }
 
+// WeightedEdge is a graph edge. In directed graphs, the direction
```
mewmew:
```diff
-// WeightedEdge is a graph edge.
+// WeightedEdge is a graph edge with an associated weight.
```

graph/graph.go
```diff
@@ -18,6 +18,11 @@ type Edge interface {
...
+// of the edge is given from -> to, otherwise the edge is semantically
+// unordered.
+type WeightedEdge Edge
```
mewmew: As already mentioned, later the `Weight` method should be added here and removed from `Edge`.

graph/graph.go
```diff
@@ -40,6 +45,27 @@ type Graph interface {
...
+	// From method.
+	WeightedEdge(u, v Node) WeightedEdge
+
+	// Weight returns the weight for the edge between
+	// x and y if Edge(x, y) returns a non-nil Edge.
+	// If x and y are the same node or there is no
+	// joining edge between the two nodes the weight
+	// value returned is implementation dependent.
+	// Weight returns true if an edge exists between
+	// x and y or if x and y have the same ID, false
```
mewmew: > ... if an edge exists between x and y or if x and y have the same ID ...

This formulation exists to support self-edges? I know simple doesn't (and shouldn't) support self-edges. Do any of the other graph implementations currently support self-edges? I'd like to play around with them in such case.

graph/graph.go
```diff
@@ -61,17 +95,17 @@ type Directed interface {
...
-	// joining edge between the two nodes the weight
-	// value returned is implementation dependent.
-	// Weight returns true if an edge exists between
-	// x and y or if x and y have the same ID, false
-	// otherwise.
-	Weight(x, y Node) (w float64, ok bool)
+// WeightedDirected is a weighted directed graph.
+type WeightedDirected interface {
+	Weighted
+
```
mewmew: I wish we could write:

```go
type WeightedDirected interface {
    Weighted
    Directed
}
```

to mean the union of `Weighted` and `Directed`.

graph/community/louvain_directed.go
```diff
@@ -111,9 +111,10 @@ type ReducedDirected struct {
 }
 
 var (
-	_ graph.Directed = (*ReducedDirected)(nil)
-	_ graph.Weighter = (*ReducedDirected)(nil)
-	_ ReducedGraph   = (*ReducedUndirected)(nil)
+	reducedDirected = (*ReducedDirected)(nil)
+
```
kortschak: Value versus compile-time type check.

graph/graph.go
kortschak: I'll fix this when `WeightedEdge` is the only weighted edge.

graph/graph.go
kortschak: Yes. This is non-trivial.

graph/graph.go
kortschak: We need them for community detection.

graph/graph.go
kortschak: Unfortunately they have common methods.

kortschak:
I am going to continue on this PR to get the entire shift in one merge commit.

graph/community/louvain_directed.go
mewmew: ok, thanks for the rationale

mewmew:
> I am going to continue on this PR to get the entire shift in one merge commit.

Sure, feel free to.

kortschak:
PTAL

I'm not wildly happy with what has happened with the spanning tree edge handling because of change to edge types in `graph/...: remove Weight method from Edge`. Suggestions welcome. The naive approach of adding something like

```
switch dst := dst.(type) {
case WeightedBuilder:
    dst.SetWeightedEdge(e)
default:
    dst.SetEdge(e)
}
```
seems a little gross.

kortschak:
The solution I have come up with is to say that the spanning tree functions have access to weights, so they should write those weights. This means the destination should be able to `SetWeightedEdge`. Clients that don't want to do this can write a simple shim struct to convert a call to `SetWeightedEdge` to `SetEdge`. The converse is not possible in the general case. The change also now means that the spanning trees exactly represent the spanning tree of `g` since they will use the nodes and edges from `g`, rather than simple representations (this is something we should probably propagate).

This is ready for review now. PTAL

graph/simple/weighted_directed.go
```diff
@@ -0,0 +1,285 @@
...
+
+	for to := range g.to[n.ID()] {
+		delete(g.from[to], n.ID())
+	}
+	delete(g.to, n.ID())
+
+	g.nodeIDs.release(n.ID())
+}
+
+// NewWeightedEdge returns a new Edge from the source to the destination node.
```
mewmew:
```diff
-NewWeightedEdge returns a new Edge from the source to the destination node.
+NewWeightedEdge returns a new weighted edge from the source to the destination node.
```

graph/simple/weighted_directed.go
```diff
@@ -0,0 +1,285 @@
...
+func (g *WeightedDirectedGraph) NewWeightedEdge(from, to graph.Node, weight float64) graph.WeightedEdge {
+	return &WeightedEdge{F: from, T: to, W: weight}
+}
+
+// SetWeightedEdge adds e, an edge from one node to another. If the nodes do not exist, they are added.
```
mewmew:
```diff
-SetWeightedEdge adds e, an edge from one node to another. If the nodes do not exist, they are added.
+SetWeightedEdge adds e, a weighted edge from one node to another. If the nodes do not exist, they are added.
```

graph/simple/weighted_directed.go
```diff
@@ -0,0 +1,285 @@
...
+	if _, ok := g.nodes[to.ID()]; !ok {
+		return
+	}
+
+	delete(g.from[from.ID()], to.ID())
+	delete(g.to[to.ID()], from.ID())
+}
+
+// Node returns the node in the graph with the given ID.
+func (g *WeightedDirectedGraph) Node(id int64) graph.Node {
```
mewmew: This method would go a long way to relieve the issues described in gonum/graph#87. Perhaps we should consider adding a `Node(id int64) graph.Node` method to the other graph.Graph implementations?

Then edge creation would be `g.NewEdge(g.Node(xid), g.Node(yid))` rather than `g.NewEdge(simple.Node(xid), simple.Node(yid))`.

But these changes can and should be part of a dedicated PR, to keep this one focused on one thing.

graph/simple/weighted_undirected.go
```diff
@@ -0,0 +1,260 @@
...
+
+	for from := range g.edges[n.ID()] {
+		delete(g.edges[from], n.ID())
+	}
+	delete(g.edges, n.ID())
+
+	g.nodeIDs.release(n.ID())
+}
+
+// NewWeightedEdge returns a new Edge from the source to the destination node.
```
mewmew:
```diff
-a new Edge
+a new weighted edge
```

graph/simple/weighted_undirected.go
```diff
@@ -0,0 +1,260 @@
...
+func (g *WeightedUndirectedGraph) NewWeightedEdge(from, to graph.Node, weight float64) graph.WeightedEdge {
+	return &WeightedEdge{F: from, T: to, W: weight}
+}
+
+// SetWeightedEdge adds an edge from one node to another. If the nodes do not exist, they are added.
```
mewmew:
```diff
-SetWeightedEdge adds an edge from one node to another.
+SetWeightedEdge adds a weighted edge from one node to another.
```

graph/simple/weighted_undirected.go
```diff
@@ -0,0 +1,260 @@
...
+	delete(g.edges[to.ID()], from.ID())
+}
+
+// Node returns the node in the graph with the given ID.
+func (g *WeightedUndirectedGraph) Node(id int64) graph.Node {
+	return g.nodes[id]
+}
+
+// Has returns whether the node exists within the graph.
+func (g *WeightedUndirectedGraph) Has(n graph.Node) bool {
```
mewmew: Also an unrelated issue to this PR, but I do believe we should look into changing the signature of `Has` to `Has(id int64) bool`. And that too, would be a separate PR.

graph/community/bisect_test.go
```diff
@@ -24,10 +24,10 @@ func ExampleProfile_simple() {
 	//  |/     \|
 	//  1       5
 	//
-	g := simple.NewUndirectedGraph(0, 0)
+	g := simple.NewUndirectedGraph()
 	for u, e := range smallDumbell {
 		for v := range e {
-			g.SetEdge(simple.Edge{F: simple.Node(u), T: simple.Node(v), W: 1})
+			g.SetEdge(simple.Edge{F: simple.Node(u), T: simple.Node(v)})
```
kortschak: Ack

graph/community/bisect_test.go
```diff
@@ -56,27 +56,27 @@ func ExampleProfile_simple() {
...
-	friends = simple.NewUndirectedGraph(0, 0)
+	friends = simple.NewWeightedUndirectedGraph(0, 0)
 	for u, e := range middleEast.friends {
 		// Ensure unconnected nodes are included.
 		if !friends.Has(simple.Node(u)) {
 			friends.AddNode(simple.Node(u))
 		}
 		for v := range e {
-			friends.SetEdge(simple.Edge{F: simple.Node(u), T: simple.Node(v), W: 1})
+			friends.SetWeightedEdge(simple.WeightedEdge{F: simple.Node(u), T: simple.Node(v), W: 1})
```
kortschak: Ack

graph/graph.go
```diff
@@ -15,6 +15,13 @@ type Node interface {
...
 	From() Node
 	To() Node
+}
+
+// WeightedEdge is a weighted graph edge. In directed graphs, the direction
+// of the edge is given from -> to, otherwise the edge is semantically
+// unordered.
+type WeightedEdge interface {
+	Edge
 	Weight() float64
```
kortschak: I'm not convinced this is necessary.

graph/graph.go
```diff
@@ -48,6 +76,14 @@ type Undirected interface {
 	EdgeBetween(x, y Node) Edge
 }
 
+// WeightedUndirected is a weighted undirected graph.
+type WeightedUndirected interface {
+	Weighted
+
+	// EdgeBetween returns the edge between nodes x and y.
```
kortschak: Done

graph/simple/weighted_directed.go
kortschak: Done

graph/simple/weighted_directed.go
kortschak: Done

graph/simple/weighted_directed.go
kortschak: Ack

graph/simple/weighted_undirected.go
kortschak: Done (and the line above).

graph/simple/weighted_undirected.go
kortschak: Ack

kortschak:
PTAL

graph/graph.go
mewmew: That is fine, I agree it is rather redundant. The main reason I prefer to have methods of commonly used interfaces documented, is so that user-defined implementations of those method sets, may include the comment easily and then focus on documenting and extending on what differs from the basic form of the method, if anything.

PR #211, opened 2017-08-25
graph/topo: add function to find k-core of undirected graph

kortschak:
This could be done using `VertexOrdering`, but this is marginally more efficient, and probably clearer to read given the naming for its use.

@mewmew Please take a look.


kortschak:
Related to this, but a separate PR is that I think `VertexOrdering` should be renamed to `DegeneracyOrdering`. This is #212.

mewmew:
I'm still too novice in graph theory to validate the implementation of this PR. Hopefully, this autumn of studies will improve that situation :)

@vladimir-ch is a better candidate for reviewing this PR.

kortschak:
No, I think you'll be fine. The PR here is fairly straight forward change factoring out a component of an already reviewed function and adding a special case of that function. The link in the code to the wiki article on graph degeneracy should help you.

This is the best way to learn.

PR #216, opened 2017-08-29
graph/{topo,community}: add clique graph and k-clique community finding

kortschak:
@mewmew @vladimir-ch Please take a look.

There are likely optimisations in `community.KCliqueCommunities` since we are using `topo.CliqueGraph` which does more work than is absolutely required, but at least for the moment the clarity in the code is worth that additional cost, which is probably significantly less than the unavoidable costs of doing the maximal clique finding in the first place. If it becomes an issue for someone, we can re-address it then.

### DIFF
--- a/graph/community/k_communities.go
+++ b/graph/community/k_communities.go
@@ -1,0 +1,98 @@
+// Copyright ©2017 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package community
+
+import (
+	"gonum.org/v1/gonum/graph"
+	"gonum.org/v1/gonum/graph/internal/set"
+	"gonum.org/v1/gonum/graph/simple"
+	"gonum.org/v1/gonum/graph/topo"
+	"gonum.org/v1/gonum/graph/traverse"
+)
+
+// KCliqueCommunities returns the k-clique communties of the undirected graph g for
+// k greater than zero. The returned communities are identified by linkage via k-clique
+// adjacency, where adjacency is defined as having k-1 common nodes. KCliqueCommunities
+// returns a single component including the full set of nodes of g when k is 1,
+// and the classical connected components of g when k is 2. Note that k-clique
+// communities may contain common nodes from g.
+//
+// k-clique communities are described in Palla et al. doi:10.1038/nature03607.
+func KCliqueCommunities(k int, g graph.Undirected) [][]graph.Node {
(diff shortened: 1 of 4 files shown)


mewmew:
@kortschak Just to let you know, I'm in my first week at Uni now, and will have a bit less time on my hands. Cheers /u

kortschak:
No worries, they're always just invitations.

graph/community/k_communities.go
```diff
@@ -0,0 +1,98 @@
...
+	"gonum.org/v1/gonum/graph/internal/set"
+	"gonum.org/v1/gonum/graph/simple"
+	"gonum.org/v1/gonum/graph/topo"
+	"gonum.org/v1/gonum/graph/traverse"
+)
+
+// KCliqueCommunities returns the k-clique communties of the undirected graph g for
+// k greater than zero. The returned communities are identified by linkage via k-clique
+// adjacency, where adjacency is defined as having k-1 common nodes. KCliqueCommunities
+// returns a single component including/ the full set of nodes of g when k is 1,
```
vladimir-ch: Extra `/`.

graph/community/k_communities_test.go
```diff
@@ -0,0 +1,132 @@
+// Copyright ©2015 The gonum Authors. All rights reserved.
```
vladimir-ch: 2015?

graph/topo/clique_graph.go
```diff
@@ -0,0 +1,106 @@
...
+		}
+	}
+
+	for _, cliques := range cliqueNodes {
+		for i, uc := range cliques {
+			for _, vc := range cliques[i+1:] {
+				// Retain the nodes that contribute to the
+				// edge between the cliques.
+				var edgeNodes []graph.Node
+				switch 1 {
```
vladimir-ch: The first time I see this construct, wow.

graph/community/k_communities.go
kortschak: Done.

graph/community/k_communities_test.go
kortschak: Done.

kortschak:
Fixes pushed, please re-approve.

graph/topo/clique_graph.go
kortschak: Yes, `switch {...` is just sugar for `switch true {...`. There are interesting an clever ways that you can abuse this.

PR #218, opened 2017-08-30
graph/path: allow Shortest to traverse negative cycles

### LINKED ISSUE #217: Add Access to the Negative Cycle found in Bellman Ford Moore
Since the Bellman Ford Moore algorithm terminates on finding a negative cycle it is often used for applications that require the detection of a negative cycle (e.g. cycle-cancelling techniques or finding arbitrage opportunities).

Currently the implementation only allows to test for the cycle. It would be great to make it accessible as well.

kortschak:
@mewmew @vladimir-ch @Noki Please take a look.

I have chosen to use NaN as the weight for negative cycle paths since a linear path that has a -Inf weight would also be found to have a -Inf path weight using the approach in the example. Intuitively also, the weight of the path seems like NaN rather than -Inf.

Fixes #217.

### DIFF
--- a/graph/path/bellman_ford_moore.go
+++ b/graph/path/bellman_ford_moore.go
@@ -58,6 +58,7 @@ func BellmanFordFrom(u graph.Node, g graph.Graph) (path Shortest, ok bool) {
 				panic("bellman-ford: unexpected invalid weight")
 			}
 			if path.dist[j]+w < path.dist[k] {
+				path.hasNegativeCycle = true
 				return path, false
 			}
 		}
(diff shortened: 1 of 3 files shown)


kortschak:
I have had to rebase to remove conflicts. Please re-approve.

retain both:
```
	"gonum.org/v1/gonum/graph"
<<<<<<< negative-cycles
	"gonum.org/v1/gonum/graph/internal/set"
=======
	"gonum.org/v1/gonum/graph/internal/ordered"
>>>>>>> master
	"gonum.org/v1/gonum/mat"
)
```

second line of negative-cycles and first line of master.
```
	}
<<<<<<< negative-cycles
	reverse(path)
	return path, math.Min(weight, p.dist[p.indexOf[v.ID()]])
=======
	ordered.Reverse(path)
	return path, p.dist[p.indexOf[v.ID()]]
>>>>>>> master
}
```

PR #225, opened 2017-09-02
blas/gonum:  Fixing NaN popagation in Dger. 

### LINKED ISSUE #224: blas/gonum:  Incorrect Dger results when NaN is an element of y.
Found while fuzzing blas:  The `if tmp != 0` short-circuit in `Dger` calculation causes incorrect handling of NaN elements in `y` vector.

Kunde21:
Fixes #224 

Added regression test and changed calculation to standardize NaN propagation in Dger.

### DIFF
--- a/blas/gonum/level2double.go
+++ b/blas/gonum/level2double.go
@@ -167,17 +167,14 @@ func (Implementation) Dger(m, n int, alpha float64, x []float64, incX int, y []f
 		x = x[:m]
 		y = y[:n]
 		for i, xv := range x {
-			tmp := alpha * xv
-			atmp := a[i*lda : i*lda+n]
-			f64.AxpyUnitaryTo(atmp, tmp, y, atmp)
+			f64.AxpyUnitary(alpha*xv, y, a[i*lda:i*lda+n])
 		}
 		return
 	}
 
 	ix := kx
 	for i := 0; i < m; i++ {
-		tmp := alpha * x[ix]
-		f64.AxpyInc(tmp, y, a[i*lda:i*lda+n], uintptr(n), uintptr(incY), 1, uintptr(ky), 0)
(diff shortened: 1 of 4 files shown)


kortschak:
Have we run out of log lines?

kortschak:
You might want to rebase this onto master now and see if that fixes the failure.

blas/gonum/level2double.go
```diff
@@ -168,20 +168,16 @@ func (Implementation) Dger(m, n int, alpha float64, x []float64, incX int, y []f
 		y = y[:n]
 		for i, xv := range x {
 			tmp := alpha * xv
-			if tmp != 0 {
-				atmp := a[i*lda : i*lda+n]
-				f64.AxpyUnitaryTo(atmp, tmp, y, atmp)
-			}
+			atmp := a[i*lda : i*lda+n]
+			f64.AxpyUnitaryTo(atmp, tmp, y, atmp)
```
kortschak: It seems that me that `tmp` could be calculated in line in the call to `f64.AxpyUnitaryTo` since the only reason it existed before was to allow the conditional.

Also, it there a reason not to use `f64.AxpyUnitary` and do the slicing in line too?

blas/gonum/level2double.go
```diff
@@ -168,20 +168,16 @@ func (Implementation) Dger(m, n int, alpha float64, x []float64, incX int, y []f
...
 		return
 	}
 
 	ix := kx
 	for i := 0; i < m; i++ {
 		tmp := alpha * x[ix]
-		if tmp != 0 {
-			f64.AxpyInc(tmp, y, a[i*lda:i*lda+n], uintptr(n), uintptr(incY), 1, uintptr(ky), 0)
-		}
+		f64.AxpyInc(tmp, y, a[i*lda:i*lda+n], uintptr(n), uintptr(incY), 1, uintptr(ky), 0)
```
kortschak: Calculate `tmp` in line here too.

PR #233, opened 2017-09-14
blas/gonum: add Zgemv

vladimir-ch:


blas/gonum/level2cmplx128.go
```diff
@@ -11,6 +11,135 @@ import (
...
+		}
+		return
+
+	case blas.ConjTrans:
+		// Form y := alpha*A^H*x + y.
+		ix := kx
+		if incY == 1 {
+			for i := 0; i < m; i++ {
+				tmp := alpha * x[ix]
+				for j := 0; j < n; j++ {
```
kortschak: Not for here, but does this warrant a `c128.AxpycUnitary` and `c128.AxpycInc`?

vladimir-ch: If it was only here, then I think not, but let's see what the upcoming routines will show. It will very possibly appear again.

PR #235, opened 2017-09-16
Fixed documentation for blas/gonum where old documentation referred to package as `native`

chewxy:
Doc still referred to this package as `native`. It's now fixed


chewxy:
Already in #193

PR #237, opened 2017-09-17
graph/encoding/dot: store graph DOT ID if present

mewmew:


mewmew:
Thanks

kortschak:
You can do the merge, but it occurs to me that maybe there should be a graph label on the test cases to check this.

mewmew:
I'll add a test case for it and push sometime next week :)

mewmew:
PTAL

PR #244, opened 2017-09-28
stat: add BivariateMoment

btracey:


btracey:
If anyone is wondering why this is useful, this computes the mu terms in https://stats.stackexchange.com/questions/48366/standard-errors-for-covariance-estimate-in-r

stat/stat.go
```diff
@@ -876,13 +911,17 @@ func Moment(moment float64, x, weights []float64) float64 {
 		}
 		return m / float64(len(x))
 	}
+	if len(x) != len(weights) {
```
kortschak: This should really happen in another PR.

stat/stat.go
```diff
@@ -876,13 +911,17 @@ func Moment(moment float64, x, weights []float64) float64 {
...
+		panic("stat: slice length mismatch")
+	}
 	var (
 		m          float64
 		sumWeights float64
 	)
 	for i, v := range x {
-		m += weights[i] * math.Pow(v-mean, moment)
-		sumWeights += weights[i]
+		w := weights[i]
```
kortschak: Same here.

stat/stat.go
```diff
@@ -862,6 +862,41 @@ func Mode(x, weights []float64) (val float64, count float64) {
 	return max, maxCount
 }
 
+// MixedMoment computes the weighted mixed moment between the samples x and y.
```
kortschak: `BivariateMixedMoment` in case we want higher order mixed moments?

btracey:
I changed to `BivariateMoment` instead of `BivariateMixedMoment`

btracey:
I'm opening an issue for the example. The variance of the covariance as liked above is a good example, but I need to test that it's correct.

Issue #47, opened 2017-06-03
stat/distmv: Adjust NewNormalChol


btracey:
Part of the intended effect of NewNormalChol was to avoid allocation when the cholesky decomposition is already known. Now that we require both `Sigma` and `Chol`, this is not sufficient to avoid allocation. We should either change this to take both `Chol` and `Sigma`, or add a different function that does.

btracey:
Instead, we have exported some common functions to avoid these allocations.

PR #249, opened 2017-09-30
stat/distmv: rename NewNormalChol and add Sigma as an input argument

btracey:


btracey:
I'm not sure this is the best name, but I can't think of anything better.

btracey:
Fixes #47

stat/distmv/normal.go
```diff
@@ -64,19 +64,25 @@ func NewNormal(mu []float64, sigma mat.Symmetric, src *rand.Rand) (*Normal, bool
 	return n, true
 }
 
-// NewNormalChol creates a new Normal distribution with the given mean and
-// covariance matrix represented by its Cholesky decomposition. NewNormalChol
-// panics if len(mu) is not equal to chol.Size().
-func NewNormalChol(mu []float64, chol *mat.Cholesky, src *rand.Rand) *Normal {
+// NewNormalData creates a new Normal distribution with the given mean and
+// covariance matrix when the Cholesky decomposition of the covariance matrix
+// is already known. NewNormalChol panics if len(mu) is not equal to chol.Size()
```
kortschak: s/NewNormalChol/NewNormalData/

Also in the test name.

btracey:
Do you have any ideas for the name?

kortschak:
No unfortunately - I would have suggested one if I did.

kortschak:
What exactly is the reason for retaining this now since #47 says it's signficant raison d'être of saving allocation no longer exists?

btracey:
It saves the cost of computing Sigma, which is an O(n^3). Thinking though, it may be better to export particular functions (like LogProb), where the Cholesky is already known. This saves all of the overhead (allocation, computation of sigma, etc.)

btracey:
We've done more of the latter. If there's a compelling use case, we can reconsider.

Issue #253, opened 2017-10-04
stat: Quantile returns wrong value when dataset is of even size


sbinet:
according to the universal source of truth:
 https://en.wikipedia.org/wiki/Median

the median of:
```go
xs := []float64{1, 3, 3, 6, 7, 8, 9}
```
should be `6`.

this program:
```go
package main

import (
	"fmt"

	"gonum.org/v1/gonum/stat"
)

func main() {
	xs := []float64{1, 3, 3, 6, 7, 8, 9}
	med := stat.Quantile(0.5, stat.Empirical, xs, nil)
	fmt.Printf("median=%v\n", med)
}
```

indeed prints:
```
$> go run ./main.go
median=6
```

but, for this dataset: `xs := []float64{1, 2, 3, 4, 5, 6, 8, 9}`, it should return `(4 + 5) ÷ 2` _ie_: `4.5`.

`gonum/stat.Quantile` returns `4`:

```go
package main

import (
	"fmt"

	"gonum.org/v1/gonum/stat"
)

func main() {
	xs := []float64{1, 2, 3, 4, 5, 6, 8, 9}
	med := stat.Quantile(0.5, stat.Empirical, xs, nil)
	fmt.Printf("median=%v\n", med)
}
```

```
$> go run ./main.go 
median=4
```

btracey:
The documentation could probably be improved, but this isn't a bug. This is the behavior set by CumulantKind. There are a bunch of different kinds https://en.wikipedia.org/wiki/Quantile#Estimating_quantiles_from_a_sample that can be implemented. You can also see our initial discussion here: https://github.com/gonum/stat/pull/1

sbinet:
ok. so perhaps have a dedicated `stat.Median` function, then?

btracey:
For example, I think the behavior you want is R2 on that list. Presumably in the weighted case it should do a weighted averaging.

sbinet:
ok. `R2Cumulant` ? or just `R2` ?

btracey:
There must be a better name than that. `AvgDiscont` (which is what it does)?

sbinet:
SGTM

sbinet:
see gonum/gonum#254.

btracey:
This was solved by https://github.com/gonum/gonum/pull/605 , right?

sbinet:
Fixed by #605, yes.

PR #254, opened 2017-10-04
stat: implement AvgDiscont CumulantKind for Quantile

sbinet:
Also add a simple example for 2-Quantile (aka "median".)

Fixes gonum/gonum#253.

### DIFF
--- a/stat/stat.go
+++ b/stat/stat.go
@@ -14,12 +14,15 @@ import (
 // CumulantKind specifies the behavior for calculating the empirical CDF or Quantile
 type CumulantKind int
 
+// List of supported CumulantKind values for the Quantile function.
+// Constant values should match the R nomenclature. See
+// https://en.wikipedia.org/wiki/Quantile#Estimating_the_quantiles_of_a_population
 const (
-	// Constant values should match the R nomenclature. See
-	// https://en.wikipedia.org/wiki/Quantile#Estimating_the_quantiles_of_a_population
-
 	// Empirical treats the distribution as the actual empirical distribution.
 	Empirical CumulantKind = 1
+
+	// AvgDiscont is like Empirical but with averaging at discontinuities.
+	AvgDiscont CumulantKind = 2
 )
 
 // bhattacharyyaCoeff computes the Bhattacharyya Coefficient for probability distributions given by:
@@ -965,6 +968,7 @@ func MomentAbout(moment float64, x []float64, mean float64, weights []float64) f
 // CumulantKind behaviors:
 //  - Empirical: Returns the lowest value q for which q is greater than or equal
 //  to the fraction p of samples
+//  - AvgDiscont: like Empirical, but it returns an averaging at discontinuities.
 func Quantile(p float64, c CumulantKind, x, weights []float64) float64 {
 	if !(p >= 0 && p <= 1) {
 		panic("stat: percentile out of bounds")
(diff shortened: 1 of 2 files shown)


stat/stat.go
```diff
@@ -1001,6 +1004,23 @@ func Quantile(p float64, c CumulantKind, x, weights []float64) float64 {
...
+		var cumsum float64
+		fidx := p * sumWeights
+		for i := range x {
+			if weights == nil {
+				cumsum++
+			} else {
+				cumsum += weights[i]
+			}
+			if cumsum >= fidx {
+				if len(x)%2 != 0 {
```
btracey: This is not right. It's not about the length of the slice being even or odd, it's about the percentile not being able to be met exactly. For instance, in []float64{1, 2}, you'd expect to get a cumulant of 1/3rd exactly. The smoothing needs to be about p_low and p_high (when you add the previous and next sample).

sbinet:
PTAL
(I'll squash if it's good)

stat/stat.go
```diff
@@ -19,7 +19,10 @@ const (
 	// https://en.wikipedia.org/wiki/Quantile#Estimating_the_quantiles_of_a_population
 
 	// Empirical treats the distribution as the actual empirical distribution.
-	Empirical CumulantKind = 1
+	Empirical CumulantKind = 1 + iota
```
kortschak: I don't think `iota` is appropriate here. The values we are using match the quantile types provided by R and are not just a sequence of enumerated values.

Would it be worth linking to the table in https://en.wikipedia.org/wiki/Quantile#Estimating_quantiles_from_a_sample if this list gets longer?

stat/stat.go
```diff
@@ -1001,6 +1005,42 @@ func Quantile(p float64, c CumulantKind, x, weights []float64) float64 {
 			}
 		}
 		panic("impossible")
+	case AvgDiscont:
```
kortschak: With this addition, I think it may be worth factoring the functions out into helpers.

stat/stat.go
sbinet: done.

stat/stat.go
sbinet: done.

sbinet:
Let me squash a bit then :)

sbinet:
(I'll wait for @btracey as well)

kortschak:
Thank you.

sbinet:
ping @btracey

stat/stat.go
```diff
@@ -988,24 +992,68 @@ func Quantile(p float64, c CumulantKind, x, weights []float64) float64 {
...
+		}
+		if cumsum >= fidx {
+			return x[i]
+		}
+	}
+	panic("impossible")
+}
+
+func avgDiscontQuantile(p float64, x, weights []float64, sumWeights float64) float64 {
+	// use algorithm from:
```
btracey: If you agree with the argument about type 4 above, then this can change to almost the algorithm above, except with the proper smoothing given the weights and the `p`.

stat/stat.go
```diff
@@ -14,12 +14,15 @@ import (
...
+// Constant values should match the R nomenclature. See
+// https://en.wikipedia.org/wiki/Quantile#Estimating_the_quantiles_of_a_population
 const (
-	// Constant values should match the R nomenclature. See
-	// https://en.wikipedia.org/wiki/Quantile#Estimating_the_quantiles_of_a_population
-
 	// Empirical treats the distribution as the actual empirical distribution.
 	Empirical CumulantKind = 1
+
+	// AvgDiscont is like Empirical but with averaging at discontinuities.
```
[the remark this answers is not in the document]
sbinet: type 2 is something people coming from `R` might expect.
also, it's the expected behaviour (at least for me) when dealing with unique, non-weighted data.

I'd leave it there and volounteer myself to implement type 4 (in the next PR)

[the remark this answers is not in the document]
btracey: We can have it for compatibility, but I really don't think it's what most people want in general. I agree that it gives the answer you're expecting for unique, non-weighted data, but Type 4 gives the same answer. I think we should encourage people to use Type 4 in documentation and examples, because it gives the same behavior as type 2 in the common case, and gives a much less surprising answer in all other cases (p != 0.5 or weights != nil).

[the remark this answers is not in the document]
btracey: I think the name should be something like `HalfBetween`, because it doesn't really average across the discontinuity in a meaningful sense, it just replaces the discontinuity with a different discontinuity. Instead, it takes half of the sample ends if the quantile falls between samples. 

With weighted data, it's not clear to me if the extension is half, or if the extension is the average of the two weights (i.e. (w_1 x_ + w_2 x_2)/(w_1 + w_2)). I think the half probably makes more sense, because that's how it would work if you replicated the data.

stat/stat.go
```diff
@@ -988,24 +992,68 @@ func Quantile(p float64, c CumulantKind, x, weights []float64) float64 {
...
+
+func avgDiscontQuantile(p float64, x, weights []float64, sumWeights float64) float64 {
+	// use algorithm from:
+	//  Hyndman, R.J and Fan, Y, (1996)
+	//  "Sample quantiles in statistical packages"
+	//  American Statistician, 50, 361-365
+	//
+	// also available from R's documentation:
+	//  https://stat.ethz.ch/R-manual/R-devel/library/stats/html/quantile.html
+	fidx := p * sumWeights
```
btracey: I can't find this algorithm in the paper you link (seems like it should be in section 2, but I don't see it). However, this really doesn't look like it works in general. If the weights are uneven, there's no reason that `p * sumWeights` should be close to the right place in the distribution. It may be a good initial guess, but this doesn't do any iteration to find the right answer. 

I think this should have a loop that looks like empiricalQuantile, but it returns `x[i] ` if `cumsum == fidx`, and returns `(x[i] + x[i+1])/2` if it's not equal.

stat/stat_test.go
```diff
@@ -1625,5 +1626,75 @@ func TestStdScore(t *testing.T) {
 			t.Errorf("StdScore mismatch case %d. Expected %v, Found %v", i, test.z, z)
 		}
 	}
+}
+
+func ExampleQuantile_median() {
```
btracey: As said above, I think we should take this out until we have the linear piecewise approximation.

stat/stat_test.go
```diff
@@ -1625,5 +1626,75 @@ func TestStdScore(t *testing.T) {
...
+	x2 := []float64{1, 2, 3, 4, 5, 6, 8, 9}
+	m2 := Quantile(0.5, AvgDiscont, x2, weights)
+	fmt.Printf("median = %v\n", m2)
+
+	// Output:
+	// median = 6
+	// median = 4.5
+}
+
+func TestQuantileMedian(t *testing.T) {
```
btracey: I think this test should be replaced with a `TestQuantileHalfBetween` (or whatever we call it). We need a more robust set of tests for this case before it can go in. Tests like `xs := []float64{10,20}` should return `10` for `p = 0`, `20` for `p = 1`, and `15` for all other cases. Similarly, there should be weighted tests, where if the `p` matches the weights perfectly, the correct `x` is returned, otherwise it's the average of the two `xs`.

kortschak:
*ping*

sbinet:
I've lost interest in this.
I'll push a version that doesn't implement 'type 2' but cleans up a little bit the code for "empirical" (so somebody else may tackle type 2, 4 or whatever).

sbinet:
abandonned in favor of https://github.com/gonum/gonum/pull/518.

PR #261, opened 2017-10-12
mat/: implement Vector convolution

kczimm:
First of all, this is my first time contributing on github so I apologize in advance if I'm going about this wrong.

I'm attempting to implement the convolution operation as a VecDense method. I thought this might belong in the vector.go file so that is where I'm putting it.

I thought that convolution was a much needed operation missing from the gonum package so I'd like to contribute by attempting to add it.


kortschak:
It would be good to have an feature request issue prior to allow us to consider the need before reviewing. Would you please do that.

kczimm:
I can add a feature request issue per your request.

PR #263, opened 2017-10-13
mat: added a few godoc examples for Dense methods

jchatkinson:
Instead of adding all examples at once, I just added a few to see if anyone has comments before working on the rest.

### DIFF
--- a/mat/dense_arithmetic_test.go
+++ b/mat/dense_arithmetic_test.go
@@ -11,32 +11,34 @@ import (
 )
 
 func ExampleDense_Add() {
-	// Initialize two matrices, a and b
+	// Initialize two matrices, a and b.
 	a := mat.NewDense(2, 2, []float64{1, 0, 1, 0})
 	b := mat.NewDense(2, 2, []float64{0, 1, 0, 1})
 
-	// Add a and b, placing the result into b
-	b.Add(a, b)
+	// Add a and b, placing the result into c.
+	// ...Notice that the size is automatically adjusted when the receiver has zero size.
+	var c mat.Dense
+	c.Add(a, b)
 
-	// print the result using the formatter.
-	fb := mat.Formatted(b, mat.Prefix("    "), mat.Squeeze())
-	fmt.Printf("Result:\nb = %v\n\n", fb)
+	// Print the result using the formatter.
+	fc := mat.Formatted(&c, mat.Prefix("    "), mat.Squeeze())
+	fmt.Printf("Result:\nc = %v\n\n", fc)
 	// Output:
 	// Result:
-	// b = ⎡1  1⎤
+	// c = ⎡1  1⎤
 	//     ⎣1  1⎦
 	//
 }
 
 func ExampleDense_Sub() {
-	// Initialize two matrices, a and b
+	// Initialize two matrices, a and b.
 	a := mat.NewDense(2, 2, []float64{1, 1, 1, 1})
 	b := mat.NewDense(2, 2, []float64{1, 0, 0, 1})
 
-	// Subtract b from a, placing the result into a
+	// Subtract b from a, placing the result into a.
 	a.Sub(a, b)
 
-	// print the result using the formatter.
+	// Print the result using the formatter.
 	fa := mat.Formatted(a, mat.Prefix("    "), mat.Squeeze())
 	fmt.Printf("Result:\na = %v\n\n", fa)
 	// Output:
@@ -47,14 +49,14 @@ func ExampleDense_Sub() {
 }
 
 func ExampleDense_MulElem() {
-	// Initialize two matrices, a and b
+	// Initialize two matrices, a and b.
 	a := mat.NewDense(2, 2, []float64{1, 2, 3, 4})
 	b := mat.NewDense(2, 2, []float64{1, 2, 3, 4})
 
-	// Muliply the elements of a and b, placing the result into a
+	// Multiply the elements of a and b, placing the result into a.
 	a.MulElem(a, b)
 
-	// print the result using the formatter.
+	// Print the result using the formatter.
 	fa := mat.Formatted(a, mat.Prefix("    "), mat.Squeeze())
 	fmt.Printf("Result:\na = %v\n\n", fa)
 	// Output:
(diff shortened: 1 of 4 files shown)


mat/dense_arithmetic_test.go
```diff
@@ -0,0 +1,109 @@
...
+
+	"gonum.org/v1/gonum/mat"
+)
+
+func ExampleDense_Add() {
+	// Initialize two matrices, a and b
+	a := mat.NewDense(2, 2, []float64{1, 0, 1, 0})
+	b := mat.NewDense(2, 2, []float64{0, 1, 0, 1})
+
+	// Add a and b, placing the result into b
```
btracey: Do you think it would be instructive here to also have a new matrix? Something like

```
// Add a and b, placing the result into c. Notice that the size is automatically adjusted when the receiver has zero size.
var c mat.Dense
c.Add(a, b)
```

mat/dense_arithmetic_test.go
```diff
@@ -0,0 +1,109 @@
...
+	b.Add(a, b)
+
+	// print the result using the formatter.
```
btracey: s/print/Print

mat/dense_arithmetic_test.go
```diff
@@ -0,0 +1,109 @@
...
+
+func ExampleDense_Sub() {
+	// Initialize two matrices, a and b
+	a := mat.NewDense(2, 2, []float64{1, 1, 1, 1})
+	b := mat.NewDense(2, 2, []float64{1, 0, 0, 1})
+
+	// Subtract b from a, placing the result into a
+	a.Sub(a, b)
+
+	// print the result using the formatter.
```
btracey: s/print/Print

mat/dense_arithmetic_test.go
```diff
@@ -0,0 +1,109 @@
...
+	//     ⎣1  0⎦
+	//
+}
+
+func ExampleDense_MulElem() {
+	// Initialize two matrices, a and b
+	a := mat.NewDense(2, 2, []float64{1, 2, 3, 4})
+	b := mat.NewDense(2, 2, []float64{1, 2, 3, 4})
+
+	// Muliply the elements of a and b, placing the result into a
```
btracey: s/Muliply/Multiply

mat/dense_arithmetic_test.go
```diff
@@ -0,0 +1,109 @@
...
+	a.MulElem(a, b)
+
+	// print the result using the formatter.
```
btracey: s/print/Print

mat/dense_arithmetic_test.go
```diff
@@ -0,0 +1,109 @@
...
+	fa := mat.Formatted(a, mat.Prefix("    "), mat.Squeeze())
+	fmt.Printf("Result:\na = %v\n\n", fa)
+	// Output:
+	// Result:
+	// a = ⎡1  2⎤
+	//     ⎣3  4⎦
+	//
+}
+
+func ExampleDense_Inverse() {
```
btracey: It's probably okay to include this, but it really should be noted that taking the explicit inverse should be avoided if at all possible. Specifically, this should point people to Solve (maybe an actual Solve example here?) for computing A^-1 b.

mat/dense_arithmetic_test.go
```diff
@@ -0,0 +1,109 @@
...
+	// Initialize two matrices, a and ia
+	a := mat.NewDense(2, 2, []float64{4, 0, 0, 4})
+	ia := mat.NewDense(2, 2, []float64{0, 0, 0, 0})
+
+	// Take the inverse of a and place the result in ia
+	ia.Inverse(a)
+
+	// print the result using the formatter.
```
btracey: s/print/Print

mat/dense_arithmetic_test.go
```diff
@@ -0,0 +1,109 @@
...
+	fa := mat.Formatted(ia, mat.Prefix("     "), mat.Squeeze())
+	fmt.Printf("Result:\nia = %.2g\n\n", fa)
+
+	// Show that [A] * [A]^(-1) = [I]
+	ia.Mul(a, ia)
```
btracey: I don't think you should overwrite `ia`, it should stay the inverse of a. Instead do
```
var result mat.Dense
result.Mul(a,ia)
```
That way it's clear one is the inverse of a, and the other is the result of a different operation.

mat/dense_arithmetic_test.go
jchatkinson: Thanks Brendan, all good comments. I have updated accordingly.

mat/dense_arithmetic_test.go
```diff
@@ -0,0 +1,128 @@
...
+	var ia mat.Dense
+
+	// Take the inverse of a and place the result in ia.
+	ia.Inverse(a)
+
+	// Print the result using the formatter.
+	fa := mat.Formatted(&ia, mat.Prefix("     "), mat.Squeeze())
+	fmt.Printf("Result:\nia = %.2g\n\n", fa)
+
+	// Show that [A] * [A]^(-1) = [I]
```
btracey: Remove the brackets.

btracey: s/Show/Confirm?

mat/dense_arithmetic_test.go
```diff
@@ -0,0 +1,128 @@
...
+	var r mat.Dense
+	r.Mul(a, &ia)
+	fr := mat.Formatted(&r, mat.Prefix("    "), mat.Squeeze())
+	fmt.Printf("Result:\nr = %v\n\n", fr)
+
+	// An alternative to using Inverse is to solve [A] * [X] = [I].
```
btracey: Sorry, this isn't what I meant. Solving A*x = I is basically equivalent to the `Inverse` function if I remember correctly.

What I meant is something like
```
The Inverse operation, however, is numerically unstable, and should typically be avoided.
For example, a common need is to find x = A^-1 * b. In this case, the SolveVec method of VecDense
(if b is a Vector) or Solve method of Dense (if b is a matrix) should used instead of computing
the Inverse of A.
```
And then give an example with a set `b` vector, and use `SolveVec`.

mat/dense_arithmetic_test.go
```diff
@@ -0,0 +1,128 @@
+// Copyright ©2013 The gonum Authors. All rights reserved.
```
btracey: Copyright date needs to change to 2017

btracey: Please rename the file to `dense_example_test.go`

mat/dense_arithmetic_test.go
jchatkinson: I've added another commit with these changes, please see if the new example is more in line with what you were thinking.

btracey:
Thank you!

PR #264, opened 2017-10-13
stat/samplemv: add Halton sampling procedure

btracey:


btracey:
While only one kind is implemented now, I know there are a few different scrambling procedures that could be useful at some point (including unscrambled for comparison purposes).   

I'm not sure how to write a meaningful test for this. I have tested this by sampling from the distribution, and visually confirming that it "looks" like it was generated from the Halton (i.e. points nicely spread out). I did find a bug in my initial implementation this way (they looked "iid random"). I don't know how to translate this intuition into a robust test that doesn't just re-implement the algorithm.

stat/samplemv/halton.go
```diff
@@ -0,0 +1,166 @@
...
+	// HaltonOwen generates (scrambled) Halton samples using the Randomized van der Corput
+	// algorithm described in
+	//  A randomized Halton algorithm
+	//  Art Owen
+	//  https://arxiv.org/pdf/1706.02808.pdf
+	// Currently limited to 1000 dimensional inputs.
+	Owen = iota + 1
+)
+
+// Halton generates rows(batch) samples using Halton sequence from
```
kortschak: Should "Halton sequence" be plural? It seems so from usage in the paper.

stat/samplemv/halton.go
```diff
@@ -0,0 +1,166 @@
...
+	perm := rand.Perm
+	if src != nil {
+		perm = src.Perm
+	}
+
+	n, d := batch.Dims()
+	// Each case should generate random numbers over the unit cube.
+	switch h {
+	default:
+		panic("halton: unknown HaltonType")
```
kortschak: s/HaltonType/HaltonKind/

kortschak:
I've looked at the implementation, but I'm finding it difficult to map between the Go and the R that Owen provides.

stat/samplemv/halton.go
```diff
@@ -0,0 +1,166 @@
...
+
+import (
+	"fmt"
+	"math/rand"
+
+	"gonum.org/v1/gonum/mat"
+	"gonum.org/v1/gonum/stat/distmv"
+)
+
+// Haltoner is a wrapper around the Halton sampling generation method.
```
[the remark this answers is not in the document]
btracey: But we have `Rejectioner`, `LatinHypercuber`, `MetropolisHastingser`. `Haltoner` is the name that fits that pattern.

stat/samplemv/halton.go
btracey: I don't know. The paper is about a generic algorithm over all Halton sequences. The wikipedia is similar. The wikipedia then talks about "let's take one dimension of the Halton sequence to be based on 2 and the other on 3". Similarly, Owen's paper says "The Halton sequence has form [blas] for different bases ... the discrepancy of the Halton sequence is [blah]".

I took this to mean that a single Halton sequence is for a specific choice of primes, and we don't give a choice of primes, so we use a specific Halton sequence. Not sure that I'm right though.

btracey:
Mapping from R to Go
1) Owen's code allocates a lot of vectors. This loops over the samples to avoid the allocations.
2) You can see on the bottom of page 25 that each dimension is generated individually, and that `b` is the n'th prime
3) Note that the permutation is only generated once for all the samples, hence outside the sample loop
4) The trickiest thing is that Owen's algorithm makes a list of residuals stored for each matrix. For each sample, it takes
```
dig := res % b
res = (res-dig)/b
```
Note that the update to res is the same as `res/b` in Go, since Go uses integer division. Next, notice that the division is by powers of `b`, so rather than store successive entries of `res/b`, we can just compute `i/b^n` directly, which is the role of `div`.

stat/samplemv/halton.go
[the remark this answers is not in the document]
kortschak: Those are unfortunate; they fit the naming convention for interfaces, which none of these is. I would name those `Rejection`, `LatinHypercube` and `MetropolisHastings`, but I think I have missed my opportunity when I should have spoken up (under that this would be `Halton`).

The comment wording should change though.

stat/samplemv/halton.go
kortschak: I searched through the Owen paper and the vast majority of uses were a noun with an article.

stat/samplemv/halton.go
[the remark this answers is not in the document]
kortschak: Looking at the docs, I see why these are the way they are. Still unfortunate, and I guess this has to be this way.

kortschak:
Yeah, 4. was the issue for me. The code looks good to me. Do you have any ideas about how to test it yet? A `spatial.GlobalMoransI` statistic might work.

stat/samplemv/halton.go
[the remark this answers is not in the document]
btracey: Let's put this in to match the pattern, and we can consider changing them en masse.

stat/samplemv/halton.go
btracey: "The" added

btracey:
Yes, I added a proper test.

There is one problem with this algorithm, which is that the first dimension always uses the first prime, which means it's going to have higher variance in it's buckets, and have a different set of points with non-zero location. I wonder if we should permute the primes or using random primes.

Maybe we just leave that to a different HaltonKind, since this is in fact Owen's algorithm

stat/samplemv/halton.go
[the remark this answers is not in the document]
btracey: In particular, maybe we just need the types and don't need the specific functions. The difference between  `LatinHypercube(batch,q, src)` and `LatinHypercube{Q: q, Src: src}.Sample(batch)` isn't so big.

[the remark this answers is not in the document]
kortschak: SGTM for further discussion.

btracey:
Sorry, I didn't see that the push failed. Updated now

PTAL @kortschak

stat/samplemv/halton_test.go
```diff
@@ -0,0 +1,44 @@
...
+
+		// In each dimension, the samples should be stratefied according to the
+		// prime for that dimension. There should be at most 1 sample per
+		// 1/b^k block, where k is log(n)/log(b).
+		for d := 0; d < test.d; d++ {
+			b := float64(nthPrime(d))
+			fk := math.Log(float64(test.n)) / math.Log(b)
+			k := math.Ceil(fk)
+			den := math.Pow(b, k)
+			m := make(map[int]struct{})
```
kortschak: Is it worth making this a `map[int]int` and counting them? In the happy case it makes no difference, but when debugging it might help. (I'm being extremely pessimistic here given that this passes).

stat/samplemv/halton_test.go
```diff
@@ -0,0 +1,44 @@
...
+	for cas, test := range []struct {
+		n int
+		d int
+	}{
+		{10, 1},
+		{100, 2},
+		{1000, 3},
+	} {
+		src := rand.New(rand.NewSource(1))
+		// Generate the samples
```
kortschak: Dot.

btracey: Done.

stat/samplemv/halton_test.go
btracey: Done.

stat/samplemv/halton_test.go
```diff
@@ -0,0 +1,46 @@
+package samplemv
```
kortschak: Here too.

stat/samplemv/halton.go
```diff
@@ -0,0 +1,166 @@
+package samplemv
```
kortschak: Ooop. License header.

stat/samplemv/halton_test.go
btracey: Shoot, sorry. Done.

btracey:
Thanks for the review.

PR #267, opened 2017-10-18
mat: Add Scale method for TriDense and Cholesky

btracey:


mat/cholesky.go
```diff
@@ -308,6 +308,31 @@ func (c *Cholesky) InverseTo(s *SymDense) error {
...
+	if !orig.valid() {
+		panic(badCholesky)
+	}
+	n := orig.Size()
+	if c.isZero() {
+		c.chol = NewTriDense(n, Upper, nil)
+	} else if c.chol.mat.N != n {
+		panic(ErrShape)
+	}
+	if f <= 0 {
```
kortschak: Move this earlier?

mat/cholesky.go
```diff
@@ -308,6 +308,31 @@ func (c *Cholesky) InverseTo(s *SymDense) error {
 	return err
 }
 
+// Scale multiplies the original matrix A by a non-negative constant using
+// its Cholesky decomposition, storing the result in-place into the receiver.
+// That is, if the original Cholesky factorization is
+//  U^T * U = A
+// the updated factorization is
+//  U'^T * U' = f A = A'
+// Scale panics if the constant is non-positive, or if the receiver is non-zero
```
kortschak: s/non-positive/negative/

mat/cholesky_test.go
```diff
@@ -421,6 +422,60 @@ func TestCholeskySymRankOne(t *testing.T) {
...
+		if !ok {
+			t.Errorf("Case %v: bad test, Cholesky factorization failed", cas)
+			continue
+		}
+
+		// Compare the update to a new Cholesky to an update in-place.
+		var cholUpdate Cholesky
+		cholUpdate.Scale(test.f, &chol)
+		chol.Scale(test.f, &chol)
+		if !equalApproxChol(&chol, &cholUpdate, 1e-14, 1e-14) {
```
kortschak: Shouldn't these be exactly equal?

mat/cholesky.go
btracey: It also panics if the constant is zero though.

mat/cholesky.go

mat/cholesky.go
kortschak: Then change the wording in the first sentence to "positive"? I'd have the wording in this line as "not positive" rather than "non-positive".

mat/cholesky_test.go
```diff
@@ -466,6 +466,14 @@ func TestCholeskyScale(t *testing.T) {
 	}
 }
 
+// equalApproxChol checks that the two Cholesky decompositions are equal.
+func equalChol(a, b *Cholesky) bool {
+	if !Equal(a.chol, b.chol) {
```
kortschak: `return Equal(a.chol, b.chol) && a.cond == b.cond`

btracey:
Thanks

PR #271, opened 2017-10-18
stat/distuv: implement Poisson with inversion method

kczimm:
This is just an update of @sbinet PR of the Poisson implementation. I updated the Rand method to use the inversion method based on Lambda. The implementation was ported from the Root TRandom. I don't know the standard for giving attribution to the original algorithm implementation.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -35,6 +35,7 @@ Jonathan Schroeder <jd.schroeder@gmail.com>
 Joseph Watson <jtwatson@linux-consulting.us>
 Julien Roland <juroland@gmail.com>
 Kent English <kent.english@gmail.com>
+Kevin C. Zimmerman <kevinczimmerman@gmail.com>
 Konstantin Shaposhnikov <k.shaposhnikov@gmail.com>
 Leonid Kneller <recondite.matter@gmail.com>
 Lyron Winderbaum <lyron.winderbaum@student.adelaide.edu.au>
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -42,6 +42,7 @@ Jonathan Schroeder <jd.schroeder@gmail.com>
 Joseph Watson <jtwatson@linux-consulting.us>
 Julien Roland <juroland@gmail.com>
 Kent English <kent.english@gmail.com>
+Kevin C. Zimmerman <kevinczimmerman@gmail.com>
 Konstantin Shaposhnikov <k.shaposhnikov@gmail.com>
 Leonid Kneller <recondite.matter@gmail.com>
 Lyron Winderbaum <lyron.winderbaum@student.adelaide.edu.au>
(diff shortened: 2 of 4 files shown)


btracey:
Sorry, I'm sorry to close this without looking, but ROOT is LGPL. We cannot use LGPL code. I am not going to look at this so I can (help) implement `Rand` from the papers that have been linked.

kczimm:
@btracey don't apologize. I didn't know what was allowed.

sbinet:
Theoretically, reimplementing an algorithm and changing the underlying language (so C++ -> Go) can not be considered derived work, according to our CNRS lawyer.
But better safe than sorry :)

btracey:
Okay. I've never been able to find anything that says that which isn't caveated by lots of "I am not a lawyer but" statements. I'd still rather be safe than sorry, especially since CNRS is unlikely to defend us in court :-) . If we ever really need to have this discussion we can.

In general though, we'd like to keep extra license attributions minimal, and I think in this case there is good enough non-licensed documentation to avoid it here.

kczimm:
I understand. I apologize for the inconvenience. I'm working on implementing the rejection method from page 511 of http://luc.devroye.org/chapter_ten.pdf as was mentioned earlier. 

I would just really like to see the Poisson implementation get included soon. It's a crucial distribution for a variety of applications and can be a show stopper for people wanting to migrate their scientific developments from language X to go.

btracey:
No problem. I don't like rejecting time people have put into the project, so I wanted to be clear as to why.

If you want (or if @sbinet wants), this PR is almost done ignoring the `Rand` method. It could be updated to remove the `Rand` method and fix the `PDF`, and then @kczimm could make a second PR that implements `Rand` and fixes the tests. This may help a lot of the git complexity.

kczimm:
Actually, the exact code that is also in ROOT is in the following book: http://www.aip.de/groups/soe/local/numres/bookcpdf/c7-3.pdf on page 294. Does this mean we can use this implementation? I've implemented the rejection method from Devroye but I cannot seem to get it to work.

btracey:
If the code is implemented from the book without using the source code then there shouldn't be issues (then it's clearly not a derivative work).

At this point it might be easiest to get the things that are working in (most of the PR), and then do `Rand` in another PR.

btracey:
Sorry, I can't re-open this PR, but @kczimm if you're still willing, could re-fork @sbinet's pull request? That way you can fix up the PR without the `Rand` method, and once it's through we can have a different PR with `Rand` since it seems tricky.

kczimm:
I'd be happy to do that. Would you mind telling me specifically what needs to be fixed up in @sbinet's pull request? Then I'll get started.

btracey:
There are a few outstanding comments there. There is the comment on LogProb being wrong, and comments on tests. There is also an outstanding discussion on Quantile, but that can also wait for a future PR if necessary.

PR #273, opened 2017-10-19
mat: Add RawU method to the Cholesky decomposition to directly expose…

### LINKED ISSUE #270: mat: Add Cholesky.RawU()
This is a suggestion to add 
```
// RawU returns the direct Triangular matrix used to store the Cholesky decomposition of
// the original matrix A. If returned matrix is modified, the decomposition is invalid and should
// not be used.
func (c *Cholesky) RawU() Triangular
```

Then, types like `Normal` can save the work of extracting the decomposition. For example, `*Normal.transformNormal` can have
```
dstVec.MulVec(n.chol.RawU().T(), srcVec)
```
and the entire `lower` field can be eliminated altogether, saving time, memory and code complexity.

btracey:
… the storage matrix

This is returned as Triangular rather than TriDense for two reasons
1) In the future, we may be able to optimize Cholesky, for instance by storing the decomposition as a banded matrix if the factorized matrix was banded. Returning as Triangular allows this to be robust to future modifications of Cholesky.
2) Normally one wants to return concrete types to allow maximum flexibility is using them, for example avoiding interface assertions. However, the opposite is true here. The user should not modify the returned data, and by returning a Triangular rather than a TriDense, this forces tricky code in order to be able to modify the returned matrix.

Fixes #270.

### DIFF
--- a/mat/cholesky.go
+++ b/mat/cholesky.go
@@ -232,6 +232,14 @@ func (c *Cholesky) SolveVec(v, b *VecDense) error {
 
 }
 
+// RawU returns the Triangular matrix used to store the Cholesky decomposition of
+// the original matrix A. The returned matrix should not be modified. If it is
+// modified, the decomposition is invalid and should
+// not be used.
+func (c *Cholesky) RawU() Triangular {
+	return c.chol
+}
+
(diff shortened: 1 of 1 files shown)


btracey:
PTAL @kortschak

PR #274, opened 2017-10-20
stat/distmv: remove use of lower and add method for directly generati…

btracey:
…ng a random number

### DIFF
--- a/stat/distmv/normal.go
+++ b/stat/distmv/normal.go
@@ -255,6 +255,13 @@ func (n *Normal) Quantile(x, p []float64) []float64 {
 	return x
 }
 
+// Rand generates a random number according to the distributon.
+// If the input slice is nil, new memory is allocated, otherwise the result is stored
+// in place.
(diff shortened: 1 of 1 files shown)


btracey:
PTAL @kortschak . I can split the added function in a different PR if you'd like, but I think it demonstrates some of the utility. I don't think this is a slippery slope to a lot of functions, I expect there will also be `LogProb` at some point, but not too many more.

stat/distmv/normal.go
```diff
@@ -258,25 +255,39 @@ func (n *Normal) Quantile(x, p []float64) []float64 {
...
-			tmp[i] = n.src.NormFloat64()
+			x[i] = src.NormFloat64()
 		}
 	}
-	n.transformNormal(x, tmp)
+	transformNormal(x, x, mean, chol)
 	return x
 }
 
+// Rand generates a random number according to the distributon.
```
kortschak: This would generate less diff noise if it were above `NormalRand` I think.

btracey:
Done.

PR #276, opened 2017-10-20
optimize: Add CMA-ES Cholesky global optimization method

btracey:
Fixes https://github.com/gonum/optimize/issues/180

### DIFF
--- a/optimize/cmaes.go
+++ b/optimize/cmaes.go
@@ -1,4 +1,4 @@
-// Copyright ©2016 The gonum Authors. All rights reserved.
+// Copyright ©2017 The gonum Authors. All rights reserved.
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
--- a/optimize/cmaes.go
+++ b/optimize/cmaes.go
@@ -41,9 +41,9 @@ import (
 //
 // The CMA-ES-Chol algorithm differs from the standard CMA-ES algorithm in that
 // it directly updates the Cholesky decomposition of the normal distribution.
-// This changes the runtime is O(dim^2*pop) rather than O(dim^3). The evolution
-// of the multi-variate normal will be similar to the baseline CMA-ES algorithm,
-// but the covariance update equation is not identical.
+// This changes the runtime from O(dimension^3) to O(dimension^2*population)
+// The evolution of the multi-variate normal will be similar to the baseline
+// CMA-ES algorithm, but the covariance update equation is not identical.
 //
 // For more information about the CMA-ES algorithm, see
 //  https://en.wikipedia.org/wiki/CMA-ES
@@ -71,8 +71,8 @@ type CmaEsChol struct {
 	// (log) "volume" of the normal distribution, and when it is too small
 	// the samples are almost the same. If the log determinant of the covariance
 	// matrix becomes less than StopLogDet, the optimization run is concluded.
-	// If StopDeterminant is 0, a default value of dim*log(1e-16) is used.
-	// If StopDeterminant is NaN, the stopping criteria is not used, though
+	// If StopLogDet is 0, a default value of dim*log(1e-16) is used.
+	// If StopLogDet is NaN, the stopping criteria is not used, though
 	// this can cause numeric instabilities in the algorithm.
 	StopLogDet float64
 	// ForgetBest, when true, does not track the best overall function value found,
@@ -103,10 +103,11 @@ type CmaEsChol struct {
 	chol     mat.Cholesky
 
 	// Parallel fields.
-	taskIdxs []int          // Stores which simulation the task ran.
-	evals    []int          // remaining evaluations in this iteration.
-	mux      sync.Mutex     // protect access to evals.
-	wg       sync.WaitGroup // wait for simulations to finish before iterating.
+	taskIdxs []int // Stores which simulation the task ran.
+	evals    []int // remaining evaluations in this iteration.
+
+	mux sync.Mutex     // protect access to evals.
+	wg  sync.WaitGroup // wait for simulations to finish before iterating.
 
 	// Overall best.
 	bestX []float64
@@ -124,7 +125,7 @@ func (cma *CmaEsChol) Needs() struct{ Gradient, Hessian bool } {
 
 func (cma *CmaEsChol) Done() {}
 
-// Status returns the status of the method. CMA
+// Status returns the status of the method.
 func (cma *CmaEsChol) Status() (Status, error) {
 	sd := cma.StopLogDet
 	switch {
@@ -229,7 +230,7 @@ func (cma *CmaEsChol) InitGlobal(dim, tasks int) int {
 	cma.bestF = math.Inf(1)
 
 	t := min(tasks, cma.pop)
-	cma.taskIdxs = make([]int, cma.pop)
+	cma.taskIdxs = make([]int, t)
 	for i := 0; i < t; i++ {
 		cma.taskIdxs[i] = -1
 	}
--- a/optimize/cmaes.go
+++ b/optimize/cmaes.go
@@ -9,7 +9,7 @@ import (
 	"sort"
 	"sync"
 
-	"math/rand"
+	"golang.org/x/exp/rand"
 
 	"gonum.org/v1/gonum/floats"
 	"gonum.org/v1/gonum/mat"
--- a/optimize/cmaes.go
+++ b/optimize/cmaes.go
@@ -50,12 +50,12 @@ import (
 //  https://arxiv.org/pdf/1604.00772.pdf
 type CmaEsChol struct {
 	// InitStepSize sets the initial size of the covariance matrix adaptation.
-	// If InitStepSize is 0, a default vaule of 0.5 is used. InitStepSize cannot
-	// be negative.
+	// If InitStepSize is 0, a default value of 0.5 is used. InitStepSize cannot
+	// be negative, or CmaEsChol will panic.
 	InitStepSize float64
 	// Population sets the population size for the algorithm. If Population is
 	// 0, a default value of 4 + math.Floor(3*math.Log(float64(dim))) is used.
-	// Population cannot be negative.
+	// Population cannot be negative or CmaEsChol will panic.
 	Population int
 	// InitMean is the initial mean of the multivariate normal for sampling
 	// input locations. If InitMean is nil, the zero vector is used. If InitMean
@@ -97,10 +97,10 @@ type CmaEsChol struct {
 	fs []float64
 
 	// Adaptive algorithm parameters.
-	sigma  float64
-	pc, ps []float64
-	mean   []float64
-	chol   mat.Cholesky
+	invSigma float64 // inverse of the sigma parameter
+	pc, ps   []float64
+	mean     []float64
+	chol     mat.Cholesky
 
 	// Parallel fields.
 	taskIdxs []int          // Stores which simulation the task ran.
@@ -114,8 +114,8 @@ type CmaEsChol struct {
 }
 
 var (
-	_ Statuser     = &CmaEsChol{}
-	_ GlobalMethod = &CmaEsChol{}
+	_ Statuser     = (*CmaEsChol)(nil)
+	_ GlobalMethod = (*CmaEsChol)(nil)
 )
 
 func (cma *CmaEsChol) Needs() struct{ Gradient, Hessian bool } {
@@ -182,10 +182,10 @@ func (cma *CmaEsChol) InitGlobal(dim, tasks int) int {
 	cma.fs = resize(cma.fs, cma.pop)
 
 	// Allocate and initialize adaptive parameters.
-	cma.sigma = cma.InitStepSize
-	if cma.sigma == 0 {
-		cma.sigma = 0.3
-	} else if cma.sigma < 0 {
+	cma.invSigma = 1 / cma.InitStepSize
+	if cma.InitStepSize == 0 {
+		cma.invSigma = 10.0 / 3
+	} else if cma.InitStepSize < 0 {
 		panic("cma-es-chol: negative initial step size")
 	}
 	cma.pc = resize(cma.pc, dim)
@@ -221,7 +221,7 @@ func (cma *CmaEsChol) InitGlobal(dim, tasks int) int {
 	}
 
 	cma.evals = make([]int, cma.pop)
-	for i := 0; i < cma.pop; i++ {
+	for i := range cma.evals {
 		cma.evals[i] = i
 	}
 
@@ -233,7 +233,8 @@ func (cma *CmaEsChol) InitGlobal(dim, tasks int) int {
 	for i := 0; i < t; i++ {
 		cma.taskIdxs[i] = -1
 	}
-	// Get a new waitgroup so that if the struct is re-used the.
+	// Get a new mutex and waitgroup so that if the structure is reused there
+	// aren't residual interractions with the previous optimization.
 	cma.mux = sync.Mutex{}
 	cma.wg = sync.WaitGroup{}
 	return t
@@ -241,7 +242,7 @@ func (cma *CmaEsChol) InitGlobal(dim, tasks int) int {
 
 func (cma *CmaEsChol) IterateGlobal(task int, loc *Location) (Operation, error) {
 	// Check the status of the incoming task. If it is a number, it means
-	// that taks contains a valid location.
+	// that task contains a valid location.
 	idx := cma.taskIdxs[task]
 	if idx != -1 {
 		cma.fs[idx] = loc.F
@@ -329,7 +330,7 @@ func (cma *CmaEsChol) update() error {
 
 	// p_{c,t+1} = (1-c_c) p_{c,t} + \sqrt(c_c*(2-c_c)*mueff) (m_{t+1}-m_t)/sigma_t
 	floats.Scale(1-cma.cc, cma.pc)
-	scaleC := math.Sqrt(cma.cc*(2-cma.cc)*cma.muEff) / cma.sigma
+	scaleC := math.Sqrt(cma.cc*(2-cma.cc)*cma.muEff) * cma.invSigma
 	floats.AddScaled(cma.pc, scaleC, meanDiff)
 
 	// p_{sigma, t+1} = (1-c_sigma) p_{sigma,t} + \sqrt(c_s*(2-c_s)*mueff) A_t^-1 (m_{t+1}-m_t)/sigma_t
@@ -342,7 +343,7 @@ func (cma *CmaEsChol) update() error {
 	if err != nil {
 		return err
 	}
-	scaleS := math.Sqrt(cma.cs*(2-cma.cs)*cma.muEff) / cma.sigma
+	scaleS := math.Sqrt(cma.cs*(2-cma.cs)*cma.muEff) * cma.invSigma
 	floats.AddScaled(cma.ps, scaleS, tmp)
 
 	// Compute the update to A.
@@ -355,12 +356,12 @@ func (cma *CmaEsChol) update() error {
 	for i, w := range cma.weights {
 		idx := indexes[i]
 		floats.SubTo(tmp, cma.xs.RawRowView(idx), meanOld)
-		cma.chol.SymRankOne(&cma.chol, cma.cmu*w/cma.sigma, tmpVec)
+		cma.chol.SymRankOne(&cma.chol, cma.cmu*w*cma.invSigma, tmpVec)
 	}
 
 	// sigma_{t+1} = sigma_t exp(c_sigma/d_sigma * norm(p_{sigma,t+1}/ E[chi] -1)
 	normPs := floats.Norm(cma.ps, 2)
-	cma.sigma *= math.Exp(cma.cs / cma.ds * (normPs/cma.eChi - 1))
+	cma.invSigma /= math.Exp(cma.cs / cma.ds * (normPs/cma.eChi - 1))
 	return nil
 }
 
--- a/optimize/cmaes.go
+++ b/optimize/cmaes.go
@@ -10,11 +10,9 @@ import (
 	"sort"
 	"sync"
 
-	"gonum.org/v1/gonum/stat/distmv"
-
-	"gonum.org/v1/gonum/mat"
-
 	"gonum.org/v1/gonum/floats"
+	"gonum.org/v1/gonum/mat"
+	"gonum.org/v1/gonum/stat/distmv"
 )
 
 // TODO(btracey): If we ever implement the traditional CMA-ES algorithm, provide
--- a/optimize/cmaes.go
+++ b/optimize/cmaes.go
@@ -6,9 +6,10 @@ package optimize
 
 import (
 	"math"
-	"math/rand"
 	"sort"
 	"sync"
+
+	"math/rand"
 
 	"gonum.org/v1/gonum/floats"
 	"gonum.org/v1/gonum/mat"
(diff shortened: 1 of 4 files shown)


btracey:
PTAL @vladimir-ch @sbinet @agrimball

optimize/cmaes.go
```diff
@@ -0,0 +1,383 @@
...
+package optimize
+
+import (
+	"math"
+	"math/rand"
+	"sort"
+	"sync"
+
+	"gonum.org/v1/gonum/stat/distmv"
+
```
kortschak: Delete the blank lines?

btracey: Huh. I don't know how that happened... I've only been using go imports.

btracey: Fixed.

kortschak: I think that happens with successive goimports additions that are not stdlib.

optimize/cmaes.go
```diff
@@ -0,0 +1,381 @@
...
+// This changes the runtime is O(dim^2*pop) rather than O(dim^3). The evolution
+// of the multi-variate normal will be similar to the baseline CMA-ES algorithm,
+// but the covariance update equation is not identical.
+//
+// For more information about the CMA-ES algorithm, see
+//  https://en.wikipedia.org/wiki/CMA-ES
+//  https://arxiv.org/pdf/1604.00772.pdf
+type CmaEsChol struct {
+	// InitStepSize sets the initial size of the covariance matrix adaptation.
+	// If InitStepSize is 0, a default vaule of 0.5 is used. InitStepSize cannot
```
sbinet: `s/vaule/value/`

optimize/cmaes.go
```diff
@@ -0,0 +1,381 @@
...
+	// be negative.
```
sbinet: perhaps tell the user we will panic otherwise?
(perhaps I am being too pedantic)

optimize/cmaes.go
```diff
@@ -0,0 +1,381 @@
...
+	InitStepSize float64
+	// Population sets the population size for the algorithm. If Population is
+	// 0, a default value of 4 + math.Floor(3*math.Log(float64(dim))) is used.
+	// Population cannot be negative.
```
sbinet: ditto? (but here again I might be too pedantic.)

optimize/cmaes.go
```diff
@@ -0,0 +1,381 @@
...
+	// this can cause numeric instabilities in the algorithm.
+	StopLogDet float64
+	// ForgetBest, when true, does not track the best overall function value found,
+	// instead returning the new best sample in each iteration. If ForgetBest
+	// is false, then the minimum value returned will be the lowest across all
+	// iterations, regardless of when that sample was generated.
+	ForgetBest bool
+	// Src allows a random number generator to be supplied for generating samples.
+	// If Src is nil the generator in math/rand is used.
+	Src *rand.Rand
```
sbinet: I don't remember the outcome of our previous discussion: do we call `rand.Rand` a `Src` or a `Source` ?

optimize/cmaes.go
```diff
@@ -0,0 +1,381 @@
...
+	mux      sync.Mutex     // protect access to evals.
+	wg       sync.WaitGroup // wait for simulations to finish before iterating.
+
+	// Overall best.
+	bestX []float64
+	bestF float64
+}
+
+var (
+	_ Statuser     = &CmaEsChol{}
```
sbinet: I would write this as:

```go
var (
    _ Statuser     = (*CmaEsChol)(nil)
    _ GlobalMethod = (*CmaEsChol)(nil)
)
```

optimize/cmaes.go
```diff
@@ -0,0 +1,381 @@
...
+		var chol mat.Cholesky
+		ok := chol.Factorize(b)
+		if !ok {
+			panic("cma-es-chol: bad cholesky. shouldn't happen")
+		}
+		cma.chol = chol
+	}
+
+	cma.evals = make([]int, cma.pop)
+	for i := 0; i < cma.pop; i++ {
```
sbinet: perhaps rewrite this as:

```go
for i := range cma.evals {
```

optimize/cmaes.go
```diff
@@ -0,0 +1,381 @@
...
+
+	cma.bestX = resize(cma.bestX, dim)
+	cma.bestF = math.Inf(1)
+
+	t := min(tasks, cma.pop)
+	cma.taskIdxs = make([]int, cma.pop)
+	for i := 0; i < t; i++ {
+		cma.taskIdxs[i] = -1
+	}
+	// Get a new waitgroup so that if the struct is re-used the.
```
sbinet: it seems we were interrupted mid-sentence :)

optimize/cmaes.go
```diff
@@ -0,0 +1,381 @@
...
+	cma.mux = sync.Mutex{}
+	cma.wg = sync.WaitGroup{}
+	return t
+}
+
+func (cma *CmaEsChol) IterateGlobal(task int, loc *Location) (Operation, error) {
+	// Check the status of the incoming task. If it is a number, it means
+	// that taks contains a valid location.
```
sbinet: `s/taks/task/`

optimize/cmaes.go
```diff
@@ -0,0 +1,381 @@
...
+	if idx != -1 {
+		cma.fs[idx] = loc.F
+		cma.wg.Done()
+	}
+
+	// Get the next task and send it to be run if there is a next task to be run.
+	// If all of the tasks have been run, perform an update step. Note that the
+	// use of this mutex means that only one task can proceed, all of the
+	// other tasks should get stuck and then get a new location.
+	cma.mux.Lock()
```
sbinet: I think I'd be inclined to use a `defer cma.mux.Unlock()` here.
(and remove the direct calls at lines 261 and 301)
what do you think?

optimize/cmaes.go
```diff
@@ -0,0 +1,381 @@
...
+
+	meanOld := make([]float64, len(cma.mean))
+	copy(meanOld, cma.mean)
+
+	// m_{t+1} = \sum_{i=1}^mu w_i x_i
+	for i := range cma.mean {
+		cma.mean[i] = 0
+	}
+	for i, w := range cma.weights {
+		idx := indexes[i] // index of teh 1337 sample.
```
sbinet: `s/teh/the/`

optimize/cmaes.go
```diff
@@ -0,0 +1,381 @@
...
+		floats.AddScaled(cma.mean, w, cma.xs.RawRowView(idx))
+	}
+	meanDiff := make([]float64, len(cma.mean))
+	floats.SubTo(meanDiff, cma.mean, meanOld)
+
+	// p_{c,t+1} = (1-c_c) p_{c,t} + \sqrt(c_c*(2-c_c)*mueff) (m_{t+1}-m_t)/sigma_t
+	floats.Scale(1-cma.cc, cma.pc)
+	scaleC := math.Sqrt(cma.cc*(2-cma.cc)*cma.muEff) / cma.sigma
```
sbinet: we divide by `cma.sigma` a lot in this `update` method (which is probably on the hot path).
perhaps we should consider storing the inverse in a local variable and multiply by it here (line-332) and at lines 345 and 358 ?

optimize/cmaes.go
btracey: Done.

optimize/cmaes.go
btracey: Done.

optimize/cmaes.go
btracey: Done.

optimize/cmaes.go
btracey: It's Src pretty much everywhere (and Source a few places)

optimize/cmaes.go
btracey: Done.

optimize/cmaes.go
btracey: Done

optimize/cmaes.go
btracey: Oh yea. Because I decided I should run the test before I spent the time writing the comment, and then the test failed :-) . Turned out to be a different bug.

btracey: Done.

optimize/cmaes.go
btracey: Done.

optimize/cmaes.go
btracey: I thought about it, but there's a few lines in the `if` statement that don't have to be protected by a mutex, and relative to the part that needs to be protected, the part that doesn't need to be protected is expensive. It doesn't really matter that much though, because this algorithm is either going to be dominated by the function evaluations, or it's going to be dominated by the parameter update, so if you'd like the extra safety I can change to defer.

optimize/cmaes.go
btracey: I can change it if you'd like, but that one was part of the joke  (https://en.wikipedia.org/wiki/Leet#Vocabulary)

optimize/cmaes.go
btracey: I think the effect on runtime is negligible, but done.

Soon we won't need a few operators, between you hating division and forcing everyone to use multiplication, and me hating multiplication and forcing everyone to add in log space.

optimize/cmaes.go
sbinet: yeah, I think I'd prefer with `defer`. (I strive to follow the rules I teach my students: first the correct, robust answer. then, faster :P)

optimize/cmaes.go
sbinet: h4! you c4n 134v3 as is.

optimize/cmaes.go
sbinet: ok.

indeed:
```
$> git grep "Source" -- *.go | grep rand.Rand
stat/distuv/bernoulli.go:17:	Source *rand.Rand
stat/distuv/beta.go:29:	Source *rand.Rand
stat/distuv/exponential.go:18:	Source *rand.Rand
stat/distuv/f.go:25:	Source *rand.Rand
stat/distuv/gamma.go:29:	Source *rand.Rand
stat/distuv/laplace.go:20:	Source *rand.Rand
stat/distuv/lognormal.go:18:	Source *rand.Rand
stat/distuv/norm.go:23:	Source *rand.Rand
stat/distuv/triangle.go:15:	Source  *rand.Rand
stat/distuv/uniform.go:20:	Source *rand.Rand
stat/distuv/weibull.go:22:	Source *rand.Rand

$> git grep "Src" -- *.go | grep rand.Rand | wc -l
9

$> git grep "Source" -- *.go | grep rand.Rand | wc -l
11
```

optimize/cmaes_test.go
```diff
@@ -0,0 +1,131 @@
...
+
+func TestCmaEsChol(t *testing.T) {
+	for i, test := range []struct {
+		dim      int
+		problem  Problem
+		method   *CmaEsChol
+		settings *Settings
+		good     func(*Result, error) error
+	}{
+		{
```
vladimir-ch: I think there should definitely be some test cases with several minima with: a) initial data (mean, variance) around the global minimum such that the algorithm is unlikely to escape to a local minimum b) initial data that cover several local minima and the global minimum such that the algorithm is very likely to find the global minimum.

optimize/cmaes.go
kortschak: It's an issue to consider the change. I've had no bites there though.

optimize/cmaes.go
btracey: Actually, I do think it's better as is. The piece in the mutex in the short path is a very small constant (slice access and slice). The piece outside the mutex is a `dim^2` and then a `dim`. I'd like to keep that outside.

optimize/cmaes.go
```diff
@@ -0,0 +1,382 @@
...
+//
+// As the normal distribution is progressively updated according to the best samples,
+// it can be that the mean of the distribution is updated in a gradient-descent
+// like fashion, followed by a shrinking covariance.
+// It is recommended that the algorithm be run multiple times (with different
+// InitMean) to have a better chance of finding the global minimum.
+//
+// The CMA-ES-Chol algorithm differs from the standard CMA-ES algorithm in that
+// it directly updates the Cholesky decomposition of the normal distribution.
+// This changes the runtime is O(dim^2*pop) rather than O(dim^3). The evolution
```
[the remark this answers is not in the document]
btracey: Done.

optimize/cmaes.go
```diff
@@ -0,0 +1,382 @@
...
+	// matrix for the initial sampling distribution. If InitCholesky is nil,
+	// a default value of I is used. If it is non-nil, then it must have
+	// InitCholesky.Size() be equal to the problem dimension.
+	InitCholesky *mat.Cholesky
+	// StopLogDet sets the threshold for stopping the optimization if the
+	// distribution becomes too peaked. The log determinant is a measure of the
+	// (log) "volume" of the normal distribution, and when it is too small
+	// the samples are almost the same. If the log determinant of the covariance
+	// matrix becomes less than StopLogDet, the optimization run is concluded.
+	// If StopDeterminant is 0, a default value of dim*log(1e-16) is used.
```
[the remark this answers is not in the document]
btracey: Done.

optimize/cmaes.go
```diff
@@ -0,0 +1,382 @@
...
+	// If StopDeterminant is NaN, the stopping criteria is not used, though
+	// this can cause numeric instabilities in the algorithm.
+	StopLogDet float64
```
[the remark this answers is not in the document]
btracey: Yes, I tried it the other way and it has issues (hence the comment name error). The problem is you want something like the identity matrix being 1e-16 * I, which is (1e-16)^dim. This underflows for medium dimension.

optimize/cmaes.go
```diff
@@ -0,0 +1,382 @@
...
+	// Fixed algorithm parameters.
+	dim                 int
+	pop                 int
+	weights             []float64
+	muEff               float64
+	cc, cs, c1, cmu, ds float64
+	eChi                float64
+
+	// Function data.
+	xs *mat.Dense
```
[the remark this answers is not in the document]
btracey: I disagree. Each row has exactly the same number of columns, which is not guaranteed by `[][]float64` or a returning method (it would be better as a multi-dim slice but Go doesn't have those). It's very common to have a list of samples stored as a matrix, and I think that fits here.

optimize/cmaes.go
```diff
@@ -0,0 +1,382 @@
...
+	// Adaptive algorithm parameters.
+	invSigma float64 // inverse of the sigma parameter
+	pc, ps   []float64
+	mean     []float64
+	chol     mat.Cholesky
+
+	// Parallel fields.
+	taskIdxs []int          // Stores which simulation the task ran.
+	evals    []int          // remaining evaluations in this iteration.
+	mux      sync.Mutex     // protect access to evals.
```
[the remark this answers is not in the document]
btracey: newline added (not sure if that's what you meant)

optimize/cmaes.go
```diff
@@ -0,0 +1,382 @@
...
+	_ GlobalMethod = (*CmaEsChol)(nil)
+)
+
+func (cma *CmaEsChol) Needs() struct{ Gradient, Hessian bool } {
+	return struct{ Gradient, Hessian bool }{false, false}
+}
+
+func (cma *CmaEsChol) Done() {}
+
+// Status returns the status of the method. CMA
```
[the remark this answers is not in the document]
btracey: Done.

optimize/cmaes.go
```diff
@@ -0,0 +1,382 @@
...
+	cma.evals = make([]int, cma.pop)
+	for i := range cma.evals {
+		cma.evals[i] = i
+	}
+
+	cma.bestX = resize(cma.bestX, dim)
+	cma.bestF = math.Inf(1)
+
+	t := min(tasks, cma.pop)
+	cma.taskIdxs = make([]int, cma.pop)
```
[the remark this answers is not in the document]
btracey: Yes, done.

optimize/cmaes_test.go
btracey: Done.

optimize/cmaes.go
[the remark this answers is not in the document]
kortschak: By convention a mutex is placed in a stanza above the fields it protects. So this would be

```
	// Parallel fields.
	mu       sync.Mutex     // protect access to evals.
	taskIdxs []int // Stores which simulation the task ran.
	evals    []int // remaining evaluations in this iteration.

	wg sync.WaitGroup // wait for simulations to finish before iterating.
```

Note, not `mux` - this is overloaded with multiplex; either `mu` or `lock`.

optimize/cmaes.go
kortschak: There is a safer alternative, which is to place a defer after the condition here. It least its visible on the same page then when it's being read.

optimize/cmaes.go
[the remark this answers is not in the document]
vladimir-ch: >an exported method

I meant 'unexported'. Sorry.

I know it is common to use Dense as a 2D array storage without needing the matrix semantics but I still disagree with it (unless it is justified by usage in an API that takes a matrix, for example stat.PC, but that is not the case here). A downside of Dense is that one has to specify which way are the samples and which way are the variables. For this `[][]float64` (or a dedicated type or something else) would be much clearer and since it is our internal code we can make sure that all the slices would have the same length. But if you wish to keep it as Dense, then it's fine by me.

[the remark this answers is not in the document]
btracey: I think it is better with Dense.

> A downside of Dense is that one has to specify which way are the samples and which way are the variables.

It is a downside, unless we're consistent in how we use it, as in "all Gonum functions have the samples as rows", which we do so far (PCA, Covariance, etc.). This also has the advantage of being the same indexing as `[][]float64`, but with the additional safety and storage guarantees.

btracey:
PTAL @vladimir-ch

optimize/cmaes.go
```diff
@@ -0,0 +1,383 @@
...
+	cma.bestX = resize(cma.bestX, dim)
+	cma.bestF = math.Inf(1)
+
+	t := min(tasks, cma.pop)
+	cma.taskIdxs = make([]int, t)
+	for i := 0; i < t; i++ {
+		cma.taskIdxs[i] = -1
+	}
+	// Get a new mutex and waitgroup so that if the structure is reused there
+	// aren't residual interractions with the previous optimization.
```
vladimir-ch: interactions

optimize/cmaes.go
```diff
@@ -0,0 +1,383 @@
+// Copyright ©2017 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package optimize
+
+import (
+	"math"
+	"math/rand"
```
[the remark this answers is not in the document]
btracey: Done.

optimize/cmaes.go
```diff
@@ -0,0 +1,383 @@
...
+	// a default value of I is used. If it is non-nil, then it must have
+	// InitCholesky.Size() be equal to the problem dimension.
+	InitCholesky *mat.Cholesky
+	// StopLogDet sets the threshold for stopping the optimization if the
+	// distribution becomes too peaked. The log determinant is a measure of the
+	// (log) "volume" of the normal distribution, and when it is too small
+	// the samples are almost the same. If the log determinant of the covariance
+	// matrix becomes less than StopLogDet, the optimization run is concluded.
+	// If StopLogDet is 0, a default value of dim*log(1e-16) is used.
+	// If StopLogDet is NaN, the stopping criteria is not used, though
```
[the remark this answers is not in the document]
btracey: Done.

optimize/cmaes.go
```diff
@@ -0,0 +1,383 @@
...
+	// this can cause numeric instabilities in the algorithm.
+	StopLogDet float64
+	// ForgetBest, when true, does not track the best overall function value found,
+	// instead returning the new best sample in each iteration. If ForgetBest
+	// is false, then the minimum value returned will be the lowest across all
+	// iterations, regardless of when that sample was generated.
+	ForgetBest bool
+	// Src allows a random number generator to be supplied for generating samples.
+	// If Src is nil the generator in math/rand is used.
```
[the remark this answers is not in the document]
btracey: Done

optimize/cmaes.go
```diff
@@ -0,0 +1,383 @@
...
+	invSigma float64 // inverse of the sigma parameter
+	pc, ps   []float64
+	mean     []float64
+	chol     mat.Cholesky
+
+	// Parallel fields.
+	taskIdxs []int // Stores which simulation the task ran.
+	evals    []int // remaining evaluations in this iteration.
+
+	mux sync.Mutex     // protect access to evals.
```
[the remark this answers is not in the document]
btracey: I put it with taskIdxs and evals, because that's what it normally protects, though in the reset almost all of the fields are modified.

optimize/cmaes.go
```diff
@@ -0,0 +1,383 @@
...
+		sd = float64(cma.dim) * -36.8413614879 // ln(1e-16)
+	}
+	if cma.chol.LogDet() < sd {
+		return MethodConverge, nil
+	}
+	return NotTerminated, nil
+}
+
+func (cma *CmaEsChol) InitGlobal(dim, tasks int) int {
+	// Initialize the parameters
```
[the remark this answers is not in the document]
btracey: Done.

optimize/cmaes.go
```diff
@@ -0,0 +1,383 @@
...
+	if cma.InitMean != nil && len(cma.InitMean) != dim {
+		panic("cma-es-chol: initial mean must be nil or have length equal to dimension")
+	}
+
+	// Set fixed algorithm parameters.
+	// Parameter values are from https://arxiv.org/pdf/1604.00772.pdf .
+	cma.dim = dim
+	cma.pop = cma.Population
+	if cma.pop == 0 {
+		cma.pop = 4 + int(3*math.Log(float64(dim))) // Note the implicit floor.
```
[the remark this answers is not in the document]
btracey: Done.

optimize/cmaes.go
```diff
@@ -0,0 +1,383 @@
...
+	cma.fs = resize(cma.fs, cma.pop)
+
+	// Allocate and initialize adaptive parameters.
+	cma.invSigma = 1 / cma.InitStepSize
+	if cma.InitStepSize == 0 {
+		cma.invSigma = 10.0 / 3
+	} else if cma.InitStepSize < 0 {
+		panic("cma-es-chol: negative initial step size")
+	}
+	cma.pc = resize(cma.pc, dim)
```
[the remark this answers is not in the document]
btracey: We should do that in another PR since it affects all of the optimizers.

optimize/cmaes.go
```diff
@@ -0,0 +1,383 @@
...
+	for i := range cma.pc {
+		cma.pc[i] = 0
+	}
+	cma.ps = resize(cma.ps, dim)
+	for i := range cma.ps {
+		cma.ps[i] = 0
+	}
+	cma.mean = resize(cma.mean, dim)
+	if cma.InitMean != nil {
+		if len(cma.InitMean) != dim {
```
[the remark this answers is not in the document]
btracey: Nice catch, done.

optimize/cmaes.go
```diff
@@ -0,0 +1,383 @@
...
+			panic("cma-es-chol: InitMean length mismatch")
+		}
+		copy(cma.mean, cma.InitMean)
+	}
+	if cma.InitCholesky != nil {
+		cma.chol = mat.Cholesky{}
```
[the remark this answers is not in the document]
btracey: Done.

btracey:
PTAL @vladimir-ch

optimize/cmaes.go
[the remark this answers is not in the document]
vladimir-ch: That's what I meant :-)

PR #280, opened 2017-10-23
mat: don't confuse notions of zero

### LINKED ISSUE #268: mat: Cholesky should not test isZero with checking chol is nil
We provide a Reset method, presumably to allow memory reuse for the Cholesky decomposition. However, we check if the decomposition is zero by checking that the matrix is `nil`. This significantly harms re-use (and the package may have bugs when using `Cholesky.Reset()`)

kortschak:
@btracey Please take a look.

Fixes #268.

### DIFF
--- a/mat/cholesky.go
+++ b/mat/cholesky.go
@@ -68,7 +68,7 @@ func (c *Cholesky) Cond() float64 {
 // factorization must not be used.
 func (c *Cholesky) Factorize(a Symmetric) (ok bool) {
 	n := a.Symmetric()
-	if c.isZero() {
+	if c.chol == nil {
 		c.chol = NewTriDense(n, Upper, nil)
 	} else {
 		c.chol = NewTriDense(n, Upper, use(c.chol.mat.Data, n*n))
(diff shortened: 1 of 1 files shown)


mat/cholesky.go
```diff
@@ -91,7 +91,7 @@ func (c *Cholesky) Factorize(a Symmetric) (ok bool) {
 // Reset resets the factorization so that it can be reused as the receiver of a
 // dimensionally restricted operation.
 func (c *Cholesky) Reset() {
-	if !c.isZero() {
+	if !c.chol.IsZero() {
```
btracey: Does this if need to be there? Can't we just call Reset?

kortschak: Looking now, the original code is correct. `Reset` cannot be called on `nil`, otherwise it will panic when it attempt to update the fields (the code here will now too, when it checks). The `!isZero()` guards that. The transformation that I've applied elsewhere would make that clear.

kortschak:
PTAL

PR #285, opened 2017-10-26
mat: added godoc examples for dense mul/exp/pow/scale

jchatkinson:
a few more examples for the godoc - Mul, Exp, Pow, and Scale methods.

### DIFF
--- a/mat/dense_example_test.go
+++ b/mat/dense_example_test.go
@@ -128,3 +128,89 @@ func ExampleDense_Inverse() {
 	//     ⎣  0  0.5⎦
 	//
 }
+
+func ExampleDense_Mul() {
+	// Initialize two matrices, a and b.
+	a := mat.NewDense(2, 2, []float64{4, 0, 0, 4})
+	b := mat.NewDense(2, 3, []float64{4, 0, 0, 0, 0, 4})
+
+	// Take the matrix product of a and b and place the result in c.
+	var c mat.Dense
+	c.Mul(a, b)
+
+	// Print the result using the formatter.
+	fc := mat.Formatted(&c, mat.Prefix("    "), mat.Squeeze())
+	fmt.Printf("Result:\nc = %v\n\n", fc)
+	// Output:
+	// Result:
+	// c = ⎡16  0   0⎤
+	//     ⎣ 0  0  16⎦
+	//
+}
+
+func ExampleDense_Exp() {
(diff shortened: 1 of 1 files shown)


mat/dense_example_test.go
```diff
@@ -128,3 +128,89 @@ func ExampleDense_Inverse() {
...
+	c.Mul(a, b)
+
+	// Print the result using the formatter.
+	fc := mat.Formatted(&c, mat.Prefix("    "), mat.Squeeze())
+	fmt.Printf("Result:\nc = %v\n\n", fc)
+	// Output:
+	// Result:
+	// c = ⎡16  0   0⎤
+	//     ⎣ 0  0  16⎦
+	//
```
kortschak: Do we need the trailing `//`?

mat/dense_example_test.go
```diff
@@ -128,3 +128,89 @@ func ExampleDense_Inverse() {
 	//     ⎣  0  0.5⎦
 	//
 }
+
+func ExampleDense_Mul() {
+	// Initialize two matrices, a and b.
+	a := mat.NewDense(2, 2, []float64{4, 0, 0, 4})
```
kortschak: I think we should go through the examples and change the linear representation of the `[]float64` to something that represents the matrix layout:

```
a := mat.NewDense(2, 2, []float64{
    4, 0,
    0, 4,
})
```

This can happen in another PR, since there are others that look the same already.

mat/dense_example_test.go
```diff
@@ -128,3 +128,89 @@ func ExampleDense_Inverse() {
...
+
+	// Take the matrix product of a and b and place the result in c.
+	var c mat.Dense
+	c.Mul(a, b)
+
+	// Print the result using the formatter.
+	fc := mat.Formatted(&c, mat.Prefix("    "), mat.Squeeze())
+	fmt.Printf("Result:\nc = %v\n\n", fc)
+	// Output:
+	// Result:
```
kortschak: I don't think we need the `// Result:` line (see that godoc already gives us something equivalent https://godoc.org/gonum.org/v1/gonum/mat#example-Dense-Add), but again this can happen in another PR.

mat/dense_example_test.go
jchatkinson: seems reasonable, it makes it easier to visualize the matrix at a glance.

PR #287, opened 2017-10-26
stat/sampleuv: add missing return

kortschak:
@btracey Please take a look.


btracey:
Can be fixed in a different PR

kortschak:
Perm is linear - it's a single swapping sweep over n elements.

btracey:
Now that you mention it, I'm sure you've pointed this out to me before. Sorry.

kortschak:
The alternative algorithm is worse than O(n^2) though, it looks like O(n^2 log n) to me (n = |idxs|) because of the sort.

PR #292, opened 2017-10-27
graph/network: add heat diffusion propagation function

kortschak:
@btracey @vladimir-ch Please take a look.

### DIFF
--- a/graph/network/diffusion.go
+++ b/graph/network/diffusion.go
@@ -20,7 +20,8 @@ import (
 // Laplacian Index field.
 //
 // Nodes without corresponding entries in h are given an initial heat of zero,
-// and entries in h without a corresponding node in g are not altered.
+// and entries in h without a corresponding node in the original graph are
+// not altered.
 func Diffuse(h map[int64]float64, by Laplacian, t float64) map[int64]float64 {
 	heat := make([]float64, len(by.Index))
 	for id, i := range by.Index {
@@ -50,7 +51,8 @@ func Diffuse(h map[int64]float64, by Laplacian, t float64) map[int64]float64 {
 // Indexing into h is defined by the Laplacian Index field.
 //
 // Nodes without corresponding entries in h are given an initial heat of zero,
-// and entries in h without a corresponding node in g are not altered.
+// and entries in h without a corresponding node in the original graph are
+// not altered.
 func DiffuseToEquilibrium(h map[int64]float64, by Laplacian, tol float64, iters int) (eq map[int64]float64, ok bool) {
 	heat := make([]float64, len(by.Index))
 	for id, i := range by.Index {
@@ -125,7 +127,7 @@ func NewLaplacian(g graph.Undirected) Laplacian {
 	return Laplacian{Matrix: l, Nodes: nodes, Index: indexOf}
 }
 
-// NewSymNormLaplacian returns a symmetric normalised Laplacian matrix for the
+// NewSymNormLaplacian returns a symmetric normalized Laplacian matrix for the
 // simple undirected graph g.
 // The Laplacian is defined by I-D^(-1/2)AD^(-1/2) where D is a diagonal matrix holding the
 // degree of each node and A is the graph adjacency matrix of the input graph.
--- a/graph/network/diffusion.go
+++ b/graph/network/diffusion.go
@@ -48,7 +48,7 @@ func Diffuse(dst, h map[int64]float64, by Laplacian, t float64) map[int64]float6
 // graph described by the given Laplacian using the initial heat
 // distribution, h, according to the Laplacian until the update function
 //  h_{n+1} = h_n - L×h_n
-// results in a 2-norm update difference within tol, or iter updates have
+// results in a 2-norm update difference within tol, or iters updates have
 // been made.
 // The resulting heat distribution is returned as eq, written into the map dst,
 // and a boolean indicating whether the equilibrium converged to within tol.
@@ -67,21 +67,21 @@ func DiffuseToEquilibrium(dst, h map[int64]float64, by Laplacian, tol float64, i
 
 	last := make([]float64, len(by.Index))
 	for id, i := range by.Index {
-		heat[i] = h[id]
+		last[i] = h[id]
 	}
 	lastV := mat.NewVecDense(len(last), last)
 
 	var tmp mat.VecDense
 	for {
+		iters--
+		if iters < 0 {
+			break
+		}
 		lastV, v = v, lastV
 		tmp.MulVec(by.Matrix, lastV)
 		v.SubVec(lastV, &tmp)
 		if normDiff(heat, last) < tol {
 			ok = true
-			break
-		}
-		iters--
-		if iters < 0 {
 			break
 		}
 	}
--- a/graph/network/diffusion.go
+++ b/graph/network/diffusion.go
@@ -38,7 +38,7 @@ func Diffuse(dst, h map[int64]float64, by Laplacian, t float64) map[int64]float6
 	if dst == nil {
 		dst = make(map[int64]float64)
 	}
-	for i, n := range v.RawVector().Data {
+	for i, n := range heat {
 		dst[by.Nodes[i].ID()] = n
 	}
 	return dst
@@ -89,7 +89,7 @@ func DiffuseToEquilibrium(dst, h map[int64]float64, by Laplacian, tol float64, i
 	if dst == nil {
 		dst = make(map[int64]float64)
 	}
-	for i, n := range v.RawVector().Data {
+	for i, n := range heat {
 		dst[by.Nodes[i].ID()] = n
 	}
 	return dst, ok
@@ -176,8 +176,8 @@ func NewSymNormLaplacian(g graph.Undirected) Laplacian {
 	return Laplacian{Matrix: l, Nodes: nodes, Index: indexOf}
 }
 
-// NewRandomWalkLaplacian returns a random walk Laplacian matrix for the
-// simple graph g.
+// NewRandomWalkLaplacian returns a damp-scaled random walk Laplacian matrix for
+// the simple graph g.
 // The random walk Laplacian is defined as I-D^(-1)A where D is a diagonal matrix
 // holding the degree of each node and A is the graph adjacency matrix of the input
 // graph.
--- a/graph/network/diffusion.go
+++ b/graph/network/diffusion.go
@@ -12,8 +12,8 @@ import (
 )
 
 // Diffuse performs a heat diffusion across nodes of the undirected
-// graph g using the initial heat distribution, h, according to the
-// given Laplacian and diffusing for time t.
+// graph described by the given Laplacian using the initial heat distribution,
+// h, according to the Laplacian and diffusing for time t.
 // The resulting heat distribution is returned, written into the map h,
 //  d = exp(-Lt)×h
 // where L is the graph Laplacian. Indexing into h is defined by the
@@ -37,6 +37,52 @@ func Diffuse(h map[int64]float64, by Laplacian, t float64) map[int64]float64 {
 		h[by.Nodes[i].ID()] = n
 	}
 	return h
+}
+
+// DiffuseToEquilibrium performs a heat diffusion across nodes of the
+// graph described by the given Laplacian using the initial heat
+// distribution, h, according to the Laplacian until the update function
+//  h_{n+1} = h_n - L×h_n
+// results in a 2-norm update difference within tol, or iter updates have
+// been made.
+// The resulting heat distribution is returned as eq, written into the map h,
+// and a boolean indicating whether the equilibrium converged.
+// Indexing into h is defined by the Laplacian Index field.
+//
+// Nodes without corresponding entries in h are given an initial heat of zero,
+// and entries in h without a corresponding node in g are not altered.
+func DiffuseToEquilibrium(h map[int64]float64, by Laplacian, tol float64, iters int) (eq map[int64]float64, ok bool) {
+	heat := make([]float64, len(by.Index))
+	for id, i := range by.Index {
+		heat[i] = h[id]
+	}
+	v := mat.NewVecDense(len(heat), heat)
+
+	last := make([]float64, len(by.Index))
+	for id, i := range by.Index {
+		heat[i] = h[id]
+	}
+	lastV := mat.NewVecDense(len(last), last)
+
+	var tmp mat.VecDense
+	for {
+		lastV, v = v, lastV
+		tmp.MulVec(by.Matrix, lastV)
+		v.SubVec(lastV, &tmp)
+		if normDiff(heat, last) < tol {
+			ok = true
+			break
+		}
+		iters--
+		if iters < 0 {
+			break
+		}
+	}
+
+	for i, n := range v.RawVector().Data {
+		h[by.Nodes[i].ID()] = n
+	}
+	return h, ok
 }
 
 // Laplacian is a graph Laplacian matrix.
@@ -94,9 +140,10 @@ func NewSymNormLaplacian(g graph.Undirected) Laplacian {
 	l := mat.NewSymDense(len(nodes), nil)
 	for j, u := range nodes {
 		to := g.From(u)
-		if len(to) != 0 {
-			l.SetSym(j, j, 1)
+		if len(to) == 0 {
+			continue
 		}
+		l.SetSym(j, j, 1)
 		uid := u.ID()
 		udeg := math.Sqrt(float64(len(to)))
 		for _, v := range to {
@@ -104,6 +151,34 @@ func NewSymNormLaplacian(g graph.Undirected) Laplacian {
 			if uid < vid {
 				l.SetSym(indexOf[vid], j, -1/(udeg*math.Sqrt(float64(len(g.From(v))))))
 			}
+		}
+	}
+
+	return Laplacian{Matrix: l, Nodes: nodes, Index: indexOf}
+}
+
+// NewRandomWalkLaplacian returns a random walk Laplacian matrix for the
+// simple graph g.
+// The Laplacian is defined by I-D^(-1)A where D is a diagonal matrix holding the
+// degree of each node and A is the graph adjacency matrix of the input graph.
+func NewRandomWalkLaplacian(g graph.Graph, damp float64) Laplacian {
+	nodes := g.Nodes()
+	indexOf := make(map[int64]int, len(nodes))
+	for i, n := range nodes {
+		id := n.ID()
+		indexOf[id] = i
+	}
+
+	l := mat.NewDense(len(nodes), len(nodes), nil)
+	for j, u := range nodes {
+		to := g.From(u)
+		if len(to) == 0 {
+			continue
+		}
+		l.Set(j, j, 1-damp)
+		rudeg := (damp - 1) / float64(len(to))
+		for _, v := range to {
+			l.Set(indexOf[v.ID()], j, rudeg)
 		}
 	}
 
--- a/graph/network/diffusion.go
+++ b/graph/network/diffusion.go
@@ -1,4 +1,4 @@
-// Copyright ©2017 The gonum Authors. All rights reserved.
+// Copyright ©2017 The Gonum Authors. All rights reserved.
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
@@ -13,10 +13,9 @@ import (
 	"gonum.org/v1/gonum/mat"
 )
 
-// HeatDiffusion performs a heat diffusion network propagation across nodes
-// of the undirected graph g using the initial heat query h provided, and
-// diffusing for time t. The resulting heat diffusion is returned, written
-// into the map h,
+// HeatDiffusion performs a heat diffusion across nodes of the undirected
+// graph g using the initial heat distribution, h, and diffusing for time t.
+// The resulting heat distribution is returned, written into the map h,
 //  d = exp(-Lt)×h
 // where L is the graph Laplacian or symmetric normalized Laplacian.
 //
--- a/graph/network/diffusion.go
+++ b/graph/network/diffusion.go
@@ -6,75 +6,106 @@ package network
 
 import (
 	"math"
-	"sort"
 
 	"gonum.org/v1/gonum/graph"
-	"gonum.org/v1/gonum/graph/internal/ordered"
 	"gonum.org/v1/gonum/mat"
 )
 
-// HeatDiffusion performs a heat diffusion across nodes of the undirected
-// graph g using the initial heat distribution, h, and diffusing for time t.
+// Diffuse performs a heat diffusion across nodes of the undirected
+// graph g using the initial heat distribution, h, according to the
+// given Laplacian and diffusing for time t.
 // The resulting heat distribution is returned, written into the map h,
 //  d = exp(-Lt)×h
-// where L is the graph Laplacian or symmetric normalized Laplacian.
-//
-// The Laplacian is defined by D-A where D is a diagonal matrix holding the
-// degree of each node and A is the graph adjacency matrix of the input graph.
-// The symmetric normalized Laplacian is defined by I-D^(-1/2)AD^(-1/2).
+// where L is the graph Laplacian. Indexing into h is defined by the
+// Laplacian Index field.
 //
 // Nodes without corresponding entries in h are given an initial heat of zero,
 // and entries in h without a corresponding node in g are not altered.
-func HeatDiffusion(g graph.Undirected, h map[int64]float64, t float64, normalize bool) map[int64]float64 {
-	// See https://doi.org/10.1371/journal.pcbi.1005598 for details.
-
-	nodes := g.Nodes()
-	sort.Sort(ordered.ByID(nodes))
-	indexOf := make(map[int64]int, len(nodes))
-	heat := make([]float64, len(nodes))
-	for i, n := range nodes {
-		id := n.ID()
-		indexOf[id] = i
+func Diffuse(h map[int64]float64, by Laplacian, t float64) map[int64]float64 {
+	heat := make([]float64, len(by.Index))
+	for id, i := range by.Index {
 		heat[i] = h[id]
 	}
 	v := mat.NewVecDense(len(heat), heat)
 
-	l := mat.NewSymDense(len(nodes), nil)
-	if normalize {
-		for j, u := range nodes {
-			to := g.From(u)
-			if len(to) != 0 {
-				l.SetSym(j, j, -t)
-			}
-			uid := u.ID()
-			udeg := math.Sqrt(float64(len(to)))
-			for _, v := range to {
-				vid := v.ID()
-				if uid < vid {
-					l.SetSym(indexOf[vid], j, t/(udeg*math.Sqrt(float64(len(g.From(v))))))
-				}
-			}
-		}
-	} else {
-		for j, u := range nodes {
-			to := g.From(u)
-			l.SetSym(j, j, -t*float64(len(to)))
-			uid := u.ID()
-			for _, v := range to {
-				vid := v.ID()
-				if uid < vid {
-					l.SetSym(indexOf[vid], j, t)
-				}
-			}
-		}
-	}
-
-	var m mat.Dense
-	m.Exp(l)
+	var m, tl mat.Dense
+	tl.Scale(-t, by)
+	m.Exp(&tl)
 	v.MulVec(&m, v)
 
 	for i, n := range v.RawVector().Data {
-		h[nodes[i].ID()] = n
+		h[by.Nodes[i].ID()] = n
 	}
 	return h
+}
+
+// Laplacian is a graph Laplacian matrix.
+type Laplacian struct {
+	// Matrix holds the Laplacian matrix.
+	mat.Matrix
+
+	// Nodes holds the input graph nodes.
+	Nodes []graph.Node
+
+	// Index is a mapping from the graph
+	// node IDs to row and column indices.
+	Index map[int64]int
+}
+
+// NewLaplacian returns a Laplacian matrix for the simple undirected graph g.
+// The Laplacian is defined by D-A where D is a diagonal matrix holding the
+// degree of each node and A is the graph adjacency matrix of the input graph.
+func NewLaplacian(g graph.Undirected) Laplacian {
+	nodes := g.Nodes()
+	indexOf := make(map[int64]int, len(nodes))
+	for i, n := range nodes {
+		id := n.ID()
+		indexOf[id] = i
+	}
+
+	l := mat.NewSymDense(len(nodes), nil)
+	for j, u := range nodes {
+		to := g.From(u)
+		l.SetSym(j, j, float64(len(to)))
+		uid := u.ID()
+		for _, v := range to {
+			vid := v.ID()
+			if uid < vid {
+				l.SetSym(indexOf[vid], j, -1)
+			}
+		}
+	}
+
+	return Laplacian{Matrix: l, Nodes: nodes, Index: indexOf}
+}
+
+// NewSymNormLaplacian returns a symmetric normalised Laplacian matrix for the
+// simple undirected graph g.
+// The Laplacian is defined by I-D^(-1/2)AD^(-1/2) where D is a diagonal matrix holding the
+// degree of each node and A is the graph adjacency matrix of the input graph.
+func NewSymNormLaplacian(g graph.Undirected) Laplacian {
+	nodes := g.Nodes()
+	indexOf := make(map[int64]int, len(nodes))
+	for i, n := range nodes {
+		id := n.ID()
+		indexOf[id] = i
+	}
+
+	l := mat.NewSymDense(len(nodes), nil)
+	for j, u := range nodes {
+		to := g.From(u)
+		if len(to) != 0 {
+			l.SetSym(j, j, 1)
+		}
+		uid := u.ID()
+		udeg := math.Sqrt(float64(len(to)))
+		for _, v := range to {
+			vid := v.ID()
+			if uid < vid {
+				l.SetSym(indexOf[vid], j, -1/(udeg*math.Sqrt(float64(len(g.From(v))))))
+			}
+		}
+	}
+
+	return Laplacian{Matrix: l, Nodes: nodes, Index: indexOf}
 }
--- a/graph/network/diffusion.go
+++ b/graph/network/diffusion.go
@@ -89,7 +89,7 @@ func DiffuseToEquilibrium(dst, h map[int64]float64, by Laplacian, tol float64, i
 	if dst == nil {
 		dst = make(map[int64]float64)
 	}
-	for i, n := range heat {
+	for i, n := range v.RawVector().Data {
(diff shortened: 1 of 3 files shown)


kortschak:
The python implementation gives me values that agree with the test expectations with increasing divergence as t increases (this is how I found the Exp behaviour), e.g. for the t=0.1 case:
```
{(0, 0): 0.82668405458647332,
 (0, 1): 0.07854806046143345,
 (0, 2): 0.0038588395017377115,
 (0, 3): 0.00012748739572128364,
 (0, 4): 3.2332177780049904e-06,
 (1, 0): 0.07854806046143345,
 (1, 1): 0.0074633081018349642,
 (1, 2): 0.00036665078612781686,
 (1, 3): 1.2113319002137123e-05,
 (1, 4): 3.0720682720649926e-07,
 (2, 0): 0.0038588395017377115,
 (2, 1): 0.00036665078612781686,
 (2, 2): 1.8012494879461547e-05,
 (2, 3): 5.9509240060162579e-07,
 (2, 4): 1.5092184747326786e-08,
 (3, 0): 0.00012748739572128364,
 (3, 1): 1.2113319002137123e-05,
 (3, 2): 5.9509240060162579e-07,
 (3, 3): 1.9660517192302685e-08,
 (3, 4): 4.986119086589226e-10,
 (4, 0): 3.2332177780049904e-06,
 (4, 1): 3.0720682720649926e-07,
 (4, 2): 1.5092184747326786e-08,
 (4, 3): 4.986119086589226e-10,
 (4, 4): 1.2645335472922438e-11}
```

graph/network/diffusion.go
```diff
@@ -0,0 +1,76 @@
...
+// license that can be found in the LICENSE file.
+
+package network
+
+import (
+	"gonum.org/v1/gonum/graph"
+	"gonum.org/v1/gonum/mat"
+)
+
+// HeatDiffucion performs a heat diffusion network propagation across nodes
```
sbinet: `s/HeatDiffucion/HeatDiffusion/`

kortschak: Fixed

kortschak:
I need to check the normalised case - the python reproducer is at home, so I'll do that after work.

kortschak:
Symmetric normalized Laplacian implementation results also confirmed with the python implementation (within an error probably due to our exp).

kortschak:
A consideration here is whether it would be better to provide an exp function here that takes the Laplacian (this would need to be a struct holding the matrix and the look-up map) and then also provide two functions (at least one more later - the random walk Laplacian) to generate the Laplacian and Symmetric normalized Laplacian. I think this might be nicer.

graph/network/diffusion_test.go
```diff
@@ -0,0 +1,139 @@
...
+
+import (
+	"math"
+	"testing"
+
+	"gonum.org/v1/gonum/floats"
+	"gonum.org/v1/gonum/graph/simple"
+)
+
+var heatDiffusionTests = []struct {
```
vladimir-ch: I think it would be worth adding two test cases:
* Initial "temperature" `h` is constant. `HeatDiffusion` then should not change `h`.
* Initial "temperature" `h` is not constant but the average temperate should be constant independently of `t`.

graph/network/diffusion.go
```diff
@@ -0,0 +1,81 @@
+// Copyright ©2017 The gonum Authors. All rights reserved.
```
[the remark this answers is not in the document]
kortschak: Done (was waiting for the capitalisation branch to be merged).

graph/network/diffusion.go
```diff
@@ -0,0 +1,81 @@
...
+	"math"
+	"sort"
+
+	"gonum.org/v1/gonum/graph"
+	"gonum.org/v1/gonum/graph/internal/ordered"
+	"gonum.org/v1/gonum/mat"
+)
+
+// HeatDiffusion performs a heat diffusion network propagation across nodes
+// of the undirected graph g using the initial heat query h provided, and
```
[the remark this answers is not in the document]
kortschak: "Query" is the term used in the paper, but I agree with your suggestion.

graph/network/diffusion.go
```diff
@@ -0,0 +1,81 @@
...
+// diffusing for time t. The resulting heat diffusion is returned, written
```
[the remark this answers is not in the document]
kortschak: Done

graph/network/diffusion_test.go
kortschak: Done. They are not happy tests due to the instability of `mat.Exp`. Worse for the symmetric normalised Laplacian (so I avoid that completely for the global temperature).

kortschak:
>Perhaps `Diffuse(Laplacian(g), h, t)` or `Evolve(Laplacian(g), h, t)` might be indeed nicer.

Of these I'd prefer `Diffuse(h map[int64]float64, by Laplacian, t float64) map[int64]float64`. I'll wait for other input.

graph/network/diffusion.go
```diff
@@ -0,0 +1,81 @@
...
+import (
+	"math"
+	"sort"
+
+	"gonum.org/v1/gonum/graph"
+	"gonum.org/v1/gonum/graph/internal/ordered"
+	"gonum.org/v1/gonum/mat"
+)
+
+// HeatDiffusion performs a heat diffusion network propagation across nodes
```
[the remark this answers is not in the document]
kortschak: Done

kortschak:
@btracey @sbinet @mewmew Do you have views on the API?

btracey:
I can try to take a look tomorrow

sbinet:
I won't be able to give it a fair look before the weekend, sorry.

mewmew:
The API of `HeatDiffusion` looks good. The only thought I had was that the key of `h` is implicitly documented in the last sentence, rather than explicitly documented that it is node IDs. This is fine though, and perhaps preferable as the last sentence also covers corner cases where node IDs are present in `h` but not `g`, and vice versa.

Great work on this! Will be fun to model some epidemic spreads :) Perhaps implement a more advanced version of the board game [pandemic](https://boardgamegeek.com/boardgame/30549/pandemic).

kortschak:
@mewmew I'm wondering about relative merits of the current API and the alternatives where there are Hermitian generating functions.

kortschak:
To expand, there is at least a third Hermitian that I want to be able to diffuse, the random walk Hermitian, though this is usually done to equilibrium. The two API options give us either the function we have and a random walk diffusion, or alternatively two diffusion functions (one timed and one with a tolerance for termination) and three (or more) Hermitian functions (of these 6 products, 3-4 are useful).

vladimir-ch:
I think that having separate functions to create some of the possible variants of the graph Laplacian is most useful. A user could want to look at its eigenvalues, for example. Then there would be functions to apply a (Laplacian) operator, just as you suggest.

I'm delighted by this PR and by the possibilities it opens. On one hand it touches on PDEs (because it is related to the heat equation du/dt = D∆u), on other hand it is related to discrete geometry of 3D surfaces, both being areas of my (past) research interest. Random walks seem to be also connected but I'm not familiar with random processes. All in all, it's pleasing for me to see an interest and activity in these areas in Gonum.

kortschak:
>I'm delighted by this PR and by the possibilities it opens. On one hand it touches on PDEs (because it is related to the heat equation du/dt = D∆u), on other hand it is related to discrete geometry of 3D surfaces, both being areas of my (past) research interest. Random walks seem to be also connected but I'm not familiar with random processes. All in all, it's pleasing for me to see an interest and activity in these areas in Gonum.

Yeah, this addresses a research problem I have been thinking about for a few years; I bumped into the paper and realised it comes close to answering the problem I had (the random walk L is closer).

kortschak:
I have revised the API. The `Laplacian` type exports its fields to allow users to write their own Laplacian construction functions - this is important because of the multiple ways a directed graph's Laplacian can be defined. I think it's best to just provide the basic undirected case and let others modify the input to their liking.

PTAL

kortschak:
I've had a go at the random walk diffusion and I'm obviously missing something. The algorithm I implemented is a loop of `v <- Lv` while the step's vector norm is above a tolerance. This won't work for the standard Laplacian since the diagonals are non-unit, but it should work for the random walk. I'll leave this for a while. @vladimir-ch if you find some time to look into that, that would be very helpful.

vladimir-ch:
I think that in the iteration `v <- Lv` the matrix `L` is the transition matrix `T := D^{-1} A` and not `I-D^{-1} A` which is what `NewRandomWalkLaplacian` currently returns. I'm not versed in random walks on graphs but you could try that.

graph/network/diffusion.go
```diff
@@ -0,0 +1,138 @@
...
+		indexOf[id] = i
+	}
+
+	l := mat.NewDense(len(nodes), len(nodes), nil)
+	for j, u := range nodes {
+		to := g.From(u)
+		if len(to) != 0 {
+			l.Set(j, j, 1-damp)
+		}
+		rudeg := (damp - 1) / float64(len(to))
```
vladimir-ch: `len(to)` can be zero, are you ok with dividing by zero even if `rudeg` is not used later?

graph/network/diffusion_test.go
```diff
@@ -0,0 +1,307 @@
...
+	"testing"
+
+	"gonum.org/v1/gonum/floats"
+	"gonum.org/v1/gonum/graph"
+	"gonum.org/v1/gonum/graph/internal/ordered"
+	"gonum.org/v1/gonum/graph/simple"
+	"gonum.org/v1/gonum/mat"
+)
+
+var heatDiffusionTests = []struct {
```
vladimir-ch: More test cases that came to my mind after some reading that you might want to add if it does not mean too much work:
* graph with isolated nodes
* graph with several components. In this case each component will have its own average.

kortschak:
>I think that in the iteration `v <- Lv` the matrix `L` is the transition matrix `T := D^{-1} A` and not `I-D^{-1} A` which is what `NewRandomWalkLaplacian` currently returns. I'm not versed in random walks on graphs but you could try that.


Yeah, I mis-wrote in the comment, I was actually doing `v <- v - Lv`. I think the issue is fixed.

graph/network/diffusion.go
kortschak: Nice, I'm fine with divinding by zero in floats, but an inverted condition above with a continue is a much better approach.

Fixed.

graph/network/diffusion_test.go
kortschak: Yeah, those are the same class of test. I'll see what I can do to expand the test logic to handle those.

kortschak: I've gone the easy route and just added graphs of that form and tested the node values explicitly. The graphs are small enough that they are Obviously Correct™.

kortschak:
This is ready. PTAL

kortschak:
*ping*

graph/network/diffusion_test.go
```diff
@@ -0,0 +1,499 @@
...
+		damp: 0.85,
+
+		want: mat.NewDense(3, 3, []float64{
+			0.15, 0, -0.15,
+			-0.15, 0.15, 0,
+			0, -0.15, 0.15,
+		}),
+	},
+	{
+		// Example graph from http://en.wikipedia.org/wiki/File:PageRanks-Example.svg 16:17, 8 July 2009
```
sbinet: `s/http:/https:/`

graph/network/diffusion_test.go
```diff
@@ -0,0 +1,499 @@
...
+			A: 0.025000, B: 0.037500, C: 0.037500, D: 0.037500, E: 0.025000,
+			F: 0.037500, G: 0.050000, H: 0.050000, I: 0.050000, J: 0.037500,
+			K: 0.037500, L: 0.050000, M: 0.050000, N: 0.050000, O: 0.037500,
+			P: 0.037500, Q: 0.050000, R: 0.050000, S: 0.050000, T: 0.037500,
+			U: 0.025000, V: 0.037500, W: 0.037500, X: 0.037500, Y: 0.025000,
+		},
+		wantOK: true,
+	},
+	{
+		// Example graph from http://en.wikipedia.org/wiki/File:PageRanks-Example.svg 16:17, 8 July 2009
```
sbinet: `s/http:/https:/`

graph/network/diffusion.go
```diff
@@ -0,0 +1,186 @@
...
+import (
+	"math"
+
+	"gonum.org/v1/gonum/graph"
+	"gonum.org/v1/gonum/mat"
+)
+
+// Diffuse performs a heat diffusion across nodes of the undirected
+// graph described by the given Laplacian using the initial heat distribution,
+// h, according to the Laplacian and diffusing for time t.
```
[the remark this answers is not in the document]
kortschak: A comma separating the description from the name is correct here.

graph/network/diffusion.go
```diff
@@ -0,0 +1,186 @@
...
+// The resulting heat distribution is returned, written into the map h,
+//  d = exp(-Lt)×h
+// where L is the graph Laplacian. Indexing into h is defined by the
+// Laplacian Index field.
+//
+// Nodes without corresponding entries in h are given an initial heat of zero,
+// and entries in h without a corresponding node in g are not altered.
```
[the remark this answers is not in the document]
kortschak: Fixed: "in the original graph".

graph/network/diffusion.go
```diff
@@ -0,0 +1,186 @@
...
+// distribution, h, according to the Laplacian until the update function
+//  h_{n+1} = h_n - L×h_n
+// results in a 2-norm update difference within tol, or iter updates have
+// been made.
+// The resulting heat distribution is returned as eq, written into the map h,
+// and a boolean indicating whether the equilibrium converged.
+// Indexing into h is defined by the Laplacian Index field.
+//
+// Nodes without corresponding entries in h are given an initial heat of zero,
+// and entries in h without a corresponding node in g are not altered.
```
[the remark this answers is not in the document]
kortschak: The use of `by` is intentional; the signature is `Diffuse(h map[int64]float64, by Laplacian, t float64) map[int64]float64` which translates almost directly to English as "diffuse h by [the] Laplacian [for time] t".

graph/network/diffusion.go
```diff
@@ -0,0 +1,186 @@
...
+func DiffuseToEquilibrium(h map[int64]float64, by Laplacian, tol float64, iters int) (eq map[int64]float64, ok bool) {
```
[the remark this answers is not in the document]
kortschak: I don't think it's a good idea to open that can of worm. It is entirely possible to construct a `Laplacian` that does not converge even after infinite time.

graph/network/diffusion.go
```diff
@@ -0,0 +1,186 @@
...
+			if uid < vid {
+				l.SetSym(indexOf[vid], j, -1)
+			}
+		}
+	}
+
+	return Laplacian{Matrix: l, Nodes: nodes, Index: indexOf}
+}
+
+// NewSymNormLaplacian returns a symmetric normalised Laplacian matrix for the
```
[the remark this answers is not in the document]
kortschak: Fixed (they sneak in without me noticing).

graph/network/diffusion_test.go
kortschak: At the time is was http:

graph/network/diffusion_test.go
kortschak: Same.

graph/network/diffusion.go
```diff
@@ -0,0 +1,188 @@
...
+	l := mat.NewSymDense(len(nodes), nil)
+	for j, u := range nodes {
+		to := g.From(u)
+		if len(to) == 0 {
+			continue
+		}
+		l.SetSym(j, j, 1)
+		uid := u.ID()
+		udeg := math.Sqrt(float64(len(to)))
+		for _, v := range to {
```
btracey: Just checking there are no self-loop issues here with overwriting the `Set`

graph/network/diffusion.go
```diff
@@ -0,0 +1,188 @@
...
+
+	l := mat.NewSymDense(len(nodes), nil)
+	for j, u := range nodes {
+		to := g.From(u)
+		if len(to) == 0 {
+			continue
+		}
+		l.SetSym(j, j, 1)
+		uid := u.ID()
+		udeg := math.Sqrt(float64(len(to)))
```
btracey: Can you rename this to `squdeg` or something, so it's clear this is the square root of the degree and not the degree?

graph/network/diffusion.go
```diff
@@ -0,0 +1,188 @@
...
+			if uid < vid {
+				l.SetSym(indexOf[vid], j, -1/(udeg*math.Sqrt(float64(len(g.From(v))))))
+			}
+		}
+	}
+
+	return Laplacian{Matrix: l, Nodes: nodes, Index: indexOf}
+}
+
+// NewRandomWalkLaplacian returns a random walk Laplacian matrix for the
```
btracey: There is no description of `damp` in the documentation.

graph/network/diffusion.go
```diff
@@ -0,0 +1,188 @@
...
+}
+
+// DiffuseToEquilibrium performs a heat diffusion across nodes of the
+// graph described by the given Laplacian using the initial heat
+// distribution, h, according to the Laplacian until the update function
+//  h_{n+1} = h_n - L×h_n
+// results in a 2-norm update difference within tol, or iter updates have
+// been made.
+// The resulting heat distribution is returned as eq, written into the map h,
+// and a boolean indicating whether the equilibrium converged.
```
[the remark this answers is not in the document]
kortschak: Done: "to within tol."

graph/network/diffusion.go
```diff
@@ -0,0 +1,188 @@
...
+// Indexing into h is defined by the Laplacian Index field.
+//
+// Nodes without corresponding entries in h are given an initial heat of zero,
+// and entries in h without a corresponding node in the original graph are
+// not altered.
+func DiffuseToEquilibrium(h map[int64]float64, by Laplacian, tol float64, iters int) (eq map[int64]float64, ok bool) {
```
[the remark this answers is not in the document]
kortschak: Don't know.

graph/network/diffusion.go
```diff
@@ -0,0 +1,188 @@
...
+			}
+		}
+	}
+
+	return Laplacian{Matrix: l, Nodes: nodes, Index: indexOf}
+}
+
+// NewSymNormLaplacian returns a symmetric normalized Laplacian matrix for the
+// simple undirected graph g.
+// The Laplacian is defined by I-D^(-1/2)AD^(-1/2) where D is a diagonal matrix holding the
```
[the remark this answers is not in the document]
kortschak: Done, and for random walk below.

graph/network/diffusion.go
kortschak: Done.

graph/network/diffusion.go
kortschak: The Laplacians cannot be constructed from graphs with self loops.

btracey: Should there be a check rather than just failing silently?  There's no comment that `Undirected` cannot have self loops.

kortschak: Yeah, we used to refuse self-loops in all (simple) graphs. The three approaches to take are 1) add an `error` return, 2) panic or 3) just document that Laplacians are defined on adjacency matrices with zero diagonals. I don't know.

graph/network/diffusion.go
kortschak: I had intended to put it in, but forgot. I needed to work out how to add it into the equation. I will do that soon.

graph/network/diffusion.go
```diff
@@ -0,0 +1,188 @@
...
+// h, according to the Laplacian and diffusing for time t.
+// The resulting heat distribution is returned, written into the map h,
+//  d = exp(-Lt)×h
+// where L is the graph Laplacian. Indexing into h is defined by the
+// Laplacian Index field.
+//
+// Nodes without corresponding entries in h are given an initial heat of zero,
+// and entries in h without a corresponding node in the original graph are
+// not altered.
+func Diffuse(h map[int64]float64, by Laplacian, t float64) map[int64]float64 {
```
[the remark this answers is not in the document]
kortschak: Done.

graph/network/diffusion.go
```diff
@@ -0,0 +1,188 @@
...
+	return h
+}
+
+// DiffuseToEquilibrium performs a heat diffusion across nodes of the
+// graph described by the given Laplacian using the initial heat
+// distribution, h, according to the Laplacian until the update function
+//  h_{n+1} = h_n - L×h_n
+// results in a 2-norm update difference within tol, or iter updates have
+// been made.
+// The resulting heat distribution is returned as eq, written into the map h,
```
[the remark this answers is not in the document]
kortschak: Done.

graph/network/diffusion.go
btracey: I think panic is fine.

kortschak: Done.

graph/network/diffusion.go
kortschak: Done.

kortschak:
PTAL

graph/network/diffusion.go
```diff
@@ -0,0 +1,212 @@
...
+	for i, n := range v.RawVector().Data {
+		dst[by.Nodes[i].ID()] = n
+	}
+	return dst, ok
+}
+
+// Laplacian is a graph Laplacian matrix.
+type Laplacian struct {
+	// Matrix holds the Laplacian matrix.
+	mat.Matrix
```
vladimir-ch: A free flow of thoughts: if we consider only `NewLaplacian` and `NewSymNormLaplacian`, they take `graph.Undirected`, so both Laplacians are actually symmetric. There could be another type, say `TransitionMatrix`, that would be used for the random walks by operating on probability vectors while the Laplacian would be used to diffuse general scalar fields on a graph. Then the interface here could be `mat.Symmetric`. This would also give an important message about its eigenvalues which are an important point in all this. By having TransitionMatrix we would offer another matrix describing the graph (even though obtainable from the random walk Laplacian), and also the iterative update would be simpler. TransitionMatrix is not symmetric and could be generated also for directed graphs (as is done now). The damping would then not be specified in `NewTransitionMatrix` but in `RandomWalk(TransitionMatrix, damp float64)`. Finally, but not so importantly, the types could live in `graph`, so to the outside they would like as `graph.Laplacian`, `graph.TransitionMatrix`.

graph/network/diffusion.go
```diff
@@ -0,0 +1,212 @@
...
+// been made.
+// The resulting heat distribution is returned as eq, written into the map dst,
+// and a boolean indicating whether the equilibrium converged to within tol.
+// Indexing into h and dst is defined by the Laplacian Index field. If dst
+// is nil, a new map is created.
+//
+// Nodes without corresponding entries in h are given an initial heat of zero,
+// and entries in h without a corresponding node in the original graph are
+// not altered when written to dst.
+func DiffuseToEquilibrium(dst, h map[int64]float64, by Laplacian, tol float64, iters int) (eq map[int64]float64, ok bool) {
```
vladimir-ch: > There isn't a more efficient way of computing the equilibrium? I thought it was one of the eigenvalues of the Laplacian?

If we consider the transition matrix T and do the iteration `p_{i+1} = T p_i`, then *if* `p_i` converges to `p`, we get `p = Tp`. This means that `p` is an eigenvector corresponding to the eigenvalue 1 and this algorithm is a way of finding it. It should converge quite fast.

The problem is in the "if" above: it may not converge. 1 is always an eigenvalue of T. If all other eigenvalues are smaller than 1 in absolute value and the graph is connected, 1 has multiplicity 1, and this converges for any initial vector. If there are more components, the multiplicity of 1 will be equal to the number of components and the limit may depend on the starting vector. If -1 is an eigenvalue, it may not converge.

Random walk with the transition matrix is useful for finding the eigenvector corrsponding to 1 (which is useful as a centrality measure) without finding the complete spectrum of the matrix, so I think this should stay.

graph/network/diffusion.go
```diff
@@ -0,0 +1,212 @@
...
+		dst[by.Nodes[i].ID()] = n
+	}
+	return dst
+}
+
+// DiffuseToEquilibrium performs a heat diffusion across nodes of the
+// graph described by the given Laplacian using the initial heat
+// distribution, h, according to the Laplacian until the update function
+//  h_{n+1} = h_n - L×h_n
+// results in a 2-norm update difference within tol, or iter updates have
```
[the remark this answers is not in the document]
kortschak: Done

graph/network/diffusion.go
```diff
@@ -0,0 +1,212 @@
...
+func DiffuseToEquilibrium(dst, h map[int64]float64, by Laplacian, tol float64, iters int) (eq map[int64]float64, ok bool) {
+	heat := make([]float64, len(by.Index))
+	for id, i := range by.Index {
+		heat[i] = h[id]
+	}
+	v := mat.NewVecDense(len(heat), heat)
+
+	last := make([]float64, len(by.Index))
+	for id, i := range by.Index {
+		heat[i] = h[id]
```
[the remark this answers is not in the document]
kortschak: Fixed

graph/network/diffusion.go
```diff
@@ -0,0 +1,212 @@
...
+	for {
+		lastV, v = v, lastV
+		tmp.MulVec(by.Matrix, lastV)
+		v.SubVec(lastV, &tmp)
+		if normDiff(heat, last) < tol {
+			ok = true
+			break
+		}
+		iters--
+		if iters < 0 {
```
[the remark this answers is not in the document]
kortschak: Moved to top of loop.

graph/network/diffusion.go
```diff
@@ -0,0 +1,212 @@
...
+			if uid < vid {
+				l.SetSym(indexOf[vid], j, -1)
+			}
+		}
+	}
+
+	return Laplacian{Matrix: l, Nodes: nodes, Index: indexOf}
+}
+
+// NewSymNormLaplacian returns a symmetric normalized Laplacian matrix for the
```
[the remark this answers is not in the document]
kortschak: I'm following the naming convention in the WP article. I suspect that the symmetricality (?) is something related to the Laplacian dynamics rather than the structure.

graph/network/diffusion.go
kortschak: Thanks

graph/network/diffusion.go
kortschak: I need to think about this. I know that L^{rw} is an abuse of the term Laplacian, but it's in use. Note that L^{rw} for undirected graphs is still not symmetric.

graph/network/diffusion.go
btracey: Okay, thanks!

graph/network/diffusion.go
```diff
@@ -0,0 +1,212 @@
...
+
+	var m, tl mat.Dense
+	tl.Scale(-t, by)
+	m.Exp(&tl)
+	v.MulVec(&m, v)
+
+	if dst == nil {
+		dst = make(map[int64]float64)
+	}
+	for i, n := range v.RawVector().Data {
```
btracey: You don't need to use this `RawVector` call here, the backing data is `heat`

graph/network/diffusion.go
```diff
@@ -0,0 +1,212 @@
...
+		if normDiff(heat, last) < tol {
+			ok = true
+			break
+		}
+	}
+
+	if dst == nil {
+		dst = make(map[int64]float64)
+	}
+	for i, n := range v.RawVector().Data {
```
btracey: Same here, just use `heat`

graph/network/diffusion.go
btracey: I don't see a description of `damp`?

graph/network/diffusion_test.go
```diff
@@ -0,0 +1,495 @@
...
+			H: 1. / 11.,
+			I: 1. / 11.,
+			J: 1. / 11.,
+			K: 1. / 11.,
+		},
+		damp: 0.85,
+		tol:  1e-6,
+		iter: 1e4,
+
+		// This does not look like Page Rank because we do not
```
btracey: Embedding homework problems in our source code? :-)

btracey:
The code mostly looks good to me, but I'll leave it to you and @vladimir-ch to decide the best locations. My two cents would be that it feels like `Laplacian` is generic enough to be at the top level, but I don't have strong opinions

btracey:
I've heard talks using these subjects, but I haven't had to implement or use them directly before, so I don't have the benefit of experience

graph/network/diffusion.go
kortschak: Done

graph/network/diffusion.go
kortschak: Done

graph/network/diffusion_test.go
kortschak: Yes.

graph/network/diffusion.go
kortschak: I wrote it and forgot to commit it.

Done

kortschak:
I'm going to leave this for a bit, the comment from @vladimir-ch https://github.com/gonum/gonum/pull/292#discussion_r151743905 looks like it's a valuable idea, but I'm unclear of the details (I'd be happy to have `Laplacian` in graph. When you have time, can you sketch it out a bit more and I'll look into it.

graph/network/diffusion_test.go
```diff
@@ -0,0 +1,495 @@
...
+		h:       map[int64]float64{A: 1, E: -10},
+		damp:    0.85,
+		tol:     1e-6,
+		iter:    1e4,
+
+		want: map[int64]float64{
+			A: 0.25, B: 0.375, C: 0.25, D: 0.125, E: -5, F: -5,
+		},
+		wantOK: true,
+	},
```
vladimir-ch: An additional test with `wantOK: false` could be:
```
g: []set{A: linksTo(B), B: linksTo(C)},
h: map[int64]float64{B: 1},
```
It should oscillate between `{0, 1, 0}` and `{0.5, 0, 0.5}`.

vladimir-ch:
With that comment I meant something like this:
```
package graph

type Laplacian struct {
	mat.Symmetric // or mat.Matrix
}

func NewLaplacian(g Undirected) Laplacian {
}

func NewSymNormLaplacian(g Undirected) Laplacian {
}

type TransitionMatrix struct {
	mat.Matrix // All columns sum to 1.
}

func NewTransitionMatrix(g graph.Graph) TransitionMatrix {
}


package network

// h is a general scalar field on the graph defined by the Laplacian.
func Diffuse(dst, h map[int64]float64, by graph.Laplacian, t float64) map[int64]float64 {
}

// h is a probability vector.
func RandomWalk(dst, h map[int64]float64, with graph.TransitionMatrix, tol float64, iters int) (eq map[int64]float64, ok bool) {
	for {
		// ...
		v, lastV = lastV, v
		v.MulVec(by.Matrix, lastV)
		// ...
	}
}
```
With this the importing of `graph` would bring the whole of `mat` and therefore `lapack` which we may want to do only in the more specialized sub-package like `network`.

Also, the distinction between TransitionMatrix and Laplacian may not be that beneficial. The usual transition matrix T is similar to the normalized Laplacian, so they have the same eigenvalues. Since the Laplacian is symmetric, it has real eigenvalue and so does T although it is not symmetric.

In short, this PR is ok as it is, my thoughts are documented and we can return to them if necessary.

graph/network/diffusion_test.go
kortschak: I've added that test here since it shows that a fix to the vector data return loop needed to happen.

PTAL

graph/network/diffusion.go
kortschak: I've reverted this one. It's a subtle point, but the vector cycling is shallow, so `heat` may not be the same as value `v.RawVector().Data`. In the case of convergence this is not visible, but for non-converging cases, it is.

kortschak:
Thanks for the detail.

graph/network/diffusion_test.go
```diff
@@ -0,0 +1,495 @@
...
+		wantTol: 1e-2, // FIXME(kortschak): High tolerance used due to instability in mat.Exp.
+		want: map[bool]map[int64]float64{
+			false: {
+				A: 1, B: 1, C: 1, D: 1, E: 1,
+				F: 1, G: 1, H: 1, I: 1, J: 1,
+				K: 1, L: 1, M: 1, N: 1, O: 1,
+				P: 1, Q: 1, R: 1, S: 1, T: 1,
+				U: 1, V: 1, W: 1, X: 1, Y: 1,
+			},
+			true: {
```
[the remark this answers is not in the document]
kortschak: If I tighten the tolerance on this case I get the following:

```
--- FAIL: TestDiffuse (0.00s)
	diffusion_test.go:171: unexpected Diffuse result for test 3 with normalize=true:
		got: [A:0.9981753 B:1.0002977 C:0.9995572 D:1.0002977 E:0.9981753 F:1.0002977 G:1.0007715 H:1.0003861 I:1.0007715 J:1.0002977 K:0.9995572 L:1.0003861 M:1.0000019 N:1.0003861 O:0.9995572 P:1.0002977 Q:1.0007715 R:1.0003861 S:1.0007715 T:1.0002977 U:0.9981753 V:1.0002977 W:0.9995572 X:1.0002977 Y:0.9981753]
		want:[A:1.0000000 B:1.0000000 C:1.0000000 D:1.0000000 E:1.0000000 F:1.0000000 G:1.0000000 H:1.0000000 I:1.0000000 J:1.0000000 K:1.0000000 L:1.0000000 M:1.0000000 N:1.0000000 O:1.0000000 P:1.0000000 Q:1.0000000 R:1.0000000 S:1.0000000 T:1.0000000 U:1.0000000 V:1.0000000 W:1.0000000 X:1.0000000 Y:1.0000000]
```

The non-normalised operation passes. I had assumed this was because of the error in `mat.Exp`, but maybe it's not. What do you propose?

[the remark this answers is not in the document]
kortschak: I have run the python code equivalent to this and the overall pattern is what we see for got, but error is still there.

```
{(0, 0): 0.98264450473308107,
 (0, 1): 1.002568278028513,
 (0, 2): 0.9958911385307706,
 (0, 3): 1.002568278028513,
 (0, 4): 0.98264450473308107,
 (1, 0): 1.002568278028513,
 (1, 1): 1.0075291695232433,
 (1, 2): 1.0038067383118021,
 (1, 3): 1.0075291695232433,
 (1, 4): 1.002568278028513,
 (2, 0): 0.9958911385307706,
 (2, 1): 1.0038067383118021,
 (2, 2): 1.0001850837547184,
 (2, 3): 1.0038067383118021,
 (2, 4): 0.9958911385307706,
 (3, 0): 1.002568278028513,
 (3, 1): 1.0075291695232433,
 (3, 2): 1.0038067383118021,
 (3, 3): 1.0075291695232433,
 (3, 4): 1.002568278028513,
 (4, 0): 0.98264450473308107,
 (4, 1): 1.002568278028513,
 (4, 2): 0.9958911385307706,
 (4, 3): 1.002568278028513,
 (4, 4): 0.98264450473308107}
```

PR #294, opened 2017-11-06
mathext: speed up Digamma for large negative arguments

person142:
Before the Digamma function could be quite slow for large negative
arguments because it had to do use the recurrence relation thousands
of times. This avoids that by instead using the reflection formula. It
also adds some explicit checks for special cases (Inf, NaN, poles).

For example, running the following code:

```go
package main

import (
	"fmt"
	"time"
	"gonum.org/v1/gonum/mathext"
)

func main () {
	start := time.Now()
	for i := 0; i < 10; i++ {
		mathext.Digamma(-1.1111111111e9)
	}
	fmt.Printf("%s elapsed\n", time.Since(start))
}
```
gives me `40.625090145s elapsed` on master and `8.615µs elapsed` on this branch.

### DIFF
--- a/mathext/digamma.go
+++ b/mathext/digamma.go
@@ -10,19 +10,35 @@ import (
 
 // Digamma returns the logorithmic derivative of the gamma function at x.
 //  ψ(x) = d/dx (Ln (Γ(x)).
-// Note that if x is a negative integer in [-7, 0] this function will return
-// negative Inf.
 func Digamma(x float64) float64 {
 	// This is adapted from
 	// http://web.science.mq.edu.au/~mjohnson/code/digamma.c
 	var result float64
-	for ; x < 7.0; x++ {
+	switch {
+	case math.IsNaN(x), math.IsInf(x, 1):
+		return x
+	case math.IsInf(x, -1):
+		return math.NaN()
+	case x == 0:
+		return math.Copysign(math.Inf(1), -x)
+	case x < 0:
+		if x == math.Floor(x) {
+			return math.NaN()
+		}
+		// Reflection formula, http://dlmf.nist.gov/5.5#E4
+		_, r := math.Modf(x)
+		result = -math.Pi / math.Tan(math.Pi * r)
+		x = 1 - x
+	}
+	for ; x < 7; x++ {
+		// Recurrence relation, http://dlmf.nist.gov/5.5#E2
 		result -= 1 / x
 	}
-	x -= 1.0 / 2.0
-	xx := 1.0 / x
+	x -= 0.5
+	xx := 1 / x
 	xx2 := xx * xx
 	xx4 := xx2 * xx2
-	result += math.Log(x) + (1./24.)*xx2 - (7.0/960.0)*xx4 + (31.0/8064.0)*xx4*xx2 - (127.0/30720.0)*xx4*xx4
+	// Asymptotic expansion, http://dlmf.nist.gov/5.11#E2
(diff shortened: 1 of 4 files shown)


mathext/digamma.go
```diff
@@ -10,19 +10,35 @@ import (
...
+		_, r := math.Modf(x)
+		result = -math.Pi / math.Tan(math.Pi * r)
+		x = 1.0 - x
+	}
 	for ; x < 7.0; x++ {
-		result -= 1 / x
+		// Recurrence relation, http://dlmf.nist.gov/5.5#E2
+		result -= 1.0 / x
 	}
 	x -= 1.0 / 2.0
```
kortschak: `x -= 0.5`

mathext/digamma.go
```diff
@@ -10,19 +10,35 @@ import (
...
 	xx := 1.0 / x
```
kortschak: s/1.0/1/

mathext/digamma.go
```diff
@@ -10,19 +10,35 @@ import (
...
 	xx2 := xx * xx
 	xx4 := xx2 * xx2
+	// Asymptotic expansion, http://dlmf.nist.gov/5.11#E2
 	result += math.Log(x) + (1./24.)*xx2 - (7.0/960.0)*xx4 + (31.0/8064.0)*xx4*xx2 - (127.0/30720.0)*xx4*xx4
```
kortschak: While you are here, please change to `1./24.` to `1.0/24.0`.

person142:
@kortschak would you prefer the fixes be squashed into one commit or kept separate?

kortschak:
There's nothing egregious here, so I reckon a squash of fixes works. Alternatively leave them separate and make a separate PR for the addition to A+C.

mathext/digamma.go
```diff
@@ -10,19 +10,35 @@ import (
 
 // Digamma returns the logorithmic derivative of the gamma function at x.
 //  ψ(x) = d/dx (Ln (Γ(x)).
-// Note that if x is a negative integer in [-7, 0] this function will return
-// negative Inf.
 func Digamma(x float64) float64 {
-	// This is adapted from
```
btracey: Thanks for the extra citations. I still feel like we owe a reference to that code. It gives the algorithm as well as the specific values for the polynomial expansion.

btracey: I also opened https://github.com/gonum/gonum/issues/302 for reference.

btracey:
Algorithm changes LGTM.

person142:
Review comments (hopefully) addressed.

Benchmarks: on this branch I get

```
go test -bench=BenchmarkDigamma
goos: darwin
goarch: amd64
pkg: gonum.org/v1/gonum/mathext
BenchmarkDigamma-8   	300000000	         4.82 ns/op
PASS
ok  	gonum.org/v1/gonum/mathext	6.297s
```

On master it times out:
```
go test -bench=BenchmarkDigamma
SIGQUIT: quit
PC=0x1105204 m=0 sigcode=0

[snip]

*** Test killed with quit: ran too long (10m0s).
FAIL	gonum.org/v1/gonum/mathext	600.011s
```

kortschak:
That's a pretty compelling benchmark! I think what has happened though is that it's timed out on the tests and failed to get to the bench. You can skip all the tests by adding `-run ^$` to regex out the full suite.

I get the following:
```
name       old time/op  new time/op  delta
Digamma-8   8.37s ± 0%   0.00s ± 2%  -100.00%  (p=0.000 n=9+10)
```

(The 0s is ~8.3ns).

kortschak:
Thanks and welcome.

PR #298, opened 2017-11-14
mat: add ability to compute the Cholesky when adding a row and column…

btracey:
… to the factorized matrix

### DIFF
--- a/mat/cholesky.go
+++ b/mat/cholesky.go
@@ -341,7 +341,7 @@ func (c *Cholesky) Scale(f float64, orig *Cholesky) {
 }
 
 // ExtendVecSym computes the Cholesky decomposition of the original matrix A,
-// whose Cholesky decomposition is in a, extended by an n×1 vector according to
+// whose Cholesky decomposition is in a, extended by a the n×1 vector v according to
 //  [A  w]
 //  [w' k]
 // where k = v[n-1] and w = v[:n-1]. The result is stored into the receiver.
@@ -349,7 +349,7 @@ func (c *Cholesky) Scale(f float64, orig *Cholesky) {
 // that n > w' A^-1 w. If this condition does not hold then ExtendVecSym will
 // return false and the receiver will not be updated.
 //
-// ExtendVecSym will panic if len(v) != a.Size()+1 or if a does not contain
+// ExtendVecSym will panic if v.Len() != a.Size()+1 or if a does not contain
 // a valid decomposition.
 func (chol *Cholesky) ExtendVecSym(a *Cholesky, v Vector) (ok bool) {
 	n := a.Size()
--- a/mat/cholesky.go
+++ b/mat/cholesky.go
@@ -340,65 +340,65 @@ func (c *Cholesky) Scale(f float64, orig *Cholesky) {
 	c.cond = orig.cond // Scaling by a positive constant does not change the condition number.
 }
 
-// AppendOneSym computes the Cholesky decomposition of the original matrix A
-// with the data appended to the original matrix A, storing the result into the
-// receiver. That is orig contains the Cholesky decomposition of A, and at completion
-// it contains the Cholesky decomposition of
-//  [A a]
-//  [a k]
-// where k = r[len(r)-1] and a = r[:len(r)-1]. In order for the updated matrix
-// to be positive definite, it must be the case that n > a' A^-1 a. If this
-// condition does not hold then AppendOneSym will return false and the receiver
-// will not be updated.
+// ExtendVecSym computes the Cholesky decomposition of the original matrix A,
+// whose Cholesky decomposition is in a, extended by an n×1 vector according to
+//  [A  w]
+//  [w' k]
+// where k = v[n-1] and w = v[:n-1]. The result is stored into the receiver.
+// In order for the updated matrix to be positive definite, it must be the case
+// that n > w' A^-1 w. If this condition does not hold then ExtendVecSym will
+// return false and the receiver will not be updated.
 //
-// AppendRow will panic if len(r) != orig.Size()+1 or if orig does not contain
+// ExtendVecSym will panic if len(v) != a.Size()+1 or if a does not contain
 // a valid decomposition.
-func (chol *Cholesky) AppendOneSym(orig *Cholesky, r Vector) (ok bool) {
-	n := orig.Size()
-	if r.Len() != n+1 {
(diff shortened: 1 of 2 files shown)


btracey:
PTAL @kortschak . This is a 70% implementation.

This arises in things like Gaussian processes, and other Kernel methods, when a new data point is acquired. This kind of update is only n^2, while computing the full Kernel matrix and taking its Cholesky is n^3.

There are two things I don't like. The first is the name. I'm not sure what a good name is when you're adding a row and column. The second is that this only allows adding one element at a time. The issue is the API for the more general version, where the 'correct' abstraction is a rectangular type, which we do not currently have, and I'm not sure if it's useful enough to fully support (though we do have BandRectangular, so maybe). The third is that this allocates every time it's called, even if orig = chol, which is a lot of unnecessary allocation and copying. We don't really have an automatic "growing" schedule like `append`, so I'm not sure what the correct fix here is.

mat/cholesky.go
```diff
@@ -340,6 +340,70 @@ func (c *Cholesky) Scale(f float64, orig *Cholesky) {
 	c.cond = orig.cond // Scaling by a positive constant does not change the condition number.
 }
 
+// AppendOneSym computes the Cholesky decomposition of the original matrix A
```
kortschak: I'm not enthralled by the doc comment; it reads as though the `orig` value is updated, but it is not.

My suggestion (still not good - I usually defer to @vladimir-ch for documentation clarity):

```
// ExtendVecSym computes the Cholesky decomposition of the matrix A, whose
// Cholesky decomposition is in a, with extended with an n×1 vector symmetrically
// according to
//  [A  v]
//  [v' k]
// where k = v[n-1] and v = v[:n-1]. The result is stored into the receiver.
// In order for the updated matrix to be positive definite, it must be the
// case that n > v' A^-1 v. If this condition does not hold then ExtendVecSym
// will return false and the receiver will not be updated.
//
// ExtendVecSym will panic if len(v) != a.Size()+1 or if a does not contain
// a valid decomposition.
```

This requires that the notation in the comments below changes, but that notation is already potentially confusing since the A/a typography suggests a relationship between them that does not exist.

mat/cholesky.go
```diff
@@ -340,6 +340,70 @@ func (c *Cholesky) Scale(f float64, orig *Cholesky) {
...
+	}
+	if !orig.valid() {
+		panic(badCholesky)
+	}
+
+	// The algorithm is commented here, but see also
+	//  https://math.stackexchange.com/questions/955874/cholesky-factor-when-adding-a-row-and-column-to-already-factorized-matrix
+	// We have A and want to compute the Cholesky of
+	//  [A a]
+	//  [a k]
```
kortschak: Where is the k from? I don't see that in the post above, also update a's to v's (and make the lower v => v').

mat/cholesky.go
```diff
@@ -340,6 +340,70 @@ func (c *Cholesky) Scale(f float64, orig *Cholesky) {
...
+	// We want
+	//  [U c]
+	//  [0 d]
+	// to be the updated Choleksy, and so it must be that
```
kortschak: Cholesky

mat/cholesky.go
```diff
@@ -340,6 +340,70 @@ func (c *Cholesky) Scale(f float64, orig *Cholesky) {
...
+	//  [A a] = [U' 0 ] [U c]
+	//  [a k]   [c' d'] [0 d]
```
kortschak: s/d'/d/ since d is a scalar.

mat/cholesky.go
```diff
@@ -340,6 +340,70 @@ func (c *Cholesky) Scale(f float64, orig *Cholesky) {
...
+	// Thus, we need
+	//  1) A = U'U (true by orig being a valid Cholesky decomposition),
+	//  2) U' * c = a  =>  c = U'^-1 a
+	//  3) c'*c + d'*d = n  =>  d = sqrt(n-c'*c)
```
kortschak: s/n/k/g?

mat/cholesky.go
```diff
@@ -340,6 +340,70 @@ func (c *Cholesky) Scale(f float64, orig *Cholesky) {
...
+
+	// First, compute c = U'^-1 a
+	a := NewVecDense(n, nil)
+	for i := 0; i < n; i++ {
```
kortschak: I'm surprised that `CopyVec` takes a `mat.VecDense`. Ah, it's one of the unfinished items in #167. Please add a TODO to change this to `a.CopyVec(...)` when that is fixed.

mat/cholesky.go
btracey: The post above assumes that k = 1, but in general there's nothing in particular requiring that to be be true. The important part is that k is positive, or more accurately that  k - a' A a is positive (and given that a' A a is always positive, this implies that k has to be positive)

mat/cholesky.go
btracey: ExtendedVecSym is a much better name, thanks.

mat/cholesky.go

mat/cholesky.go

mat/cholesky.go

btracey:
PTAL @kortschak

mat/cholesky.go
```diff
@@ -340,6 +340,70 @@ func (c *Cholesky) Scale(f float64, orig *Cholesky) {
 	c.cond = orig.cond // Scaling by a positive constant does not change the condition number.
 }
 
+// ExtendVecSym computes the Cholesky decomposition of the original matrix A,
+// whose Cholesky decomposition is in a, extended by an n×1 vector according to
```
kortschak: "...an n×1 vector, **v**, according..."?

mat/cholesky.go
```diff
@@ -340,6 +340,70 @@ func (c *Cholesky) Scale(f float64, orig *Cholesky) {
...
+//  [A  w]
+//  [w' k]
+// where k = v[n-1] and w = v[:n-1]. The result is stored into the receiver.
+// In order for the updated matrix to be positive definite, it must be the case
+// that n > w' A^-1 w. If this condition does not hold then ExtendVecSym will
+// return false and the receiver will not be updated.
+//
+// ExtendVecSym will panic if len(v) != a.Size()+1 or if a does not contain
```
kortschak: s/len(v)/v.Len()/

btracey:
Done. Thanks. PTAL

PR #299, opened 2017-11-16
floats: add functions for handling NaN payloads

kortschak:
Addresses https://github.com/gonum/gonum/pull/50#issuecomment-306936995.

Please take a look.

### DIFF
--- a/floats/floats.go
+++ b/floats/floats.go
@@ -497,7 +497,7 @@ const (
 
 // NaNWith returns an IEEE 754 "quiet not-a-number" value with the
 // payload specified in the low 51 bits of payload.
-// The NaN returned by math.NaN has a bit pattern equal to NaN(1).
+// The NaN returned by math.NaN has a bit pattern equal to NaNWith(1).
 func NaNWith(payload uint64) float64 {
 	return math.Float64frombits(nanBits | (payload &^ nanMask))
 }
--- a/floats/floats.go
+++ b/floats/floats.go
@@ -502,9 +502,9 @@ func NaNWith(payload uint64) float64 {
 	return math.Float64frombits(nanBits | (payload &^ nanMask))
 }
 
-// NaNPayload returns the lowest 51 bits of an IEEE 754 "quiet
-// not-a-number" and true, or zero and false if f is not NaN
-// or is not quiet.
+// NaNPayload returns the lowest 51 bits payload of an IEEE 754 "quiet
+// not-a-number". For values of f other than quiet-NaN, NaNPayload
+// returns zero and false.
 func NaNPayload(f float64) (payload uint64, ok bool) {
 	b := math.Float64bits(f)
 	if b&nanBits != nanBits {
--- a/floats/floats.go
+++ b/floats/floats.go
@@ -490,6 +490,30 @@ func MulTo(dst, s, t []float64) []float64 {
 	return dst
 }
 
+const (
+	nanBits = 0x7ff8000000000000
+	nanMask = 0xfff8000000000000
+)
+
+// NaN returns an IEEE 754 "quiet not-a-number" value with the
+// payload specified in the low 51 bits of payload.
+// The NaN returned by math.NaN has a bit pattern equal to NaN(1).
+func NaN(payload uint64) float64 {
+	payload &^= nanMask
+	return math.Float64frombits(nanBits | payload)
+}
+
+// NaNPayload returns the lowest 51 bits of an IEEE 754 "quiet
+// not-a-number" and true, or zero and false if f is not NaN
+// or is not quiet.
+func NaNPayload(f float64) (payload uint64, ok bool) {
+	b := math.Float64bits(f)
+	if b&nanBits != nanBits {
+		return 0, false
+	}
+	return b &^ nanMask, true
+}
+
 // Nearest returns the index of the element in s
 // whose value is nearest to v.  If several such
 // elements exist, the lowest index is returned.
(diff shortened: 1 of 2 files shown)


kortschak:
We can add sign handling if wanted `NaN(sign int, payload uint64) float64`; sign is sometimes used for things here apparently, though `math.Copysign` will do that for both directions (NaN->payload and payload->NaN).

floats/floats.go
```diff
@@ -490,6 +490,30 @@ func MulTo(dst, s, t []float64) []float64 {
...
+const (
+	nanBits = 0x7ff8000000000000
+	nanMask = 0xfff8000000000000
+)
+
+// NaN returns an IEEE 754 "quiet not-a-number" value with the
+// payload specified in the low 51 bits of payload.
+// The NaN returned by math.NaN has a bit pattern equal to NaN(1).
+func NaN(payload uint64) float64 {
+	payload &^= nanMask
```
btracey: Is there a reason to have this two lines?
`math.Float64frombits(nanBits | (payload &^ nanMask))`

floats/floats.go
```diff
@@ -490,6 +490,30 @@ func MulTo(dst, s, t []float64) []float64 {
...
+
+// NaNPayload returns the lowest 51 bits of an IEEE 754 "quiet
+// not-a-number" and true, or zero and false if f is not NaN
+// or is not quiet.
+func NaNPayload(f float64) (payload uint64, ok bool) {
+	b := math.Float64bits(f)
+	if b&nanBits != nanBits {
+		return 0, false
+	}
+	return b &^ nanMask, true
```
btracey: Sorry, is this right? I'm not great at bitwise arithmetic yet. `b &^ nanMask` keeps all of the `nanMask` bits of `b` and turns all of the rest to 0, doesn't it? It seems like that's exactly the opposite of what we want. The test passes, so I must not be understanding something ...

floats/floats.go
```diff
@@ -490,6 +490,30 @@ func MulTo(dst, s, t []float64) []float64 {
...
 
+const (
+	nanBits = 0x7ff8000000000000
+	nanMask = 0xfff8000000000000
+)
+
+// NaN returns an IEEE 754 "quiet not-a-number" value with the
+// payload specified in the low 51 bits of payload.
+// The NaN returned by math.NaN has a bit pattern equal to NaN(1).
+func NaN(payload uint64) float64 {
```
btracey: Just wondering if this should be `NaNWith` or something to help distinguish it a bit from `NaN`

floats/floats.go
```diff
@@ -490,6 +490,30 @@ func MulTo(dst, s, t []float64) []float64 {
...
 }
 
+const (
+	nanBits = 0x7ff8000000000000
+	nanMask = 0xfff8000000000000
+)
+
+// NaN returns an IEEE 754 "quiet not-a-number" value with the
+// payload specified in the low 51 bits of payload.
+// The NaN returned by math.NaN has a bit pattern equal to NaN(1).
```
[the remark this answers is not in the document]
kortschak: It's not guaranteed. Nor is the behaviour of `math.Copysign`.

floats/floats.go
kortschak: Nice. Done.

floats/floats.go
kortschak: Fear of parentheses? Done.

floats/floats.go
kortschak: I think you are confusing `&` and `&^`. We want to keep all bits in `b` that are zero in `nanMask`.

kortschak:
PTAL

floats/floats.go
btracey: Ah, yes. Thanks.

kortschak:
Sorry, just fixed a comment that became incorrect after the last commit. PTAL.

btracey:
Oops. Sorry

kortschak:
An alternative signature pair is `NaNWith(payload uin64, signalling bool) float64` and `NaNPayload(f float64) (payload uint64, signalling, ok bool)`. In this case the payloads would be normalised by doing a shift of `payload` by one bit if `signalling` is true. This obviates the issue of failing, but adds complexity. I think probably we just leave it as it is now.

btracey:
I think it's better as is for now. As far as I understand, one of the main things this attempts to enable is to use these as a "not available" value. Additionally, as I understand, those kinds of values should also use quiet NaNs. Seems best to leave simple until we have another reason.

floats/floats.go
```diff
@@ -490,6 +490,30 @@ func MulTo(dst, s, t []float64) []float64 {
...
+func NaN(payload uint64) float64 {
+	payload &^= nanMask
+	return math.Float64frombits(nanBits | payload)
+}
+
+// NaNPayload returns the lowest 51 bits of an IEEE 754 "quiet
+// not-a-number" and true, or zero and false if f is not NaN
+// or is not quiet.
```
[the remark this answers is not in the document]
kortschak: The size of the payload is important and the phrasing "X returns ...." is conventional. How about:

```
// NaNPayload returns the lowest 51 bits payload of an IEEE 754 "quiet not-a-number".
// For values of f other than quiet-NaN, NaNPayload returns zero and false.
```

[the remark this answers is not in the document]
Kunde21: That reads much better.  endorsed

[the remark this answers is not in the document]
kortschak: Sent. Please reapprove.

Issue #228, opened 2017-09-05
all: consider generalising the random source parameter type


kortschak:
We use a `*rand.Rand` as the source of randomness for routines we provide that need this. Recently, a new rand package was added to the golang/exp repo. For the most part this is a drop in replacement, but gonum clients can not use this. If we define a rand interface that matches the method set of `*rand.Rand` then that package will be usable. Is this something that we want to do? The alternative is to back up to the `rand.Source`-ish level and take that as the argument - this needs a shim to convert the signatures between the two.

kortschak:
@btracey Do you have a view on this?

sbinet:
I am ok with that even if we should also perhaps check going through an interface call doesn't impact too negatively a few meaningful (real life?) benchmarks.

note there are a few interesting differences between the two, and we should be careful not to paint ourself into a corner while creating our own `Rander` interface.

----

`exp/rand.Rand` and `math/rand.Rand` do have a few differences:

- https://godoc.org/golang.org/x/exp/rand#Rand
- https://godoc.org/math/rand#Rand

ie: `exp/rand.Rand` has:
- `func (r *Rand) Uint64n(n uint64) uint64`
- `func (r *Rand) Seed(seed uint64)`

while `math/rand.Rand` doesn't have `Uint64n` and has the following signature:
- `func (r *Rand) Seed(seed int64)`

btracey:
Yes, I think we should do this.

kortschak:
We should wait for the proposal document that @robpike has said is coming.

robpike:
The proposal is #21835.

kortschak:
Thanks, Rob.

kortschak:
On the basis of the proposal, I'd like to suggest that we use the exp/rand.Source and change from math/rand to exp/rand, and then when the Go standard library makes the shift, change the import paths. This seems like the easiest path to follow.

@robpike The proposal is almost silent on the behaviour of the `rand.Source` that would by used in the Go2 math/rand package; if the implementation being used in exp/rand is used verbatim, there would be a change in the source stream method's signature from `Int64() int64` to `Uint64() uint64`, but this is not explicitly discussed in the proposal. Would that change be part of the proposal (the comment above is contingent on this being true).

robpike:
@kortschak That question is in fact unresolved, but perhaps it will end up that the new Source contains both.

I would be happy to work with you to figure out what exp/rand should look like for you. That is somewhat independent of the proposal.

[...]

PR #301, opened 2017-11-17
all: migrate to golang.org/x/exp/rand

kortschak:
Please take a look.

This is the first step in addressing #228.

The first commit is:
```
sed -i 's|"math/rand"|"golang.org/x/exp/rand"|g' $(grep -Rl '"math/rand"')
goimports -w .
.travis/check-generate.sh
```
followed by a visual check and single manual edit to put "golang.org/x/exp/rand" in the same import group as another "golang.org/x/" import in stat/distmv/studentst.go.

The second commit is generally adjusting `want` values, loosening tolerances, increasing `n`s for samples or fixing non-deterministic tests (`Rejection` was nasty here). Only one change was needed to get compilation to succeed.

A concern here is that the signatures do not visually change, and I'm not sure that we can paint a bright enough sign to ensure that people use "exp/rand" rather than "math/rand".

Updates #228.


kortschak:
Note that "autogen_bench_level1double.go" is incorrect, but I don't understand that file and it seems to be still unaware of the separation of Go-native and Cgo packages.

kortschak:
Well damn. It looks like we can't do this for about six months; exp/rand depends on math/bits which does not exist until 1.9. The breakage *could* be fixed by vendoring, but that is a no no since the global state of the package become different for other imports of exp/rand.

kortschak:
Yay for open source!

PTAL

btracey:
Is the plan to migrate back when the changes are ported?

kortschak:
Ported? The math/rand change is a Go2 milestone.

stat/samplemv/sample_test.go
```diff
@@ -94,15 +94,15 @@ func TestRejection(t *testing.T) {
...
+		min := src.NormFloat64()
+		max := src.NormFloat64()
 		if min > max {
 			min, max = max, min
 		}
 		bounds[i].Min = min
 		bounds[i].Max = max
 	}
-	target := distmv.NewUniform(bounds, nil)
+	target := distmv.NewUniform(bounds, src)
```
btracey: I guess it doesn't hurt for this to be here, though this was intentionally kept as `nil` because no random numbers are generated from `target`.

graph/community/louvain_undirected_test.go
```diff
@@ -153,13 +153,16 @@ var communityUndirectedQTests = []communityUndirectedQTest{
...
-				q: 0.39907955292570674,
+				q: 0.3496877054569362,
 				communities: [][]graph.Node{
 					{simple.Node(0), simple.Node(1), simple.Node(2), simple.Node(3), simple.Node(7), simple.Node(11), simple.Node(12), simple.Node(13), simple.Node(17), simple.Node(19), simple.Node(21)},
 					{simple.Node(4), simple.Node(10)},
 					{simple.Node(5), simple.Node(6), simple.Node(16)},
-					{simple.Node(8), simple.Node(9), simple.Node(14), simple.Node(15), simple.Node(18), simple.Node(20), simple.Node(22), simple.Node(26), simple.Node(29), simple.Node(30), simple.Node(32), simple.Node(33)},
-					{simple.Node(23), simple.Node(24), simple.Node(25), simple.Node(27), simple.Node(28), simple.Node(31)},
+					{simple.Node(8), simple.Node(9), simple.Node(14), simple.Node(15), simple.Node(18), simple.Node(20), simple.Node(22), simple.Node(30), simple.Node(32), simple.Node(33)},
+					{simple.Node(23), simple.Node(25)},
```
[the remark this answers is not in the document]
kortschak: It's not a rewrite, I needed to change intermediate community reduction results to get the test to pass. This is OK because the algorithm is randomised and so doesn't guarantee the same results and more significantly the intermediate result shown here are not necessarily meaningful as a community modularisation.

stat/samplemv/sample_test.go
kortschak: OK. I can revert this if you want; the test still passes without the change (the changes as 97 and 98 were necessary for the test to pass in ways that were difficult to understand - revert those locally to see what happens - you need to run the entire package, not just TestRejection).

I'll wait.

graph/community/louvain_undirected_test.go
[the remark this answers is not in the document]
btracey: Great. Just wanted to check.

stat/samplemv/sample_test.go
btracey: Nah, I don't think it needs to be reverted.

PR #303, opened 2017-11-17
stat/combin: add Cartesian for constructing the cartesian product of …

btracey:
…a set of slices

### DIFF
--- a/stat/combin/combin.go
+++ b/stat/combin/combin.go
@@ -187,7 +187,7 @@ func nextCombination(s []int, n, k int) {
 
 // Cartesian returns the cartesian product of the slices in data. The Cartesian
 // product of two sets is the set of all combinations of the items. For example,
-// gives the input
+// given the input
 //  [][]float64{{1,2},{3,4},{5,6}}
 // the returned matrix will be
 //  [ 1 3 5 ]
@@ -198,7 +198,7 @@ func nextCombination(s []int, n, k int) {
 //  [ 2 3 6 ]
 //  [ 2 4 5 ]
 //  [ 2 4 6 ]
-// If dst is nil, a new matrix will be allocated and used, otherwise the number
+// If dst is nil, a new matrix will be allocated and returned, otherwise the number
 // of rows of dst must equal \prod_i len(data[i]), and the number of columns in
 // dst must equal len(data). Cartesian also panics if len(data) = 0.
 func Cartesian(dst *mat.Dense, data [][]float64) *mat.Dense {
--- a/stat/combin/combin.go
+++ b/stat/combin/combin.go
@@ -4,12 +4,17 @@
 
 package combin
 
-import "math"
+import (
+	"math"
+
+	"gonum.org/v1/gonum/mat"
+)
 
 const (
-	badNegInput = "combin: negative input"
-	badSetSize  = "combin: n < k"
-	badInput    = "combin: wrong input slice length"
+	badNegInput          = "combin: negative input"
+	badSetSize           = "combin: n < k"
+	badInput             = "combin: wrong input slice length"
+	nonpositiveDimension = "combin: non-positive dimension"
 )
 
 // Binomial returns the binomial coefficient of (n,k), also commonly referred to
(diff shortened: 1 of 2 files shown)


btracey:
PTAL @sbinet  @kortschak

stat/combin/combin.go
```diff
@@ -179,3 +184,115 @@ func nextCombination(s []int, n, k int) {
...
+//  [ 1 4 5 ]
+//  [ 1 4 6 ]
+//  [ 2 3 5 ]
+//  [ 2 3 6 ]
+//  [ 2 4 5 ]
+//  [ 2 4 6 ]
+// If dst is nil, a new matrix will be allocated and returned, otherwise the number
+// of rows of dst must equal \prod_i len(data[i]), and the number of columns in
+// dst must equal len(data). Cartesian also panics if len(data) = 0.
+func Cartesian(dst *mat.Dense, data [][]float64) *mat.Dense {
```
kortschak: I think an iterator API would be better here; it's not alway the case that you would want a matrix for these values, you may want to iterate over the products. So `Cartesian(fn func([]float64), data [][]float64)` might be better.

btracey: In that package, we also have Combination and CombinationGenerator. I can add CartesianGenerator in a future PR.

kortschak: SGTM

stat/combin/combin_test.go
```diff
@@ -8,9 +8,24 @@ import (
 	"math/big"
 	"testing"
 
+	"gonum.org/v1/gonum/mat"
+
 	"gonum.org/v1/gonum/floats"
 )
 
+// intsEqual returns whether the two slices of int are equal.
```
kortschak: Why not use `reflect.DeepEqual`?

stat/combin/combin_test.go
```diff
@@ -179,3 +194,96 @@ func TestCombinationGenerator(t *testing.T) {
...
+		}
+		to := make([]int, len(dims))
+		for idx := 0; idx < maxIdx; idx++ {
+			sub := SubFor(nil, idx, dims)
+			for i := range sub {
+				if sub[i] < 0 || sub[i] >= dims[i] {
+					t.Errorf("cas %v: bad subscript. dims: %v, sub: %v", cas, dims, sub)
+				}
+			}
+			SubFor(to, idx, dims)
```
kortschak: Can you make this more explicit that the test is looking at side-effect changes to `to`? Maybe `to = SubFor(to, idx, dims)`.

btracey: I don't think that is good style. If `to` is nil, then it silently hides that new data is allocated. I could change the name to `into` if that  would help.

stat/combin/combin_test.go
btracey: Done.

stat/combin/combin_test.go
btracey: Changed to into

kortschak: I don't understand, `to`/`into` are never nil here, since you have called `make` [here](https://github.com/gonum/gonum/pull/303/files/1aa30b0e5a03cb39cf9e0c2280fe932709e65783#diff-ae5dcb80ac292c22ab37585192b8fe1eR259).

btracey: I agree. I just think it's bad practice to have  `dst = Func(dst, data)`

I've found it best to have it either be `Func(dst,data)` or `dst = Func(nil, data)`. One reason is accidentally obscuring nil data.

kortschak: Yeah, I'm not convinced, but I'm not going to push it.

stat/combin/combin_test.go
```diff
@@ -6,8 +6,11 @@ package combin
 
 import (
 	"math/big"
+	"reflect"
 	"testing"
 
+	"gonum.org/v1/gonum/mat"
+
```
kortschak: Delete blank line (goimports scar).

stat/combin/combin_test.go
```diff
@@ -179,3 +182,96 @@ func TestCombinationGenerator(t *testing.T) {
...
+		1, 4, 5,
+		1, 4, 6,
+		2, 3, 5,
+		2, 3, 6,
+		2, 4, 5,
+		2, 4, 6,
+	})
+	got := Cartesian(nil, data)
+	if !mat.Equal(want, got) {
+		t.Errorf("cartesian data mismatch. \nwant:\n%v\ngot:\n%v", mat.Formatted(want), mat.Formatted(got))
```
kortschak: Here and in the next errorf remove the space before the `\n`.

stat/combin/combin_test.go
btracey: Done.

stat/combin/combin_test.go
btracey: Done.

btracey:
PTAL

PR #306, opened 2017-11-18
mathext: use relative error in Digamma tests

person142:
Relative error is the more important measure of accuracy for special
functions except for sometimes around zeros.

### DIFF
--- a/mathext/digamma_test.go
+++ b/mathext/digamma_test.go
@@ -29,7 +29,7 @@ func TestDigamma(t *testing.T) {
 		{-1.111111111e9, 30.497454343508262861},
 	} {
 
-		if got := Digamma(test.x); math.Abs(got-test.want) > 1e-10 {
+		if got := Digamma(test.x); math.Abs(got-test.want) > 1e-10*math.Abs(test.want) {
 			t.Errorf("test %d Digamma(%g) failed: got %g want %g", i, test.x, got, test.want)
 		}
 	}


person142:
For example, consider `Digamma(1.46)` (here `1.46` was chosen because it is close to a zero). If I run:

```go
package main

import (
	"fmt"
	"math"
	"gonum.org/v1/gonum/mathext"
)

func main () {
	want := -0.0015805619870834176761
	got := mathext.Digamma(1.46)
	fmt.Printf("Absolute error is %v\n", math.Abs(want-got))
	fmt.Printf("Relative error is %v\n", math.Abs((want-got)/want))
}
```

then I get:

```
Absolute error is 2.6834123681776512e-11
Relative error is 1.6977583860088294e-08
```

The current test would miss that the number of significant figures is slipping, and it could be doing much better:

```python
>>> import scipy
>>> scipy.__version__
'1.0.0'
>>> import scipy.special as sc
>>> want = -0.0015805619870834176761
>>> got = sc.digamma(1.46)
>>> abs(want - got)
3.4477629085039041e-17
>>> abs((want - got)/want)
2.1813525421207923e-14
```

mathext/digamma_test.go
```diff
@@ -29,7 +29,7 @@ func TestDigamma(t *testing.T) {
 		{-1.111111111e9, 30.497454343508262861},
 	} {
 
-		if got := Digamma(test.x); math.Abs(got-test.want) > 1e-10 {
+		if got := Digamma(test.x); math.Abs(got-test.want) > 1e-10*math.Abs(test.want) {
```
kortschak: We have `floats.EqualWithin*` for this kind of thing (they should have been used here in the original code). See [EqualWithinAbs](https://godoc.org/gonum.org/v1/gonum/floats#EqualWithinAbs), [EqualWithinRel](https://godoc.org/gonum.org/v1/gonum/floats#EqualWithinRel),  [EqualWithinRel](https://godoc.org/gonum.org/v1/gonum/floats#EqualWithinAbsOrRel),  and [EqualWithinULP](https://godoc.org/gonum.org/v1/gonum/floats#EqualWithinULP): take your pick.

btracey:
Ping?

PR #313, opened 2017-11-25
graph/topo: add 2-sat example

kortschak:
Systems in example checked with minisat.

Please take a look.


kortschak:
Just for interest, I've used a simpler version of this taking an almost-dimacs input of 1e6 clauses and 8.5e5 variables and it has not problem.

PR #315, opened 2017-11-27
Update roc_example_test.go

MaxHalford:
Just a little typo I noticed in the ROC AUC basic example.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -3,7 +3,7 @@
 # See the latter for an explanation.
 
 # Names should be added to this file as
-#	Name or Organization <email address>
+#   Name or Organization <email address>
 # The email address is not required for organizations.
 
 # Please keep the list sorted.
@@ -42,6 +42,7 @@ Konstantin Shaposhnikov <k.shaposhnikov@gmail.com>
 Leonid Kneller <recondite.matter@gmail.com>
 Lyron Winderbaum <lyron.winderbaum@student.adelaide.edu.au>
 Matthieu Di Mercurio <matthieu.dimercurio@gmail.com>
+Max Halford <maxhalford25@gmail.com>
 MinJae Kwon <k239507@gmail.com>
 Or Rikon <rikonor@gmail.com>
 Pontus Melke <pontusmelke@gmail.com>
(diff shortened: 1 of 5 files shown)


btracey:
I'm pretty sure this example is correct. The ROC curve is usually plotted with the true positive rate on the y axis, and the false positive rate on the x axis (see, e.g. https://en.wikipedia.org/wiki/Receiver_operating_characteristic), which is how the information is passed to Trapezoidal.

MaxHalford:
Maybe I was too hasty. The thing is I ran the example with sklearn and got 0.25.

```python
>>> from sklearn import metrics
>>> y = [0.1, 0.35, 0.4, 0.8]
>>> classes = [1, 0, 1, 0]
>>> metrics.roc_auc_score(classes, y)
0.25
```

What's more with the order of the example a perfect match between `y` and `classes` yields 0 and not 1. Maybe this is correct but it isn't very intuitive.

btracey:
Thanks for the report. Seems like something is wrong then. I'm not sure it's the line you changed though.

btracey:
If you look at the Scikit learn example (http://scikit-learn.org/stable/modules/generated/sklearn.metrics.roc_auc_score.html) they have the same data we do, except their values for `true` and `false` are flipped. 

Our documentation says 
```
// For a given cutoff value, observations corresponding to entries in y
// greater than the cutoff value are classified as false, while those
// below (or equal to) the cutoff value are classified as true. 
```

I believe this means that the behavior as documented is correct, but I wonder if this is a backward convention from most implementations. I would have assumed that `y` roughly corresponds to `p(true|y)` (that is, the y values can be interpreted as probability of being true), but it seems like we treat it as the opposite.

btracey:
@kortschak @sbinet @Armadilloa16

MaxHalford:
Indeed I guess this is just a matter of conventions. I think your assumption is correct and I believe it would make sense to most people. However if that assumption holds then intuitively a perfect match between two arrays/slices should be equal to 1 (because `p(true|y) = 1`).

btracey:
Agreed. If we run the example with
```
y := []float64{0, 0, 1, 1}
classes := []bool{true, true, false, false}
```
Then you get an AUC of 1. This feels backward to me. 

Could you please open an issue instead of this PR? Title would be something like "stat: Change parity convention in ROC", you can cite the example, and link this discussion.

MaxHalford:
Of course, no problem. Sorry for making a pull request before opening an issue.

btracey:
No problem. You thought it was a simple fix, but it turns out the actual fix requires discussion.

PR #317, opened 2017-11-27
stat: implement Kendall correlation

MaxHalford:
Hey,

I'm really interested by gonum so I wanted to start with a little PR to implement Kendall correlation. I not there hasn't been any issue asking for it but I need it for one of my projects and I though I might as well make a PR.

### DIFF
--- a/stat/stat.go
+++ b/stat/stat.go
@@ -217,11 +217,14 @@ func Correlation(x, y, weights []float64) float64 {
 	return (sxy - xcompensation*ycompensation/sumWeights) / math.Sqrt(sxx*syy)
 }
 
-// KendallCorrelation returns the weighted Tau-a Kendall correlation between the
-// samples of x and y.
+// Kendall returns the weighted Tau-a Kendall correlation between the
+// samples of x and y. The Kendall correlation measures the quantity of
+// concordant and discordant pairs of numbers. If weights are specified then
+// each pair is weighted by weights[i] * weights[j] and the final sum is
+// normalized to stay between -1 and 1.
 // The lengths of x and y must be equal. If weights is nil then all of the
 // weights are 1. If weights is not nil, then len(x) must equal len(weights).
-func KendallCorrelation(x, y, weights []float64) float64 {
+func Kendall(x, y, weights []float64) float64 {
 	if len(x) != len(y) {
 		panic("stat: slice length mismatch")
 	}
@@ -250,15 +253,12 @@ func KendallCorrelation(x, y, weights []float64) float64 {
 
 	var sumWeights float64
 
-	var c int
-
 	for i := 0; i < n; i++ {
 		for j := i; j < n; j++ {
 			if i == j {
 				continue
 			}
-			c++
-			var weight = weights[i] * weights[j]
+			weight := weights[i] * weights[j]
 			if math.Signbit(x[j]-x[i]) == math.Signbit(y[j]-y[i]) {
 				cc += weight
 			} else {
--- a/stat/stat.go
+++ b/stat/stat.go
@@ -217,6 +217,35 @@ func Correlation(x, y, weights []float64) float64 {
 	return (sxy - xcompensation*ycompensation/sumWeights) / math.Sqrt(sxx*syy)
 }
 
+// KendallCorrelation returns the weighted Kendall correlation between the
+// samples of x and y.
+// The lengths of x and y must be equal.
+func KendallCorrelation(x, y []float64) float64 {
+	if len(x) != len(y) {
+		panic("stat: slice length mismatch")
+	}
+
+	var (
+		cc int // number of concordant pairs
+		dc int // number of discordant pairs
+		n  = len(x)
+	)
+
+	for i := 0; i < n; i++ {
+		for j := i; j < n; j++ {
+			if i == j {
+				continue
+			}
+			if math.Signbit(x[j]-x[i]) == math.Signbit(y[j]-y[i]) {
+				cc++
+			} else {
+				dc++
+			}
+		}
+	}
+	return float64(cc-dc) / float64(n*(n-1)/2)
+}
+
 // Covariance returns the weighted covariance between the samples of x and y.
 //  sum_i {w_i (x_i - meanX) * (y_i - meanY)} / (sum_j {w_j} - 1)
 // The lengths of x and y must be equal. If weights is nil then all of the
--- a/stat/stat.go
+++ b/stat/stat.go
@@ -220,30 +220,53 @@ func Correlation(x, y, weights []float64) float64 {
 // KendallCorrelation returns the weighted Kendall correlation between the
 // samples of x and y.
 // The lengths of x and y must be equal.
-func KendallCorrelation(x, y []float64) float64 {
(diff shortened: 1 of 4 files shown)


btracey:
Thanks for the PR. I have a really busy week this week, but I'll take a look when I can.

sbinet:
a quick pass-by comment: could you squash the first 2 commits into one with the following commit message:
`A+C: add MaxHalford`

(we haven't been completely consistent wrt the formating of that authoring thing but I think we kind of converged on that wording. and it's easier to grep for it, history-wise.)

thanks.

MaxHalford:
I've never done a squash so I don't know if I did it well. The squashed commit seems to be there but it didn't remove the old commits. Is that normal?

stat/stat.go
```diff
@@ -217,6 +217,35 @@ func Correlation(x, y, weights []float64) float64 {
 	return (sxy - xcompensation*ycompensation/sumWeights) / math.Sqrt(sxx*syy)
 }
 
+// KendallCorrelation returns the weighted Kendall correlation between the
+// samples of x and y.
+// The lengths of x and y must be equal.
+func KendallCorrelation(x, y []float64) float64 {
```
btracey: The comment says that it's a weighted Kendall correlation, but this doesn't have weights. My impression is that this test doesn't make sense with weighted data, but I'm not an expert in this test. Do you agree @sbinet @kortschak @jonlawlor ?

btracey:
I approved, but waiting for a second opinion on the weighting. If there is a weighted version, that would be better.

stat/stat.go
kortschak: @Armadilloa16 Do you have some time for this?

I found [this](http://au.mathworks.com/matlabcentral/fileexchange/27361-weighted-kendall-rank-correlation-matrix) that suggests a weighted form is possible, but I don't know enough about it to be able to comment otherwise. Also [this](http://www.sciencedirect.com/science/article/pii/S0167715298000066) - apologies for anyone not at an institution.

Armadilloa16:
Hey sorry for not being active in awhile.

Yeah a weighted form is possible, and fairly straightforward. That proposed in equation (1) of the paper by Shieh you linked in [your comment](https://github.com/gonum/gonum/pull/317#discussion_r155397955) @kortschak is the intuitive way to do this. Essentially it just takes the unweighted Kendall's Tau, weights each pairwise comparison by it's weight, and normalises by the sum of weights.
Should be fairly straightforward, the only complication I see is, as the other link you provided (the mathworks) points out, if you are calculating the correlation between two length-n vectors x and y, then the vector of weights needs to be length n * (n - 1) / 2, as what you are weighting is not values in x and y, but rather the pairwise comparisons between the values in x to the values in y. I believe there is a standard method for unwrapping the triangular matrix of pairwise weights into a vector, as in functions such as pdist in MATLAB, or you could pass the weights as a triangular matrix, but that would add dependancy on matrix. Not sure which is the better option here for a weighted version.

Armadilloa16:
Also, you should probably add mention of which Kendall Tau you are calculating, there are two popular ones, Tau-a and Tau-b, see [the wiki page](https://en.wikipedia.org/wiki/Kendall_rank_correlation_coefficient#Accounting_for_ties). At the moment you are calculating Tau-a. The difference with Tau-b is essentially it adjusts for the possibility of ties. I think just clarifying which one you are calculating is good enough, but if you wanted too you could provide the option to calculate either version, in which case you could make a weighted form of the Tau-b analogous to the weighted form of Tau-a shown in the paper by Shieh. 

Also just for completeness there is also a Tau-c, which is apparently used mostly for rectangular contingency tables, although I'm not sure I understand the intuition behind it.

MaxHalford:
Hey all, thanks for the insightful and detailed comments. I get that a weighted version exists, but it doesn't seem very popular and simple to implemented -- at least in terms of memory allocation. Moreover it doesn't seem to be part of the [scipy library](https://docs.scipy.org/doc/scipy-0.19.1/reference/generated/scipy.stats.kendalltau.html) so I guess that there hasn't been a big need from their community to implement it. However I understand that for completeness and to provide a consistent API this should be implemented. I'll go over it when I get some time -- most likely next week.

stat/stat.go
```diff
@@ -217,6 +217,35 @@ func Correlation(x, y, weights []float64) float64 {
...
+	if len(x) != len(y) {
+		panic("stat: slice length mismatch")
+	}
+
```
btracey: Don't we need to verify that `x` is sorted? Otherwise this test returns meaningless results I think.

MaxHalford:
Okay so I implemented the weighted version with symmetric weights. Tell me what you think about it. I also edited the comment of `KendallCorrelation` to indicate that it's the Tau-a version that is calculated.

stat/stat.go
```diff
@@ -217,6 +217,59 @@ func Correlation(x, y, weights []float64) float64 {
...
+	var c int
+
+	for i := 0; i < n; i++ {
+		for j := i; j < n; j++ {
+			if i == j {
+				continue
+			}
+			c++
+			var weight = weights[i] * weights[j]
```
btracey: `weight := `

stat/stat.go
```diff
@@ -217,6 +217,59 @@ func Correlation(x, y, weights []float64) float64 {
...
+					dc++
+				}
+			}
+		}
+		return (cc - dc) / float64(n*(n-1)/2)
+	}
+
+	var sumWeights float64
+
+	var c int
```
btracey: This is unused?

stat/stat.go
```diff
@@ -217,6 +217,59 @@ func Correlation(x, y, weights []float64) float64 {
 	return (sxy - xcompensation*ycompensation/sumWeights) / math.Sqrt(sxx*syy)
 }
 
+// KendallCorrelation returns the weighted Tau-a Kendall correlation between the
```
btracey: Is the work "Correlation" needed? Can we just call it Kendall?

stat/stat.go
```diff
@@ -217,6 +217,59 @@ func Correlation(x, y, weights []float64) float64 {
...
+// samples of x and y.
```
btracey: I think there needs to be some comment that the weight of a pair is equal to weights[i] * weights[j]. This may require an extended explanation of what Kendal is.

stat/stat.go
MaxHalford: Sure I'm fine with that.

stat/stat.go
MaxHalford: Yep, sorry.

stat/stat.go
MaxHalford: Is this a style issue?

MaxHalford:
@btracey I took your comments into consideration in the latest commit.

stat/stat.go
btracey: Yes.

kortschak: Given the life-span of the temporary, I'd even go so far as to say `w := weights[i] * weights[j]`, but it's not necessary to change that now.

kortschak:
Thank you for doing this.

PR #322, opened 2017-12-04
blas/gonum: add Ztrsv

vladimir-ch:
I had to fix `zsame` which in turn required modifying tests for `Zher` and `Zher2`. This in turn revealed that their behavior is not as documented when `alpha == 0`. I postpone that to another PR. If you wish I can cut out the commits to a prequel PR.

### DIFF
--- a/blas/testblas/common.go
+++ b/blas/testblas/common.go
@@ -254,22 +254,46 @@ func allPairs(x, y []int) [][2]int {
 	return p
 }
 
+func sameFloat64(a, b float64) bool {
+	return a == b || math.IsNaN(a) && math.IsNaN(b)
+}
+
+func sameComplex128(x, y complex128) bool {
+	return sameFloat64(real(x), real(y)) && sameFloat64(imag(x), imag(y))
+}
+
 func zsame(x, y []complex128) bool {
 	if len(x) != len(y) {
 		return false
 	}
 	for i, v := range x {
 		w := y[i]
-		if math.IsNaN(real(v)) && math.IsNaN(imag(v)) && math.IsNaN(real(w)) && math.IsNaN(imag(w)) {
+		if !sameComplex128(v, w) {
+			return false
+		}
+	}
+	return true
+}
+
+// zEqualApprox returns whether vectors x and y with stride inc
+// are approximately equal within tol. Elements at non-strided
+// positions must be same in both x and y.
+func zEqualApprox(x, y []complex128, inc int, tol float64) bool {
+	if len(x) != len(y) {
+		return false
+	}
+	if inc < 0 {
+		inc = -inc
+	}
+	for i, v := range x {
+		w := y[i]
+		if i%inc == 0 {
+			if cmplx.Abs(v-w) > tol {
+				return false
+			}
 			continue
 		}
-		if math.IsNaN(real(v)) && math.IsNaN(real(w)) && imag(v) == imag(w) {
-			continue
-		}
-		if math.IsNaN(imag(v)) && math.IsNaN(imag(w)) && real(v) == real(w) {
-			continue
-		}
-		if v != w {
+		if !sameComplex128(v, w) {
 			return false
 		}
 	}
--- a/blas/testblas/common.go
+++ b/blas/testblas/common.go
@@ -260,7 +260,16 @@ func zsame(x, y []complex128) bool {
 	}
 	for i, v := range x {
 		w := y[i]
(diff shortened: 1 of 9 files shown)


blas/testblas/common.go
```diff
@@ -260,7 +260,16 @@ func zsame(x, y []complex128) bool {
 	}
 	for i, v := range x {
 		w := y[i]
-		if v != w && !math.IsNaN(real(v)) && !math.IsNaN(imag(v)) && !math.IsNaN(real(w)) && !math.IsNaN(imag(w)) {
+		if math.IsNaN(real(v)) && math.IsNaN(imag(v)) && math.IsNaN(real(w)) && math.IsNaN(imag(w)) {
```
kortschak: Could we not have something like 

```
func sameFloat64(a, b float64) bool {
}
```

and then

```
for i, v := range x {
    w := y[i]
    if !sameFloat64(real(v), real(w)) || !sameFloat64(imag(v), imag(w)) {
        return false
    }
}
```

blas/testblas/common.go
```diff
@@ -260,7 +260,16 @@ func zsame(x, y []complex128) bool {
 	}
```
kortschak: Can we have an explanation in the commit message for this for why the change was necessary.

blas/testblas/common.go
```diff
@@ -276,6 +276,28 @@ func zsame(x, y []complex128) bool {
 	return true
 }
```
kortschak: This probably should have been committed before the tests, but github will never render it that way now.

blas/testblas/common.go
```diff
@@ -276,6 +276,28 @@ func zsame(x, y []complex128) bool {
...
 
+func zequal(x, y []complex128, tol float64) bool {
+	if len(x) != len(y) {
+		return false
+	}
+	for i, v := range x {
```
kortschak: Can we have a similar thing here too?

blas/testblas/common.go
```diff
@@ -276,6 +276,28 @@ func zsame(x, y []complex128) bool {
 	return true
 }
 
+func zequal(x, y []complex128, tol float64) bool {
```
kortschak: It seems that this is more `zsameApprox` rather than `zequal`.

blas/testblas/common.go
vladimir-ch: Done.

blas/testblas/common.go
vladimir-ch: I reordered the commits, so maybe now they are rendered as you wished?

blas/testblas/common.go
vladimir-ch: I added `sameFloat64` and then `sameComplex128` which uses it.

blas/testblas/common.go
vladimir-ch: Done.

blas/testblas/common.go
vladimir-ch: I changed the signature and the semantics of this function. I think it is better now.

vladimir-ch:
PTAL

blas/testblas/common.go
kortschak: It will make the commits all pass which is the important thing. Github for some reason renders commits by creation time, not graph order.

blas/testblas/common.go
```diff
@@ -254,13 +254,46 @@ func allPairs(x, y []int) [][2]int {
...
 		w := y[i]
-		if v != w && !math.IsNaN(real(v)) && !math.IsNaN(imag(v)) && !math.IsNaN(real(w)) && !math.IsNaN(imag(w)) {
+		if !sameComplex128(v, w) {
+			return false
+		}
+	}
+	return true
+}
+
+// zEqualApprox returns whether vectors x and y with stride inc
```
kortschak: Thanks. I'm not convinced of the name, but I can't think of anything better and it doesn't face the public, so we'll leave it.

vladimir-ch: What about the name worries you? I'll merge but I think there may be a chance to revisit this later. In general, the supporting functions for blas testing could be cleaned up and unified.

kortschak: The name is indicating that it's testing for approximate equality, but there is no tolerance that will allow NaN == NaN, but conversely, those elements are not the elements that are being tested for approximate equality. A name that does cover it is `zApproxEqualInElementsSameOtherwise`; this is obviously not a good name though.

vladimir-ch: I think that having two functions, one for stride elements and one for out-of-stride elements, would be better. I'll try to do something about it.

Issue #310, opened 2017-11-21
mat: simplify and harmonise shadow detection


kortschak:
Currently we have three different routines to determined element shadowing in mat, one for rectangular data, one for triangular data and one for linear (vector) data. The last one is an optimisation for the first.

The separate approaches taken for rectangular and triangular data are intended to allow space optimisations for triangles, but impose either a combinatoric explosion or (what we have chosen) a restriction on where we guarantee safety (only within values within a particular `Raw*` type - and as a special case for all with vectors). A problem with this approach has arisen in #213 where it makes sense to be able to gain a diagonal of a matrix (`*Dense` or otherwise) into a `*VecDense` and then potentially use that vector as the underlying data in a `*SymBandDense`. There is no way under the current implementation to allow safe ways of doing this while preventing unsafe ways, without always allocating for what should be a zero-allocation operation.

To address this, I think we should drop the triangular shadow detection and only do rectangular (with vector shadowing being a simplifying special case of rectangular for performance reasons). This would simplify the rules:

```
 mat will use the following rules to detect overlap between the receiver and one of the inputs:

- the input and receiver implement one of the Raw methods, and
- the address ranges of the backing data slices overlap, and
- the strides differ or there is an overlap in the data elements.

 If such an overlap is detected, the method will panic.

The following cases will not panic:

- the data slices do not overlap,
- there is pointer identity between the receiver and input values after
  the value has been untransposed if necessary.
```

Some finessing of "there is an overlap in the data elements" may need to happen since in reality it is "there is an overlap in the data elements when the values are converted to a `*Dense`".

btracey:
This is fixed, right?

kortschak:
Closed by #325.

PR #325, opened 2017-12-05
mat: use generalised rectangular shadow check

kortschak:
This is eases shadow detection between the complete set of BLAS backing matrix types without combinatoric explosion. The second commit changes the `checkOverlap` method signatures to have all take `blas64.General` and then have conversion calls at the method call sites.

Using the changes here allows shadow detection for all BLAS types, including packed matrices.

Please take a look.

Updates #310.


kortschak:
@btracey *ping*

PR #331, opened 2017-12-07
travis: make import of math/rand fail build

kortschak:
Please take a look.


btracey:
LGTM, though we may have some comments that contain that string (still not bad to prevent that though)

kortschak:
We could do something with `go list` if it becomes necessary. I'm just looking at whether there is a good format that makes this easy.

kortschak:
This command outputs a list of packages that import "math/rand":

```
go list -f '{{$p := .}}{{range .Imports}}{{if eq . "math/rand"}}{{$p.ImportPath}}{{end}}{{end}}' ./...
```

We could use this instead, but the existing code gives the file name.

PR #333, opened 2017-12-07
stat: implement Poisson rejection sampling.

kczimm:


stat/distuv/poisson.go
```diff
@@ -65,28 +65,55 @@ func (p Poisson) Prob(x float64) float64 {
...
-	for {
-		prod *= rnd()
-		if prod <= exp {
-			return x
+	var sq, alxm float64
+	var em, t, y float64
+
+	var lg float64
+
+	if p.Lambda < 12.0 {
```
btracey: Not sure which is better, but Numpy has 10 here.

stat/distuv/poisson.go
```diff
@@ -65,28 +65,55 @@ func (p Poisson) Prob(x float64) float64 {
...
 	rnd := rand.Float64
 	if p.Source != nil {
 		rnd = p.Source.Float64
 	}
 
-	for {
-		prod *= rnd()
-		if prod <= exp {
-			return x
+	var sq, alxm float64
```
btracey: Create these inside the `if` when they're needed, not up here. (this goes for all of them, not just these two)

stat/distuv/poisson.go
```diff
@@ -65,28 +65,55 @@ func (p Poisson) Prob(x float64) float64 {
...
+		// Use direct method.
+		em = -1
+		t = 1.0
+		for {
+			em++
+			t *= rnd()
+			if t <= g {
+				break
+			}
+		}
```
btracey: `return em` here.

stat/distuv/poisson.go
```diff
@@ -65,28 +65,55 @@ func (p Poisson) Prob(x float64) float64 {
...
+	} else {
```
btracey: Then, no need for `else` branch

stat/distuv/poisson.go
```diff
@@ -65,28 +65,55 @@ func (p Poisson) Prob(x float64) float64 {
...
-		if prod <= exp {
-			return x
+	var sq, alxm float64
+	var em, t, y float64
+
+	var lg float64
+
+	if p.Lambda < 12.0 {
+		// Use direct method.
+		em = -1
```
btracey: `var em int`

stat/distuv/poisson.go
```diff
@@ -65,28 +65,55 @@ func (p Poisson) Prob(x float64) float64 {
...
+		t = 1.0
+		for {
+			em++
+			t *= rnd()
+			if t <= g {
+				break
+			}
```
btracey: `em++` here.

stat/distuv/poisson.go
kczimm: Would you rather Rand() return type `int` or a `float64` that is an integer?

btracey: `float64`

kczimm: If `em` is of type `int` it will need to go through a number of type conversions.

btracey: So, `var em float64` is better, sorry.

kczimm: No worries. I was just asking for confirmation. I assumed I was likely overlooking something but wanted to confirm.

kczimm:
I'm guessing I may have missed some feedback as it was originally intended. Please let me know what I let fall through this last commit and I will be sure it get it in the next one.

stat/distuv/poisson.go
btracey: I think this would be better as  `var em float64`  and then putting the `em++` after the `if` statement.

stat/distuv/poisson.go
```diff
@@ -66,26 +66,49 @@ func (p Poisson) Prob(x float64) float64 {
...
-	//  Luc Devroye (p504)
-	//  http://www.eirene.de/Devroye.pdf
-	x := 0.0
-	prod := 1.0
-	exp := math.Exp(-p.Lambda)
+	// NUMERICAL RECIPES IN C: THE ART OF SCIENTIFIC COMPUTING (ISBN 0-521-43108-5)
+	// p. 294
+	// <http://www.aip.de/groups/soe/local/numres/bookcpdf/c7-3.pdf>
+
+	g := math.Exp(-p.Lambda)
```
btracey: Like the previous PR, this should work in log space. Specifically, there's no need for `g`, and then we can have `rnd := rand.ExpFloat64`, `t += rnd()`, `if t >= p.Lambda`

kczimm: Do you want the rejection method to work in log space, as well?

stat/distuv/poisson.go
```diff
@@ -70,30 +70,30 @@ func (p Poisson) Rand() float64 {
...
+			t += rnd()
+			if t >= p.Lambda {
 				break
 			}
+			em++
 		}
 		return em
 	}
 	// Use rejection method.
+	rnd = rand.Float64
```
btracey: This also needs to use `Source` when appropriate.

PR #335, opened 2017-12-08
graph: change Edge interface to include ID method

kortschak:
This is a major breaking change in the graph packages.

Updates gonum/graph#109.

Please take a look.

### DIFF
--- a/graph/undirect.go
+++ b/graph/undirect.go
@@ -214,6 +214,16 @@ func (e EdgePair) To() Node {
 	return nil
 }
 
+// ID returns the ID of the first non-nil edge, or 0.
+func (e EdgePair) ID() int64 {
+	if e[0] != nil {
+		return e[0].ID()
+	} else if e[1] != nil {
+		return e[1].ID()
+	}
+	return 0
+}
+
 // WeightedEdgePair is an opposed pair of directed edges.
 type WeightedEdgePair struct {
 	EdgePair
--- a/graph/community/louvain_common.go
+++ b/graph/community/louvain_common.go
@@ -259,6 +259,7 @@ type edge struct {
 
 func (e edge) From() graph.Node { return e.from }
 func (e edge) To() graph.Node   { return e.to }
+func (e edge) ID() int64        { return 0 }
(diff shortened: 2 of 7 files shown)


graph/undirect.go
```diff
@@ -214,6 +214,16 @@ func (e EdgePair) To() Node {
...
 }
 
+// ID returns the ID of the first non-nil edge, or 0.
+func (e EdgePair) ID() int64 {
+	if e[0] != nil {
+		return e[0].ID()
+	} else if e[1] != nil {
+		return e[1].ID()
+	}
+	return 0
```
mewmew: How would the returned `0` be differentiated from `e[0].ID()` or `e[1].ID()` if either of those edges have the *valid* `0` edge ID? Will this be an issue? Is the `EdgePair` a valid edge if it contains two `nil` edges?

The other methods on EdgePair (e.g. `From`) return `nil` when both the edges are `nil`, and `nil` can be differentiated from valid nodes which are non-nil, whereas `0` is a valid edge ID.

kortschak:
>Would this change allow graph with self-loops?

There is currently no in principle restriction on self-edges, the choice is by policy, but yes, the multigraphs should allow self-edges.

graph/undirect.go
kortschak: The code here only deals with simple graphs. We may need to look into how some of the code that exists needs to be changed to cope with mutligraphs when they appear.

kortschak:
Something that we had not considered previously was using the same kind of approach that we now use for weighting edges; the equivalent here would have been to keep `Edge` as it is and add a new `MultiEdge` (or something with a less horrible name) with the `ID() int64` method. This would have had the advantage of not breaking the current API and not requiring the useless `ID` method for edges used in simple graphs, but the disadvantage of making interop between simple graphs and multi graphs require a wrapper adding an `ID` method.

Having merged this change and the rudeness that it brought, I'm reluctant to revert this and do that, but if @mewmew and @vladimir-ch think that that is a better approach, I'll go ahead and do it. I'm pretty evenly balanced on which is better even without considering the breakage.

Thoughts?

Issue #277, opened 2017-10-22
Adding a method to VectorDense for element-wise operations


fasmat:
Is there a way to apply an operation to every element of a vector without extracting the underlying slice and manually iterating over it? Something like:
```
func (v *VectorDense) Apply(func (float64) float64)
```
that applies the provided function to every element would allow vectors to be used like this for instance:
```
var v VectorDense 
// ... (populate with values)

// elementwise apply the sigmoid function
v.Apply(func(x float64) {
  return 1.0 / (1.0 - math.Exp(-x))
})
```

btracey:
I could be convinced otherwise, I'm not sure this is really that beneficial. It's generally more go-like to just write the loop. It's more legible, and often more performant.

```
for i := 0; i < v.Len(); i++ {
    v.SetVec(i, f(v.At(i,0))
}
```
doesn't seem that burdensome to me.

kortschak:
Thanks, I was thinking the same, but I often do about these kinds of helpers so having another voice makes me feel better.

Edenhofer:
I recently faced the same _"problem"_ and IMHO it would make a lot of sense to implement this method. For sure there is some performance to gain considering each execution of the loop is independent from its previous run.
At the very least it would be as useful as the already existing `Apply` method for matrices.

kortschak:
There is a significant difference in the complexity of ranging over all elements of a matrix and all elements of a vector.

Please understand that Gonum does not (unlike other libraries and languages) try to do everything for the user. We provide tools for building on, particularly where these tools are onerous to implement locally or difficult to implement correctly. Two ideas that are fairly central to Go design principles are ["Do Less. Enable More."](https://blog.golang.org/open-source) and [adding features adds benefit, but also costs, and these must weigh in favour of benefit for the change to be worth while](https://golang.org/doc/faq#Why_doesnt_Go_have_feature_X). I don't think this does since it is easy to write a loop that operates through the API (almost inlined by the compiler if called on `*VecDense` and inlined if you are prepared to work on the `blas64.Vector` type - these can be hidden in your own helper).

>For sure there is some performance to gain considering each execution of the loop is independent from its previous run.

Sorry, I don't understand what you mean here.

sbinet:
>> For sure there is some performance to gain considering each execution of the loop is independent from its previous run.
>
> Sorry, I don't understand what you mean here.

I guess what's implied here is SIMD (or a gang of goroutines, each working on an element of the vector: something similar to OpenMP)
[...]

PR #336, opened 2017-12-10
generalize VecDense Copy method and LU SolveVec method, add ApplySym and ApplyVec methods

dezmondgoff:
Pull requests associated with these two issued: #199 and #277 and some tests.


dezmondgoff:
I see what happened...I must have pulled your vectors branch rather than my own. Sorry about that.

PR #337, opened 2017-12-10
mat: generalize VecDense Copy method and LU SolveVec method; add ApplyVec and ApplySym

dezmondgoff:
Reference #277 

But considering the discussion elsewhere, maybe there should just be a group of functions that act on Mutable interfaces.

### DIFF
--- a/mat/lu.go
+++ b/mat/lu.go
@@ -46,7 +46,11 @@ func (lu *LU) updateCond(anorm float64, norm lapack.MatrixNorm) {
 		anorm = unorm * lnorm
 	}
 	v := lapack64.Gecon(norm, lu.lu.mat, anorm, work, iwork)
-	lu.cond = 1 / v
+	if v != 0 {
+		lu.cond = 1 / v
+	} else {
+		lu.cond = math.Inf(1)
+	}
 }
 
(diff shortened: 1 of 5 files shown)


mat/lu.go
```diff
@@ -46,7 +46,11 @@ func (lu *LU) updateCond(anorm float64, norm lapack.MatrixNorm) {
 		anorm = unorm * lnorm
 	}
 	v := lapack64.Gecon(norm, lu.lu.mat, anorm, work, iwork)
-	lu.cond = 1 / v
+	if v != 0 {
```
kortschak: Why?

mat/lu.go
```diff
@@ -366,8 +369,9 @@ func (lu *LU) SolveVec(v *VecDense, trans bool, b *VecDense) error {
 		t = blas.Trans
 	}
 	lapack64.Getrs(t, lu.lu.mat, vMat, lu.pivot)
-	if lu.cond > ConditionTolerance {
-		return Condition(lu.cond)
-	}
+	// (dezmondgoff) is this final check now redundant?
```
kortschak: This should not be left like this.

btracey:
Closing because, as Dan said, there's too much going on. Feel free to discuss specific changes before sending PRs.

PR #343, opened 2017-12-12
blas/gonum: add Zhpr2

vladimir-ch:


vladimir-ch:
Ah, I can see now what you meant about the order of commits displayed by GitHub. It does not match the actual order. Interesting.

kortschak:
Yes, I think github just keeps a date-ordered list here rather than a proper DAG, and the fact that an interactive rebase to revise history order does not alter author date.

kortschak:
Please remember to send a PR against netlib (otherwise I can do that). Also if you have a small moment, can you take a look at #345.

PR #351, opened 2017-12-14
mat: add vector helpers

kortschak:
I found this lying around due to another PR. I didn't remember writing it, but on thinking I vaguely remember having some discussion with @btracey about some things that appear here, so I'm sending it with a view to clarifying that. I imagine I was planning on adding this as a prelude to attacking more of #167.

Updates #167.

Please take a look.


kortschak:
So obviously I've either confabulated the discussion or it was on a related by distinct topic.

Issue #339, opened 2017-12-11
optimize: Hard to enforce happens before behavior in global optimization


btracey:
The current implementation of `GuessAndCheck` is subtly wrong.

In IterateGlobal, there is a mutex to make sure the current task gets the best location, and then returns this with MajorIteration. This doesn't actually behave as expected. The problem is that when this returns, there is no guarantee that `updateOptLoc` is called immediately. So, in particular, we can have TaskA that gets the current best location, and returns MajorIteration. Now, TaskB finds a better value, so it returns MajorIteration with that better value. Assume that now, `updateOptLoc` is called for TaskB, updating the optimum location to the best found. Then, TaskA is resumed, and `updateOptLoc` and the best overall value is overwritten with the value in TaskB. 

I don't know if there is a better way to structure on the Global side, but if there isn't we need to document that the Method needs to synchronize calls to MajorIteration itself. For example, this is not a problem for CMA-ES, since the MajorIteration uses are synchronized (when the population finishes).

I'm working on a third optimizer at the moment that will try to implement this synchronization (how I found this bug), and we can see how tedious the fix is.

btracey:
This is worse than I thought. CMA-ES does try to synchronize, but incorrectly. It's entirely possible that

- TaskA updates, giving a list of new locations and returning MajorIteration 
- Tasks B-E evaluate all of the elements in the new list. 
- TaskB updates, giving a list of new locations and returning MajorIteration
- TaskB has `updateOptLoc` called
- TaskA has `updateOptLoc` called

I think we need to enforce some kind of FIFO order on `MajorIteration` at least.

btracey:
This was fixed with #352

Issue #344, opened 2017-12-12
optimize: Redesign the Global method interface and calling convention


btracey:
I realized the current design is broken trying to implement a new algorithm (see: #339 #340 #341 ). I made a writeup with a more full explanation why, and I'm posting it here as kind of a tracking issue.

Redoing the Global Optimization Scheme

# Introduction

## Calling routines
The optimize package has two main calling routines. The first is `Local`, which implements methods that do a search for a local minumum. These searches usually update parameters after every function evaluation. For example, in a quasi-Newton method, each new function evaluation updates the Hessian estimate and provides a new search direction. As local methods are usually inherently serial (or mostly serial), `optimize` provides no option for parallel function evaluations.

The second routine is `Global`, which seeks to find the global optimum of a function. Unlike local routines, global routines can frequently evaluate multiple locations before needing to pause to assess where the next batch of evaluations should be. In many optimization problems, evaluating the objective function is the dominant cost, and so it is highly beneficial to enable concurrent function evaluations within the interface.

## Reverse Communication
The other important piece is that `optimize` uses a "reverse communication" interface. In some optimization libraries, each method is effectively a full program, each with its own possible options. In `optimize`, we instead use `Local` and `Global` as driver scripts. The driver scripts are responsible for evaluating the actual objective function, and for checking the optimization settings (i.e. function evaluation limit). The actual optimization method is only responsible for returning the next location to evaluate.

This approach has a number of important advantages. The first is consistency. All optimization methods (including those implemented outside of `optimize`) have the same settings, and as the code is shared the settings are implemented correctly (if any of them are). Importantly, since the methods can not evaluate the function directly (as the information is not passed to them), we do not rely on the method to perform correct reporting of function evaluations (and gradient evaluations, etc.). 

## Operation
[...]

Issue #341, opened 2017-12-11
optimize: Done have a task ID?


btracey:
I think the `Done()` signature probably needs a task ID. Right now, in something like GuessAndCheck, the last function evaluation can never be sent to MajorIteration, because the limit will always be hit. Maybe we need to give control to the optimizer when `Done` happens in order to allow cleanup, but at that point `Global` refuses to do any more evaluations. 

This is somewhat related to the issue in #340 , and could help make the behavior more clear.

btracey:
This was fixed in a different way with #352

Issue #340, opened 2017-12-11
optimize: Can CMA-ES bleed goroutines?


btracey:
I'm a little confused about the interaction of the garbage collector and goroutines.

In `IterateGlobal` there is a `WaitGroup` to make sure that the whole list of evaluations is finished before the parameters can be updated. However, we can also stop because of function evaluations. Let's say that the number of allowed function evaluations is 199, and there are 200 functions in the list. Let's also say `nTasks = 8`. If the first 7 hit that waitgroup because there are no more function evaluations, but then the 8th one hits the function evaluation limit, and so it never calls IterateGlobal again, are the other goroutines waiting forever on that waitgroup?

btracey:
This was fixed with #352

PR #352, opened 2017-12-14
optimize: completely overhaul Global

btracey:
The previous implementation of Global was a minefield for incorrectly implementing global optimization methods. It was very difficult to correctly implement methods (both of the provided methods were incorrect), and the resulting code is very ugly. This commit switches to use channels to communicate, allowing a more clear ordering of concurrent code. This also enables better shutdown of methods.

In addition to the main fix of Global, this refactors the two Global methods to use the updated interface, and makes some small improvements that were previously not possible. In addition, there are some small cleanups of Local to better match between the two calls.

If anyone has been curious about what is meant by 'Don't communicate by sharing memory, share memory by communicating' this is it, and why.

### DIFF
--- a/optimize/cmaes.go
+++ b/optimize/cmaes.go
@@ -107,7 +107,7 @@ type CmaEsChol struct {
 	bestF float64
 
 	// Synchronization.
-	taskIdx     int
+	sentIdx     int
 	receivedIdx int
 	operation   chan<- GlobalTask
 	updateErr   error
@@ -233,7 +233,7 @@ func (cma *CmaEsChol) InitGlobal(dim, tasks int) int {
 	cma.bestX = resize(cma.bestX, dim)
 	cma.bestF = math.Inf(1)
 
-	cma.taskIdx = 0
+	cma.sentIdx = 0
 	cma.receivedIdx = 0
 	cma.operation = nil
 	cma.updateErr = nil
@@ -245,7 +245,7 @@ func (cma *CmaEsChol) sendInitTasks(tasks []GlobalTask) {
 	for i, task := range tasks {
 		cma.sendTask(i, task)
 	}
-	cma.taskIdx = len(tasks)
+	cma.sentIdx = len(tasks)
 }
 
 // sendTask generates a sample and sends the task. It does not update the cma index.
@@ -266,7 +266,8 @@ func (cma *CmaEsChol) bestIdx() int {
 		if math.IsNaN(v) {
 			continue
 		}
-		if v < bestVal {
+		// Use equality in case somewhere evaluates to +inf.
+		if v <= bestVal {
 			best = i
 			bestVal = v
 		}
@@ -280,8 +281,12 @@ func (cma *CmaEsChol) findBestAndUpdateTask(task GlobalTask) GlobalTask {
 	// Find and update the best location.
 	// Don't use floats because there may be NaN values.
 	best := cma.bestIdx()
-	bestF := cma.fs[best]
-	bestX := cma.xs.RawRowView(best)
+	bestF := math.NaN()
+	bestX := cma.xs.RawRowView(0)
+	if best != -1 {
+		bestF = cma.fs[best]
+		bestX = cma.xs.RawRowView(best)
+	}
 	if cma.ForgetBest {
 		task.F = bestF
 		copy(task.X, bestX)
@@ -292,52 +297,49 @@ func (cma *CmaEsChol) findBestAndUpdateTask(task GlobalTask) GlobalTask {
 		}
 		task.F = cma.bestF
 		copy(task.X, cma.bestX)
-
 	}
 	return task
 }
 
-func (cma *CmaEsChol) RunGlobal(operation chan<- GlobalTask, result <-chan GlobalTask, tasks []GlobalTask) {
-	cma.operation = operation
+func (cma *CmaEsChol) RunGlobal(operations chan<- GlobalTask, results <-chan GlobalTask, tasks []GlobalTask) {
+	cma.operation = operations
 	// Send the initial tasks. We know there are at most as many tasks as elements
 	// of the population.
 	cma.sendInitTasks(tasks)
-	cma.taskIdx = len(tasks)
 
-Outer:
+Loop:
 	for {
-		task := <-result
-		switch task.Op {
+		result := <-results
+		switch result.Op {
 		default:
 			panic("unknown operation")
 		case PostIteration:
-			break Outer
+			break Loop
 		case MajorIteration:
 			// The last thing we did was update all of the tasks and send the
 			// major iteration. Now we can send a group of tasks again.
 			cma.sendInitTasks(tasks)
-			cma.taskIdx = len(tasks)
 		case FuncEvaluation:
 			cma.receivedIdx++
-			cma.fs[task.ID] = task.F
+			cma.fs[result.ID] = result.F
 			switch {
-			case cma.taskIdx < cma.pop:
+			case cma.sentIdx < cma.pop:
 				// There are still tasks to evaluate. Send the next.
-				cma.sendTask(cma.taskIdx, task)
-				cma.taskIdx++
+				cma.sendTask(cma.sentIdx, result)
+				cma.sentIdx++
 			case cma.receivedIdx < cma.pop:
 				// All the tasks have been sent, but not all of them have been received.
 				// Need to wait until all are back.
-				continue Outer
+				continue Loop
 			default:
 				// All of the evaluations have been received.
 				if cma.receivedIdx != cma.pop {
 					panic("bad logic")
 				}
 				cma.receivedIdx = 0
-				cma.taskIdx = 0
+				cma.sentIdx = 0
 
-				task = cma.findBestAndUpdateTask(task)
+				task := cma.findBestAndUpdateTask(result)
 				// Update the parameters and send a MajorIteration or a convergence.
 				err := cma.update()
 				// Kill the existing data.
@@ -355,7 +357,7 @@ Outer:
 					task.Op = MajorIteration
 					task.ID = -1
 				}
-				operation <- task
+				operations <- task
 			}
 		}
 	}
@@ -363,7 +365,7 @@ Outer:
 	// Been told to stop. Clean up.
 	// Need to see best of our evaluated tasks so far. Should instead just
 	// collect, then see.
-	for task := range result {
+	for task := range results {
 		switch task.Op {
 		case MajorIteration:
 		case FuncEvaluation:
@@ -373,18 +375,20 @@ Outer:
 		}
 	}
 	// Send the new best value if the evaluation is better than any we've
-	// found so far.
+	// found so far. Keep this separate from findBestAndUpdateTask so that
+	// we only send an iteration if we find a better location.
 	if !cma.ForgetBest {
 		best := cma.bestIdx()
 		if best != -1 && cma.fs[best] < cma.bestF {
 			task := tasks[0]
-			task = cma.findBestAndUpdateTask(task)
+			task.F = cma.fs[best]
+			copy(task.X, cma.xs.RawRowView(best))
 			task.Op = MajorIteration
 			task.ID = -1
-			operation <- task
+			operations <- task
 		}
 	}
-	close(operation)
+	close(operations)
 }
 
 // update computes the new parameters (mean, cholesky, etc.). Does not update
--- a/optimize/cmaes.go
+++ b/optimize/cmaes.go
@@ -7,7 +7,6 @@ package optimize
 import (
 	"math"
 	"sort"
-	"sync"
 
 	"golang.org/x/exp/rand"
 
@@ -103,15 +102,15 @@ type CmaEsChol struct {
 	mean     []float64
 	chol     mat.Cholesky
 
-	// Parallel fields.
-	mux      sync.Mutex     // protect access to evals.
-	wg       sync.WaitGroup // wait for simulations to finish before iterating.
-	taskIdxs []int          // Stores which simulation the task ran.
-	evals    []int          // remaining evaluations in this iteration.
-
 	// Overall best.
 	bestX []float64
 	bestF float64
+
+	// Synchronization.
+	taskIdx     int
+	receivedIdx int
+	operation   chan<- GlobalTask
+	updateErr   error
 }
 
 var (
@@ -123,21 +122,26 @@ func (cma *CmaEsChol) Needs() struct{ Gradient, Hessian bool } {
 	return struct{ Gradient, Hessian bool }{false, false}
 }
 
-func (cma *CmaEsChol) Done() {}
-
-// Status returns the status of the method.
-func (cma *CmaEsChol) Status() (Status, error) {
+func (cma *CmaEsChol) methodConverged() Status {
 	sd := cma.StopLogDet
 	switch {
 	case math.IsNaN(sd):
-		return NotTerminated, nil
+		return NotTerminated
 	case sd == 0:
 		sd = float64(cma.dim) * -36.8413614879 // ln(1e-16)
 	}
 	if cma.chol.LogDet() < sd {
-		return MethodConverge, nil
+		return MethodConverge
 	}
-	return NotTerminated, nil
+	return NotTerminated
+}
+
+// Status returns the status of the method.
+func (cma *CmaEsChol) Status() (Status, error) {
+	if cma.updateErr != nil {
+		return Failure, cma.updateErr
+	}
+	return cma.methodConverged(), nil
 }
 
 func (cma *CmaEsChol) InitGlobal(dim, tasks int) int {
@@ -226,90 +230,167 @@ func (cma *CmaEsChol) InitGlobal(dim, tasks int) int {
 		cma.chol = chol
 	}
 
-	cma.evals = make([]int, cma.pop)
-	for i := range cma.evals {
-		cma.evals[i] = i
-	}
-
 	cma.bestX = resize(cma.bestX, dim)
 	cma.bestF = math.Inf(1)
 
+	// New ones TODO: reorient
+	cma.taskIdx = 0
+	cma.receivedIdx = 0
+	cma.operation = nil
+	cma.updateErr = nil
 	t := min(tasks, cma.pop)
-	cma.taskIdxs = make([]int, t)
-	for i := 0; i < t; i++ {
-		cma.taskIdxs[i] = -1
-	}
-	// Get a new mutex and waitgroup so that if the structure is reused there
-	// aren't residual interactions with the previous optimization.
-	cma.mux = sync.Mutex{}
-	cma.wg = sync.WaitGroup{}
 	return t
 }
 
-func (cma *CmaEsChol) IterateGlobal(task int, loc *Location) (Operation, error) {
-	// Check the status of the incoming task. If it is a number, it means
-	// that task contains a valid location.
-	idx := cma.taskIdxs[task]
-	if idx != -1 {
-		cma.fs[idx] = loc.F
-		cma.wg.Done()
+func (cma *CmaEsChol) sendInitTasks(tasks []GlobalTask) {
+	for i, task := range tasks {
+		cma.sendTask(i, task)
 	}
+	cma.taskIdx = len(tasks)
+}
 
-	// Get the next task and send it to be run if there is a next task to be run.
-	// If all of the tasks have been run, perform an update step. Note that the
-	// use of this mutex means that only one task can proceed, all of the
-	// other tasks should get stuck and then get a new location.
-	cma.mux.Lock()
-	if len(cma.evals) != 0 {
-		// There are still tasks to evaluate. Grab one and remove it from the list.
-		newIdx := cma.evals[len(cma.evals)-1]
-		cma.evals = cma.evals[:len(cma.evals)-1]
-		cma.wg.Add(1)
-		cma.mux.Unlock()
+// sendTask generates a sample and sends the task. It does not update the cma index.
+func (cma *CmaEsChol) sendTask(idx int, task GlobalTask) {
+	task.Index = idx
+	task.Operation = FuncEvaluation
+	distmv.NormalRand(cma.xs.RawRowView(idx), cma.mean, &cma.chol, cma.Src)
+	copy(task.X, cma.xs.RawRowView(idx))
+	cma.operation <- task
+}
 
-		// Sample x and send it to be evaluated.
-		distmv.NormalRand(cma.xs.RawRowView(newIdx), cma.mean, &cma.chol, cma.Src)
-		copy(loc.X, cma.xs.RawRowView(newIdx))
-		cma.taskIdxs[task] = newIdx
-		return FuncEvaluation, nil
+// bestIdx returns the best index in the functions. Returns -1 if all values
+// are NaN.
+func (cma *CmaEsChol) bestIdx() int {
+	best := -1
+	bestVal := math.Inf(1)
+	for i, v := range cma.fs {
+		if math.IsNaN(v) {
+			continue
+		}
+		if v < bestVal {
+			best = i
+			bestVal = v
+		}
 	}
-	// There are no more tasks to evaluate. This means the iteration is over.
-	// Find the best current f, update the parameters, and re-establish
-	// the evaluations to run.
+	return best
+}
 
-	// Wait for all of the outstanding tasks to finish, so the full set of functions
-	// has been evaluated.
-	cma.wg.Wait()
-
-	// Find the best f out of all the tasks.
-	best := floats.MinIdx(cma.fs)
+// findBestAndUpdateTask finds the best task in the current list, updates the
+// new best overall, and then stores the best location into task.
+func (cma *CmaEsChol) findBestAndUpdateTask(task GlobalTask) GlobalTask {
+	// Find and update the best location.
+	// Don't use floats because there may be NaN values.
+	best := cma.bestIdx()
 	bestF := cma.fs[best]
 	bestX := cma.xs.RawRowView(best)
 	if cma.ForgetBest {
-		loc.F = bestF
-		copy(loc.X, bestX)
+		task.F = bestF
+		copy(task.X, bestX)
 	} else {
 		if bestF < cma.bestF {
 			cma.bestF = bestF
 			copy(cma.bestX, bestX)
 		}
-		loc.F = cma.bestF
-		copy(loc.X, cma.bestX)
+		task.F = cma.bestF
+		copy(task.X, cma.bestX)
+
 	}
-
-	cma.taskIdxs[task] = -1
-
-	// Update the parameters of the distribution
-	err := cma.update()
-
-	// Reset the tasks
-	cma.evals = cma.evals[:cma.pop]
-
-	cma.mux.Unlock()
-	return MajorIteration, err
+	return task
 }
 
-// update computes the new parameters (mean, cholesky, etc.)
+func (cma *CmaEsChol) RunGlobal(operation chan<- GlobalTask, result <-chan GlobalTask, tasks []GlobalTask) {
+	cma.operation = operation
+	// Send the initial tasks. We know there are at most as many tasks as elements
+	// of the populaiton.
+	cma.sendInitTasks(tasks)
+	cma.taskIdx = len(tasks)
+
+Outer:
+	for {
+		task := <-result
+		switch task.Operation {
+		default:
+			panic("unknown operation")
+		case PostIteration:
+			break Outer
+		case MajorIteration:
+			// The last thing we did was update all of the tasks and send the
+			// major iteration. Now we can send a group of tasks again.
+			cma.sendInitTasks(tasks)
+			cma.taskIdx = len(tasks)
+		case FuncEvaluation:
+			// fmt.Println("fun received", cma.taskIdx, cma.receivedIdx)
+			cma.receivedIdx++
+			cma.fs[task.Index] = task.F
+			switch {
+			case cma.taskIdx < cma.pop:
+				// There are still tasks to evaluate. Send the next.
+				cma.sendTask(cma.taskIdx, task)
+				cma.taskIdx++
+			case cma.receivedIdx < cma.pop:
+				// All the tasks have been sent, but not all of them have been received.
+				// Need to wait until all are back.
+				continue Outer
+			default:
+				// All of the evaluations have been received.
+				if cma.receivedIdx != cma.pop {
+					panic("bad logic")
+				}
+				cma.receivedIdx = 0
+				cma.taskIdx = 0
+
+				task = cma.findBestAndUpdateTask(task)
+				// Update the parameters and send a MajorIteration or a convergence.
+				err := cma.update()
+				// Kill the existing data.
+				for i := range cma.fs {
+					cma.fs[i] = math.NaN()
+					cma.xs.Set(i, 0, math.NaN())
+				}
+				switch {
+				case err != nil:
+					cma.updateErr = err
+					task.Operation = MethodDone
+				case cma.methodConverged() != NotTerminated:
+					task.Operation = MethodDone
+				default:
+					task.Operation = MajorIteration
+					task.Index = -1
+				}
+				operation <- task
+			}
+		}
+	}
+
+	// Been told to stop. Clean up.
+	// Need to see best of our evaluated tasks so far. Should instead just
+	// collect, then see.
+	for task := range result {
+		switch task.Operation {
+		case MajorIteration:
+		case FuncEvaluation:
+			cma.fs[task.Index] = task.F
+		default:
+			panic("unknown operation")
+		}
+	}
+	// Send the new best value if the evaluation is better than any we've
+	// found so far.
+	if !cma.ForgetBest {
+		best := cma.bestIdx()
+		if best != -1 && cma.fs[best] < cma.bestF {
+			task := tasks[0]
+			task = cma.findBestAndUpdateTask(task)
+			task.Operation = MajorIteration
+			task.Index = -1
+			operation <- task
+		}
+	}
+	close(operation)
+}
+
+// update computes the new parameters (mean, cholesky, etc.). Does not update
+// any of the synchronization parameters (taskIdx).
 func (cma *CmaEsChol) update() error {
 	// Sort the function values to find the elite samples.
 	ftmp := make([]float64, cma.pop)
--- a/optimize/cmaes.go
+++ b/optimize/cmaes.go
@@ -233,7 +233,6 @@ func (cma *CmaEsChol) InitGlobal(dim, tasks int) int {
 	cma.bestX = resize(cma.bestX, dim)
 	cma.bestF = math.Inf(1)
 
-	// New ones TODO: reorient
 	cma.taskIdx = 0
 	cma.receivedIdx = 0
 	cma.operation = nil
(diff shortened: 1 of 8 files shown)


btracey:
Fixes #339 #340 #341 #344

btracey:
PTAL @vladimir-ch @sbinet 

Sorry for the huge PR , but a major overhaul was needed, and it's impossible to do piecemeal. This new code is not only correct, but MUCH cleaner and much easier to write new algorithms.

vladimir-ch:
I'll need time to go through this.

optimize/minimize.go
```diff
@@ -13,6 +13,18 @@ import (
 	"gonum.org/v1/gonum/mat"
 )
 
+var (
+	nonpositiveDimension string = "optimize: non-positive input dimension"
```
sbinet: I would call them `errNonPositiveDimension` and `errNegativeTasks`.
and create them with `errors.New(...)`

optimize/global.go
```diff
@@ -10,106 +10,74 @@ import (
...
-// and consistent.
-//
-// IterateGlobal must not return InitIteration and PostIteration operations. These are
-// reserved for the clients to be passed to Recorders. A Method must also not
-// combine the Evaluation operations with the Iteration operations.
+// GlobalTask is a task to communicate between the GlobalMethod and
+type GlobalTask struct {
+	Index int
+	Operation
+	*Location
```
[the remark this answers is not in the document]
btracey: Changed `Operation` to be a field rather than a method. `Location` doesn't have methods, but it does have fields, and so this allows writing `task.X` instead of `task.Loc.X`. I can change Location if you want too, it just seemed clearer this way.

optimize/global.go
```diff
@@ -10,106 +10,74 @@ import (
...
-//
-// When IterateGlobal declares MajorIteration, the caller updates the optimal
-// location to the values in Location, and checks for convergence. The type
-// implementing GlobalMethod must make sure that the fields of Location are valid
-// and consistent.
-//
-// IterateGlobal must not return InitIteration and PostIteration operations. These are
-// reserved for the clients to be passed to Recorders. A Method must also not
-// combine the Evaluation operations with the Iteration operations.
+// GlobalTask is a task to communicate between the GlobalMethod and
```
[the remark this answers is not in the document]
btracey: Done.

optimize/global.go
```diff
@@ -10,106 +10,74 @@ import (
...
-// documentation of Result for more information.
-//
-// Be aware that the default behavior of Global is to find the minimum.
-// For certain functions and optimization methods, this process can take many
-// function evaluations. The Settings input struct can be used to limit this,
-// for example by modifying the maximum runtime or maximum function evaluations.
-//
-// Global cannot guarantee strict adherence to the bounds specified in Settings
-// when performing concurrent evaluations and updates.
 func Global(p Problem, dim int, settings *Settings, method GlobalMethod) (*Result, error) {
```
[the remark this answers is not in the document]
btracey: Agreed. I'm not sure how that got deleted.

optimize/global.go
```diff
@@ -184,158 +140,195 @@ func minimizeGlobal(p *Problem, method GlobalMethod, settings *Settings, stats *
...
-	g.mux.RLock()
-	defer g.mux.RUnlock()
-	status = g.status
-	return status
-}
+	operations := make(chan GlobalTask, nTasks) // GlobalMethod sends tasks.
+	results := make(chan GlobalTask, nTasks)    // return results to GlobalMethod.
+	workerChan := make(chan GlobalTask)         // Delegate tasks to the workers.
+	evalStatsChan := make(chan GlobalTask)      // Send evaluation updates.
+	//iterStatsChan := make(chan GlobalTask)      // Send iteration updates.
```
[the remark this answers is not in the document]
btracey: Done.

optimize/global.go
```diff
@@ -184,158 +140,195 @@ func minimizeGlobal(p *Problem, method GlobalMethod, settings *Settings, stats *
...
+	statusChan := make(chan final)    // Send a termination status.
+	statusSent := make(chan struct{}) // Communicate the optimizaiton is done.
```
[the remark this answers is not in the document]
btracey: Done.

optimize/global.go
```diff
@@ -184,158 +140,195 @@ func minimizeGlobal(p *Problem, method GlobalMethod, settings *Settings, stats *
...
+	//statsDone := make(chan struct{})            // No further stats to update.
```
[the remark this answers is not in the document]
btracey: Done.

optimize/minimize.go
kortschak: These should be constants.

optimize/types.go
```diff
@@ -26,10 +26,12 @@ const (
 	// NoOperation specifies that no evaluation or convergence check should
 	// take place.
 	NoOperation Operation = 0
+	// signal is used interally to signal completion.
+	signalOperation Operation = 1 << (iota - 1)
```
vladimir-ch: This does not seem to be used. Apart from that, having non-exported operations would be a strange idea.

optimize/guessandcheck.go
```diff
@@ -27,34 +23,63 @@ func (g *GuessAndCheck) Needs() struct{ Gradient, Hessian bool } {
...
+	}
+
+	// Read from the channel until PostIteration is sent.
+Outer:
+	for {
+		task := <-result
+		switch task.Operation {
+		default:
+			panic("unknown operation")
+		case PostIteration:
```
vladimir-ch: This is mis/abusing `PostIteration` for a purpose for which it is not intended or documented. I have not read the diff completely yet but this looks something to think about.

optimize/cmaes.go
```diff
@@ -226,90 +230,167 @@ func (cma *CmaEsChol) InitGlobal(dim, tasks int) int {
...
+			panic("unknown operation")
+		case PostIteration:
+			break Outer
+		case MajorIteration:
+			// The last thing we did was update all of the tasks and send the
+			// major iteration. Now we can send a group of tasks again.
+			cma.sendInitTasks(tasks)
+			cma.taskIdx = len(tasks)
+		case FuncEvaluation:
+			// fmt.Println("fun received", cma.taskIdx, cma.receivedIdx)
```
vladimir-ch: Delete.

vladimir-ch:
Just gave it a quick pass, this will take me some time.

optimize/guessandcheck.go
btracey: You're right that the documentation for `PostIteration` needs to be updated, but the intended new behavior is documented https://github.com/gonum/gonum/pull/352/files#diff-a7f524fa342e38670f472a79c2779b82R57

btracey:
Thanks for starting to take a look. I'd recommend starting at Global, and then at GuessAndCheck. Sorry that the PR is so large, but this new design is much better than the old design (even considering the old design was incorrect). Happy to discuss if there are ways to simplify.

If we like this new design, I think we may even be able to combine this with Local to just have one implementation.

vladimir-ch:
So I've gone through `Global` and `GuessAndCheck` completely and have been pondering the design for the past couple of days. On the highest level, I like and completely agree with the new design (`GlobalMethod -> distributor -> (worker) -> stats combiner` communicating through channels). Using `PostIteration` in the new context also makes sense.

Before I get to commenting on the code level, I wonder whether the terminating logic isn't unnecessarily complicated. At least for me it's difficult to reason about it with confidence. I'm sure you've given it much thought but still I'd like to confirm: After the distributor gets notified that PostIteration has been sent, it could close the workerChan and start ignoring evaluation operations instead of terminating. This would enable it to range over `operations` (instead of breaking Outer) and close evalStatsChan on return. It would also eliminate the need for `distWG` and `statWG` and the stats combiner would not need to run in a go routine. The range loop over `evalStatsChan` would become the main control loop. Am I missing something?

btracey:
I really did try to have the stats combiner be the outer loop, and I eventually decided it wouldn't work. I'm not saying it can't work, but I remember it being trickier than I thought it would be. I believe the problem with your suggestion was deadlock, in coordinating between the distributor and statsCombiner, as well as making all of the workers shut down. I feel like the problem was establishing a HappensBefore relationship when no more evaluations would happen. That is, it's hard to synchronize when the statsCombiner triggers the ending, and where the operations actually stop being sent.

It does look, however, that a lot of that should be able to happen in the `<-statusSent` case of the distributor, which may be able to eliminate one of the goroutines and one waitgroup. I should have better commented what the difficulty was, but I'll give that alternative a try.

vladimir-ch:
`distWG` can be avoided by moving
```
	close(workerChan)
	// Read the final operations, only performing an action if MajorIteration.
	for task := range operations {
		switch task.Operation {
		case MajorIteration:
			evalStatsChan <- task
		}
	}
	return b
	// All stats have been sent. Close the channel and wait until all are updated.
	close(evalStatsChan)
```
into the distributor. This is actually quite nice because the distributor does not return after PostIteration, just its function changes in a way that is explicit. and closing `evalStatsChan` and `workerChan` is done in the one goroutine that writes to it. With this even `statWG` can be avoided by simply waiting on reading from `statusChan`. The end of `global` then becomes very simple:
```
	// Wait until a termination is sent.
	f := <-statusChan
	return f.Status, f.Err
}
```
The tests pass with these changes.

vladimir-ch:
Argh, the last part about statWG and the end of `global` is of course not true, we have to wait until the stats combiner finishes, not only that a status is sent:
```
	// Wait until a termination is sent.
	f := <-statusChan
	statWG.Wait()
	return f.Status, f.Err
}
```

optimize/cmaes.go
btracey: Done.

optimize/types.go
btracey: It is used (as you probably see now). We can re-assess if it needs to be after I make the fixes you suggest below.

btracey:
Implemented your idea. Thanks for the simplification. It still feels like we can move the distributor out of the goroutine, but the obvious transformation causes deadlocks. I'll try to take a look tomorrow to see if I can fix the issue, but for now at least that's simplified.

btracey:
Oh right, so here's the basic problem with moving the distributor out of a goroutine. We want to send exactly one "final answer". Let's say we change `case <- statusSent` in the distributor to `f := <- statusChan`. This causes a deadlock, because the distributor will likely be in a different case, say, the `MajorIteration` case, where it needs to send on `evalStatsChan`. So, the distributor is waiting to send on `evalStatsChan`, but the stats combiner is waiting to send on `statusSent`, hence deadlock. This is why I had to create the third outlet, and put the distributor in a goroutine.

 I think this could be improved by having `statusChan` be buffered. Our current semantics guarantee that that send happens before a new task is read by the stats combiner, and the switch semantics guarantee that the distributor checks `statusSent` before trying to read from `operations` again, so therefore there is a happens before relationship for shutting down.

kortschak:
Yes, this is what I was just playing with. It looks to me like the stats combiner does not need to be a go routine, and in fact that statusChan and statusSent are not required, though there is a little bit in the clean up that I'm not completely clear on.

BTW The fixes comment above will not be recognised by GitHub; they must be on separate lines.

btracey:
> BTW The fixes comment above will not be recognised by GitHub; they must be on separate lines.

Thanks. Do you know if I can fix it by editing, or do I have to post a new comment?

kortschak:
I think you can just edit it.

Arrgh, I though I had the buffered approach working, but now not. What would be lovely when this is done is a gonum.org blog post on the design.

btracey:
Ah. I misunderstood select semantics. The easy transformation using buffering (with me trying to move the distributor out of the goroutine) means that it's possible to send on that channel twice (with the single buffer), and so now there's a panic with trying to close the channel twice.

kortschak:
I have a solution without the stats combiner goroutine, but I can't post it right now.

btracey:
Sounds great, Mr. Fermat :-)

kortschak:
Here's my margin. I think it can be done without `statusChan`. If it can, I'll update this.

```diff
diff --git a/optimize/global.go b/optimize/global.go
index ea30fbd..9ac4991 100644
--- a/optimize/global.go
+++ b/optimize/global.go
@@ -6,7 +6,6 @@ package optimize
 
 import (
        "math"
-       "sync"
        "time"
 )
 
@@ -224,9 +223,8 @@ func minimizeGlobal(prob *Problem, method GlobalMethod, settings *Settings, stat
        results := make(chan GlobalTask, nTasks)    // return results to GlobalMethod.
        workerChan := make(chan GlobalTask)         // Delegate tasks to the workers.
        evalStatsChan := make(chan GlobalTask)      // Send evaluation updates.
-       statusChan := make(chan final)              // Send a termination status.
+       statusChan := make(chan final, 1)           // Send a termination status.
        statusSent := make(chan struct{})           // Communicate the optimization is done.
-       var statWG sync.WaitGroup                   // All stats are updated.
 
        // Launch the method.
        go func() {
@@ -250,77 +248,6 @@ func minimizeGlobal(prob *Problem, method GlobalMethod, settings *Settings, stat
                }()
        }
 
-       // Launch the stats combiner.
-       statWG.Add(1)
-       go func() {
-               defer statWG.Done()
-               var workerDone int // effective wg for the workers
-               var status Status
-               var err error
-               for task := range evalStatsChan {
-                       switch task.Operation {
-                       default:
-                               updateEvaluationStats(stats, task.Operation)
-                               status, err = checkEvaluationLimits(prob, stats, settings)
-                       case signalOperation:
-                               workerDone++
-                               if workerDone == nTasks {
-                                       close(results)
-                               }
-                               continue
-                       case NoOperation:
-                               // Just send the task back.
-                       case MajorIteration:
-                               status = performMajorIteration(optLoc, task.Location, stats, startTime, settings)
-                       case MethodDone:
-                               statuser, ok := method.(Statuser)
-                               if !ok {
-                                       panic("optimize: global method returned MethodDone is not a Statuser")
-                               }
-                               status, err = statuser.Status()
-                               if status == NotTerminated {
-                                       panic("optimize: global method returned MethodDone but a NotTerminated status")
-                               }
-                       }
-                       if settings.Recorder != nil && status == NotTerminated && err == nil {
-                               stats.Runtime = time.Since(startTime)
-                               // Allow err to be overloaded if the Recorder fails.
-                               err = settings.Recorder.Record(task.Location, task.Operation, stats)
-                               if err != nil {
-                                       status = Failure
-                               }
-                       }
-                       // If the optimization should be over, send the signal.
-                       if status != NotTerminated || err != nil {
-                               f := final{
-                                       Status: status,
-                                       Err:    err,
-                               }
-                               // Two cases: 1) A NotTerminated status has already been sent on
-                               // statusChan, and so we can send the first here. 2) A NotTerminated
-                               // status has already been sent. This means statusSent is closed
-                               // and can be received from.
-                               select {
-                               case <-statusSent:
-                               case statusChan <- f:
-                                       // Tell the distributor to quit, and send a PostIteration
-                                       close(statusSent)
-                                       results <- GlobalTask{
-                                               Operation: PostIteration,
-                                       }
-                               }
-                       }
-
-                       // If there are still workers remaining, send the result back.
-                       // Don't send MethodDone back.
-                       if task.Operation != MethodDone {
-                               if workerDone != nTasks {
-                                       results <- task
-                               }
-                       }
-               }
-       }()
-
        // Launch the distributor
        go func() {
                // Note: This cannot simply be a range loop over operations because we want to
@@ -354,11 +281,77 @@ func minimizeGlobal(prob *Problem, method GlobalMethod, settings *Settings, stat
                }
        }()
 
+       // Launch the stats combiner.
+       var workerDone int // effective wg for the workers
+       var status Status
+       var err error
+       for task := range evalStatsChan {
+               switch task.Operation {
+               default:
+                       updateEvaluationStats(stats, task.Operation)
+                       status, err = checkEvaluationLimits(prob, stats, settings)
+               case signalOperation:
+                       workerDone++
+                       if workerDone == nTasks {
+                               close(results)
+                       }
+                       continue
+               case NoOperation:
+                       // Just send the task back.
+               case MajorIteration:
+                       status = performMajorIteration(optLoc, task.Location, stats, startTime, settings)
+               case MethodDone:
+                       statuser, ok := method.(Statuser)
+                       if !ok {
+                               panic("optimize: global method returned MethodDone is not a Statuser")
+                       }
+                       status, err = statuser.Status()
+                       if status == NotTerminated {
+                               panic("optimize: global method returned MethodDone but a NotTerminated status")
+                       }
+               }
+               if settings.Recorder != nil && status == NotTerminated && err == nil {
+                       stats.Runtime = time.Since(startTime)
+                       // Allow err to be overloaded if the Recorder fails.
+                       err = settings.Recorder.Record(task.Location, task.Operation, stats)
+                       if err != nil {
+                               status = Failure
+                       }
+               }
+               // If the optimization should be over, send the signal.
+               if status != NotTerminated || err != nil {
+                       f := final{
+                               Status: status,
+                               Err:    err,
+                       }
+                       // Two cases: 1) A NotTerminated status has already been sent on
+                       // statusChan, and so we can send the first here. 2) A NotTerminated
+                       // status has already been sent. This means statusSent is closed
+                       // and can be received from.
+                       select {
+                       case <-statusSent:
+                       case statusChan <- f:
+                               // Tell the distributor to quit, and send a PostIteration
+                               close(statusSent)
+                               results <- GlobalTask{
+                                       Operation: PostIteration,
+                               }
+                       }
+               }
+
+               // If there are still workers remaining, send the result back.
+               // Don't send MethodDone back.
+               if task.Operation != MethodDone {
+                       if workerDone != nTasks {
+                               results <- task
+                       }
+               }
+       }
+
        // Wait until a termination is sent.
        f := <-statusChan
        finalStatus = f.Status
        finalError = f.Err
        // Wait for all the stats to be updated.
-       statWG.Wait()
        return finalStatus, finalError
 }
```

optimize/types.go
vladimir-ch: Yes, sorry, it was in the diff for global.go which is hidden unless clicked.

vladimir-ch:
@kortschak 's solution is what I was aiming for but didn't know how to get there. Very nice. In that diff, it is not necessary to put `status` and `err` into a `final`, meaning that `final` type can also be removed, simplifying the code further.

I think the next conceptual change should be the removal of `signalOperation` although I could be convinced that it's not worth it.

btracey:
Sorry, I tried copying and pasting that diff into a file, and then tried `git apply`. I get a `corrupt patch at line 181`. I tried both with and without the first line starting at `index`, and I tried both with and without adding a newline to the end of that. Any ideas?

kortschak:
The output of `git diff` is not a formatted patch. If you want to apply a patch, it's easier if I just push that change as a commit to this branch. Is that OK? I'll also add in the change of removing the final type.

btracey:
Sgtm

kortschak:
Done

kortschak:
The question that arises is how many tasks will arrive on `evalStatsChan` after `statusSent` has been closed and how long these tasks take; what is the amount of time that the caller will have to wait cleaning up tasks before the `evalStatsChan` loop has reached the required result. The answer to this may suggest that the stats collector goroutine is actually the better solution. Is there likely to be a wait for completion?

vladimir-ch:
After getting `PostIteration` the method should switch to a shutdown mode and drain the results channel until it's closed. This will then close the `operations` channel and eventually end the `evalStatsChan` loop. So it all depends on the `results` channel getting closed which will happen when all workers finish. So your comment brings me to an idea: We cannot and don't want to terminate ongoing evaluations but perhaps it is a good idea not to start new evaluations which are already in `workerChan` because evaluations are where most time is spent. I think that even now the method cannot rely on all evaluations being done and getting the result from them.

btracey:
We cannot guarantee (because of concurrency) that no other function evaluations are started after convergence happens. We can try to limit it with some kind of select statement in the workers.

 >  I think that even now the method cannot rely on all evaluations being done and getting the result from them.

Just to make sure we're on the same page, the method cannot guarantee that any evaluation it sends out is actually evaluated. However, currently the method / algorithm as a whole does guarantee that any evaluation which is actually started can be processed. 

Right now we don't really actually guarantee that the shutdown happens at a reasonable time, we only guarantee it with high probability. Specifically, the distributor randomly selects one of its cases to evaluate, which hypothetically could be reading from `operations` many times in a row. I wonder if it's worth thinking about re-ordering so it must read from `statusSent` if it can.

vladimir-ch:
> We cannot guarantee (because of concurrency) that no other function evaluations are started after convergence happens. We can try to limit it with some kind of select statement in the workers.

I was referring to a situation when all workers are busy evaluating and there are more evaluation tasks waiting in `workChan`. However, since the number of tasks is limited and known, this cannot happen.

> Just to make sure we're on the same page, the method cannot guarantee that any evaluation it sends out is actually evaluated. However, currently the method / algorithm as a whole does guarantee that any evaluation which is actually started can be processed.

Yes, that is also my understanding.

I will start commenting on the code.

optimize/types.go
vladimir-ch: I think that for now we can leave it in, although after merging this PR I would like to return to this to see if it can be avoided. However, until then I would:
* rename it to `signalDone`
* move it to the end of Operations
* fix spelling of `internally` in the comment.

optimize/global.go
```diff
@@ -184,158 +177,169 @@ func minimizeGlobal(p *Problem, method GlobalMethod, settings *Settings, stats *
...
+	//   PostIteration to be sent to Method, and triggers a series of shutdown steps.
 
-		// Evaluate location and/or update stats.
-		status := g.globalOperation(op, loc, x)
-		if status != NotTerminated {
-			break
-		}
-	}
-}
+	var (
```
vladimir-ch: Move this to the variables declared before the `evalStatsChan` loop.

optimize/global.go
```diff
@@ -184,158 +177,169 @@ func minimizeGlobal(p *Problem, method GlobalMethod, settings *Settings, stats *
...
-	settings  *Settings
-	method    GlobalMethod
-	statuser  Statuser
-	err       error
-}
+	operations := make(chan GlobalTask, nTasks) // GlobalMethod sends tasks.
+	results := make(chan GlobalTask, nTasks)    // return results to GlobalMethod.
+	workerChan := make(chan GlobalTask)         // Delegate tasks to the workers.
+	evalStatsChan := make(chan GlobalTask)      // Send evaluation updates.
+	statusSent := make(chan struct{})           // Communicate the optimization is done.
```
vladimir-ch: I'd rename this to `done`.

optimize/global.go
```diff
@@ -184,158 +177,169 @@ func minimizeGlobal(p *Problem, method GlobalMethod, settings *Settings, stats *
...
-func (g *globalStatus) updateStatus(s Status, err error) {
-	g.mux.Lock()
-	defer g.mux.Unlock()
-	if s != NotTerminated {
-		g.status = s
-		g.err = err
+	// Launch the workers. After workerChan is closed, each worker signals
+	// to the stats updater that they have successfully terminated.
+	for worker := 0; worker < nTasks; worker++ {
+		go func() {
```
vladimir-ch: I'd assign the function literal to a variable and have the comment `After workerChan is closed ...` before it. Commenting `// Launch the workers.` would then be unnecessary.

optimize/global.go
```diff
@@ -184,158 +177,169 @@ func minimizeGlobal(p *Problem, method GlobalMethod, settings *Settings, stats *
...
 	}
-}
 
-func (g *globalStatus) finishIteration(status Status, err error, loc *Location, op Operation) (Status, error) {
-	g.mux.Lock()
-	defer g.mux.Unlock()
-	return finishIteration(status, err, g.stats, g.settings, g.statuser, g.startTime, loc, op)
-}
+	// Launch the distributor
+	go func() {
```
vladimir-ch: Same here, I'd assign the function to a variable and remove the comment.

optimize/global.go
```diff
@@ -184,158 +177,169 @@ func minimizeGlobal(p *Problem, method GlobalMethod, settings *Settings, stats *
...
+		// Note: This cannot simply be a range loop over operations because we want to
```
vladimir-ch: I'm not sure whether this comment is useful after the recent changes.

optimize/global.go
```diff
@@ -184,158 +177,169 @@ func minimizeGlobal(p *Problem, method GlobalMethod, settings *Settings, stats *
...
+		// still send to operation even after a termination Status has been sent.
+		for {
+			select {
+			case <-statusSent:
+				close(workerChan)
+				// Read the final operations, only performing an action if MajorIteration.
+				for task := range operations {
+					switch task.Operation {
```
vladimir-ch: This could be a simple `if`.

optimize/global.go
```diff
@@ -184,158 +177,169 @@ func minimizeGlobal(p *Problem, method GlobalMethod, settings *Settings, stats *
...
+					case MajorIteration:
+						evalStatsChan <- task
+					}
+				}
+				// All stats have been sent. Close the channel and wait until all are updated.
```
vladimir-ch: Is this comment still useful, I think the code is clear. At least, there is no wait here anymore.

optimize/global.go
```diff
@@ -184,158 +177,169 @@ func minimizeGlobal(p *Problem, method GlobalMethod, settings *Settings, stats *
...
+				close(evalStatsChan)
+				return
+			case task := <-operations: // Delegate the Operation.
```
vladimir-ch: Either remove the comment or separate it to its own line.

optimize/global.go
```diff
@@ -184,158 +177,169 @@ func minimizeGlobal(p *Problem, method GlobalMethod, settings *Settings, stats *
...
+				switch task.Operation {
+				case InitIteration:
+					panic("optimize: GlobalMethod returned InitIteration")
+				case PostIteration:
+					panic("optimize: GlobalMethod returned PostIteration")
+				case NoOperation, MajorIteration, MethodDone:
+					evalStatsChan <- task
+				default: // Any of the Evaluation operations.
+					workerChan <- task
```
vladimir-ch: I would add
```
if !task.Operation.isEvaluation() {
    panic
}
```
and remove the comment.

optimize/global.go
```diff
@@ -184,158 +177,169 @@ func minimizeGlobal(p *Problem, method GlobalMethod, settings *Settings, stats *
...
-// When modifying this function, keep in mind that it can be called concurrently.
-// Uses of the internal fields should be through the methods of globalStatus and
-// protected by a mutex where appropriate.
-func (g *globalStatus) globalOperation(op Operation, loc *Location, x []float64) Status {
-	// Do a quick check to see if one of the other workers converged in the meantime.
-	status := g.getStatus()
-	if status != NotTerminated {
-		return status
-	}
+	// Launch the stats combiner.
```
vladimir-ch: Either remove or rephrase, we are not launching a goroutine here anymore.

optimize/global.go
```diff
@@ -184,158 +177,169 @@ func minimizeGlobal(p *Problem, method GlobalMethod, settings *Settings, stats *
...
+			}
+			continue
+		case NoOperation:
+			// Just send the task back.
+		case MajorIteration:
+			status = performMajorIteration(optLoc, task.Location, stats, startTime, settings)
+		case MethodDone:
+			statuser, ok := method.(Statuser)
+			if !ok {
+				panic("optimize: global method returned MethodDone is not a Statuser")
```
vladimir-ch: Missing "but"? `global method returned MethodDone but is not Statuser`?

optimize/global.go
```diff
@@ -184,158 +177,169 @@ func minimizeGlobal(p *Problem, method GlobalMethod, settings *Settings, stats *
...
+			stats.Runtime = time.Since(startTime)
+			// Allow err to be overloaded if the Recorder fails.
+			err = settings.Recorder.Record(task.Location, task.Operation, stats)
+			if err != nil {
+				status = Failure
+			}
+		}
+		// If the optimization should be over, send the signal.
+		if status != NotTerminated || err != nil {
+			// Two cases: 1) A NotTerminated status has already been sent on
```
vladimir-ch: This comment should be updated to the current behavior or maybe even removed.

optimize/global.go
```diff
@@ -184,158 +177,169 @@ func minimizeGlobal(p *Problem, method GlobalMethod, settings *Settings, stats *
...
-func DefaultSettingsGlobal() *Settings {
-	return &Settings{
-		FunctionThreshold: math.Inf(-1),
-		FunctionConverge: &FunctionConverge{
-			Absolute:   1e-10,
-			Iterations: 100,
-		},
+		// If there are still workers remaining, send the result back.
+		// Don't send MethodDone back.
+		if task.Operation != MethodDone {
```
vladimir-ch:
```
if workerDone != nTasks && tast.Operation != MethodDone {
    ...
}
```

optimize/global.go
```diff
@@ -6,65 +6,76 @@ package optimize
...
-// and consistent.
-//
-// IterateGlobal must not return InitIteration and PostIteration operations. These are
-// reserved for the clients to be passed to Recorders. A Method must also not
-// combine the Evaluation operations with the Iteration operations.
+// GlobalTask is a type to communicate between the GlobalMethod and the outer
+// calling script.
+type GlobalTask struct {
+	Index     int
+	Operation Operation
```
vladimir-ch: `Op Operation`

btracey:
Took another pass at writing the documentation.

optimize/minimize.go
```diff
@@ -134,70 +117,77 @@ func updateEvaluationStats(stats *Stats, op Operation) {
...
 	}
-
 	if settings.HessEvaluations > 0 && stats.HessEvaluations >= settings.HessEvaluations {
-		return HessianEvaluationLimit
+		return HessianEvaluationLimit, nil
 	}
+	return NotTerminated, nil
+}
 
+// checkIterationLimits checks if any of the limits on iterations affected by
```
vladimir-ch: Something wrong with this sentence.

optimize/minimize.go
```diff
@@ -134,70 +117,77 @@ func updateEvaluationStats(stats *Stats, op Operation) {
...
+// MajorIteration.
+func checkIterationLimits(loc *Location, stats *Stats, settings *Settings) Status {
+	if settings.MajorIterations > 0 && stats.MajorIterations >= settings.MajorIterations {
+		return IterationLimit
+	}
 	// TODO(vladimir-ch): It would be nice to update Runtime here.
```
vladimir-ch: Could you please remove this TODO for me?

optimize/guessandcheck.go
```diff
@@ -27,34 +23,63 @@ func (g *GuessAndCheck) Needs() struct{ Gradient, Hessian bool } {
 	return struct{ Gradient, Hessian bool }{false, false}
 }
 
-func (g *GuessAndCheck) Done() {
-	// No cleanup needed
-}
-
 func (g *GuessAndCheck) InitGlobal(dim, tasks int) int {
```
vladimir-ch: `dim` and `tasks` should be checked as in `CmaEsChol`.

optimize/cmaes.go
```diff
@@ -226,90 +230,165 @@ func (cma *CmaEsChol) InitGlobal(dim, tasks int) int {
...
+func (cma *CmaEsChol) RunGlobal(operation chan<- GlobalTask, result <-chan GlobalTask, tasks []GlobalTask) {
+	cma.operation = operation
+	// Send the initial tasks. We know there are at most as many tasks as elements
+	// of the population.
+	cma.sendInitTasks(tasks)
+	cma.taskIdx = len(tasks)
+
+Outer:
+	for {
+		task := <-result
```
vladimir-ch: Seeing the implementation I think I would globally rename the `result` channel to `results`, and so `operation` to `operations`. Then here `result := <-results` and below `task := findBestAndUpdateTask(result)`.

optimize/cmaes.go
```diff
@@ -226,90 +230,165 @@ func (cma *CmaEsChol) InitGlobal(dim, tasks int) int {
...
 	} else {
 		if bestF < cma.bestF {
 			cma.bestF = bestF
 			copy(cma.bestX, bestX)
 		}
-		loc.F = cma.bestF
-		copy(loc.X, cma.bestX)
-	}
+		task.F = cma.bestF
+		copy(task.X, cma.bestX)
```
vladimir-ch: Delete empty line.

optimize/cmaes.go
```diff
@@ -226,90 +230,165 @@ func (cma *CmaEsChol) InitGlobal(dim, tasks int) int {
...
+		case MajorIteration:
+		case FuncEvaluation:
+			cma.fs[task.ID] = task.F
+		default:
+			panic("unknown operation")
+		}
+	}
+	// Send the new best value if the evaluation is better than any we've
+	// found so far.
+	if !cma.ForgetBest {
```
vladimir-ch: Up to here the code reads very smoothly but the logic of this final part is not clear. Steps done here and in `findBestAndUpdateTask` are very similar. Could you please summarize?

optimize/cmaes.go
```diff
@@ -103,15 +102,15 @@ type CmaEsChol struct {
...
-	wg       sync.WaitGroup // wait for simulations to finish before iterating.
-	taskIdxs []int          // Stores which simulation the task ran.
-	evals    []int          // remaining evaluations in this iteration.
-
 	// Overall best.
 	bestX []float64
 	bestF float64
+
+	// Synchronization.
+	taskIdx     int
```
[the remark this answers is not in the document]
btracey: Done.

optimize/cmaes.go
```diff
@@ -226,90 +230,165 @@ func (cma *CmaEsChol) InitGlobal(dim, tasks int) int {
...
+}
 
-	// Find the best f out of all the tasks.
-	best := floats.MinIdx(cma.fs)
+// findBestAndUpdateTask finds the best task in the current list, updates the
+// new best overall, and then stores the best location into task.
+func (cma *CmaEsChol) findBestAndUpdateTask(task GlobalTask) GlobalTask {
+	// Find and update the best location.
+	// Don't use floats because there may be NaN values.
+	best := cma.bestIdx()
```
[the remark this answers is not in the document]
btracey: Done.

optimize/cmaes.go
btracey: Done.

optimize/cmaes.go
```diff
@@ -226,90 +230,165 @@ func (cma *CmaEsChol) InitGlobal(dim, tasks int) int {
...
+}
 
-	// Reset the tasks
-	cma.evals = cma.evals[:cma.pop]
+func (cma *CmaEsChol) RunGlobal(operation chan<- GlobalTask, result <-chan GlobalTask, tasks []GlobalTask) {
+	cma.operation = operation
+	// Send the initial tasks. We know there are at most as many tasks as elements
+	// of the population.
+	cma.sendInitTasks(tasks)
+	cma.taskIdx = len(tasks)
```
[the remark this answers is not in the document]
btracey: Done.

optimize/cmaes.go
```diff
@@ -226,90 +230,165 @@ func (cma *CmaEsChol) InitGlobal(dim, tasks int) int {
...
+
+Outer:
```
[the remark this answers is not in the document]
btracey: Done.

optimize/cmaes.go
```diff
@@ -226,90 +230,165 @@ func (cma *CmaEsChol) InitGlobal(dim, tasks int) int {
...
+		switch task.Op {
+		default:
+			panic("unknown operation")
+		case PostIteration:
+			break Outer
+		case MajorIteration:
+			// The last thing we did was update all of the tasks and send the
+			// major iteration. Now we can send a group of tasks again.
+			cma.sendInitTasks(tasks)
+			cma.taskIdx = len(tasks)
```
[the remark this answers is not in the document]
btracey: Done.

optimize/cmaes.go
btracey: Done.

optimize/cmaes.go
btracey: Done.

vladimir-ch:
Have you pushed the changes?

btracey:
Sorry, didn't realize my push had failed. Now they are uploaded.

btracey:
No problem. It's a big PR and big PRs take time.

PR #359, opened 2017-12-21
mat: generalise basic arithmetic vector operations

kortschak:
Please take a look.

Updates #167.

### DIFF
--- a/mat/vector.go
+++ b/mat/vector.go
@@ -97,11 +97,11 @@ func NewVecDense(n int, data []float64) *VecDense {
 	}
 }
 
-// SliceVec returns a new VecDense that shares backing data with the receiver.
+// SliceVec returns a new Vector that shares backing data with the receiver.
 // The returned matrix starts at i of the receiver and extends k-i elements.
 // SliceVec panics with ErrIndexOutOfRange if the slice is outside the capacity
 // of the receiver.
-func (v *VecDense) SliceVec(i, k int) *VecDense {
+func (v *VecDense) SliceVec(i, k int) Vector {
 	if i < 0 || k <= i || v.Cap() < k {
 		panic(ErrIndexOutOfRange)
 	}
--- a/mat/vector.go
+++ b/mat/vector.go
@@ -568,12 +568,18 @@ func (v *VecDense) MulVec(a Matrix, b Vector) {
 	case RawSymmetricer:
 		if fast {
 			amat := aU.RawSymmetric()
+			// We don't know that a is a *SymDense, so make
+			// a temporary SymDense to check overlap.
+			(&SymDense{mat: amat}).checkOverlap(v.asGeneral())
 			blas64.Symv(1, amat, bmat, 0, v.mat)
 			return
 		}
 	case RawTriangular:
 		v.CopyVec(b)
 		amat := aU.RawTriangular()
+		// We don't know that a is a *TriDense, so make
+		// a temporary TriDense to check overlap.
+		(&TriDense{mat: amat}).checkOverlap(v.asGeneral())
 		ta := blas.NoTrans
 		if trans {
 			ta = blas.Trans
--- a/mat/vector.go
+++ b/mat/vector.go
@@ -196,32 +196,51 @@ func (v *VecDense) RawVector() blas64.Vector {
 // CopyVec makes a copy of elements of a into the receiver. It is similar to the
 // built-in copy; it copies as much as the overlap between the two vectors and
 // returns the number of elements it copied.
-func (v *VecDense) CopyVec(a *VecDense) int {
+func (v *VecDense) CopyVec(a Vector) int {
 	n := min(v.Len(), a.Len())
-	if v != a {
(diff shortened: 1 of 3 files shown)


kortschak:
`MulVec` is more difficult to do in a sensible way.

mat/vector.go
```diff
@@ -503,105 +503,118 @@ func (v *VecDense) DivElemVec(a, b Vector) {
...
 	case *VecDense:
-		if v != a {
-			v.checkOverlap(a.mat)
+		if v != aU {
+			v.checkOverlap(aU.mat)
 		}
 
-		if a.Len() == 1 {
+		if aU.Len() == 1 {
 			// {1,1} x {1,n}
```
vladimir-ch: `{1,1} x {n,1}`?

mat/vector.go
```diff
@@ -503,105 +503,118 @@ func (v *VecDense) DivElemVec(a, b Vector) {
...
+				return
+			}
 			for i := 0; i < b.Len(); i++ {
-				v.mat.Data[i*v.mat.Inc] = av * b.mat.Data[i*b.mat.Inc]
+				v.mat.Data[i*v.mat.Inc] = av * b.AtVec(i)
 			}
 			return
 		}
 		if b.Len() == 1 {
 			// {1,n} x {1,1}
```
vladimir-ch: `{n,1} x {1,1}`?

mat/vector.go
```diff
@@ -503,105 +503,118 @@ func (v *VecDense) DivElemVec(a, b Vector) {
...
-			bv := b.At(0, 0)
-			for i := 0; i < a.Len(); i++ {
-				v.mat.Data[i*v.mat.Inc] = bv * a.mat.Data[i*a.mat.Inc]
+			bv := b.AtVec(0)
+			for i := 0; i < aU.Len(); i++ {
+				v.mat.Data[i*v.mat.Inc] = bv * aU.mat.Data[i*aU.mat.Inc]
 			}
 			return
 		}
 		// {n,1} x {1,n}
```
vladimir-ch: `{1,n} x {n,1}`?

mat/vector.go
```diff
@@ -97,11 +97,11 @@ func NewVecDense(n int, data []float64) *VecDense {
 	}
 }
 
-// SliceVec returns a new VecDense that shares backing data with the receiver.
+// SliceVec returns a new Vector that shares backing data with the receiver.
 // The returned matrix starts at i of the receiver and extends k-i elements.
 // SliceVec panics with ErrIndexOutOfRange if the slice is outside the capacity
 // of the receiver.
-func (v *VecDense) SliceVec(i, k int) *VecDense {
+func (v *VecDense) SliceVec(i, k int) Vector {
```
vladimir-ch: Why this change?

kortschak: Interface satisifaction; if there are multiple `SliceVec` implementers they can only satisfy an interface if there is an equal signature. This is what we did with `Slice` on `Dense`. Maybe it needs rediscussion?

mat/vector.go
kortschak: Interesting. If columns of A is 1, rows of b must also be 1. Now we specify that `VecDense` is a column vectore, which means that here we cannot have `{1,1} x {1,n}` unless n is 1.

/cc @btracey

mat/vector.go
kortschak: Fixed

mat/vector.go
kortschak: Fixed

mat/vector.go
kortschak: I'm tightening this up, and as a second commit removing this case since it doesn't improve much over the fall through at the bottom.

mat/vector.go
```diff
@@ -242,247 +261,362 @@ func (v *VecDense) AddScaledVec(a *VecDense, alpha float64, b *VecDense) {
...
+			amat := aU.RawMatrix()
+			// We don't know that a is a *Dense, so make
+			// a temporary Dense to check overlap.
+			(&Dense{mat: amat}).checkOverlap(v.asGeneral())
+			t := blas.NoTrans
+			if trans {
+				t = blas.Trans
+			}
+			blas64.Gemv(t, 1, amat, bmat, 0, v.mat)
+			return
```
btracey: This needs a fallthrough, or other code for the non fast path.

mat/vector.go
```diff
@@ -242,247 +261,362 @@ func (v *VecDense) AddScaledVec(a *VecDense, alpha float64, b *VecDense) {
...
+		}
+	} else {
+		fast = false
 	}
 
 	v.reuseAs(ar)
 
 	switch {
 	case alpha == 0: // v <- a
+		if v == a {
```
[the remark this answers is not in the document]
kortschak: It has to come after the `reuseAs` call.

mat/vector.go
kortschak: Like the one after the switch?

btracey: Oh, sorry.

PR #361, opened 2017-12-26
mat: relax list test type restrictions for Vector parameters

kortschak:
Please take a look.

### DIFF
--- a/mat/list_test.go
+++ b/mat/list_test.go
@@ -169,7 +169,7 @@ func legalTypesSym(a, b Matrix) bool {
 	return true
 }
 
-// legalTypeVec returns whether v is a Vector.
+// legalTypeVector returns whether v is a Vector.
 func legalTypeVector(v Matrix) bool {
 	_, ok := v.(Vector)
 	return ok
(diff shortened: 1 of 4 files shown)


mat/list_test.go
```diff
@@ -169,14 +169,20 @@ func legalTypesSym(a, b Matrix) bool {
 	return true
 }
 
+// legalTypeVec returns whether v is a Vector.
```
vladimir-ch: legalTypeVector

mat/list_test.go
```diff
@@ -197,9 +203,16 @@ func legalTypesVecDenseVecDense(a, b Matrix) bool {
 	return true
 }
 
-// legalTypesNotVecVec returns whether the first input is an arbitrary Matrix
+// legalTypesNotVectorVector returns whether the first input is an arbitrary Matrix
+// and the second input is a Vector.
+func legalTypesNotVectorVector(a, b Matrix) bool {
```
vladimir-ch: How about renaming to legalTypesMatrixVector?

mat/list_test.go
```diff
@@ -197,9 +203,16 @@ func legalTypesVecDenseVecDense(a, b Matrix) bool {
...
+	_, ok := b.(Vector)
+	return ok
+}
+
+// legalTypesNotVectorVecDense returns whether the first input is an arbitrary Matrix
 // and the second input is a *VecDense.
-func legalTypesNotVecVec(a, b Matrix) bool {
+func legalTypesNotVectorVecDense(a, b Matrix) bool {
```
vladimir-ch: legalTypesMatrixVecDense?

mat/list_test.go
kortschak: Done

mat/list_test.go
kortschak: Thanks, I was looking for something like that.

mat/list_test.go
kortschak: Done

kortschak:
PTAL

PR #366, opened 2017-12-27
mat: generalise Inner vector parameters

kortschak:
Please take a look.

Updates #167.

### DIFF
--- a/mat/inner.go
+++ b/mat/inner.go
@@ -6,16 +6,17 @@ package mat
 
 import (
 	"gonum.org/v1/gonum/blas"
+	"gonum.org/v1/gonum/blas/blas64"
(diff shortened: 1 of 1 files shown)


mat/inner.go
```diff
@@ -36,66 +37,84 @@ func Inner(x *VecDense, A Matrix, y *VecDense) float64 {
 			// Panic as a string not a mat.Error.
```
vladimir-ch: I'd rather have `switch a := A.(type) {` and `amat` but it's outside of the diff scope and I think we usually use lower case even for matrix parameters. It's up to you.

kortschak: I'll do that in another PR. Yes, this capital parameter snuck through.

PR #372, opened 2017-12-31
simple: fix nil edge returns

kortschak:
Please take a look.


kortschak:
We have obviously only got some really crappy travis hardware to run on at the moment. These tests pass locally and in short time, but timeout either with 10min silence or with the global time limit.

kortschak:
I'll merge when travis-ci fix whatever it is that they're not doing right at the moment and I can get the tests to pass.

kortschak:
Woot. That was way harder than it should have been.

PR #374, opened 2018-01-02
mat: retain type and dimension attributes in encodings

### LINKED ISSUE #373: mat: add type annotation to binary encoding
See https://github.com/gonum/gonum/issues/319#issuecomment-352218292

/cc @sbinet

kortschak:
Please take a look.

Closes #373.


mat/io.go
```diff
@@ -29,22 +32,27 @@ var (
 // MarshalBinary encodes the receiver into a binary form and returns the result.
 //
 // Dense is little-endian encoded as follows:
-//   0 -  7  number of rows    (int64)
-//   8 - 15  number of columns (int64)
-//  16 - ..  matrix data elements (float64)
+//   0 -  1  'D', 'G' - type annotation (int8)
```
sbinet: We should probably document somewhere what these type annotations are and their complete list.

sbinet: Also, their expected order for the Dense case.

kortschak: Yeah, there isn't a complete list yet. I should probably go through blas64 and devise one.

Note that the dense case does explain the order, the 'D' and 'G' are not members of a set.

mat/io.go
```diff
@@ -20,6 +20,9 @@ var (
 	sizeInt64   = binary.Size(int64(0))
 	sizeFloat64 = binary.Size(float64(0))
 
+	errWrongType  = errors.New("mat: wrong data type")
```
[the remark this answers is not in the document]
kortschak: What would the user do with the information?

mat/io.go
```diff
@@ -68,8 +76,16 @@ func (m Dense) MarshalBinary() ([]byte, error) {
 func (m Dense) MarshalBinaryTo(w io.Writer) (int, error) {
 	var n int
 	var buf [8]byte
+	buf[0] = 'D'
```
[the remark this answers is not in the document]
kortschak: Marshaling does not mutate the receiver.

mat/io.go
```diff
@@ -198,23 +239,32 @@ func (m *Dense) UnmarshalBinaryFrom(r io.Reader) (int, error) {
...
 	return n, nil
 }
 
 // MarshalBinary encodes the receiver into a binary form and returns the result.
 //
 // VecDense is little-endian encoded as follows:
-//   0 -  7  number of elements     (int64)
-//   8 - ..  vector's data elements (float64)
+//
+//   0       'D' - type annotation  (int8)
```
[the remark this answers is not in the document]
kortschak: Does this extend to including vectors in that?

The scheme that looks like needs to exist is:

|Type|Form|Packing|Uplo|Unit|Rows|Columns|kU |kL |
|---|---|---|---|---|---|---|---|---|
| |`uint8` [GST]|`uint8` [BPF]|`uint8` [AUL]|`bool`|`int64`|`int64`|`int64`|`int64`|
|General|'G'|'F'|'A'|`false`|`r`|`c`|0|0|
|Band|'G'|'B'|'A'|`false`|`r`|`c`|`kU`|`kL`|
|Symmetric|'S'|'F'|`ul`|`false`|`n`|`n`|0|0|
|SymmetricBand|'S'|'B'|`ul`|`false`|`r`|`c`|`k`|`k`|
|SymmetricPacked|'S'|'P'|`ul`|`false`|`r`|`c`|0|0|
|Triangular|'T'|'F'|`ul`|`Diag==Unit`|`n`|`n`|0|0|
|TriangularBand|'T'|'B'|`ul`|`Diag==Unit`|`r`|`c`|`k`|`k`|
|TriangularPacked|'T'|'P'|`ul`|`Diag==Unit`|`r`|`c`|0|0|

This can be unmarshaled into a common internal struct
```
type storage struct {
    Form    byte
    Packing byte
    Uplo    byte
    Unit    bool
    Rows    int
    Cols    int
    KU      int
    KL      int
}
```

Does this sound reasonable?

kortschak:
PTAL

kortschak:
Failure is real but no longer relevant.

kortschak:
*ping*

mat/io.go
```diff
@@ -370,3 +427,53 @@ func readFull(r io.Reader, buf []byte) (int, error) {
 	}
 	return n, err
 }
+
+// storage is the internal representation of the storage format of a
+// serialised matrix.
+type storage struct {
+	Version uint32 // Keep this first.
+	Form    byte   // [GST]
```
sbinet: I still feel that the meaning of the `'G'`, `'S'`, `'T'`, ... letters could be explained somewhere.
presumably:
- `G` -> General
- `S` -> Symmetric
- `T` -> Triangular
- `B` -> Band ?
- `P` -> Packed ?
- `F` -> Fortran?
- `A` -> Any?
- `U` -> Up
- `L` -> Low

mat/io.go
```diff
@@ -370,3 +427,53 @@ func readFull(r io.Reader, buf []byte) (int, error) {
...
+	Unit    bool
+	Rows    int64
+	Cols    int64
+	KU      int64
+	KL      int64
+}
+
+func (s storage) marshalBinaryTo(w io.Writer) (int, error) {
+	buf := bytes.NewBuffer(make([]byte, 0, headerSize))
+	err := binary.Write(buf, binary.LittleEndian, s)
```
sbinet: this is using a lot of (slow) reflection.
could we at least add a `TODO/FIXME` to replace this with direct calls to `LittleEndian.PutUintXX` ?

mat/io.go
```diff
@@ -370,3 +427,53 @@ func readFull(r io.Reader, buf []byte) (int, error) {
...
+	if err != nil {
+		return 0, err
+	}
+	return w.Write(buf.Bytes())
+}
+
+func (s *storage) unmarshalBinary(buf []byte) error {
+	err := binary.Read(bytes.NewReader(buf), binary.LittleEndian, s)
```
sbinet: ditto (slow reflection)

mat/io.go
```diff
@@ -370,3 +427,53 @@ func readFull(r io.Reader, buf []byte) (int, error) {
...
+	if err != nil {
+		return err
+	}
+	if s.Version != version {
```
sbinet: I would perhaps add a comment about (perhaps) add code to handle earlier versions when the time comes.
(just as a kind note for our future selves who will end up landing here)

mat/io.go
```diff
@@ -370,3 +427,53 @@ func readFull(r io.Reader, buf []byte) (int, error) {
...
+		return fmt.Errorf("mat: incorrect version: %d", s.Version)
+	}
+	return nil
+}
+
+func (s *storage) unmarshalBinaryFrom(r io.Reader) (int, error) {
+	buf := make([]byte, headerSize)
+	n, err := readFull(r, buf)
```
sbinet: IIRC, I added `readFull` because we were supporting a Go version which didn't ship with `io.ReadFull`.
perhaps drop it ?

mat/io.go
```diff
@@ -370,3 +427,53 @@ func readFull(r io.Reader, buf []byte) (int, error) {
...
+	if err != nil {
+		return n, err
+	}
+	err = binary.Read(bytes.NewReader(buf), binary.LittleEndian, s)
```
sbinet: couldn't this be replaced with a call to `storage.unmarshalBinary` ?

mat/io.go
kortschak: These are in https://github.com/gonum/gonum/pull/374#discussion_r187238701. I can put that text in the source.

Full,Band,Packed
All,Upper,Lower.

mat/io.go
kortschak: Yeah, this is a first iteration, so since I didn't have any indication of whether the proposed approach was OK, I did the simplest (for me) to write. These can be optimised, though if you I/O is your major issue, you're lucky.

mat/io.go
kortschak: Maybe. I'll think about it.

mat/io.go
kortschak: I was wondering why it was there.

mat/io.go
kortschak: I can't remember why I did it this way. It should just use `r`.

kortschak: Ah, I remember, binary.Read and binary.Write don't return the number of bytes. So this does that, but yes, this can be changed.

mat/io.go
kortschak: Done

mat/io.go
kortschak: Added TODO.

mat/io.go
kortschak: Added table of scheme.

mat/io.go
kortschak: The one here behaves differently to io.ReadFull. I'm going to leave this for another PR.

kortschak:
PTAL

kortschak:
They get dismissed by new commit in gonum/gonum, so I need to ask.

sbinet:
I know. I wasn't complaining :) (apologies if it seemed like I was)

kortschak:
No worries.

PR #378, opened 2018-01-03
mat: generalise outer and rank update vector parameters

### LINKED ISSUE #167: all: generalise vector functions

kortschak:
Please take a look.

Closes #167.

### DIFF
--- a/mat/cholesky.go
+++ b/mat/cholesky.go
@@ -423,12 +423,12 @@ func (chol *Cholesky) ExtendVecSym(a *Cholesky, v Vector) (ok bool) {
 //
 // SymRankOne updates a Cholesky factorization in O(n²) time. The Cholesky
 // factorization computation from scratch is O(n³).
-func (c *Cholesky) SymRankOne(orig *Cholesky, alpha float64, x *VecDense) (ok bool) {
+func (c *Cholesky) SymRankOne(orig *Cholesky, alpha float64, x Vector) (ok bool) {
 	if !orig.valid() {
 		panic(badCholesky)
 	}
 	n := orig.Size()
-	if x.Len() != n {
+	if r, c := x.Dims(); r != n || c != 1 {
 		panic(ErrShape)
 	}
 	if orig != c {
@@ -472,7 +472,15 @@ func (c *Cholesky) SymRankOne(orig *Cholesky, alpha float64, x *VecDense) (ok bo
 
 	work := getFloats(n, false)
 	defer putFloats(work)
-	blas64.Copy(n, x.RawVector(), blas64.Vector{1, work})
+	var xmat blas64.Vector
+	if rv, ok := x.(RawVectorer); ok {
+		xmat = rv.RawVector()
+	} else {
+		var tmp *VecDense
+		tmp.CopyVec(x)
+		xmat = tmp.RawVector()
+	}
+	blas64.Copy(n, xmat, blas64.Vector{1, work})
 
 	if alpha > 0 {
 		// Compute rank-1 update.
--- a/mat/dense_arithmetic.go
+++ b/mat/dense_arithmetic.go
@@ -679,13 +679,22 @@ func (m *Dense) RankOne(a Matrix, alpha float64, x, y *VecDense) {
 	*m = w
 }
 
-// Outer calculates the outer product of x and y, and stores the result
-// in the receiver.
+// Outer calculates the outer product of the column vectors x and y,
+// and stores the result in the receiver.
 //  m = alpha * x * y'
 // In order to update an existing matrix, see RankOne.
-func (m *Dense) Outer(alpha float64, x, y *VecDense) {
-	r := x.Len()
-	c := y.Len()
+func (m *Dense) Outer(alpha float64, x, y Vector) {
+	xr, xc := x.Dims()
+	if xc != 1 {
+		panic(ErrShape)
+	}
+	yr, yc := y.Dims()
+	if yc != 1 {
+		panic(ErrShape)
+	}
+
+	r := xr
+	c := yr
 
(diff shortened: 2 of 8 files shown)


btracey:
I got through most of it, but I'm not sure I fully understand our policy on Vectors. For instance, here a lot of things check for the number of columns equaling one. I think instead our policy should be that a `Vector` is always a column vector, and so we can use `Len`. In our current API, `VecDense.T()` returns a `Matrix`, not a `Vector`, and we do not have a `TVec`, like we do in the other types. I think this is a feature.

kortschak:
>In our current API, VecDense.T() returns a Matrix, not a Vector,

This is the nature of the `Matrix` interface, everything that is a matrix must satisfy that.

>and we do not have a TVec

Yes, we do. There is a glitch in the current code that means that you can't `v.T().(Vector)`, but that is an oversight.

kortschak:
The documentation for `VecDense` and `Vector` say they are column vector and vector respectively. One of the issues is that there is only one type of concrete vector and that that is deemed to be column, independent of it's origin, so `m.RowView(i)` returns a vector of some sort, under the model that all vectors are columns, this surprisingly is a transpose of the row. This surprise was mitigated by making the row and column accessors return `Vector` which is merely defined as a column.

The underlying problem is that the Go type system is not strong enough to capture the richness of the operations here without some lenience.

btracey:
>> and we do not have a TVec
> Yes, we do. There is a glitch in the current code that means that you can't v.T().(Vector), but that is an oversight.

You're right, sorry. `VecDense` does not have a `TVec` method (though it probably should).

Let me rephrase my comment, in light of yours. I don't have a problem with saying that a `Vector` is just a vector, and not a column vector. I do think it's a problem to generalize `Outer` and say that the vectors have to be column vectors. There's no need to force that, and it's a bit weird mathematically (why not just insist that y be a row vector?). Instead, I think we should just say `Outer calculates the outer product of the vectors x and y, where both are interpreted as a column vector`. (or, `where x is treated as a column vector and y is treated as a row vector`). This way it doesn't matter if `RowView` returns a row vector or a column vector, you can still use it wherever you use a `Vector`. I'd rather it be, as much as possible, that when the input parameter is a `Vector`, it doesn't matter which orientation (though obviously the orientation does matter when it will be used as `Matrix`).

kortschak:
That sounds reasonable to me; I actually was initially thinking to do that approach, but the documentation was difficult to properly specify so I went the more constrained route that you see here. Note that other PRs have already been merged that have the dimension checks on the input vectors. Because of this, I'd like to merge this as is and have a completely consistent state to more forward from. Do you think that you would be able to do the dimension relaxation PR and add appropriate documentation?

Also see #384 and #385.

btracey:
SGTM. I'll take a look  at the full PR when I get a chance, on travel at the moment.

PR #379, opened 2018-01-03
graph: add multigraph implementation

### LINKED ISSUE #342: graph: multigraph API
This comes from thought arising from https://github.com/gonum/gonum/pull/335#issuecomment-350939998.

If we make a new `type Line interface { Edge; ID() int64 }` (line is used in some graph terminologies for an edge that is distinguishable other than by its end points), then we can define a multigraph API that looks like this:

```
// Line is an edge in a multigraph. A Line returns an ID that must
// distinguish Lines sharing Node end points.
type Line interface {
	Edge
	ID() int64
}

// WeightedLine is a weighted multigraph edge.
type WeightedLine interface {
	Line
	Weight() float64
}

// Multigraph is a generalized multigraph.
type Multigraph interface {
	// Has returns whether the node exists within the multigraph.
	Has(Node) bool

	// Nodes returns all the nodes in the multigraph.
	Nodes() []Node

	// From returns all nodes that can be reached directly
	// from the given node.
	From(Node) []Node

	// HasEdgeBetween returns whether an edge exists between
	// nodes x and y without considering direction.
	HasEdgeBetween(x, y Node) bool

	// Lines returns the lines from u to v if any such lines
	// exist and nil otherwise. The node v must be directly
	// reachable from u as defined by the From method.
	Lines(u, v Node) []Line
}

// WeightedMultigraph is a weighted multigraph.
type WeightedMultigraph interface {
	Multigraph

	// WeightedEdge returns the weighted edge from u to v if
	// such an edge exists and nil otherwise. The node v must
	// be directly reachable from u as defined by the
	// From method.
	WeightedLines(u, v Node) []WeightedLine
}

// UndirectedMultigraph is an undirected multigraph.
type UndirectedMultigraph interface {
	Multigraph

	// LinesBetween returns the lines between nodes x and y.
	LinesBetween(x, y Node) []Line
}

// WeightedUndirectedMultigraph is a weighted undirected multigraph.
type WeightedUndirectedMultigraph interface {
	WeightedMultigraph

	// WeightedEdgeBetween returns the edge between nodes
	// x and y.
	WeightedLinesBetween(x, y Node) []WeightedLine
}

// DirectedMultigraph is a directed multigraph.
type DirectedMultigraph interface {
	Multigraph

	// HasEdgeFromTo returns whether an edge exists
	// in the multigraph from u to v.
	HasEdgeFromTo(u, v Node) bool

	// To returns all nodes that can reach directly
	// to the given node.
	To(Node) []Node
}

// WeightedDirectedMultigraph is a weighted directed multigraph.
type WeightedDirectedMultigraph interface {
	WeightedMultigraph

	// HasEdgeFromTo returns whether an edge exists
	// in the multigraph from u to v.
	HasEdgeFromTo(u, v Node) bool

	// To returns all nodes that can reach directly
	// to the given node.
	To(Node) []Node
}

// LineAdder is an interface for adding lines to a multigraph.
type LineAdder interface {
	// NewLine returns a new Line from the source to the destination node.
	NewLine(from, to Node) Edge

	// SetLine adds an edge from one node to another.
	// If the multigraph supports node addition the nodes
	// will be added if they do not exist, otherwise
	// SetLine will panic.
	SetLine(l Line)
}

// WeightedLineAdder is an interface for adding liness to a multigraph.
type WeightedLineAdder interface {
	// NewWeightedLine returns a new WeightedLine from
	// the source to the destination node.
	NewWeightedLine(from, to Node, weight float64) WeightedLine

	// SetWeightedEdge adds an edge from one node to
	// another. If the multigraph supports node addition
	// the nodes will be added if they do not exist,
	// otherwise SetWeightedEdge will panic.
	SetWeightedLine(e WeightedLine)
}

// LineRemover is an interface for removing lines from a multigraph.
type LineRemover interface {
	// RemoveLine removes the given line, leaving the
	// terminal nodes. If the line does not exist it
	// is a no-op.
	RemoveLine(Line)
}

// MultigraphBuilder is a multigraph that can have nodes and lines added.
type MultigraphBuilder interface {
	NodeAdder
	LineAdder
}

// WeightedMultigraphBuilder is a multigraph that can have nodes and weighted lines added.
type WeightedMultigraphBuilder interface {
	NodeAdder
	WeightedLineAdder
}

// UndirectedMultgraphBuilder is an undirected multigraph builder.
type UndirectedMultigraphBuilder interface {
	UndirectedMultigraph
	MultigraphBuilder
}

// UndirectedWeightedMultigraphBuilder is an undirected weighted multigraph builder.
type UndirectedWeightedMultigraphBuilder interface {
	UndirectedMultigraph
	WeightedMultigraphBuilder
}

// DirectedMultigraphBuilder is a directed multigraph builder.
type DirectedMultigraphBuilder interface {
	DirectedMultigraph
	MultigraphBuilder
}

// DirectedWeightedMultigraphBuilder is a directed weighted multigraph builder.
type DirectedWeightedMultigraphBuilder interface {
	DirectedMultigraph
	WeightedMultigraphBuilder
}
```

This is almost a complete reflection of the simple graph API with one minor reduction; the `WeightedMultigraph` interface does not have a `Weight(x, y Node) (w float64, ok bool)` method as it's not clear to me how this would work (it pretty much an optimisation in the simple graphs anyway).

Also note that is is possible for a multigraph to return something like a `type Edges []Line` that satisfies the `Edge` interface to trivially satisfy the simple graph interfaces (or just return the first found edge). For the `[]WeightLine` case, the returned type could calculate the sum, mean, min etc weights for the `Edge` `Weight` method.

If this seems good, I'll revert #335 and add this.

/cc @mewmew @vladimir-ch

kortschak:
Please take a look.

Note that dates are old since the code is largely cribbed from simple.

Closes #342.

### DIFF
--- a/graph/multi/directed.go
+++ b/graph/multi/directed.go
@@ -1,0 +1,275 @@
+// Copyright ©2014 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package multi
+
+import (
+	"fmt"
+
+	"gonum.org/v1/gonum/graph"
+	"gonum.org/v1/gonum/graph/internal/uid"
+)
+
+var (
+	dg *DirectedGraph
+
+	_ graph.Graph              = dg
+	_ graph.Directed           = dg
+	_ graph.Multigraph         = dg
+	_ graph.DirectedMultigraph = dg
+)
+
+// DirectedGraph implements a generalized directed graph.
+type DirectedGraph struct {
+	nodes map[int64]graph.Node
+	from  map[int64]map[int64]map[int64]graph.Line
+	to    map[int64]map[int64]map[int64]graph.Line
+
+	nodeIDs uid.Set
+	lineIDs uid.Set
+}
+
+// NewDirectedGraph returns a DirectedGraph.
+func NewDirectedGraph() *DirectedGraph {
+	return &DirectedGraph{
+		nodes: make(map[int64]graph.Node),
+		from:  make(map[int64]map[int64]map[int64]graph.Line),
+		to:    make(map[int64]map[int64]map[int64]graph.Line),
+
+		nodeIDs: uid.NewSet(),
+		lineIDs: uid.NewSet(),
+	}
+}
+
+// NewNode returns a new unique Node to be added to g. The Node's ID does
+// not become valid in g until the Node is added to g.
+func (g *DirectedGraph) NewNode() graph.Node {
+	if len(g.nodes) == 0 {
+		return Node(0)
+	}
+	if int64(len(g.nodes)) == uid.Max {
+		panic("simple: cannot allocate node: no slot")
+	}
+	return Node(g.nodeIDs.NewID())
+}
+
+// AddNode adds n to the graph. It panics if the added node ID matches an existing node ID.
+func (g *DirectedGraph) AddNode(n graph.Node) {
+	if _, exists := g.nodes[n.ID()]; exists {
+		panic(fmt.Sprintf("simple: node ID collision: %d", n.ID()))
+	}
+	g.nodes[n.ID()] = n
+	g.from[n.ID()] = make(map[int64]map[int64]graph.Line)
+	g.to[n.ID()] = make(map[int64]map[int64]graph.Line)
+	g.nodeIDs.Use(n.ID())
+}
+
+// RemoveNode removes n from the graph, as well as any edges attached to it. If the node
+// is not in the graph it is a no-op.
+func (g *DirectedGraph) RemoveNode(n graph.Node) {
+	if _, ok := g.nodes[n.ID()]; !ok {
+		return
+	}
+	delete(g.nodes, n.ID())
+
+	for from := range g.from[n.ID()] {
+		delete(g.to[from], n.ID())
+	}
+	delete(g.from, n.ID())
+
+	for to := range g.to[n.ID()] {
+		delete(g.from[to], n.ID())
+	}
+	delete(g.to, n.ID())
+
+	g.nodeIDs.Release(n.ID())
+}
+
+// NewLine returns a new Line from the source to the destination node.
+// The returned Line will have a graph-unique ID.
+// The Line's ID does not become valid in g until the Line is added to g.
+func (g *DirectedGraph) NewLine(from, to graph.Node) graph.Line {
+	return &Line{F: from, T: to, UID: g.lineIDs.NewID()}
+}
+
+// SetLine adds l, a line from one node to another. If the nodes do not exist, they are added.
+func (g *DirectedGraph) SetLine(l graph.Line) {
+	var (
+		from = l.From()
+		fid  = from.ID()
+		to   = l.To()
+		tid  = to.ID()
+		lid  = l.ID()
+	)
+
+	if !g.Has(from) {
+		g.AddNode(from)
+	}
+	if g.from[fid][tid] == nil {
+		g.from[fid][tid] = make(map[int64]graph.Line)
+	}
+	if !g.Has(to) {
+		g.AddNode(to)
+	}
+	if g.to[tid][fid] == nil {
+		g.to[tid][fid] = make(map[int64]graph.Line)
+	}
+
+	g.from[fid][tid][lid] = l
+	g.to[tid][fid][lid] = l
+	g.lineIDs.Use(lid)
+}
+
+// RemoveLine removes l from the graph, leaving the terminal nodes. If the line does not exist
+// it is a no-op.
+func (g *DirectedGraph) RemoveLine(l graph.Line) {
+	from, to := l.From(), l.To()
+	if _, ok := g.nodes[from.ID()]; !ok {
+		return
+	}
+	if _, ok := g.nodes[to.ID()]; !ok {
+		return
+	}
+
+	delete(g.from[from.ID()][to.ID()], l.ID())
+	if len(g.from[from.ID()][to.ID()]) == 0 {
+		delete(g.from[from.ID()], to.ID())
+	}
+	delete(g.to[to.ID()][from.ID()], l.ID())
+	if len(g.to[to.ID()][from.ID()]) == 0 {
+		delete(g.to[to.ID()], from.ID())
+	}
+	g.lineIDs.Release(l.ID())
+}
+
+// Node returns the node in the graph with the given ID.
+func (g *DirectedGraph) Node(id int64) graph.Node {
+	return g.nodes[id]
+}
+
+// Has returns whether the node exists within the graph.
+func (g *DirectedGraph) Has(n graph.Node) bool {
+	_, ok := g.nodes[n.ID()]
+	return ok
+}
+
+// Nodes returns all the nodes in the graph.
+func (g *DirectedGraph) Nodes() []graph.Node {
+	if len(g.nodes) == 0 {
+		return nil
+	}
+	nodes := make([]graph.Node, len(g.nodes))
+	i := 0
+	for _, n := range g.nodes {
+		nodes[i] = n
+		i++
+	}
+	return nodes
+}
+
+// Edges returns all the edges in the graph. Each edge in the returned slice
+// is a multi.Edge.
+func (g *DirectedGraph) Edges() []graph.Edge {
+	var edges []graph.Edge
+	for _, u := range g.nodes {
+		for _, e := range g.from[u.ID()] {
+			var lines Edge
+			for _, l := range e {
+				lines = append(lines, l)
+			}
+			if len(lines) != 0 {
+				edges = append(edges, lines)
+			}
+		}
+	}
+	return edges
+}
+
+// From returns all nodes in g that can be reached directly from n.
+func (g *DirectedGraph) From(n graph.Node) []graph.Node {
+	if _, ok := g.from[n.ID()]; !ok {
+		return nil
+	}
+
+	from := make([]graph.Node, len(g.from[n.ID()]))
+	i := 0
+	for id := range g.from[n.ID()] {
+		from[i] = g.nodes[id]
+		i++
+	}
+	return from
+}
+
+// To returns all nodes in g that can reach directly to n.
+func (g *DirectedGraph) To(n graph.Node) []graph.Node {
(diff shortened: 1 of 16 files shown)


graph/multi/directed_test.go
```diff
@@ -0,0 +1,65 @@
...
+import (
+	"testing"
+
+	"gonum.org/v1/gonum/graph"
+)
+
+var (
+	directedGraph = (*DirectedGraph)(nil)
+
+	_ graph.Graph    = directedGraph
```
mewmew: Same as above. If the `directedGraph` global variable is not used, then perform assert that the interfaces are implemented using `(*DirectedGraph)(nil)` directly.

graph/multi/multi.go
```diff
@@ -0,0 +1,112 @@
...
+// license that can be found in the LICENSE file.
+
+package multi
+
+import "gonum.org/v1/gonum/graph"
+
+// Node here is a duplication of simple.Node
+// to avoid needing to import both packages.
+
+// Node is a simple graph node.
```
mewmew: Should it still state that Node is a *simple* node?

graph/multi/multi.go
```diff
@@ -0,0 +1,112 @@
...
+
+// ID returns the ID number of the node.
+func (n Node) ID() int64 {
+	return int64(n)
+}
+
+// Edge is a collection of multigraph edges sharing end points.
+type Edge []graph.Line
+
+// From returns the from-node of the edge.
```
mewmew: Add a comment along the lines of "If more than one edge exist, the from node of the first edge is returned." Likewise for `To`.

graph/multi/multi.go
```diff
@@ -0,0 +1,112 @@
...
+	Lines []graph.WeightedLine
+
+	// WeightFunc calculates the aggregate
+	// weight of the lines in Lines. If
+	// WeightFunc is nil, the sum of weights
+	// is used as the edge weight.
+	WeightFunc func([]graph.WeightedLine) float64
+}
+
+// From returns the from-node of the edge.
```
mewmew: Add comment that the from-node is returned from the first line, if more than one line exist.

graph/multi/undirected.go
```diff
@@ -0,0 +1,250 @@
...
+
+import (
+	"fmt"
+
+	"gonum.org/v1/gonum/graph"
+	"gonum.org/v1/gonum/graph/internal/uid"
+)
+
+var (
+	ug *UndirectedGraph
```
mewmew: Same comment as above. Remove `ug` if not used.

graph/multi/directed.go
```diff
@@ -0,0 +1,275 @@
...
+	"fmt"
+
+	"gonum.org/v1/gonum/graph"
+	"gonum.org/v1/gonum/graph/internal/uid"
+)
+
+var (
+	dg *DirectedGraph
+
+	_ graph.Graph              = dg
```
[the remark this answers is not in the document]
kortschak: I'm OK with this small additional space cost for the benefit of readability. I should change how simple does this - I prefer to have it in the built package and not just in the test files since tests may not be run.

graph/multi/directed_test.go
kortschak: Ack

graph/multi/multi.go
kortschak: This is simple as in a simple thing, not simple as in a simple graph. This is true in simple as well since there is no such thing as a simple node in graph theory.

graph/multi/multi.go
kortschak: This is an implementation detail that I would rather not document; if people are doing things that make that behaviour distinguishable, they are holding it wrong. This does suggest to me though a way that we can deal with the node edge ordering subtlety. I am thinking that perhaps graph.Edge should be able to set nodes irrespective of whether it is a constructor type. This would mean that a properly written edge type would always have the correct node behaviour. I'll send comments to that issue later.

graph/multi/multi.go
kortschak: As above.

graph/multi/undirected.go
kortschak: Ack

kortschak:
I'll wait for others as well.

graph/multi/directed.go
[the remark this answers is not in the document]
mewmew: Actually, I `dg` would add to the space of the executable was not part of my rationale for removing it. Rather, I consider it code hygiene to limit the number of global variables. And seeing as this code base is spread across multiple files, for a user not yet accustomed to the code base it may take some time to figure out that `dg` is not used. That was the rationale for removing it entirely, and simply using `var _ graph.Graph = (*DirectedGraph)(nil)` as the `_` identifier can never be used.

graph/multi/multi.go
mewmew: Sounds good. Thanks for the explanation.

graph/multi/directed.go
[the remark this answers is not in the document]
kortschak: The var is unexported and in a var block. This is a reasonably commonly used convention.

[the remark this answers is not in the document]
sbinet: the "trick" @mewmew refers to is also a very reasonable and commonly used convention :)
(one could also add a little comment explaining what it does and why)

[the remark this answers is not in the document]
kortschak: It's the same 'trick', the only difference is the dialect. We have used the dialect here in other packages, largely because of the verbosity of doing it the other way. The existence of an apparently no-op var block like this reads as clearly as a comment; I really don't think a change is necessary here.

kortschak:
*ping* @someone_with_interest_in_graphs

graph/multi/undirected.go
```diff
@@ -0,0 +1,250 @@
...
+// UndirectedGraph implements a generalized undirected graph.
+type UndirectedGraph struct {
+	nodes map[int64]graph.Node
+	lines map[int64]map[int64]map[int64]graph.Line
+
+	nodeIDs uid.Set
+	lineIDs uid.Set
+}
+
+// NewUndirectedGraph returns an UndirectedGraph with the specified self and absent
```
vladimir-ch: No weights here.

kortschak: I'll wait for the rest of the review before I fix this.

graph/multi/weighted_directed_test.go
```diff
@@ -0,0 +1,66 @@
...
+	weightedDirectedGraph = (*WeightedDirectedGraph)(nil)
+
+	_ graph.Graph            = weightedDirectedGraph
+	_ graph.Directed         = weightedDirectedGraph
+	_ graph.WeightedDirected = weightedDirectedGraph
+)
+
+// Tests Issue #27
+func TestWeightedEdgeOvercounting(t *testing.T) {
+	g := generateDummyGraph()
```
vladimir-ch: Should this use `generateDummyWeightedGraph`?

graph/multi/undirected.go
kortschak: Done

graph/multi/weighted_directed.go
```diff
@@ -0,0 +1,331 @@
...
+	if g.to[tid][fid] == nil {
+		g.to[tid][fid] = make(map[int64]graph.WeightedLine)
+	}
+
+	g.from[fid][tid][lid] = l
+	g.to[tid][fid][lid] = l
+	g.lineIDs.Use(l.ID())
+}
+
+// RemoveLine removes l from the graph, leaving the terminal nodes. If the line does not exist
```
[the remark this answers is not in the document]
kortschak: Done

graph/multi/weighted_directed.go
```diff
@@ -0,0 +1,331 @@
...
+	for _, n := range g.nodes {
+		nodes[i] = n
+		i++
+	}
+
+	return nodes
+}
+
+// Edges returns all the edges in the graph. Each edge in the returned slice
+// is a multi.WeighteEdge.
```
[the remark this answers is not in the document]
kortschak: Done

graph/multi/weighted_directed.go
```diff
@@ -0,0 +1,331 @@
...
+				lines.WeightFunc = g.EdgeWeightFunc
+				edges = append(edges, lines)
+			}
+		}
+	}
+	return edges
+}
+
+// WeightedEdges returns all the edges in the graph. Each edge in the returned slice
+// is a multi.WeighteEdge.
```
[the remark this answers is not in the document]
kortschak: Done

graph/multi/weighted_directed_test.go
kortschak: Yes, thanks.

Done

graph/multi/weighted_undirected.go
```diff
@@ -0,0 +1,308 @@
...
+	EdgeWeightFunc func([]graph.WeightedLine) float64
+
+	nodes map[int64]graph.Node
+	lines map[int64]map[int64]map[int64]graph.WeightedLine
+
+	nodeIDs uid.Set
+	lineIDs uid.Set
+}
+
+// NewWeightedUndirectedGraph returns an WeightedUndirectedGraph with the specified self and absent
```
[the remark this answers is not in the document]
kortschak: Done

Yes, those default values in graphs in simple allow the `graph.Weighter` interface to be satisfied with user-defined values. That user hook here is provided through the `EdgeWeightFunc` field. I could have done the same thing in the constructor here as in simple - providing the func via the constructor, but this seemed nicer since there is only one value to pass and you don't need to be able to see two consistent values being passed at the same time. If you think I should change this, please let me know.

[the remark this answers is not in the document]
vladimir-ch: Yes, this is nicer.

graph/multi/weighted_directed_test.go
vladimir-ch: I think that `simple/weighted_directed_test.go` has the same oversight. You may want to fix it later.

kortschak: I'll fix that here and if you don't mind, squash the last two commits here into one so I can leave the three intended commits separate when I mergre. This will need another approval when it's done.

kortschak:
Review comment commit squashed.

kortschak:
Thanks

PR #381, opened 2018-01-04
diff: Derivative Performance

### LINKED ISSUE #380: diff: Derivative Performance
I created benchmarks comparing the serial and concurrent Derivative function implementations. It seems the serial implementation will always be faster. 

Here is the benchmark result:
```
# go test -benchtime=10s -bench=.
BenchmarkDerivativeSerial-8          	200000000	        64.5 ns/op
BenchmarkDerivativeConcurrent-8      	10000000	      1937 ns/op
BenchmarkDerivative2ndSerial-8       	100000000	       113 ns/op
BenchmarkDerivative2ndConcurrent-8   	 5000000	      2865 ns/op
PASS
ok  	gonum.org/v1/gonum/diff/fd	69.481s
```

It seems the concurrent [Derivative](https://github.com/gonum/gonum/blob/master/diff/fd/derivative.go#L17) function merely calculates the differential between each `Stencil` point (sans origin) in a separate goroutine. Unfortunately, there are only two points in the first-order approximations, and three points in the second-order. If there were many points, the concurrent implementation could potentially enhance performance. As it is, there is a large overhead for the concurrent path, as we create a `sync.Waitgroup` and lock/unlock a `sync.Mutex` for each of the points. I propose we remove the concurrent path, and always use the unencumbered sequential calculation.

The benchmark code is as follows:
```
func BenchmarkDerivativeSerial(b *testing.B) {
	// run the Derivative benchmark b.N times
	for n := 0; n < b.N; n++ {
		Derivative(xSquared, 10, &Settings{Concurrent: false})
	}
}

func BenchmarkDerivativeConcurrent(b *testing.B) {
	// run the Derivative benchmark b.N times
	for n := 0; n < b.N; n++ {
		Derivative(xSquared, 10, &Settings{Concurrent: true})
	}
}

func BenchmarkDerivative2ndSerial(b *testing.B) {
	// run the Derivative benchmark b.N times
	for n := 0; n < b.N; n++ {
		Derivative(xSquared, 10, &Settings{Formula: Forward2nd, Concurrent: false})
	}
}

func BenchmarkDerivative2ndConcurrent(b *testing.B) {
	// run the Derivative benchmark b.N times
	for n := 0; n < b.N; n++ {
		Derivative(xSquared, 10, &Settings{Formula: Forward2nd, Concurrent: true})
	}
}
```

SellJamHere:
Fixes #380 

The PR removes the concurrent code path in the `Derivative` function. To maintain backwards compatibility, the concurrent value in `Settings` is unchanged. However, the setting is now a no-op. 

`benchstat` comparison of the Derivative pre/post PR:
```
# benchstat old.txt new.txt 
name                       old time/op  new time/op  delta
DerivativeSerial-8         67.0ns ± 0%  28.3ns ± 0%   ~     (p=1.000 n=1+1)
DerivativeConcurrent-8     1.98µs ± 0%  0.03µs ± 0%   ~     (p=1.000 n=1+1)
Derivative2ndSerial-8       116ns ± 0%    74ns ± 0%   ~     (p=1.000 n=1+1)
Derivative2ndConcurrent-8  2.88µs ± 0%  0.07µs ± 0%   ~     (p=1.000 n=1+1)
```

### DIFF
--- a/diff/fd/derivative.go
+++ b/diff/fd/derivative.go
@@ -6,8 +6,6 @@ package fd
 
 import (
 	"math"
-	"runtime"
-	"sync"
 )
 
 // Derivative estimates the derivative of the function f at the given location.
@@ -19,7 +17,7 @@ func Derivative(f func(float64) float64, x float64, settings *Settings) float64 
 	formula := Forward
 	step := formula.Step
 	var originValue float64
-	var originKnown, concurrent bool
+	var originKnown bool
 
 	// Use user settings if provided.
 	if settings != nil {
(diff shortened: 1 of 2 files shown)


kortschak:
I have not touched this code, but it seems to me that a concurrent implementation that performs better is easily achievable with minimal effort here; instead of the goroutine-per-point approach used here which is obviously inefficient, `runtime.GOMAXPROCS(0)` goroutines should be started, each with a work unit of `n/runtime.GOMAXPROCS(0)` points, these can then either update `$GOMAXPROCS` elements of a slice to be summed after all workers return, or be reduced some other way.

BTW, `benchstat` should be given more than 1 replicate run (use `go test -run ^$ -bench <foo> -count 10`, e.g.).

SellJamHere:
Closing, after discussion in #381

Issue #354, opened 2017-12-18
optimize: Constrained optimization


dezmondgoff:
What are people's thoughts on how constrained optimization should be handled? I ask because I'm writing an implementation of the parametric simplex and affine scaling methods for linear optimization, and would be interested in writing an SQP method. I can imagine a interface similar to optimize, but with constraint values and Jacobians within the Location struct. I think this could cover the most popular algorithms for general constrained optimization. There should probably also be flags for when gradients and hessians are constant.

btracey:
I would love there to be constrained optimization in Gonum. 

We already have an implementation of the Simplex algorithm, https://godoc.org/gonum.org/v1/gonum/optimize/convex/lp . It should be updated to work better with interfaces and with @vladimir-ch 's work on iterative linear solving. But still, it's a working implementation.

What we are missing is a (convex) QP solver. There are different algorithms for the different cases (unconstrained, only equality, only bounds, equality+bounds, general inequality). This is quite a bit of work (if Simplex is any guide), and I have have a partially done implementation I keep coming back to. One major stalling point is that both QP and Simplex have to solve the Phase 1 problem, and ideally that code should be shared between the implementations, which is a (second) refactoring of Simplex. 

Once those are in place, and we can actually support constrained optimization, then we can think about writing SLSQP (or other SQP solver), and add to optimize to provide support. See https://github.com/gonum/optimize/issues/198 for my past thoughts on how to do this (which I think are still accurate, but I haven't thought super deeply). One difficulty I'm seeing is defining `Operation` properly, which isn't really addressed.

dezmondgoff:
Very cool. I will take a look. Parametric simplex is actually a slightly different algorithm that doesn't require a Phase I and is particularly useful in sparse selection. But I mostly wanted to see if there was an established API yet. So thank you.

btracey:
There's room in `convex/lp` more more algorithms. I'd be happy to work help change the API that's more amenable to sparse matrices if you're interested in helping build the package.

dezmondgoff:
[...]

PR #382, opened 2018-01-06
lp: Add implementation of parametric simplex algorithm

deasmhumhna:
#354


btracey:
Please run gofmt on the code.

btracey:
Hi @dezmondgoff , sorry I haven't gotten to this at all. I've been having trouble finding time to do PR reviews recently, especially for large ones like this. I haven't forgotten about it though.

dezmondgoff:
No problem. Thanks for letting me know.

btracey:
Sorry this has been so long. I'll try and get to this soon. Thanks for the time you've put in so far

optimize/convex/lp/parametric.go
```diff
@@ -0,0 +1,448 @@
...
+type Trace struct {
+	Lambda, X []float64
+	Idx       []int
+}
+
+// Parametric solves a linear program in standard form using the parametric
+// simplex method, a primal-dual algorithm. A description of the algorithm can
+// be found in Pang, H. et al. (2017). "Parametric Simplex Method for Sparse Learning".
+// Advances in Neural Information Processing Systems(30): p. 187-196.
+func Parametric(c []float64, A mat.Matrix, b []float64, tol float64, initialBasic []int, dense bool, rnd *rand.Rand) (optF float64, optX []float64, basicIdxs []int, err error) {
```
btracey: This needs a description of all of the inputs, especially `dense`

optimize/convex/lp/parametric.go
```diff
@@ -0,0 +1,448 @@
...
+// function is useful for homotopy in L1 regularized minimization problems. Note that
+//  minimize	L(θ) + λ‖θ‖_1
+// can be written equivalently as
+//  minimize ‖θ‖_1
+//  s.t. 		‖∇L(θ)‖_∞ ≤ λ.
+// Either fomulation can both be expressed as a linear program when L(θ)
+// is piecewise linear or quadratic, respectively. Therefore the parametric simplex
+// method allows for recovery of the entire regularization path of at the cost of
+// solving a single linear program.
+func ParametricWithTrace(c []float64, A mat.Matrix, b, cbar, bbar []float64, tol float64, initialBasic []int, rnd *rand.Rand) (optTr *Trace, err error) {
```
btracey: Is there a reason not to return the optimal solution as well?

optimize/convex/lp/parametric.go
```diff
@@ -0,0 +1,448 @@
...
+// simplex method and returns the output of each iteration in a Trace object. This
+// function is useful for homotopy in L1 regularized minimization problems. Note that
+//  minimize	L(θ) + λ‖θ‖_1
+// can be written equivalently as
+//  minimize ‖θ‖_1
+//  s.t. 		‖∇L(θ)‖_∞ ≤ λ.
+// Either fomulation can both be expressed as a linear program when L(θ)
+// is piecewise linear or quadratic, respectively. Therefore the parametric simplex
+// method allows for recovery of the entire regularization path of at the cost of
+// solving a single linear program.
```
btracey: Could you add some description of Trace, either here or for the comment of trace? I was expecting this to return something like `[]Trace`, but I see below that x actually has length `4*m` which is not clear to me why.

optimize/convex/lp/parametric.go
```diff
@@ -0,0 +1,448 @@
...
+
+	"gonum.org/v1/gonum/floats"
+	"gonum.org/v1/gonum/mat"
+)
+
+var (
+	ErrDegenerate = errors.New("lp: current perturbation is degenerate")
+)
+
+const (
```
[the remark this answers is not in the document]
deasmhumnha: Now I'm the one who doesn't have time to review my own code! As far as this method goes, it is nearly impossible for the randomly perturbed system to become degenerate.  In the rare case that this does happen, we can simply choose another random perturbation without changing anything else. However, the `ParametricWithTrace` method allows the user to provide their own perturbations when, for example, one wants to perform L1 minimization. These perturbations are by nature deterministic, so if they cause degeneracy (in the form of unordered `lambdaMin` and `lambdaMax`), then there is something infeasible about the underlying minimization task. I suppose I could still allow this behavior to be dictated by the user, though I don't really see the necessity.

optimize/convex/lp/parametric.go
deasmhumnha: I've changed this. 
```
// Solution represents a lambda-optimal solution to an LP.
type Solution struct {
	// Lambda is the perturbation coefficient/regularization parameter.
	Lambda float64
	// The values of the active variables.
	X []float64
	// The indices of the active variables.
	Idx []float64
}

// Path stores a list of lambda-optimal intermediate solutions found
// over the course of the optimization.
type Path []*Solution
```

optimize/convex/lp/parametric.go
deasmhumnha: Mostly because that solution is the final entry in the `Path` do it seemed redundant. Not to mention when using the method for minimization, all solutions are optimal for different values of the regularization parameter.

kortschak:

deasmhumhna:
Yes, I am still working on this. I have a update I forget to push. I'll do that later today.

vladimir-ch:

PR #392, opened 2018-01-12
Refactor initial Shortest

vlastv:


kortschak:
Not LGTM

This is not a refactor, it's a behaviour change. The use of the graph-owned node instead of the user-provide node is important for correct client use.

Is there some problem that you were facing that you think this fixes?

PR #398, opened 2018-01-24
lapack/testlapack: fix and extend test for Dlarfg

vladimir-ch:
Coverage for `Dlarfg` is complete with this.


lapack/testlapack/dlarfg.go
```diff
@@ -56,6 +66,9 @@ func DlarfgTest(t *testing.T, impl Dlarfger) {
 				x[i] = rnd.Float64()
 			}
 		} else {
+			if len(test.x) != n-1 {
```
kortschak: Here is a reason why it would be nice to have precise vector lengths, but I agree that it's a difficult problem.

PR #401, opened 2018-01-29
integrate: Add tests for panics in trapezoidal

millerick:
Increases coverage to 100%

### DIFF
--- a/integrate/trapezoidal_test.go
+++ b/integrate/trapezoidal_test.go
@@ -11,6 +11,15 @@ import (
 	"gonum.org/v1/gonum/floats"
 )
 
+func assertPanic(t *testing.T, f func() float64, desiredPanic string) {
+	defer func() {
+		if r := recover(); r != desiredPanic {
+			t.Errorf("The code did not panic with \"%s\"", desiredPanic)
+		}
+	}()
+	f()
+}
+
 func TestTrapezoidal(t *testing.T) {
 	const N = 1e6
 	x := floats.Span(make([]float64, N), 0, 1)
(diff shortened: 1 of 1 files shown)


integrate/trapezoidal_test.go
```diff
@@ -11,6 +11,15 @@ import (
 	"gonum.org/v1/gonum/floats"
 )
 
+func assertPanic(t *testing.T, f func() float64, desiredPanic string) {
```
kortschak: The idiom we use for this is not an assert, but rather a function that will returns whether a function call panicked, and what message the panic was, like so https://github.com/gonum/gonum/blob/287df04a63144ae62d3f7a85dc313836f665c85b/mat/matrix_test.go#L18-L26

millerick:
> A complete test coverage statistic is not a goal here.

I agree that 100% coverage is not necessarily representative of good tests, but I think that test coverage is an indicator for the relative health of a project, and I (perhaps wrongly) assumed that the maintainers of gonum would welcome changes that increased test coverage given that there is a coverage indicator present in the root README.
image

At a minimum, I'll make the required changes, if only for my own benefit.

integrate/trapezoidal_test.go
millerick: Looks like `panics` is present in several different test files in very similar forms.  Would you advise putting a single copy of the function under `gonum/internal`?

kortschak:
Please don't take this as a dismissal, it's just that this is a very minor contribution to the test coverage, increasing coverage for code that is obviously correct.

If you are interested in improving coverage (we do take this seriously), then picking good target from the list of source files that show very low (or no) coverage [here](https://coveralls.io/github/gonum/gonum?branch=master) would probably be more productive. An Easy™ way to do this would be to generate reliable single precision tests from the existing double precision tests in gonum/blas, since the the single precision tests are not tested (rather being implied from the fact that the code is generated from the double precision code). There are many other places that could do with test coverage improvement (and just out and out test improvement).

integrate/trapezoidal_test.go
kortschak: Yeah, maybe. If we make a test helpers package in internal that might be worthwhile, but I think that should be a separate PR, with a proposal issue first.

millerick: Agreed.

millerick:
Thanks for the link.  I'll find a different area of the code base that interests me and lacks coverage.

PR #404, opened 2018-01-29
floats: fix Same for NaN/N

kortschak:
Please take a look.

Fixes #403.

PR #406, opened 2018-01-29
floats: Same() with unmatched NaNs

### LINKED ISSUE #403: floats: Same() returns true when false is expected
Here is an example to demonstrates the issue:
```log.Println(floats.Same([]float64{math.NaN()}, []float64{3}))```

aleh-null:
Closes #403.


btracey:
It looks like there was a race condition and this is fixed by https://github.com/gonum/gonum/pull/404 ? Thanks for the bug report and PR.

aleh-null:
yes, looks good and  #404 is more idiomatic.

kortschak:
Closing in favour of #404.

PR #411, opened 2018-01-31
github: direct questions to gonum-dev

kortschak:
Please take a look.

### DIFF
--- a/.github/ISSUE_TEMPLATE
+++ b/.github/ISSUE_TEMPLATE
@@ -1,6 +1,11 @@
 <!--
 Please make sure your issue title matchs the Go convention; a summary
 of the problem, prefixed by the primary affected package.
+
+If the issue is a question about usage or similar, please ask at the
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE
```diff
@@ -1,6 +1,11 @@
 <!--
 Please make sure your issue title matchs the Go convention; a summary
 of the problem, prefixed by the primary affected package.
+
+If the issue is a question about usage or similar, please ask at the
+Gonum discussion rather than fileing an issue:
```
sbinet: `s/Gonum discussion/Gonum discussion list/` or something similar?
(to the non-native english speaker that I am, it sounded weird, like a word was missing.)

kortschak: Hah! I had it that way before and edited to something else and then didn't revert it properly.

Also spelling of "filing".

Fixed

PR #412, opened 2018-02-04
fourier: new package for fourier analysis and related functions

kortschak:
Please take a look.

So far I have only added the backing code. I want to add CFFT[IFB] and then I will dismantle some of the `array.go` support (remove 1-based indexing and the linear support types). It makes sense to add the API in this PR, so suggestions welcome here.


kortschak:
From an API perspective, I think that given there are work arrays kept, a type holding those that can be reused would be the most sensible approach:

```
type FFT struct

// NewFFT returns an FFT initialized for work on sequences of length n.
func NewFFT(n int) *FFT

// Len returns the length of the acceptable input.
func (t *FFT) Len() int

// Reset reinitializes the FFT for work on sequences of length n.
func (t *FFT) Reset(n int)

// FFT computes the Fourier coefficients of the input sequence, s,
// placing the result in dst and returning it.
// If dst is nil, a new slice is allocated and returned. If dst is not nil and
// the length of dst does not equal the length of s, FFT will panic.
func (t *FFT) FFT(dst, s []float64) []float64

// IFFT computes the real perodic sequence from the Fourier coefficients,
// c, placing the result in dst and returning it.
// If dst is nil, a new slice is allocated and returned. If dst is not nil and
// the length of dst does not equal the length of c, IFFT will panic.
func (t *FFT) IFFT(dst, c []float64) []float64
```

fourier/rfft.go
```diff
@@ -0,0 +1,1134 @@
...
+		ds2 := ai1
+		ar2 := ar1
+		ai2 := ai1
+		for j := 3; j <= ipph; j++ {
+			jc := ipp2 - j
+			ar2h := dc2*ar2 - ds2*ai2
+			ai2 = dc2*ai2 + ds2*ar2
+			ar2 = ar2h
+			for ik := 1; ik <= idl1; ik++ {
+				ch2m.set(ik, l, ch2m.at(ik, l)+ar2*c2m.at(ik, j))
```
kortschak: These two lines suggest that a `twoArray.add(i, j int, v float64` method would be a worthwhile optimisation.

fourier/rfft.go
```diff
@@ -0,0 +1,1134 @@
...
+				ch2m.set(ik, lc, ch2m.at(ik, lc)+ai2*c2m.at(ik, jc))
+			}
+		}
+	}
+	for j := 2; j <= ipph; j++ {
+		for ik := 1; ik <= idl1; ik++ {
+			ch2m.set(ik, 1, ch2m.at(ik, 1)+c2m.at(ik, j))
```
kortschak: And here.

fourier/rfft.go
```diff
@@ -0,0 +1,1134 @@
...
+		ds2 := ai1
+		ar2 := ar1
+		ai2 := ai1
+		for j := 3; j <= ipph; j++ {
+			jc := ipp2 - j
+			ar2h := dc2*ar2 - ds2*ai2
+			ai2 = dc2*ai2 + ds2*ar2
+			ar2 = ar2h
+			for ik := 1; ik <= idl1; ik++ {
+				c2m.set(ik, l, c2m.at(ik, l)+ar2*ch2m.at(ik, j))
```
kortschak: And here.

fourier/rfft.go
```diff
@@ -0,0 +1,1134 @@
...
+				c2m.set(ik, lc, c2m.at(ik, lc)+ai2*ch2m.at(ik, jc))
+			}
+		}
+	}
+
+	for j := 2; j <= ipph; j++ {
+		for ik := 1; ik <= idl1; ik++ {
+			ch2m.set(ik, 1, ch2m.at(ik, 1)+ch2m.at(ik, j))
```
kortschak: And here.

kortschak:
Damn. Not quite correct yet.

fourier/rfft.go
kortschak: There are done.

btracey:
I'll get to this when I can, I'm on travel at the moment.

kortschak:
No worries. Despite my best efforts, I managed to sneak some off-by-one errors into this - when you read the subtle one-based indexing removal, also read the fix off-by-one bugs commit. The sign bug goes with the use of the sign parameter.

kortschak:
Apologies for the unhelpful ordering - I revised the author name, and github insists on presenting PR changes in chronological order in a way that made that revision show differences.

btracey:
I understand the desire for having the types to reuse the work arrays. From what I understand, this is how FFTW works as well. Do you think it's desirable to also have some high level functions for ease of use? I haven't looked at all of the intended functions, but I wonder if we could get away with `Float64` and `Complex128` as types that hold the temporary memory, and the could have `FFT` be a function that allocates the temporary memory. 

Just a thought. I'll take a more detailed look soon hopefully.

kortschak:
They are an easy addition, so I think we should leave that for when someone asks for it.

kortschak:
I have put together the implementation and API for DCT, DST and QW-FFT (tests will come later), but I won't include them in this PR. However, after seeing how the API looks I'm going to push a little harder towards English for the method names. I have two proposale (in preference order):

1. `FFT` -> `Coefficients` and `IFFT` -> `Sequence`.
2. `FFT` -> `Analysis` and `IFFT` -> `Synthesis` (or the verbs).

This gives us nicer API for the new routines mentioned above, e.g. for QW-FFT, `SinCoefficients` and `CosCoefficients` (the methods are on the same type, `QuarterWaveFFT`).

fourier/rfft.go
```diff
@@ -0,0 +1,1151 @@
...
+	if len(ifac) < 15 {
+		panic("fourier: short ifac")
+	}
+	if n == 1 {
+		return
+	}
+	rffti1(n, work[n:2*n], ifac[:15])
+}
+
+func rffti1(n int, wa []float64, ifac []int) {
```
btracey: Checked this one and looks good.

fourier/rfft.go
```diff
@@ -0,0 +1,1151 @@
...
+	if len(ifac) < 15 {
+		panic("fourier: short ifac")
+	}
+	if n == 1 {
+		return
+	}
+	rfftf1(n, r[:n], work[:n], work[n:2*n], ifac[:15])
+}
+
+func rfftf1(n int, c, ch, wa []float64, ifac []int) {
```
btracey: This looks good too

fourier/rfft.go
```diff
@@ -0,0 +1,1151 @@
...
+		ch3.set(0, 0, k, cc3.at(0, k, 0)+cc3.at(0, k, 1))
+		ch3.set(ido-1, 1, k, cc3.at(0, k, 0)-cc3.at(0, k, 1))
+	}
+	if ido < 2 {
+		return
+	}
+	if ido > 2 {
+		idp2 := ido + 1
+		for k := 0; k < l1; k++ {
+			for i := 2; i < ido; i += 2 {
```
btracey: Shouldn't this be `i <= ido`? Looks like `idp2` is zero-indexed (only `ido + 1`)

kortschak: That would suggest all of the loops with an iteration bound of `ido` would be converted to `<=`. That change results in a panic.

kortschak: `idp2` is badly named, but the name is retained to help track the code.

kortschak: I should probably explain the process I used to make the transformation from 1-based to 0-based. For each function in turn I swapped from the 1-based types to the 0-based types, fixing up the indexing at the call site by adding `+1` to the indexing parameters. I then algebraically smoothed out the change, simplifying as much as I could. Also note that `ido` is defined in the outer functions as `n / l2`. These functions are not modified in either the trivial or subtle commits.

fourier/rfft.go
```diff
@@ -0,0 +1,1151 @@
...
+		ti1 := -hsqt2 * (cc3.at(ido-1, k, 1) + cc3.at(ido-1, k, 3))
+		tr1 := hsqt2 * (cc3.at(ido-1, k, 1) - cc3.at(ido-1, k, 3))
+		ch3.set(ido-1, 0, k, tr1+cc3.at(ido-1, k, 0))
+		ch3.set(ido-1, 2, k, cc3.at(ido-1, k, 0)-tr1)
+		ch3.set(0, 1, k, ti1-cc3.at(ido-1, k, 2))
+		ch3.set(0, 3, k, ti1+cc3.at(ido-1, k, 2))
+	}
+}
+
+func radf5(ido, l1 int, cc, ch, wa1, wa2, wa3, wa4 []float64) {
```
btracey: This looks good too.

kortschak:
*ping*

btracey:
I haven't forgotten, just a really big PR. Sorry it's taking so long.

kortschak:
No worries. Sorry about the size, there was not really much option on that.

btracey:
It's okay. I should be more proactive about getting through it. It's a very good thing to have in the suite.

fourier/rfft.go
```diff
@@ -0,0 +1,1151 @@
...
+			ch3.set(ic, 1, k, ti5-ti2)
+			ch3.set(i-1, 4, k, tr3+tr4)
+			ch3.set(ic-1, 3, k, tr3-tr4)
+			ch3.set(i, 4, k, ti3+ti4)
+			ch3.set(ic, 3, k, ti4-ti3)
+		}
+	}
+}
+
+func radfg(ido, ip, l1, idl1 int, cc, c1, c2, ch, ch2, wa []float64) {
```
btracey: Looks good

fourier/rfft.go
```diff
@@ -0,0 +1,1151 @@
...
+//
+//  This transform is unnormalized since a call of rfftf
+//  followed by a call of rfftb will multiply the input
+//  sequence by n.
+//
+//  work   Contains results which must not be destroyed between
+//         calls of rfftf or rfftb.
+//  ifac   Contains results which must not be destroyed between
+//         calls of rfftf or rfftb.
+func rfftb(n int, r, work []float64, ifac []int) {
```
btracey: Looks good.

fourier/rfft.go
```diff
@@ -0,0 +1,1151 @@
...
+	if len(ifac) < 15 {
+		panic("fourier: short ifac")
+	}
+	if n == 1 {
+		return
+	}
+	rfftb1(n, r[:n], work[:n], work[n:2*n], ifac[:15])
+}
+
+func rfftb1(n int, c, ch, wa []float64, ifac []int) {
```
btracey: Looks good

fourier/rfft.go
```diff
@@ -0,0 +1,1151 @@
...
+	}
+
+	if na {
+		for i := 0; i < n; i++ {
+			c[i] = ch[i]
+		}
+	}
+}
+
+func radb2(ido, l1 int, cc, ch, wa1 []float64) {
```
btracey: Looks good except...

fourier/rfft.go
```diff
@@ -0,0 +1,1151 @@
...
+				ch3.set(i, k, 1, wa1[i-2]*ti2+wa1[i-1]*tr2)
+			}
+		}
+
+		if ido%2 == 1 {
+			return
+		}
+	}
+	for k := 0; k < l1; k++ {
+		ch3.set(ido-1, k, 0, cc3.at(ido-1, 0, k)+cc3.at(ido-1, 0, k))
```
btracey: Shouldn't these just be `2*cc3.at(` looks like the same call twice.

kortschak: Yes, damn fortraners.

kortschak:
BTW, I plan to rewrite history before merge, but not just a complete squash since I want to retain the 1-based code in the tree in case we find a bug later. So an informal approval first is fine and then after I rewrite the history a real approval.

btracey:
That sounds good. I hope you don't mind the partial reviews, it's just easiest to remember what I've looked at.

Was the cfft file independently translated? Should I go through that once I'm through rfft?

kortschak:
I'm very happy with how you're doing the review. Yes, cfft was translated independently so it also needs a review.

fourier/rfft.go
```diff
@@ -0,0 +1,1151 @@
...
+		tr1 := cc3.at(ido-1, 0, k) - cc3.at(ido-1, 2, k)
+		tr2 := cc3.at(ido-1, 0, k) + cc3.at(ido-1, 2, k)
+		ch3.set(ido-1, k, 0, tr2+tr2)
+		ch3.set(ido-1, k, 1, math.Sqrt2*(tr1-ti1))
+		ch3.set(ido-1, k, 2, ti2+ti2)
+		ch3.set(ido-1, k, 3, -math.Sqrt2*(tr1+ti1))
+	}
+}
+
+func radb5(ido, l1 int, cc, ch, wa1, wa2, wa3, wa4 []float64) {
```
btracey: Looks good except...

fourier/rfft.go
```diff
@@ -0,0 +1,1151 @@
...
+			di3 := ci2 - cr3
+			ch3.set(i-1, k, 1, wa1[i-2]*dr2-wa1[i-1]*di2)
+			ch3.set(i, k, 1, wa1[i-2]*di2+wa1[i-1]*dr2)
+			ch3.set(i-1, k, 2, wa2[i-2]*dr3-wa2[i-1]*di3)
+			ch3.set(i, k, 2, wa2[i-2]*di3+wa2[i-1]*dr3)
+		}
+	}
+}
+
+func radb4(ido, l1 int, cc, ch, wa1, wa2, wa3 []float64) {
```
btracey: Looks good.

fourier/rfft.go
```diff
@@ -0,0 +1,1151 @@
...
+			return
+		}
+	}
+	for k := 0; k < l1; k++ {
+		ch3.set(ido-1, k, 0, 2*cc3.at(ido-1, 0, k))
+		ch3.set(ido-1, k, 1, -2*cc3.at(0, 1, k))
+	}
+}
+
+func radb3(ido, l1 int, cc, ch, wa1, wa2 []float64) {
```
btracey: Looks good.

fourier/rfft.go
```diff
@@ -0,0 +1,1151 @@
...
+		ch3.set(ido-1, k, 2, ti2+ti2)
+		ch3.set(ido-1, k, 3, -math.Sqrt2*(tr1+ti1))
+	}
+}
+
+func radb5(ido, l1 int, cc, ch, wa1, wa2, wa3, wa4 []float64) {
+	const (
+		tr11 = 0.309016994374947
+		ti11 = 0.951056516295154
+		tr12 = -0.809016994374947
```
btracey: I'm confused. The fortran line is `  1-.809016994374947,.587785252292473/`, which implies `tr12` is incorrect, but there's also a random `1` in the line above it. I've been searching for fortran line continuation information, but haven't found anything. You're sure this is right?

fourier/rfft.go
```diff
@@ -0,0 +1,1151 @@
...
+			ch3.set(i, k, 2, wa2[i-2]*di3+wa2[i-1]*dr3)
+			ch3.set(i-1, k, 3, wa3[i-2]*dr4-wa3[i-1]*di4)
+			ch3.set(i, k, 3, wa3[i-2]*di4+wa3[i-1]*dr4)
+			ch3.set(i-1, k, 4, wa4[i-2]*dr5-wa4[i-1]*di5)
+			ch3.set(i, k, 4, wa4[i-2]*di5+wa4[i-1]*dr5)
+		}
+	}
+}
+
+func radbg(ido, ip, l1, idl1 int, cc, c1, c2, ch, ch2, wa []float64) {
```
btracey: Looks good except...

fourier/rfft.go
```diff
@@ -0,0 +1,1151 @@
...
+	}
+
+	is := -ido - 1
+	if nbd > l1 {
+		for j := 1; j < ip; j++ {
+			is += ido
+			for k := 0; k < l1; k++ {
+				idij := is
+				for i := 2; i < ido; i += 2 {
+					idij = idij + 2
```
btracey: `idij += 2`

fourier/rfft.go
kortschak: The Fortran
```
      DATA TR11,TI11,TR12,TI12 /.309016994374947,.951056516295154,
     1-.809016994374947,.587785252292473/
```
is a line continuation (the 1 is in the continuation column), so if the language did not have its archaic line length limits it would be

```
      DATA TR11,TI11,TR12,TI12 /.309016994374947,.951056516295154,-.809016994374947,.587785252292473/
```

which is expressible as pseudo go as

```
const tr11, ri11, tr12, ri12 = 0.309016994374947, 0.951056516295154, -0.809016994374947, 0.587785252292473
```

fourier/rfft.go
kortschak: Fixed

fourier/cfft.go
```diff
@@ -0,0 +1,646 @@
...
+			tr2 := cc3.at(i-1, 0, k) - cc3.at(i-1, 1, k)
+			ch3.set(i, k, 0, cc3.at(i, 0, k)+cc3.at(i, 1, k))
+			ti2 := cc3.at(i, 0, k) - cc3.at(i, 1, k)
+			ch3.set(i, k, 1, wa1[i-1]*ti2+sign*wa1[i]*tr2)
+			ch3.set(i-1, k, 1, wa1[i-1]*tr2-sign*wa1[i]*ti2)
+		}
+	}
+}
+
+func pass3(ido, l1 int, cc, ch, wa1, wa2 []float64, sign float64) {
```
btracey: Looks good

fourier/cfft.go
```diff
@@ -0,0 +1,646 @@
...
+			di3 := ci2 - cr3
+			ch3.set(i, k, 1, wa1[i-1]*di2+sign*wa1[i]*dr2)
+			ch3.set(i-1, k, 1, wa1[i-1]*dr2-sign*wa1[i]*di2)
+			ch3.set(i, k, 2, wa2[i-1]*di3+sign*wa2[i]*dr3)
+			ch3.set(i-1, k, 2, wa2[i-1]*dr3-sign*wa2[i]*di3)
+		}
+	}
+}
+
+func pass4(ido, l1 int, cc, ch, wa1, wa2, wa3 []float64, sign float64) {
```
btracey: Looks good.

fourier/cfft.go
```diff
@@ -0,0 +1,646 @@
...
+			ch3.set(i, k, 1, wa1[i-1]*ci2+sign*wa1[i]*cr2)
+			ch3.set(i-1, k, 2, wa2[i-1]*cr3-sign*wa2[i]*ci3)
+			ch3.set(i, k, 2, wa2[i-1]*ci3+sign*wa2[i]*cr3)
+			ch3.set(i-1, k, 3, wa3[i-1]*cr4-sign*wa3[i]*ci4)
+			ch3.set(i, k, 3, wa3[i-1]*ci4+sign*wa3[i]*cr4)
+		}
+	}
+}
+
+func pass5(ido, l1 int, cc, ch, wa1, wa2, wa3, wa4 []float64, sign float64) {
```
btracey: Looks good here.

fourier/cfft.go
```diff
@@ -0,0 +1,646 @@
...
+			ch3.set(i, k, 2, wa2[i-1]*di3+sign*wa2[i]*dr3)
+			ch3.set(i-1, k, 3, wa3[i-1]*dr4-sign*wa3[i]*di4)
+			ch3.set(i, k, 3, wa3[i-1]*di4+sign*wa3[i]*dr4)
+			ch3.set(i-1, k, 4, wa4[i-1]*dr5-sign*wa4[i]*di5)
+			ch3.set(i, k, 4, wa4[i-1]*di5+sign*wa4[i]*dr5)
+		}
+	}
+}
+
+func pass(ido, ip, l1, idl1 int, cc, c1, c2, ch, ch2, wa []float64, sign float64) (nac bool) {
```
btracey: Looks good.

kortschak:
>I think we should put the fftpack code in its own subfolder.

Sure, I'll put it in `./internal/fftpack`.

btracey:
Taking a quick glance, this does feel somewhat overengineered to me, though I don't know how it looks when you add in the other transforms. Two things in particular

1) It feels like `work` should be its own type. It have `Reset` and `Len` as methods. Then the API would be something like

```
func FFT(dst []complex128, seq []float64, work *Work) []complex128
```
The user can use `nil` if they don't care about saving the memory, or can allocate a work vector as now. The work struct can be shared between `FFT` and `CmplxFFT`

2) I suspect that `FFT.Freq` and `CmplxFFT.Shift` are useful on their own outside the context of the type, and likely have analogies in other transforms (happy to be corrected if wrong though). Seems better to have them be standalone functions.

fourier/fourier.go
```diff
@@ -0,0 +1,254 @@
...
+
+// NewFFT returns an FFT initialized for work on sequences of length n.
+func NewFFT(n int) *FFT {
+	var t FFT
+	t.Reset(n)
+	return &t
+}
+
+// Len returns the length of the acceptable input.
+func (t *FFT) Len() int { return len(t.work) / 2 }
```
btracey: Why not `len(t.real)`?

fourier/fourier.go
```diff
@@ -0,0 +1,254 @@
...
+	} else {
+		t.work = make([]float64, 2*n)
+		t.real = make([]float64, n)
+	}
+	rffti(n, t.work, t.ifac[:])
+}
+
+// FFT computes the Fourier coefficients of the input sequence, seq,
+// placing the result in dst and returning it. This transform is
+// unnormalized since a call to FFT followed by a call of IFFT will
```
btracey: "This transform is unnormalized, _so_ a call"?

fourier/fourier.go
```diff
@@ -0,0 +1,254 @@
...
+//
+// If the length of seq is not t.Len(), FFT will panic.
+// If dst is nil, a new slice is allocated and returned. If dst is not nil and
+// the length of dst does not equal t.Len()/2+1, FFT will panic.
+func (t *FFT) FFT(dst []complex128, seq []float64) []complex128 {
+	if len(seq) != t.Len() {
+		panic("fourier: sequence length mismatch")
+	}
+	if dst == nil {
+		dst = make([]complex128, t.Len()/2+1)
```
btracey: It's weird thet t.Len() and len(seq) are both being used here for the same reason. I would pick one.

fourier/fourier.go
```diff
@@ -0,0 +1,254 @@
...
+	}
+	for i := 1; i < len(dst)-1; i++ {
+		dst[i] = complex(t.real[2*i-1], t.real[2*i])
+	}
+	return dst
+}
+
+// IFFT computes the real perodic sequence from the Fourier coefficients,
+// coeff, placing the result in dst and returning it. This transform is
+// unnormalized since a call to FFT followed by a call of IFFT will
```
btracey: so

fourier/fourier.go
```diff
@@ -0,0 +1,254 @@
...
+
+// FFT computes the Fourier coefficients of a complex input sequence,
+// seq, placing the result in dst and returning it. This transform is
+// unnormalized since a call to FFT followed by a call of IFFT will
+// multiply the input sequence by the length of the sequence.
+//
+// If the length of seq is not t.Len(), FFT will panic.
+// If dst is nil, a new slice is allocated and returned. If dst is not nil and
+// the length of dst does not equal the length of seq, FFT will panic.
+// It is safe to use the same slice for dst and seq.
```
btracey: Is it not safe for other calls? It's weird that this one specifically says it but the others don't.

fourier/fourier.go
```diff
@@ -0,0 +1,254 @@
...
+	cfftf(len(dst), t.real, t.work, t.ifac[:])
+	for i := range dst {
+		dst[i] = complex(t.real[2*i], t.real[2*i+1])
+	}
+	return dst
+}
+
+// IFFT computes the complex perodic sequence from the Fourier coefficients,
+// coeff, placing the result in dst and returning it. This transform is
+// unnormalized since a call to FFT followed by a call of IFFT will
```
btracey: so

fourier/fourier_example_test.go
```diff
@@ -0,0 +1,187 @@
...
+	"fmt"
+	"math/cmplx"
+
+	"gonum.org/v1/gonum/floats"
+	"gonum.org/v1/gonum/fourier"
+	"gonum.org/v1/gonum/mat"
+)
+
+func ExampleFFT_FFT() {
+	// Period is a set of samples over a given period.
```
btracey: Period is a set of values observed at regular intervals over a given period.

fourier/fourier_example_test.go
```diff
@@ -0,0 +1,187 @@
...
+	// freq=0 cycles/period, magnitude=9, phase=0
+	// freq=0.125 cycles/period, magnitude=3, phase=3.142
+	// freq=0.25 cycles/period, magnitude=1, phase=-0
+	// freq=0.375 cycles/period, magnitude=3, phase=3.142
+	// freq=0.5 cycles/period, magnitude=9, phase=0
+}
+
+func ExampleCmplxFFT_FFT() {
+	// Period is a set of samples over a given period.
+	period := []complex128{1, 0, 2, 0, 4, 0, 2, 0}
```
btracey: Maybe have some values that are actually complex?

Observed values and phases?

fourier/fourier_example_test.go
```diff
@@ -0,0 +1,187 @@
...
+	// freq=-0.375 cycles/period, magnitude=3, phase=3.142
+	// freq=-0.25 cycles/period, magnitude=1, phase=0
+	// freq=-0.125 cycles/period, magnitude=3, phase=3.142
+	// freq=0 cycles/period, magnitude=9, phase=0
+	// freq=0.125 cycles/period, magnitude=3, phase=3.142
+	// freq=0.25 cycles/period, magnitude=1, phase=0
+	// freq=0.375 cycles/period, magnitude=3, phase=3.142
+}
+
+func Example_fFT2() {
```
btracey: Are these supposed to have more in the name? Like `ExampleFFT_fFT2`?

fourier/fourier_example_test.go
```diff
@@ -0,0 +1,187 @@
+// Copyright ©2018 The Gonum Authors. All rights reserved.
```
btracey: To be honest, I don't understand the image example, though I'm sure it's just unfamiliarity with the domain. I do think it would be useful to have a simple FFT of a sine wave or something to show the peak.

fourier/fourier_test.go
```diff
@@ -0,0 +1,234 @@
+// Copyright ©2018 The Gonum Authors. All rights reserved.
```
btracey: It would be nice to have one bigger input that we compare with numpy, like a square wave example or something.

fourier/fourier_example_test.go
kortschak: Sure, I can add that, though that is essentially what the examples on the types show - these sequence is a sum of periodic functions in both of those.

fourier/fourier_test.go
kortschak: OK

fourier/fourier_example_test.go
kortschak: This is a package-level example for people wanting 2D FFT since we don't provide it.

fourier/fourier_example_test.go
kortschak: >Maybe have some values that are actually complex?

From what I've seen that's actually pretty rare except when the input to the complex FFT is an output from another FFT.

>Observed values and phases?

?

fourier/fourier.go
```diff
@@ -0,0 +1,254 @@
...
+// Freq will panic if i is negative or greater than or equal to t.Len().
+func (t *FFT) Freq(i int) float64 {
+	if i < 0 || t.Len() <= i {
+		panic("fourier: index out of range")
+	}
+	step := 1 / float64(t.Len())
+	return step * float64(i)
+}
+
+// CmplxFFT implements Fast Fourier Transform and its inverse for real sequences.
```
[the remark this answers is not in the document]
kortschak: Done

fourier/fourier.go
```diff
@@ -0,0 +1,254 @@
...
+	}
+	return step * float64(i-t.Len())
+}
+
+// Shift returns returns a shifted index into a slice of
+// coefficients returned by the CmplxFFT so that indexing
+// into the coefficients places the zero frequency component
+// at the center of the spectrum. Shift will panic if i is
+// negative or greater than or equal to t.Len().
+func (t *CmplxFFT) Shift(i int) int {
```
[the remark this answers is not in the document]
kortschak: OK, but what is the purpose of `Shift`?

Done

fourier/fourier_example_test.go
btracey: It just seems weird to have an example that basically uses the same data and gives the same answers as the FFT one.

fourier/fourier.go
kortschak: Changed

fourier/fourier_example_test.go
btracey: Oh, I see. Could you provide a header description to that effect? It's not at all clear to me that is what I was supposed to learn.

fourier/fourier.go
kortschak: I'm reflecting the original FFTPACK documentation. I think it comes down to a question of causality; the behaviour of the implementation is that is causes a multiplication, so it is therefore unnormalised. The the alternative wording is that has unnormalised behaviour and so it is therefor causing a multiplication. I think the former is marginally tighter.

fourier/fourier.go
kortschak: Done

fourier/fourier.go
kortschak: Argument above.

fourier/fourier.go
kortschak: This is the only call that can possibly have `dst` and `seq` be the same slice since the real FFT calls don't have matching types for these two parameters.

fourier/fourier.go
kortschak: As above.

fourier/fourier_example_test.go
kortschak: The normal terminology is "sample".

fourier/fourier_example_test.go
kortschak: :) I'd be very upset if they didn't.

kortschak:
Do you have a view on the method naming I suggested above?

btracey:
Could we leave it to the next PR? I have a strong bias toward FFT because that is what everyone will be looking for. I'm not opposed to better names, but it's so established that I think alternate names will more confusing to read for many. I assume your argument makes more sense in the broader context though.

kortschak:
OK. With the current API, `FFT` and `CmplxFFT` are clearly present though.

btracey:
Oh, I see, the idea is to keep the type names, but change the method names. I misunderstood.

fourier/fourier.go
btracey: maybe "in that"? Doesn't assign any causality, just a fact.

kortschak:
PTAL

kortschak:
>Oh, I see, the idea is to keep the type names, but change the method names. I misunderstood.

Yes, so a call would look like `fft.Coefficients(nil, samples)` and the reverse would be `fft.Sequence(nil, coeffs)`.

fourier/fourier.go
kortschak: I'm a fan of the ';' for this; that is a what a semicolon means.

btracey:
Oh, sorry, there was also the comment about moving the FFT code into ./internal/fftpack . That can either happen now or in a future PR.

fourier/fourier_example_test.go
```diff
@@ -0,0 +1,223 @@
...
+	// freq=-0.375 cycles/period, magnitude=3, phase=3.142
+	// freq=-0.25 cycles/period, magnitude=1, phase=0
+	// freq=-0.125 cycles/period, magnitude=3, phase=3.142
+	// freq=0 cycles/period, magnitude=9, phase=0
+	// freq=0.125 cycles/period, magnitude=3, phase=3.142
+	// freq=0.25 cycles/period, magnitude=1, phase=0
+	// freq=0.375 cycles/period, magnitude=3, phase=3.142
+}
+
+func Example_fFT2() {
```
vladimir-ch: Lower-case f in fFT2

btracey: That's to give it package scope. See the example names section https://blog.golang.org/examples

vladimir-ch: Oh, there's always something to learn. Thanks.

kortschak:
Thanks. There are a few of things to do; I forgot the comment that the cfft backend code is refactored to allow direction being set by sign, the change `s/since/;/g` (@btracey are you happy with this?) and a lot of delicate history revision including creating internal/fftpack (I'm not confident about whether this is possible in this PR).

On the subject of the internal/fftpack package: I know we've done this kind of thing in mathext, but I'm really not convinced it's a good idea. The package boundary is to define separation of concerns and modularity. I don't see that division here except as a consequence of history. Though maybe it does help in debugging with other implementations. Comments?

kortschak:
I've done as much history condensation as I think is achievable while retaining the capacity to identify bugs later.

btracey:
Semicolon sounds great.

I think it's a separation of concerns. The code base isn't really "native go", just translated fortran, while the main package is novel code. The FFTPack code has additional "license" documentation that goes with it that the other code does not have. I think it's also good to separate for other implementations, in particular it would be nice to clean up a lot of the FFTPack code (i.e. using `complex128` where appropriate) at which point those alternate functions could live in the main package, preferably with comment styles that match gonum rather than matching netlib. In short, the FFTPACK code feels out of place, and putting it in internal is a clear signal that it's somehow different.

kortschak:
Actually, the history revision is non-trivial to the extent that I think I will just add another commit moving the fftpack code out to the internal package. There is too much breakage opportunity doing it the way I initially wanted.

kortschak:
PTAL

kortschak:
Thanks for the review.

PR #414, opened 2018-02-09
Fixes #413

### LINKED ISSUE #413: floats: Nearest() produces strange results with NaNs
When running this:
```
a := []float64{math.NaN(), 1, 2, 3}
b := 2
c := floats.Nearest(a, b)
```
Nearest returns 0, where I would expect 2. 
Since it's a simple matching, NaN can be also matched to a first occurrence in a slice like 
```
a := []float64{math.NaN(), 1, 2, 3}
b := math.NaN()
c := floats.Nearest(a, b)
// c == 0
```

aleh-null:
Please take a look.

Fixes #413

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -16,6 +16,7 @@
 # Please keep the list sorted.
 
 Andrew Brampton <brampton@gmail.com>
+Aleh Plashchynski <plashchynskialeh@gmail.com>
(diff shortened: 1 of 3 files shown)


kortschak:
This fix is premature since the correct behaviour has not yet been specified.

(Also, please follow the pull request template)

kortschak:
Closing in favour of #416.

PR #416, opened 2018-02-13
floats: fix nearest, span and min/max index behaviours

### LINKED ISSUE #413: floats: Nearest() produces strange results with NaNs

kortschak:
Please take a look.

I have not really touched the documentation, but I think that will probably need work too. Comments welcome for that.

Fixes #413.

/cc @aleh-null

### DIFF
--- a/floats/floats.go
+++ b/floats/floats.go
@@ -433,13 +433,13 @@ func MaxIdx(s []float64) int {
 	if len(s) == 0 {
 		panic("floats: zero slice length")
 	}
-	max := math.Inf(-1)
+	max := math.NaN()
 	var ind int
 	for i, v := range s {
 		if math.IsNaN(v) {
 			continue
 		}
-		if v > max {
+		if v > max || math.IsNaN(max) {
 			max = v
 			ind = i
 		}
@@ -459,13 +459,13 @@ func MinIdx(s []float64) int {
 	if len(s) == 0 {
 		panic("floats: zero slice length")
 	}
-	min := math.Inf(1)
+	min := math.NaN()
 	var ind int
 	for i, v := range s {
 		if math.IsNaN(v) {
 			continue
 		}
-		if v < min {
+		if v < min || math.IsNaN(min) {
 			min = v
 			ind = i
 		}
@@ -538,14 +538,14 @@ func NearestIdx(s []float64, v float64) int {
 		return MinIdx(s)
 	}
 	var ind int
-	dist := math.Inf(1)
+	dist := math.NaN()
 	for i, val := range s {
 		newDist := math.Abs(v - val)
 		// A NaN distance will not be closer.
 		if math.IsNaN(newDist) {
 			continue
 		}
-		if newDist < dist {
+		if newDist < dist || math.IsNaN(dist) {
 			dist = newDist
 			ind = i
 		}
@@ -815,9 +815,14 @@ func Scale(c float64, dst []float64) {
 // Span returns a set of N equally spaced points between l and u, where N
 // is equal to the length of the destination. The first element of the destination
 // is l, the final element of the destination is u.
+//
+// If l and u  are infinities of opposite sign, then on return dst will contain half
+// of one infinity, and half the other, with the middle element being zero if dst is odd.
+//
 // Panics if len(dst) < 2.
 //
-// Also returns the mutated slice dst, so that it can be used in range expressions, like:
+// Span also returns the mutated slice dst, so that it can be used in range expressions,
+// like:
 //
 //     for i, x := range Span(dst, l, u) { ... }
 func Span(dst []float64, l, u float64) []float64 {
@@ -861,7 +866,7 @@ func Span(dst []float64, l, u float64) []float64 {
 		return dst
 	case math.IsInf(u, 0):
 		for i := range dst[1:] {
-			dst[i] = u
+			dst[i+1] = u
 		}
 		dst[0] = l
 		return dst
--- a/floats/floats.go
+++ b/floats/floats.go
@@ -521,11 +521,11 @@ func NaNPayload(f float64) (payload uint64, ok bool) {
 	return b &^ nanMask, true
 }
 
-// Nearest returns the index of the element in s
+// NearestIdx returns the index of the element in s
 // whose value is nearest to v.  If several such
 // elements exist, the lowest index is returned.
-// Nearest panics if len(s) == 0.
-func Nearest(s []float64, v float64) int {
+// NearestIdx panics if len(s) == 0.
+func NearestIdx(s []float64, v float64) int {
 	if len(s) == 0 {
 		panic("floats: zero length slice")
 	}
@@ -538,18 +538,14 @@ func Nearest(s []float64, v float64) int {
 		return MinIdx(s)
 	}
 	var ind int
-	dist := math.Abs(v - s[0])
+	dist := math.Inf(1)
 	for i, val := range s {
 		newDist := math.Abs(v - val)
 		// A NaN distance will not be closer.
 		if math.IsNaN(newDist) {
 			continue
 		}
-		// The inverted test here is used to
-		// ensure that a test of newDist against
-		// a NaN dist results in a change to dist
-		// and the current index value.
-		if !(newDist >= dist) {
+		if newDist < dist {
 			dist = newDist
 			ind = i
 		}
@@ -557,10 +553,13 @@ func Nearest(s []float64, v float64) int {
 	return ind
 }
 
-// NearestIdx return the index of a hypothetical vector created
+// NearestIdxForSpan return the index of a hypothetical vector created
 // by Span with length n and bounds l and u whose value is closest
-// to v. NearestIdx panics if n is zero.
-func NearestIdx(n int, l, u float64, v float64) int {
+// to v. That is, NearestIdxForSpan(n, l, u, v) is equivalent to
+// Nearest(Span(make([]float64, n),l,u),v) without an allocation and
+// with a relaxation for the paramater n.
+// NearestIdxForSpan panics if n is less than one.
+func NearestIdxForSpan(n int, l, u float64, v float64) int {
 	if n == 0 {
 		panic("floats: zero length span")
 	}
@@ -578,8 +577,14 @@ func NearestIdx(n int, l, u float64, v float64) int {
 		if l == u {
 			return 0
 		}
-		if n%2 == 1 && !math.IsInf(v, 0) {
-			return n / 2
+		if n%2 == 1 {
+			if !math.IsInf(v, 0) {
+				return n / 2
+			}
+			if math.Copysign(1, v) == math.Copysign(1, l) {
+				return 0
+			}
+			return n/2 + 1
 		}
 		if math.Copysign(1, v) == math.Copysign(1, l) {
 			return 0
--- a/floats/floats.go
+++ b/floats/floats.go
@@ -433,9 +433,12 @@ func MaxIdx(s []float64) int {
 	if len(s) == 0 {
 		panic("floats: zero slice length")
 	}
-	max := s[0]
+	max := math.Inf(-1)
 	var ind int
 	for i, v := range s {
+		if math.IsNaN(v) {
+			continue
+		}
 		if v > max {
 			max = v
 			ind = i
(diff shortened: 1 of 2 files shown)


floats/floats.go
```diff
@@ -515,31 +515,109 @@ func NaNPayload(f float64) (payload uint64, ok bool) {
...
+// Nearest panics if len(s) == 0.
 func Nearest(s []float64, v float64) int {
+	if len(s) == 0 {
+		panic("floats: zero length slice")
+	}
+	switch {
+	case math.IsNaN(v):
+		return 0
+	case math.IsInf(v, 1):
+		return MaxIdx(s)
```
aleh-null: There is an issue with `MaxInd` as it does not handle NaNs "properly" in case first element is NaN(). Following tests will fail:
```
		{
			in:    []float64{math.NaN(), 0, -1, 2},
			query: math.Inf(1),
			want:  3,
			desc:  "Wrong index returned when query is +Inf and no +Inf element exists",
		},
		{
			in:    []float64{math.NaN(), 0, -1, 2},
			query: math.Inf(-1),
			want:  2,
			desc:  "Wrong index returned when query is -Inf and no -Inf element exists",
		}
```

Changing [this](https://github.com/gonum/gonum/blob/master/floats/floats.go#L436) to something like this
```
	var (
		max float64
		idx int
	)
	allnan := true
	for k, v := range x {
		if !math.IsNaN(v) {
			allnan = false
			idx, max = k, v
			break
		}
	}
	if allnan {
		return 0
	}
```

floats/floats.go
```diff
@@ -515,31 +515,109 @@ func NaNPayload(f float64) (payload uint64, ok bool) {
...
+		if l == u {
+			return 0
+		}
+		if n%2 == 1 && !math.IsInf(v, 0) {
+			return n / 2
+		}
+		if math.Copysign(1, v) == math.Copysign(1, l) {
+			return 0
+		}
+		return n / 2
```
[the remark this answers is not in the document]
btracey: Are we sure this works for the case where n odd, v = +inf, and range = (-inf, +inf)?  Seems like it would return the index for 0 when it should be returning n/2+1

floats/floats.go
```diff
@@ -515,31 +515,109 @@ func NaNPayload(f float64) (payload uint64, ok bool) {
...
+// NearestIdx return the index of a hypothetical vector created
 // by Span with length n and bounds l and u whose value is closest
-// to v. NearestWithinSpan panics if u < l. If the value is greater than u or
-// less than l, the function returns -1.
-func NearestWithinSpan(n int, l, u float64, v float64) int {
-	if u < l {
-		panic("floats: upper bound greater than lower bound")
+// to v. NearestIdx panics if n is zero.
+func NearestIdx(n int, l, u float64, v float64) int {
+	if n == 0 {
```
btracey: I think we should also panic if `l > u` (and in `Span`)

floats/floats.go
```diff
@@ -733,6 +811,48 @@ func Span(dst []float64, l, u float64) []float64 {
 	if n < 2 {
 		panic("floats: destination must have length >1")
 	}
+
```
btracey: We'll need to add commentary about the NaN and Inf cases.

btracey: Panic if `l > u`?

floats/floats.go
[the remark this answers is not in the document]
btracey: Oh, looks like this was already pointed out. Thanks @aleh-null

floats/floats.go
```diff
@@ -515,31 +515,109 @@ func NaNPayload(f float64) (payload uint64, ok bool) {
 // Nearest returns the index of the element in s
 // whose value is nearest to v.  If several such
 // elements exist, the lowest index is returned.
-// Panics if len(s) == 0.
+// Nearest panics if len(s) == 0.
```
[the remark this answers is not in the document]
kortschak: Ah, that's what you meant.

floats/floats.go
```diff
@@ -515,31 +515,109 @@ func NaNPayload(f float64) (payload uint64, ok bool) {
...
+		return MaxIdx(s)
+	case math.IsInf(v, -1):
+		return MinIdx(s)
+	}
 	var ind int
 	dist := math.Abs(v - s[0])
 	for i, val := range s {
 		newDist := math.Abs(v - val)
-		if newDist < dist {
+		// A NaN distance will not be closer.
```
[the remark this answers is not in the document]
kortschak: The first loop panics for a completely NaN slice.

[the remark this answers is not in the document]
btracey: Yes, sorry, it should have `for i = 0; i < len(s); i++`

floats/floats.go
kortschak: Fixed.

floats/floats.go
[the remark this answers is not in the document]
kortschak: Done

floats/floats.go
```diff
@@ -515,31 +515,109 @@ func NaNPayload(f float64) (payload uint64, ok bool) {
...
+		if !(newDist >= dist) {
 			dist = newDist
 			ind = i
 		}
 	}
 	return ind
 }
 
-// NearestWithinSpan return the index of a hypothetical vector created
+// NearestIdx return the index of a hypothetical vector created
```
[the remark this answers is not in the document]
kortschak: Done

floats/floats.go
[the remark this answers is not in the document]
kortschak: Fixed

floats/floats.go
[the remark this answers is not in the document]
kortschak: The copysign is so that -0 is closer to -Inf.

floats/floats.go
kortschak: I don't see any basis for prohibiting a descending span.

floats/floats.go
kortschak: Same here.

floats/floats.go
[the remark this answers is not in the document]
kortschak: This is fixed.

kortschak:
PTAL

kortschak:
>I'll look at tests when we agree on other behavior

It's probably worth looking at the tests first since they form a pseudo specification. TDD actually work quite well in this situation.

aleh-null:
@kortschak @btracey can I help you in some way to close this one? It seems that everything was fixed besides documentation.

kortschak:
You can give an informal approval, but I'm waiting on a review from another committer.

aleh-null:
@kortschak LGTM

floats/floats_test.go
```diff
@@ -919,8 +1038,57 @@ func TestNearestWithinSpan(t *testing.T) {
...
+			upper:  math.Inf(1),
+			value:  0,
+			idx:    2,
+		},
+		{
+			length: 4,
+			lower:  math.Inf(-1),
+			upper:  math.Inf(1),
+			value:  math.Inf(1),
+			idx:    2,
```
vladimir-ch: Why is the expected `idx` not 3? It seems wrong that `0` and `inf` lead to same `idx`.

btracey: The (hypothetical) slice is []float64{-inf, -inf, inf, inf}. -0 is closer to -inf, +0 is closer to inf. This way they both return idx = 2.

vladimir-ch: Thanks, seeing the hypothetical slice explains it. But documenting the behavior regarding all the possible combinations of corner values and for the user to understand it, should this function really be in Gonum?

floats/floats_test.go
```diff
@@ -26,6 +27,26 @@ func AreSlicesEqual(t *testing.T, truth, comp []float64, str string) {
 	}
 }
 
+func AreSlicesSame(t *testing.T, truth, comp []float64, str string) {
```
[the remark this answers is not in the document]
kortschak: fd4ca875, it was copy pasted.

floats/floats_test.go
```diff
@@ -1159,6 +1327,57 @@ func TestSpan(t *testing.T) {
...
+			want: []float64{math.Inf(1), math.NaN(), math.NaN(), math.NaN(), math.NaN()},
+		},
+		{
+			n: 5, l: math.NaN(), u: math.Inf(-1),
+			want: []float64{math.NaN(), math.NaN(), math.NaN(), math.NaN(), math.Inf(-1)},
+		},
+		{
+			n: 5, l: math.NaN(), u: math.Inf(1),
+			want: []float64{math.NaN(), math.NaN(), math.NaN(), math.NaN(), math.Inf(1)},
+		},
```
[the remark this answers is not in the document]
kortschak: Done x6

floats/floats.go
```diff
@@ -433,9 +433,12 @@ func MaxIdx(s []float64) int {
 	if len(s) == 0 {
 		panic("floats: zero slice length")
 	}
-	max := s[0]
+	max := math.Inf(-1)
 	var ind int
 	for i, v := range s {
+		if math.IsNaN(v) {
```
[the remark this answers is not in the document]
kortschak: Also MinAndIdx

floats/floats_test.go
vladimir-ch: Why the hypothetical slice is `[]float64{-inf, -inf, inf, inf}`? If `n` is odd, does it have 0 in the middle? Why?

btracey: It seems like the only logical behavior since it's halfway between the two.

`Span` shouldn't be used if there are NaN and Inf involved. It is a very reasonable function without those limiting cases. However, the input cannot be finite numbers, so we have to have behavior with inf and nan inputs. We can either panic with those inputs, or try to do something sensible. It's not clear to me which is worse here, but neither is good.

floats/floats.go
```diff
@@ -512,34 +521,117 @@ func NaNPayload(f float64) (payload uint64, ok bool) {
...
 // by Span with length n and bounds l and u whose value is closest
-// to v. NearestWithinSpan panics if u < l. If the value is greater than u or
-// less than l, the function returns -1.
-func NearestWithinSpan(n int, l, u float64, v float64) int {
-	if u < l {
-		panic("floats: upper bound greater than lower bound")
+// to v. That is, NearestIdxForSpan(n, l, u, v) is equivalent to
+// Nearest(Span(make([]float64, n),l,u),v) without an allocation and
+// with a relaxation for the paramater n.
+// NearestIdxForSpan panics if n is less than one.
```
vladimir-ch: This should be `n is less than two`, shouldn't it?

floats/floats_test.go
vladimir-ch: > logical behavior since it's halfway between the two.

That's probably the essence of my unease with this. I don't think it's logical because I don't think there's anything halfway between -inf and +inf. Or that it is clear how a hypothetical slice from -inf to 3 (for example) looks like (I really don't know, I guess it would be something like `[]float{-inf, -inf, -inf, 3}`).

The output seems to be useful only if the inputs are finite numbers, so the user will not be able to blindly pass numbers in and use the result anyway, so I would vote for panicking. But much work has went into this and there are rules with reasoning behind them, so if they are documented, then LGTM too.

floats/floats.go
vladimir-ch: Or is this what "a relaxation" means? How the hypothetical vector looks like for `n = 1`?

btracey: Yes, I agree, this should panic for `n  < 2`, so that it matches `Span`

floats/floats_test.go
vladimir-ch: SGTM

floats/floats.go
kortschak: OK

floats/floats_test.go
kortschak: I will remove the documentation of the special cases.

kortschak:
PTAL

PR #418, opened 2018-02-14
graph: fix undirected to directed copy behaviour

### LINKED ISSUE #417: graph: Copy() of simple.UndirectedGraph to simple.DirectedGraph does not match documentation
Apply `graph.Copy()` with an `simple.UndirectedGraph` source and a `simple.DirectedGraph` destination.

(I imagine that this also affects `graph.CopyWeighted()` for the same reason outlined below.)

### What did you do?

https://gist.github.com/DryHumour/e735ebcfea35b0a73eed36eb357369b8

The documentation for `graph.Copy()` says, in part,

> If the source is undirected and the destination is directed both directions will be present in the destination after the copy is complete.

### What actually happened?
Only a single direction is present in the destination.

This seems to happen because the call by `graph.Copy()` to `src.Edge()` (in this case `simple.UndirectedGraph.Edge()`) will return exactly the same edge for both `Edge(u, v)` and `Edge(v, u)` (_i.e._ the endpoints in the result are always in the order supplied when the edge was originally created, regardless of the order in the arguments when retrieving it).  This results in `dst.SetEdge()` (in this case `simple.DirectedGraph.SetEdge()`) effectively overwriting the edge it set on the first call with the (apparently identical) edge on the second call.

I hesitate to suggest a PR since it is not clear to me where the responsibility for ordering the endpoints ought to lie.  Also, as always, any change to existing behaviour runs the risk of breaking existing applications relying on the current implementation.

I would guess that `graph.Copy()` should be using something more like `dst.SetEdge(dst.NewEdge(u, v))`, which would decouple the source graph's notion of edges from that of the destination graph's.  (Similarly for `graph.CopyWeighted()`.)  However, there may well be some good reason why this is not already so.

But it could also possibly be argued that `simple.UndirectedGraph` violates the contract for `Graph.Edge()` which says "... returns the edge from u to v ..." and so implies that `Edge(u, v).From()` should always return `u` and never `v`, which is not currently the case.

Or maybe both?  Neither?  Bueller?

`go version go1.9.2 linux/amd64`
996b88e8f8941c9defdbb338bea5e239a230e742

Yes

kortschak:
Please take a look.

Fixes #417.

/cc @DryHumour

### DIFF
--- a/graph/graph.go
+++ b/graph/graph.go
@@ -218,7 +218,7 @@ func Copy(dst Builder, src Graph) {
 	}
 	for _, u := range nodes {
 		for _, v := range src.From(u) {
-			dst.SetEdge(src.Edge(u, v))
+			dst.SetEdge(dst.NewEdge(u, v))
 		}
 	}
 }
@@ -241,7 +241,7 @@ func CopyWeighted(dst WeightedBuilder, src Weighted) {
 	}
 	for _, u := range nodes {
 		for _, v := range src.From(u) {
-			dst.SetWeightedEdge(src.WeightedEdge(u, v))
+			dst.SetWeightedEdge(dst.NewWeightedEdge(u, v, src.WeightedEdge(u, v).Weight()))
 		}
 	}
 }
--- a/graph/graph_test.go
+++ b/graph/graph_test.go
@@ -1,0 +1,298 @@
+// Copyright ©2017 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package graph_test
+
+import (
+	"sort"
+	"testing"
+
+	"gonum.org/v1/gonum/graph"
+	"gonum.org/v1/gonum/graph/internal/ordered"
+	"gonum.org/v1/gonum/graph/simple"
+)
+
+type graphBuilder interface {
+	graph.Graph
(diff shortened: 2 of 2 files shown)


kortschak:
PTAL

kortschak:
*ping*

graph/graph.go
```diff
@@ -218,7 +218,7 @@ func Copy(dst Builder, src Graph) {
 	}
 	for _, u := range nodes {
 		for _, v := range src.From(u) {
-			dst.SetEdge(src.Edge(u, v))
+			dst.SetEdge(dst.NewEdge(u, v))
```
mewmew: I've been curious to know, what would happen if we accidentally did this, but for nodes? That is add the `graph.Node` of one graph `G1` to another `G2`, without creating a new dedicated node for `G2`.

Good find with fixing the edges here @kortschak.

kortschak: >I've been curious to know, what would happen if we accidentally did this, but for nodes? That is add the `graph.Node` of one graph `G1` to another `G2`, without creating a new dedicated node for `G2`.

There is no reason not to do this - in fact the original code is not incorrect when the destinations and source match for whether they are directed or undirected. There are subtleties that are touched on in gonum/graph#87 though.

Not my find, #417 was found by @DryHumour.

PR #420, opened 2018-02-16
blas/gonum: rewrite drotmg

### LINKED ISSUE #415: blas/gonum: check if drotmg needs this fix
Some debugging happened on Drotmg in the OpenBLAS implementation. We should see if our implementation has the same issues 
https://github.com/xianyi/OpenBLAS/issues/1452

kortschak:
Please take a look.

This is a complete rewrite of Drotmg from the Fortran, staying closer to the structure [there](http://www.netlib.org/lapack/explore-html/df/deb/drotmg_8f_source.html). If fixes the behaviour that is referenced in #415, but introduces a need for changes to the expected values in testblas (and doesn't fix the errors in a translation of the netlib Fortran test code - which may have been translated wrong), and causes the OpenBLAS back-end to fail in netlib. However, it looks very much more like the original code.

Fixes #415.


btracey:
I'm pretty confused actually ... I wrote these tests comparing to the mac Accelerate framework, and found the OpenBLAS bug because the behavior is different. When I originally pulled this PR, I was getting a bunch of failures with both Accelerate and OpenBLAS. Then, when I pulled a fresh copy of OpenBLAS, I now only get a single error with both implementations. Is Accelerate somehow loading from OpenBLAS? If so, why was I seeing implementation differences before? Very confused.

blas/testblas/level1double.go
```diff
@@ -1772,7 +1772,7 @@ var DrotmgTests = []DrotmgTestStruct{
 		Name: "RD1_Big_RD2_Big_Flag_0",
 		P: &blas.DrotmParams{
 			Flag: blas.Rescaling,
-			H:    [4]float64{4096, -3584, 1792, 4096},
+			H:    [4]float64{4096, -4096, 1, 4096},
```
btracey: I'm getting failures here:

```
level1double.go:2109: drotm y_0 mismatch RD1_Big_RD2_Big_Flag_0: expected 0, found -4096
```

blas/gonum/level1double.go
```diff
@@ -326,122 +326,121 @@ func (Implementation) Drotg(a, b float64) (c, s, r, z float64) {
...
 // for more details.
 func (Implementation) Drotmg(d1, d2, x1, y1 float64) (p blas.DrotmParams, rd1, rd2, rx1 float64) {
-	var p1, p2, q1, q2, u float64
-
 	const (
 		gam    = 4096.0
-		gamsq  = 16777216.0
-		rgamsq = 5.9604645e-8
+		gamsq  = gam * gam
+		rgamsq = 1.0 / gamsq
```
btracey: It's not clear this should change. If the goal is to make it more like Fortran, rgamsq is not the same here because the Fortran rounds poorly.

blas/gonum/level1double.go
```diff
@@ -326,122 +326,121 @@ func (Implementation) Drotg(a, b float64) (c, s, r, z float64) {
...
-			p.H[1] = -1
-			p.H[2] = 1
-			p.Flag = blas.Rescaling
-		}
-		if rd1 <= rgamsq {
-			rd1 *= gam * gam
-			rx1 /= gam
-			p.H[0] /= gam
-			p.H[2] /= gam
+	if d1 != 0 {
```
btracey: Okay, so this doesn't exactly match the reference behavior, as commented in https://github.com/xianyi/OpenBLAS/issues/1452#issuecomment-366558013 . I agree that this behavior seems more reasonable.

blas/testblas/level1double.go
btracey: And I see that this is https://github.com/xianyi/OpenBLAS/issues/1452#issuecomment-366565894

btracey: Is it expected that our tests also fail?

kortschak: The OpenBLAS implementation fails in exactly the same way that this code does. So for some value of expectation, yes, it's expected. I'd like this not to fail and for us to be able to provide a solution to the OpenBLAS failure.

blas/gonum/level1double.go
kortschak: The goal is not to be more like the Fortran, since it is demonstrably wrong. I started with that as a way to identify where the problems were.

blas/testblas/level1double.go
kortschak: This is the only part that we need to resolve. It is related though to the new test that I have added in below (at L2104) and am not sure about; it does not look correct wrt the documentation, but is the only way I can get Drotm to zero out the y1 component. If you can help resolve that, we can probably finish this up.

kortschak: By way of notes, prior to the rescaling section of `Drotmg` the values of H would work correctly as a `blas.Rescaling` H for this input, so the issue is again in the rescaling part.

blas/gonum/level1double.go
btracey: Great.

blas/testblas/level1double.go
btracey: This is the penalty we reap for making independent tests

kortschak: Can you confirm that I have interpreted the documentation of `DROTMG` correctly in the additional test stanza at [L2104](https://github.com/gonum/gonum/blob/783e09b9285a8f952d51b722de5c168a0a0c5529/blas/testblas/level1double.go#L2104-L2111).

If that's not correct, we can just remove it or fix it and maybe this failure goes away. If it is correct, it should not be too difficult to figure out what in the rescaling code is being broken.

btracey: Thanks for the more direct pointer. I was looking this morning trying to figure it out, but did not succeed in the time I had. I'll take another stab.

blas/testblas/level1double.go
```diff
@@ -1971,6 +2100,15 @@ func DrotmgTest(t *testing.T, d Drotmger) {
 		if !dTolEqual(rx1, test.Rx1) {
 			t.Errorf("drotmg rx1 mismatch %v: expected %v, found %v", test.Name, test.Rx1, rx1)
 		}
+
+		y := []float64{test.Y1}
```
btracey: Sorry, I'm confused. Isn't this a test of `Drotm` and not `Drotmg`? As far as I can tell this only uses data from Drotm

btracey: Nope, I'm an idiot, sorry.

blas/testblas/level1double.go
btracey: I must be wrong, because my interpretation fails all of the tests. But, if we look at this documentation https://software.intel.com/en-us/mkl-developer-reference-fortran-rotmg , it seems like the vector of interest is `x1*sqrt(d1)` and `y1*sqrt(d2)` which is loosely confirmed by the fortran documentation. Because of that, I don't see why the vector you're supplying is `x1` and `y1`

kortschak: Indeed. Mind you the Fortran documentation is pretty terrible. I also looked at the [cublas documentation](https://pyculib.readthedocs.io/en/latest/cublas.html#pyculib.blas.Blas.rotmg) which is what I was working from - this still doesn't agree with what I have here, but what I have here is the only thing that zeros out the second entry (I tried what is actually described - unless I'm wildly off the track).

btracey: The code is certainly doing something with the terms `x1 * sqrt(d1)` and such, but it's convoluted.

Rather than read what was going on, I tried to just solve the problem myself, but I really don't get it. `H` has four degrees of freedom. If the requirement is just to zero out the second component, then there is one constraint. If there is also a requirement on the first component (say, equal to x1), then there are two constraints. Either way, this seems like an underconstrained problem . A normal givens rotation only has two elements, `c` and `s`, but that doesn't seem like what's going on here.

kortschak: I've pushed a change that reflects the documentation in the use of`Drotm`, which fails more, and also causes NaNs to be introduced when `d1` or `d2` are negative. It would be great if the original documentation were more informative. Is there a paper for this?

vladimir-ch: I don't have a paper but there is a paragraph in Björck (2015): Numerical methods in matrix computations:
```
It is possible to rearrange the plane rotations so that only two instead of four multi-
plications per element are used and no square roots are required. These modified trans-
formations, called fast Givens transformations, are due to Gentleman [115, 1973]
and Hammarling [146, 1974]. The basic idea is to take out a scaling factor and write
G = cQ = c
1 s/c
−s/c 1


or G = s Q = s
c/s 1
−1 c/s,


(2.3.18)
depending on whether |c| > |s| or |c|≤|s|. In a product of rotations G1 ··· Gk the
scaling factors ci and si are accumulated separately. A dynamic scaling has been
suggested by Anda and Park [1, 1994]. On modern processors the gain in speed is
modest. Because of this and the nontrivial amount of monitoring needed to avoid
overflow and underflow, the usefulness of fast Givens transformations appears to be
limited and LAPACK does not make use of them.
```
The Anda and Park reference is:
Fast plane rotations with dynamic scaling. SIAM J. Matrix Anal Appl 15, 162-714
But I don't have access to it so not sure if relevant or helpful. Perhaps the older papers.

blas/testblas/level1double.go
```diff
@@ -1971,6 +2100,22 @@ func DrotmgTest(t *testing.T, d Drotmger) {
...
+
+		// Drotmg routines compute the components of a modified Givens transformation
+		// matrix H that zeros the y-component of the resulting vector,
+		//
+		//  [x1; 0] := H[x1 sqrt(d1); y1 sqrt(d2)].
+		//
+		// Drotm performs a modified Givens rotation of points in the plane,
+		//
+		//  [x1; y1] := H[x1; y1].
+		y := []float64{test.Y1 * math.Sqrt(test.D2)}
```
vladimir-ch: I think the `Sqrt` should not be here based at least on the rest of the test. We have `test.Rx1` that we want and `rx1` matches it. If I remove the sqrt, `x[0]` also matches it.

I will also look at this but I already concur that the documentation is quite bad.

blas/testblas/level1double.go
```diff
@@ -1971,6 +2100,22 @@ func DrotmgTest(t *testing.T, d Drotmger) {
...
+		d.Drotm(1, []float64{test.X1 * math.Sqrt(test.D1)}, 1, y, 1, p)
+		for i, v := range y {
+			if !dTolEqual(v, 0) {
+				t.Errorf("drotm y_%d mismatch %v: expected 0, found %v", i, test.Name, y[0])
```
vladimir-ch: I think you want `s/y[0]/v`.

blas/testblas/level1double.go
vladimir-ch: > H has four degrees of freedom.

I don't think that's the case. Drotmg still computes a Givens rotation, just in a way that reduces the number of arithmetical operations. It still has only only degree of freedom, the angle of rotation, and one constraint, to zero out the second component.

blas/testblas/level1double.go
kortschak: Yes, though not that it makes any difference here.

Fixed.

blas/testblas/level1double.go
kortschak: Yes, that's what I had previously, but that is at odds with the documentation. If I had some basis for removing the sqrt other than "it works better that way", I'd be pretty happy to make the change. They we can get down to fixing the single failing case.

vladimir-ch:
Another reference that looks very useful and may shed some light is https://kar.kent.ac.uk/14202/1/Raermdaarnk.pdf

vladimir-ch:
The most relevant reference is this:
Lawson, C. L., Hanson, R. J., Kinaid, D. R., and Krogh, F. T. 1979b. Basic linear algebra subprograms for Fortran usage. ACM Trans. Math. Softw. 5, 3 (Sept.), 308-323
https://doi.org/10.1145/355841.355847

blas/testblas/level1double.go
vladimir-ch: The BLAS reference paper clarifies this. The input is a scaling matrix D defined by d1, d2, and a vector [x1,y1], and H rotates [x1,y1] (to zero out y1).

kortschak: Any idea where the sqrts that are ubiquitous in the various drotmg documentation pages come from?

vladimir-ch: The inputs to Drotmg define a vector `[a,b]^T` as
```
[a] = [ sqrt(d1)     0 ] * [ x1 ]
[b]   [ 0     sqrt(d2) ]   [ y1 ]
```
Drotmg rotates `[ x1, y1 ]^T` and updates the scaling. It could work directly with the square roots but only squares appear in the formulas, so it's more economical to define d1,d2 as squares of the actual coefficients. Taking the square root is necessary only when you want to form the vector `[a,b]^T` which is not the case in our tests.

Also, regarding the value of `gam`, in the Hanson-Hopkins reference you can read that the value of 4096 is very conservative and can be chosen much larger. This would reduce the occurence of rescaling which is good because the main saving in arithmetic operations comes from Drotm that uses the fact that H has normally two units (and rescaling spoils that). Except that I've just noticed that our Drotm does not! Feel free to create an issue, otherwise I'll do it tomorrow.

This `gam` value story also means that our `blas.Rescaling` test cases rely completely on the hardcoded value. If `gam` is as large as suggested, rescaling would maybe not take place and the test case would fail even though the result would be correct. We could modify `gam` which would support the reason of existence of Drotm(g) (performance) but then we would probably deviate from all other (public) blas implementations out there.

I haven't studied yet why the one test case fails, I may (or may not) have time for it tomorrow.

vladimir-ch:
One final reference, more code oriented with an alternative implementation:
Tim Hopkins. 1997. Restructuring the BLAS level 1 routine for computing the modified givens transformation. SIGNUM Newsl. 32, 4 (October 1997), 2-14. DOI=http://dx.doi.org/10.1145/289251.289253
https://kar.kent.ac.uk/21699/

blas/testblas/level1double.go
```diff
@@ -1902,7 +1902,7 @@ var DrotmgTests = []DrotmgTestStruct{
 		Name: "D1_Small_D2_Big_Flag_1",
```
vladimir-ch: This seems to be named in opposite way.

blas/testblas/level1double.go
```diff
@@ -1830,7 +1830,7 @@ var DrotmgTests = []DrotmgTestStruct{
 		Name: "D1_Big_D2_Small_Flag_1",
```
vladimir-ch: This seems to be named in opposite way.

blas/testblas/level1double.go
```diff
@@ -1772,7 +1772,7 @@ var DrotmgTests = []DrotmgTestStruct{
 		Name: "RD1_Big_RD2_Big_Flag_0",
```
vladimir-ch: How are the test cases named? Flag_0 but Rescaling is -1 ...
Also in other test cases.

blas/testblas/level1double.go
kortschak: @btracey ?

kortschak:
@vladimir-ch Do you have access to Lawson et al and Hopkins?

blas/testblas/level1double.go
btracey: I'm trying to remember what I meant when I wrote that. I remember that I was working on this at my parents' house and trying to find cases to get 100% test coverage... My suspicion looking at it now is that Flag is a misnomer, and I suspect it is really referring to the `if absQ1 > absQ2` statement.

btracey: I just read Dan and my emails (from 2013!) on Drotmg and while there was complaining about the documentation, there is no reference to what I meant by that `Flag` value. I still stand by my suspicion, and it seems to match when I look at a few of the values.

blas/testblas/level1double.go
vladimir-ch: Regarding the failure, I think this comment https://github.com/xianyi/OpenBLAS/issues/1452#issuecomment-368057808 is right and the logic is most likely wrong.

It's easy to see it when you comment out all other tests and only log the error so that the test passes. First, H is set to zero out y1. Then rescaling in the first component is done, so flag is Rescaling. Then rescaling for the second component is done, but since flag is Rescaling, h12 is set to 1 and h21 is set to -1. This destroys the scaling in x1 and zeroing out of y1. It makes no sense.

If I had time, I would consider rewriting this based on the Appendix A in the Hopkins paper.

kortschak:
Here is the Hopkins implementation. There are too many failures for me to want to wade through it - some presumably are just correct difference in the result, others probably not.

```
// Drotmg computes the modified Givens rotation. See
// http://www.netlib.org/lapack/explore-html/df/deb/drotmg_8f.html
// for more details.
func (Implementation) Drotmg(d1, d2, x1, y1 float64) (p blas.DrotmParams, rd1, rd2, rx1 float64) {
	// The implementation of Drotmg used here is taken from Hopkins 1997
	// Appendix A: https://doi.org/10.1145/289251.289253

	// TODO(kortschak): Consider placing this in an init func
	// and making gamsq, gam and rgamsq package globals.
	//
	// The original Fortran makes a comparison of math.MaxFloat64 and
	// 1/math.SmallestNonzeroFloat64. This is not possible here
	// as the second value overflows float64.
	gamsq := math.MaxFloat64
	gam := math.Sqrt(gamsq)
	rgamsq := 1 / gamsq

	if d1 < 0 {
		p.Flag = blas.Error
		return p, 0, 0, 0
	}

	var h11, h12, h21, h22 float64
	if d2 == 0 || y1 == 0 {
		p.Flag = blas.Identity
	} else if (d1 == 0 || x1 == 0) && d2 > 0 {
		p.Flag = blas.Diagonal
		h12 = 1
		h21 = -1
		x1 = y1
		d1, d2 = d2, d1
	} else {
		p2 := d2 * y1
		p1 := d1 * x1
		q2 := p2 * y1
		q1 := p1 * x1
		if math.Abs(q1) > math.Abs(q2) {
			p.Flag = blas.OffDiagonal
			h11 = 1
			h22 = 1
			h21 = -y1 / x1
			h12 = p2 / p1
			u := 1 - h12*h21
			if u <= 0 {
				// This requires that a blas.Error (2) be added to blas.Flag's values
				p.Flag = blas.Error
				return p, 0, 0, 0
			} else {
				d1 /= u
				d2 /= u
				x1 *= u
			}
		} else {
			if q2 < 0 {
				p.Flag = blas.Error
				return p, 0, 0, 0
			} else {
				p.Flag = blas.Diagonal
				h21 = -1
				h12 = 1
				h11 = p1 / p2
				h22 = x1 / y1
				u := 1 + h11*h22
				d1, d2 = d2/u, d1/u
				x1 = y1 * u
			}
		}
	}

	if p.Flag != blas.Error && p.Flag != blas.Identity {
		for d1 <= rgamsq && d1 != 0 {
			p.Flag = blas.Rescaling
			d1 = (d1 * gam) * gam
			x1 /= gam
			h11 /= gam
			h12 /= gam
		}
		for d1 > gamsq {
			p.Flag = blas.Rescaling
			d1 = (d1 / gam) / gam
			x1 *= gam
			h11 *= gam
			h12 *= gam
		}

		for math.Abs(d2) <= rgamsq && d2 != 0 {
			p.Flag = blas.Rescaling
			d2 = (d2 * gam) * gam
			h21 /= gam
			h22 /= gam
		}
		for math.Abs(d2) > gamsq {
			p.Flag = blas.Rescaling
			d2 = (d2 / gam) / gam
			h21 *= gam
			h22 *= gam
		}
	}

	switch p.Flag {
	case blas.Diagonal:
		p.H = [4]float64{h11, -1, 1, h22}
	case blas.OffDiagonal:
		p.H = [4]float64{1, h21, h12, 1}
	case blas.Rescaling:
		p.H = [4]float64{h11, h21, h12, h22}
	case blas.Identity:
		// This is [1, 0, 1, 0] in the paper
		// which is obviously incorrect.
		p.H = [4]float64{1, 0, 0, 1}
	case blas.Error:
		// Do nothing; p.H is already zero.
	default:
		panic("blas: unexpected blas.Flag")
	}

	return p, d1, d2, x1
}
```

vladimir-ch:
I will go through it today. Hopefully most (all?) failures will be due to a
different value of gam. If not, it should be easy to spot the differences
or at least it could help us restructure the existing Drotmg.

vladimir-ch:
If I do the following changes:
- replace blas.Error with blas.Rescaling just to make the tests pass, this could be handles later.
- set `gam` and related values to the original ones
- remove the (incorrect) scaling by sqrt in the test and instead scale v by sqrt(rd2) as in
```
if rd2 >= 0 {
    v *= math.Sqrt(rd2)
}
```
then I get the following test failures:
```
	level1double.go:2090: drotmg H mismatch ZeroD1: expected [0 0 0 2], found [0 0 0 0]
	level1double.go:2090: drotmg H mismatch RD1_Big_RD2_Big_Flag_0: expected [4096 -4096 1 4096], found [4096 -3584 1792 4096]
	level1double.go:2090: drotmg H mismatch RD1_Big_RD2_Big_Flag_1: expected [2340.5714285714284 -4096 1 4681.142857142857], found [2340.5714285714284 -4096 4096 4681.142857142857]
	level1double.go:2090: drotmg H mismatch D1_Big_D2_Small_Flag_1: expected [2.8671999999999997e-26 -0.000244140625 1 2.44140625e-16], found [2.8671999999999997e-26 -0.000244140625 4096 2.44140625e-16]
	level1double.go:2090: drotmg H mismatch D1_Small_D2_Big_Flag_1: expected [2.3731773997569866e+10 -1.6777216e+07 1 1.6777216e-07], found [2.3731773997569866e+10 -1.6777216e+07 0.000244140625 1.6777216e-07]
	level1double.go:2090: drotmg H mismatch OpenBLAS#1452: expected [1.6110934624105326e-06 -0.000244140625 1 1.6276041666666668e-06], found [1.6110934624105326e-06 -0.000244140625 0.000244140625 1.6276041666666668e-06]
```

The first failure is obviously a bad expectation:
```
	level1double.go:2090: drotmg H mismatch ZeroD1: expected [0 0 0 2], found [0 0 0 0]
```
Here the test case is:
```
{
		Name: "ZeroD1",
		P: &blas.DrotmParams{
			Flag: blas.Diagonal,
			H:    [4]float64{0, 0, 0, 2},
		},
		D1:  0,
		D2:  2,
		X1:  8,
		Y1:  4,
		Rd1: 2,
		Rd2: 0,
		Rx1: 4,
	},
```
The expectation on H makes no sense. d1 zeros out the first component, so flag should be Diagonal and H should be just a 90 degree rotation, ie, all zeros, then d1 goes into rd2 (and so the rotated scaled y is zero), and y1 into rx1. Also Drotm ignores H[3] for flag Diagonal.

Then
```
	level1double.go:2090: drotmg H mismatch RD1_Big_RD2_Big_Flag_0: expected [4096 -4096 1 4096], found [4096 -3584 1792 4096]
	level1double.go:2090: drotmg H mismatch RD1_Big_RD2_Big_Flag_1: expected [2340.5714285714284 -4096 1 4681.142857142857], found [2340.5714285714284 -4096 4096 4681.142857142857]

	level1double.go:2090: drotmg H mismatch D1_Small_D2_Big_Flag_1: expected [2.3731773997569866e+10 -1.6777216e+07 1 1.6777216e-07], found [2.3731773997569866e+10 -1.6777216e+07 0.000244140625 1.6777216e-07]

```
would pass if it wasn't for the changes to the expected values made in this PR.

The other two failing test cases show the same pattern: the expected H[2] is 1, the gotten H[2] is different, flag is Rescaling in all cases and also looking at the scales d1,d2 I think rescaling took place in both components triggering the nonsensical behavior I described yesterday. So the question is where the expected H values come from but seeing the situation I'm skeptical these are correct.

kortschak:
Thanks for that. I trust the programmatic tests much more than I trust the golden values - this passes with the changes you describe. Though the weirdness with the sqrt irritates me. Can you explain that?

PTAL

vladimir-ch:
Let `w := [w1,w2]^T` be a vector in R^2, and `G` a Givens rotation that zeros out w2. We can decompose `w` as `w = S v` where `S` is a scaling matrix (this can always be done by taking `S = I`) and `v := [x1,y1]^T`.

So to zero out `w2`, we do `G w = G S v`. It can be shown that `G S` can be rewritten as `D H` where `D` is another scaling matrix and H has two unit entries. As is written in the BLAS paper, only the squares of the scale factors are involved in the nonunit elements of H and so square roots can be avoided by updating the squares of the scale factors, i.e., by updating D^2. Knowing this, if we go back and start with D^{1/2} (in `D^{1/2} H v`), then we end up updating D (which is what Drotmg does). But this also means that after applying `H v`, if you want to recover the rotated `w` itself, you have to multiply by `D^{1/2}` and that's where the sqrt comes from.

btracey:
I personally would wait for some discussion with OpenBLAS on the value of `gam`. I'm not sure the gains gotten from changing `gam` outweigh the costs of returning different values than all other representations. All these bugs suggest to me at least that `drotmg` is not particularly widely used, and so it's especially unclear to me if it's worth the headache.

PR #426, opened 2018-02-27
blas/gonum: use p.Flag in rotm more carefully

### LINKED ISSUE #425: blas/gonum: Drotm doesn't use special structure of H

Call gonum.Drotm

Drotm uses a fact that the input matrix H most commonly has a special structure (two unit elements) that allows it to reduce the number of multiplications.

### What actually happened?

Drotm forms the matrix H explicitly and then uses a generic loop over the two input vectors, thus defeating the purpose of Drotm/Drotmg pair. Instead it should special-case based on the value of Flag.

Yes.

kortschak:
Please take a look.

Fixes #425.

### DIFF
--- a/blas/gonum/level1double.go
+++ b/blas/gonum/level1double.go
@@ -511,47 +511,88 @@ func (Implementation) Drotm(n int, x []float64, incX int, y []float64, incY int,
 		panic(badY)
 	}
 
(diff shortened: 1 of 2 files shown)


kortschak:
There's a fair amount of duplication here, but since it is an optimsation routine, this seems to be the intention.

PR #433, opened 2018-03-05
travis: shard tags over separate builds

kortschak:
Travis seems to have been throwing weaker hardware at problems recently, so we have seen many test builds timing out.

This change shards the four different tags we use in testing, and additionally also adds tagged builds to the test coverage. This makes it possible for the tests to complete and has the additional advantage that tagged tests status is easier to see, and tagged coverage is now available.

Please take a look.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,12 @@ go:
  - 1.10.x
  - master
 
+env:
+ - TAGS=""
+ - TAGS="-tags bounds"
+ - TAGS="-tags noasm"
+ - TAGS="-tags appengine"
+
 matrix:
  fast_finish: true
  allow_failures:
@@ -31,10 +37,7 @@ script:
  - ${TRAVIS_BUILD_DIR}/.travis/check-formatting.sh
  - go get -d -t -v ./...
  - go build -v ./...
- - go test -v ./...
- - go test -a -tags bounds -v ./...
- - go test -a -tags noasm -v ./...
- - go test -a -tags appengine -v ./...
+ - go test -a $TAGS -v ./...
  - if [[ $TRAVIS_SECURE_ENV_VARS = "true" ]]; then bash ./.travis/test-coverage.sh; fi
  - ${TRAVIS_BUILD_DIR}/.travis/check-rand.sh
  # This is run last since it alters the tree.
--- a/.travis/test-coverage.sh
+++ b/.travis/test-coverage.sh
@@ -13,7 +13,7 @@ testCover() {
 	# switch to the directory to check
 	pushd $d > /dev/null
 	# create the coverage profile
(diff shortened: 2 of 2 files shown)


btracey:
Thanks! That's great

kortschak:
@sbinet Where do those numbers come from?

sbinet:
Randomly selecting a previous PR and computing with that one.
It's true the sigma of the distribution of this variable is quite large, though...

sbinet:
**comparing with that one

kortschak:
The overall time for the new approach is longer though (Total time 6 hrs 45 min 25 sec), and even the wall time is not that much better (Ran for 47 min 50 sec) since there is a cap of 50 min. I'm wondering about the precise details for the calculation you used.

kortschak:
Note that I'm not worried about the total run time, since we get better information from having run test coverage analysis for all tags (4x the previous setup).

kortschak:
Also note that travis may lie. [This page](https://travis-ci.org/gonum/gonum/builds/349213635?utm_source=github_status&utm_medium=notification) shows a "Rand for 21 min 55 sec", but the longest build job ran for 36 min 8 sec. I suspect this is a result of failing to correctly handle timing when a job is rerun to fix a transient error (it should just add the additional time on to the existing time, but 21 min 55 sec is a pretty good match for the g1.10.x TAGS="-tags bounds" job which I had to restart and ran for 21 min 58 sec.

sbinet:
No calculation involved, really.
Just random MonteCarlo/Brownian walk :)

I looked at this PR:
- https://travis-ci.org/gonum/gonum/builds/349097714 (`Ran for 13 min 37 sec`)

and compared to that one:
- https://travis-ci.org/gonum/gonum/builds/347968277 (`Ran for 1 hr 26 min 7 sec`)

looking at others, I did notice (as I was hinting at in https://github.com/gonum/gonum/pull/433#issuecomment-370568568) that the spread of this distribution would be quite large, *if* somebody cared to plot it.

as you note, the measurements displayed by travis-ci may not be the most accurate, though...

in conclusion: I was expecting a reduction in wall time and stopped my non-scientific survey as soon as I got data agreeing with my expectations.

kortschak:
Yeah, I wouldn't trust the "Ran for ..." comments at all. Clearly the the one that ran for "13" minutes is a lie since one of the jobs was running for 42 minutes.

If travis get their act together with sources capable metal (this is probably why the variance is so huge), then we'll see possibly reductions in time, but the overall time spent (CPU hours) will go up because we are asking for more work to be done (and more of the most expensive work since coverage is much more costly than plain testing. Looking now, we should probably add a TAGS="-tags race" and not run coverage on that.

PR #434, opened 2018-03-06
travis: add race test to matrix

kortschak:
Please take a look.


kortschak:
After correcting the errors in my original push, and fiddling around with various things to try to get this to work, it's pretty clear to me that this is not feasible; the race built tests take too long to complete within the allowable travis timeouts.

Issue #184, opened 2017-08-08
blas/gonum: add complex BLAS Level 2 functions


vladimir-ch:
- [x] ZGEMV - matrix vector multiply
- [x] ZGBMV - banded matrix vector multiply
- [x] ZHEMV - hermitian matrix vector multiply
- [x] ZHPMV - hermitian packed matrix vector multiply
- [x] ZHBMV - hermitian banded matrix vector multiply
- [x] ZTRMV - triangular matrix vector multiply
- [x] ZTPMV - triangular packed matrix vector multiply
- [x] ZTBMV - triangular banded matrix vector multiply
- [x] ZTRSV - solving triangular matrix problems
- [x] ZTPSV - solving triangular packed matrix problems
- [ ] ZTBSV - solving triangular banded matrix problems
- [x] ZGERU - performs the rank 1 operation `A := alpha*x*y' + A`
- [x] ZGERC - performs the rank 1 operation `A := alpha*x*conjg( y' ) + A`
- [x] ZHER - hermitian rank 1 operation `A := alpha*x*conjg(x') + A`
- [x] ZHPR - hermitian packed rank 1 operation `A := alpha*x*conjg( x' ) + A`
- [x] ZHER2 - hermitian rank 2 operation
- [x] ZHPR2 - hermitian packed rank 2 operation

btracey:
Almost there!

PR #435, opened 2018-03-07
blas/gonum: add Zgbmv

vladimir-ch:
Please take a look.

Updates #184


vladimir-ch:
I'll take the approval as a green light to fix #436. I put it in my task queue.

PR #438, opened 2018-03-07
optimize: add ListSearch for finding the optimum over a specific list…

btracey:
… of values

### DIFF
--- a/optimize/listsearch.go
+++ b/optimize/listsearch.go
@@ -11,11 +11,11 @@ import (
 )
 
 // ListSearch finds the optimum location from a specified list of possible
-// optimum locations. The dimension specified in init must match the number
-// of columns of Locs, and Locs must have at least one row.
+// optimum locations.
 type ListSearch struct {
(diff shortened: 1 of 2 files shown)


btracey:
PTAL @vladimir-ch . This is the method I was implementing when I discovered the hidden bugs, so it's good to have.

This is useful for doing something like a grid search (though we may want to also have a particular grid search method). I've used it in high-D spaces to get a good initial guess for a local optimization

optimize/listsearch.go
```diff
@@ -0,0 +1,119 @@
...
+package optimize
+
+import (
+	"math"
+
+	"gonum.org/v1/gonum/mat"
+)
+
+// ListSearch finds the optimum location from a specified list of possible
+// optimum locations. The dimension specified in init must match the number
```
vladimir-ch: s/init/InitGlobal/
Although I'm not sure whether these comments wouldn't better go to InitGlobal and Locs.

PR #440, opened 2018-03-13
stat/distuv: Add Gumbel distribution

btracey:

### DIFF
--- a/stat/distuv/general.go
+++ b/stat/distuv/general.go
@@ -4,8 +4,6 @@
 
 package distuv
 
-import "math"
-
 // Parameter represents a parameter of a probability distribution
 type Parameter struct {
 	Name  string
@@ -19,8 +17,8 @@ var (
 	badNoSamples  = "distuv: must have at least one sample"
 )
 
-var (
-	expNegOneHalf   = math.Exp(-0.5)
-	eulerMascheroni = 0.5772156649015328606065120900824024310421
-	apery           = 1.2020569031595942853997381615114499907649
+const (
+	expNegOneHalf   = 0.6065306597126334236037995349911804534419 // https://oeis.org/A092605
+	eulerMascheroni = 0.5772156649015328606065120900824024310421 // https://oeis.org/A001620
+	apery           = 1.2020569031595942853997381615114499907649 // https://oeis.org/A002117
 )
--- a/stat/distuv/general.go
+++ b/stat/distuv/general.go
@@ -20,5 +20,7 @@ var (
 )
 
 var (
-	expNegOneHalf = math.Exp(-0.5)
+	expNegOneHalf   = math.Exp(-0.5)
+	eulerMascheroni = 0.5772156649015328606065120900824024310421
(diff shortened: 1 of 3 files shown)


btracey:
Currently the pdf is only tested with `mu=0, beta =1` because that's what scipy has, and my R installation is having issues. If anyone wants to provide other numbers I'll gladly add them.

stat/distuv/general.go
```diff
@@ -20,5 +20,7 @@ var (
 )
 
 var (
-	expNegOneHalf = math.Exp(-0.5)
+	expNegOneHalf   = math.Exp(-0.5)
+	eulerMascheroni = 0.5772156649015328606065120900824024310421
```
kortschak: These can be consts. Add `// https://oeis.org/A001620`?

stat/distuv/gumbel.go
```diff
@@ -0,0 +1,116 @@
...
+
+// Gumbel implements the Gumbel distribution, a two-parameter continuous
+// distribution with support over the real numbers.
+//
+// The Gumbel distribution has density function
+//  1/beta * exp(-(z + exp(-z)))
+//  z = (x - mu)/beta
+// Beta must be greater than 0.
+//
+// For more information, see https://en.wikipedia.org/wiki/Gumbel_distribution .
```
kortschak: I think you can omit the space before the period, but please check.

stat/distuv/gumbel.go
```diff
@@ -0,0 +1,116 @@
...
+}
+
+// Prob computes the value of the probability density function at x.
+func (g Gumbel) Prob(x float64) float64 {
+	return math.Exp(g.LogProb(x))
+}
+
+// Quantile returns the inverse of the cumulative probability distribution.
+func (g Gumbel) Quantile(p float64) float64 {
+	if p < 0 || p > 1 {
```
kortschak: `if p < 0 || 1 < p {`

stat/distuv/general.go
```diff
@@ -20,5 +20,7 @@ var (
...
+	apery           = 1.2020569031595942853997381615114499907649
```
kortschak: `// https://oeis.org/A002117`

btracey:
I think we should rename this to GumbelLeft. The wikipedia definition is the one we have here, but Wolfram's GumbelDistribution is flipped, and scipy has gumbel_l and gumbel_r .

btracey:
or rather, GumbelRight, as this is right-skewed

stat/distuv/gumbel.go
btracey: Done.

stat/distuv/general.go
btracey: Done.

stat/distuv/gumbel.go
btracey: Done.

stat/distuv/general.go
btracey: Done.

PR #443, opened 2018-03-15
optimize/convex/lp: fix Simplex when number of constraints is number …

btracey:
…of variables

Fixes 441.

### DIFF
--- a/optimize/convex/lp/simplex.go
+++ b/optimize/convex/lp/simplex.go
@@ -103,11 +103,16 @@ func simplex(initialBasic []int, c []float64, A mat.Matrix, b []float64, tol flo
 	if m == n {
 		// Problem is exactly constrained, perform a linear solve.
 		bVec := mat.NewVecDense(len(b), b)
-		x := make([]float64, m)
+		x := make([]float64, n)
 		xVec := mat.NewVecDense(n, x)
 		err := xVec.SolveVec(A, bVec)
 		if err != nil {
 			return math.NaN(), nil, nil, ErrSingular
+		}
+		for _, v := range x {
+			if v < 0 {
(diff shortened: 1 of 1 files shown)


btracey:
PTAL @vladimir-ch @sbinet

optimize/convex/lp/simplex.go
```diff
@@ -100,6 +100,19 @@ func simplex(initialBasic []int, c []float64, A mat.Matrix, b []float64, tol flo
 	}
 	m, n := A.Dims()
 
+	if m == n {
+		// Problem is exactly constrained, perform a linear solve.
+		bVec := mat.NewVecDense(len(b), b)
+		x := make([]float64, m)
```
vladimir-ch: `m == n` here but still it is strange to use `m` and `n` on the next line.

optimize/convex/lp/simplex.go
```diff
@@ -100,6 +100,19 @@ func simplex(initialBasic []int, c []float64, A mat.Matrix, b []float64, tol flo
...
+		xVec := mat.NewVecDense(n, x)
+		err := xVec.SolveVec(A, bVec)
+		if err != nil {
+			return math.NaN(), nil, nil, ErrSingular
+		}
+		f := floats.Dot(x, c)
+		return f, x, nil, nil
```
vladimir-ch: What if not `x >= 0`?

optimize/convex/lp/simplex.go
btracey: Done.

optimize/convex/lp/simplex.go
btracey: Good point. Thanks.

PR #444, opened 2018-03-16
mat: fix Exp function behaviour for large norms

### LINKED ISSUE #291: mat: Exp gives crazy results for some matrices
We have a problem with the Exp function that I suspect is due to overflow. Compare this scipy output:

```
>>> from scipy.linalg import expm
>>> import numpy as np
>>> expm(np.array([[-200, 100, 100, 0], [100, -200, 0, 100], [100, 0, -200, 100], [0, 100, 100, -200]]))
array([[ 0.25,  0.25,  0.25,  0.25],
       [ 0.25,  0.25,  0.25,  0.25],
       [ 0.25,  0.25,  0.25,  0.25],
       [ 0.25,  0.25,  0.25,  0.25]])
```

and the gonum result for the same problem:
```
        a := mat.NewDense(4, 4, []float64{
                -200, 100, 100, 0,
                100, -200, 0, 100,
                100, 0, -200, 100,
                0, 100, 100, -200,
        })
        a.Exp(a)
        fmt.Printf("a = %4.2g", mat.Formatted(a, mat.Prefix("    ")))
```
```
a = ⎡ 3.6e+109  -3.6e+109  -3.6e+109   3.6e+109⎤
    ⎢-3.6e+109   3.6e+109   3.6e+109  -3.6e+109⎥
    ⎢-3.6e+109   3.6e+109   3.6e+109  -3.6e+109⎥
    ⎣ 3.6e+109  -3.6e+109  -3.6e+109   3.6e+109⎦
```

kortschak:
Please take a look.

Fixes #291.


mat/dense_arithmetic.go
```diff
@@ -457,67 +457,162 @@ func strictCopy(m *Dense, a Matrix) {
...
-	const (
-		terms   = 10
-		scaling = 4
-	)
-
-	small := getWorkspace(r, r, false)
-	small.Scale(math.Pow(2, -scaling), a)
-	power := getWorkspace(r, r, false)
-	power.Copy(small)
+	pade := []struct {
```
sbinet: perhaps consider lifting this outside of `Expm` ?

kortschak: I'll run benchmarks to see if the compiler puts these in a read only segment/doesn't reallocate, but I'd rather that stayed within the function scope if possible for clarity.

If the compiler *does* reallocate on each call even though the data are not mutated, I think that's a worthwhile issue against golang/go.

kortschak: I've tried a number of permutations including moving this data outside the function, changing the `[]struct...` to a `[...]struct{...` (this is probably worthwhile anyway) and doing the same with a range over the pointer to the array. None make a major significant difference (maxing out at 2% improvement on bytes allocated but not changing number of allocations and making virtually no difference to time performance - max of ~2% here too, though this is statistically significant).

kortschak: This is the combination of all of them: move the data segments outside the function, make pade a `[...]struct` and range over the pointer to it.

```
name       old time/op    new time/op    delta
Exp10-8      18.6µs ± 5%    18.7µs ± 6%    ~     (p=0.171 n=10+10)
Exp100-8     3.85ms ± 3%    3.80ms ± 1%  -1.34%  (p=0.011 n=10+10)
Exp1000-8     1.99s ± 1%     2.00s ± 1%  +0.64%  (p=0.001 n=10+9)

name       old alloc/op   new alloc/op   delta
Exp10-8      1.41kB ± 2%    1.41kB ± 2%    ~     (p=1.000 n=10+10)
Exp100-8     95.6kB ± 2%    93.2kB ± 1%  -2.44%  (p=0.000 n=9+10)
Exp1000-8    91.6MB ± 0%    91.6MB ± 0%    ~     (p=0.724 n=8+9)

name       old allocs/op  new allocs/op  delta
Exp10-8        18.6 ± 8%      18.6 ± 8%    ~     (p=1.000 n=10+10)
Exp100-8       49.0 ± 0%      49.0 ± 0%    ~     (all equal)
Exp1000-8       140 ± 4%       141 ± 6%    ~     (p=0.992 n=10+10)
```

I think we just leave it as is.

mat/dense_arithmetic.go
```diff
@@ -457,67 +457,162 @@ func strictCopy(m *Dense, a Matrix) {
...
-	)
-	for i := 1.; i < terms; i++ {
-		factI *= i
-
-		// This is OK to do because power and tmp are
-		// new Dense values so all rows are contiguous.
-		// TODO(kortschak) Make this explicit in the NewDense doc comment.
-		for j, v := range power.mat.Data {
-			tmp.mat.Data[j] = v / factI
+		for i := 0; i < r; i++ {
```
btracey: Could you use k here so it's obviously different than the outer loop i?

mat/dense_arithmetic.go
```diff
@@ -457,67 +457,162 @@ func strictCopy(m *Dense, a Matrix) {
...
-		tmp.Mul(w, w)
-		tmp, w = w, tmp
-	}
-	putWorkspace(tmp)
 
-	if w != m {
-		m.Copy(w)
-		putWorkspace(w)
+	// Remaining Padé table line.
+	const phi13 = 5.4
```
btracey: theta

graph/network/diffusion_test.go
```diff
@@ -102,7 +102,7 @@ var diffuseTests = []struct {
 			}
 			return m
 		}(),
-		t: 0.01, // FIXME(kortschak): Low t used due to instability in mat.Exp.
+		t: 1, // FIXME(kortschak): Low t used due to instability in mat.Exp.
```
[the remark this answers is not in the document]
kortschak: Compared to the values above, yes. The scipy algorithm is significantly more complicated though, so I think this is OK.

mat/dense_arithmetic.go
```diff
@@ -457,67 +457,162 @@ func strictCopy(m *Dense, a Matrix) {
 
 // Exp calculates the exponential of the matrix a, e^a, placing the result
 // in the receiver. Exp will panic with matrix.ErrShape if a is not square.
-//
-// Exp uses the scaling and squaring method described in section 3 of
-// http://www.cs.cornell.edu/cv/researchpdf/19ways+.pdf.
 func (m *Dense) Exp(a Matrix) {
+	// The implementation used here is from Functions of Matrices: Theory and Computation
```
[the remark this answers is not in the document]
kortschak: It seems to be one that several people use, but that would probably be worth while.

mat/dense_arithmetic.go
```diff
@@ -457,67 +457,162 @@ func strictCopy(m *Dense, a Matrix) {
...
-
-	small := getWorkspace(r, r, false)
-	small.Scale(math.Pow(2, -scaling), a)
-	power := getWorkspace(r, r, false)
-	power.Copy(small)
+	pade := []struct {
+		phi float64
+		b   []float64
+	}{
+		{phi: 0.015, b: []float64{
```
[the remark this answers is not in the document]
kortschak: My eyes!

mat/dense_arithmetic.go
kortschak: Done.

mat/dense_arithmetic.go
kortschak: Fixed

kortschak:
PTAL

kortschak:
I'll wait for comment after you read the chapter.

graph/network/diffusion_test.go
```diff
@@ -102,7 +102,7 @@ var diffuseTests = []struct {
...
 
 		wantTol: 1e-1, // FIXME(kortschak): High tolerance used due to instability in mat.Exp.
```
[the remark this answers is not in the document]
kortschak: I tried.

mat/dense_arithmetic.go
sbinet: SGTM

kortschak:
Thanks

PR #424, opened 2018-02-27
graph: use int64 for node retrieval

kortschak:
Please take a look.

This is a major breaking change to the graph packages.

This was not done mechanically as I don't think `eg` is able to handle some of the subtlety here. Because of this it needs to be read reasonably carefully in some places, particularly in graph/community where I introduced a bug that was only found because of tests (I would have hated to do this in a dynamically typed language).

Updates #31.

Issue #230, opened 2017-09-06
graph: add NewNodeID(id int64) Node


kortschak:
We need a way for some classes of random graph generation (and other things) to be able to request a specific node ID (we can get away with out it, but it becomes less efficient). Since we removed the `NewNodeID() intT` method, we can now use that name to return a new node with a given ID. There are three obvious possible signatures:

```
// NewNodeID returns a node with the given ID.
// If the node exists in graph it is returned
// otherwise a new Node is returned.
NewNodeID(id int64) Node
```
```
// NewNodeID returns a new node with the given
// ID. If the node exists in the graph n is returned
// nil and ok false.
NewNodeID(id int64) (n Node, ok bool)
```
```
// NewNodeID returns a new node with the given ID if possible.
NewNodeID(id int64) (Node, error)
```

I think I prefer the last one, it is the most general and clearest. The new method would go in `NodeAdder` and be added to the sparse graphs we have in simple.

/cc @mewmew

mewmew:
I prefer the signature

```go
// NewNodeID returns a node with the given ID.
// If the node exists in graph it is returned
// otherwise a new Node is returned.
NewNodeID(id int64) Node
```

mainly because there is currently no way of retrieving a Node from a given node ID. In addition, we may wish to look at refining `Has` to

```go
// Has returns whether the node exists within the graph.
Has(id int64) bool
```

mewmew:
> ... what to do when an ID is not valid for the graph (e.g. the dense matrix-based graphs) ...

For this reason, I think it should be up to the individual graph implementation whether to support a `NewNodeID` method, as such a method would give users control over node ID generation. Graph implementations which support arbitrary node IDs (e.g. maps/adjacency lists) may provide a `NewNodeID` method.

Do you foresee many users of this functionality? Random graph generation and test cases would be the top two I can think of. For test cases at least, it seems reasonable to use a specific graph type, e.g. simple which may provide a `NewNodeID` method; so no need to add it to the `NodeAdder` interface.

In general, I believe we should try to keep the interfaces minimal, intersect of functionality rather than union. This was the same rationale applied when splitting the Graph interface into `graph.Weighted` and `graph.Graph`.

kortschak:
An alternative is to partially go back to where we were with `NewNodeID`. Under this model, `NodeAdder` becomes

```
type NodeAdder interface {
    // NewNode returns a new unique arbitrary ID.
    // The returned ID is only guaranteed to be valid
[...]

PR #448, opened 2018-03-25
graph: use int64 for node, edge and line removal indexing

kortschak:
Please take a look.

Updates #424.

### DIFF
--- a/graph/multi/undirected_test.go
+++ b/graph/multi/undirected_test.go
@@ -24,9 +24,9 @@ func TestMaxID(t *testing.T) {
 		g.AddNode(i)
 		nodes[i] = struct{}{}
 	}
-	g.RemoveNode(Node(0))
+	g.RemoveNode(int64(0))
 	delete(nodes, Node(0))
-	g.RemoveNode(Node(2))
+	g.RemoveNode(int64(2))
 	delete(nodes, Node(2))
 	n := g.NewNode()
 	g.AddNode(n)
@@ -53,7 +53,7 @@ func TestIssue123UndirectedGraph(t *testing.T) {
 	n1 := g.NewNode()
 	g.AddNode(n1)
 
-	g.RemoveNode(n0)
+	g.RemoveNode(n0.ID())
 
 	n2 := g.NewNode()
 	g.AddNode(n2)
--- a/graph/multigraph.go
+++ b/graph/multigraph.go
@@ -125,10 +125,10 @@ type WeightedLineAdder interface {
 
 // LineRemover is an interface for removing lines from a multigraph.
 type LineRemover interface {
-	// RemoveLine removes the given line, leaving the
-	// terminal nodes. If the line does not exist it
-	// is a no-op.
-	RemoveLine(Line)
+	// RemoveLine removes the line with the given end
+	// and line IDs, leaving the terminal nodes. If
+	// the line does not exist it is a no-op.
+	RemoveLine(fid, tid, id int64)
 }
 
 // MultigraphBuilder is a multigraph that can have nodes and lines added.
--- a/graph/graph.go
+++ b/graph/graph.go
@@ -169,10 +169,10 @@ type WeightedEdgeAdder interface {
 
 // EdgeRemover is an interface for removing nodes from a graph.
 type EdgeRemover interface {
-	// RemoveEdge removes the given edge, leaving the
-	// terminal nodes. If the edge does not exist it
-	// is a no-op.
-	RemoveEdge(Edge)
+	// RemoveEdge removes the edge with the given end
+	// IDs, leaving the terminal nodes. If the edge
+	// does not exist it is a no-op.
+	RemoveEdge(fid, tid int64)
 }
 
 // Builder is a graph that can have nodes and edges added.
--- a/graph/graph.go
+++ b/graph/graph.go
@@ -129,10 +129,11 @@ type NodeAdder interface {
 
 // NodeRemover is an interface for removing nodes from a graph.
 type NodeRemover interface {
-	// RemoveNode removes a node from the graph, as
-	// well as any edges attached to it. If the node
-	// is not in the graph it is a no-op.
-	RemoveNode(Node)
+	// RemoveNode removes the node with the given ID
+	// from the graph, as well as any edges attached
+	// to it. If the node is not in the graph it is
+	// a no-op.
+	RemoveNode(id int64)
 }
 
 // EdgeAdder is an interface for adding edges to a graph.
--- a/graph/multi/directed.go
+++ b/graph/multi/directed.go
@@ -121,26 +121,25 @@ func (g *DirectedGraph) SetLine(l graph.Line) {
 	g.lineIDs.Use(lid)
 }
 
-// RemoveLine removes l from the graph, leaving the terminal nodes. If the line does not exist
-// it is a no-op.
-func (g *DirectedGraph) RemoveLine(l graph.Line) {
-	from, to := l.From(), l.To()
-	if _, ok := g.nodes[from.ID()]; !ok {
+// RemoveLine removes the line with the given end point and line IDs from the graph, leaving
+// the terminal nodes. If the line does not exist it is a no-op.
+func (g *DirectedGraph) RemoveLine(fid, tid, id int64) {
+	if _, ok := g.nodes[fid]; !ok {
 		return
 	}
-	if _, ok := g.nodes[to.ID()]; !ok {
+	if _, ok := g.nodes[tid]; !ok {
 		return
 	}
 
-	delete(g.from[from.ID()][to.ID()], l.ID())
-	if len(g.from[from.ID()][to.ID()]) == 0 {
-		delete(g.from[from.ID()], to.ID())
+	delete(g.from[fid][tid], id)
+	if len(g.from[fid][tid]) == 0 {
+		delete(g.from[fid], tid)
 	}
-	delete(g.to[to.ID()][from.ID()], l.ID())
-	if len(g.to[to.ID()][from.ID()]) == 0 {
-		delete(g.to[to.ID()], from.ID())
+	delete(g.to[tid][fid], id)
+	if len(g.to[tid][fid]) == 0 {
+		delete(g.to[tid], fid)
 	}
-	g.lineIDs.Release(l.ID())
+	g.lineIDs.Release(id)
 }
 
 // Node returns the node in the graph with the given ID.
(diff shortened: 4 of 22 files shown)


graph/multigraph.go
```diff
@@ -125,10 +125,10 @@ type WeightedLineAdder interface {
...
 // LineRemover is an interface for removing lines from a multigraph.
 type LineRemover interface {
-	// RemoveLine removes the given line, leaving the
-	// terminal nodes. If the line does not exist it
-	// is a no-op.
-	RemoveLine(Line)
+	// RemoveLine removes the line with the given end
+	// and line IDs, leaving the terminal nodes. If
+	// the line does not exist it is a no-op.
+	RemoveLine(fid, tid, id int64)
```
mewmew: Something about the `RemoveLine` method doesn't feel right to me. It is too many IDs to keep track of, and not entirely clear what each one is referring to.

I know, that `fid` is from ID, `tid` is to ID and `id` is the line ID. but this is clear only after a moments thought. Also, the line `id` is not analogous to the node ID which increases the cognitive load.

Perhaps this would be a good time to discuss whether the line ID should be unique in the entire graph, rather than just for edges with the same end points in a multigraph. What would the pros and cons be?

Pro:
* reduced cognitive load as `line.ID()` would by analogous to `node.ID()`.
* the `RemoveLine` method could be simplified to `RemoveLine(id int64)`

Cons:
* imposes constraints on implementations to keep track of the line IDs for the entire graph.

I would be curious to learn more about various implementations, is the con true for real world code? Would the implementation be much simpler (or efficient) of the ID only needs to be kept distinct for lines sharing the same end point?

On an extended note (one I don't know if it adds any benefit, but definitely worth discussing, and potentially rejecting). If after discussion we reach consensus that the line ID should be unique in the entire graph and not just for lines sharing the same end points, then it would be reasonable to open the question of whether we can to represent edges with an edge ID?

If that would be the case than the entire distinction between edges and lines could be removed.

I understand this opens a larger question than perhaps this PR intended. However, it is better to get the API right now, rather than at some later point in the future, as both this PR and the suggested change to make line IDs unique in the graph cause API breakage.

What are your thoughts on this?

kortschak: >I know, that `fid` is from ID, `tid` is to ID and `id` is the line ID. but this is clear only after a moments thought. Also, the line `id` is not analogous to the node ID which increases the cognitive load.

I disagree here; the line ID is exactly analogous to a node ID, though in a different namespace.

>Perhaps this would be a good time to discuss whether the line ID should be unique in the entire graph, rather than just for edges with the same end points in a multigraph. What would the pros and cons be?

Ah, OK, I see what you mean by not analogous. If edges had an ID then all edge removals would be by edge ID.

>Pro:
>* reduced cognitive load as `line.ID()` would by analogous to `node.ID()`.
>* the `RemoveLine` method could be simplified to `RemoveLine(id int64)`

>Cons:
>* imposes constraints on implementations to keep track of the line IDs for the entire graph.

Note that this is actually how the graphs implement them, though that is not currently required.

**EDIT:** and of course, point out more pros and cons if you think of any.

>I would be curious to learn more about various implementations, is the con true for real world code? Would the implementation be much simpler (or efficient) of the ID only needs to be kept distinct for lines sharing the same end point?

In some cases it would be a lot simpler to not bother with retaining a graph-wide ID list; the most obvious case is when lines are stored in a `[]graph.Line` and the ID is just the slice index - this is not currently prohibited (i.e. we don't say that the line ID must persist between calls).

>On an extended note (one I don't know if it adds any benefit, but definitely worth discussing, and potentially rejecting). If after discussion we reach consensus that the line ID should be unique in the entire graph and not just for lines sharing the same end points, then it would be reasonable to open the question of whether we can also represent edges with an edge ID?

>If that would be the case than the entire distinction between edges and lines could be removed.

This is actually how graph was for a brief period after #335 went in. I prefer the simpler computational, cognitive and SLOC burden on the common simple case, I'm also reasonably happy with the separation between edges and lines and how they are reintegrated as a `[]graph.Line` being able to be a `graph.Edge`.

>I understand this opens a larger question than perhaps this PR intended. However, it is better to get the API right now, rather than at some later point in the future, as both this PR and the suggested change to make line IDs unique in the graph cause API breakage.

Agreed.

Overall, I'm not sure about what to do with this. @vladimir-ch?

DryHumour:
Since the API is still open for modification, might I suggest that distinct integer types (NodeID, EdgeID, LineID etc.) be utilised for the different uses?  In general over the years I've found that to be a useful way to reduce programming errors (i.e. inadvertently using the wrong ID type).

mewmew:
> Since the API is still open for modification, might I suggest that distinct integer types (NodeID, EdgeID, LineID etc.) be utilised for the different uses? In general over the years I've found that to be a useful way to reduce programming errors (i.e. inadvertently using the wrong ID type).

@DryHumour Thanks for the suggestion. Using `NodeID` has been proposed before, https://github.com/gonum/graph/issues/87#issuecomment-299039592 and at that time it was determined not to carry its weight.

Comment by @kortschak for when introducing only `NodeID`, https://github.com/gonum/graph/issues/87#issuecomment-299097832. May be different now, if there exist both a `NodeID` and an `EdgeID`.
> My main question is why it is necessary to have a NodeID type? This limits some things that are currently possible.

Perhaps, now we can revisit the decision if edge IDs are indeed added, as there then will exist more than one kind of ID, and the benefits of having distinct `NodeID` and `EdgeID` types becomes clearer.

Also, for background on edge ID, see https://github.com/gonum/graph/issues/87#issuecomment-299339036
>> @kortschak:
>> The main issue with a new named type for node ID is that at some stage I suspect that edges will get IDs (I think this may be the only sensible way to handle multiple edge mutation). This brings with it the possibility that edges in one graph can be nodes in another. This kind of edge annotation is used elsewhere and allows very rich data modelling.
>
> @mewmew:
> Yes, while playing with the new API for node IDs, it also appeared natural to me that an edge ID would make sense at some point, as that would unify edge retrieval with node retrieval operations.

DryHumour:
One other thing I noticed while porting my algorithms to the new API was that it would have been convenient if `graph.Graph` had `Graph.Node(id int64) graph.Node` (just as the simple concrete implementations already provide).  There were several places where I was carrying around full `graph.Node`s only so I wouldn't "lose" access to them -- otherwise my algorithms could have operated solely with the id, only converting back at the end where necessary.

kortschak:
>Since the API is still open for modification, might I suggest that distinct integer types (NodeID, EdgeID, LineID etc.) be utilised for the different uses? In general over the years I've found that to be a useful way to reduce programming errors (i.e. inadvertently using the wrong ID type).

I have been resistant to this, but I have to say that during the recent change to `int64` node and edge retrieval, the only issue that I had that was a little subtle was in community where this change in approach might have helped.

Given a situation where there is `graph.NodeID` and `graph.LineID` (and possibly `graph.EdgeID`, though I am reluctant to add this since a simple graph is intrinsically an {V_int, (V_int,V_int)} set rather than an {V_int, (V_int,V_int,E_int)} set (which is a multigraph). Adding edge IDs to simple graphs adds either additional layers of API (where accessing/mutating a graph's edges requires getting their IDs first), or multiple ways of doing things (you can access either with or without an edge ID).

kortschak:
>One other thing I noticed while porting my algorithms to the new API was that it would have been convenient if graph.Graph had Graph.Node(id int64) graph.Node (just as the simple concrete implementations already provide). There were several places where I was carrying around full graph.Nodes only so I wouldn't "lose" access to them -- otherwise my algorithms could have operated solely with the id, only converting back at the end where necessary.

Agreed. I just have not got around to adding this. I'm not sure if there is an issue for it. If not, would you add one?

mewmew:
> Agreed. I just have not got around to adding this. I'm not sure if there is an issue for it. If not, would you add one?

While not entirely the same, I believe we should discuss solutions to #230 in tandem with discussions of introducing a `Node(id int64) graph.Node` method, as the two are closely related.

graph/multigraph.go
kortschak: Can we move forward with this?

kortschak:
Addressing the issue of making reference to edges via an edge ID would change one-off look-up from a `map[int64]map[int64]` index operation to a `map[int64]` look-up after the previous type of operation (since we need to get the edge ID from the node IDs and then get the edge for that node). However, this would amortize. From a brief mental introspection, I think the majority of edge lookups, at least in our code, would not offer the amortization benefit, so the majority of edge look-up would have an additional map index operation with the proposed change.

kortschak:
*ping*

kortschak:
Yeah, understood. I just want someone to allow me to move forward with this, since I still have large scale changes to make and having two in the air at the same time is not feasible.

graph/multigraph.go
vladimir-ch: I agree that the signature of `RemoveLine` is not very nice, I also had pause to parse the parameters and think which is which. If lines had graph-unique IDs, then edges should have IDs, then it would not make sense/be necessary to distinguish between lines and edges. On the downside, without `Lines` there would be no clear distinguishing method for `Multigraph`. The question is if `Multigraph` distinction would still be needed. So the consequences are big and it's not clear that merging Edges and Lines and giving them IDs were beneficial, so I'd leave it as is for the moment.

kortschak: The signature here very closely matches the mathematics though (in both the simple and multigraph cases).

graph/multi/undirected_test.go
```diff
@@ -24,9 +24,9 @@ func TestMaxID(t *testing.T) {
 		g.AddNode(i)
 		nodes[i] = struct{}{}
 	}
-	g.RemoveNode(Node(0))
+	g.RemoveNode(int64(0))
```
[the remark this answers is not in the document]
vladimir-ch: I suspected you had a reason.

graph/multigraph.go
vladimir-ch: Yes, let's keep this signature.

graph/multi/undirected_test.go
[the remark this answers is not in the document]
kortschak: Overall, these tests need to be rewritten since they are very ad hoc, but whenever I think of doing it, I can always find something else that I need to do more urgently.

kortschak:
@mewmew I you happy to dismiss your review?

mewmew:
> @mewmew Are you happy to dismiss your review?

Sure, I'll dismiss my review. I think we may revisit this for a v2 of the graph API of gonum, but feel it requires experience with both approaches to really get a feel for which would be better (more consistent, less cognitive load, less boiler plate, etc, etc).

So for now, lets focus on a simple implementation for simple graphs (by not including edge IDs). I still feel the RemoveLine methods are a bit complicated (three IDs rather than one global edge ID), etc. But that's fine.

kortschak:
Thanks

PR #449, opened 2018-03-29
stat: add matrix conditioning functions for MCA and MDS

kortschak:
DO NOT SUBMIT

See https://groups.google.com/d/topic/gonum-dev/gEiEWpCDssU/discussion

### DIFF
--- a/stat/pca_helpers.go
+++ b/stat/pca_helpers.go
@@ -1,0 +1,125 @@
+// Copyright ©2018 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package stat
+
+import (
+	"math"
+	"sort"
+
+	"gonum.org/v1/gonum/blas/blas64"
+	"gonum.org/v1/gonum/mat"
+)
+
+// TransformDisjunctive transforms a Complete Disjunctive Table to a Transformed
+// Complete Disjunctive Table in order to be able to perform Multiple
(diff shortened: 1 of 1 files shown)


btracey:
I think this idea looks good. I don't know if there's a better place for this to be than stat, I'm not sure there's a class of these transformations that works for many algorithms (though I can try and read more about this if you'd like). If this is a more general transformation approach, we could consider putting it into a sub-package. 

I don't like the name "ClassicalScaling" , it's pretty general for a specific idea. I'd vote TorgersonScaling as it's immediately googleable and clearly some kind of specific algorithm. I'd also side to TransformDisjunctive for similar reasons, more clearly legible to someone who isn't already familiar.

kortschak:
Agreed on the naming. I'm somewhat confused about the behaviour though. From my reading the Togerson scaling should just work as an MDS by doing a PCA, but it doesn't seem to here (it disagrees with R's MDS).

kortschak:
I think maybe there should be an MDS package, "gonum.org/v1/gonum/stat/mds", since there are too may types of this.

kortschak:
OK, I think I misunderstood what people were saying about Torgerson Scaling. I have fixed that, but it no longer fits in this PR (though I'll park it here for a while until we decide what to do). I suspect it should go in the above mentioned mds package, and the `TransformDisjunctive` should go here, though I'm still waiting on input from the original requester (I don't see any way to summon him here).

kortschak:
OK. I'm going to abandon this and open two PRs for the MCA in stat and for TCS in stat/mds.

Issue #450, opened 2018-04-04
optimize/convex/lp: Simplex should panic if m > n


jjhbw:
Thank you for this amazing package. Really nice work. I has become my go-to library for when I need to 'rescue' scientific code from R or matlab.
I am currently implementing a simple integer linear programming routine on top of simplex, and came across the following error. 

In short: `lp.Simplex` tells me my input matrix `A` is singular but I don't think it actually is.

### What are you trying to do?
Solve the linear problem specified below.

### What did you do?
```go
func TestSingularityReproduction(t *testing.T) {
	A := mat.NewDense(16, 14, []float64{
		1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
		1, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
		1, 0, 0, 1, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0,
		0, 1, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
		0, 1, 0, 0, 0, 0, -1, 0, 0, 0, 0, 0, 0, 0,
		0, 1, 0, 1, 1, 1, 0, 0, 0, 0, 0, 0, 0, 0,
		0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
		0, 0, 0, 1, 1, 1, 0, 0, 0, 0, 0, 0, 0, 0,
		0, 0, 0, 0, 1, 1, 0, 0, 0, 0, 0, 0, 0, 0,
		1, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0,
		0, 1, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0,
		0, 0, 1, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0,
		0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0,
		0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 1, 0, 0,
		0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 1, 0,
		0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 1,
	})

[...]

PR #451, opened 2018-04-04
optimize/lp: make simplex panic when there are more equality constrai…

btracey:


btracey:
PTAL @vladimir-ch @jjhbw

btracey:
Fixes #450 .

optimize/convex/lp/simplex.go
```diff
@@ -70,10 +70,10 @@ const (
 //
 // The Convert function can be used to transform a general LP into standard form.
 //
-// The input matrix A must have full rank and may not contain any columns with
-// all zeros. Furthermore, len(c) must equal the number of columns of A, and len(b)
-// must equal the number of rows of A. Simplex will panic if these conditions are
-// not met.
+// The input matrix A must have at least as many columns as rows, must have
+// full row rank and may not contain any columns with all zeros. Furthermore,
```
vladimir-ch: > ... must have full row rank ...

If we do panic on this, I think it should be an error instead.

btracey: Agreed. Fixed comment to match the actual behavior.

btracey:
Always :)

PR #456, opened 2018-04-11
Add Go import comments.

functionary:
This pull request was made with an automated tool.

The suggested change adds an import comment to a file in each package
to ensure that the package is always imported using its canonical
import path. With this comment in place, the go tool guarantees
that the package can be imported only from this path.

For more information, see https://golang.org/s/go14customimport,
in particular the "Proposal" section.

### DIFF
--- a/graph/formats/dot/internal/errors/errors.go
+++ b/graph/formats/dot/internal/errors/errors.go
@@ -10,7 +10,7 @@
 // This file is made available under a Creative Commons CC0 1.0
 // Universal Public Domain Dedication.
 
-package errors
+package errors // import "gonum.org/v1/gonum/graph/formats/dot/internal/errors"
 
 import (
 	"bytes"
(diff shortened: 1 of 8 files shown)


mewmew:
@kortschak As far as I know, they should have auto-gen markings all of them. If you find one that doesn't please list which file.

E.g. https://github.com/functionary/gonum/blob/513f351b296a4ddd59556b321e9f492004a89595/graph/formats/dot/internal/errors/errors.go#L1
```go
// Code generated by gocc; DO NOT EDIT.
```

mewmew:
I also believe that adding vanity import paths for internal packages provides little value. The internal packages have no "contract" towards end users, and are merely implementation details. End-users should not be able to import them in either case, so the vanity path does nothing as far as I can tell?

kortschak:
Thanks for pointing [that out](https://github.com/gonum/gonum/pull/456#issuecomment-380408370), I missed it in the diff.

PR #458, opened 2018-04-12
mds: new package for multidimensional scaling

kortschak:
Please take a look.


kortschak:
Tests added. PTAL

stat/mds/mds.go
```diff
@@ -0,0 +1,88 @@
...
+	for i := 0; i < n; i++ {
+		for j := i; j < n; j++ {
+			v := dis.At(i, j)
+			v *= v
+			b.SetSym(i, j, v)
+		}
+	}
+	c := mat.NewSymDense(n, nil)
+	for i := 0; i < n; i++ {
+		s := -1 / float64(n)
```
sbinet: this could probably be hoisted off the loop

kortschak: Thanks. Done

There are probably many other optimisations here as well - I was just aiming at something that works.

stat/mds/mds.go
```diff
@@ -0,0 +1,88 @@
...
+		for j := i; j < n; j++ {
+			b.SetSym(i, j, -0.5*dst.At(i, j))
+		}
+	}
+
+	var ed mat.EigenSym
+	ed.Factorize(b, true)
+	dst.EigenvectorsSym(&ed)
+	vals := ed.Values(nil)
+	sort.Sort(byValues{
```
kortschak: @vladimir-ch This sort is necessary to get the Eigen values to be ordered appropriately. Is there a better way of doing this? Other ED implementations return the values and vectors ordered such that the values are decreasing, but our documentation is silent on the ordering.

vladimir-ch: Yes, I think this would be the way to do it but EigenSym uses Dsyev which promises to return the eigenvalues in ascending order, so this sorting is not necessary. The documentation should be updated.

kortschak: Great. So I can just do a order reverse. I'll make that change soon.

Also, #461.

kortschak: Done

stat/mds/mds.go
```diff
@@ -0,0 +1,83 @@
...
+	vals := ed.Values(nil)
+	reverse(vals, dst.RawMatrix())
+	for i, v := range vals {
+		if v < 0 {
+			vals[i] = 0
+			continue
+		}
+		vals[i] = math.Sqrt(v)
+	}
+	dst.Mul(dst, mat.NewDiagonal(len(vals), vals))
```
vladimir-ch: If m is the number of negative eigenvalues, then `dst` will have last m columns zero. Isn't this information worth returning somehow?

kortschak: We could return the Eigenvalues. Some packages in other languages off this as an option.

vladimir-ch: All of them or only the positive?

kortschak: Don't know, probably all of them, but we could slice to the positive only (this would make the negative values available by slicing to `cap(v)` (which we'd have to document).

vladimir-ch: Not sure how much value the negative eigenvalues have, so slicing to the positive values only seems better, also so that the number of non-zero columns is more readily available. Slicing to `cap(v)` is an interesting idea but isn't it a bit unconventional?

kortschak: >Slicing to `cap(v)` is an interesting idea but isn't it a bit unconventional?

I think it's horrific.

I think adding an input destination slice and returning it is probably the way to go (allocating if it's `nil`). We check that it's adequately long (or that it's exactly correct length for the input - any preference?) and slice to include only the positive elements and return it.

kortschak:
*ping*

stat/mds/mds.go
vladimir-ch: If the eigenvalues are not extremely useful, then just returning the number of non-zero columns would be the easiest. If they are useful, then adding a destination slice sound good to me. Or, having a dedicated type for this?

kortschak: The API here is tricky. I think the simplest thing to do is take a Eigenvalues destination slice and return it and the number of positive values (probably in place of the `ok bool`). Then the user can slice to the correct length if they want (that's essentially what we're getting them to do with the `mds` value too).

vladimir-ch: Just to be clear, by "returning the number of non-zero columns" I mean having a `cols int` return.

The simplest thing would be not take the destinations and just allocate and return, but since we want to reuse, then the destination slice is probably the simplest.

kortschak: Yeah, what I've done is perhaps odd, but please take a look anyway.

stat/mds/mds_test.go
```diff
@@ -0,0 +1,143 @@
...
+			587.675679, 81.18224, -75.88485, 13.080496, 114.688714, -185.462926, -47.4707495, 209.170525, 39.452744, -65.352092, -23.127035,
+			-156.836257, -211.13911, 131.30852, 27.089432, -100.800743, 6.794791, -25.6913210, -7.904762, 39.731234, -79.091817, -75.116800,
+			709.413282, 1109.36665, -179.83052, -109.895049, -90.243055, 316.665218, 9.9070412, 9.976867, 6.316941, -13.285987, 1.496742,
+			839.445911, -1836.79055, -541.35188, -108.755016, 25.155430, 41.399927, 71.8793196, 9.438407, 73.554358, -35.415426, -6.021569,
+			911.230500, 205.93020, 98.02313, 62.375253, 198.960032, -525.197308, -13.9515306, -183.562155, -44.881159, -43.133195, 15.060856,
+		}),
+		wantVals: []float64{
+			1.953837708954e+07, 1.185655533400e+07, 1.528844467987e+06, 1.118741950509e+06,
+			7.893472026801e+05, 5.816552067198e+05, 2.623192077011e+05, 1.925975616762e+05,
+			1.450845349644e+05, 1.079673069262e+05, 5.139484110774e+04, 7.598368987955e-11,
```
kortschak: Note that the R implementation gives k=11 beacuse 7.6e-11 is considered to be zero.

stat/mds/mds_test.go
```diff
@@ -123,7 +123,7 @@ func TestTorgersonScaling(t *testing.T) {
 			t.Errorf("unexpected result for test %d:\ngot:\n%.4f\nwant:\n%.4f",
 				i, mat.Formatted(got.Slice(0, n, 0, wc)), mat.Formatted(test.want))
 		}
-		if !floats.EqualApprox(gotVals, test.wantVals, 1e-12) {
+		if !floats.EqualApprox(gotVals, test.wantVals, 1e-10) {
```
kortschak: This is necessary on travis infrastructure, but not locally on my machine.

PR #459, opened 2018-04-12
stat: add matrix conditioning functions for MCA

kortschak:
Replaces #449.

WIP - needs tests.

### DIFF
--- a/stat/pca_helpers.go
+++ b/stat/pca_helpers.go
@@ -1,0 +1,45 @@
+// Copyright ©2018 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package stat
+
+import (
+	"gonum.org/v1/gonum/mat"
+)
+
+// TransformDisjunctive transforms a Complete Disjunctive Table to a Transformed
+// Complete Disjunctive Table in order to be able to perform Multiple
+// Correspondence Analysis using PC.PrincipalComponents. TransformDisjunctive
+// places the TCDT in dst and returns it.
+//
+// If dst is nil, a new mat.Dense is allocated. If dst is not a zero matrix,
+// the dimensions of dst and cdt must match otherwise TransformDisjunctive will
+// panic. If cdt contains values other 0 or 1 TransformDisjunctive will panic.
+//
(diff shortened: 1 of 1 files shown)


stat/pca_helpers.go
```diff
@@ -0,0 +1,45 @@
...
+)
+
+// TransformDisjunctive transforms a Complete Disjunctive Table to a Transformed
+// Complete Disjunctive Table in order to be able to perform Multiple
+// Correspondence Analysis using PC.PrincipalComponents. TransformDisjunctive
+// places the TCDT in dst and returns it.
+//
+// If dst is nil, a new mat.Dense is allocated. If dst is not a zero matrix,
+// the dimensions of dst and cdt must match otherwise TransformDisjunctive will
+// panic. If cdt contains values other 0 or 1 TransformDisjunctive will panic.
```
vladimir-ch: s/other 0/other than 0/ ?

kortschak: Fixed

btracey:
The code looks good. Is there a link to somewhere to find what this does? Even Googling left me unable to find more information.

kortschak:
I'm waiting for the OP to come and comment. I think I got the approach from the [WP article](https://en.wikipedia.org/wiki/Multiple_correspondence_analysis).

nathanhack:
@btracey and @kortschak sorry for the delayed response.  

The point/benefit of MCA is that it enables analysis of qualitative data sets.  The TransformDisjunctive function in this commit is the second step (in MCA) required to transform a **qualitative** data set so that it can be analyzed by PCA (**quantitative** analysis tool).  

Many attempts to use PCA on qualitative values over the years have shown that without this change the analysis results are skewed which can yield incorrect assumptions/results.

kortschak:
@nathanhack, please understand that this is a low priority for me, so without more significant input this is likely to languish.

kortschak:
I have had a look at the links posted above; the FactoMineR code is not well aligned with the Gonum approach and will require significant investment in time to be able to untangle, and neither link is particularly helpful for developing tests for an implementation.

In the absence of any significant input here, I'm going to abandon this work. I will leave the branch so it can be picked up again if the interest does actually present itself.

PR #460, opened 2018-04-12
encoding/dot: fix ports on undirected graphs

32bitkid:
Please take a look.

### DIFF
--- a/graph/encoding/dot/encode.go
+++ b/graph/encoding/dot/encode.go
@@ -40,6 +40,10 @@ type Attributers interface {
 // connection ports for their end points. The returned port corresponds
 // to the the DOT node port to be used by the edge, compass corresponds
 // to DOT compass point to which the edge will be aimed.
+//
+// When encoded, the From() and To() sides of the graph.Edge will receive the
+// FromPort() and ToPort() annotations respectively, even when the edge is
+// represented in an UndirectedGraph.
 type Porter interface {
 	FromPort() (port, compass string)
 	ToPort() (port, compass string)
--- a/graph/encoding/dot/encode.go
+++ b/graph/encoding/dot/encode.go
@@ -219,9 +219,14 @@ func (p *printer) print(g graph.Graph, name string, needsIndent, isSubgraph bool
 			} else {
 				p.writeNode(n)
 			}
-			e, edgeIsPorter := g.Edge(nid, tid).(Porter)
+			e := g.Edge(nid, tid)
+			porter, edgeIsPorter := e.(Porter)
 			if edgeIsPorter {
-				p.writePorts(e.FromPort())
+				if e.From().ID() == nid {
+					p.writePorts(porter.FromPort())
+				} else {
+					p.writePorts(porter.ToPort())
+				}
 			}
 
 			if isDirected {
@@ -241,7 +246,11 @@ func (p *printer) print(g graph.Graph, name string, needsIndent, isSubgraph bool
 				p.writeNode(t)
 			}
 			if edgeIsPorter {
-				p.writePorts(e.ToPort())
+				if e.From().ID() == nid {
+					p.writePorts(porter.ToPort())
+				} else {
+					p.writePorts(porter.FromPort())
+				}
 			}
 
 			if a, ok := g.Edge(nid, tid).(encoding.Attributer); ok {
--- a/graph/encoding/dot/encode.go
+++ b/graph/encoding/dot/encode.go
@@ -40,12 +40,13 @@ type Attributers interface {
 // connection ports for their end points. The returned port corresponds
 // to the the DOT node port to be used by the edge, compass corresponds
 // to DOT compass point to which the edge will be aimed.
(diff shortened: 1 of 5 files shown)


kortschak:
What's the rationale for using the edge for this? I'm not saying that it's wrong, but I'm not convinced that it's right. At the very least this behaviour should be documented.

32bitkid:
the rationale for using the edge for this is borrowing the argument semantics from the`EdgeAdder` interface https://github.com/gonum/gonum/blob/master/graph/graph.go#L139, and enforcing the same lookup behavior when encoding the graph: edges, even within a undirected graph, have a From side and a To side.

So, even though the graph—_itself_, in the case of a undirected graph—doesn't enforce directionality of an edge, the associated metadata _can_, and _should_ be.

For instance, here is the simplest case that exacerbates the problem:

```go
type Node struct {
	graph.Node
	name string
}

func (n *Node) DOTID() string { return n.name }

type Edge struct {
	graph.Edge
	ports struct{ from, to string }
}

func (e *Edge) FromPort() (port, compass string) { return e.ports.from, "" }
func (e *Edge) ToPort() (port, compass string) { return e.ports.to, "" }

func main() {
	g := simple.NewUndirectedGraph()

	a := &Node{g.NewNode(), "a"}
	g.AddNode(a)

	b := &Node{g.NewNode(), "b"}
	g.AddNode(b)

	e := &Edge{g.NewEdge(b, a), struct{ from, to string }{"p1", "p2"}}
	g.SetEdge(e)

	result, _ := dot.Marshal(g, "", "", "  ", false)
	println(string(result))
}
```

results in
```gv
graph {
  // Node definitions.
  a;
  b;

  // Edge definitions.
  a:p1 -- b:p2;
  // obviously, should be b:p1 -- a:p2 or a:p2 -- b:p2, but not what it is *for sure*
}
```

By not enforcing the from and to sides of the edges, the lookups have been swapped because of the implicit iteration order of the nodes. So, that's not cool. 

But, I think, once you look a little closer at the `Edge` interfaces, there is a _from_ and _to side of a edge: even the implemented interface for adding an edge to an undirected graph is `NewEdge(from, to Node) Edge`,  the `Edge` itself exposes `To()` and `From()`. Which I think does not _ necessary imply_ directionality, but does provide enough of a context for the `Porter` `FromPort()` and `ToPort()` methods to be interpreted predictably and correctly.

This is similar and related to another PR that I'm working on that deal with the encoding of `Multigraphs` which, which have similar problems for UndirectedMultigraphs, but thats a PR for another day.

**tl;dr** Because the interface for edges enforce `From()` and `To()`, and edges, even within undirected graphs, are built _from_ and _to_ nodes. Using that relationship to predictiably derive edge metadata like ports and compass directionality, seem like a good idea. 

But thats just my thoughts, I'd love to get feedback if you have something different in mind. but yea, I got bit by this the other day, and eventually stumbled onto the TODO in the tests before realizing what was going on.

kortschak:
Yes, that all sounds reasonable (I really appreciate the effort you put into the comment above - it really helps when we come back to things like this). Can you add a note in the documentation though.

kortschak:
In `Porter` seems fair.

graph/encoding/dot/encode.go
```diff
@@ -40,6 +40,10 @@ type Attributers interface {
 // connection ports for their end points. The returned port corresponds
 // to the the DOT node port to be used by the edge, compass corresponds
 // to DOT compass point to which the edge will be aimed.
+//
+// When encoded, the From() and To() sides of the graph.Edge will receive the
```
kortschak: I'd place this as separate comments on each of `FromPort` and `ToPort` rather than here.

```
type Porter interface {
    // FromPort returns the port and compass for the
    // From node of a graph.Edge.
    FromPort() (port, compass string)

    // ToPort returns the port and compass for the
    // To node of a graph.Edge.
    ToPort() (port, compass string)
}
```

kortschak:
Thanks

graph/encoding/dot/example_test.go
```diff
@@ -0,0 +1,44 @@
+package dot_test
```
kortschak: Sorry, I missed this. Can you please add a copyright header.

graph/encoding/dot/example_test.go
```diff
@@ -0,0 +1,44 @@
...
+
+func (e *edgeWithPorts) ToPort() (string, string) {
+	return e.toPort, ""
+}
+
+func ExamplePorter() {
+	g := simple.NewUndirectedGraph()
+	g.SetEdge(&edgeWithPorts{
+		Edge: simple.Edge{ simple.Node(1), simple.Node(0) },
+		fromPort: "p1",
```
kortschak: This looks like it has not be gofmt'd. Can you do that please.

graph/encoding/dot/example_test.go
32bitkid: resolved

PR #464, opened 2018-04-14
encoding/dot: support edge ports on Unmarshal/decode 

### LINKED ISSUE #462: graph/encoding: [proposal] port support in the dot decoder
While the dot ast correctly handles/parses dot ports _and_ there is the `Porter` interface for _encoding_ ports to a dot graph, there is no interface/hooks to decode that data from `Unmarshal()`

This is a proposal to add a `PortSetter` interface to expose callbacks for setting the parsed `port:compass` information:

```go
type PortSetter interface {
	SetFromPort(port, compass string)
	SetToPort(port, compass string)
}
```

If an `graph.Edge` implements this interface, it will be invoked during decode after the edge has been added to the graph, similar to how the `DOTIDSetter` interface is handled. If the edge has no ports, then the callbacks will not be called.

---

There is [a proposed branch](https://github.com/gonum/gonum/compare/master...32bitkid:decode-edge-ports?expand=1) with these changes ready for further review and PR, if advisable. 

Other potentially controversial changes in that branch include altering the default `CompassPoint` in `ast/ast.go` to be the token `CompassPointNone`, rather than `CompassPointDefault`. This allows for cleaner replication of when the default port is used explicitly, _i.e_ `foo:p:_ -> bar` can be detected distinctly from `foo:p -> bar` and cleans up some other tests that would strip the default `_` direction off. This would be done in a real-world graph to clear a top-level `edge` declaration.

32bitkid:
This PR introduces a new interface for emitting DOT `:port:compass` information on `Unmarshal()` (resolves #462)

Internal changes:

The parser now returns the `ast.CompassPointNone` if a compass direction is undefined in the source graph. This allows for disambiguation between a non-defined compass direction and the explicit default override `_`.

Potentially interesting/related changes:

This change apllies ports through an edge chain, however, I don't think the general attributes get applied in the same way. For instance, the follow graphs are equivalent:

```
digraph {
  a:p1 -> b:p2 -> c:p3
}

digraph {
  a:p1 -> b:p2 
  b:p2 -> c:p3
}
```
However, the same is not true for edge attributes, which I think the reference DOT implementation does:

```
digraph {
  a -> b -> c [color=red]
}

digraph {
  a -> b [color=red]
  b -> c [color=red]
}
```

However, I think attribute application across a chained-edge should probably be addressed in a different PR. But I'm open to rethink that/discuss it.


kortschak:
>I think attribute application across a chained-edge should probably be addressed in a different PR.

SGTM

graph/encoding/dot/decode.go
```diff
@@ -26,6 +26,17 @@ type DOTIDSetter interface {
 	SetDOTID(id string)
 }
 
+// PortSetter is implemented by graph.Edge and graph.Line that want to
+// access to the DOT port and compass directions of an edge.
+type PortSetter interface {
+	// SetFromPort will receive the port and compass direction of the
```
kortschak:
```
// SetFromPort sets the From port and
// compass direction of the receiver.
```

graph/encoding/dot/decode.go
```diff
@@ -26,6 +26,17 @@ type DOTIDSetter interface {
...
+	// From side of an edge, if defined.
+	SetFromPort(port, compass string)
+	// SetTo will receive the port and compass direction of the To side
```
kortschak:
```

// SetToPort sets the To port and compass
// direction of the receiver.
```

graph/encoding/dot/decode.go
```diff
@@ -26,6 +26,17 @@ type DOTIDSetter interface {
 	SetDOTID(id string)
 }
 
+// PortSetter is implemented by graph.Edge and graph.Line that can set
+// the DOT port and compass directions of an edge.
+type PortSetter interface {
+	// SetFromPort sets the From port and
+	// compass direction of the receiver.
+	SetFromPort(port, compass string) error
+	// SetToPort sets the To port and compass
```
kortschak: Final nit. Please place a blank line between the first method and the following comment.

kortschak:
Thank you for bearing with me.

Issue #467, opened 2018-04-15
travis: implement a generalised import blacklist checker


kortschak:
See #466.

sbinet:
That could be fun to implement.
Let me see if I can come up with something by tomorrow?

kortschak:
There is an obvious approach that is unfortunately incorrect; there are generated files that get modified by goimports (which is probably what we want to check against). So files containing the codegen warning should be skipped.

sbinet:
why?
I mean, IIUC, `goimports` introduced `github.com/gonum/floats` on your setup because it's a shorter path than `gonum.org/v1/gonum/floats` (and shorter paths win when there are duplicates) and because you still have either the `floats.a` binary installed or the sources of that old package somewhere in your `$GOPATH`(s).

the way I see it, it's a problem on your end, like forgetting to run `gofmt` before submitting some human written code as a PR :)

kortschak:
I'm not sure I understand the "why?"

Generated code does not have to follow the same rules as other code. The generated code we produce does follow `gofmt`, but it does not follow `goimports` import block structure rules. That's the problem.

sbinet:
I thought this issue was about checking that we didn't import, say, `"github.com/gonum/floats"` by mistake (either from autocompletion, (old) muscle memory, or some other way).
I don't see how this has anything to do with `goimports` not following Gonum's preferred block structure.

kortschak:
It's been a long day. I had two thoughts in my mind and erroneously thought people would practice telepathy and fix the conflation.

When I was following up the incorrect import, I noticed that doing to goimports check in .travis/check-formatting.sh is not used, and looked to see why. The idea of a generalised import checker and generalised goimports fmt checker merged in my mind and only one thought came out. This is still useful, but we need the other as well. It could be done in the same code you are working on.

Apologies for the confusion.

PR #469, opened 2018-04-16
gonum,internal/imports: add check against imports of deprecated packages

sbinet:
This CL adds a check against Gonum packages importing deprecated or
blacklisted packages.

Fixes gonum/gonum#467.

Please take a look.


gonum.go
```diff
@@ -0,0 +1,9 @@
+// Copyright ©2018 The Gonum Authors. All rights reserved.
```
kortschak: I really don't want a "gonum.org/v1/gonum" to exist until it has a reason to exist for users.

internal/imports/import.go
```diff
@@ -0,0 +1,95 @@
+// Copyright ©2018 The Gonum Authors. All rights reserved.
```
kortschak: This does pretty much what we'd want, though I'd rather it were just a `go run`-able file in `.travis/`

internal/imports/import.go
```diff
@@ -0,0 +1,95 @@
...
+func checkImports(fset *token.FileSet, src []byte, fname string) error {
+	f, err := parser.ParseFile(fset, fname, src, parser.ImportsOnly)
+	if err != nil {
+		return err
+	}
+
+	imp := Error{File: fname}
+	for _, s := range f.Imports {
+		path := strings.Trim(s.Path.Value, `"`)
+		if strings.HasPrefix(path, "github.com/gonum/") {
```
kortschak: Possibly for later, this could be configurable with a blacklist file since we already check for import of "math/rand".

sbinet:
PTAL

internal/imports/import.go
kortschak: Placing it as a test in a package seems odd to me. To my mind, a test is a test of a package (golang/go's all.bash notwithstanding), so a test in internal/imports seems like a test of that package. What we want here is a test of the suite, which should be outside of the suite (.travis/ is such a location). The code is simple enough that it could be a single binary, but if you want it tested then I guess it needs to be in internal/imports, but the actual testing of the suite's imports can still be in a go runnable program in .travis/.

sbinet: My issue with putting that inside '.travis' is twofold:

- it's only run during CI,
- it's run very late, duplicated over all the various 'TAGS=xyz' variants we test against, essentially wasting a lot of resources.

Whereas if it were to be run as part of regular 'go test', it could be caught much earlier, even *before* sending the PR.

I agree, putting that test inside 'internal/imports' isn't ideal and that's why I initially put it at the 'gonum' top-level :)

kortschak: We can make git hooks to do this on the submitter side. We can add others in too.

On the issue of resource waste, how much work is actually done in terms of time? We do waste a lot already.

kortschak:
I'd still prefer this were an executable rather than a testing package.

sbinet:
PTAL.
if everything's OK, I'll neatly squash everything together.

internal/imports/check-imports.go
```diff
@@ -0,0 +1,37 @@
...
+package main
+
+import (
+	"log"
+	"os"
+
+	"gonum.org/v1/gonum/internal/imports"
+)
+
+var (
```
kortschak: This is a single var, so it doesn't need a paren block (same in the test file).

sbinet: done

sbinet:
do you mean some `gonum.org/v1/tools/...` repo ?
- `gonum.org/v1/tools/cmd/check-imports`
- `gonum.org/v1/tools/imports`

sure, why not.
perhaps bubble this up to the `gonum-dev` list ?

kortschak:
SGTM

kortschak:
@sbinet Do you want to close this, or just update it to make use of the new tool?

sbinet:
I wanted to update it once `gonum.org/v1/tools` was in.
(then I got side tracked...)

sbinet:
ah... shoot.
`gonum.org` doesn't serve `gonum.org/v1/tools` :)

sbinet:
should be addressed by https://github.com/gonum/gonum.github.io/pull/9

sbinet:
should be addressed by https://github.com/gonum/website/pull/16

sbinet:
PTAL.

now that `gonum.org/v1/tools` is properly served, everything works :)

sbinet:
oops, I didn't notice I had left the intermediate CLs.
sorry for messing up the history...:
```
* 8b277688 (HEAD -> master, origin/master, origin/HEAD) travis: use check-imports from gonum.org/v1/tools
* de950b78 PS2
* dd368688 travis,internal/imports: check imports with a command, not a test
* ece6203f all: remove redundant math/rand check from travis
* 4110b3f8 internal/imports: use a regexp instead
* 9bb94577 internal/imports: add check against imports of blacklisted packages
* 5530c0a2 blas/gonum: integrate asm calls to Gemv kernels
* 19002801 remove build flags for unsupported versions

```

btracey:
Thanks!

PR #477, opened 2018-05-01
all: replace publicly facing *rand.Rand with rand.Source

### LINKED ISSUE #228: all: consider generalising the random source parameter type

kortschak:
Please take a look.

Closes #228.

### DIFF
--- a/graph/graphs/gen/holme_kim.go
+++ b/graph/graphs/gen/holme_kim.go
@@ -21,10 +21,10 @@ import (
 // additional edges joining existing nodes with probability proportional to the nodes'
 // degrees. The edges are formed as a triad with probability, p.
 // If src is not nil it is used as the random source, otherwise rand.Float64 and
-// rand.Intn are used.
+// rand.Intn are used for the random number generators.
 //
 // The algorithm is essentially as described in http://arxiv.org/abs/cond-mat/0110452.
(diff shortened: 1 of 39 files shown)


btracey:
Why?

btracey:
@sbinet care to comment?

graph/graphs/gen/holme_kim.go
```diff
@@ -116,11 +117,11 @@ func permute(n []graph.Node, rnd func(int) int) []graph.Node {
 // node having degree m-1. At each iteration of graph addition, one node is added
 // with m additional edges joining existing nodes with probability proportional
 // to the nodes' degrees. If src is not nil it is used as the random source,
-// otherwise rand.Float64 is used.
+// otherwise rand.Float64 is for the random number generator.
```
sbinet: `s/is for/is used for/`

Issue #482, opened 2018-05-08
optimize: Combine Global and Local


btracey:
Following some offline discussions with @vladimir-ch , I think `optimize.Global` and `optimize.Local` should be combined into a single call, `optimize.Minimize`. The two function calls use a largely overlapping code and interface at the moment. The two major differences are that 1) Local is only serial 2) Local takes in an initial location while global does not. The second is easily adjusted  by adding a field to Settings, and saying that all 0s is the initial location by default. The first will add a bit of overhead to calls to local, but I think this cost will be made up for by making the optimize package easier to use (a single function rather than two, a single `Method` interface rather than two, etc.), making there be less code overall, and I suspect there will be significant improvement in the legibility of Local methods that comes along.

btracey:
A couple weeks ago I spent a bit of time with this. I decided that doing this all at once would be a massive change that would be extremely difficult to review. I suggest a couple of steps

- Replace the code in Local with a call to Global, and adding a shell `GlobalMethod`. Add an InitLocation field to `GlobalSettings` to allow this to happen
- Remove all of the Local specific code, much of this will just be moving to Global
- Fix the current Local methods to "natively" use the Global interface 
- Delete `Method`, `Settings`, etc. and rename `GlobalMethod` to `Method` etc.

vladimir-ch:
SGTM

btracey:
I'm going to update the above list as PRs are submitted and I find things to add to the list

PR #485, opened 2018-05-08
optimize: remove Local implementation and replace with a call to Global

btracey:
This PR starts the process described in #482. It removes the existing Local implementation, replacing with a function that wraps Method to act as a GlobalMethod. This PR also adds a hack to fix an inconsistency with FunctionConverge between Global and Local (and a TODO to make it not a hack in the future). Finally it improves the Global documentation to help with shutdown.

### DIFF
--- a/optimize/local.go
+++ b/optimize/local.go
@@ -4,10 +4,7 @@
 
 package optimize
 
-import (
-	"math"
-	"time"
-)
+import "math"
 
 // Local finds a local minimum of a minimization problem using a sequential
 // algorithm. A maximization problem can be transformed into a minimization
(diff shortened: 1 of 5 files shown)


optimize/local.go
```diff
@@ -59,181 +56,176 @@ import (
...
-		if loc.Gradient != nil {
-			initG := settings.InitialGradient
-			if initG == nil {
+func (l *localGlobal) getStartingLocation(operation chan<- GlobalTask, result <-chan GlobalTask, task GlobalTask) Operation {
+	copy(task.X, l.InitX)
+	if l.Settings.UseInitialData {
+		task.F = l.Settings.InitialValue
+		if task.Gradient != nil {
+			g := l.Settings.InitialGradient
+			if task.Gradient != nil && g == nil {
```
vladimir-ch: task.Gradient is already checked for nil above.

optimize/unconstrained_test.go
```diff
@@ -1182,36 +1182,36 @@ func testLocal(t *testing.T, tests []unconstrainedTest, method Method) {
 
 		result, err := Local(test.p, test.x, settings, method)
 		if err != nil {
-			t.Errorf("error finding minimum (%v) for:\n%v", err, test)
+			t.Errorf("Cas %d: error finding minimum (%v) for:\n%v", cas, err, test)
```
vladimir-ch: Why not spell these out as "Case"?

optimize/local.go
btracey: Done.

optimize/unconstrained_test.go
btracey: Done.

PR #489, opened 2018-05-09
Remove init location from FunctionConverge

btracey:


btracey:
PTAL @vladimir-ch . Sorry, should have done this in the last PR, but I thought it was going to be a bigger change.

PR #490, opened 2018-05-12
unit: move autogen script to internal package

### LINKED ISSUE #17: unit: Fix placement of autogen script
It currently comes across as a sub-package, it should be moved into an internal location.

vladimir-ch:
Please take a look.

Closes #17


vladimir-ch:
Done. PTAL.

PR #493, opened 2018-05-14
stat: handle non-unique results in test for PC

### LINKED ISSUE #21: stat: PCA fails with noasm or appengine build tags
The failure is due to numerical stability and only really shows up in the lowest valued column.

vladimir-ch:
Please take a look.

Closes #21

### DIFF
--- a/stat/pca_test.go
+++ b/stat/pca_test.go
@@ -180,6 +180,18 @@ tests:
 				t.Errorf("%d use %d: unexpected variance result got:%v, want:%v",
 					i, j, vars, test.wantVars)
 			}
+
+			// Check that the set of principal vectors is
+			// orthonormal by comparing V^T*V to the identity matrix.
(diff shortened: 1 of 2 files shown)


stat/pca_test.go
```diff
@@ -163,14 +159,37 @@ tests:
 				t.Errorf("unexpected SVD failure for test %d use %d", i, j)
 				continue tests
 			}
-			if !mat.EqualApprox(vecs, test.wantVecs, test.epsilon) {
+
+			// Find the number of non-zero variances to handle
+			// non-uniqueness in SVD result (issue #21).
+			nnz := len(vars)
+			for k, v := range vars {
+				if math.Abs(v) < 1e-15 {
```
kortschak: Can we change this to 
```
const epsilon 1e-15
if math.Abs(v) < epsilon {
```

vladimir-ch: Done

PR #496, opened 2018-05-16
optimize: Change Settings to allow InitialLocation

btracey:
This modifies Settings to allow specifying an initial location and properties of the function (value, gradient, etc.). This allows to work with local optimizers that are seeded with initial settings. This has two fields that must be specified (both Location and Operation). The difficulty is that the default value of the function is 0, so we either must require the user to specify it is set (somehow), or require the user to change the default value away if it is not set. The former seems much safer.

Please take a look.

### DIFF
--- a/optimize/global.go
+++ b/optimize/global.go
@@ -58,6 +58,12 @@ type GlobalMethod interface {
 	// The last parameter to RunGlobal is a slice of tasks with length equal to
 	// the return from InitGlobal. GlobalTask has an ID field which may be
 	// set and modified by GlobalMethod, and must not be modified by the caller.
+	// The first element of tasks contains information about the initial location
+	// if any is specified in Settings. The Location field will contain the value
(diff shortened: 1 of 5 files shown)


btracey:
It's unfortunate that this version requires users to deal with `Operations`, which otherwise are only needed by the methods. I think the other viable alternative is to have 
```
type Settings {
    InitX []float64
    InitValues *Location
}
```
where InitValues ignores the `X` location (using `InitX`) and must have the initial function value set if it is non-nil. It's not clear which is less confusing. Maybe this current way is worse because setting the initial X will be the most common thing to do (far more common than setting any of the values).

Thoughts @vladimir-ch @kortschak ?

btracey:
As an addendum, we only really need to know a boolean for FunctionValue, so that's an additional argument that this other approach is better, since this current approach makes setting everything a two-line process and introduces additional complexity to the user, while the alternate approach only has one odd pathology.

I'll switch and resubmit a PR

Issue #486, opened 2018-05-09
path: allow implicit graph path search for A*, Dijkstra and Bellman-Ford


kortschak:
~~All of A*,~~ Dijkstra ~~and Bellman-Ford~~ do not in principle need to know the identities of the nodes from the outset of the algorithm. Currently this is known and that is made use of by the constructor for the `Shortest` that will be returned. However, I suspect that we can construct this successively in cases where it's not known. This would allow us to trim down the interface taken by these functions to something similar to the [recent change in traverse](https://github.com/gonum/gonum/issues/481). There are two possible approaches if there is a feasible way of incrementally constructing the `Shortest`:

1. Have a separate function for each of these types that takes the reduced interface.
2. Conditionally use the `Nodes` method to construct the `Shortest` and the `Has` method to determine existence, falling back to incremental construction and `From(u) != nil` if they don't exist.

PR #498, opened 2018-05-17
graph/path: allow incremental path building for Dijkstra

kortschak:
Please take a look.

BellmanFord is more subtle and I need to think about it some more. There is room for optimisation here, but I think overall this is a win in terms of flexibility and in cases where there are large disconnected graphs, a reduction in overall space.

Updates #486


kortschak:
Actually re-reading Bellman-Ford and this is not possible there except very inefficiently. So I will no do that change.

kortschak:
Closes #486.

Issue #461, opened 2018-04-13
mat: document that EigenSym decomposition returns values in ascending order


kortschak:
See https://github.com/gonum/gonum/pull/458#discussion_r181452960.

kortschak:
Also, it appears that the documentation for `lapack64.Geev` does not state the order of Eigen values.

vladimir-ch:
For `EigenSym` this has been done but for `Eigen`/`Dgeev` there is no defined order. I think this can be closed.

PR #501, opened 2018-05-21
mat: note that EigenSym orders values ascending

kortschak:
Please take a look.

Updates #461.

### DIFF
--- a/mat/eigen.go
+++ b/mat/eigen.go
@@ -27,8 +27,9 @@ type EigenSym struct {
 // The Eigen decomposition is defined as
 //  A = P * D * P^-1
 // where D is a diagonal matrix containing the eigenvalues of the matrix, and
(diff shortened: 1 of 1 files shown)


mat/eigen.go
```diff
@@ -27,8 +27,9 @@ type EigenSym struct {
 // The Eigen decomposition is defined as
 //  A = P * D * P^-1
 // where D is a diagonal matrix containing the eigenvalues of the matrix, and
-// P is a matrix of the eigenvectors of A. If the vectors input argument is
-// false, the eigenvectors are not computed.
+// P is a matrix of the eigenvectors of A. Factorize computes the eigenvalue
```
vladimir-ch: Why not just "Factorize computes the eigenvalues in ascending order."?

kortschak: Good question. Changed.

Issue #442, opened 2018-03-15
mat: No panic on zero sized matrices


btracey:
The code in issue #441 panics in `blas`, but it should have panicked in `mat` first somewhere. I'm not exactly sure where. The issue there is that `Simplex` is calling MulVec with a 0x2 matrix times a length 2 vector. The dimensions work, and so MulVec doesn't panic, but since there's no real receiver, it does panic inside BLAS. I'm not sure what our policy should be on things like this. Should we forbade making matrices with 0 size, or should we be doing a case-by-case decision on what should happen with zero sized matrices (for example, 0 size could make sense with something like `Stack`?

kortschak:
This should probably a shape error panic, i.e. we disallow all n×0 and 0×m matrices. I don't think `Stack` or `Augment` gain anything by being allowed to take these shaped matrices as inputs.

vladimir-ch:
I agree, I wouldn't allow zero dimensions. This should panic.

btracey:
SGTM

PR #502, opened 2018-05-21
mat: disallow zero dimension results

kortschak:
Please take a look.

I would have liked to add a zero-size check in the construction functions for each of these types. But there are a number of places (particularly in list_test.go) that use calls to `New*(0,...` (list_test can special-case 0 and just return the type, which would be better anyway since we say the zero value is usable). The approach here covers the majority of cases though (note that it does not catch stack and augment with one zero and one non-zero dimensioned input - this is arguably OK).

The io routines do not panic, since the user cannot know prior to reading whether the data are zero.

I'd like to be able to differentiate zero values that the user has made from zero values that we've incorrectly passed into a routine and panic harder in the latter case, but this doesn't seem easily doable. I may come back to it.

I have also added in missing negative dimension checks where needed.

Updates #442.

### DIFF
--- a/mat/dense.go
+++ b/mat/dense.go
@@ -42,6 +42,9 @@ type Dense struct {
 // The data must be arranged in row-major order, i.e. the (i*c + j)-th
 // element in the data slice is the {i, j}-th element in the matrix.
 func NewDense(r, c int, data []float64) *Dense {
+	if r < 0 || c < 0 {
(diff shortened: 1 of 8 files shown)


kortschak:
I'll wait for @btracey and @sbinet.

kortschak:
Yes, or at least we discuss whether that's wanted (I'm ambivalent about it).

Issue #377, opened 2018-01-03
mat: shadow detection misses some cases


kortschak:
I have made some optimisation-driven mistakes in the checks for shadow detection; it is the case that these checks for many cases are only made when all input values are `Raw*` types, this means that we are not doing what we say in the documentation, that all `Raw*` types are checked.

btracey:
Should I look into somehow integrating this into `ListTest` so that we can test that the right overlaps panic?

kortschak:
I don't think this is API stability. And I think I recall that I well through that check whether there were any more cases. I don't think there are.

PR #503, opened 2018-05-21
mat: improve mat element shadowing detection

kortschak:
Please take a look.

Updates #377.


btracey:
Is there utility in trying to add a test of this to list_test (or something)? It's easy to review what is here, but harder to review what's missing. Maybe we could automatically check that functions correctly panic with overlap.

kortschak:
I think so - this is why I wrote "Updates" rather than "Fixes".

PR #504, opened 2018-05-21
mat: test that SymEigen eigenvalues are in ascending order

vladimir-ch:
Please take a look.

### DIFF
--- a/mat/eigen_test.go
+++ b/mat/eigen_test.go
@@ -5,6 +5,7 @@
 package mat
 
 import (
+	"sort"
 	"testing"
 
 	"golang.org/x/exp/rand"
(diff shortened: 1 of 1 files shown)


mat/eigen_test.go
```diff
@@ -160,6 +160,13 @@ func TestSymEigen(t *testing.T) {
 					t.Errorf("Eigenvalue does not match")
 				}
 			}
+
+			// Check that the eigenvalues are in ascending order.
```
kortschak: `if !sort.Float64sAreSorted(es.values) {`

vladimir-ch: Oh. Done.

kortschak:
Is the same thing true of Eigen? I don't think `lapacke.Dgeev` says so.

vladimir-ch:
`Dgeev` gives complex eigenvalues in general but there is no natural ordering for complex numbers. We could do _something_ but I think it's not worth it.

kortschak:
Good point.

PR #510, opened 2018-05-25
stat/distuv: change KL comment spelling and distance to divergence

btracey:
Please take a look.


btracey:
PTAL @kortschak  @vladimir-ch

PR #512, opened 2018-05-27
gonum: add initial vgo support

sbinet:
Please take a look.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,0 +1,6 @@
+module gonum.org/v1/gonum
+
+require (
+	golang.org/x/exp v0.0.0-20180321215751-8460e604b9de
+	golang.org/x/tools v0.0.0-20180525024113-a5b4c53f6e8b
+)


kortschak:
Is there some testing of `vgo` with this `go.mod` that you can add that makes sure it's working correctly?

sbinet:
I've added some tentative code that verifies the vgo module deps.
but for a complete test, I suppose one would need to test with *clients* of `gonum.org/v1/gonum` :)

sbinet:
arf... `vgo` can only be compiled with Go >= 1.10.x...

kortschak:
Can the tests not be put in a script that runs them conditionally on `$(go version)`?

sbinet:
Sure. Would be easier to have them written directly in Go and put in the 'tools' repo :)

kortschak:
In this case I'm not sure it would, but maybe.

sbinet:
I'd be tempted to remove (for now) the `vgo` test and only enable them when `Go-1.11` hits the door.
having a `go.mod` file now will allow people living with `go-1.9.x`, `go-1.10.x` and `master` to experiment and play with `vgo`.

what do you think?

kortschak:
SGTM

sbinet:
updated with latest `vgo`:
```
$> vgo version
go version devel +b3562658fd Sun Jun 10 17:04:22 2018 +0000 linux/amd64 vgo:2018-02-20.1
```

sbinet:
PTAL

kortschak:
I take it the netlib dep is due to the cblas test flag. Will this force users to pull that? Maybe edit that line out.

sbinet:
done.

PR #514, opened 2018-05-31
stat/distuv: Add Bhattacharyya and Hellinger distances for Beta and N…

btracey:
…ormal distributions

Please take a look.

### DIFF
--- a/stat/distuv/statdist.go
+++ b/stat/distuv/statdist.go
@@ -18,7 +18,7 @@ import (
 //  BC = \int_-∞^∞ (p(x)q(x))^(1/2) dx
 // Where BC is known as the Bhattacharyya coefficient.
 // The Bhattacharyya distance is related to the Hellinger distance by
-//  H = sqrt(1-BC)
+//  H(l,r) = sqrt(1-BC(l,r))
 // For more information, see
 //  https://en.wikipedia.org/wiki/Bhattacharyya_distance
 type Bhattacharyya struct{}
@@ -52,7 +52,7 @@ func (Bhattacharyya) DistNormal(l, r Normal) float64 {
 // and is bounded between 0 and 1. Note the above formula defines the squared
 // Hellinger distance, while this returns the Hellinger distance itself.
 // The Hellinger distance is related to the Bhattacharyya distance by
-//  H^2 = 1 - exp(-Db)
+//  H^2 = 1 - exp(-D_B)
 // For more information, see
 //  https://en.wikipedia.org/wiki/Hellinger_distance
 type Hellinger struct{}
--- a/stat/distuv/statdist.go
+++ b/stat/distuv/statdist.go
@@ -10,6 +10,69 @@ import (
 	"gonum.org/v1/gonum/mathext"
 )
 
+// Bhattacharyya is a type for computing the Bhattacharyya distance between
+// probability distributions.
+//
+// The Bhattacharyya distance is defined as
+//  D_B = -ln(BC(l,r))
+//  BC = \int_-∞^∞ (p(x)q(x))^(1/2) dx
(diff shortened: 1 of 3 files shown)


stat/distuv/statdist.go
```diff
@@ -10,6 +10,69 @@ import (
 	"gonum.org/v1/gonum/mathext"
 )
 
+// Bhattacharyya is a type for computing the Bhattacharyya distance between
+// probability distributions.
+//
+// The Bhattacharyya distance is defined as
+//  D_B = -ln(BC(l,r))
+//  BC = \int_-∞^∞ (p(x)q(x))^(1/2) dx
```
sbinet: `\int_-∞^∞` looks a bit funny.
I am not strongly opinionated on this issue but perhaps we could just use a latex-ish syntax here as well?

btracey: It is latex-ish. It could be made real latex with `\int_{-\infty}^\infty` but that seems like the wrong balance in clarity.

stat/distuv/statdist.go
```diff
@@ -10,6 +10,69 @@ import (
...
+}
+
+// DistNormal returns the Bhattacharyya distance Normal distributions l and r.
+// For Normal distributions, the Bhattacharyya distance is given by
+//  s = (σ_l^2 + σ_r^2)/2
+//  BC = 1/8 (μ_l-μ_r)^2/s + 1/2 ln(s/(σ_l*σ_r))
+func (Bhattacharyya) DistNormal(l, r Normal) float64 {
+	// Reference: https://en.wikipedia.org/wiki/Bhattacharyya_distance
+	m := l.Mu - r.Mu
+	s := (l.Sigma*l.Sigma + r.Sigma*r.Sigma) / 2
```
kortschak: I think this would be more clear (given the ref to the example at WP) if the division here were removed and done in the expression below, `0.25*m*m/s + 0.5*math.Log(s/2) - 0.5*math.Log(l.Sigma) - 0.5*math.Log(r.Sigma)`. No?

stat/distuv/statdist.go
```diff
@@ -10,6 +10,69 @@ import (
...
+
+// Hellinger is a type for computing the Hellinger distance between probability
+// distributions.
+//
+// The Hellinger distance is defined as
+//  H^2(l,r) = 1/2 * int_x (\sqrt(l(x)) - \sqrt(r(x)))^2 dx
+// and is bounded between 0 and 1. Note the above formula defines the squared
+// Hellinger distance, while this returns the Hellinger distance itself.
+// The Hellinger distance is related to the Bhattacharyya distance by
+//  H^2 = 1 - exp(-Db)
```
kortschak: s/Db/D_B/ as used above.

stat/distuv/statdist.go
```diff
@@ -10,6 +10,69 @@ import (
...
 
+// Bhattacharyya is a type for computing the Bhattacharyya distance between
+// probability distributions.
+//
+// The Bhattacharyya distance is defined as
+//  D_B = -ln(BC(l,r))
+//  BC = \int_-∞^∞ (p(x)q(x))^(1/2) dx
+// Where BC is known as the Bhattacharyya coefficient.
+// The Bhattacharyya distance is related to the Hellinger distance by
+//  H = sqrt(1-BC)
```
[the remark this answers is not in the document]
btracey: Done.

stat/distuv/statdist.go
btracey: I see. I was simplifying the multivariate normal formula (and missed the equation for the single). I personally think the multivariate formula is much clearer what is going on (in that the terms are clearer), but I'm happy to do it either way

kortschak: This is fine.

PR #515, opened 2018-06-03
lapack/gonum: fix doc comment for Dtrevc3

vladimir-ch:
Please take a look.

### DIFF
--- a/lapack/gonum/dtrevc3.go
+++ b/lapack/gonum/dtrevc3.go
@@ -21,9 +21,8 @@ import (
 // The right eigenvector x of T corresponding to an
 // eigenvalue λ is defined by
 //  T x = λ x,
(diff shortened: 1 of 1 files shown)


kortschak:
Please remember to send a PR with the change to netlib.

vladimir-ch:
I think your local copy of Reference LAPACK is not up-to-date, this was fixed recently there. My mind must have been wandering somewhere else when I was adapting their comment.

Dtrevc3 is not part of lapacke, so this time there no change to netlib.

kortschak:
Thanks. The Dtrevc3 I was looking at was [from the NETLIB site](http://www.netlib.org/lapack/explore-html/d5/d0e/dtrevc3_8f_source.html) rather than local. I suppose that is updated a little slower?

PR #516, opened 2018-06-03
graph/formats/dot: allow unicode inside html labels

egonelbre:
Please take a look.

I added support for unicode characters inside html labels.

I also replaced the shell script for pasting the copyright tag with go code, so that code could be generated on Windows.


kortschak:
I didn't realise you hadn't made a contribution before. Can you add yourself to A+C?

PR #518, opened 2018-06-05
stat: refactor Quantile to prepare for more CumulantKinds

sbinet:
This CL exposes some hidden documentation about stat.CumulantKinds and
refactors the current empirical Quantile into its own function.

Please take a look.


sbinet:
I'll wait a bit more for @kortschak :)

PR #519, opened 2018-06-07
stat: fix panic message

kortschak:
Please take a look.

Closes #252.


kortschak:
I'm seeing a number of failures (panics) on go master. They seem to be related to shadow detection. I will look into it.

PR #520, opened 2018-06-07
all: fix up godoc rendering of package lists

kortschak:
Please take a look.


mathext/internal/amos/doc.go
```diff
@@ -2,5 +2,5 @@
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
-// package amos implements functions originally in the Netlab code by Donald Amos.
+// Package amos implements functions originally in the Netlab code by Donald Amos.
```
btracey: Netlib

PR #523, opened 2018-06-07
blas/gonum internal/asm/f64: add Dgemv assembly kernels

Kunde21:
Two kernels for calculating Dgemv, depending upon the matrix tranpose:  

GemvN - No transpose of matrix A  y = alpha * A * x + beta * y  
GemvT - Transpose matrix A        y = alpha * A^T * x + beta * y

PR #524, opened 2018-06-08
blas/gonum: remove redundant assignments

kortschak:
Prompted by looking at `f64.Dgemv` code in #523.

Please take a look.

### DIFF
--- a/blas/gonum/level2double.go
+++ b/blas/gonum/level2double.go
@@ -58,14 +58,10 @@ func (Implementation) Dgemv(tA blas.Transpose, m, n int, alpha float64, a []floa
 	}
 
 	var kx, ky int
-	if incX > 0 {
-		kx = 0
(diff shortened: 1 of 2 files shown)


blas/gonum/level2double.go
```diff
@@ -58,14 +58,10 @@ func (Implementation) Dgemv(tA blas.Transpose, m, n int, alpha float64, a []floa
 	}
 
 	var kx, ky int
-	if incX > 0 {
-		kx = 0
-	} else {
+	if incX <= 0 {
```
Kunde21: `incX == 0` also results in `kx = 0`.  Makes more sense to use `<` test here.

Same in the other changes.

kortschak: Good point. I was doing a blind inversion.

Issue #474, opened 2018-04-23
graph/path: feature request: K shortest path implementation


gsidebottom:
Nice library, but I didn't see a K shortest paths algorithm such as Yen's algorithm (maybe I missed it).  https://github.com/starwander/goraph has one.  Any interested in adding KSP to gonum?

kortschak:
Functions are usually added on a need basis by those who want them or (more than occasionally) me. However my time is limited at the moment.

The API for addition of those would be to take a `path.Shortest` and perform Yen on it if you want to pick it up.

apache8080:
I am interested in working on this. I will keep you guys updated on how far I get.

apache8080:
@kortschak Does gonum have a shortest path from nodeA to nodeB? I can't seem to find it in the dijkstra.go file. I am also using the [Wikipedia](https://en.wikipedia.org/wiki/Yen%27s_algorithm) implementation for Yen.

sergeymsui:
Hi! If the task is still relevant, then I can take it on.

kortschak:
This was added in 0d927069219651367007a1ff1931bf06245acfeb. Closing.

PR #525, opened 2018-06-08
graph/path: add Yen's k-shortest paths implementation

apache8080:
This PR addresses issue #474. Currently it has my initial attempt at implementing Yen's algorithm for k-shortest path. I still need to add test cases but I wanted to get some initial feedback on my current progress.


graph/path/yen_ksp.go
```diff
@@ -0,0 +1,81 @@
...
+			var rootWeight float64
+			for x := 1; x < len(root); x++ {
+				rootWeight += g.Weight(root[x - 1], root[x])
+			}
+			
+			var edges []graph.Edge
+			var nodes []graph.Node
+
+			for _, p := range paths {
+				if reflect.DeepEqual(p.nodes[:i], root) {
```
kortschak: You can't do this. There is no requirement in the `graph.Node` contract that nodes are comparable. The only test of equality of nodes is that their IDs match.

graph/path/yen_ksp.go
```diff
@@ -0,0 +1,81 @@
...
+			for _, node := range root {
+				if (node != spur) {
+					append(nodes, node)
+					g.RemoveNode(node)
+				}
+			}
+
+			spath, weight := DjikstraFrom(spur, g).To(sink.ID())
+			var total []graph.Node
+			append(total, root, spath)
```
kortschak: `append` returns the modified slice, this will leave `total` unmodified (throughout).

graph/path/yen_ksp.go
```diff
@@ -0,0 +1,81 @@
...
+			potential := YenShortest {total, weight + rootWeight}
+			append(pot, potential)
```
kortschak: Same here.

graph/path/yen_ksp.go
```diff
@@ -0,0 +1,81 @@
+// Copyright ©2015 The Gonum Authors. All rights reserved.
```
kortschak: 2018

graph/path/yen_ksp.go
```diff
@@ -0,0 +1,81 @@
...
+	"gonum.org/v1/gonum/graph"
+)
+
+type YenShortest struct {
+	path []graph.node,
+	weight float64
+}
+
+func YenKSP(source graph.Node, sink graph.Node, g graph.Graph, k int) []Shortest {
+	paths := make([]Shortest, k)
```
kortschak: This doesn't seem like the right thing to do. We don't know that there will be `k` paths, just that there will be at most `k` paths. The WP-published algorithm does this (sort of), but it is horrifically inefficient as an approach to returning `k` paths from s to t. The Dijkstra that is being used in the article is just returning the shortest path from s to t, while we return all shortest paths from s. What should happen here is that you reuse the `Shortest` you have to recalculate the shortest path after the graph pseudomutation and then keep the actual path (a `[]graph.Node` is fine for this) and finally return a `[][]graph.Node`.

graph/path/yen_ksp.go
```diff
@@ -0,0 +1,81 @@
...
+				rootWeight += g.Weight(root[x - 1], root[x])
+			}
+			
+			var edges []graph.Edge
+			var nodes []graph.Node
+
+			for _, p := range paths {
+				if reflect.DeepEqual(p.nodes[:i], root) {
+					append(edges, g.Edge(i, i + 1))
+					g.RemoveEdge(i, i + 1)
```
kortschak: I don't think it's reasonable to mutate he graph being handed to the routine. What you need to do is wrap the graph so that you can filter the edges that the graph believes exist.

graph/path/yen_ksp.go
```diff
@@ -0,0 +1,81 @@
...
+				}
+			}
+
+			for _, node := range root {
+				if (node != spur) {
+					append(nodes, node)
+					g.RemoveNode(node)
```
kortschak: Same here

graph/path/yen_ksp.go
```diff
@@ -0,0 +1,81 @@
...
+				g.SetEdge(edge)
+			}
+
+		}
+
+		if (len(pot) == 0) {
+			break
+		}
+
+		sort.Slice(pot, func(a, b int) bool {
```
kortschak: Please write a sorting type for this.

graph/path/yen_ksp.go
```diff
@@ -0,0 +1,81 @@
...
+
+import (
+	"reflect"
+	"sort"
+	
+	"gonum.org/v1/gonum/graph"
+)
+
+type YenShortest struct {
+	path []graph.node,
```
kortschak: No comma in a struct type declaration.

apache8080:
Thanks for the comments. I will go over it tonight.

apache8080:
I was first trying to implement the WP solution and then improve that. I am a bit stuck on adding back the graph edges that I removed. On one of your comments @kortschak, you were saying to wrap the graph; what do you mean by this, like should I just copy the graph to a local variable.

kortschak:
Graph types are interfaces in gonum/graph. This means that you can make a new type that holds the original graph and alters the behaviour. There are examples of this in path where a graph is wrapped to allows the S node to be added in [Johnson's All Pairs Shortest Paths](https://github.com/gonum/gonum/blob/master/graph/path/johnson_apsp.go#L68-L142). We don't copy things unless we can't avoid it.

apache8080:
@kortschak, I have the Wikipedia solution working now.

graph/path/yen_ksp.go
```diff
@@ -0,0 +1,144 @@
...
+package path
+
+import (
+	"fmt"
+	"sort"
+
+	"gonum.org/v1/gonum/graph"
+)
+
+type YenShortest struct {
```
kortschak: Please unexport this.

graph/path/yen_ksp.go
```diff
@@ -0,0 +1,144 @@
...
+	path   []graph.Node
+	weight float64
+}
+
+func YenKSP(source graph.Node, sink graph.Node, g graph.Graph, k int) [][]graph.Node {
```
kortschak: This needs documentation; a doc comment above the func for godoc and a comment including a link to the source of the algorithm used here.

graph/path/yen_ksp.go
```diff
@@ -0,0 +1,144 @@
...
+	yk := yenKSPAdjuster{
+		g:       g,
+		visited: make([]int64, 0),
+		from:    g.From,
+		weight:  g.(graph.Weighted).Weight,
+	}
+
+	paths := make([][]graph.Node, k)
+
```
kortschak: Delete this blank line.

graph/path/yen_ksp.go
```diff
@@ -0,0 +1,144 @@
...
+	paths[0], _ = DijkstraFrom(source, yk).To(sink.ID())
+
+	var pot []YenShortest
+
```
kortschak: Same here.

graph/path/yen_ksp.go
```diff
@@ -0,0 +1,144 @@
...
+	for i := int64(1); i < int64(k); i++ {
+		fmt.Printf("%d \n", i)
```
kortschak: Delete this print.

graph/path/yen_ksp.go
```diff
@@ -0,0 +1,144 @@
...
+		for n := 0; n < len(paths[i-1])-2; n++ {
+			fmt.Printf("%d \n", n)
```
kortschak: Same here.

graph/path/yen_ksp.go
```diff
@@ -0,0 +1,144 @@
...
+			spur := paths[i-1][n]
+			root := paths[i-1][:n]
+
+			var rootWeight float64
+			for x := 1; x < len(root); x++ {
+				w, _ := g.(graph.Weighted).Weight(root[x-1].ID(), root[x].ID())
+				rootWeight += w
+			}
+
+			//var edges []graph.Edge
```
kortschak: Delete these commented out decls.

graph/path/yen_ksp.go
```diff
@@ -0,0 +1,144 @@
...
+			for _, path := range paths {
+				for x := 0; x < len(root); x++ {
+					if x < len(path) {
+						if path[x].ID() != root[x].ID() {
+							break
+						}
+					}
+				}
+
+				//if contains {
```
kortschak: Delete this and other commented-out code.

graph/path/yen_ksp.go
```diff
@@ -0,0 +1,144 @@
...
+			//	g.(graph.WeightedEdgeAdder).SetWeightedEdge(edge.(graph.WeightedEdge))
+			//}
+
+		}
+
+		if len(pot) == 0 {
+			break
+		}
+
+		sort.Slice(pot, func(a, b int) bool {
```
kortschak: Please use a `sort.Interface` rather than `sort.Slice`.

graph/path/yen_ksp.go
```diff
@@ -0,0 +1,144 @@
...
+		})
+
+		paths[i] = pot[0].path
+		pot = pot[1:]
+	}
+
+	return paths
+}
+
+// TODO Wrap graph in YenKSPAdjuster interface and override the From method
```
kortschak: TODOs should be in the form of `TODO(name): ...` where name is the person most likely to know what is going on. If that is you, please then also expand on the text here so that I know what you are intending.

graph/path/yen_ksp.go
```diff
@@ -0,0 +1,144 @@
...
+
+	from   func(id int64) []graph.Node
+	edgeTo func(uid, vid int64) graph.Edge
+	weight func(xid, yid int64) (w float64, ok bool)
+}
+
+func (g yenKSPAdjuster) From(id int64) []graph.Node {
+	if contains(g.visited, id) {
+		return nil
+	} else {
```
kortschak: Remove the else and deindent; with a return there is no need for the else block.

graph/path/yen_ksp.go
```diff
@@ -0,0 +1,144 @@
...
+		paths[i] = pot[0].path
+		pot = pot[1:]
+	}
+
+	return paths
+}
+
+// TODO Wrap graph in YenKSPAdjuster interface and override the From method
+
+type yenKSPAdjuster struct {
```
kortschak: Given that this is an important part of the implementation, please document the function of `yenKSPAdjuster` and its fields.

graph/path/yen_ksp.go
```diff
@@ -0,0 +1,144 @@
...
+	path   []graph.Node
+	weight float64
+}
+
+func YenKSP(source graph.Node, sink graph.Node, g graph.Graph, k int) [][]graph.Node {
+	yk := yenKSPAdjuster{
+		g:       g,
+		visited: make([]int64, 0),
+		from:    g.From,
+		weight:  g.(graph.Weighted).Weight,
```
kortschak: This will panic if g is not a `graph.Weighted`. If it's the case that you always expect g to be a weighted graph, specify that in the signature.

graph/path/yen_ksp.go
```diff
@@ -0,0 +1,144 @@
...
+
+type YenShortest struct {
+	path   []graph.Node
+	weight float64
+}
+
+func YenKSP(source graph.Node, sink graph.Node, g graph.Graph, k int) [][]graph.Node {
+	yk := yenKSPAdjuster{
+		g:       g,
+		visited: make([]int64, 0),
```
kortschak: This is unnecessary.

graph/path/yen_ksp.go
```diff
@@ -0,0 +1,144 @@
...
+
+func (g yenKSPAdjuster) Edge(uid, vid int64) graph.Edge {
+	return g.edgeTo(uid, vid)
+}
+
+func (g yenKSPAdjuster) Weight(xid, yid int64) (w float64, ok bool) {
+	return g.weight(xid, yid)
+}
+
+func contains(visited []int64, id int64) bool {
```
kortschak: This becomes problematic for large graphs. I'd suggest using a [`set.Int64s`](https://godoc.org/gonum.org/v1/gonum/graph/internal/set#Int64s).

graph/path/yen_ksp.go
```diff
@@ -0,0 +1,144 @@
...
+		pot = pot[1:]
+	}
+
+	return paths
+}
+
+// TODO Wrap graph in YenKSPAdjuster interface and override the From method
+
+type yenKSPAdjuster struct {
+	g       graph.Graph
```
kortschak: If you embed `graph.Graph` rather than naming it (`g`), you get all the methods you want for free and you only need to provide those that are changed.

graph/path/yen_ksp_test.go
```diff
@@ -0,0 +1,34 @@
+package path
```
kortschak: Please add a copyright notice.

graph/path/yen_ksp_test.go
```diff
@@ -0,0 +1,34 @@
...
+		{F: simple.Node(4), T: simple.Node(5), W: 2},
+	}
+
+	for _, edge := range edges {
+		g.SetWeightedEdge(edge)
+	}
+
+	shortests := YenKSP(simple.Node(0), simple.Node(5), g, 3)
+	
+	fmt.Printf("%v", shortests)
```
kortschak: Please use "testing" rather than "fmt".

Also, read [this](https://github.com/golang/go/wiki/TableDrivenTests).

apache8080:
Realized that my original implementation was actually incorrect. I have corrected this, but I still need to do some of the cleanups that @kortschak asked for earlier.

graph/path/yen_ksp.go
```diff
@@ -0,0 +1,130 @@
...
+	}
+
+	return paths
+}
+
+type yenKSPAdjuster struct {
+	g graph.Weighted
+	visited map[int64]set.Int64s
+
+	from   func(id int64) []graph.Node
```
kortschak: This field is not necessary, you can just use `g.g.From()` below.

graph/path/yen_ksp.go
```diff
@@ -0,0 +1,130 @@
...
+package path
+
+import (
+	"sort"
+
+	"gonum.org/v1/gonum/graph"
+	"gonum.org/v1/gonum/graph/internal/set"
+)
+
+type yenShortest struct {
```
kortschak: This appears to be dead. Please remove it.

graph/path/yen_ksp.go
```diff
@@ -0,0 +1,130 @@
...
+			nodes[int64(i)] = nodes[len(nodes)-1]
+			nodes = nodes[:len(nodes)-1]
+			i--;
+		}
+	}
+
+	return nodes
+}
+
+func (g yenKSPAdjuster) AddVisited(parent, id int64) {
```
kortschak: This method does not need to be exported.

I would implement the method as
```
visited, ok := g.visited[parent]
if ok {
    visited = make(set.Int64)
    g.visited[parent] = visited
}
visited.Add(id)
```

This saves 1 of 3 map index operations.

graph/path/yen_ksp.go
```diff
@@ -0,0 +1,130 @@
...
+
+type yenKSPAdjuster struct {
+	g graph.Weighted
+	visited map[int64]set.Int64s
+
+	from   func(id int64) []graph.Node
+}
+
+func (g yenKSPAdjuster) From(id int64) []graph.Node {
+	nodes := g.from(id)
```
kortschak: `nodes := g.g.From(id)`

graph/path/yen_ksp.go
```diff
@@ -0,0 +1,130 @@
...
+type yenShortest struct {
+	p   []graph.Node
+	weight float64
+}
+
+func YenKSP(source graph.Node, sink graph.Node, g graph.Weighted, k int) [][]graph.Node {
+	yk := yenKSPAdjuster{
+		g:       g,
+		visited: make(map[int64]set.Int64s),
+		from:    g.From,
```
kortschak: Delete this line.

graph/path/yen_ksp.go
```diff
@@ -0,0 +1,130 @@
...
+	"gonum.org/v1/gonum/graph"
+	"gonum.org/v1/gonum/graph/internal/set"
+)
+
+type yenShortest struct {
+	p   []graph.Node
+	weight float64
+}
+
+func YenKSP(source graph.Node, sink graph.Node, g graph.Weighted, k int) [][]graph.Node {
```
kortschak: Conventionally we use `s` and `t`, and I think YenKSP as a name could be improved on: `YenKShortestPaths(g graph.Graph, k int, s, t graph.Node) [][]graph.Node`.

The weighting should be determined dynamically using unit cost in the case that `g` is not a `graph.Weighted`.

graph/path/yen_ksp.go
```diff
@@ -0,0 +1,130 @@
...
+	yk := yenKSPAdjuster{
+		g:       g,
+		visited: make(map[int64]set.Int64s),
+		from:    g.From,
+	}
+
+	paths := make([][]graph.Node, k)
```
kortschak: k is a maximum, so use `paths := make([][]graph.Node, 0, k)`.

graph/path/yen_ksp.go
```diff
@@ -0,0 +1,130 @@
...
+
+		if len(pot) == 0 {
+			break
+		}
+
+		sort.Slice(pot, func(a, b int) bool {
+			return pot[a].weight < pot[b].weight
+		})
+
+		paths[i] = pot[0].p
```
kortschak: `paths = append(paths, pot[0].p)`

graph/path/yen_ksp.go
apache8080: No, yenShortest is still being used to sort the found shortest in order.

kortschak: Ah, sorry yes. Can you move it below the main function when you add the `sort.Interface` implementation. Normally we introduce things after they are used rather than before.

graph/path/yen_ksp.go
apache8080: How does uniform cost work? Currently, I am adding up the root weight to each computed spur weight; is it possible to do that using uniform cost?

kortschak: Each edge just has a weight of 1. If you handed in a weighted graph where each edge had unit weight, that would be the same.

In practice, make a `Weighting` (`func(xid, yid int64) (w float64, ok bool)`) and assign it either the `g.Weight` method or `UniformCost(g)` like so:
```
	var weight Weighting
	if wg, ok := g.(Weighted); ok {
		weight = wg.Weight
	} else {
		weight = UniformCost(g)
	}
```

Then use `weight` to determine weights where you currently use `g.Weight`.

graph/path/yen_ksp.go
```diff
@@ -123,7 +130,13 @@ func (g yenKSPAdjuster) Edge(uid, vid int64) graph.Edge {
 }
 
 func (g yenKSPAdjuster) Weight(xid, yid int64) (w float64, ok bool) {
-	return g.g.Weight(xid, yid)
+	var weight Weighting
```
kortschak: Don't do this each time. Add a `weight func(xid, yid int64) (w float64, of bool)` field that you determined in the preamble for `YenKShortest` and call that here as

```
func (g yenKSPAdjuster) Weight(xid, yid int64) (w float64, ok bool) {
    return g.weight(xid, yid)
}
```

graph/path/yen_ksp.go
kortschak: This has not been done. Please use `sort.Interface` instead of `sort.Slice`.

graph/path/yen_ksp.go
```diff
@@ -0,0 +1,153 @@
...
+
+import (
+	"sort"
+
+	"gonum.org/v1/gonum/graph"
+	"gonum.org/v1/gonum/graph/internal/set"
+)
+
+
+func YenKShortestPath(g graph.Graph, k int, s, t graph.Node) [][]graph.Node {
```
kortschak: This needs some documentation.

graph/path/yen_ksp_test.go
```diff
@@ -0,0 +1,64 @@
...
+import (
+	"math"
+	"testing"
+
+	"gonum.org/v1/gonum/graph/simple"
+	"gonum.org/v1/gonum/graph"
+)
+
+
+func TestYenKSP(t *testing.T) {
```
kortschak: I'd like to see more test cases here, so a table of test cases would be the way to go. At the minimum, we need some cases where there are other implementations or obviously correct solutions. I would also like to see some test cases where there are fewer than k possible paths from s to t to see how the implementation deals with that, and cases where there are k equally weighted paths and k paths with some more costly paths before the k runout.

apache8080: Yeah, I am working on moving the test cases to a table, and then I will add some more. Thanks for all the help with this.

apache8080:
I added a simple test case into a table. I may need some help coming up with extra test cases for this.

graph/path/internal/testgraphs/yen_shortest.go
```diff
@@ -0,0 +1,59 @@
...
+package testgraphs
+
+import (
+	"math"
+	
+	"gonum.org/v1/gonum/graph"
+	"gonum.org/v1/gonum/graph/simple"
+)
+
+func init() {
```
kortschak: This is not needed here.

graph/path/internal/testgraphs/yen_shortest.go
```diff
@@ -0,0 +1,59 @@
+// Copyright ©2018 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package testgraphs
```
kortschak: The tests here are pretty specific to YSK, so having them here is not necessary. I'd just have the test cases in the actual test file.

graph/path/yen_ksp_test.go
```diff
@@ -5,60 +5,32 @@
...
+		}
 
-	for i, sp := range shortests {
-		e := expected[i]
-		for n, p := range sp {
-			if (e[n].ID() != p.ID()) {
-				t.Errorf("ERROR: path #%d expected: %d, got: %d", i+1, e[n].ID(), p.ID())
+		paths := YenKShortestPath(g.(graph.Graph), test.K, test.Query.From(), test.Query.To())
+		for i := 0; i < len(test.WantPaths); i++ {
+			expected := test.WantPaths[i]
```
kortschak: s/expected/want/g
s/WantPaths/want/g

graph/path/yen_ksp_test.go
```diff
@@ -5,60 +5,32 @@
...
+			path := paths[i]
+			if (len(expected) == len(path)) {
+				for n := 0; n < len(path); n++ {
+					if (expected[n] != path[n].ID()) {
+						t.Errorf("ERROR: path #%d expected: %d, got: %d", i+1, expected[n], path[n].ID())
```
kortschak: `"unexpected path for %q: got:%v want:%v", test.name, got, test.want`

graph/path/yen_ksp_test.go
```diff
@@ -5,60 +5,32 @@
...
 
-	for i, sp := range shortests {
-		e := expected[i]
-		for n, p := range sp {
-			if (e[n].ID() != p.ID()) {
-				t.Errorf("ERROR: path #%d expected: %d, got: %d", i+1, e[n].ID(), p.ID())
+		paths := YenKShortestPath(g.(graph.Graph), test.K, test.Query.From(), test.Query.To())
+		for i := 0; i < len(test.WantPaths); i++ {
+			expected := test.WantPaths[i]
+			path := paths[i]
```
kortschak: This is not safe if the result is shorter than want.

TBH, the best way to do this would be to make a `func([][]graph.Node) [][]int` helper and then just use `reflect.DeepEqual`.

graph/path/yen_ksp_test.go
```diff
@@ -5,60 +5,32 @@
...
+		for _, e := range test.Edges {
+			g.SetWeightedEdge(e)
+		}
 
-	for i, sp := range shortests {
-		e := expected[i]
-		for n, p := range sp {
-			if (e[n].ID() != p.ID()) {
-				t.Errorf("ERROR: path #%d expected: %d, got: %d", i+1, e[n].ID(), p.ID())
+		paths := YenKShortestPath(g.(graph.Graph), test.K, test.Query.From(), test.Query.To())
```
kortschak: s/paths/got/g

The type assertion is not needed.

graph/path/internal/testgraphs/yen_shortest.go
```diff
@@ -0,0 +1,59 @@
...
+	Name              string
+	Graph             func() graph.WeightedEdgeAdder
+	Edges             []simple.WeightedEdge
+
+	Query         simple.Edge
+	K int
+	WantPaths     [][]int64
+}{
+	// Positive weighted graphs.
+	{
```
kortschak: I'd like to see tests that have k > |paths|, tests on a G={V={0,1}, E={(0,1)}}, on a G={V={0}, E=∅}, on a G=star_n, and on a G={V={-1,1,2..n}, E={(-1,m),(m,1)}} for m∈[2,n] and s=-1, t=1. for k<|{m}|, k = |{m}| and k>|{m}|.

That should probably cover it unless you have some other test cases that are well known, which would be good to have if they exist.

graph/path/yen_ksp_test.go
```diff
@@ -5,32 +5,81 @@
...
+		WantPaths: [][]int64{
+			{0, 2, 3, 5},
+			{0, 2, 4, 5},
+			{0, 1, 3, 5},
+		},	
+	},
+}
+
+func toIntPath(nodePaths [][]graph.Node) [][]int64 {
+	paths := make([][]int64, 0)
```
kortschak: If you do this, it could be `var paths [][]int64`, but since you know the length of the input, you can make it the right length.

```
func pathIDs(paths [][]graph.Node) [][]int64 {
	ids := make([][]int64, len(paths)
	for i, p := range paths {
		ids[i] = make([]int64, len(p))
		for j, n := range p {
			ids[i][j] = n.ID()
		}
	}
	return ids
}
```

graph/path/yen_ksp_test.go
```diff
@@ -5,32 +5,81 @@
...
+	"reflect"
 	"testing"
 
-	"gonum.org/v1/gonum/graph/path/internal/testgraphs"
 	"gonum.org/v1/gonum/graph"
+	"gonum.org/v1/gonum/graph/simple"
 )
 
 
+var YenShortestPathTests = []struct {
```
kortschak: Please unexport this.

graph/path/yen_ksp_test.go
```diff
@@ -5,32 +5,81 @@
...
-		paths := YenKShortestPath(g.(graph.Graph), test.K, test.Query.From(), test.Query.To())
-		for i := 0; i < len(test.WantPaths); i++ {
-			expected := test.WantPaths[i]
-			path := paths[i]
-			if (len(expected) == len(path)) {
-				for n := 0; n < len(path); n++ {
-					if (expected[n] != path[n].ID()) {
-						t.Errorf("ERROR: path #%d expected: %d, got: %d", i+1, expected[n], path[n].ID())
-					}
+		paths := toIntPath(YenKShortestPath(g.(graph.Graph), test.K, test.Query.From(), test.Query.To()))
```
kortschak:
```
got := YenKShortestPath(g.(graph.Graph), test.K, test.Query.From(), test.Query.To())
gotIDs := pathIDs(got)
```

graph/path/yen_ksp_test.go
```diff
@@ -5,32 +5,81 @@
...
+
+		expected := test.WantPaths
```
kortschak: Please delete this line.

graph/path/yen_ksp_test.go
```diff
@@ -5,32 +5,81 @@
...
+		
+		if len(expected) != len(paths) {
```
kortschak: Please replace all this with
```
if !reflect.DeepEqual(got, test.want) {
	t.Errorf("unexpected result for %q:\ngot: %v\nwant:%v", test.name, gotIDs, test.want)
}
```

graph/path/yen_ksp_test.go
```diff
@@ -5,32 +5,81 @@
...
+	"gonum.org/v1/gonum/graph/simple"
 )
 
 
+var YenShortestPathTests = []struct {
+	Name              string
+	Graph             func() graph.WeightedEdgeAdder
+	Edges             []simple.WeightedEdge
+
+	Query         simple.Edge
```
kortschak: Please gofmt your code.

graph/path/yen_ksp_test.go
kortschak: Please use this code.

graph/path/yen_ksp_test.go
kortschak: And this.

graph/path/yen_ksp.go
```diff
@@ -0,0 +1,152 @@
...
+	"gonum.org/v1/gonum/graph/internal/set"
+)
+
+func YenKShortestPath(g graph.Graph, k int, s, t graph.Node) [][]graph.Node {
+	yk := yenKSPAdjuster{
+		g:       g,
+		visited: make(map[int64]set.Int64s),
+	}
+
+	var weight Weighting
```
kortschak: Since `yk.weight` is going to hold this `func`, just use that here.

graph/path/yen_ksp.go
```diff
@@ -0,0 +1,152 @@
...
+	if wg, ok := g.(Weighted); ok {
+		weight = wg.Weight
+	} else {
+		weight = UniformCost(g)
+	}
+
+	yk.weight = weight
```
kortschak: Delete this.

graph/path/yen_ksp.go
```diff
@@ -0,0 +1,152 @@
...
+
+	paths := make([][]graph.Node, 0, k)
+
+	shortest, _ := DijkstraFrom(s, yk).To(t.ID())
+	paths = append(paths, shortest)
+
+	var pot []yenShortest
+
```
kortschak: Delete this blank line.

graph/path/yen_ksp.go
```diff
@@ -0,0 +1,156 @@
...
+		weight = wg.Weight
+	} else {
+		weight = UniformCost(g)
+	}
+
+	yk.weight = weight
+
+	var paths [][]graph.Node
+
+	shortest, _ := DijkstraFrom(s, yk).To(t.ID())
```
kortschak:
```
shortest, _ := DijkstraFrom(s, yk).To(t.ID())
if len(shortest) == 0 {
    return nil
}
paths := [][]graph.Node{shortest}
```

graph/path/yen_ksp.go
```diff
@@ -0,0 +1,156 @@
...
+	if len(shortest) != 0 {
+		paths = append(paths, shortest)
+	} else {
+		return paths
+	}
+
+	var pot []yenShortest
+
```
kortschak: Please delete this blank line.

graph/path/yen_ksp.go
```diff
@@ -0,0 +1,156 @@
...
+	g       graph.Graph
+	visited map[int64]set.Int64s
+	weight  Weighting
+}
+
+func (g yenKSPAdjuster) From(id int64) []graph.Node {
+	nodes := g.g.From(id)
+
+	for i := 0; i < len(nodes); i++ {
+		if g.visited[id].Has(int64(nodes[i].ID())) {
```
kortschak: Remove the unneeded `int64` conversion.

graph/path/yen_ksp.go
```diff
@@ -0,0 +1,156 @@
...
+func (s byPathWeight) Swap(i, j int) {
+	s[i], s[j] = s[j], s[i]
+}
+
+func (s byPathWeight) Less(i, j int) bool {
+	return s[i].weight < s[j].weight
+}
+
+type yenKSPAdjuster struct {
+	g       graph.Graph
```
kortschak: Please embed `graph.Graph`.

graph/path/yen_ksp.go
```diff
@@ -0,0 +1,156 @@
...
+	visited, ok := g.visited[parent]
+	if !ok {
+		visited = make(set.Int64s)
+		g.visited[parent] = visited
+	}
+
+	visited.Add(id)
+}
+
+func (g yenKSPAdjuster) Edge(uid, vid int64) graph.Edge {
```
kortschak: Delete this method.

graph/path/yen_ksp.go
```diff
@@ -0,0 +1,156 @@
...
+	return g.g.Edge(uid, vid)
+}
+
+func (g yenKSPAdjuster) Weight(xid, yid int64) (w float64, ok bool) {
+	return g.weight(xid, yid)
+}
+
+func (g yenKSPAdjuster) HasEdgeBetween(xid, yid int64) bool {
```
kortschak: Delete this method.

graph/path/yen_ksp_test.go
```diff
@@ -0,0 +1,112 @@
...
+	"math"
+	"reflect"
+	"testing"
+
+	"gonum.org/v1/gonum/graph"
+	"gonum.org/v1/gonum/graph/simple"
+)
+
+var yenShortestPathTests = []struct {
+	Name  string
```
kortschak: Unexport these fields.

graph/path/yen_ksp.go
```diff
@@ -0,0 +1,156 @@
...
+package path
+
+import (
+	"sort"
+
+	"gonum.org/v1/gonum/graph"
+	"gonum.org/v1/gonum/graph/internal/set"
+)
+
+func YenKShortestPath(g graph.Graph, k int, s, t graph.Node) [][]graph.Node {
```
kortschak: Please add documentation for this:
```
// YenKShortestPath returns the k-shortest loopless paths from s to t in g. YenKShortestPath will
// panic if g contains a negative edge weight.
```

graph/path/yen_ksp.go
```diff
@@ -0,0 +1,156 @@
...
+						}
+					}
+					if ok {
+						yk.addVisited(path[n].ID(), path[n+1].ID())
+					}
+				}
+			}
+
+			spath, weight := DijkstraFrom(spur, yk).To(t.ID())
+			size := len(root) - 1
```
kortschak: `size` is only used in the first condition branch below (and only once), so just write `nroot := root[:len(root)-1]`. In fact, `nroot` and `root` are the same and `root` is discarded at the end of the iteration (more later), so the condition below becomes:

```
if len(root) > 1 {
    root = append(root[:len(root)-1], spath...)
    pot = append(pot, yenShortest{root, weight + rootWeight})
} else {
    pot = append(pot, yenShortest{spath, weight})
}
```

graph/path/yen_ksp.go
```diff
@@ -0,0 +1,156 @@
...
+			}
+
+			yk.visited = make(map[int64]set.Int64s)
+		}
+
+		if len(pot) == 0 {
+			break
+		}
+
+		sort.Sort(byPathWeight(pot))
```
kortschak: Please remove the blank lines in this section. They constitute a stanza.

graph/path/yen_ksp.go
```diff
@@ -0,0 +1,156 @@
...
+		paths = append(paths, pot[0].p)
+
+		pot = pot[1:]
+	}
+
+	return paths
+}
+
+type yenShortest struct {
+	p      []graph.Node
```
kortschak: s/p/path/g

graph/path/yen_ksp.go
```diff
@@ -0,0 +1,156 @@
...
+				nroot := root[:size]
+				nroot = append(nroot, spath...)
+				potential := yenShortest{nroot, weight + rootWeight}
+				pot = append(pot, potential)
+			} else {
+				potential := yenShortest{spath, weight}
+				pot = append(pot, potential)
+			}
+
+			yk.visited = make(map[int64]set.Int64s)
```
kortschak: I don't believe this is correct; the algorithm removes edges from G over the life of the algorithm, while you are removing nodes over the life of a single iteration. This is completely difference and, without rigorous analysis to back this up, likely to result in re-walking paths in adequately small graphs. At least one of the tests I suggested should fail because of this.

apache8080: After reading through the wikipedia page, the way I understand it is that the algorithm is removing edges over the life of each spur path the algorithm is checking so after it has finished iterating through each spur path it resets the graph. [Wikipedia Link](https://en.wikipedia.org/wiki/Yen%27s_algorithm)

graph/path/yen_ksp.go
apache8080: What do you mean by this?

kortschak: Please replace the `p` field label with `path`. The name exists in enough locations at a distance from its definition, that a richer name is worthwhile.

graph/path/yen_ksp.go
kortschak: Sorry, you're right; I missed the section below under "// Add back the edges and nodes that were removed from the graph."

kortschak: However, you need a set of edges, not a set of nodes.

apache8080: I am technically storing edges because I only use the visited map in the From function which is looking for edges. Also, the map key is the From node and all the nodes it has already visited in the past.

kortschak: Again, sorry. You are right - I missed it because it's not a common pattern here.

I will ask that you use a `map[[2]int64]struct{}` or `map[struct{ f, t int64 ]struct{}` though since that requires a single look-up rather than the two here. Note that the edges stored change depending on whether the input `g` is directed or undirected (both directions need to be stored in the undirected case).

graph/path/yen_ksp.go
apache8080: Do you have any good examples of this? I looked online and experimented a little bit but was running into weird issues.

kortschak: What did you see? You should be able to just write `type yenKSPAdjuster struct { graph.Graph; ... `.

graph/path/yen_ksp.go
```diff
@@ -0,0 +1,148 @@
...
+
+	shortest, _ := DijkstraFrom(s, yk).To(t.ID())
+	if len(shortest) == 0 {
+		return nil
+
+	}
+	paths := [][]graph.Node{shortest}
+
+	var pot []yenShortest
+
```
kortschak: Please delete this blank line.

graph/path/yen_ksp.go
```diff
@@ -0,0 +1,148 @@
...
+						}
+					}
+					if ok {
+						yk.addVisited(path[n].ID(), path[n+1].ID())
+					}
+				}
+			}
+
+			spath, weight := DijkstraFrom(spur, yk).To(t.ID())
+
```
kortschak: Please delete this blank line.

graph/path/yen_ksp.go
```diff
@@ -0,0 +1,148 @@
...
+func (s byPathWeight) Swap(i, j int) {
+	s[i], s[j] = s[j], s[i]
+}
+
+func (s byPathWeight) Less(i, j int) bool {
+	return s[i].weight < s[j].weight
+}
+
+type yenKSPAdjuster struct {
```
kortschak: Given that this is an important part of the implementation, please document the function of `yenKSPAdjuster` and its fields.

graph/path/yen_ksp.go
```diff
@@ -0,0 +1,148 @@
...
+	g       graph.Graph
+	visited map[int64]set.Int64s
```
kortschak: Please use a named `map[[2]int64]struct{}` or `map[struct{ f, t int64}]struct{}` here.

graph/path/yen_ksp.go
```diff
@@ -0,0 +1,148 @@
...
+		paths = append(paths, pot[0].path)
+		pot = pot[1:]
+	}
+
+	return paths
+}
+
+type yenShortest struct {
+	path      []graph.Node
+	weight float64
```
kortschak: Please gofmt your code (can I recommend that you add an on-save hook to goimports or gofmt - then you don't need to think about doing this)

graph/path/yen_ksp.go
```diff
@@ -0,0 +1,148 @@
...
+			nodes[int64(i)] = nodes[len(nodes)-1]
+			nodes = nodes[:len(nodes)-1]
+			i--
+		}
+	}
+
+	return nodes
+}
+
+func (g yenKSPAdjuster) addVisited(parent, id int64) {
```
kortschak:
```
func (g yenKSPAdjuster) addVisited(u, v int64) {
	g.visited[[2]int64{u, v}] = struct{}{}
	if !g.isDirected {
		g.visited[[2]int64{v, u}] = struct{}{}
	}
}
```

or the `struct` equivalent. This requires that you add an `isDirected bool` to `yenKSPAdjuster`.

graph/path/yen_ksp_test.go
```diff
@@ -0,0 +1,113 @@
...
+
+func TestYenKSP(t *testing.T) {
+	for _, test := range yenShortestPathTests {
+		g := test.graph()
+		for _, e := range test.edges {
+			g.SetWeightedEdge(e)
+		}
+
+		got := YenKShortestPath(g.(graph.Graph), test.k, test.query.From(), test.query.To())
+		gotIds := toIntPath(got)
```
kortschak: `gotIDs`

graph/path/yen_ksp_test.go
```diff
@@ -0,0 +1,113 @@
...
+}
+
+func TestYenKSP(t *testing.T) {
+	for _, test := range yenShortestPathTests {
+		g := test.graph()
+		for _, e := range test.edges {
+			g.SetWeightedEdge(e)
+		}
+
+		got := YenKShortestPath(g.(graph.Graph), test.k, test.query.From(), test.query.To())
```
kortschak: The type assertion is not needed here.

graph/path/yen_ksp_test.go
```diff
@@ -0,0 +1,113 @@
...
+		},
+		query: simple.Edge{F: simple.Node(0), T: simple.Node(1)},
+		k:     1,
+		wantPaths: [][]int64{
+			{0, 1},
+		},
+	},
+}
+
+func toIntPath(nodePaths [][]graph.Node) [][]int64 {
```
kortschak:
```
func pathIDs(paths [][]graph.Node) [][]int64 {
	ids := make([][]int64, len(paths)
	for i, p := range paths {
		ids[i] = make([]int64, len(p))
		for j, n := range p {
			ids[i][j] = n.ID()
		}
	}
	return ids
}
```

graph/path/yen_ksp.go
```diff
@@ -0,0 +1,148 @@
...
+}
+
+type yenShortest struct {
+	path      []graph.Node
+	weight float64
+}
+
+type byPathWeight []yenShortest
+
+func (s byPathWeight) Len() int {
```
kortschak: These are easier to read as a unit if they are formatted as a stanza.
```
func (s byPathWeight) Len() int           { len(s) }
func (s byPathWeight) Swap(i, j int)      { s[i], s[j] = s[j], s[i] }
func (s byPathWeight) Less(i, j int) bool { return s[i].weight < s[j].weight }
```

graph/path/yen_ksp_test.go
```diff
@@ -0,0 +1,113 @@
...
+
+var yenShortestPathTests = []struct {
+	name  string
+	graph func() graph.WeightedEdgeAdder
+	edges []simple.WeightedEdge
+
+	query     simple.Edge
+	k         int
+	wantPaths [][]int64
+}{
```
kortschak: I'd like to see tests that have k > |paths|, tests on a G={V={0,1}, E={(0,1)}}, on a G={V={0}, E=∅}, on a G=star_n, and on a G={V={-1,1,2..n}, E={(-1,m),(m,1)}} for m∈[2,n] and s=-1, t=1. for k<|{m}|, k = |{m}| and k>|{m}|.

graph/path/yen_ksp_test.go
```diff
@@ -0,0 +1,113 @@
...
+	// Positive weighted graphs.
+	// https://en.wikipedia.org/wiki/Yen%27s_algorithm
```
kortschak: This link should be on the element for the wikipedia example, and be `https://en.wikipedia.org/w/index.php?title=Yen%27s_algorithm&oldid=841018784#Example`.

graph/path/yen_ksp_test.go
```diff
@@ -0,0 +1,113 @@
...
+	{
+		name:  "wikipedia example",
+		graph: func() graph.WeightedEdgeAdder { return simple.NewWeightedDirectedGraph(0, math.Inf(1)) },
+		edges: []simple.WeightedEdge{
+			{F: simple.Node(0), T: simple.Node(1), W: 3},
```
kortschak: Please make the edge and node definitions in the form e.g. `{F: simple.Node('C'), T: simple.Node('D'), W: 3},`,  `Query: simple.Edge{F: simple.Node('C'), T: simple.Node('H')},` and the wanted values the same, e.g. `{'C', 'E', 'F', 'H'},`.

kortschak:
Can we get on to closing this out soon please. It is blocking some major changes that I am planning.

apache8080:
Yeah, I have been busy with work. Will try to get as much done by Friday.

kortschak:
OK. In this PR add yourself to A+C. I want to start work on some breaking changes to graph on the weekend, so if you haven't finished I'll finish this, but I'll need your A+C change to do that.

graph/path/yen_ksp.go
apache8080: Is there a function in graph that checks if a graph is Directed or Undirected?

apache8080:
When I change the yenKSPAdjuster to embed graph.Graph I am getting a segfault. Is there an easy way to use gdb to debug segfaults in golang?

apache8080:
So I have fixed the formatting changes and I have also changed the implementation of the visited map. I wasn't able to get the embedded graph.Graph  working and I haven't written the final test case you have asked for. I will continue to work and see if I can get it done by Friday.

graph/path/yen_ksp_test.go
kortschak: Sorry, I forgot that was an interface type previously.

graph/path/yen_ksp_test.go
apache8080: I have a star n graph test cases, but I don't quite get the second test case you are asking for.

kortschak: I'd like to abandon this PR in favour of #525.

The bipartite graph tests look at a variety of cases where I was expecting the implementation to break down (the k > |paths| too, which I altered in #525 where you had k = |paths|). The algorithm as stated (?) and implemented attempts to give back k paths, but does this with such force that it gives back k possibly paths, in the case that there are fewer paths than k, some returned paths are ∅ and some do not contain t. This looks like a failure to me. It does not look like there is a test in the algorithm at WP for this case, but it is fairly trivial to add. Those changes have been made in the new PR, along with some additional docs and a minor optimisation for allocations. Note that the bipartite graph tests can be extended to include non-equally weighted s-t paths, which is probably worth doing.

Please feel free to comment at #525, but I also want input from others, since, on the face of it, the algorithm seems to have defects and what I've done feels like papering over them.

apache8080:
When I embed graph.Graph into the the yenKSPAdjuster struct and pass it to the DijkstraFrom function, DijkstraFrom is casting the passed graph to graph.Graph and that seems to be returning nil. When I try to do that same conversion locally it errors but I don't have the step of converting to a traverse.Graph when testing locally.

`--- FAIL: TestYenKSP (0.00s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x28 pc=0x535c82]

goroutine 17 [running]:
testing.tRunner.func1(0xc42d1760f0)
	/usr/local/go/src/testing/testing.go:742 +0x29d
panic(0x563180, 0x67aad0)
	/usr/local/go/src/runtime/panic.go:502 +0x229
gonum.org/v1/gonum/graph/path.(*yenKSPAdjuster).Has(0xc42d152840, 0x43, 0xc42d152840)
	<autogenerated>:1 +0x32
gonum.org/v1/gonum/graph/path.DijkstraFrom(0x5b6060, 0x664878, 0x5b68a0, 0xc42d152840, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
	/home/apache8080/go/src/gonum.org/v1/gonum/graph/path/dijkstra.go:25 +0x123
gonum.org/v1/gonum/graph/path.YenKShortestPath(0x5b6f60, 0xc42005a840, 0x3, 0x5b6060, 0x664878, 0x5b6060, 0x664880, 0x623cc0, 0xf, 0x660a5b)
	/home/apache8080/go/src/gonum.org/v1/gonum/graph/path/yen_ksp.go:28 +0x145
gonum.org/v1/gonum/graph/path.TestYenKSP(0xc42d1760f0)
	/home/apache8080/go/src/gonum.org/v1/gonum/graph/path/yen_ksp_test.go:104 +0x1df
testing.tRunner(0xc42d1760f0, 0x5a0650)
	/usr/local/go/src/testing/testing.go:777 +0xd0
created by testing.(*T).Run
	/usr/local/go/src/testing/testing.go:824 +0x2e0
exit status 2
FAIL	gonum.org/v1/gonum/graph/path	10.303s
`

kortschak:
Sorry, my comments above referenced this PR instead of #565. Please look at the code there, then you can see how the embedding works.

apache8080:
@kortschak Closing this PR in favor of #565.

PR #527, opened 2018-06-11
blas/gonum: cleanup in doc comments, panic strings, and test for Dgemm

### LINKED ISSUE #243: blas/gonum: 6c3eea introduced inconsistency to ?gemv comments
The change to comments for ?gemv in 6c3eea8080f72373ec13c659363559501f3a4cc0 is inconsistent with others, and also makes the netlib generation check fail.

Do we want to revert this comment formatting change or propagate it around the entire blas method set?

/cc @vladimir-ch @btracey

vladimir-ch:
Please take a look.

Closes #243


blas/gonum/level2double.go
```diff
@@ -1343,10 +1345,10 @@ func (Implementation) Dtbsv(ul blas.Uplo, tA blas.Transpose, d blas.Diag, n, k i
 	}
 }
 
-// Dsbmv performs
+// Dsbmv performs the matrix-vector operation
 //  y = alpha * A * x + beta * y
-// where A is an n×n symmetric banded matrix, x and y are vectors, and alpha
-// and beta are scalars.
+// where A is an n×n symmetric band matrix with k super-diagonals, x and y are
```
kortschak: "k super-diagonals" or "k+1 diagonals"?

vladimir-ch: `k` does not include the main diagonal and since the matrix is symmetric, the total bandwidth is `k + 1 + k`. This is consistent with `Dgbmv`. `k+1` diagonals is used in `Dtbmv` which also fits the same scheme. So "k super-diagonals" is correct.

kortschak: Thanks.

PR #535, opened 2018-06-21
 all: add missing copyright headers

vladimir-ch:
I checked only `*.go` files.

Please take a look.


btracey:
Can this be added to an auto-check?

sbinet:
now there's a natural home for this kind of tool :)

PR #536, opened 2018-06-21
wip add LDA algorithms

Eleanor2:
- There are 2 new files added: lda.go and lda_test.go
- lda.go computes the linear discriminate analysis in golang
- lda.go was based on a file that computed LDA in Java: https://github.com/haifengl/smile/blob/master/core/src/main/java/smile/classification/LDA.java
-lda_test.go tests lda.go with sample data


btracey:
Hi @Eleanor2 , thanks for the engagement. Is there a particular kind of feedback you're looking for at this point? I ask because the implementation does not look done (still contains print statements, etc.)

Eleanor2:
Hi this was accidentally being asked to be merged into gonum's master branch instead of our own fork. Yes it is a work in progress so I will close this pr and will open a new one when it's ready.

kortschak:
Hi @Eleanor2, we will not be able to accept a PR for this code as it will come with Apache License 2.0 encumbrance which we feel will unnecessarily complicate the licensing of Gonum. If you would like to see this algorithm implemented in Gonum probably the best way to achieve that would be to file a feature request issue. In the issue a broad overview of the algorithm (and some test cases), without implementation details would help us perform the implementation. I would be happy to take that on if you file that request.

PR #539, opened 2018-07-05
all: conform build tag comment to convention

kortschak:
I've resisted this for a long time, but it seems that the convention in the Go tree is to have a space between the comment token and the `+build` (I think this is odd given that the `//go:` directives are abutted, but the data are there:
```
~/go [(go1.10.3)*]$ grep -Rl '// +build' | wc -l
936
~/go [(go1.10.3)*]$ grep -Rl '//+build' | wc -l
4
```

This change conforms all the build tag comments to this convention.

Done by `sed -i -e 's|//+build|// +build|' $(grep -l '//+build' -R|grep \.go$)`.

Please take a look.

### DIFF
--- a/graph/internal/set/same.go
+++ b/graph/internal/set/same.go
@@ -2,7 +2,7 @@
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
-//+build !appengine
+// +build !appengine
 
 package set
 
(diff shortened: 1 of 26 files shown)


btracey:
I echo @vladimir-ch 's sentiments entirely. I like the missing space as a reminder it's something special, but that we should be following the Go team convention.

kortschak:
See https://groups.google.com/d/topic/golang-dev/bq4xA40CITw/discussion

PR #540, opened 2018-07-10
lapack/gonum, mat: add Dpotrs and use it in Cholesky

vladimir-ch:
Not really a necessary addition but I felt like adding something new to our lapack implementation and this looked easy enough to do within my free time constrains. Apart from that, I do think it is good to move as much of the computational stuff to lapack as possible and let `mat` only do the matrix handling around it.

Please take a look.

### DIFF
--- a/lapack/gonum/dpotrs.go
+++ b/lapack/gonum/dpotrs.go
@@ -1,0 +1,46 @@
+// Copyright ©2018 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package gonum
+
+import (
+	"gonum.org/v1/gonum/blas"
+	"gonum.org/v1/gonum/blas/blas64"
+)
+
+// Dpotrs solves a system of n linear equations A*X = B where A is an n×n
+// symmetric positive definite matrix and B is an n×nrhs matrix. The matrix A is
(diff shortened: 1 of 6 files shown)


lapack/gonum/dpotrs.go
```diff
@@ -0,0 +1,39 @@
...
+// license that can be found in the LICENSE file.
+
+package gonum
+
+import (
+	"gonum.org/v1/gonum/blas"
+	"gonum.org/v1/gonum/blas/blas64"
+)
+
+func (Implementation) Dpotrs(uplo blas.Uplo, n, nrhs int, a []float64, lda int, b []float64, ldb int) {
```
btracey: Function comment?

lapack/lapack64/lapack64.go
```diff
@@ -37,6 +37,10 @@ func Potrf(a blas64.Symmetric) (t blas64.Triangular, ok bool) {
 	return
 }
 
+func Potrs(t blas64.Triangular, b blas64.General) {
```
btracey: Comment

vladimir-ch:
Oh, I knew there was something I wanted to do before creating a PR.

vladimir-ch:
Doc comments added. Please take a look.

vladimir-ch:
Thanks.

PR #543, opened 2018-07-11
Improve parameters when dimension of NelderMead is one. In particular…

### LINKED ISSUE #542: optimize: NelderMead has bad default parameters when dim = 1
NelderMead cites a paper for the default values of the parameters, but the paper explicitly states it's for `dim >= 2`.  In particular, this makes the shrink parameter 0 for `dim = 1`, and so the shrink step does not behave properly.

btracey:
… the shrink parameter is set to 0, which causes the algorithm to fail to converge.

Fixes #542.

### DIFF
--- a/optimize/neldermead.go
+++ b/optimize/neldermead.go
@@ -108,14 +108,23 @@ func (n *NelderMead) Init(loc *Location) (Operation, error) {
 	n.expansion = n.Expansion
 	if n.expansion == 0 {
 		n.expansion = 1 + 2/float64(dim)
+		if dim == 1 {
+			n.expansion = 2
+		}
 	}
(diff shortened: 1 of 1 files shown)


btracey:
It turns out it's actually quite tricky to trigger the case (which is why we haven't seen it). I can work harder if you'd like to make one.

vladimir-ch:
It does not have to be a test that fails without this PR. I thought in general there could be a 1d test case so that these special-case values are covered. You don't have to do it now if there is no easy way to add such test.

btracey:
Done. Thanks, sorry for being obtuse.

vladimir-ch:
Not at all, I didn't explain myself clearly.

Issue #548, opened 2018-07-16
mat: cblas_test.go imports gonum.org/v1/netlib/blas that has no Go code


ChrisHines:
### What are you trying to do?

Run `go mod -sync` on module that depends on package `gonum.org/v1/plot/plotter`.

### What did you do?

**main.go:**
```go
package main

import (
	_ "gonum.org/v1/plot/plotter"
)

func main() {
}
```

**go.mod:**
```
module github.com/ChrisHines/gonum-bug
```

**Run command:**
```
>go mod -v -sync
go: finding gonum.org/v1/plot/plotter latest
go: finding gonum.org/v1/plot latest
go: finding gonum.org/v1/netlib/blas latest
go: finding gonum.org/v1/netlib latest
go: import "github.com/ChrisHines/gonum-bug" ->
        import "gonum.org/v1/plot/plotter" ->
        test ->
        import "gonum.org/v1/gonum/mat" ->
        test ->
        import "gonum.org/v1/netlib/blas": cannot find module providing package gonum.org/v1/netlib/blas
```

### What did you expect to happen?

The `go mod` command to succeed.

### What actually happened?

`go mod` failed.

### What version of Go and Gonum are you using?

go version devel +a241922 Fri Jul 13 00:03:31 2018 +0000 windows/amd64

### Does this issue reproduce with the current master?

Yes.

### Additional information

Searching through the source of Gonum I found what appears to be an erroneous import that is the root cause: https://github.com/gonum/gonum/blob/master/mat/cblas_test.go#L11. It imports `gonum.org/v1/netlib/blas`, which is not a valid Go package, it is a directory with no .go files.

PR #550, opened 2018-07-17
Update go.mod with go mod -sync

ChrisHines:


kortschak:
Is it the case that the go mod tooling is unaware of tags and testing? I don't really want netlib to be marked as a dependency; it's not unless it's chosen to be.

ChrisHines:
I don't know. I'm still trying to get the issue described in https://github.com/gonum/gonum/issues/548 to work. Fixing the import path was only part of the problem.

At this point the go.mod file in the gonum/plot repo is referencing an old version of gonum/gonum with the bad import statement before your fix. My approach to updating the go.mod file is to change the version in go.mod to `latest` and run `go mod -sync`. Maybe that's too aggressive at this stage of the module tooling, but when I do that in the gonum/plot repo then netlib gets added with an `// indirect` dependency. So that led me to the idea that the gonum/gonum/go.mod file needed `go mod -sync` to declare all its dependencies.

I defer to your decision as a maintainer. Feel free to close this PR if it's not valid.

kortschak:
You've peaked my interest and now I'm trying to get the the `go mod` command to work - without success; I just get "go mod: cannot use outside module". I guess I'm holding it wrong.

ChrisHines:
Are you getting that on the gonum/plot repo or this repo? I've found an obsolete import statement in gonum/plot that causes some problems: https://github.com/gonum/plot/blob/master/palette/moreland/luminance_test.go#L13

kortschak:
In gonum/gonum and gonum/plot (using `go version devel +c2ba0ca`).

ChrisHines:
That's the error I get if I try to use go mod inside GOPATH without setting the environment variable GO111MODULE=on.

kortschak:
Yeah, I saw that on golang-nuts just before. Thanks.

kortschak:
In the absence of clear use documentation in the go mod help, I fell back on tried and true unix approaches. I hand edited the go.mod for gonum/plot, removing the gonum/gonum line, and then re-ran go mod -sync. This fixed the problem.

kortschak:
I have sent a PR to gonum/plot that should fix the problem, gonum/plot#458. Please try it out.

ChrisHines:
It works. Thanks.

PR #553, opened 2018-07-18
Change call to Local to call to Global

btracey:


btracey:
My fault, there were no merge conflicts, but there was a skew in the test. I'll submit this when the tests pass to fix the build .

PTAL @vladimir-ch @kortschak

kortschak:
I'm not qualified to review this. I am assuming that the confusion that I feel in looking at what the code does will be alleviated when the names are changed.

btracey:
Yes, it will be clearer. I included you because I was going to merge without an okay if Vladimir couldn't review quickly to fix the build.

PR #556, opened 2018-07-23
lapack/testlapack: clean up and use isOrthogonal helper

vladimir-ch:
Motivated by https://github.com/gonum/lapack/issues/234

At the moment I would not add a tolerance parameter to `isOrthogonal`. `1e-13` is pretty tight and works for all current tests on my machine, so there is no real need for it. It can be added later if some finer control is necessary.

Please take a look.


lapack/testlapack/dgeql2.go
```diff
@@ -71,7 +71,7 @@ func Dgeql2Test(t *testing.T, impl Dgeql2er) {
 			copy(qTmp.Data, q.Data)
 			blas64.Gemm(blas.NoTrans, blas.NoTrans, 1, h, qTmp, 0, q)
 		}
-		if !isOrthonormal(q) {
+		if !isOrthogonal(q) {
```
sbinet: perhaps we should update the error message to say `orthogonal` instead of `orthonormal` ?

vladimir-ch: Thanks, I missed that. I will change the error message.

vladimir-ch: Doing this I noticed that `isOrthogonal` can be also used in `Dsytd2Test`. It has also brought to my attention the existence of `hasOrthonormalColumns`. All in all, I smuggled in some additional changes in the last 4 commits.

Please take a look.

lapack/testlapack/dorg2l.go
```diff
@@ -5,11 +5,9 @@
 package testlapack
 
 import (
-	"math"
 	"testing"
 
 	"golang.org/x/exp/rand"
-
```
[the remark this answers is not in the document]
vladimir-ch: Done.

Issue #551, opened 2018-07-17
mat: Better panic message when asking to slice to zero


btracey:
The following code

```
package main

import "gonum.org/v1/gonum/mat"

func main() {
	a := mat.NewDense(5, 3, nil)
	a = a.Slice(0, 0, 0, 3).(*mat.Dense)
}
```

Gives the panic of `panic: matrix: index out of range`. A better panic would be about a zero-sized slice.

ahlusar1989:
@btracey Is this where you would like more informed messages: https://github.com/gonum/gonum/blob/master/mat/dense.go#L49

kortschak:
No, it would be a check for i relative to k and j relative to l [before this](https://github.com/gonum/gonum/blob/72e79ca98526cae0807f57ef69033b767f715ea7/mat/dense.go#L285) (with a change there of s/k <= i/k < i/ and s/l <= j/l < j/.

```
	if i == k || j == l {
		panic(ErrZeroLength)
	}
```

ahlusar1989:
Gotcha. @btracey Would you like for me to take a stab at it, or do you feel that you already have a solution in mind. I am looking to learn and grow my skillset, so it will be a learning experience for me, either way :)

btracey:
It would be great if you would want to look into this @ahlusar1989

ahlusar1989:
@kortschak I am working on the solution for this. Do you mind clarifying for me this:

```

 s/k <= i/k < i/ and s/l <= j/l < j/

```

kortschak:
The substitutions above are to change `k <= i` to `k < i` and the same for `l` and `j`. These are since the equality is already checked. The alternative that @btracey has suggested is to leave those the same and move the `ErrZeroLength` panic check inside the conditional below like so:
```
	mr, mc := m.Caps()
	if i < 0 || mr <= i || j < 0 || mc <= j || k < i || mr < k || l < j || mc < l {
		if i == k || j == l {
			panic(ErrZeroLength)
		}
		panic(ErrIndexOutOfRange)
	}
```

This has a minor reduction in comparisons for the fast path, so probably do that.

kortschak:
This was fixed by #558.

PR #558, opened 2018-07-27
mat: add error handling when asking to slice to 0

ahlusar1989:
Addresses #551 .

WIP - needs tests and clarifications on any other edge cases I have missed from @kortschak. So far I have only one condition to check if i is relative to k and j is relative to l. 

All tests pass and code is formatted.


mat/dense.go
```diff
@@ -281,6 +281,10 @@ func (m *Dense) rawRowView(i int) []float64 {
 // Slice panics with ErrIndexOutOfRange if the slice is outside the capacity
 // of the receiver.
 func (m *Dense) Slice(i, k, j, l int) Matrix {
+	if i == k || j == l {
+		panic(ErrZeroLength)
+	}
+
 	mr, mc := m.Caps()
 	if i < 0 || mr <= i || j < 0 || mc <= j || k <= i || mr < k || l <= j || mc < l {
```
kortschak: The condition relating i/k and j/l can be changed to `k < i` and `l < j` since the equality case is handled above.

I'm wondering now about the order of checks though. @btracey What do you think about this? Is an out of range check a higher priority than a zero length? With the change in the para above, that would work.

CONTRIBUTORS
```diff
@@ -71,3 +71,4 @@ Takeshi Yoneda <cz.rk.t0415y.g@gmail.com>
 Tobin Harding <me@tobin.cc>
 Vladimír Chalupecký <vladimir.chalupecky@gmail.com>
 Yevgeniy Vahlis <evahlis@gmail.com>
+Saran Ahluwalia <ahlusar.ahluwalia@gmail.com>
```
kortschak: Please add yourself to AUTHORS as well.

btracey:
I think it probably doesn't matter which we do first. What we could do, however, is keep the check as is, and then inside the if statement test for equality to throw the right panic. This way there is only one test for correct slice actions.

ahlusar1989:
I'll do just that @btracey . I think @kortschak has an example in the issue #551 . I appreciate all of your feedback.

AUTHORS
```diff
@@ -68,3 +68,5 @@ The University of Washington
 Tobin Harding <me@tobin.cc>
 Vladimír Chalupecký <vladimir.chalupecky@gmail.com>
 Yevgeniy Vahlis <evahlis@gmail.com>
+Saran Ahluwalia <ahlusar.ahluwalia@gmail.com>
```
btracey: Sorry @ahlusar1989 , please add yourself alphabetically in the list (after Sam Zaydel)

CONTRIBUTORS
btracey: Alphabetically, as above.

kortschak:
Can you please add back the final newline.

CONTRIBUTORS
```diff
@@ -70,4 +71,4 @@ Steve McCoy <mccoyst@gmail.com>
 Takeshi Yoneda <cz.rk.t0415y.g@gmail.com>
 Tobin Harding <me@tobin.cc>
 Vladimír Chalupecký <vladimir.chalupecky@gmail.com>
-Yevgeniy Vahlis <evahlis@gmail.com>
+Yevgeniy Vahlis <evahlis@gmail.com>
```
kortschak: Here too please.

PR #560, opened 2018-08-01
floats: use asm-backed AddConst

Kunde21:
Replace native go loop with its asm function.


Kunde21:
Closed in favor of #561

PR #562, opened 2018-08-01
floats,asm/f64: add asm routine for sum

Kunde21:
Add Sum(x []float64) float64 which computes the sum of all elements in the slice.

### DIFF
--- a/internal/asm/f64/sum_amd64.s
+++ b/internal/asm/f64/sum_amd64.s
@@ -1,0 +1,100 @@
+// Copyright ©2016 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// +build !noasm,!appengine,!safe
+
+#include "textflag.h"
+
+#define X_PTR SI
+#define IDX AX
+#define LEN CX
+#define TAIL BX
+#define SUM X0
+#define SUM_1 X1
+#define SUM_2 X2
+#define SUM_3 X3
+
+// func Sum(x []float64) float64
+TEXT ·Sum(SB), NOSPLIT, $0
+	MOVQ x_base+0(FP), X_PTR // X_PTR = &x
+	MOVQ x_len+8(FP), LEN    // LEN = len(x)
+	XORQ IDX, IDX            // i = 0
+	PXOR SUM, SUM            // p_sum_i = 0
+	CMPQ LEN, $0             // if LEN == 0 { return 0 }
+	JE   sum_end
+
+	PXOR SUM_1, SUM_1
+	PXOR SUM_2, SUM_2
+	PXOR SUM_3, SUM_3
+
+	MOVQ X_PTR, TAIL // Check memory alignment
+	ANDQ $15, TAIL   // TAIL = &y % 16
+	JZ   no_trim     // if TAIL == 0 { goto no_trim }
+
+	// Align on 16-byte boundary
+	ADDSD (X_PTR), X0 // X0 += x[0]
+	INCQ  IDX         // i++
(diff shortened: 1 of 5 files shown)


internal/asm/f64/sum_amd64.s
```diff
@@ -0,0 +1,100 @@
...
+	DECQ  TAIL        // TAIL--
+	JZ    sum_end     // if TAIL == 0 { return }
+
+no_trim:
+	MOVQ LEN, TAIL
+	SHRQ $4, LEN        // LEN = floor( n / 16 )
+	JZ   sum_tail_start // if LEN == 0 { goto tail_start }
+
+sum_loop: // sum 16x wide do {
+	ADDPD (SI)(AX*8), SUM      // sum_i += x[i:i+1]
```
vladimir-ch: The comment is confusing because in Go `x[i:i+1]` is a 1-element slice. I think that `sum += x[i] + x[i+1]` would be better although this also isn't completely accurate. Also below.

internal/asm/f64/sum_amd64.s
```diff
@@ -0,0 +1,100 @@
...
+	ADDPD 48(SI)(AX*8), SUM_3
+	ADDPD 64(SI)(AX*8), SUM
+	ADDPD 80(SI)(AX*8), SUM_1
+	ADDPD 96(SI)(AX*8), SUM_2
+	ADDPD 112(SI)(AX*8), SUM_3
+	ADDQ  $16, IDX             // i += 16
+	DECQ  LEN
+	JNZ   sum_loop             // } while --CX > 0
+
+sum_tail_start: // Reset loop registers
```
vladimir-ch: "Reset loop registers", what does it refer to?
Also, rename the label to sum_tail8?

internal/asm/f64/sum_amd64.s
```diff
@@ -0,0 +1,100 @@
+// Copyright ©2016 The Gonum Authors. All rights reserved.
```
vladimir-ch: Is there any reason for 2016?

kortschak:
No worries. Just checking.

internal/asm/f64/sum_amd64.s
Kunde21: I try to use the common vector programming syntax in the comments, because it's difficult to easily display the two slots in an XMM register (or 4 in YMM).  Using `x[i] + x[i+1]` breaks down when we arrive at the `HADD` instructions, because the comments haven't been showing the two elements.

Kunde21:
Sorry for the delays on this one.  I addressed @vladimir-ch comments.  Let me know if you want me to sqash/rebase.

Kunde21:
Travis failure was a TLS hiccup in one of the 1.8 builds.  Not a code error.

kortschak:
What failure :slightly_smiling_face: ?

internal/asm/f64/sum_amd64.s
kortschak: Maybe `sum_i += sum(x[i:i+2])`?

Kunde21: `sum(x[i:i+2])` The two elements aren't being added to each other in this op.  Each is being added to a different element of the vector `sum_i`.

kortschak: OK, sorry. Then without the `sum()`, but still using the `[i:i+2]` which I think is the concern that @vladimir-ch had here; the use of `:` in an indexing context look very much like Go, where it doesn't mean what you want here.

Kunde21:
@kortschak Yes.  Thank you.

kortschak:
Thank you for doing this.

PR #565, opened 2018-08-03
graph/path: add Yen k-shortest paths implementation

kortschak:
Please take a look.

/cc @apache8080


apache8080:
The change seems to make sense. Thanks for all the help with this PR.

kortschak:
>It might make sense to update the Wikipedia page with this.

I'd like to see what other people here think first. I may have missed something in the description in WP. Looking through the original paper, I can see why there could possibly be some lack of clarity.

vladimir-ch:
I will take a look soon.

graph/path/yen_ksp.go
```diff
@@ -0,0 +1,152 @@
...
+	weight Weighting
+
+	// visitedEdges holds the edges that have
+	// been removed by Yen's algorithm.
+	visitedEdges map[[2]int64]struct{}
+}
+
+func (g yenKSPAdjuster) From(id int64) []graph.Node {
+	nodes := g.Graph.From(id)
+
```
vladimir-ch: Remove the empty lines here as well?

graph/path/yen_ksp.go
```diff
@@ -0,0 +1,152 @@
...
+	for i := 0; i < len(nodes); i++ {
+		if !g.canWalk(id, nodes[i].ID()) {
+			continue
+		}
+		nodes[int64(i)] = nodes[len(nodes)-1]
```
vladimir-ch: Why the cast to `int64`?

graph/path/yen_ksp.go
```diff
@@ -0,0 +1,152 @@
...
+	// visitedEdges holds the edges that have
+	// been removed by Yen's algorithm.
+	visitedEdges map[[2]int64]struct{}
+}
+
+func (g yenKSPAdjuster) From(id int64) []graph.Node {
+	nodes := g.Graph.From(id)
+
+	for i := 0; i < len(nodes); i++ {
+		if !g.canWalk(id, nodes[i].ID()) {
```
vladimir-ch: This is confusing, currently it reads as: if cannot walk from `id` to `nodes[i]`, keep `nodes[i]` as reachable from `id`. `canWalk` should return `!ok` and then this can change to `if g.canWalk ... continue`.

graph/path/yen_ksp.go
```diff
@@ -0,0 +1,152 @@
...
+			continue
+		}
+		nodes[int64(i)] = nodes[len(nodes)-1]
+		nodes = nodes[:len(nodes)-1]
+		i--
```
vladimir-ch: This loop index manipulation made me pause for a while. For me it would be clearer to omit the post statement, increment `i` manually before the `continue` above and remove this line.

graph/path/yen_ksp.go
```diff
@@ -0,0 +1,152 @@
...
+		yk.weight = UniformCost(g)
+	}
+
+	shortest, _ := DijkstraFrom(s, yk).To(t.ID())
+	switch len(shortest) {
+	case 0:
+		return nil
+	case 1:
+		return [][]graph.Node{shortest}
+
```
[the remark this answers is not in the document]
kortschak: Done.

graph/path/yen_ksp.go
```diff
@@ -0,0 +1,152 @@
...
+	var root []graph.Node
+	for i := int64(1); i < int64(k); i++ {
+		for n := 0; n < len(paths[i-1])-1; n++ {
+			yk.reset()
+
+			spur := paths[i-1][n]
+			root := append(root[:0], paths[i-1][:n+1]...)
+
+			for _, path := range paths {
+				if len(path) > n {
```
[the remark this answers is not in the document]
kortschak: Done.

graph/path/yen_ksp.go
kortschak: Done.

graph/path/yen_ksp.go
kortschak: Removed.

graph/path/yen_ksp.go
kortschak: Sorry. Fixed.

graph/path/yen_ksp.go
kortschak: I missed that. This is not how I would have written it. Thanks.

Fixed.

kortschak:
There is logic to handle the deficiencies. The logic does not appear to be in the WP pseudo-code, but the WP pseudo-code does not bear more than a passing resemblance to the original algorithm anyway - the original is very hard to read and does not make use of dijkstra.

PR #567, opened 2018-08-08
graph: add node, edge and line iterators

kortschak:
I have largely left the changes in the client code as mechanical, using the graph package's helper, except where there's an obvious simplification or improvement in performance. The changes here should have reasonably limited negative impact on performance and will allow improvments in performance in some cases (when reflect gets map iterator support we'll see a benefit, and I'll write an unsafe equivalent of that API).

Please take a look.

This should go in after #565 since I'll update this to add the necessary changes in that code, so please review that first.

Fixes gonum/graph#18.
Fixes gonum/graph#46.
Updates #570.


kortschak:
I'm not sure what to do about the [`multi.Edge`](https://github.com/gonum/gonum/blob/9837c06d8caf08c2f0a57732da870dac50ded3ec/graph/multi/multi.go#L21) and [`multi.WeightedEdge`](https://github.com/gonum/gonum/blob/9837c06d8caf08c2f0a57732da870dac50ded3ec/graph/multi/multi.go#L56-L64) types. These still hold a slice of the relevant line type. The can be converted to an iterator-based type, but they need to reset before and after iterating once to get their representative ends (e.g. for the [`Edge.From` method](https://github.com/gonum/gonum/blob/9837c06d8caf08c2f0a57732da870dac50ded3ec/graph/multi/multi.go#L24-L29)). I think this is probably OK, though not ideal.

kortschak:
Is someone able to take a look at this? It's relatively straight forward (though largish) and would be suitable for someone who does not necessarily have a strong background in graph algorithms/implementations.

vladimir-ch:
If nobody is faster, I'll be able to take a look after Sept 10.

kortschak:
Thanks

graph/graph_test.go
```diff
@@ -272,9 +272,9 @@ func same(a, b graph.Graph) bool {
 			return false
 		}
 	}
-	for _, u := range a.Nodes() {
-		aFromU := a.From(u.ID())
-		bFromU := b.From(u.ID())
+	for _, u := range graph.NodesOf(a.Nodes()) {
```
mewmew: Why use range and not iterator here?

graph/iterator/edges.go
```diff
@@ -0,0 +1,129 @@
...
+package iterator
+
+import "gonum.org/v1/gonum/graph"
+
+// OrderedEdges implements the graph.Edges and graph.EdgeSlicer interfaces.
+// The iteration order of OrderedEdges is the order of edges passed to
+// NewEdgeIterator.
+type OrderedEdges struct {
+	idx   int
+	edges []graph.Edge
```
mewmew: There is currently no way to append new edges to `OrderedEdges`, and that's probably fine.

graph/multi/directed.go
```diff
@@ -247,26 +248,11 @@ func (g *DirectedGraph) Lines(uid, vid int64) []graph.Line {
...
+	return iterator.NewOrderedLines(lines)
 }
 
 // HasEdgeFromTo returns whether an edge exists in the graph from u to v.
 func (g *DirectedGraph) HasEdgeFromTo(uid, vid int64) bool {
 	_, ok := g.from[uid][vid]
 	return ok
 }
-
-// Degree returns the in+out degree of n in g.
```
mewmew: Was this removed as the functionality is now provided by iterators? If so, how.

graph/nodes_edges.go
```diff
@@ -0,0 +1,203 @@
+// Copyright ©2018 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package graph
+
+type Iterator interface {
```
mewmew: Missing doc comment for `Iterator`.

graph/nodes_edges.go
```diff
@@ -0,0 +1,203 @@
...
+	// NodeSlice returns the set of nodes remaining
+	// to be iterated by a Nodes iterator.
+	// The holder of the iterator may arbitrarily
+	// change elements in the returned slice, but
+	// those changes may be reflected to other
+	// iterators.
+	NodeSlice() []Node
+}
+
+// NodesOf returns it.Len() nodes from it. If it is an NodeSlicer, the NodeSlice method
```
mewmew:
```diff
-an NodeSlicer
+a NodeSlicer
```

graph/nodes_edges.go
```diff
@@ -0,0 +1,203 @@
...
+
+// Lines is an Line iterator.
+type Lines interface {
+	Iterator
+
+	// Line returns the current Line from the iterator.
+	Line() Line
+}
+
+type LineSlicer interface {
```
mewmew: Missing doc comment for `LineSlicer`.

graph/nodes_edges.go
```diff
@@ -0,0 +1,203 @@
...
+
+// Edges is an Edge iterator.
+type Edges interface {
+	Iterator
+
+	// Edge returns the current Edge from the iterator.
+	Edge() Edge
+}
+
+type EdgeSlicer interface {
```
mewmew: Missing doc comment for `EdgeSlicer`.

graph/nodes_edges.go
```diff
@@ -0,0 +1,203 @@
...
+
+// Nodes is a Node iterator.
+type Nodes interface {
+	Iterator
+
+	// Node returns the current Node from the iterator.
+	Node() Node
+}
+
+type NodeSlicer interface {
```
mewmew: Missing doc comment for NodeSlicer.

graph/nodes_edges.go
```diff
@@ -0,0 +1,203 @@
...
+
+// WeightedLines is an Line iterator.
+type WeightedLines interface {
+	Iterator
+
+	// Line returns the current Line from the iterator.
+	WeightedLine() WeightedLine
+}
+
+type WeightedLineSlicer interface {
```
mewmew: Missing doc comment for WeightedLineSlicer, etc (seems to be all FooSlicer's).

graph/simple/simple.go
```diff
@@ -18,6 +18,10 @@ func (n Node) ID() int64 {
 	return int64(n)
 }
 
+func newSimpleNode(id int) graph.Node {
```
mewmew: Why use this instead of simple calling `Node(id)`?

kortschak:
I'll get back to the specific point later, but to address the major issues.

There is necessarily going to be some additional syntax when we move away from types that can be arguments to a `range` operator. This is demonstrated by the use of the fully specified iterator use in the `graph.Copy*` functions. However these will have performance benefits in future (they do not currently unless the `src` is a dense graph without stored nodes. As you point out, the idiom of the `bufio,Scanner` (and others in the broader golang.org-provided libraries) makes this less foreign than might otherwise be. In addition to this, the iterators approach here gives us cheap to ~free access to data that we did not previously have and harmonises approaches to getting those data (graph order and size, node degree — in this case allowing the user to get in, out and in-out degree without us adding arbitrarily many degree methods).

The memory impact is, at the moment, approximately zero (determined by Mental Static Analysis™). This is based on the observation that all graphs currently have to allocate a `[]graph.Node`, `[]graph.Edge` or `[]graph.Line` for iterations. This is still done for the most part, and that allocation is passed unaltered through the respective iterator via the `NodeSlice` call when it can be. Some (two) graph implementations defer this allocation or omit it depending on what iterator (ours or `range`) is used, so these may reduce the allocation load. After go1.12 (it is in master), there will exist a [`reflect.MapIter`](https://tip.golang.org/pkg/reflect/#MapIter) type that is a low-allocation map iterator that works on an already existing map. I intend to replace many of the `iterator.OrderedNodes` returns with a, yet-to-be-written `iterator.Nodes` type that wraps this or an unsafe equivalent (prior to the inclusion of `reflect.MapIter` I had already written a similar unsafe iterator, but could not use it in Gonum code because we have to support the `safe` build tag. These changes will further reduce allocations across a wider range of graph implementations.

graph/multi/directed.go
kortschak: `id := n.ID(); g.From(id).Len()+g.To(id).Len()` gives us this. This is currently more costly, but will be less so later (not quite down to what it is right now, but I think worth it because of the readability; there are many ways to interpret degree, and a method name of `Degree` does not say which, the code above does.

graph/iterator/edges.go
kortschak: This is intentional.

graph/graph_test.go
kortschak: In a test this is simpler and the cost is not an issue.

mewmew:
> There is necessarily going to be some additional syntax when we move away from types that can be arguments to a range operator.

Shame to see the use of `range` go. Oh well. Very unlikely, but that could also be a *Go 2 proposal*. Provide access to iterator like behaviour by implementing an `Iterator` interface defined by the Go std to get syntactic sugar for `range` iterators.

> In addition to this, the iterators approach here gives us cheap to ~free access to data that we did not previously have and harmonises approaches to getting those data (graph order and size, node degree — in this case allowing the user to get in, out and in-out degree without us adding arbitrarily many degree methods).

I would agree, this is good. Also clarifies the removal of the `Degree` methods in this PR.

> The memory impact is, at the moment, approximately zero (determined by Mental Static Analysis™). This is based on the observation that all graphs currently have to allocate a []graph.Node, []graph.Edge or []graph.Line for iterations. This is still done for the most part, and that allocation is passed unaltered through the respective iterator via the NodeSlice call when it can be. Some (two) graph implementations defer this allocation or omit it depending on what iterator (ours or range) is used, so these may reduce the allocation load. After go1.12 (it is in master), there will exist a reflect.MapIter type that is a low-allocation map iterator that works on an already existing map. I intend to replace many of the iterator.OrderedNodes returns with a, yet-to-be-written iterator.Nodes type that wraps this or an unsafe equivalent (prior to the inclusion of reflect.MapIter I had already written a similar unsafe iterator, but could not use it in Gonum code because we have to support the safe build tag. These changes will further reduce allocations across a wider range of graph implementations.

That's really great. Thanks for the write-up. Also, I had not heard about reflect.MapIter yet, so thanks for the link. The API used there is mostly identical to the Iterator interface you defined here. Not a coincidence I would guess ^^

graph/iterator/edges.go
kortschak: Though it may be worth adding a `Swap(i, j int)` method here (and to the other ordered iterators) that bases `i` and `j` on `idx`. This would allow arbitrary user orderings if the user embeds the value in a struct with a `Less`method.

graph/nodes_edges.go
kortschak: Fixed.

graph/nodes_edges.go
kortschak: Fixed.

graph/nodes_edges.go
kortschak: Fixed.

graph/nodes_edges.go
kortschak: Indeed. <kbd>CTRL-C</kbd><kbd>CTRL-V</kbd>.

Fixed.

graph/simple/simple.go
kortschak: Because the `iterator.NewImplicitNodes` function takes a `func(int) graph.Node` and `graph.Node()` is not a function.

graph/nodes_edges.go
kortschak: Fixed.

graph/nodes_edges.go
```diff
@@ -0,0 +1,209 @@
...
+	// The holder of the iterator may arbitrarily
+	// change elements in the returned slice, but
+	// those changes may be reflected to other
+	// iterators.
+	NodeSlice() []Node
+}
+
+// NodesOf returns it.Len() nodes from it. If it is a NodeSlicer, the NodeSlice method
+// is used to obtain the nodes. It is safe to pass a nil Nodes to NodesOf.
+func NodesOf(it Nodes) []Node {
```
kortschak: Maybe this should be `func NodesOf(dst []Node, it Nodes) []Node` and append to `dst` and return it so that we can avoid allocations. Same for `EdgesOf` etc.

graph/simple/simple.go
mewmew: Oh, I see. Thanks for clarifying.

graph/nodes_edges.go
mewmew: That would make sense, and then passing nil for dst allocates a new slice. When if cap(dst) is less than the number of nodes in `it`?

kortschak: The behaviour would be `l := len(dst); dst = append(dst, make([]Node, it.Len())...)[:l]` followed by iterative addition of nodes for the non-slicer and just `dst = append(dst, it.NodeSlice()...)` for the slicer case.

This reduces the number of reallocations in the iterative case. It also gives a way to collect nodes from multiple `Nodes` values.

graph/nodes_edges.go
```diff
@@ -0,0 +1,209 @@
+// Copyright ©2018 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package graph
+
+// Iterator is a item iterator.
```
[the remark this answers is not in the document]
kortschak: Done.

graph/nodes_edges.go
```diff
@@ -0,0 +1,209 @@
...
+	// LineSlice returns the set of lines remaining
+	// to be iterated by an Lines iterator.
+	// The holder of the iterator may arbitrarily
+	// change elements in the returned slice, but
+	// those changes may be reflected to other
+	// iterators.
+	LineSlice() []Line
+}
+
+// NodesOf returns it.Len() nodes from it. If it is an LineSlicer, the LineSlice method is used
```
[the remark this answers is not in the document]
kortschak: Done (and corrected name and "an" -> "a" in similar docs).

graph/nodes_edges.go
```diff
@@ -0,0 +1,209 @@
...
+		return it.LineSlice()
+	}
+	n := make([]Line, 0, it.Len())
+	for it.Next() {
+		n = append(n, it.Line())
+	}
+	return n
+}
+
+// WeightedLines is an Line iterator.
```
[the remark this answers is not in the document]
kortschak: Done

kortschak:
Thank you both for the reviews.

PR #571, opened 2018-08-16
blas/blas64: reorder struct fields

vladimir-ch:
Reorder vector and matrix struct fields to match the order in which BLAS
routines accept them. This makes calls with inline declarations with
unnamed fields consistent with the `blas/gonum` package and so easier to understand.

Please take a look.

Taken over from https://github.com/gonum/blas/issues/201

### DIFF
--- a/blas/blas64/blas64.go
+++ b/blas/blas64/blas64.go
@@ -28,72 +28,72 @@ func Implementation() blas.Float64 {
 
 // Vector represents a vector with an associated element increment.
 type Vector struct {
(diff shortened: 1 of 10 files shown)


vladimir-ch:
keyify seems to require a byte offset that points into the struct literal, which makes it not so useful for us, unless I'm missing something.

kortschak:
No, you're correct. I missed that. This is a pity. I'll ask if there is another one that will just work on a file.

PR #586, opened 2018-09-07
Entropy estimators

kzahedi:
Please take a look.

### DIFF
--- a/stat/counts.go
+++ b/stat/counts.go
@@ -1,0 +1,50 @@
+// Copyright ©2018 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package stat
+
+import "gonum.org/v1/gonum/mat"
+
+// CountsToDist returns a probability distribution from a discrete vector of
+// counts.
+//   p(x) = c_x / N
+// where N = \sum c_x. The function returns mat.Vector
+func CountsToDist(c []int) mat.Vector {
+	var n float64
+	r := make([]float64, len(c), len(c))
+
+	for _, v := range c {
+		n += float64(v)
+	}
+
+	for i, v := range c {
+		r[i] = float64(v) / n
+	}
+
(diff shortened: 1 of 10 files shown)


stat/counts.go
```diff
@@ -0,0 +1,50 @@
...
+	for i, v := range c {
+		r[i] = float64(v) / n
+	}
+
+	p := mat.NewVecDense(len(r), r)
+
+	return p
+}
+
+// Counts returns an integer vector with number of occurrences of each number in
```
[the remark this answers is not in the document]
btracey: Hold off though until I can look at the entropy estimators more closely.

stat/it/probabilityestimators.go
```diff
@@ -0,0 +1,156 @@
+package it
```
btracey: We should leave these off this PR since they aren't used here.

stat/it/mutualinformation.go
```diff
@@ -0,0 +1,43 @@
+package it
+
+import (
+	"math"
+
+	"gonum.org/v1/gonum/mat"
+)
+
+// MutualInformation calculates the mutual information with the given lnFunc function
```
btracey: I think we can just move this into `gonum/stat`. It's cleanly defined, and simple.

kortschak:
Where are we up to with this?

Note that the failure here is due to missing copyright headers.

kzahedi:
The main problem is that I don't understand how to have my own fork of gonum and don't mess up the import paths.

kortschak:
When you locally clone out your repo from your fork don't use `go get` from your fork, use git and put it in the correct path for the gonum project,
```
mkdir -p $GOPATH/src/gonum.org/v1
$GOPATH/src/gonum.org/v1
git clone git@github.com:kzahedi/gonum.git
```

Alternatively, if you already have gonum installed in the correct location, add your fork as a tracked remote to that and work that way (this is how I usually work on third-party projects).

vladimir-ch:

PR #587, opened 2018-09-11
quat: new package for quaternions

### LINKED ISSUE #229: Quaternion math
Great lib! But using it I found one part missing which I could not find in code/issues. A fast google gave me this [westphae/quaternion](https://github.com/westphae/quaternion) which I'll use for now. Im using Quaternion as a part to test a simulation of a robot moving. Would be great to also have in this lib.

kortschak:
This is not quite ready; I need to add tests for the higher functions.

The code for higher functions is taken from the April 1 joke for 2017, but on the face of it looks sound. Tests are needed.

Please take a look.

Closes #229.


btracey:
I wonder if it's worth having another level of heirarchy? I have this from a long time ago: https://github.com/btracey/hyperdual  , and `dual` numbers are quite useful as well.

kortschak:
How would you like to fit this together, or are you suggesting quat should not go in?

btracey:
I was mostly suggesting that we do something like `gonum/num/quat` so that way the alternate number implementations don't have to all be top-level packages.

btracey:
I'll review when it's ready, but I agree with the high-level API decisions from a quick glance

kortschak:
How many number types do you think there will be?

btracey:
I'm not sure, but there are already three that I know are useful (quat, dual, hyperdual)

kortschak:
Fair enough. I'll move it.

num/quat/quat.go
```diff
@@ -0,0 +1,105 @@
...
+import (
+	"bytes"
+	"fmt"
+)
+
+var zero Quat
+
+// Quat is a float64 precision quaternion.
+type Quat struct {
+	Real, Imag, Jmag, Kmag float64
```
kortschak: Another possible set of field names are R, I, J and K.

kortschak:
@btracey I have added tests for `Exp`, `Log`, `Pow` (partial) and `Sqrt`. There are trivial tests for the trig functions, but I don't see other implementations that provide those, so I don't have a reference (same for the non-real powers for the `Pow` test). If you can suggest places to look or relationships I can test (other than the obvious of round-tripping inverses) that would be very helpful. Otherwise, please feel free to start reviewing.

num/quat/exp.go
```diff
@@ -0,0 +1,65 @@
...
+		return lift(math.Exp(w))
+	}
+	v := Abs(uv)
+	e := math.Exp(w)
+	s, c := math.Sincos(v)
+	return Add(lift(e*c), Scale(e*s/v, uv))
+}
+
+// Log returns the natural logarithm of q.
+
```
vladimir-ch: Extra line

num/quat/quat.go
```diff
@@ -0,0 +1,400 @@
...
+}
+
+// Format implements fmt.Formatter.
+func (q Quat) Format(fs fmt.State, c rune) {
+	prec, pOk := fs.Precision()
+	if !pOk {
+		prec = -1
+	}
+	width, wOk := fs.Precision()
```
vladimir-ch: I think this should say `fs.Width()`.

num/quat/quat.go
vladimir-ch: During the review I got used to the `mag` version, initially I thought that `Jmag` and `Kmag` carry too much cleverness and play-on-wordiness.

kortschak: Yes, recall that the original code this is based on was an April Fool's joke.

I'd change the suggestion above so that it's Real instead of R.

However, there is an alternative pronunciation that makes it sensible this way. Most people read `imag` is `imag(inary)`, but it could easily be `imag(nitude)`. This second pronunciation would probably be clearer that I, J and K.

num/quat/exp.go
kortschak: Fixed

num/quat/quat.go
kortschak: Indeed. Thanks.

Fixed

btracey:
I personally prefer the less verbose I J K. It more closely matches teh complex implementation, and I think it's clear for anyone who knows quaternions. I've gone through a chunk of the code, and hopfeully will get to the rest soon.

kortschak:
>I personally prefer the less verbose I J K. It more closely matches teh complex implementation

I'm not sure this is true; we don't parse source to get a value the same way that Go does for complex so the equivalence is not really there, but we do return a component with a word which mimics the `imag` and `real` built-ins. The parsing input that we do do works the same way as the fmt and the compiler with the single letter basis symbols. I have thought about this a bit since @vladimir-ch's comment (and my own origin suggestion) and I think that the single letter upper case do not tell the story properly: they are the wrong case, they do not actually return the value that they say (but rather the magnitude of a scalar multiplied by the basis) and they are small enough to get lost in source when used in the most common case of accessing the value (rather than as a struct composite literal key).

btracey:
Okay, sounds reasonable. I'll pay attention when I review the code.

num/quat/conj.go
```diff
@@ -0,0 +1,20 @@
+// Copyright ©2018 The Gonum Authors. All rights reserved.
```
btracey: This file looks good.

num/quat/conj_test.go
```diff
@@ -0,0 +1,41 @@
+// Copyright ©2018 The Gonum Authors. All rights reserved.
```
btracey: This file looks good.

num/quat/doc.go
```diff
@@ -0,0 +1,6 @@
+// Copyright ©2018 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Package quat provides quaternion functions.
```
btracey: `numeric type and functions`?

num/quat/util.go
```diff
@@ -0,0 +1,21 @@
...
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Copyright 2017 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package quat
+
+func lift(v float64) Quat {
```
btracey: I'm sure this is a good name, but why `lift`? I feel like there's some concept I'm missing.

num/quat/exp.go
```diff
@@ -0,0 +1,64 @@
...
+// Exp returns e**q, the base-e exponential of q.
+func Exp(q Quat) Quat {
+	w, uv := split(q)
+	if uv == zero {
+		return lift(math.Exp(w))
+	}
+	v := Abs(uv)
+	e := math.Exp(w)
+	s, c := math.Sincos(v)
+	return Add(lift(e*c), Scale(e*s/v, uv))
```
btracey: Instead of these `Add`, I think it would be clearer to have a `pack` function that is the opposite of `split`. It makes it clearer that the relevant fields are explicitly zero.

num/quat/exp.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+// Pow return q**r, the base-q exponential of r.
+// For generalized compatibility with math.Pow:
+//      Pow(0, ±0) returns 1+0i+0j+0k
+//      Pow(0, c) for real(c)<0 returns Inf+0i+0j+0k if imag(c), jmag(c), kmag(c) are zero,
+//          otherwise Inf+Inf i+Inf j+Inf k.
+func Pow(q, r Quat) Quat {
+	if q == zero {
+		w, uv := split(r)
```
btracey: I don't think this switch statement does what the documentation says. The documentation is about the power, but the switch tests the real part of the quaternion.

num/quat/quat.go
```diff
@@ -0,0 +1,400 @@
...
+	}
+}
+
+// Scale returns q scaled by f.
+func Scale(f float64, q Quat) Quat {
+	return Quat{Real: f * q.Real, Imag: f * q.Imag, Jmag: f * q.Jmag, Kmag: f * q.Kmag}
+}
+
+// Parse converts the string s to a Quat. The string may be parenthesized and
+// has the format [±]N±Ni±Nj±Nk. The order of the components is not strict.
```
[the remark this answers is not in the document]
kortschak: If you parse "5+6j+7k", you would expect to get 5+0i+6j+7k, which is what you get. I'm sorry, I don't really understand this.

num/quat/abs.go
```diff
@@ -0,0 +1,52 @@
+// Copyright ©2018 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Copyright 2017 The Go Authors. All rights reserved.
```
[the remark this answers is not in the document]
kortschak: The files with copyright Go authors contain code that is largely derived from the April Fools CL. The rest are my own with no contribution from those filed.

num/quat/doc.go
kortschak: Done

num/quat/util.go
kortschak: These are as existed in the the original CL, however, my understanding is that that conventional direction for the real basis vector is upwards, with the three imaginary bases orthogonal to that.

num/quat/exp.go
kortschak: I don't understand; the switch tests the real part of the quaternion that q is raised to. There is some leeway; the doc says Pow(0, ±0) returns 1+0i+0j+0k, but the switch only tests the real part. I don't have enough knowledge of quaternion math to knwo whether this is OK.

num/quat/exp.go
kortschak: Do we need to make that clearer? The expressions here reflect the maths more closely. I have added this is a separate commit.

num/quat/exp.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+// Log returns the natural logarithm of q.
+func Log(q Quat) Quat {
+	w, uv := split(q)
+	if uv == zero {
+		return lift(math.Log(w))
+	}
+	v := Abs(uv)
+	return join(math.Log(Abs(q)), Scale(math.Atan2(v, w)/v, uv))
```
vladimir-ch: Where does the `Atan2` come from? The reference from Wikipedia https://users.aalto.fi/~ssarkka/pub/quat.pdf has `Acos`.

num/quat/exp.go
vladimir-ch: The documentation regarding `±0` here is unclear and even misleading but since the implementation follows `cmplx.Pow`, I think I'd leave it as is.

num/quat/quat.go
vladimir-ch: I also agree with the `mag` version although the connection to magnitude is unfortunate because magnitude is inherently positive. But since Go has already `real` and `imag`, and since quaternions are an extension of complex numbers, it makes sense to continue the trend and have `Jmag` and `Kmag`. For me other convincing arguments are that `I, J, K` are small enough to get lost in source and that they are not the quaternion units as the name would suggest.

num/quat/exp.go
vladimir-ch: That's a nice reference, thanks. I had misread the arguments.

num/quat/quat.go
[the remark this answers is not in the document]
kortschak: Ah, I see what you mean now. The documentation does not fully describe the flexibility. If you have wording for that that doesn't make it less clear, please suggest.

The reason for the non-strictness of ordering is that it doesn't harm anything. It could be made strict but it would be a little more complicated.

kortschak:
*ping*

num/quat/trig.go
```diff
@@ -0,0 +1,143 @@
+// Copyright ©2018 The Gonum Authors. All rights reserved.
```
btracey: Do you have a reference for these functions?

kortschak: These were in the original ["math/quat" CL](https://go-review.googlesource.com/c/39220/1). Other than that, no I don't (and yes it does trouble me that the only reference is an April fools joke).

num/quat/quat.go
```diff
@@ -0,0 +1,400 @@
...
```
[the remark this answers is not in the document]
kortschak: That does not properly handle cases where there is an exponent part to the float; what is "(1e-10i)"? An error, `1-10i`, or the correct `0+1e-10i+0j+0k`? The state machine here is the simplest it can be.

num/quat/trig.go
kortschak: @mdempsky Do you remember where you got the references for the quaternion [April fools joke for 2017](https://go-review.googlesource.com/c/39220/1)? In particular, the trig functions.

mdempsky: I'd be skeptical of code from that CL. My intention was primarily to inspire a "wait, is this serious?" reaction, and not expecting any of the code would ever be used in production.

I don't remember the exact origin of the code either. Trying to retrace my steps, what I do vaguely remember at least is reading Dorai Sitaram's "On adding quaternions to Scheme" (http://www.ccs.neu.edu/home/dorai/squat/squat.html), and comparing it with Go's math/cmplx package. I believe the math/quat package was largely my own best-guess extrapolation from those two sources.

kortschak: Thanks. That scheme reference does look promising.

Interestingly, while the remainder of the math/quat looks good as far as I can tell, there was an error in the quaternion multiplcation code.

I like the idea that an April fools joke will get into a codebase — properly vetted.

mdempsky: > there was an error in the quaternion multiplcation code.

Out of curiosity, what/where was this error?

kortschak: The error was in the mul node [here](https://go-review.googlesource.com/c/go/+/39216/1/src/go/constant/value.go#1266). From memory, it gives the wrong sign sometimes.

kortschak: @btracey I've had a look through the Scheme paper and what it covers (circular trig functions) looks OK.

For inverse trig functions, there is [this chapter](https://link.springer.com/content/pdf/10.1007%2F978-3-0348-0622-0_8.pdf) from [Real Quaternionic Calculus Handbook](https://www.springerprofessional.de/en/real-quaternionic-calculus-handbook/1981012) which I don't have access to via our library for other chapters, although Google books has a [reasonable coverage of it](https://books.google.com/books?id=YnS8BAAAQBAJ&printsec=frontcover&source=gbs_ge_summary_r&cad=0#v=onepage&q&f=false) (but not hyperbolic trig).

kortschak: @btracey I've had a look through the Scheme paper and what it covers
(circular trig functions) looks OK.

For inverse trig functions, there is [this chapter](https://link.springer.com/
content/pdf/10.1007%2F978-3-0348-0622-0_8.pdf) from [Real Quaternionic
Calculus Handbook](https://www.springerprofessional.de/en/real-
quaternionic-calculus-handbook/1981012) which I don't have access to
via our library for other chapters, although Google books has a
[reasonable coverage of
it](https://books.google.com/books?id=YnS8BAAAQBAJ&printsec=frontcover&
source=gbs_ge_summary_r&cad=0#v=onepage&q&f=false).

GitHub has been sick, apologies for all those (and this one which I need to add back because it was not handling deletes well either).

btracey: I don't have access either, and I can't find a refernce to the arc functions. They look pretty reasonable though.

kortschak: I'll file an issue to improve thetesting when wecan.

kortschak: I've found a copy of the book and have added at least a test for Sin and Cos, but I suspect there are errors in the book.

If I use the literal translation of definition for sin given by them (using our functions)
```
	sgn := unit(uv)
	return Scale(-0.5, (Mul(sgn, Sub(Exp(Mul(q, sgn)), Exp(Scale(-1, (Mul(q, sgn))))))))
```
I get the same result for the tests as with the current code, but it disagrees with an exercise solution by replacement of sin for cos in one term.

They give sin(1 + i + j + k) = -sin 1 sinh sqrt 3 + (cos 1 sinh sqrt 3)/sqrt 3 (i + j + k), but this does not agree with the code above (though with the real term being -sin 1 cosh sqrt 3 it does).

kortschak:
I have added some additional tests which largely pass (there are a couple of tests that would come from the book, but fail as written there and seem incorrect), others are likely failing in our code, but I don't understand why.

PTAL

kortschak:
What do you think about adding a link to https://eater.net/quaternions to the example code. I know that it's not really our job to be teaching concepts (though I try to in the code I add), but this resource is very nice.

btracey:
We could add it to the example code, but it may make sense to add it to the package documentation.

kortschak:
Done.

PR #589, opened 2018-09-12
mat: avoid hard coding expected output that depends on weak promises

kortschak:
This fixes the formatting issues that have been causing go:master to fail.

Please take a look.


kortschak:
Sorry, I missed one. Now added.

PR #594, opened 2018-09-21
mat: Add Diagonal interface and DiagDense type

### LINKED ISSUE #592: Proposal: Add Diagonal type to mat
I think we should add a Diagonal type to mat. If we also add TriBanded, then I think that will be a complete set of types for real-valued matrices in mat.

Diagonal matrices are a pretty core concept, and I think it'll be something users look for. Actually having it there will make it easy. We've also had a bunch of requests for extracting the diagonal of a matrix (and related), and having a Diagonal type is a clear means to do this.

What is special about Diagonal is that it implements both Triangular and Symmetric, which cannot be done by either TriBanded or SymBanded. This makes it difficult when the task is to extract and then use for a specific purpose, as we can't guess in advance which the user wants. A Diagonal type would implement both, and so there is no need to know.

There is one problem, which is that a Diagonal is both upper and lower triangular, and that is impossible to represent in a boolean. We could redefine the type to be Upper/Lower/Both, but this would have a lot of ramifications on the current codebase, and it would be generally uglier and harder to deal with. I think there is a clean solution though. NewDiagDense  (and associated constructions) return the Diagonal as an upper triangular matrix, just like the default position for Triangular (such as in Cholesky) is to use an upper triangular matrix. The Transpose method on Diagonal would then swap Upper to Lower (and vis-versa), but otherwise be a no-op. This would make it easy to use a Diagonal as a Triangular matrix that requires a specific orientation, but does not change any of our existing code.

btracey:
Fixes #592 .

Please take a look.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -19,6 +19,7 @@ Dan Kortschak <dan.kortschak@adelaide.edu.au> <dan@kortschak.io>
 Daniel Fireman <danielfireman@gmail.com>
 David Samborski <bloggingarrow@gmail.com>
 Davor Kapsa <davor.kapsa@gmail.com>
+DeepMind Technologies
 Egon Elbre <egonelbre@gmail.com>
 Ekaterina Efimova <katerina.efimova@gmail.com>
 Ethan Burns <burns.ethan@gmail.com>
--- a/mat/diagonal.go
+++ b/mat/diagonal.go
@@ -1,0 +1,89 @@
+// Copyright ©2018 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package mat
+
+var (
+	diagDense *DiagDense
+	_         Matrix          = diagDense
+	_         Diagonal        = diagDense
+	_         MutableDiagonal = diagDense
+	_         Triangular      = diagDense
+	_         Symmetric       = diagDense
+)
+
+// Diagonal represents a diagonal matrix, that is a square matrix that only
+// has non-zero terms on the diagonal.
+type Diagonal interface {
+	Matrix
+	Symmetric() int
+	TTri() Triangular
+	// Triangle implements the Triangular interface. Implementers of Diagonal
+	// should return Upper as the default TriKind.
+	Triangle() (int, TriKind)
+	// Diag returns the number of rows/columns in the matrix
+	Diag() int
+}
+
+// MutableDiagonal is a Diagonal matrix whose elements can be set.
+type MutableDiagonal interface {
+	Diagonal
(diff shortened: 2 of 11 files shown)


kortschak:
```
# gonum.org/v1/gonum/optimize
optimize/cmaes.go:215:5: b.SetSymBand undefined (type *mat.DiagDense has no field or method SetSymBand)
```

kortschak:
Please fix the link to the issue.

I'm not entirely sure I understand the strategy behind the methods in the interface. ISTM that it should just be interface{ Matrix; Diag() int }, though it could satisfy relevant interfaces, including RawBanded, RawSymetricBanded and RawTriangularBanded.

I also think that DenseDiag should just be a []float64 with methods in the pointer.

btracey:
Sorry about the failure.

I waffled on the methods in the interface. I suspect we do want them, as it means algorithms that take in a `Diagonal` can use them as a `Symmetric`, just as now we want algorithms that take in a `Symmetric` to also use them as a `Matrix`. (and similarly, we don't have a SymBanded interface, but we probably want it to satisfy `Symmetric`)

btracey:
Although, actually, MutableSymBanded requires satisfying `Symmetric`, and not just `Matrix`, which means at some point we agreed with the stricter interfaces :-)

Added methods to satisfy the proper Raw types

I removed `cap` and `len` because they clearly don't need to be there. I'm not sure that we want it to be `type DiagDense []float64` though, as that would allow casting and such. I can modify if you disagree with that assessment.

kortschak:
I think we do want want on `[]float64` and, correcting myself since we're not doing resizes, not on the pointer. This gives a nice way of doing a diagonal multiply with a `[]float64`. It also removes the need for a `NewDenseDiag` and lets people do slicing and appending.

You can also add `RawTriBand() blas64.RawTriangularBand`.

btracey:
I can add `RawTriBand` if you'd like, though it seemed more appropriate to add when we had a `TriBandDense` type. 

I don't think we want it to be a named type of `[]float64`. It means that `DiagDense` would behave very differently than the other matrix types. In particular, my plan for the next commit was to add an `Extract` command that gets the diagonal from a matrix. This would require a pointer reciever. And so it would mean that for all of the other matrix types you construct with `NewMatrix` or `var a Matrix`, with `DiagDense` you would have to do `diag = &make([]float64,n)` and `var a DiagDense` doesn't work. I think the consistency is worth more than the couple of bells and whistles. 

We also should think about making Diag strided to allow for `DiagView` methods.

kortschak:
Can you add issues for making `DiagDense` strided and adding `RawTriBand() blas64.RawTriangularBand` when there is a `TriBandDense` type?

PR #595, opened 2018-09-24
lapack: rename DecompUpdate type, remove None const

vladimir-ch:
* Split DecompUpdate type into two types, one for applying an orthogonal matrix, one for generating an orthogonal matrix
* Add SchurComp and SchurUpdateComp types and consts
* Rename EVComp consts
* Remove the generic None const, it has now been replaced with type-specific None consts.

Please take a look.


vladimir-ch:
Thanks.

Issue #570, opened 2018-08-14
all: API stability metaissue


kortschak:
This is an issue to aggregate API stability issues from deprecated repos, culled by looking for the "API stability" milestone (and owner added).

Please add any you think are important or mark as completed those that are done.

[graph](https://github.com/gonum/graph/milestone/1):
- [x] https://github.com/gonum/graph/issues/87 @kortschak
- [x] https://github.com/gonum/graph/issues/46 @kortschak
- [x] https://github.com/gonum/graph/issues/18 @kortschak

[lapack](https://github.com/gonum/lapack/milestone/1):
- [x] https://github.com/gonum/lapack/issues/163 @vladimir-ch

[mat](https://github.com/gonum/matrix/milestone/1):
- [x] https://github.com/gonum/gonum/issues/756 @kortschak
- [x] https://github.com/gonum/gonum/issues/748 @btracey
- [x] https://github.com/gonum/gonum/issues/738 @btracey

kortschak:
These are done.

PR #597, opened 2018-09-28
graph: avoid issues around graph node retrieval subtleties

### LINKED ISSUE #31: graph: [proposal] use node ID for retrieval operations and graph.Node for creation

kortschak:
With the approach to graph node mutation on edge setting the previously existed there was an issue that the edge last used connect a pair of nodes could result in a difference in the nodes being returned by a node query compared to the same node associated with edges returned from an edge query.

This change avoids dealing with that by making it implementation dependent and stating this, and by making all the node-storing graphs we provide mutate the nodes when edges are set.

There are a number of ways to deal with the situation of maintaining consistency between edges and nodes, so I think being open to that is the best approach, and while using `int64` for edges did seem plausible, it became difficult very quickly.

Please take a look.

Closes #31.
Fixes gonum/graph#87.
Updates #570.

### DIFF
--- a/graph/graph.go
+++ b/graph/graph.go
@@ -32,7 +32,7 @@ type Graph interface {
 	Has(id int64) bool
 
 	// Node returns the node with the given ID if it exists
-	// in the graph.
+	// in the graph, and nil otherwise.
 	Node(id int64) Node
 
 	// Nodes returns all the nodes in the graph.
--- a/graph/graph.go
+++ b/graph/graph.go
@@ -31,6 +31,10 @@ type Graph interface {
 	// within the graph.
 	Has(id int64) bool
 
+	// Node returns the node with the given ID if it exists
+	// in the graph.
+	Node(id int64) Node
+
 	// Nodes returns all the nodes in the graph.
 	Nodes() Nodes
 
@@ -146,8 +150,10 @@ type EdgeAdder interface {
 	// will be added if they do not exist, otherwise
 	// SetEdge will panic.
 	// The behavior of an EdgeAdder when the IDs
-	// returned by e.From and e.To are equal is
+	// returned by e.From() and e.To() are equal is
 	// implementation-dependent.
+	// Whether e, e.From() and e.To() are stored
+	// within the graph is implementation dependent.
 	SetEdge(e Edge)
 }
 
@@ -162,8 +168,10 @@ type WeightedEdgeAdder interface {
 	// the nodes will be added if they do not exist,
 	// otherwise SetWeightedEdge will panic.
 	// The behavior of a WeightedEdgeAdder when the IDs
-	// returned by e.From and e.To are equal is
+	// returned by e.From() and e.To() are equal is
 	// implementation-dependent.
+	// Whether e, e.From() and e.To() are stored
+	// within the graph is implementation dependent.
 	SetWeightedEdge(e WeightedEdge)
 }
 
--- a/graph/multi/weighted_directed.go
+++ b/graph/multi/weighted_directed.go
@@ -103,7 +103,7 @@ func (g *WeightedDirectedGraph) NewWeightedLine(from, to graph.Node, weight floa
 	return &WeightedLine{F: from, T: to, W: weight, UID: g.lineIDs.NewID()}
 }
 
-// SetWeightedLine adds l, a line from one node to another. If the nodes do not exist, they are added.
+// SetWeightedLine adds l, a line from one node to another. If the nodes do not exist, they are added
 // and are set to the nodes of the line otherwise.
 func (g *WeightedDirectedGraph) SetWeightedLine(l graph.WeightedLine) {
 	var (
--- a/graph/multi/weighted_directed.go
+++ b/graph/multi/weighted_directed.go
@@ -157,16 +157,17 @@ func (g *WeightedDirectedGraph) RemoveLine(fid, tid, id int64) {
 	g.lineIDs.Release(id)
 }
 
-// Node returns the node in the graph with the given ID.
-func (g *WeightedDirectedGraph) Node(id int64) graph.Node {
-	return g.nodes[id]
-}
-
 // Has returns whether the node exists within the graph.
 func (g *WeightedDirectedGraph) Has(id int64) bool {
 	_, ok := g.nodes[id]
 
 	return ok
+}
+
+// Node returns the node in the graph with the given ID if it exists,
+// and nil otherwise.
+func (g *WeightedDirectedGraph) Node(id int64) graph.Node {
+	return g.nodes[id]
 }
 
 // Nodes returns all the nodes in the graph.
--- a/graph/multi/weighted_directed.go
+++ b/graph/multi/weighted_directed.go
@@ -104,6 +104,7 @@ func (g *WeightedDirectedGraph) NewWeightedLine(from, to graph.Node, weight floa
 }
 
 // SetWeightedLine adds l, a line from one node to another. If the nodes do not exist, they are added.
+// and are set to the nodes of the line otherwise.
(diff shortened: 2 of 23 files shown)


graph/graph.go
```diff
@@ -31,6 +31,10 @@ type Graph interface {
 	// within the graph.
 	Has(id int64) bool
 
+	// Node returns the node with the given ID if it exists
+	// in the graph.
```
mewmew:
```diff
 // Node returns the node with the given ID if it exists
-// in the graph.
+// in the graph, and nil otherwise.
```

Also update docs of concrete implementations.

graph/multigraph.go
```diff
@@ -23,6 +23,10 @@ type Multigraph interface {
 	// within the multigraph.
 	Has(id int64) bool
 
+	// Node returns the node with the given ID if it exists
+	// in the multigraph.
```
mewmew:
```diff
-// in the multigraph.
+// in the multigraph, and nil otherwise.
```

graph/path/internal/testgraphs/limited.go
```diff
@@ -166,6 +166,14 @@ func (l *LimitedVisionGrid) Has(id int64) bool {
 	return 0 <= id && id < int64(len(l.Grid.open))
 }
 
+// Node return a node from g if it exists.
```
mewmew: Not sure if we should mention node ID here? Right now its implicit from the context. Same question regarding `Grid.Node`.

graph/graph.go
kortschak: Done.

graph/multigraph.go
kortschak: Done.

graph/path/internal/testgraphs/limited.go
kortschak: Fixed in both.

kortschak:
PTAL

graph/multigraph.go
```diff
@@ -24,7 +24,7 @@ type Multigraph interface {
 	Has(id int64) bool
 
 	// Node returns the node with the given ID if it exists
```
mewmew: For consistency with `// Node returns the node in the graph with the given ID if it exists, and nil otherwise.`

```diff
- // Node returns the node with the given ID if it exists in the multigraph, and nil otherwise.
+// Node returns the node in the multigraph with the given ID if it exists, and nil otherwise.
```

Actually, come to think of it. I prefer the way multigraph was stated. I.e. `Node returns the node with the given ID if it exists in the multigraph, and nil otherwise`.

So, for consistency, either update the multigraph comment, or better yet (in my opinion), the other comments to `// Node returns the node with the given ID if it exists in the graph, and nil otherwise.`

graph/path/internal/testgraphs/limited.go
```diff
@@ -161,12 +161,13 @@ func (l *LimitedVisionGrid) NodeAt(r, c int) graph.Node {
...
 
-// Has returns whether n is a node in the grid.
+// Has returns whether the node with the given ID is a node in the grid.
 func (l *LimitedVisionGrid) Has(id int64) bool {
 	return 0 <= id && id < int64(len(l.Grid.open))
 }
 
-// Node return a node from g if it exists.
+// Node returns the node in the graph with the given ID if it exists,
+// and nil otherwise.
```
mewmew: As `LimitedVisionGrid` does not actually store the node, should this be pointed out in the comment? The same goes for `Grid` (both of which return a new `simple.Node`).

graph/multi/weighted_directed.go
```diff
@@ -104,6 +104,7 @@ func (g *WeightedDirectedGraph) NewWeightedLine(from, to graph.Node, weight floa
 }
 
 // SetWeightedLine adds l, a line from one node to another. If the nodes do not exist, they are added.
```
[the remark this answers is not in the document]
kortschak: Fixed. Thanks

graph/path/internal/testgraphs/limited.go
kortschak: Whether the node is stored or not is not relevant here. But also, these are internal testing packages, so I'm not particularly fussed.

graph/multigraph.go
kortschak: Done latter.

kortschak:
PTAL

graph/undirect.go
```diff
@@ -14,6 +14,9 @@ var _ Undirected = Undirect{}
 // Has returns whether the node exists within the graph.
 func (g Undirect) Has(id int64) bool { return g.G.Has(id) }
 
+// Node returns the node with the given ID from the graph if it exists.
```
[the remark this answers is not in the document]
kortschak: I'll do that later. Each change dismisses approvals.

PR #598, opened 2018-09-28
blas: change matrix description types to byte

vladimir-ch:
Please take a look.


vladimir-ch:
I'd still like to review the type and constants not related to eigendecomposition before closing https://github.com/gonum/lapack/issues/163 but that should be quick.

Issue #602, opened 2018-09-30
graphs/gen: TunableClusteringScaleFree does not correctly set up initial state for graph


kortschak:
The algorithm asks that the graph be initialised with m nodes and no edges. Currently it will panic if given a `dst` with any nodes in {1, 2, ... m-1} because of a typo in that loop.

```
	// Initial condition.
	wt := make([]float64, n)
	for u := 0; u < m; u++ {
		if !dst.Has(int64(0)) {
			dst.AddNode(simple.Node(u))
		}
		// We need to give equal probability for
		// adding the first generation of edges.
		wt[u] = 1
	}
```

While fixing that, `dst.NewNode` should also be used in place of `simple.Node` (and `simple.Edge` elsewhere).

PR #603, opened 2018-09-30
graph: remove Has method from Graph and Multigraph

### LINKED ISSUE #601: proposal: graph: remove Has methods
### Background
I have just updated the graph API to address a long standing issue of a subtle and potentially surprising intereaction between node and edge retrieval from graphs. In order to allow users to manage this subtlety I needed to add a method to retrieve a node from the graph — as distinct from obtaining a potentially new node via the `NewNode` method. This means there are now two approaches to determining whether a node exists in the graph:
```
exists := g.Has(id)
```
and
```
exists := g.Node(id) != nil
```
When I added the `Node` method, I was aware of this, but my thinking was that `g.Has` is potentially cheaper than returning a node. However, whether this is significant, or even true, is not something I looked at. I doubt in many cases whether either of these is true.

The existence of two approaches to obtaining the knowledge of existence of a node allows skew between them and reduces code consistency. I also think that while `g.Node(id) != nil` is longer it arguably more clearly states what is being asked to the graph.

### Proposal
I would like to suggest that the `Has` method be removed from the `Graph` and `Multigraph` interfaces and all implementations we provide.

This will be quite an invasive, but largely mechanical change (made less amenable to mechanisation by the existence of a `Has` method on our set types).

It is however a breaking change since `Has` has been an integral part of the graph API and node existence testing is central to many graph querying algorithms.

I will do the work.

/cc @vladimir-ch @mewmew

kortschak:
Please take a look.

Closes #601.


kortschak:
Absolutely.

Note also that I've just cleaned up a couple of cases that I missed (and deleted a `// +build ignore`'d file that was written against the pre-iterator API (and I presumably left accidentally while debugging).

graph/graphs/gen/holme_kim.go
```diff
@@ -48,7 +48,7 @@ func TunableClusteringScaleFree(dst graph.UndirectedBuilder, n, m int, p float64
 	// Initial condition.
 	wt := make([]float64, n)
 	for u := 0; u < m; u++ {
-		if !dst.Has(int64(0)) {
+		if dst.Node(int64(0)) == nil {
```
kortschak: This is incorrect (#602), but will be fixed in another PR.

kortschak:
I should add. Having the changes here, I have tested gorgonia to see if it breaks. It doesn't. I suspect that few people actually use `g.Has(id)` outside gonum/graph since usually people are either using our routines or doing `g.Nodes()` and graph traversal, which together imply that `g.Has(id)` is true.

graph/community/louvain_undirected.go
```diff
@@ -296,20 +296,20 @@ func reduceUndirected(g graph.Undirected, communities [][]graph.Node) *ReducedUn
...
-	if g.Has(id) {
+	if g.has(id) {
 		return g.nodes[id]
 	}
 	return nil
 }
 
+// has returns whether the node exists within the graph.
+func (g *ReducedUndirected) has(id int64) bool {
+	return 0 <= id || id < int64(len(g.nodes))
```
vladimir-ch: `&&`

kortschak: Thanks. Fixed

kortschak:
PTAL

PR #605, opened 2018-10-04
stat: add support for LinInterp CumulantKind for Quantile

vthiery:
This PR adds the support of a new kind of `CumulantKind`: `LinInterp`
The name was suggested [here](https://github.com/gonum/gonum/pull/254/files#r144047424) so I used it.

I also took the liberty to split the test on the `Quantile` function to make the test intention clearer (functional / invalid input checks / invalid cumulant kind)

### DIFF
--- a/stat/stat.go
+++ b/stat/stat.go
@@ -20,6 +20,8 @@ type CumulantKind int
 const (
 	// Empirical treats the distribution as the actual empirical distribution.
 	Empirical CumulantKind = 1
+	// LinInterp use a linear interpolation on the empirical distribution.
+	LinInterp CumulantKind = 2
 )
 
 // bhattacharyyaCoeff computes the Bhattacharyya Coefficient for probability distributions given by:
@@ -1028,6 +1030,7 @@ func MomentAbout(moment float64, x []float64, mean float64, weights []float64) f
 // CumulantKind behaviors:
 //  - Empirical: Returns the lowest value q for which q is greater than or equal
 //  to the fraction p of samples
+//  - LinInterp: Returns the linearly interpolated value
 func Quantile(p float64, c CumulantKind, x, weights []float64) float64 {
 	if !(p >= 0 && p <= 1) {
 		panic("stat: percentile out of bounds")
@@ -1052,6 +1055,8 @@ func Quantile(p float64, c CumulantKind, x, weights []float64) float64 {
 	switch c {
 	case Empirical:
 		return empiricalQuantile(p, x, weights, sumWeights)
+	case LinInterp:
+		return linInterpQuantile(p, x, weights, sumWeights)
 	default:
 		panic("stat: bad cumulant kind")
 	}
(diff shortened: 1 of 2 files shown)


btracey:
Thanks for the PR.

I stand by my comments in https://github.com/gonum/gonum/pull/254/files though, which is that `LinInterp` is really type 4, and we should be doing that instead of the actual algorithm for 2.  I haven't looked at the code yet, but given your comment, I assume the code implements type 4, and so we should number it 4 :-)

sbinet:
needs https://github.com/gonum/gonum/pull/612

stat/stat.go
```diff
@@ -20,6 +20,8 @@ type CumulantKind int
 const (
 	// Empirical treats the distribution as the actual empirical distribution.
 	Empirical CumulantKind = 1
+	// LinInterp use a linear interpolation on the empirical distribution.
```
sbinet: `s/use/uses/`

stat/stat.go
```diff
@@ -1073,6 +1078,30 @@ func empiricalQuantile(p float64, x, weights []float64, sumWeights float64) floa
...
+		if weights == nil {
+			cumsum++
+		} else {
+			cumsum += weights[i]
+		}
+		if cumsum >= fidx {
+			if i == 0 {
+				return x[i]
+			}
+			width := 1.0
```
kortschak:
```
t := cumsum - fidx
if weights != nil {
    t /= weights[i]
}
return t*x[i-1] + (1-t)*x[i]
```

stat/stat.go
```diff
@@ -20,6 +20,8 @@ type CumulantKind int
 const (
 	// Empirical treats the distribution as the actual empirical distribution.
 	Empirical CumulantKind = 1
+	// LinInterp uses a linear interpolation on the empirical distribution.
```
btracey: Maybe

```
LinInterp linearly interpolates the empirical distribution between sample values, where small p are given the first value of x, and p=1 is the last value of x
```

I'm not picky about the exact wording, I mostly want the documentation to convey that this interpolation is direction asymmetric.

vthiery: Maybe a clearer way to express `where small p are given the first value of x, and p=1 is the last value of x`  would be `with a flat extrapolation`?

btracey:
Thank you for the pull request!

sbinet:
could you rebase and then squash into one commit with a nice commit message?
thanks :)

sbinet:
@vthiery I can do it for you if you prefer, but here are the commands for reference:
```
$> git fetch --all -p
$> git co add-lin-interp-quantile
$> git rebase -i origin/master
## squash into one commit, then edit commit message
$> git push -f vthiery add-lin-interp-quantile
```

btracey:
We can also do it here from the github API though, right?

sbinet:
it's here: https://github.com/gonum/gonum/compare/master...sbinet-gonum:add-lin-interp-quantile

I must admit I am not completely sure what the "squash and merge" button of github will do :}

kortschak:
Squash and merge does a squash and rebase onto master.

vthiery:
@sbinet and @kortschak thanks for merging! (sry for the delay, was not online this we)

PR #606, opened 2018-10-04
Add doc related to sumWeights <= 1 case when computing variance

vthiery:
When the sum of weights equals 1, the unbiased estimator used to compute the variance is ill-defined. This PR adds some doc to inform the user that a biased estimator should be used instead

### DIFF
--- a/stat/stat.go
+++ b/stat/stat.go
@@ -1076,6 +1076,7 @@ func empiricalQuantile(p float64, x, weights []float64, sumWeights float64) floa
 // Skew computes the skewness of the sample data.
 // If weights is nil then all of the weights are 1. If weights is not nil, then
 // len(x) must equal len(weights).
+// Note that in case of weights summing to 1, a biaised variance estimator must be used instead
 func Skew(x, weights []float64) float64 {
 
 	mean, std := MeanStdDev(x, weights)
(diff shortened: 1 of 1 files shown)


stat/stat.go
```diff
@@ -1076,6 +1076,8 @@ func empiricalQuantile(p float64, x, weights []float64, sumWeights float64) floa
 // Skew computes the skewness of the sample data.
 // If weights is nil then all of the weights are 1. If weights is not nil, then
 // len(x) must equal len(weights).
+// Note that in case of weights summing to less than (or equal) to 1,
```
kortschak: Please remove the parentheses here and below.

I would prefer something like "`// When weights sum to 1 or less, a biased variance estimator should be used.`"

vthiery: That is indeed clearer, thanks!

btracey:
It looks like they (or we) changed a setting where all reviewers have to approve? Not sure we want our policy to be that way.

kortschak:
>It looks like they (or we) changed a setting where all reviewers have to approve? Not sure we want our policy to be that way.

No, I'm pretty sure that is there's a request for changes that's always been blocking. If worst comes to worst, another person can always dismiss the review.

Issue #614, opened 2018-10-05
graph: decide whether empty collections of nodes, edges and lines are nil or just empty


kortschak:
With the old approach we could always return a `nil` and people could still iterate over it without needing to check for `nil`. This is no longer true since the container types returned by these methods are interface values now. The `graph.<Types>Of` helpers hide this for the most part at the moment because they are guaranteed to work with `nil` input, but user code is not.

We currently don't specify the `nil`-ness of returns for collections, so this is not terrible, but it might be an idea to tighten that down.

There are, I think. three approaches:

1. leave it as is and let users handle the `nil` values.
2. require that containers are always non-`nil`.
3. provide an iterator helper functions that wraps `it.Next` and performs the check for `nil` like so:
```
func Next(it Iterator) bool {
    return it != nil && it.Next()
}
```

kortschak:
> Is it reasonable to assume that users would type assert to check whether a returned iterator is `graph.Empty`? Otherwise, perhaps it belongs in the `iterator` package.

They might, but we are not requiring it. Going in iterator is a possibility, but I feel that it is less discoverable there and is not near the definitions of the interfaces. The types in interator are helpers for us and possibly for others. `Empty` is a first class, though optional, entity.

> Also, while `Empty` signal intent very well, would it suffice to use e.g. `iterator.NewOrderedWeightedEdges(nil)` and the respective counter-parts for nodes, lines, etc?

That is possible, but requires more work and space. The `Empty` entity has a clear meaning as well.

PR #615, opened 2018-10-05
graph/multi: clean up missed changes and bugs

kortschak:
This fixes some bugs in the multigraph handling of edges from lines and finishes changes that I had intended to go into the iterator PR (the change in container for `graph.*Lines` in the `multi.*Edge` types).

It has also raised an issue of how I have been dealing with empty collections. With the old approach we could always return a `nil` and people could still iterate over it without needing to check for `nil`. This is no longer true. We currently don't specify the `nil`-ness of returns for collections, so this is not terrible, but it might be an idea to tighten that down. I'm raising this here since you can see the impacts with these changes, but I will file an issue (#614).

Please take a look.

### DIFF
--- a/graph/multi/directed.go
+++ b/graph/multi/directed.go
@@ -193,7 +193,8 @@ func (g *DirectedGraph) Edges() graph.Edges {
 			}
 			if len(lines) != 0 {
 				edges = append(edges, Edge{
-					F: g.Node(u.ID()), T: g.Node(lines[0].To().ID()),
+					F:     g.Node(u.ID()),
+					T:     g.Node(lines[0].To().ID()),
 					Lines: iterator.NewOrderedLines(lines),
 				})
 			}
--- a/graph/multi/directed.go
+++ b/graph/multi/directed.go
@@ -181,15 +181,21 @@ func (g *DirectedGraph) Nodes() graph.Nodes {
 // Edges returns all the edges in the graph. Each edge in the returned slice
 // is a multi.Edge.
 func (g *DirectedGraph) Edges() graph.Edges {
+	if len(g.nodes) == 0 {
+		return nil
+	}
 	var edges []graph.Edge
 	for _, u := range g.nodes {
 		for _, e := range g.from[u.ID()] {
-			var lines Edge
+			var lines []graph.Line
 			for _, l := range e {
 				lines = append(lines, l)
 			}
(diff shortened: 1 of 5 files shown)


kortschak:
Updates #31.

kortschak:
*ping*

graph/multi/multi.go
```diff
@@ -54,41 +48,42 @@ func (l Line) ID() int64 { return l.UID }
...
 // Weight returns the weight of the edge.
 func (e WeightedEdge) Weight() float64 {
+	if e.WeightedLines == nil {
+		return 0
+	}
 	if e.WeightFunc == nil {
 		var w float64
-		for _, l := range e.Lines {
-			w += l.Weight()
+		for e.Next() {
```
vladimir-ch: Is it ok to assume that `e` is at the first line? Above it is documented for the input to `WeightFunc` but what about `Weight`?

graph/multi/weighted_directed.go
```diff
@@ -30,7 +30,11 @@ var (
 
 // WeightedDirectedGraph implements a generalized directed graph.
 type WeightedDirectedGraph struct {
-	EdgeWeightFunc func([]graph.WeightedLine) float64
+	// EdgeWEightFunc is used to provide
```
vladimir-ch: `EdgeWeightFunc` (lower-case e)

graph/multi/directed.go
```diff
@@ -181,15 +181,21 @@ func (g *DirectedGraph) Nodes() graph.Nodes {
...
 		for _, e := range g.from[u.ID()] {
-			var lines Edge
+			var lines []graph.Line
 			for _, l := range e {
 				lines = append(lines, l)
 			}
 			if len(lines) != 0 {
-				edges = append(edges, lines)
+				edges = append(edges, Edge{
+					F: g.Node(u.ID()), T: g.Node(lines[0].To().ID()),
```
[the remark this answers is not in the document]
kortschak: Changed

graph/multi/multi.go
kortschak: Added documentation making this clear.

graph/multi/weighted_directed.go
kortschak: Thanks. Fixed.

kortschak:
PTAL

PR #618, opened 2018-10-10
graph/formats/gexf12: new package for handling GEXF1.2 graph description files

kortschak:
This was largely written by aqwari.net/xml/cmd/xsdgen working on the XSD files available from the [gephy/gexf specs repository](https://github.com/gephi/gexf/tree/81ba4e7ccdc25631f836fc5caa4ed64ba5300379/specs/1.2draft), with some help; the specs use includes which cause xsdgen to fail, so I manually in-lined the includes.

The generated code was then edited to make optional attributes exist on pointers, to reduce the length of labels to within reasonably normal Go idiom and removing unnecessary namespacing.

The package is versioned since the format appears to arbitrarily be updated.

Please take a look.

### DIFF
--- a/graph/formats/gexf12/gexf.go
+++ b/graph/formats/gexf12/gexf.go
@@ -1,0 +1,304 @@
+// Copyright ©2018 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Package gexf12 implements marshaling and unmarshaling of GEXF1.2 documents.
+//
+// For details of GEXF see https://gephi.org/gexf/format/.
+package gexf12 // import "gonum.org/v1/gonum/graph/formats/gexf12"
+
+import (
+	"bytes"
+	"encoding/xml"
+	"time"
+)
+
+// BUG(kortschak): The namespace for GEFX1.2 is 1.2draft, though it has
+// already been deprecated. There is no specification for 1.3, although
+// it is being used in the wild.
+
+// Content holds a GEFX graph and metadata.
+type Content struct {
+	XMLName xml.Name `xml:"http://www.gexf.net/1.2draft gexf"`
+	Meta    *Meta    `xml:"meta,omitempty"`
+	Graph   Graph    `xml:"graph"`
+	// Version must be "1.2".
+	Version string  `xml:"version,attr"`
+	Variant *string `xml:"variant,attr,omitempty"`
+}
+
+// Meta holds optional metadata associated with the graph.
+type Meta struct {
+	Creator      string    `xml:"creator,omitempty"`
+	Keywords     string    `xml:"keywords,omitempty"`
+	Description  string    `xml:"description,omitempty"`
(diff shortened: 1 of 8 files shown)


kortschak:
I'm not sure whether I will write the cognate package in encoding. XML is too flexible for a sensible general approach there.

kortschak:
Updates gonum/graph#9.

kortschak:
@mewmew I've reconsidered the zero-value encoding and I think you're right. There are two parts to this: the first deals with the node/edge count — it's a hint, which means that if it's zero there is no gain or loss of information from it's presence/absence when it's zero, and also despite the documentation saying it should be used, I've not often seen it in documents; the second part is the defaults case — all the attributes are encoded stringly typed, so if a default is not a string, it's default would be actively encoded, and if it _is_ a string, the default (at least in Go) is `""`.

For struct values, I think the situation leans more towards the use of pointers, but we get Go syntax support there.

I can either add this to #619 or send a separate PR. Views?

mewmew:
> I can either add this to #619 or send a separate PR. Views?

Glad you are looking into using zero values. Feel free to add it either in the open PR, or a new one. I don't think I'll have any time to review in the foreseeable future, as Uni is ramping up for exams.

Cheers,
Robin

edward1958:
Hi! i hope that you don't mind my posting here. I have text network output files in xml and would like to import them into Gephi to generate graphs and statistics. Can I convert the xml file to gexf or is there another preferred method? Any guidance is appreciated. Thanks a bunch!

kortschak:
Sure. Though the container format is irrelevant. If you have knowledge of the node and edge structure of your network data then you can construct the GEXF relationships (for example see [here](https://github.com/kortschak/mbg/blob/c5a94a4f907848b4d08ab5843fb0bb14fa3eeb18/mbg.go#L265-L331)). Rather than thinking about your data as XML, I would just unmarshal it using the encoding/xml package, then fill a gexf12 data structure and marshal that.

Note that we only write GEXF1.2, which due to an honestly bizarre deprecation policy by Gephi is no longer supported even though the GEXF1.3 spec has not been release (the GEXF1.3 parser is hand rolled in gephi making it something that is not workable from). Gephi will read this, but emits a warning.

edward1958:
Hi Dan! Thanks a bunch for your response. I have limited knowledge with this application and do appreciate your guidance. Attached is an xml network output file (zipped). If you don't mind my asking, how would I convert it so that I can run and generate output in Gephi? Again, thank you very much. Ed

[Gephi Working Sample Text.zip](https://github.com/gonum/gonum/files/2637022/Gephi.Working.Sample.Text.zip)

kortschak:
There's really not enough information for me to work from there. There's no link to an XSD or DTD so I don't know what schema is actually being used or who wrote it. If you do have this information, you will be able to write a set of types to make use of the [xml package](https://golang.org/pkg/encoding/xml/) to unmarshal the data. You can then transcribe the relevant parts into the types in the gexf12 package and marshal that collection.

edward1958:
I found the schema. I will attempt to follow your instructions. Thank you for your time and guidance.

PR #623, opened 2018-10-16
Lapack/types

vladimir-ch:
Closes https://github.com/gonum/lapack/issues/163
Updates #570 

Please take a look.

### DIFF
--- a/lapack/gonum/dgesvd.go
+++ b/lapack/gonum/dgesvd.go
@@ -26,7 +26,7 @@ const noSVDO = "dgesvd: not coded for overwrite"
 // jobU and jobVT are options for computing the singular vectors. The behavior
 // is as follows
 //  jobU == lapack.SVDAll       All m columns of U are returned in u
-//  jobU == lapack.SVDInPlace   The first min(m,n) columns are returned in u
+//  jobU == lapack.SVDStore     The first min(m,n) columns are returned in u
 //  jobU == lapack.SVDOverwrite The first min(m,n) columns of U are written into a
 //  jobU == lapack.SVDNone      The columns of U are not computed.
 // The behavior is the same for jobVT and the rows of V^T. At most one of jobU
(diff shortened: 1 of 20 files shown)


lapack/gonum/dgesvd.go
```diff
@@ -26,7 +26,7 @@ const noSVDO = "dgesvd: not coded for overwrite"
 // jobU and jobVT are options for computing the singular vectors. The behavior
 // is as follows
 //  jobU == lapack.SVDAll       All m columns of U are returned in u
-//  jobU == lapack.SVDInPlace   The first min(m,n) columns are returned in u
+//  jobU == lapack.SVDStore     The first min(m,n) columns are returned in u
```
vladimir-ch: For me "in-place" and "overwrite" are synonyms, "store" indicates that the result is, well, stored in a designated destination.

lapack/lapack.go
```diff
@@ -108,21 +108,21 @@ const (
 type SVDJob byte
 
 const (
-	SVDAll       SVDJob = 'A' // Compute all singular vectors
-	SVDInPlace   SVDJob = 'S' // Compute the first singular vectors and store them in provided storage.
-	SVDOverwrite SVDJob = 'O' // Compute the singular vectors and store them in input matrix
-	SVDNone      SVDJob = 'N' // Do not compute singular vectors
+	SVDAll       SVDJob = 'A' // Compute all columns of the matrix U or V.
```
vladimir-ch: Perhaps it is better to avoid referring to specific matrices and use "of the orthogonal matrix" or "on the input matrix" instead.

vladimir-ch:
There is one more place in lapack constants where I'd like to do something but don't know what is "best". There are four `MatrixNorm` constants. `NormFrob` is the only one with `Norm` prefix and I don't like this anomaly. So either 1) prepend `Norm` to the other constants and shorten `MaxColumnSum` to `MaxColSum`, or 2) rename `NormFrob` to `Frob`, or 3) rename `NormFrob` to `Frobenius`. Suggestions? This would be the final nitpick from me on this topic.

kortschak:
On the last comment, I'd be happy for `NormFrob` to be renamed `Frobenius`.

lapack/gonum/dgesvd.go
kortschak: This seems reasonable.

lapack/lapack.go
kortschak: A, U and V make good short hand and are well establish terminology. Perhaps add a note on the type that U and V correspond to the orthogonal matrix and A to the input matrix to make that absolutely clear?

kortschak:
This all looks good to me. Let me know if you're happy with it as it stands and I'll approve, otherwise send more here and I'll take another look.

lapack/lapack.go
vladimir-ch: I didn't change the comment for the type, I just added the adjective "orthogonal" where U and V are mentioned in the const comments.

vladimir-ch:
I renamed `NormFrob` to `Frobenius`, and unified the style of comments (XXX are computed -> Compute XXX).

Please take a look.

Issue #396, opened 2018-01-20
graph/encoding: update serialisers and deserialisers to handle multrgraphs


kortschak:
/cc @mewmew

mewmew:
@kortschak It's been great to follow your work on implementing multigraph support in gonum. Would you wish to propose updates to the API? I'd be happy to give some feedback on the needed API changes, however, I can't lead the work as I have quite a bit of projects at Uni this spring. Fun stuff for sure, and also time consuming. Just to let you know.

kortschak:
The only change will be the addition of `encoding.MultiBuilder` and adding the appropriate wiring for that. The serialisation is simpler, a graph can just be type switched and use the `Lines` as edges if they exist. Does that sound reasonable to you?

mewmew:
> The only change will be the addition of encoding.MultiBuilder and adding the appropriate wiring for that. The serialisation is simpler, a graph can just be type switched and use the Lines as edges if they exist. Does that sound reasonable to you?

Sounds reasonable.

At first I was not sure how to keep the edges distinct when deserialising, however seeing as the edge IDs only have to be unique then that should not matter. E.g. when parsing the following multigraph, the first edge from A to B would get edge ID 1 (*the first unused ID*) and the second edge would get edge ID 2, and their attributes would when creating the graph during deserialisation be attached to the associated edges; in this case label *edge 2* is assigned to edge with ID 1.

```dot
digraph {
	A -> B [label='edge 2']
	A -> B [label='edge 1']
}
```

I don't think the above thoughts provided much useful information, however feel free to take a stab at defining the MultiBuilder interface and fleshing out the serialiser, and wiring up the deserialiser so it passes a round-trip test.

Let me know if you wish to bounce some ideas, or hit any stumbling blocks when dealing with the DOT decoder.

Cheers
/u

32bitkid:
I have a branch that takes a stab at dot encoding of multigraphs: 

https://github.com/32bitkid/gonum/tree/multigraph-encoding

The external API doesn't change on `Marshal()` and internally the implementation is abstracted to either output the simply edge if `g` is not a `Multigraph`, or iterate and output the lines if it is. It supports both attributes on edges/lines as well as `ports:compass` annotations (the same way that `strict` graphs currently do), if they implemented on the `graph.Line`.

I would love to get some feedback on the implementation and other possible tests that would make sense before queueing up a PR. Thanks!

mewmew:
[...]

PR #624, opened 2018-10-18
graph/encoding/dot: add multigraph serialization and deserialization support 

32bitkid:
This should resolve issue #396, in the way that has been discussed there. This change introduces a new pair of `MarshalMulti()` and `UnmarshalMulti()` functions, both which take `graph.Multigraph`. Please take a look.

There are a couple outstanding questions on this particular implementation:

- The arguments for `Marshal()` have remained unchanged. However, the new function `MarshalMulti()` does not include the `strict` argument, because it doesn't _really_ make any sense in that context. But, I wonder if `Marshal()` should _always_ emit a strict DOT graph, because thats what it really represents. However, that might be too big of a breaking change here.

-  In the `multiprinter.print()` function, I need to iterate lines, which I do via `lines := graph.LinesOf(g.Lines(nid, tid))`, I'm not sure if I need to _explicitly_ sort the lines here to get a stable output, like one does a few lines up on `NodesOf()`, and if so, then what would be the preferred sorting strategy to use here?

If the approach looks worth moving forward, then I'll fill in some more testing coverage before finalizing the PR.

Thanks!

### DIFF
--- a/graph/encoding/dot/decode.go
+++ b/graph/encoding/dot/decode.go
@@ -50,6 +50,19 @@ func Unmarshal(data []byte, dst encoding.Builder) error {
 	return copyGraph(dst, file.Graphs[0])
 }
 
+// UnmarshalMulti parses the Graphviz DOT-encoded data as a multigraph and
+// stores the result in dst.
+func UnmarshalMulti(data []byte, dst encoding.MultiBuilder) error {
+	file, err := dot.ParseBytes(data)
+	if err != nil {
+		return err
+	}
+	if len(file.Graphs) != 1 {
+		return fmt.Errorf("invalid number of graphs; expected 1, got %d", len(file.Graphs))
+	}
+	return copyMultigraph(dst, file.Graphs[0])
+}
+
 // copyGraph copies the nodes and edges from the Graphviz AST source graph to
 // the destination graph. Edge direction is maintained if present.
 func copyGraph(dst encoding.Builder, src *ast.Graph) (err error) {
@@ -62,9 +75,41 @@ func copyGraph(dst encoding.Builder, src *ast.Graph) (err error) {
 			panic(e)
 		}
 	}()
-	gen := &generator{
-		directed: src.Directed,
-		ids:      make(map[string]graph.Node),
+	gen := &graphGenerator{
+		generator: generator{
+			directed: src.Directed,
+			ids:      make(map[string]graph.Node),
+		},
+	}
+	if dst, ok := dst.(DOTIDSetter); ok {
+		dst.SetDOTID(src.ID)
+	}
+	if a, ok := dst.(AttributeSetters); ok {
+		gen.graphAttr, gen.nodeAttr, gen.edgeAttr = a.DOTAttributeSetters()
+	}
+	for _, stmt := range src.Stmts {
+		gen.addStmt(dst, stmt)
+	}
+	return err
+}
+
+// copyMultigraph copies the nodes and edges from the Graphviz AST source graph to
+// the destination graph. Edge direction is maintained if present.
+func copyMultigraph(dst encoding.MultiBuilder, src *ast.Graph) (err error) {
+	defer func() {
+		switch e := recover().(type) {
+		case nil:
+		case error:
+			err = e
+		default:
+			panic(e)
+		}
+	}()
+	gen := &multigraphGenerator{
+		generator: generator{
+			directed: src.Directed,
+			ids:      make(map[string]graph.Node),
+		},
 	}
 	if dst, ok := dst.(DOTIDSetter); ok {
 		dst.SetDOTID(src.ID)
@@ -97,7 +142,7 @@ type generator struct {
 
 // node returns the gonum node corresponding to the given dot AST node ID,
 // generating a new such node if none exist.
-func (gen *generator) node(dst encoding.Builder, id string) graph.Node {
+func (gen *generator) node(dst graph.NodeAdder, id string) graph.Node {
 	if n, ok := gen.ids[id]; ok {
 		return n
 	}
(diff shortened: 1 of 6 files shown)


kortschak:
Addressing your questions before I take a look.

There are reasons why you might not want `strict` on a simple graph and why you might want `strict` on a DOT rendering of a multigraph (in this latter case, you have a multigraph internally that has had some work done on it that you expect to leave it as a simple graph in the multigraph container). Whether this check is work that should happen in the graph as opposed to the serialisation is not immediately obvious to me. The other option is for neither serialisation routine to prepend with `strict ` and do as the xml package does and let the user just do `"strict %s"` or similar. I think this warrants further discussion, and I'm not particularly worried about breaking things if we get it right.

The lines will need to be ordered. The appropriate ordering would be lexical on DOTID with the from node being the first key and the to node being the second key.

32bitkid:
Re `strict`: I think the solution that I would, personally, be most in favor of would be to remove the `strict` parameter, and be to have the `Marshal(graph.Graph)` method _always_ output a `strict graph` and `MarshalMulti(graph.Multigraph)` _never_ output a `strict` and drop the parameter for both, and then introduce a new method for converting between `graph.Graph` and `graph.Multigraph` where it would return an `error` if it is not convertible. But thats just me, and again, I feel like its beyond the scope of this PR specifically.

As such, I think your opinion on why one might serialize a `strict` multigraph is well-taken. For the short-term, I think it's probably best to re-introduce the argument here, and then the signatures and functionality will be the same for both. And a more thoughtful discussion on what the right path is forward can result in another PR.

On the latter, I'll make some updates and tests to verify the ordering is correct and stable; I agree they need to be stable, just wasn't sure if it needed _explicit_ sorting in the MarshalMulti() loop, or if just asking for `g.Lines(from, to)` was enough to guarantee correct ordering.

kortschak:
Thanks.

In the ordering of things like lines between a pair of node, I go out of my way to not guarantee any ordering unless it makes sense to have an ordering (like here). The rationale for this is that having an ordering on items that are not intrinsically ordered leads to bad graph algorithm implementations.

kortschak:
BTW Would you file an issue to look at the strict situation here. Add the <kbd>stability</kbd> label.

32bitkid:
@kortschak I created an issue, and described it the best I can. #625 However, as a contributor, I don't think I'm privileged enough to assign labels. Feel free to reword it or change it however you see fit. Thanks!

kortschak:
Is this ready for review?

32bitkid:
@kortschak i think this is ready for review now. Thanks!

kortschak:
Thanks. I'll be able to get to this in a few days.

graph/encoding/dot/decode.go
```diff
@@ -50,6 +50,19 @@ func Unmarshal(data []byte, dst encoding.Builder) error {
 	return copyGraph(dst, file.Graphs[0])
 }
 
+// UnmarshalMulti parses the Graphviz DOT-encoded data as a multigraph and
+// stores the result in dst.
```
kortschak: We should probably have a note that only the first graph of one can be stored (same for `Unmarshal`).

I'll file an issue.

graph/encoding/dot/decode.go
```diff
@@ -117,8 +162,85 @@ func (gen *generator) node(dst encoding.Builder, id string) graph.Node {
 	return n
 }
 
+// applyPortsToEdge applies the available port metadata from an ast.Edge
```
kortschak: Can you reorder this to reduce diff noise please.

graph/internal/ordered/sort.go
```diff
@@ -74,3 +74,20 @@ func Reverse(nodes []graph.Node) {
 		nodes[i], nodes[j] = nodes[j], nodes[i]
 	}
 }
+
+// Lines implements the sort.Interface sorting a slice of graph.Line
```
kortschak: Please rename this to `LinesByIDs`.

32bitkid:
Will do.

kortschak:
See @mewmew's comment on self-loop test cases [here](https://github.com/gonum/gonum/issues/648#issuecomment-438578846).

32bitkid:
@kortschak I rebased and reorganized the code a bit to minimize the diff noise. I also added some tests/assertions for multigraph self-loops.

Though, while the re-organization of the code makes the diff a little easier to parse/approve this PR, it probably need to be _actually_ be re-ogranized into a more idiomatic form after with proper method grouping.

graph/encoding/dot/encode.go
```diff
@@ -265,11 +282,44 @@ func (p *printer) print(g graph.Graph, name string, needsIndent, isSubgraph bool
...
+	if name != "" {
+		p.buf.WriteByte(' ')
+		p.buf.WriteString(name)
+	}
+
+	p.openBlock(" {")
 	return nil
 }
 
+func (p *printer) printBackMatter() {
```
kortschak: Do we need this?

graph/encoding/encoding.go
```diff
@@ -12,6 +12,12 @@ type Builder interface {
 	graph.Builder
 }
 
+// Builder is a graph that can have user-defined nodes and edges added.
```
kortschak: `MultiBuilder`

graph/encoding/dot/encode.go
32bitkid: I went back and forth on this; and I'm not sure... It felt weird to _not_ have it given that it starts with a common printFrontMatter(), it felt unsymmetrical to unroll it and print a seemingly unmatched `}` here. But, as I think you are implying it feels superfluous to print a single `}`. So, I'm not sure and I'll defer to whatever you think is best.

graph/encoding/dot/decode.go
```diff
@@ -117,8 +162,10 @@ func (gen *generator) node(dst encoding.Builder, id string) graph.Node {
 	return n
 }
 
+type graphGenerator struct{ generator }
```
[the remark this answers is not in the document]
32bitkid: I'm okay with this, especially because they are package private.

graph/encoding/dot/encode.go
kortschak: Can you remove it.

kortschak:
Thanks for doing this.

32bitkid:
Happy to help. Thanks to everybody for all the work they’ve put in.

PR #626, opened 2018-10-21
graph/graphs/gen: use graph construction interfaces correctly

### LINKED ISSUE #602: graphs/gen: TunableClusteringScaleFree does not correctly set up initial state for graph

kortschak:
This makes the simple graph generators use the `NewNode` and `NewEdge` methods rather than always using `simple.Node` and `simple.Edge`. It also makes the (sub)graphs that are generated actually reflect the theoretical statistics. This means that they no longer adds edges to already existing nodes. We may want to add in functions that do do that at some stage (or change the behaviour of what I have here to allow that).

Please take a look.

Fixes #602.


graph/graphs/gen/batagelj_brandes.go
```diff
@@ -290,6 +295,17 @@ func SmallWorldsBB(dst GraphBuilder, n, d int, p float64, src rand.Source) error
 	return nil
 }
 
+// edgeNodesFor returns the pair of nodes for the ith edge in a simple
+// undirected graph. The pair is returned such that the index of w in
+// nodes is less than the index of v in nodes.
+func edgeNodesFor(i int, nodes []graph.Node) (v, w graph.Node) {
+	// This is an algebraic simplification of the expressions described
+	// on p3 of http://algo.uni-konstanz.de/publications/bb-eglrn-05.pdf
+	vi := int(0.5 + math.Sqrt(float64(1+8*i))/2)
```
vladimir-ch: This is a clever mapping. Just a remark.

Issue #616, opened 2018-10-06
graph/{simple,multi}: clean up implementation organisation and do more rigorous testing


kortschak:
The implementations provided in simple and multi have a lot of code duplication, but it has grown rather organically and without consistency between the implementations. This has resulted in a number of occasions where methods that should have been added to all implementations have been missed on some. This has been compounded by the sparsity of testing on those types.

So the files should be harmonised to enable ready comparison between the types and there should be a common test suite that exercises the behaviour we claim, similar to the approach in testblas and testlapack.

### Code structure
- [x] Harmonise the code arrangement; using the godoc order seems the most sensible.

### Testing general graphs
- [x] Return all nodes
- [x] Return correct presence status of non/existing nodes
- [x] Return all edges/lines
- [x] Return correct presence status of non/existing edges/lines
- [x] Return all adjacent nodes relevant to type (`From`, `To`)
- [x] Compare graph weight method with corresponding edge/line weight
In addition, there are some type-specific tests 
- [x] Test matrix input/output from dense graphs

### Testing mutable graphs
- [x] Test node addition
- [x] Test node creation behaviour
- [x] Test node deletion
- [x] Test node deletion causes attached edges to be deleted
- [x] Test edge/line setting
- [x] Confirm simple graphs panic on self-edge setting
- [x] Confirm multi graphs accept self-line setting
- [x] Test edge/line creation behaviour
- [x] Test edge/line deletion
- [x] Test edge/line deletion does not cause edge/line endpoints to be deleted
- [x] Test nodes corresponding to set edges/lines are updated in set edge operation

PR #634, opened 2018-11-05
graph/{multi,simple}: harmonise code organisation

kortschak:
This helps prevent the kind of errors that I found while doing this; the incomplete fix-up when removing the `Has` method.

Review requires no specific graph experience as it's almost entirely just text reordering.

Please take a look.

Updates #616.


kortschak:
Thanks you.

PR #640, opened 2018-11-07
graph: add temporal graph earliest-arrival path algorithm

phyrwork:
Please take a look.

Implentation of one-pass earliest-arrival algorithm described in: https://www.vldb.org/pvldb/vol7/p721-wu.pdf

### DIFF
--- a/graph/iterator/lines.go
+++ b/graph/iterator/lines.go
@@ -4,7 +4,10 @@
 
 package iterator
 
-import "gonum.org/v1/gonum/graph"
+import (
+	"gonum.org/v1/gonum/graph"
+	"sort"
+)
 
 // OrderedLines implements the graph.Lines and graph.LineSlicer interfaces.
 // The iteration order of OrderedLines is the order of lines passed to
@@ -66,6 +69,25 @@ func (e *OrderedLines) LineSlice() []graph.Line {
 // Reset returns the iterator to its initial state.
 func (e *OrderedLines) Reset() {
 	e.idx = -1
+}
+
+// LineStream implements the LineStreamer interface.
+type LineStream struct {
+	lines graph.Lines
+}
+
+// Create a LineStreamer that meets the given ordering specification.
+func NewLineStream(lines []graph.Line, less func(a, b graph.Line) bool) *LineStream {
+	if !sort.SliceIsSorted(lines, func (i, j int) bool {
+		return less(lines[i], lines[j])
+	}) {
+		return nil
+	}
+	return &LineStream{NewOrderedLines(lines)}
+}
+
+func (s *LineStream) LineStream() graph.Lines {
+	return s.lines
 }
 
 // OrderedWeightedLines implements the graph.Lines and graph.LineSlicer interfaces.
--- a/graph/path/earliest.go
+++ b/graph/path/earliest.go
@@ -1,0 +1,110 @@
+// Copyright ©2018 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package path
+
+import (
+	"gonum.org/v1/gonum/graph"
+	"gonum.org/v1/gonum/graph/temporal"
+	"math"
+)
+
+// Earliest stores the earliest-arrival paths
+type Earliest struct {
+	from  graph.Node
+	at    uint64
+	until uint64
+	nodes map[int64]struct{
+		earliest uint64
+		via      []graph.Node
+	}
+}
+
+func (e *Earliest) set(v graph.Node, t uint64, p []graph.Node) {
+	e.nodes[v.ID()] = struct{
+		earliest uint64
+		via      []graph.Node
+	}{
+		t,
+		p,
+	}
+}
+
+func (e *Earliest) From() graph.Node { return e.from }
+
+func (e *Earliest) At() uint64 { return e.at }
+
+func (e *Earliest) Until() uint64 { return e.until }
+
+func (e *Earliest) To(uid int64) (path []graph.Node, duration uint64) {
+	eu, ok := e.nodes[uid]
+	if !ok {
+		return nil, ^uint64(0)
+	}
+	duration = eu.earliest
+	u := temporal.Node(uid)
+	path = append(eu.via, &u)
+	return
+}
+
+func (e *Earliest) Len() int {
+	return len(e.nodes)
+}
+
+func (e *Earliest) Weight(uid, vid int64) float64 {
+	if uid != e.from.ID() {
+		return math.Inf(1)
+	}
+	ev, ok := e.nodes[vid]
+	if !ok {
+		return math.Inf(1)
+	}
+	return float64(ev.earliest - e.at)
+}
+
+// EarliestArrivalFrom computes the earliest-arrival paths to all nodes reachable from u where
+// from <= t <=until. It is an implementation of Algorithm 1. described in
+// https://www.vldb.org/pvldb/vol7/p721-wu.pdf
+func EarliestArrivalFrom(g graph.LineStreamer, from graph.Node, at uint64, until uint64) Earliest {
+	earliest := Earliest{
+		from:  from,
+		at:    at,
+		until: until,
+		nodes: make(map[int64]struct{
+			earliest uint64
+			via      []graph.Node
+		}),
+	}
+	earliest.set(from, at, []graph.Node{})
+	s := g.LineStream()
+	for s.Next() {
+		l := s.Line()
+		u := l.From()
+		uid := u.ID()
+		eu, ok := earliest.nodes[uid]
+		if !ok {
+			continue
+		}
+		var ets, ete uint64
+		if tl, ok := l.(graph.TemporalLine); ok {
+			ets, ete = tl.Interval()
+		} else {
+			// When not dealing with temporal edges degnerates to case
+			// that gives reachability of nodes in the static graph
+			// from u
+			ets, ete = uint64(at), uint64(at)
+		}
+		if eu.earliest <= ets && ete <= until {
+			v := l.To()
+			vid := v.ID()
+			ev, ok := earliest.nodes[vid]
+			if !ok || ete < ev.earliest {
+				earliest.set(v, ete, append(eu.via, u))
+			}
+		} else if ets >= until {
+			break
+		}
+	}
+	return earliest
+}
--- a/graph/path/earliest.go
+++ b/graph/path/earliest.go
@@ -60,7 +60,8 @@ func (e *Earliest) Weight(uid, vid int64) float64 {
 	if !ok {
 		return math.Inf(1)
 	}
-	return float64(ev.earliest - e.at)
+
+	return float64(ev.earliest)
 }
 
 // EarliestArrivalFrom computes the earliest-arrival paths to all nodes reachable from u where
--- a/graph/path/earliest_test.go
+++ b/graph/path/earliest_test.go
@@ -1,0 +1,160 @@
+// Copyright ©2018 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package path
+
+import (
+	"gonum.org/v1/gonum/graph/temporal"
+	"testing"
+	"gonum.org/v1/gonum/graph"
+	"reflect"
+)
+
+var (
+	// sketchStream is the edge stream for the arbitrary
+	// graph I sketched in a notebook to demonstrate that
+	// my method for earliest arrival path construction
+	// worked OK.
+	sketchStream = []struct{
+		F, T int64
+		S, E uint64
+	}{
+		{1, 2, 1, 1},
+		{1, 3, 1, 1},
+		{4, 7, 1, 1},
+		{1, 4, 2, 2},
+		{2, 6, 3, 3},
+		{2, 5, 4, 4},
+		{5, 8, 5, 5},
+		{6, 8, 6, 6},
+		{7, 9, 7, 7},
+		{6, 9, 8, 8},
+	}
+
+	// wuChengHuangKeLuXuStream is the example temporal graph from
+	// figure 1(a) of https://www.vldb.org/pvldb/vol7/p721-wu.pdf
+	wuChengHuangKeLuXuStream = []struct{
+		F, T int64
+		S, E uint64
+	}{
+		{1, 2,  1,  1 },
+		{1, 2,  2,  2 },
+		{7, 10, 2,  2 },
+		{1, 6,  3,  3 },
+		{2, 7,  3,  3 },
+		{2, 8,  3,  3 },
+		{1, 3,  4,  4 },
+		{6, 9,  5,  5 },
+		{3, 8,  6,  6 },
+		{7, 11, 6,  6 },
+		{8, 9,  7,  7 },
+		{9, 12, 8,  8 },
+		{9, 12, 9,  9 },
+		{1, 9,  10, 10},
+	}
+)
+
+func lineAt(fid, tid int64, s, e uint64) temporal.TemporalLine {
+	f := temporal.Node(fid)
+	t := temporal.Node(tid)
+	return temporal.TemporalLine{
+		F: &f,
+		T: &t,
+		S: s,
+		E: e,
+	}
+}
+
+func idsOf(nodes []graph.Node) []int64 {
+	ids := make([]int64, len(nodes))
+	for i := range nodes {
+		ids[i] = nodes[i].ID()
+	}
+	return ids
+}
+
+func TestEarliestArrivalFrom(t *testing.T) {
+	tests := []struct {
+		name string
+		stream []struct{
+			F, T int64
+			S, E uint64
+		}
+		nodes map[int64]struct{
+			earliest uint64
+			path     []int64
+		}
+		from  int64
+		at    uint64
+		until uint64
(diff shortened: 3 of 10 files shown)


graph/.idea/graph.iml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
kortschak: Delete these `.idea` files please.

kortschak:
I use temporal state in graphs with the current API. It may be that there needs some help, but it's not currently impossible to implement the algorithm here with the current API, without expanding the graph first.

I like the idea of adding more temporal support, but I think there needs to be a plan first.

The issues that need to be addressed are where it should live (I don't think temporal graph algorithms should go in the same name spaces as general graphs, so I'd suggest graph/temporal/...) the combinatoric explosion of introducing another edge type in "gonum.org/v1/gonum/graph" is untenable. Either it and its support types/function are unexported and defined in the packages where they are used or if used over a number of packages then exported from graph/temporal. This is the kind of discussion that needs to be had.

Keeping a graph API sane is hard, adding parts is difficult.

phyrwork:
> I use temporal state in graphs with the current API. It may be that there needs some help, but it's not currently impossible to implement the algorithm here with the current API, without expanding the graph first.

How?

> I don't think temporal graph algorithms should go in the same name spaces as general graphs

I don't really see it as all that different from a directed multigraph with some additional traversal logic.

> the combinatoric explosion of introducing another edge type in "gonum.org/v1/gonum/graph" is untenable

Perhaps.

On top of what there already is added in the commit we would end up with additionally WeightedTemporalLine and related interfaces and...?

They would need to go somewhere. I'm not sure separating them makes any more sense than separating multi completely too. Temporal is to directed multi as multi is to simple.

> Keeping a graph API sane is hard, adding parts is difficult.

*strokes beard*

Do you have any other thoughts?

kortschak:
The point of the graph API is to provide generic commonly used graph primitives, but they are provided as interfaces so that additional data and behaviours can be added. In things that I do, I just add those data and behaviours and type assert when they are needed.

The difference between temporal and mult is that mutligraphs are commonly used. Temporal graphs are less so. But maybe others feel that the additional weight in graph is warranted. Hence the need for discussion.

Is there a reason simple and undirected graphs are not temporal? No reason that I can see, so that doubles the number of types and support functions.

I would try to implement the algorithm with the mininum addition to the API and see how it goes. But also post to gonum-dev/post a proposal here for handling of temporal graphs. Unfortunately the number of people available at the moment to discuss graph-related code is limited because of work/study.

/cc @vladimir-ch @mewmew if you have time to consider these things.

kortschak:
In the mean time, I'll review this on the understanding that it may move and may be significantly simplified.

graph/path/earliest.go
```diff
@@ -0,0 +1,100 @@
+package path
```
kortschak: Add copyright header please.

graph/path/earliest.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+}
+
+func (e *Earliest) From() graph.Node { return e.from }
+
+func (e *Earliest) At() uint64 { return e.at }
+
+func (e *Earliest) Until() uint64 { return e.until }
+
+func (e *Earliest) To(u interface{}) (path []graph.Node, duration uint64) {
```
kortschak: Unless it's strictly necessary we don't use `interface{}`. Here it's purely a convenience to let the user not do `To(n.ID())`, so please just make this `uid int64`.

graph/path/earliest.go
```diff
@@ -0,0 +1,100 @@
...
+	var uid int64
```
kortschak: So this type switch goes away and the import of "log" (we don't log in libraries).

graph/path/earliest.go
```diff
@@ -0,0 +1,100 @@
...
+	case graph.Node:
+		uid = t.ID()
+	case int64:
+		uid = t
+	default:
+		log.Panicln("not supported")
+	}
+	if eu, ok := e.nodes[uid]; !ok {
+		return nil, temporal.MaxTime
+	} else {
```
kortschak: Remove redundant else block and deindent.

graph/path/earliest.go
```diff
@@ -0,0 +1,100 @@
+package path
+
+import (
+	"gonum.org/v1/gonum/graph"
+	"log"
+	"gonum.org/v1/gonum/graph/temporal"
+)
+
+type Earliest struct {
```
kortschak: This type needs to be documented so I can see what the intention is.

graph/path/earliest_test.go
```diff
@@ -0,0 +1,160 @@
+package path
```
kortschak: Copyright header.

graph/path/earliest_test.go
```diff
@@ -0,0 +1,160 @@
...
+
+import (
```
kortschak: Please use goimports throughout.

graph/path/earliest_test.go
```diff
@@ -0,0 +1,160 @@
...
+			}
+			for vid, ev := range test.nodes {
+				path, earliest := out.To(vid)
+				if path == nil {
+					t.Fatalf("expected reachable node: %v", vid)
+				}
+				if earliest != ev.earliest {
+					t.Fatalf("expected earliest arrival time: (%v) expected %v, got %v", vid, ev.earliest, earliest)
+				}
+				if len(path) != len(ev.path) {
```
kortschak: Use `reflect.DeepEqual` instead of `goto`.

graph/temporal.go
```diff
@@ -0,0 +1,17 @@
+package graph
```
kortschak: Copyright header.

graph/temporal.go
```diff
@@ -0,0 +1,17 @@
...
+
+	// Duration is the discrete time which elapses while traversing the line
+	Duration() uint64
+}
+
+type TemporalStream interface {
+	// LineStream returns a TemporalLine iterator ordered first by start
+	// time and then by duration.
+	LineStream() TemporalLines
+}
```
kortschak: gofmt

graph/temporal.go
```diff
@@ -0,0 +1,17 @@
+package graph
+
+type TemporalLine interface {
```
kortschak: This type needs thought (and a doc comment).

From what I can see, `At` and `Duration` (more below) are always called together, so they could be merged to a single method, `Interval() (start, end T)`(?) reducing the interface method dispatch cost.

`Duration` is suggestive that it would return a `time.Duration`. So a different name would be better even without the above change. Alternatively, `T` above is made a `time.Duration` or `time.Time`.

graph/temporal.go
```diff
@@ -0,0 +1,17 @@
...
+	Line
```
kortschak: The `Line.ID` method is never used here, so this could actually be `Edge`.

graph/temporal/temporal.go
```diff
@@ -0,0 +1,40 @@
+package temporal
```
kortschak: Copyright header.

graph/temporal/temporal.go
```diff
@@ -0,0 +1,40 @@
...
+
+import "gonum.org/v1/gonum/graph"
+
+const MaxTime = ^uint64(0)
```
kortschak: This doesn't need to be exported (or even provided).

graph/temporal/temporal.go
```diff
@@ -0,0 +1,40 @@
...
+func (l TemporalLine) To() graph.Node { return l.T }
+
+// ID returns the ID of the line.
+func (l TemporalLine) ID() int64 { return l.UID }
+
+// At returns the start time of the line.
+func (l TemporalLine) At() uint64 { return l.A }
+
+// During returns the traversal time of the line.
+func (l TemporalLine) Duration() uint64 { return l.D }
```
kortschak: gofmt.

graph/path/earliest.go
```diff
@@ -0,0 +1,100 @@
...
+}
+
+func (e *Earliest) Len() int {
+	return len(e.nodes)
+}
+
+// EarliestArrivalFrom computes the earliest-arrival paths to all nodes reachable from u where
+// from <= t <=until. It is an implementation of Algorithm 1. described in
+// https://www.vldb.org/pvldb/vol7/p721-wu.pdf
+func EarliestArrivalFrom(g graph.TemporalStream, from graph.Node, at uint64, until uint64) Earliest {
```
kortschak: This could just be `EarliestArrivalFrom(g graph.Graph, from graph.Node, at uint64, until uint64) Earliest` and then type switch on the capacity to get a temporal stream, defaulting to [-inf,inf] temporal intervals when unable to do that.

graph/path/earliest.go
```diff
@@ -0,0 +1,100 @@
...
+	earliest.set(from, at, []int64{})
+	s := g.LineStream()
+	for s.Next() {
+		l := s.TemporalLine()
+		u := l.From()
+		uid := u.ID()
+		eu, ok := earliest.nodes[uid]
+		tl := l.At()
+		dtl := l.Duration()
+		if !ok {
```
kortschak: This should happen prior to the calls to `At` and `Duration`; they are discarded if `!ok`.

graph/path/earliest.go
```diff
@@ -0,0 +1,100 @@
...
+			ev, ok := earliest.nodes[vid]
+			if !ok || tl + dtl < ev.earliest {
+				earliest.set(v, tl + dtl, append(eu.via, uid))
+			}
+		} else if tl >= until {
+			break
+		}
+	}
+	return earliest
+}
```
kortschak: gofmt

graph/path/earliest.go
```diff
@@ -0,0 +1,100 @@
...
+		l := s.TemporalLine()
+		u := l.From()
+		uid := u.ID()
+		eu, ok := earliest.nodes[uid]
+		tl := l.At()
+		dtl := l.Duration()
+		if !ok {
+			continue
+		}
+		if tl + dtl <= until && tl >= eu.earliest {
```
kortschak: `if eu.earliest <= tl && tl + dtl <= until {`

graph/iterator/lines.go
```diff
@@ -129,3 +132,87 @@ func (e *OrderedWeightedLines) WeightedLineSlice() []graph.WeightedLine {
...
+		return true
+	}
+	if e.lines[i].Duration() < e.lines[j].Duration() {
+		return true
+	}
+	return false
+}
+
+func (e *OrderedTemporalLines) LineStream() graph.TemporalLines {
+	stream := make([]graph.TemporalLine, len(e.lines))
```
kortschak: I don't understand why this needs a copy. If you want to lazily sort, then just sort lazily on the unused edges. Otherwise just make it a requirement that the edges be sorted and do a `sort.IsSorted` when `New...` is called. If the latter is done then you can guarantee that the edges will always come out in order and this method can go away.

phyrwork:
> Is there a reason simple and undirected graphs are not temporal? No reason that I can see, so that doubles the number of types and support functions.

Yes. A temporal edge *must* be directed because time has direction and magnitude, even if that magnitude is zero.

If you consider a regular undirected edge to be infinitely traversible between the nodes it connects then an "undirected" temporal edge would have to have zero duration. The fact that you could not give this edge arbitrary (discrete non-zero) duration would make it by definition not a temporal edge.

graph/path/earliest.go
phyrwork: What? No.

Even with [-inf,inf] time constraint the edges still need to have start time and otherwise it's not earliest-arrival so much as some sort of weird DFS based on in what order the edges are provided.

kortschak:
>Yes. A temporal edge must be directed because time has direction and magnitude, even if that magnitude is zero.

Ah, OK. So this is a different definition to a temporal graph than what I was using; a dynamic graph where edges exist within a set duration but may not necessarily have a direction in the period. This is where things get complicated. This adds to the argument that we need discussion to explore what's actually wanted.

graph/path/earliest.go
kortschak: The paper says that co-occurrent edges can be arbitrarily ordered. That's what this does.

phyrwork: No. It is setting the time bounds of the search, which has no effect on the start times of the edges. It only affects the scope of which edges may be traversed.

kortschak: In the case that no temporal information is know, the graph is a `graph.Directed` only, then the algorithm degenerates to the case I am talking about. Sure, it's not necessarily useful, but I'm considering the API here as well. If this were all in temporal, I'd be pushing back less on this.

The gains that the algorithm gives you depend on a specified order. We have other algorithms here that have similar invariants and they either do the work themselves or require that the graph be handed to them that way (this varies). Here we have a time ordering invariant, but the graph itself does not do that work, instead it's done lazily by the `OrderedTemporalLines`. This means that (unless you are planning on changing that behaviour), it could be done initially by `EarliestArrivalFrom` after obtaining the list of edges. There is the possibility that you are considering the use of a non-memory edge store though (that breaks the assumption that we can fit all the edges into memory). If that's the case, then this is reasonable.

kortschak:
A point to consider here is that if the container for paths you use for holding paths has a `Weight(uid, vid int64) float64` (where the returned `float64` is the path duration or distance depending on which you are finding), then you get the closeness centrality for free if we change the signatures there for an interface instead of the current `path.AllShortest`.

phyrwork:
I understand the desire to encourage code reuse, but I'm not keen on the idea of float64 time.

All of the literature I'm familiar with deals with discrete time. Making an erroneous choice to traverse (or not) an edge due to a rounding error could have significant implications compared to, say, having two (e.g. shortest distance) paths in a static with slightly incorrect but very similar costs.

kortschak:
I'm not suggesting that the time for the edges be float, but container that holds the result can give time between nodes in `float64`. This is purely so that the distance [centrality code](https://github.com/gonum/gonum/blob/48288cca5b5efb2c3c3b0d8b48a6b01b37831bf8/graph/network/distance.go#L14-L41) can be used on the  results here (you'll find it difficult to calculate closeness centrality with ints).

graph/path/earliest_test.go
phyrwork: `reflect.DeepEqual` does not work. (array of interfaces vs. array of values)

graph/temporal.go
phyrwork: But given that temporal extends directed multi, it should satisfy the `Line` interface for consistency, no?

graph/path/earliest.go
phyrwork: > In the case that no temporal information is know, the graph is a graph.Directed only, then the algorithm degenerates to the case I am talking about.

Yes, after another look this is clear now.

> Here we have a time ordering invariant, but the graph itself does not do that work, instead it's done lazily by the OrderedTemporalLines

My intention is that the `TemporalStream` implementor will provide an already-ordered iterator for the reason you mentioned.

Having `OrderedTemporalLines` do this is
- for convenience, no point for the user to reimplement this.
- lazily, with a copy, for immutability. The type name suggests the order is important and there is no other means to change it for any of the 'Ordered' types.

graph/temporal/temporal.go
```diff
@@ -0,0 +1,40 @@
+package temporal
+
+import "gonum.org/v1/gonum/graph"
+
+const MaxTime = ^uint64(0)
+
+// Node here is a duplication of simple.Node
```
kortschak: This is not needed.

graph/path/earliest_test.go
kortschak: Either make `ev.path` a `[]simple.Node` or make a helper to give you a slice of IDs from a `[]graph.Node`.

I see that you retain only int64 in the `Earlier`. We don't do this. We keep the nodes since they may be used by the client for annotation.

graph/path/earliest.go
kortschak: Yeah, don't copy here.

I've been thinking about this over the day and ISTM that all we need to add is a single interface (`StreamGraph interface { graph.DirectedMultigraph; Stream() graph.Lines }`) in a new package temporal which would also hold the actual functionality, allow with the container types. It's arguable that even an extra iterator is not really required, but that can be explored (`Stream` either returns the new type or has documentation indicating that the `graph.Lines` must return `graph.Line`s that can be type asserted to the `temporal.Line` type).

phyrwork: SGTM, however

> a single interface (StreamGraph interface { graph.DirectedMultigraph; Stream() graph.Lines }) in a new package temporal

Maybe it's overkill to require implementation of `graph.DirectedMultigraph` when we already know we can run algorithms only with the `Stream()` representation of the graph.

And for this reason I think that the interface we want is `LineStreamer interface { LineStream() graph.Lines }`

kortschak: That seems fair.

kortschak:
Yes, that depends on whether those measures are actually useful though. In the first instance they are not needed, but simply implementing the pairwise weight should be reasonably easy to implement.

kortschak:
Yes, that's understood. The paper does not mention betweenness centrality and I don't think an implementation is going to be trivial. I think that is future work if at all.

phyrwork:
Ignore latest - WIP

kortschak:

phyrwork:
It’s on hold until I have time to look at it again.

vladimir-ch:

PR #645, opened 2018-11-12
blas/gonum: remove check helpers

vladimir-ch:
Please take a look.


vladimir-ch:
Oh, the mat package tests against exact panic message text from blas. That's unexpected.

kortschak:
The check there being what it was was to distinguish between our panic and a possible runtime panic. So to guard against future change-causing problems, either the check in mat could be changed to 
```
		if !panicked || !strings.HasPrefix(message, "blas:") {
			if message != "" {
				t.Errorf("expected blas panic for nil data slice: got %q", message)
			} else {
				t.Error("expected blas panic for nil data slice")
			}
		}
```
or make a check that the panic is not a `runtime.Error`. I think the former is easier given what's already written.

Or just update the expected string.

blas/gonum/dgemm.go
```diff
@@ -21,25 +21,75 @@ import (
...
+	if ldc < max(1, n) {
+		panic(badLdC)
+	}
+
+	// Quick return if possible.
+	if m == 0 || n == 0 || ((alpha == 0 || k == 0) && beta == 1) {
+		return
+	}
+
+	if aTrans {
```
kortschak: Can you add a comment to the cases where you have moved checks to after quick returns explaining that the failure is not possible for the quick return case (with nicer words than I have access to right now obviously).

Also for other commits where these checks have moved.

vladimir-ch: As if I knew how to say it with nicer words :-)

But looking at the code, I think I will change some of the quick return checks before merging. Using `Zgbmv` as an example, I'll change it to:
```
if m == 0 || n == 0 {
    return
}
// Length checks
if alpha == 0 && beta == 1 {
    return
}

**Comment by vladimir-ch** [comment]
> Or just update the expected string.

I just updated the expected string.

File: blas/gonum/dgemm.go (hunk 1 above)
Comment (vladimir-ch) [reply blas/gonum/dgemm.go]: How about
```
// For zero matrix size the following slice length checks are trivially satisfied.
```
?

kortschak: SGTM

kortschak: >As if I knew how to say it with nicer words :-)

You do the best documentation.

vladimir-ch:
PTAL

vladimir-ch:
Do we want to change any of the panic messages in errors.go?

kortschak:
They all seem OK to me.

PR #654, opened 2018-11-20
mat: make Diagonal satisfy Symmetric

### LINKED ISSUE #610: mat: Banded interfaces should require satisfying other matrix interfaces
For example any type that implements `SymBanded` should also be required to implement `Symmetric`, and similarly, any type that implements `Diagonal` should also implement `Symmetric`.

The argument for this is to allow the same flexibility that the `Matrix` interface allows. `Symmetric` matrices also implement `Matrix` by requirement, which allows them to be used anywhere that accepts a `Matrix` as an argument. As of right now, that flexibility is limited to `Matrix`, it doesn't extend to the other (sub) types. This means, for example, you cannot use a `Diagonal` as a `Symmetric`, only `DiagDense`.

An example of a useful algorithm is the following, pretending that `mat.SymDense` has the method `DiagView() Diagonal` (which is the natural design).

```
// ProductGaussian returns a product-distribution from the given Gaussian
// where all of the components are independent.
func ProductGaussian(norm *distmv.Normal) {
  mu := norm.Mean(nil)
  cov := norm.Covariance(nil)
  n, _ := distmv.NewNormal(mu, cov.DiagView(), nil)
  return n
}
```

This function would not compile, because `Diagonal` does not implement `Symmetric`, even though all `Diagonal` matrices are symmetric. I don't think this requirement comes with significant cost, just a few extra methods for alternative implementations.

kortschak:
Please take a look.

As far as I can see, this is the only addition that is needed to satisfy the linked issue (the `Symmetric` interface is already satisfied by symmetric banded implementation, and there is no single interface for that).

Fixes #610.

### DIFF
--- a/mat/diagonal.go
+++ b/mat/diagonal.go
@@ -25,8 +25,11 @@ var (
 // has non-zero terms on the diagonal.
 type Diagonal interface {
 	Matrix
-	// Diag returns the number of rows/columns in the matrix
+	// Diag and Symmetric return the number of rows/columns in
+	// the matrix. Both methods are included to allow diagonal
+	// matrices to be used in functions taking symmetric inputs.
 	Diag() int
+	Symmetric() int
 }
 
 // MutableDiagonal is a Diagonal matrix whose elements can be set.


vladimir-ch:
If there is `MutableSymBanded` interface, should there also be `SymBanded`?

kortschak:
I'll add it.

PR #655, opened 2018-11-20
mat: disallow New calls with zero length

### LINKED ISSUE #442: mat: No panic on zero sized matrices

kortschak:
Please take a look.

I'm still ambivalent about this, but here it is if we want it.

Fixes  #442.


mat/band.go
```diff
@@ -126,7 +127,10 @@ func (t TransposeBand) UntransposeBand() Banded {
 // which is passed to NewBandDense as []float64{*, 1, 2, 3, 4, ...} with kl=1 and ku=2.
 // Only the values in the band portion of the matrix are used.
 func NewBandDense(r, c, kl, ku int, data []float64) *BandDense {
-	if r < 0 || c < 0 || kl < 0 || ku < 0 {
+	if r <= 0 || c <= 0 || kl < 0 || ku < 0 {
```
sbinet: I think I'd go with less indentation:
```go
switch {
case r == 0 || c == 0:
    panic(ErrZeroLength)
case r < 0 || c < 0 || kl < 0 || ku < 0:
    panic("mat: negative dimension")
}
```

(same for the other nested `if`s)
but ok.

kortschak: The choice to use repeated checks was made to reduce compares on the fast path. The use of `switch` would undo that.

btracey:
The main use I've needed for zero-sized matrices is as a placeholder for concatenating data. Something like

```
m := mat.NewDense(0, c)
for { 
  // Get new row of data
  // Stack row with m
}
```
Where using `for` because it's some unknown number of rows total. It's definitely possible to work around, but zero-sized matrices nicely solve the loop-and-a-half problem

kortschak:
Could we not allow `Stack` to correctly stack with at most one zero matrix input? Then the code becomes
```
var m Dense
for {
  // Get new row of data
  m.Stack(&m, data)
}
```

Same for `Augment`.

btracey:
Yes, I think `Stack` could be fixed to allow that. That's what I intended to say above, though re-reading my comment I realize it's not really english

kortschak:
I'm now confused. I have just gone through making changes to do this, but `Stack` and `Augment` are size constrained operations. Are you doing a `Grow` for each row you add? Can you post code rather than pseudo-code?

btracey:
Yes, I was doing a Grow (so maybe it's Grow and not Stack that was where I wanted it). I appear to have removed the code that was using it, so I can't point to it directly.

kortschak:
But Grow is valid on the zero value.

PR #657, opened 2018-11-20
lapack/gonum: use crossover point for blocked/unblocked code; clean up parameter checks

### LINKED ISSUE #83: lapack/gonum: dgebrd.go has ineffectual assignement
This is something that I left out of #78.

dgebrd.go has a report of ineffasign for the `nx` variable [here](https://github.com/gonum/gonum/blob/master/lapack/gonum/dgebrd.go#L88-L105). This appears to be because of change to the choice in blocking choice parameters.

It seems to me that this could be rewritten as
```
        if nb > 1 && nb < minmn {
                if max(nb, impl.Ilaenv(3, "DGEBRD", " ", m, n, -1, -1)) < minmn {
                        ws = (m + n) * nb
                        if lwork < ws {
                                nbmin := impl.Ilaenv(2, "DGEBRD", " ", m, n, -1, -1)
                                if lwork >= (m+n)*nbmin {
                                        nb = lwork / (m + n)
                                } else {
                                        nb = minmn
                                }
                        }
                }
        }
```

vladimir-ch:
This got bigger than expected. The starting point was checking the list of functions from #83:
* Dgebrd
* Dorgql
* Dsytrd
* Dgeqrf
* Dgerqf
* Dorgqr
* Dgehrd
* Dorglq
* Dgelqf
that call `Ilanenv(3, ...)` and sometime chose to ignore the returned value.

While at that I cleaned up the parameter checks. Doing this
* made the checks stricter for workspace query calls and so broke some tests which I had to fix,
* revealed some inconsistencies, especially regarding `work[0]` value upon return in case of _non_-workspace query call. Some functions return `lwkopt` while other compute a separate `iws` value. I changed the code to follow the reference but we can either ask the reference about this or just use `lwkopt` everywhere.

In general, I think that the following order of initial checks is the most reasonable:
1. Check numerical parameters like `uplo`, `trans`, `m`, `n`, `k`, `lda`, `lwork` and the length of `work`.
2. Quick return if possible with setting `work[0] = 1`.
3. Compute optimal work length and return if query call
4. Check slice lengths

I could be convinced to move leading dimension checks to 4, or also in the opposite direction to move slice length checks to 1 although it might not be a good idea to be so strict as exemplified by Dorgqr.

Unlike with the other functions, in case of Dorgqr I had to move the lda check after the workspace query return. The reason is Dgesvd. Dorgqr is queried at several places in Dgesvd for workspace size with varying arguments. It is called as `Dorgqr(m, n, ..., a, lda, ..., -1)` and also as `Dorgqr(m, m, ..., a, lda, ..., -1)`. 
This is obviously impossible to pass. The reference gets around this because it's column-major
and the number of rows is the same for each call. I think that it would be more correct to call it in the second case with `u, ldu` but then the test for Dlapll fails because it does not need `U` and passes `nil, 0` for it.

Eventually we could delete `checkMatrix` also here in lapack.

The unified diff is big but the individual commits are focused.

Fixes #83 

Please take a look.

### DIFF
--- a/lapack/gonum/dgebrd.go
+++ b/lapack/gonum/dgebrd.go
@@ -53,46 +53,62 @@ import (
 //
 // Dgebrd is an internal routine. It is exported for testing purposes.
 func (impl Implementation) Dgebrd(m, n int, a []float64, lda int, d, e, tauQ, tauP, work []float64, lwork int) {
-	checkMatrix(m, n, a, lda)
-	// Calculate optimal work.
-	nb := impl.Ilaenv(1, "DGEBRD", " ", m, n, -1, -1)
-	var lworkOpt int
-	if lwork == -1 {
-		if len(work) < 1 {
-			panic(badWork)
-		}
-		lworkOpt = ((m + n) * nb)
-		work[0] = float64(max(1, lworkOpt))
+	switch {
+	case m < 0:
+		panic(mLT0)
+	case n < 0:
+		panic(nLT0)
+	case lda < max(1, n):
+		panic(badLdA)
+	case lwork < max(1, max(m, n)) && lwork != -1:
+		panic(badWork)
+	case len(work) < max(1, lwork):
+		panic(shortWork)
+	}
+
+	// Quick return if possible.
+	minmn := min(m, n)
+	if minmn == 0 {
+		work[0] = 1
 		return
 	}
-	minmn := min(m, n)
-	if len(d) < minmn {
+
+	nb := impl.Ilaenv(1, "DGEBRD", " ", m, n, -1, -1)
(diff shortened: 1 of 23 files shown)


kortschak:
For another PR. I think it would be nice to have meaningful names for the ILAENV ispec constants.

vladimir-ch:
Any opinion about the `lwktop`/`iws` question?

btracey:
I forgot about this issue. `Ilaenv` is a strange function and doesn't seem to make sense (not adding anything new, just re-inventing past discussions). 

Why is check matrix bad? Seems like it can cut down on a bunch of boilerplate

PR #659, opened 2018-11-20
mat: add SymBanded type

kortschak:
Please take a look.

### DIFF
--- a/mat/symband.go
+++ b/mat/symband.go
@@ -30,18 +30,10 @@ type SymBandDense struct {
 
 // SymBanded is a symmetric band matrix interface type.
 type SymBanded interface {
-	Matrix
+	Banded
 
 	// Symmetric returns the number of rows/columns in the matrix.
 	Symmetric() int
-
-	// Bandwidth returns the lower and upper bandwidth values for
-	// the matrix. The total bandwidth of the matrix is kl+ku+1.
-	Bandwidth() (kl, ku int)
-
-	// TBand is the equivalent of the T() method in the Matrix
-	// interface but guarantees the transpose is of banded type.
-	TBand() Banded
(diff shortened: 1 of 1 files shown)


mat/symband.go
```diff
@@ -14,6 +14,7 @@ var (
 	_            Matrix           = symBandDense
 	_            Symmetric        = symBandDense
 	_            Banded           = symBandDense
+	_            SymBanded        = symBandDense
```
btracey: Diagonal should also have the methods added to satisfy this

mat/symband.go
```diff
@@ -27,6 +28,22 @@ type SymBandDense struct {
 	mat blas64.SymmetricBand
 }
 
+// SymBanded is a symmetric band matrix interface type.
+type SymBanded interface {
```
btracey: Wouldn't this be better expressed as `Banded` and then the extra `Symmetric`?

btracey: (in terms of the expression of the interface methods)

mat/symband.go
kortschak: I think that's another PR.

mat/symband.go
kortschak: I couldn't decide which way that should go, so I avoided it. I guess this way is the lesser number of methods literally present. I'll change it.

kortschak: Though it's at odds with how the mutable type is expressed.

btracey: Is it? MutableBanded is Banded plus a Mutable method

kortschak: `MuatbleSymBanded` is `Symmetric`+some methods. This is `Banded`+some methods.

PR #660, opened 2018-11-21
mat: add Banded methods to Diagonal

kortschak:
Please take a look.

I'm not at all convinced by this. The concrete type already implements the methods. It seems like interface bloat. Are you sure you want this?

### DIFF
--- a/mat/diagonal.go
+++ b/mat/diagonal.go
@@ -16,6 +16,7 @@ var (
 	_         MutableDiagonal = diagDense
 	_         Triangular      = diagDense
 	_         Symmetric       = diagDense
+	_         SymBanded       = diagDense
 	_         Banded          = diagDense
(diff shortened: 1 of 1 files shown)


btracey:
I think it is a good idea, but that doesn't mean I'm right, and it doesn't mean it has to happen now.

I understand it feels like a lot of methods, but the question is if one should be able to use a Diagonal anywhere one uses a {Triangular, Banded, Symmetric, etc.). It feels to me like you should, as all Diagonal are those other things. If Diagonal does not implement those other interfaces, then you have to cast to a concrete type in order to pass in onward, which in turn requires knowing the concrete type. If all of the more specialized types implement the less specialized types, then it all just works.

kortschak:
That seems fair. Can you approve.

PR #661, opened 2018-11-21
mat: add striding to DiagDense and capacity to extract diags from other matrices

### LINKED ISSUE #213: mat: get diagonal of a matrix
as per discussion with @btracey on slack, it'd very handy to be able to extract the diagonal of a matrix for e. g. extracting the variance from a varcov matrix (cf. `numpy.diag` when operating on a matrix)

possibly as a view directly on the matrix:

```go
func (m *Dense) DiagView() SymBandDense {}
```

or

```
func (m *Dense) DiagView() Dense {}
```

or

```go
func (s *SymBandDense) GetDiag(a Matrix)
```

### LINKED ISSUE #609: mat: Diagonal should be strided
This would allow for `Diag.DiagView()`, as an example.

kortschak:
Please take a look.

I'm not really all that happy with the approach in `DiagOf` because of the mixing of view and copy, but the alternatives that I can see are less attractive; have a collection of methods for each raw type, either on `DiagDense` or (less likely) the raw types, or alternatively panicking when a non-raw type is handed in.

Fixes #609.
Updates #213.

### DIFF
--- a/mat/diagonal.go
+++ b/mat/diagonal.go
@@ -121,62 +121,63 @@ func (d *DiagDense) Reset() {
 	d.mat.Data = d.mat.Data[:0]
 }
 
-// DiagOf extracts the diagonal of m into the receiver. If m can return
-// its raw representation d will be a view of the diagonal of m. Otherwise
-// the receiver will be a copy of the input's diagonal.
-// The receiver must be min(r, c) long or zero. Otherwise DiagOf will panic.
-func (d *DiagDense) DiagOf(m Matrix) {
-	// reuseAs is not used here since
-	// sensible cases should not allocate,
-	// but would uselessly if we did.
+// DiagFrom copies the diagonal of m into the receiver. The receiver must
+// be min(r, c) long or zero. Otherwise DiagOf will panic.
+func (d *DiagDense) DiagFrom(m Matrix) {
 	n := min(m.Dims())
-	if !d.IsZero() && n != d.n {
-		panic(ErrShape)
-	}
-	d.n = n
+	d.reuseAs(n)
 
+	var vec blas64.Vector
 	switch r := m.(type) {
 	case *DiagDense:
-		d.mat = r.mat
+		vec = r.mat
 	case RawBander:
 		mat := r.RawBand()
-		d.mat = blas64.Vector{
+		vec = blas64.Vector{
 			Inc:  mat.Stride,
 			Data: mat.Data[mat.KL : (n-1)*mat.Stride+mat.KL+1],
 		}
 	case RawMatrixer:
 		mat := r.RawMatrix()
-		d.mat = blas64.Vector{
+		vec = blas64.Vector{
 			Inc:  mat.Stride + 1,
 			Data: mat.Data[:(n-1)*mat.Stride+n],
 		}
 	case RawSymBander:
 		mat := r.RawSymBand()
-		d.mat = blas64.Vector{
+		vec = blas64.Vector{
 			Inc:  mat.Stride,
 			Data: mat.Data[:(n-1)*mat.Stride+1],
 		}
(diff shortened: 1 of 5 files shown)


mat/diagonal.go
```diff
@@ -85,41 +87,150 @@ func (d *DiagDense) TTri() Triangular {
...
+		d.n = r
+		return
+	}
+	if r != d.n {
+		panic(ErrShape)
+	}
+}
+
+// IsZero returns whether the receiver is zero-sized. Zero-sized vectors can be the
+// receiver for size-restricted operations. VecDenses can be zeroed using Reset.
```
vladimir-ch: `DiagDenses`

mat/diagonal.go
```diff
@@ -40,7 +40,8 @@ type MutableDiagonal interface {
 
 // DiagDense represents a diagonal matrix in dense storage format.
 type DiagDense struct {
-	data []float64
+	mat blas64.Vector
```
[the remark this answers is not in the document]
kortschak: The convention throughout is that the blas64 field is called `mat`.

mat/diagonal.go
```diff
@@ -60,18 +61,19 @@ func NewDiagonal(n int, data []float64) *DiagDense {
 		panic(ErrShape)
 	}
 	return &DiagDense{
-		data: data,
+		mat: blas64.Vector{Inc: 1, Data: data},
```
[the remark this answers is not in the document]
kortschak: Changed.

mat/diagonal.go
kortschak: Fixed.

mat/diagonal.go
```diff
@@ -85,41 +87,150 @@ func (d *DiagDense) TTri() Triangular {
...
+// See the Reseter interface for more information.
+func (d *DiagDense) Reset() {
+	// No change of Inc or n to 0 may be
+	// made unless both are set to 0.
+	d.mat.Inc = 0
+	d.n = 0
+	d.mat.Data = d.mat.Data[:0]
+}
+
+// DiagOf extracts the diagonal of m into the receiver. If m can return
```
btracey: IMO, this is wrong. This should be `DiagFrom`, and copy the result into the receiver. This works the same for any `Matrix`. Then, on `Dense`, `Symmetric`, etc. there should be a `func (d *Dense) DiagView() Diagonal` method that returns a view of the diagonal.

mat/diagonal.go
```diff
@@ -85,41 +87,150 @@ func (d *DiagDense) TTri() Triangular {
...
+		d.mat.Data = d.mat.Data[:1]
+	default:
+		d.reuseAs(n)
+		for i := 0; i < n; i++ {
+			d.setDiag(i, m.At(i, i))
+		}
+	}
+}
+
+// RawBand returns the underlying data used by the receiver represented
```
btracey: I know this function is already here, but is it a good idea? Currently in the addition types (or whatever), we switch on the type by which `Raw` it can implement. If methods implement multiple `RawBand` types, then we should make sure to order from the most specific to the least specific. Given that there's nothing wrong with implementing this method, I tend to lean toward fixing our ordering, but it's worth asking about.

kortschak: Agreed. We can have a note in the package level documentation about this.

I don't think most specific to least specific if the correct ordering though (Is there such an ordering in the general case? - I'm not sure there is). The ordering I think should exist is one that maintains correctness and gives the best performance available.

mat/diagonal.go
kortschak: No worries. I'll change this.

kortschak:
PTAL

kortschak:
I'm changing this to updates #213 since we don't have the view option yet.

Issue #494, opened 2018-05-16
testlapack: add implementation comments to some tests


vladimir-ch:
In https://github.com/gonum/gonum/pull/465#discussion_r186274052 it was suggested that the clarity of the following LAPACK tests could be improved by adding some implementation comments to them:

- [x] Dgebd2
- [x] Dgelqf
- [x] Dgeqp3
- [x] Dlags2
- [x] Dlange
- [x] Dlanst
- [x] Dlansy
- [x] Dlapll
- [x] Dlartg
- [x] Dlasr
- [x] Dlatrd
- [x] Dorg2r
- [x] Dorgtr
- [x] Dtrti2
- [x] Dtrtri
- [x] Dlaln2
- [x] Dlanv2
- [x] Dlaqr23
- [x] Dlarfx
- [ ] Dpotf2
- [ ] Dorgbr

vladimir-ch:
I have a PR ready which adds comments to the remaining tests except for `Dpotf2` and `Dorgbr`. I'd like to skip those two, it's too onerous to add comments to those two, I'd rather rewrite them but that would be a bigger task for a separate ticket.

kortschak:
OK

PR #689, opened 2018-11-23
lapack/testlapack: add implementation comments to some tests

vladimir-ch:
Updates #494 

Please take a look.


vladimir-ch:
Could you please reapprove? I have pushed changes for another test without noticing your review.

Issue #692, opened 2018-11-24
mat: triangular matrices don't respect the .mat.Diag field


kortschak:
The `.mat.Diag` field of the triangular types is used by nearly none of the operations in mat (one internal method checks it to panic and the new diagonal extraction method on `Diagonal` allows unit diagonals to be obtained (FWIW). However indexing a `TriDense` for example does not check `t.mat.Diag`. This should probably be done (although there is currently no way for a user to set the state to unit, that may happen).

btracey:
I don't think there's any pressing need to expose the `Diag` field as part of the interface. The reason to have the field is so you can put two triangles together in one Dense, or to put a triangular and a symmetric together. That's a big part of why BLAS works, but it's somewhat counter the spirit of `mat`. For instance, it's impossible in general to maintain the `Diag` property when the `TriDense` is a receiver for most operations. That said, we support `RawTriangular`, and so we should respect the full interface. This would also enable a user to have a type for this kind of storage (that implements both `RawTriangular` and `RawSymmetric` for instance.

If that's the policy, presumably that also means we should be supporting `Lower` symmetric matrices as well?

kortschak:
The issue here is to ensure that the check is done correctly if the `Diag` field is set differently to how the package corrently does by default. As I wrote above, this is not urgent since there is no safe way for a user to do that contrary setting of `Diag` since we don't have a raw setter for these types. However, when we do, this will have to happen.

BTW, storing two triangles|symmetics in a dense is potentially hazardous if one is a parameter and one is a receiver since that will fail a shadow check.

kortschak:
By "receiver" what do you mean?

The reason for having a raw setter is to allow users to fill in areas of BLAS API that we don't cover. This was an integral part of the API design from the start. Shadow detection prevents egregiously unsafe uses, but there are safe uses that are made available by being able to do this.

Note that with a raw setter, once the `blas64.Triangular` or `blas64.TriangularBand` has been handed in, the `Diag` field is not mutable externally, so the maintenance issue is not a problem.

btracey:
By receiver, I mean something like
```
t.FromRawTriangular(raw)
t.AddTri(t1, t2)
```
It is impossible to maintain `t` as a Diag in this case. If we want to support `Diag=true`, then we have to add in many panic checks. I do not think we want to go the route of "unless t1 and t2 have zeros on the diagonal"

kortschak:
[...]

PR #693, opened 2018-11-24
mat: Add TriBanded interface and TriBandDense type

btracey:
Please take a look.

### DIFF
--- a/mat/diagonal.go
+++ b/mat/diagonal.go
@@ -15,6 +15,7 @@ var (
 	_         Diagonal        = diagDense
 	_         MutableDiagonal = diagDense
 	_         Triangular      = diagDense
+	_         TriBanded       = diagDense
 	_         Symmetric       = diagDense
 	_         SymBanded       = diagDense
 	_         Banded          = diagDense
@@ -89,7 +90,7 @@ func (d *DiagDense) T() Matrix {
 }
 
 // TTri returns the transpose of the matrix. Note that Diagonal matrices are
-// Upper by default
+// Upper by default.
 func (d *DiagDense) TTri() Triangular {
 	return TransposeTri{d}
 }
@@ -98,6 +99,12 @@ func (d *DiagDense) TTri() Triangular {
 // TransposeBand.
 func (d *DiagDense) TBand() Banded {
 	return TransposeBand{d}
+}
+
+// TTriBand performs an implicit transpose by returning the receiver inside a
+// TransposeTriBand. Note that Diagonal matrices are Upper by default.
+func (d *DiagDense) TTriBand() TriBanded {
+	return TransposeTriBand{d}
 }
 
 // Bandwidth returns the upper and lower bandwidths of the matrix.
@@ -111,9 +118,19 @@ func (d *DiagDense) Symmetric() int {
 	return d.n
 }
 
+// SymBand returns the number of rows/columns in the matrix, and the size of
+// the bandwidth.
+func (d *DiagDense) SymBand() (n, k int) {
+	return d.n, 0
+}
+
 // Triangle implements the Triangular interface.
 func (d *DiagDense) Triangle() (int, TriKind) {
 	return d.n, Upper
+}
+
+func (d *DiagDense) TriBand() (n, k int, kind TriKind) {
+	return d.n, 0, Upper
 }
 
 // Reset zeros the length of the matrix so that it can be reused as the
--- a/mat/index_no_bound_checks.go
+++ b/mat/index_no_bound_checks.go
@@ -236,6 +236,65 @@ func (s *SymBandDense) set(i, j int, v float64) {
 	s.mat.Data[i*s.mat.Stride+pj] = v
 }
 
+func (t *TriBandDense) At(i, j int) float64 {
+	if uint(i) >= uint(t.mat.N) {
+		panic(ErrRowAccess)
+	}
+	if uint(j) >= uint(t.mat.N) {
+		panic(ErrColAccess)
+	}
+	return t.at(i, j)
+}
+
+func (t *TriBandDense) at(i, j int) float64 {
+	isUpper := t.isUpper()
+	if (isUpper && i > j) || (!isUpper && i < j) {
+		return 0
+	}
+	kl := t.mat.K
+	ku := 0
+	if isUpper {
+		ku = t.mat.K
+		kl = 0
+	}
+	pj := j + kl - i
+	if pj < 0 || kl+ku+1 <= pj {
+		return 0
+	}
+	return t.mat.Data[i*t.mat.Stride+pj]
+}
+
+func (t *TriBandDense) SetTriBand(i, j int, v float64) {
+	if uint(i) >= uint(t.mat.N) {
+		panic(ErrRowAccess)
+	}
+	if uint(j) >= uint(t.mat.N) {
+		panic(ErrColAccess)
+	}
+	isUpper := t.isUpper()
+	if (isUpper && i > j) || (!isUpper && i < j) {
+		panic(ErrTriangleSet)
+	}
+	kl, ku := t.mat.K, 0
+	if isUpper {
+		kl, ku = 0, t.mat.K
+	}
+	pj := j + kl - i
+	if pj < 0 || kl+ku+1 <= pj {
+		panic(ErrBandSet)
+	}
+	t.mat.Data[i*t.mat.Stride+pj] = v
+}
+
+func (t *TriBandDense) setTriBand(i, j int, v float64) {
+	var kl int
+	if !t.isUpper() {
+		kl = t.mat.K
+	}
+	pj := j + kl - i
+	t.mat.Data[i*t.mat.Stride+pj] = v
+}
+
 // At returns the element at row i, column j.
 func (d *DiagDense) At(i, j int) float64 {
 	if uint(i) >= uint(d.n) {
--- a/mat/list_test.go
+++ b/mat/list_test.go
@@ -234,7 +234,7 @@ func legalDims(a Matrix, m, n int) bool {
 			return false
 		}
 		return true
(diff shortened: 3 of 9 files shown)


btracey:
This is pretty tedious, but the good news is that this is the last `float64` fundamental type, and it's as bad as it'll get.

kortschak:
Unless we add packed types.

mat/diagonal.go
```diff
@@ -100,6 +123,12 @@ func (d *DiagDense) TBand() Banded {
 	return TransposeBand{d}
 }
 
+// TTriBand performs an implicit transpose by returning the receiver inside a
+// TransposeTriBand. Note that Diagonal matrices are Upper by default.
+func (d *DiagDense) TTriBand() TriBanded {
+	return TransposeTriBand{d}
```
kortschak: It can just return `d`.

btracey:
That's true, but at least those won't be more complicated

mat/index_no_bound_checks.go
```diff
@@ -236,6 +236,65 @@ func (s *SymBandDense) set(i, j int, v float64) {
...
+	if uint(i) >= uint(t.mat.N) {
+		panic(ErrRowAccess)
+	}
+	if uint(j) >= uint(t.mat.N) {
+		panic(ErrColAccess)
+	}
+	return t.at(i, j)
+}
+
+func (t *TriBandDense) at(i, j int) float64 {
```
kortschak: Note #692 here.

mat/index_no_bound_checks.go
```diff
@@ -236,6 +236,65 @@ func (s *SymBandDense) set(i, j int, v float64) {
...
+		kl = 0
+	}
+	pj := j + kl - i
+	if pj < 0 || kl+ku+1 <= pj {
+		return 0
+	}
+	return t.mat.Data[i*t.mat.Stride+pj]
+}
+
+func (t *TriBandDense) SetTriBand(i, j int, v float64) {
```
kortschak: And here (presumably a panic for set, or a change of state from `blas.Unit` to `blas.NonUnit` and set all other to 1. (Probably the former).

mat/list_test.go
```diff
@@ -625,6 +661,8 @@ var testMatrices = []Matrix{
 	&DiagDense{},
 	&DiagDense{mat: blas64.Vector{Inc: 1}},
 	&DiagDense{mat: blas64.Vector{Inc: 10}},
+	NewTriBandDense(6, 2, Upper, nil),
```
kortschak: Zero values here too?

mat/symband.go
```diff
@@ -118,6 +121,10 @@ func (s *SymBandDense) Bandwidth() (kl, ku int) {
 	return s.mat.K, s.mat.K
 }
 
+func (s *SymBandDense) SymBand() (n, k int) {
```
kortschak: Doc comment.

mat/triband.go
```diff
@@ -0,0 +1,292 @@
...
+
+// MutableTriBanded is a triangular band matrix interface type that allows
+// elements to be altered.
+type MutableTriBanded interface {
+	TriBanded
+	SetTriBand(i, j int, v float64)
+}
+
+var (
+	_ Matrix               = TransposeTriBand{}
```
kortschak: `triBand` prototype value instead of repeated `TransposeTriBand{}`?

mat/triband.go
```diff
@@ -0,0 +1,292 @@
...
+func (t TransposeTriBand) TTriBand() TriBanded {
+	return t.TriBanded
+}
+
+// Untranspose returns the Triangular field.
+func (t TransposeTriBand) Untranspose() Matrix {
+	return t.TriBanded
+}
+
+func (t TransposeTriBand) UntransposeTri() Triangular {
```
kortschak: Doc comments for these.

mat/triband.go
```diff
@@ -0,0 +1,292 @@
...
+
+// T performs an implicit transpose by returning the receiver inside a Transpose.
+func (t *TriBandDense) T() Matrix {
+	return Transpose{t}
+}
+
+// IsZero returns whether the receiver is zero-sized. Zero-sized matrices can be the
+// receiver for size-restricted operations. TriBandDense matrices can be zeroed using Reset.
+func (t *TriBandDense) IsZero() bool {
+	// It must be the case that t.Dims() returns
```
kortschak: Add `Reset` method.

mat/diagonal.go
btracey: The proposal would be that this is implicit, with the signature to `NewDiagonal` unchanged. It just gets modified during the call to Transpose.

kortschak: If this is needed, I'd prefer it not go in the `DiagDense`. It seems out of place there.

mat/symband.go
btracey: Done.

mat/triband.go
btracey: Done.

mat/triband.go
btracey: Done.

mat/triband.go
btracey: Done.

mat/diagonal.go
btracey: I'll leave it this way for now. It's not super clear what the costs are, or that it matters that much. I think it's important to have because there are functions where you need the Triangular to be lower triangular (for instance) ad this would be otherwise impossible.

kortschak: SGTM

kortschak: To clarify, I think the approach here is better than having an `Uplo` field in the `DiagDense`.

mat/list_test.go
btracey: The note above is
`// The TriDense types have actual sizes because the return from Triangular is
// only valid when n == 0.`

We intentionally corrupt the return type when n is zero. We can fix this by not doing this corruption, for example changing the call to `Triangle` to be `return t.mat.N, t.triKind()`. This is simpler, and either way the user is warned not to rely on the value. If we want to keep the corruption, then we have to make the Triangular non-zero valued.

kortschak: OK, we might want to revisit this then, but not here.

mat/triband.go
```diff
@@ -0,0 +1,307 @@
...
+	return t.TriBanded
+}
+
+// Bandwidth returns the upper and lower bandwidths of the matrix.
+func (t TransposeTriBand) Bandwidth() (kl, ku int) {
+	kl, ku = t.TriBanded.Bandwidth()
+	return ku, kl
+}
+
+// TTri performs an implicit transpose by returning the TriBand field.
```
vladimir-ch: TBand

mat/triband.go
```diff
@@ -0,0 +1,307 @@
...
+	tTriBand TransposeTriBand
+	_        Matrix               = tTriBand
+	_        TriBanded            = tTriBand
+	_        Untransposer         = tTriBand
+	_        UntransposeTrier     = tTriBand
+	_        UntransposeBander    = tTriBand
+	_        UntransposeTriBander = tTriBand
+)
+
+// TransposeBandTri is a type for performing an implicit transpose of a TriBanded
```
vladimir-ch: TransposeTriBand

mat/triband.go
```diff
@@ -0,0 +1,307 @@
...
+	return TransposeBand{t}
+}
+
+// TriBand returns the number of rows/columns in the matrix, the
+// size of the bandwidth, and the orientation.
+func (t *TriBandDense) TriBand() (n, k int, kind TriKind) {
+	return t.mat.N, t.mat.K, TriKind(!t.IsZero()) && t.triKind()
+}
+
+// TBandTri performs an implicit transpose by returning the receiver inside a TransposeTriBand.
```
vladimir-ch: TTriBand

mat/diagonal.go
```diff
@@ -15,28 +15,51 @@ var (
...
 	Diag() int
-	Symmetric() int
 
 	// Bandwidth and TBand are included in the Diagonal interface
 	// to allow the use of Diagonal types in banded functions.
 	// Bandwidth will always return (0, 0).
 	Bandwidth() (kl, ku int)
 	TBand() Banded
+
+	// Triangular and TTri are included in the Diagonal interface
```
vladimir-ch: Triangle

mat/triband.go
```diff
@@ -0,0 +1,307 @@
...
+	return t.mat.N, t.mat.N
+}
+
+// T performs an implicit transpose by returning the receiver inside a Transpose.
+func (t *TriBandDense) T() Matrix {
+	return Transpose{t}
+}
+
+// IsZero returns whether the receiver is zero-sized. Zero-sized matrices can be the
+// receiver for size-restricted operations. TriBandDense matrices can be zeroed using Reset.
```
vladimir-ch: The comment to Reset uses "dimensionally restricted operation", so it would be good to unify.

vladimir-ch:
I wouldn't add the packed types.

kortschak:
> I wouldn't add the packed types.

Sorry, my delivery was flat. That was intended as a joke.

If we do add anything like that it would have to be proper sparse. I think that's a long way off.

btracey:
Though it's a long way off, I think sparse should go in `mat/sparse` rather than in the main thing. They don't interact very well with the Dense types and have their own sets of methods that are useful.

mat/diagonal.go
btracey: Done

mat/triband.go
btracey: Done

mat/triband.go
btracey: Done

mat/triband.go
btracey: This was copied from the documentation on `Dense`. I'll leave this and open an issue.

mat/triband.go
btracey: Done.

Issue #648, opened 2018-11-14
graph/encoding/dot: Unmarshal docs do not say only one graph may be unmarshaled


kortschak:

mewmew:
I'm fine with having dot.Unmarshal return a non-nil error if it encounters more than one graph in the input file.

I took a quick look at #624, and from what I can tell (by a quick glance at the test cases) @32bitkid implemented support for unmarshaling multigraphs that handle multiple edges from the same src to dst nodes. However, I saw no test cases for multiple graphs in the same source DOT file, is this supposed to be supported? Also, would self-loops be supported in the multi-graph unmarshalling? If so, that's great!

```dot
digraph {
	a -> b
}

digraph {
	c -> d
}
```

mewmew:
Ok, great. Also, good you have better memory than I do :)

32bitkid:
@mewmew The changes in #624 _do_, or at least it is my intention that it should, support support self-loops on multi-graphs. 👍

kortschak:
I think #694 actually fixed this. Do you agree @mewmew? I don't know why I said "updates".

mewmew:
>  If the number of graphs encoded in data is not one, an error is returned and dst will hold the first graph in data.

Yes, I would consider this issue fixed. Feel free to close.

PR #694, opened 2018-11-24
encoding/dot: minor fixups

### LINKED ISSUE #651: graph/encoding/dot: error returned by calls to encoding.SetAttributes are not passed back to caller
In `Unmarshal` are a few calls to the `encoding.SetAttributes` method of `dst` that have their error checked. The error is not included in the error returned by `Unmarshal`. It should be.

kortschak:
Please take a look.

Updates #648.
Fixes #651.

### DIFF
--- a/graph/encoding/dot/decode.go
+++ b/graph/encoding/dot/decode.go
@@ -39,28 +39,34 @@ type PortSetter interface {
 }
 
 // Unmarshal parses the Graphviz DOT-encoded data and stores the result in dst.
+// If the number of graphs encoded in data is not one, an error is returned and
+// dst will hold the first graph in data.
 func Unmarshal(data []byte, dst encoding.Builder) error {
 	file, err := dot.ParseBytes(data)
 	if err != nil {
 		return err
 	}
-	if len(file.Graphs) != 1 {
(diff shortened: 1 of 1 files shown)


graph/encoding/dot/decode.go
```diff
@@ -39,28 +39,34 @@ type PortSetter interface {
 }
 
 // Unmarshal parses the Graphviz DOT-encoded data and stores the result in dst.
+// If the number of graphs encoded in data is not one, an error is returned and
```
vladimir-ch: Is this useful? The caller cannot distinguish it from other error situations. Or is the intention different?

kortschak: The error is different and in the case of other errors there will be no graph written. My intention in the longer run is to use a rich error that allows the user to actually find out how many graphs there were.

I don't know how useful it is. Maybe it would be better to have something like a `Decoder` type that will iterate over the input and write graphs while they exist.

vladimir-ch: Thanks, both SGTM.

kortschak: I can back this out and just leave the doc change and deal with what we want to do here later.

vladimir-ch: I don't think it's necessary, just leave it in and we can still revisit this later.

PR #714, opened 2018-11-26
Listtypes

btracey:
Please take a look.

### DIFF
--- a/mat/list_test.go
+++ b/mat/list_test.go
@@ -563,11 +563,17 @@ func makeCopyOf(a Matrix) Matrix {
 		case *basicBanded:
 			band = (*BandDense)(s)
 		}
-		data := make([]float64, len(band.mat.Data))
-		copy(data, band.mat.Data)
-		r, c := band.Dims()
-		kl, ku := band.Bandwidth()
-		m := NewBandDense(r, c, kl, ku, data)
+		m := &BandDense{
+			mat: blas64.Band{
+				Rows:   band.mat.Rows,
+				Cols:   band.mat.Cols,
+				KL:     band.mat.KL,
+				KU:     band.mat.KU,
+				Data:   make([]float64, len(band.mat.Data)),
+				Stride: band.mat.Stride,
+			},
+		}
+		copy(m.mat.Data, band.mat.Data)
 		return returnAs(m, t)
 	case *SymBandDense, *basicSymBanded:
 		var sym *SymBandDense
--- a/mat/list_test.go
+++ b/mat/list_test.go
@@ -510,11 +510,27 @@ func makeRandOf(a Matrix, m, n int) Matrix {
 		if m != n {
 			panic("bad size")
 		}
-		diag := NewDiagonal(m, nil)
-		for i := 0; i < n; i++ {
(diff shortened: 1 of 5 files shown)


btracey:
I think this is right, but it's late here. I'll take another glass in the morning.

kortschak:
>I'll take another glass in the morning.

Are you learning rhyming slang?

btracey:
Yea, I am "glass gobbledegook" -> look

btracey:
Okay, I think this is ready for review.

kortschak:
Ahh. I just found out why our slang for look, "squiz", is what it is; it looks like it might have once been rhyming slang. Squizzy Taylor was a notorious gangster in the 20s in Melbourne.

mat/list_test.go
```diff
@@ -648,45 +769,86 @@ func underlyingData(a Matrix) []float64 {
...
+// correct TriKind to be read).
 var testMatrices = []Matrix{
 	&Dense{},
-	&SymDense{},
-	NewTriDense(3, true, nil),
-	NewTriDense(3, false, nil),
-	&basicVector{},
-	&VecDense{mat: blas64.Vector{Inc: 1}},
-	&VecDense{mat: blas64.Vector{Inc: 10}},
-	&DiagDense{},
```
kortschak: Why does the zero value go away?

btracey: It went away because it's redundant with having an incremented and a non-incremented `Dense` type.

btracey:
Could you elaborate on what you mean by "here"?

btracey:
If by "here" you mean "This whole file", then I would suggest that this PR stay compact at what it is (shoring up some current inconsistencies), and then I'll write a big doc comment for the file

mat/list_test.go
btracey: Okay, sorry, fixed an issue with the increments, and now the zero value is back (this is much better)

kortschak:
By here I mean the PR. Just add a GitHub comment about what is happening, particularly with the rearrangement of types which I don't understand without going through it in detail. And then we add that comment to the commit message when the squash happens.

btracey:
Oh, weird, I thought I had that, sorry

```
mat: Add missing types to list test

With the recent additions of Diagonal and TriBanded, there were some matrix types that were not added to the list of testing types. There were additionally some older types that had been left out. This PR completes the set of types, their transposes, and their basic counterparts to the set of matrices. Changes are made to the makeNew etc. routines in order to support this.
```

btracey:
The re-arrangement of types was just to make it organized which was previously pretty haphazard

mat/triangular.go
```diff
@@ -154,7 +154,7 @@ func (t *TriDense) Dims() (r, c int) {
 // Triangle returns the dimension of t and its orientation. The returned
 // orientation is only valid when n is not zero.
 func (t *TriDense) Triangle() (n int, kind TriKind) {
-	return t.mat.N, TriKind(!t.IsZero()) && t.triKind()
+	return t.mat.N, t.triKind()
```
kortschak: Can you add a note to the commit message about why this and the same for the banded type happens.

PR #723, opened 2018-12-01
quat: rename Quat to Number

kortschak:
This reduces stutter and parallels the convention in the proposed dual and hyperdual packages.

I was originally reluctant to do this, but I think the reduction of stutter and the now found mathworks [page](https://au.mathworks.com/help/robotics/ref/quaternion.html) that uses "quaternion number" make it OK in my mind.

Please take a look.


num/quat/trig_test.go
```diff
@@ -11,25 +11,25 @@ import (
...
-	{q: Quat{Jmag: math.Pi / 2}, want: Quat{Jmag: imag(cmplx.Sin(complex(0, math.Pi/2)))}},
-	{q: Quat{Kmag: math.Pi / 2}, want: Quat{Kmag: imag(cmplx.Sin(complex(0, math.Pi/2)))}},
+	{q: Number{}, want: Number{}},
+	{q: Number{Real: math.Pi / 2}, want: Number{Real: 1}},
+	{q: Number{Imag: math.Pi / 2}, want: Number{Imag: imag(cmplx.Sin(complex(0, math.Pi/2)))}},
+	{q: Number{Jmag: math.Pi / 2}, want: Number{Jmag: imag(cmplx.Sin(complex(0, math.Pi/2)))}},
+	{q: Number{Kmag: math.Pi / 2}, want: Number{Kmag: imag(cmplx.Sin(complex(0, math.Pi/2)))}},
 
-	// Exercises from Real Quaternionic Calculus Handbook doi:10.1007/978-3-0348-0622-0
+	// Exercises from Real Numberernionic Calculus Handbook doi:10.1007/978-3-0348-0622-0
```
btracey: The downsides of find and replace

num/quat/trig_test.go
```diff
@@ -79,20 +79,20 @@ func TestSinh(t *testing.T) {
...
-	{q: Quat{Jmag: math.Pi / 2}, want: Quat{Real: real(cmplx.Cos(complex(0, math.Pi/2)))}},
-	{q: Quat{Kmag: math.Pi / 2}, want: Quat{Real: real(cmplx.Cos(complex(0, math.Pi/2)))}},
+	{q: Number{}, want: Number{Real: 1}},
+	{q: Number{Real: math.Pi / 2}, want: Number{Real: 0}},
+	{q: Number{Imag: math.Pi / 2}, want: Number{Real: real(cmplx.Cos(complex(0, math.Pi/2)))}},
+	{q: Number{Jmag: math.Pi / 2}, want: Number{Real: real(cmplx.Cos(complex(0, math.Pi/2)))}},
+	{q: Number{Kmag: math.Pi / 2}, want: Number{Real: real(cmplx.Cos(complex(0, math.Pi/2)))}},
 
-	// Example from Real Quaternionic Calculus Handbook doi:10.1007/978-3-0348-0622-0
+	// Example from Real Numberernionic Calculus Handbook doi:10.1007/978-3-0348-0622-0
```
btracey: Here too.

num/quat/trig_test.go
kortschak: He! Thanks.

num/quat/trig_test.go
kortschak: Fixed.

kortschak:
>And we think this is better than `num.Quat`, `num.Dual`, etc.?

Yes. That approach would handle the types, but not the functions. With the approach here we have `quat.Sin`, but with them all in num we would have `num.SinQuat`, with all the functions mixed together, or `num.QuatSin` which I don't think is much better (and probably harder to discover).

btracey:
That's a solid argument.

Issue #677, opened 2018-11-22
Optimize: Idea to remove DefaultSettings()


btracey:
Currently, we have `DefaultSettingsLocal` and `DefaultSettingsGlobal` for users to get easy defaults. In Go generally, and also in Gonum specifically, we try to have the zero value be the default value. This issue is a proposal to make that true

Aside from being un-go-like, there are two gotchas based on the current design. First, we have to choose some default value, and right now that default is `DefaultLocalSettings`. If users change the method to a global method, but do not also change the settings, they are going to get weird behavior. Second, because not all of the settings can use the zero value, you cannot naively just create the zero value of settings and modify the settings you care about. We do this in the tests, but we know what we're doing. For a typical user who probably does not want to understand all of the settings and their interactions, this is easy to mess up, in particular the `FunctionTolerance` setting could easily lead to mistakes.

There are a couple challenges to having a zero-value default struct.
- We have the `FunctionThreshold` setting, which has a meaningful value for zero 
- There are two categories of default settings, Local and Global, and how do you make a reasonable value for both of them?
- In particular for the above, Local optimization techniques typically define convergence by having the gradient be a small value. Global optimization techniques typically have some nebulous definition of convergence based on number of function evaluations.

I think there are two parts to the solution. The first is to adopt #488  . This would replace both the existing `FunctionThreshold` and `FunctionConverge` with an abstraction that takes in updates, and returns if convergence happens. By default, we could use the existing `FunctionConverge`. Using an interface makes it really easy for the user to implement different criteria, for example `FunctionThreshold`. 

[...]

PR #724, opened 2018-12-02
optimize: move global code to minimize

### LINKED ISSUE #482: optimize: Combine Global and Local

btracey:
Fixes #482.

Please take a look.


btracey:
This is just a straight cut and paste. Leaving GlobalSettings in global, because that makes sense, but I suspect the file will go away with #677

PR #728, opened 2018-12-08
optimize: make function converger an interface

### LINKED ISSUE #488: optimize: Replace FunctionConverge with an interface
Right now we only allow a specific kind of function convergence. It seems like this should be replaced with an interface. An extended idea is to pass `*Location` and not just `f`. If we can figure out how to make this easy, we could possibly wrap FunctionThreshold and GradientThreshold into this (or just FunctionThreshold since it's less common to use).

btracey:
Fixes #488.
Updates #677.

Please take a look.

### DIFF
--- a/optimize/types.go
+++ b/optimize/types.go
@@ -186,7 +186,7 @@ type Settings struct {
 	GradientThreshold float64
 
 	// Converger checks if the optimization has converged based on the (history
-	// of) locations found during the optimizaiton. Minimize will pass the
+	// of) locations found during the optimization. Minimize will pass the
 	// Location at every MajorIteration to the Converger.
 	//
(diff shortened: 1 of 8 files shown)


optimize/types.go
```diff
@@ -185,18 +185,18 @@ type Settings struct {
 	// The default value is 1e-6.
 	GradientThreshold float64
 
-	// FunctionConverge tests that the function value decreases by a
-	// significant amount over the specified number of iterations.
+	// Converger checks if the optimization has converged based on the (history
+	// of) locations found during the optimizaiton. Minimize will pass the
```
vladimir-ch: optimization

btracey: Done.

PR #753, opened 2018-12-09
blas64: add length field N to Vector

### LINKED ISSUE #736: blas/blas64: Vector doesn't contain length?
Is there a good reason `blas64.Vector` doesn't contain information about it's length? This seems inconsistent with the rest of the types, and I don't see any discussion on the matter.

btracey:
Alongside, fix the implementation of mat.VecDense and mat.Diagonal, as well as other changes needed to fix this change.

Fixes #736.

Please take a look.


lapack/testlapack/general.go
```diff
@@ -828,18 +828,16 @@ func isOrthogonal(q blas64.General) bool {
...
 		}
 		if math.Abs(nrm-1) > tol {
 			return false
 		}
 		for j := i + 1; j < n; j++ {
-			dot := blas64.Dot(n,
-				blas64.Vector{Data: q.Data[i*q.Stride:], Inc: 1},
-				blas64.Vector{Data: q.Data[j*q.Stride:], Inc: 1},
-			)
+			dot := blas64.Dot(blas64.Vector{Data: q.Data[i*q.Stride:], Inc: 1},
```
kortschak: Where is `n`?

mat/cholesky.go
```diff
@@ -503,9 +503,9 @@ func (c *Cholesky) SymRankOne(orig *Cholesky, alpha float64, x Vector) (ok bool)
 				// Multiply the extended factorization matrix by
 				// the Givens matrix from the left. Only
 				// the i-th row and x are modified.
-				blas64.Rot(n-i-1,
-					blas64.Vector{Data: umat.Data[i*stride+i+1 : i*stride+n], Inc: 1},
-					blas64.Vector{Data: work[i+1 : n], Inc: 1},
+				blas64.Rot(
+					blas64.Vector{N: n - i - 1, Data: umat.Data[i*stride+i+1 : i*stride+n], Inc: 1},
```
kortschak: I think this is the kind of case we were considering as well.

mat/dense_arithmetic.go
```diff
@@ -504,12 +504,12 @@ func (m *Dense) Exp(a Matrix) {
 	a1.Copy(a)
 	v := getWorkspace(r, r, true)
 	vraw := v.RawMatrix()
-	vvec := blas64.Vector{Inc: 1, Data: vraw.Data}
+	vvec := blas64.Vector{N: r * r, Inc: 1, Data: vraw.Data}
```
kortschak: This gets recalculated a lot; `n := r*r`? (though it did previously just differently).

mat/matrix.go
```diff
@@ -345,7 +343,7 @@ func Dot(a, b Vector) float64 {
 	}
```
kortschak: In the condition above, we are now checking twice.

mat/vector.go
```diff
@@ -175,7 +174,7 @@ func (v *VecDense) Reset() {
 	// No change of Inc or n to 0 may be
```
kortschak: "or N"

mat/matrix.go
btracey: Done.

mat/vector.go
btracey: Done.

mat/cholesky.go
btracey: Sorry, that is meant to be read as "I agree with you this is a net win"

kortschak: >The fact that the size is now doubled?

No, that we are passing the length twice.

blas/blas64/blas64.go
```diff
@@ -98,64 +99,79 @@ type SymmetricPacked struct {
...
+	}
+	blas64.Dswap(x.N, x.Data, x.Inc, y.Data, y.Inc)
 }
 
 // Copy copies the elements of x into the elements of y:
 //  y[i] = x[i] for all i.
-func Copy(n int, x, y Vector) {
-	blas64.Dcopy(n, x.Data, x.Inc, y.Data, y.Inc)
+func Copy(x, y Vector) {
+	if x.N != y.N {
```
kortschak: I forgot to comment here. Since BLAS doesn't hold the notion of independent lengths here. The behaviour for `copy([n]T, [m]T)` is both unspecified and not obviously defined from external context. For example the built in `copy` does min elements, while here we panic. For the other diadic functions here there is a reasonable obvious choice, so I think they are fine. This one needs discussion and more documentation.

lapack/testlapack/general.go
kortschak: This is the foot gun that exists with this approach.

blas/blas64/blas64.go
btracey: I believe blas has an implied equality. It is always used as such in blas and lapack as far as I can remember.

lapack/testlapack/general.go
btracey: Agreed. The previous foot gun was having one size though the two vectors were the same size. Unfortunately, this is a problem with testing Lapack, which doesn't get the niceties `mat` does. There are many fewer hand-coded `blas.Vector` in `mat`

mat/matrix.go
kortschak: No, sorry. That was not something to change. It's an unfortunate consequence of the checks in blas64 now. We have to check twice. We promise to `panic(ErrShape)` in mat for this situation. As it stands now, we don't.

blas/blas64/blas64.go
kortschak: OK. That needs to be documented, at least on `Copy` since there is a semantic collision/overloading with `copy`.

mat/matrix.go
btracey: Good catch. Reverted.

blas/blas64/blas64.go
btracey: Done.

blas/blas64/blas64.go
```diff
@@ -158,6 +158,8 @@ func Swap(x, y Vector) {
 
 // Copy copies the elements of x into the elements of y:
 //  y[i] = x[i] for all i.
+// Note that the behavior of Copy behaves differently than the built-in copy.
```
kortschak: `// Copy requires that the lengths of x and y match and will panic otherwise.` (avoiding behaviour behaving and also the more stark disparities in American v British English).

kortschak:
When you squash, I think this should be "blas64,mat: add length field N to blas64.Vector" or something similar (even "blas64: add length field N to Vector" given that the flow on to mat is obvious).

PR #762, opened 2018-12-12
graph: add Empty universal iterator for empty returns

### LINKED ISSUE #614: graph: decide whether empty collections of nodes, edges and lines are nil or just empty

kortschak:
Please take a look.

I think this is the best approach, but it is here for discussion.

Closes #614.


graph/nodes_edges.go
```diff
@@ -227,3 +227,40 @@ func WeightedLinesOf(it WeightedLines) []WeightedLine {
...
+	_ EdgeSlicer         = Empty
+	_ WeightedEdges      = Empty
+	_ WeightedEdgeSlicer = Empty
+	_ Lines              = Empty
+	_ LineSlicer         = Empty
+	_ WeightedLines      = Empty
+	_ WeightedLineSlicer = Empty
+)
+
+const nothing = empty(true)
```
vladimir-ch: What is the purpose of `nothing`?

PR #764, opened 2018-12-14
mat: rename NewDiagonal to NewDiagDense

### LINKED ISSUE #662: mat: NewDiagonal should be called NewDiagDense
All the rest of the implementations name the concrete type. This one should too.

kortschak:
Please take a look.

Closes #662.

### DIFF
--- a/mat/diagonal.go
+++ b/mat/diagonal.go
@@ -73,10 +73,10 @@ type DiagDense struct {
 	mat blas64.Vector
 }
 
-// NewDiagonal creates a new Diagonal matrix with n rows and n columns.
-// The length of data must be n or data must be nil, otherwise NewDiagonal
-// will panic. NewDiagonal will panic if n is zero.
-func NewDiagonal(n int, data []float64) *DiagDense {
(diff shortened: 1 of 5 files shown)


vladimir-ch:
Sorry for being late but I'm confused by the name. The name suggests that the underlying storage is a dense matrix which is not. This is basically a sparse matrix and only the diagonal is stored.

The same applies to all the other `*BandDense` types. `TriBandDense` even says that the matrix is in dense storage format which is not correct.

It does not apply to TriDense and SymDense where the underlying storage is truly dense.

btracey:
I understand your point, but ... One of the nice things about the current nomenclature is that FooDense is the actual type, while Foo is the interface. We could be more accurate with BandPacked, and DiagPacked, but I'm not sure if it's worth it.

vladimir-ch:
I don't like replacing `Dense` with `Packed` either but the nomenclature is actually that BandDense is the type and Banded is the interface, so dropping Dense from BandDense could be an option, giving Band as a type. I'm not sure if that's better, just wanted to raise the question.

PR #766, opened 2018-12-14
stat/mds: reduce work in TorgersonScaling and fix k return

kortschak:
The fix for k papers over inaccuracies in near zero Eigenvalues. So this may want further work there.

Please take a look.


kortschak:
Working on this has reminded me of some weirdness in TorgersonScaling that I had forgotten about to do in part with ED and differences in behaviour between pure Go and ASM impllementations. I'm going to fix(-ish) this up here and add a TODO/FIXME and file an issue to address the underlying cause more deeply when I have a clue about what is going on.

kortschak:
Is someone able to review this?

btracey:
Sorry, this totally slipped past my radar. Apologies.

PR #772, opened 2018-12-21
optimize: Refactor gradient convergence and remove DefaultSettings

btracey:
The current API design makes it easy to make a mistake in not using the DefaultSettings. This change makes the zero value of Settings do the 'right thing'. The remaining setting that is used by the DefaultSettings is to change the behavior of the GradientTolerance. This was necessary because gradient-based Local methods (BFGS, LBFGS, CG, etc.) typically _define_ convergence by the value of the gradient, while Global methods (CMAES, GuessAndCheck) are defined by _not_ converging when the gradient is small. The problem is to have two completely different default behaviors without knowing the Method. The solution is to treat a very small value of the gradient as a method-based convergence, in the same way that a small spread of data is a convergence of CMAES. Thus, the default behavior, from the perspective of Settings, is never to converge based on the gradient, but all of the Local methods will converge when a value close to the minimum is found. This default value is set to a very small value, such that users should not want a smaller value. A user can thus still set a (more reasonable) convergence value through settings.

Please take a look.


optimize/types.go
```diff
@@ -161,28 +160,23 @@ func (p Problem) satisfies(method Needser) error {
 }
 
 // Settings represents settings of the optimization run. It contains initial
-// settings, convergence information, and Recorder information. In general, users
-// should use DefaultSettings rather than constructing a Settings literal.
-//
-// If Recorder is nil, no information will be recorded.
+// settings, convergence information, and Recorder information. Convergence
+// settings are only checked at MajorIterations, while Evaluation thresholds
+// are checked every Operation. See the field comments for default values.
```
vladimir-ch: are checked _at_ every Operation

PR #773, opened 2018-12-21
 mat: rewrite Cholesky.ToSym without calling SymOuterK

vladimir-ch:
A substitute for the Dlauum attempt. `Cholesky.ToSym` is not super important but with this PR it becomes slightly faster.

```
name                    old time/op  new time/op  delta
CholeskyToSym/n=10-4    2.65µs ± 2%  0.78µs ± 0%  -70.75%  (p=0.029 n=4+4)
CholeskyToSym/n=100-4    329µs ± 1%   124µs ± 0%  -62.33%  (p=0.029 n=4+4)
CholeskyToSym/n=1000-4   241ms ± 2%   175ms ± 0%  -27.34%  (p=0.029 n=4+4)
```

Please take a look.

### DIFF
--- a/mat/cholesky.go
+++ b/mat/cholesky.go
@@ -287,11 +287,25 @@ func (c *Cholesky) ToSym(dst *SymDense) *SymDense {
 	}
 	n := c.chol.mat.N
 	if dst == nil {
-		dst = NewSymDense(n, make([]float64, n*n))
+		dst = NewSymDense(n, nil)
 	} else {
 		dst.reuseAs(n)
 	}
-	dst.SymOuterK(1, c.chol.T())
+	// Copy c.chol into a TriDense U with dst's backing slice.
+	// Operations on u are reflected in dst.
+	u := *c.chol
+	u.mat.Data = dst.mat.Data
+	u.Copy(c.chol)
+	// Compute the product U^T*U using the algorithm from LAPACK/TESTING/LIN/dpot01.f
+	a := u.mat.Data
+	lda := u.mat.Stride
+	bi := blas64.Implementation()
+	for k := n - 1; k >= 0; k-- {
+		a[k*lda+k] = bi.Ddot(k+1, a[k:], lda, a[k:], lda)
+		if k > 0 {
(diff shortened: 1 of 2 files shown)


mat/cholesky.go
```diff
@@ -287,11 +287,24 @@ func (c *Cholesky) ToSym(dst *SymDense) *SymDense {
 	}
 	n := c.chol.mat.N
 	if dst == nil {
-		dst = NewSymDense(n, make([]float64, n*n))
+		dst = NewSymDense(n, nil)
 	} else {
 		dst.reuseAs(n)
 	}
-	dst.SymOuterK(1, c.chol.T())
+	// Copy c.chol into a TriDense U with dst's backing slice.
```
vladimir-ch: I'm not very happy about this part, so if you have any suggestion how to do this better, I'll appreciate it.

kortschak: It is a bit funky. It took me a while to figure out what is going on, though I don't see a better way to do it. Maybe just an additional comment saying that work on `u` is reflected into `dst` and do the work on `u`, like so:
```
	// Copy c.chol into a TriDense U with dst's backing slice.
	// Operations on u are reflected in dst.
	u := *c.chol
	u.mat.Data = dst.mat.Data
	u.Copy(c.chol)
	// Compute the product U^T*U using the algorithm from LAPACK/TESTING/LIN/dpot01.f
	bi := blas64.Implementation()
	a := u.mat.Data
	lda := u.mat.Stride
	for k := n - 1; k >= 0; k-- {
		a[k*lda+k] = bi.Ddot(k+1, a[k:], lda, a[k:], lda)
		if k > 0 {
			bi.Dtrmv(blas.Upper, blas.Trans, blas.NonUnit, k, a, lda, a[k:], lda)
		}
	}
	return dst
```

vladimir-ch: Funky is the word :-) I added the comment and do the work on u, thanks.

vladimir-ch:
I will wait also for @btracey

PR #775, opened 2018-12-22
encoding/dot: set DOT node ID before adding node to graph

mewmew:
This makes it possible to implement the AddNode method of
graph.NodeAdder to track the nodes of a graph by DOT node
ID, as they are added.

Please take a look.

### DIFF
--- a/graph/encoding/dot/decode.go
+++ b/graph/encoding/dot/decode.go
@@ -153,10 +153,10 @@ func (gen *generator) node(dst graph.NodeAdder, id string) graph.Node {
 		return n
 	}
 	n := dst.NewNode()
-	dst.AddNode(n)
 	if n, ok := n.(DOTIDSetter); ok {
 		n.SetDOTID(id)
 	}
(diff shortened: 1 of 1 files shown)


mewmew:
For reference, one such implementation could look something as follows:

```

type Graph struct {
	// Underlying simple.DirectedGraph.
	*simple.DirectedGraph
	// nodes maps from DOT node ID to associated node.
	nodes map[string]graph.Node
}

// AddNode adds a node to the graph. AddNode panics if the added node ID matches
// an existing node ID.
func (g *Graph) AddNode(n graph.Node) {
	nn := n.(dot.Node)
	dotID := nn.DOTID()
	if prev, ok := g.nodes[dotID]; ok {
		panic(fmt.Errorf("node with DOT node ID %q already present; prev `%v`, new `%v`", dotID, prev, nn))
	}
	g.nodes[dotID] = nn
	g.DirectedGraph.AddNode(nn)
}
```

Issue #627, opened 2018-10-24
graph: consider changing semantics of Len() for graph.Iterator


kortschak:
### Background
The graph node and edge iterators free the graph packages from the size limits of `[]graph.Node` and `[]graph.Edge` when querying graphs. The freedom to be able to perform graph operations on very large graphs that are backed by disk was one of the reasons for changing to `int64` ID values rather than `int` although at the time I was envisioning only working on subgraphs that had been pull from those larger backing stores.

With store-agnostic iterators we can conceivably work directly (though slowly) with nodes and edges that are not-memory resident with the exception that the iterator `Len` method must have a value to will fit inside an `int`. This won't work for 32 bit archs and will also not work for very large graphs. Given that the `Len` method is also used to give the user information about potential allocations either directly or via `graph.NodesOf` _etc._ it also doesn't work for very large graphs that may be materialised into a slice.

### Proposal
I suggest that we add a minor change to the semantics of the `Len` method's returned value.

> If the size/order of the graph is either unknown, too costly to calculate or likely to be too large to materialise (all loosely defined) then `Len` may return `-1`.
> The iterator's `Next` loop would however still be able to calculate the number of items.

An alternative to this is to add an `ok bool` to `Len` method, though it is unclear to me how that will help except in the case of 32 archs, and then we would also nee to change the return integer type to `int64` which I don't think is tenable.

### Potential impact of proposal

The only change to the graph packages would be an addition to the documentation for `graph.Iterator` on the `Len` method and a defensive check for negative length in `graph.NodesOf`, `graph.EdgesOf`, `graph.WeightedEdgesOf`, `graph.LinesOf` and `graph.WeightedLinesOf` after the relevant `*Slicer` type switch, returning a nil slice if the length is negative.

There are no types in the graph packages that would be affected by any of these changes.

I would do the work.

tzneal:
+1 from a random user.  I needed a descendants iterator for a directed graph, and thought about cleaning it up and sending as a PR, but I didn't want to implement Len().

kortschak:
Can you elaborate? "Didn't want to implement Len" is not a hugely compelling argument. In fact this is one thing I am concerned about - see second para just added to the impact section.

tzneal:
In my mind, iterator is a loose concept to walk a series in a particular order.  The length doesn't come into it much, for the functions that require length they can just traverse the iterator and construct a list.

[...]

PR #776, opened 2018-12-22
graph: allow indeterminate iterator length

kortschak:
Please take a look.

I have decided to use a panic in the slicer functions rather than returning nil to avoid silent failure from incorrect use.

The next step for this is to go through the functions that we have an document those that are able to work on indeterminate iterators. Then fix those that could but currently can not.

The intention with the semantics around the magnitude of the length return value are that for disk-backed graph the size can be known, but it may not be materialisabe. This is useful information, just not information that can be used directly by most of the analytical functions.

Updates #627.

### DIFF
--- a/graph/nodes_edges.go
+++ b/graph/nodes_edges.go
@@ -20,6 +20,15 @@ type Iterator interface {
 
 	// Len returns the number of items remaining in the
 	// iterator.
+	//
+	// If the number of items in the iterator is unknown,
+	// too large to materialize or too costly to calculate
+	// then Len may return a negative value.
+	// In this case the consuming function must be able
+	// to operate on the items of the iterator directly
+	// without materializing the items into a slice.
+	// The magnitude of a negative length has
+	// implementation-dependent semantics.
 	Len() int
 
 	// Reset returns the iterator to its start position.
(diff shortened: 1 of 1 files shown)


graph/nodes_edges.go
```diff
@@ -47,18 +56,23 @@ type NodeSlicer interface {
 
 // NodesOf returns it.Len() nodes from it. If it is a NodeSlicer, the NodeSlice method
 // is used to obtain the nodes. It is safe to pass a nil Nodes to NodesOf.
+//
+// If the Nodes has en indeterminate length, NodesOf will panic.
```
vladimir-ch: s/en/an throughout

kortschak: Thanks. Fixed.

kortschak: Not throughout - I'll be back in a sec.

kortschak: Done

kortschak:
I'm still not entirely sure of the approach. So if you would have a think about potential problems, that would be helpful.

kortschak:
An alternative/extension is requiring/stating (choice TBD) that if `Len` is negative, then there must/may (choice TBD) be another method `Size() int64`(?) that will return the actual size of the iterator.

mewmew:
> An alternative/extension is requiring/stating (choice TBD) that if Len is negative, then there must/may (choice TBD) be another method Size() int64(?) that will return the actual size of the iterator.

Not sure I follow. Wasn't the idea to use a negative value for Len when the length could not be determined? If so, how would you return a value from Size? I may just have misunderstood something.

kortschak:
There are two conditions when the type is allowed to return a negative value:

1. too large to materialise
2. too costly to determine

These are quite different and the `Size` method would only be provided in the first case (I guess this answers the TBD). Say you have a graph stored in a kv store, that graph may have more nodes than can be stored in a slice (easily imaginable on a 32 bit arch), so it would return len=-1 and have `Size` which would return the actual number of values held. `Size` needs to return an `int64` for this reason. In the second case, there would be no `Size` method.

mewmew:
Thanks for the explanation.

kortschak:
Thinking more, the current PR is consistent with the may have a `Size` method case; -len has implementation dependent meaning. It can be up to the implementation to decide that -len means there is a `Size` method. I think that we should go with this and see how it works out.

Issue #223, opened 2017-09-02
optimize,diff/fd: impedance mismatch in Hessian and mat.MutableSymmetric


sbinet:
recently, I wanted to numerically compute a hessian, so I used:
https://godoc.org/gonum.org/v1/gonum/diff/fd#Hessian
```go
// Hessian approximates the Hessian matrix of the multivariate function f
// at the location x. That is
//  H_{i,j} = ∂^2 f(x)/∂x_i ∂x_j
// If dst is not nil, the resulting H will be stored in-place into dst
// and returned, otherwise a new matrix will be allocated first. Finite difference
// formula and other options are specified by settings. If settings is nil,
// the Hessian will be estimated using the Forward formula and a default step size.
//
// Hessian panics if the size of dst and x is not equal, or if the derivative
// order of the formula is not 1.
func Hessian(
    dst *mat.SymDense, f func(x []float64) float64, 
    x []float64, settings *Settings,
) *mat.SymDense
```

so I could feed that into `optimize.Problem.Hess`:
https://godoc.org/gonum.org/v1/gonum/optimize#Problem
```go
type Problem struct {
    // Func evaluates the objective function at the given location. Func
    // must not modify x.
    Func func(x []float64) float64

    // Grad evaluates the gradient at x and stores the result in-place in grad.
    // Grad must not modify x.
    Grad func(grad []float64, x []float64)

    // Hess evaluates the Hessian at x and stores the result in-place in hess.
    // Hess must not modify x.
    Hess func(hess mat.MutableSymmetric, x []float64)

    // Status reports the status of the objective function being optimized and any
    // error. This can be used to terminate early, for example when the function is
    // not able to evaluate itself. The user can use one of the pre-provided Status
    // constants, or may call NewStatus to create a custom Status value.
    Status func() (Status, error)
}
```

so, I had to resort to:

```go
p.Hess = func(hess mat.MutableSymmetric, x []float64) {
	fd.Hessian(hess.(*mat.SymDense), f.fct, x, nil)
}
```

looking at `fd.Hessian`, it seems to me the only requirement for using `*mat.SymDense` is coming from the ability to pass `nil` as an optional parameter (as a Hessian matrix) and return it (as an output value.)

couldn't we just use `mat.MutableSymmetric` everywhere ?

btracey:
[...]

Issue #593, opened 2018-09-20
optimize: Change Hessian function signature?


btracey:
I've been thinking for a while that we might want to change the signature for the Hessian method. The current signature forces the Hessian to be a `*SymDense`, while for many problems the Hessian is sparse, or block structured, or something. Directly, this limits the size of problems that can be solved due to memory constraints. Indirectly, this also significantly limits optimizations available. For instance, right now `Newton` calls Solve, which means the matrix is always using a `Dense` routine. There could be simplifications available. For instance, `Newton` really only requires a `SolveGradToer`, something like

```
type SolveVecToer interface {
    Symmetric()
    SolveGradTo(v []float64) err
}
```
which is easily implemented by `SymDense`, but also enables taking advantage of sparse structure (or approximate solves).

cc @vladimir-ch

vladimir-ch:
SGTM. If we do this change, then I would change also the `Grad` function from
```
Grad func(grad []float64, x []float64)
```
to
```
Grad func(x []float64) []float64
```

kortschak:
Is this closed by #955?

btracey:
Yes. The dream is dead

kortschak:
Are there other approaches you could use to get what you want?

btracey:
Not within the current design of optimize, that I see. The problem is that anything that uses the Hessian but wants to stay sparse needs to know more than just that it's a Symmetric. It needs more specialized things, like `MulVecTo` or `SolveWithRegularize`. As far as I can see, this is incompatible with `optimize` which by design only has a couple ways of specifying functions. The upside to this is that the methods are guaranteed to be consistent in how they treat `Settings`, since they all use identical code to do so. This also gives a common place to do the evaluations, which is otherwise tricky to do concurrently. 

There are a couple ways to fix this. One is to add a lot of fields to Problem containing all the methods one might want. This of course does not scale. The second is to ditch the idea of `Minimize` altogether, and instead say that every `Method` has its own entry function where it can ask for a type that satisfies a certain interface (Functioner, FunGrader, FunGradHesser, etc.). I feel like the current design is quite usable, and I'm not sure how we would make `Settings` common and useful, but maybe there's a clean design I'm missing.

A third solution, and my current thoughts on the matter, is to generalize the "back-end" of optimize, and then provide different front ends (for which `optimize` is one of them). This is #925 . The general idea is that we (or someone) could have a different package like `sparsenewton`. `sparsenewton` could have it's own `Problem`, which would be something like:
[...]

PR #779, opened 2019-01-03
optimize: Change initialization, remove Needser, and update Problem f…

### LINKED ISSUE #727: optimize: Replace Needser
I think `Needser` needs to be replaced. The fundamental problem is that it's only a one-way communication, but it needs to be two-way. There are a bunch of optimization algorithms that can make use of a part of the `Problem`, if it is available, but do not need to. This is impossible to communicate with the current interface. I think it should be something like

```
type Useser interface {
    Uses(has ProbFuns) (uses ProbFuns, err error)
}
```
where ProbFuns is the name of the current unnamed structure returned by Needser. The error is non-nil if the `Problem` doesn't have the needed methods.

btracey:
…unction calls

We need a better way to express the Hessian function call so that sparse Hessians can be provided. This change updates the Problem function definitions to allow an arbitrary Symmetric matrix. With this change, we need to change how Location is used, so that we do not allocate a SymDense. Once this location is changed, we no longer need Needser to allocate the appropriate memory, and can shift that to initialization, further simplifying the interfaces.

A 'fake' Problem is passed to Method to continue to make it impossible for the Method to call the functions directly.

Fixes #727, #593.

Please take a look.

### DIFF
--- a/optimize/bfgs.go
+++ b/optimize/bfgs.go
@@ -10,6 +10,11 @@ import (
 	"gonum.org/v1/gonum/mat"
 )
 
+var (
+	_ Method      = &BFGS{}
+	_ localMethod = &BFGS{}
+)
+
 // BFGS implements the Broyden–Fletcher–Goldfarb–Shanno optimization method. It
 // is a quasi-Newton method that performs successive rank-one updates to an
 // estimate of the inverse Hessian of the objective function. It exhibits
@@ -46,10 +51,13 @@ func (b *BFGS) Status() (Status, error) {
 	return b.status, b.err
 }
 
-func (b *BFGS) Init(dim, tasks int) int {
+func (b *BFGS) Init(dim, tasks int, prob *Problem) (int, error) {
 	b.status = NotTerminated
 	b.err = nil
-	return 1
+	if prob.Grad == nil {
+		return 1, ErrMissingGrad
+	}
+	return 1, nil
 }
 
 func (b *BFGS) Run(operation chan<- Task, result <-chan Task, tasks []Task) {
--- a/optimize/minimize.go
+++ b/optimize/minimize.go
@@ -33,17 +33,32 @@ type Task struct {
 	*Location
 }
 
+// Location represents a location in the optimization procedure.
+type Location struct {
+	X        []float64
+	F        float64
+	Gradient []float64
+	Hessian  mat.Symmetric
+}
+
 // Method is a type which can search for an optimum of an objective function.
 type Method interface {
-	Needser
-	// Init takes as input the problem dimension and number of available
-	// concurrent tasks, and returns the number of concurrent processes to be used.
-	// The returned value must be less than or equal to tasks.
-	Init(dim, tasks int) int
+	// Init initializes the method for optimization. The first two inputs are
+	// the problem dimension and number of available concurrent tasks. The third
+	// input communicates to the Method the available functions to call. The
+	// fields of method are non-nil if the function is present, and nil otherwise.
+	// Init should return an error if the fields of Problem do not match the
+	// requirements of the Method.
+	// Note that the functions passed in are _NOT_ the actual functions specified
+	// in problem, and should not be called.
+	//
+	// Init returns the number of current processes to use, which must be
+	// less than or equal to tasks.
+	Init(dim, tasks int, has *Problem) (int, error)
 	// Run runs an optimization. The method sends Tasks on
 	// the operation channel (for performing function evaluations, major
 	// iterations, etc.). The result of the tasks will be returned on Result.
-	// See the documentation for Operation types for the possible tasks.
+	// See the documentation for Operation types for the possible operations.
 	//
 	// The caller of Run will signal the termination of the optimization
 	// (i.e. convergence from user settings) by sending a task with a PostIteration
@@ -69,6 +84,13 @@ type Method interface {
 	// the values should be used, otherwise the Evaluation operations will be
 	// composed to specify the valid fields. Methods are free to use or
 	// ignore these values.
+	//
+	// Successful execution of an Operation typically requires the Method to
+	// modify the X field of Location, and MajorIteration operations may use
+	// the information in the remaining fields. Methods are encouraged to leave
+	// the remaining fields untouched so that the memory may be re-used by the
+	// problem. Methods should not allocate memory themselves for a Location,
+	// instead the memory should be left untouched or set to nil.
 	//
 	// Method may have its own specific convergence criteria, which can
 	// be communicated using a MethodDone operation. This will trigger a
@@ -131,15 +153,15 @@ func Minimize(p Problem, initX []float64, settings *Settings, method Method) (*R
 	}
 	stats := &Stats{}
 	dim := len(initX)
-	err := checkOptimization(p, dim, method, settings.Recorder)
+	err := checkOptimization(p, dim, settings.Recorder)
 	if err != nil {
 		return nil, err
 	}
 
-	optLoc := newLocation(dim, method)
+	optLoc := newLocation(dim) // This must have an allocated X field.
 	optLoc.F = math.Inf(1)
 
-	initOp, initLoc := getInitLocation(dim, initX, settings.InitValues, method)
+	initOp, initLoc := getInitLocation(dim, initX, settings.InitValues)
 
 	converger := settings.Converger
 	if converger == nil {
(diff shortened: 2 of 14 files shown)


btracey:
@vladimir-ch @sbinet (and maybe @kortschak  for memory allocation discussion)

This is still a WIP, but there needs to be a lot of modifications to `optimize/functions` in order for the tests to pass. I want to get feedback here before going through that effort.

To recap, the major problem is that our current implementation forces allocations of `mat.SymDense` of size `dim x dim`, which can have a very significant cost. Many large-scale problems have a sparse Hessian matrix, and we want optimize to be able to be used for large-scale problems with a known Hessian.

I think there are two main choices for Problem. One way is implemented here, where the memory can be passed back to `Problem` allowing reuse. The other was is to not allow reuse.
```
type Problem struct {
    Func(x []float64) float64
    Grad(x []float64) []float64
    Hess(x []float64) mat.Symmetric
}
```
The advantage here is that the signature is simpler (though not simpler to implement, since one could always ignore the input data). The second advantage is that with this implementation, we could also change the signatures of `diff/fd` to fix #223 . I do not think it is possible to fix #223 while re-using memory, as `fd` needs to take in a `*mat.SymDense`, while `Problem.Hessian` needs to be agnostic to the implementation. If we choose this second representation, which does not re-use memory,  I do not think it is possible to re-use memory behind the scenes by being clever. The Problem does not know when the memory is effectively "freed", and so a pool of some sort cannot be used (when would the data be added back into the pool?).

I don't know how to weigh the considerations. I lean toward the implementation in this PR. The extra complexity in the signatures is in-line with many Gonum functions, and so shouldn't be surprising to (non-new) users. Second, there will still be a mismatch between `problem.Hess` and `fd.Hessian` since `fd.Hessian` will still need a settings structure, and so a wrapper will be needed in any case. I think the ability to re-use memory could be significant, especially for large problems where there is a known and fixed sparsity structure. I could be convinced though. 

For reference, with this implementation, the similar wrapper would need to be
```
p.Hess = func(hess mat.Symmetric, x []float64) mat.Symmetric {
        if hess == nil {
            return fd.Hessian(nil, f.fct, x, nil)
       }
	return fd.Hessian(hess.(*mat.SymDense), f.fct, x, nil)
}
```
We could provide this as a convenience function in `diff/fd` to make that easier. This could look something along the lines of `problem.Hess = fd.HessOpt{f.fct, settings}.Hess`  (happy to detail this more)

sbinet:
@pa-m may have also some interesting feedback (from implementing sklearn)

btracey:
@kortschak the core thing I'd like your opinion on is the choice between 

```
type Problem struct {
    Func(x []float64) float64
    Grad(x []float64) []float64
    Hess(x []float64) mat.Symmetric
}
```
and

```
type Problem struct {
    Func(x []float64) float64
    Grad(grad, x []float64) []float64
    Hess(hess mat.Symmetric, x []float64) mat.Symmetric
}
```

In the first case, the functions always have to allocate their memory. The memory is stored in `minimize` and possibly the `Method`, so there is no safe way to re-use memory. In the second case, users have the option to reuse `grad`/`hess` if they are non-nil. We will guarantee that the type passed into `Hess` is the same as returned from `Hess`, so a user could use a type assertion to get back their particular format.

kortschak:
The reusability of allocations seems fair. What are the performance impacts?

btracey:
None in optimize itself. Extra if statement if we allow memory re-use and the function decides to do that

btracey:
I drew up a benchmark case, but the performance impacts are negligible. I think there's enough other places to not make it matter so much

optimize/bfgs.go
```diff
@@ -10,6 +10,11 @@ import (
 	"gonum.org/v1/gonum/mat"
 )
 
+var (
+	_ Method      = &BFGS{}
```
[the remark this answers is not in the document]
btracey: Done.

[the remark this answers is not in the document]
btracey: Done.

optimize/minimize.go
```diff
@@ -33,17 +33,32 @@ type Task struct {
...
-	// concurrent tasks, and returns the number of concurrent processes to be used.
-	// The returned value must be less than or equal to tasks.
-	Init(dim, tasks int) int
+	// Init initializes the method for optimization. The first two inputs are
+	// the problem dimension and number of available concurrent tasks. The third
+	// input communicates to the Method the available functions to call. The
+	// fields of method are non-nil if the function is present, and nil otherwise.
+	// Init should return an error if the fields of Problem do not match the
+	// requirements of the Method.
+	// Note that the functions passed in are _NOT_ the actual functions specified
```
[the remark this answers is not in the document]
vladimir-ch: I'm not a fan of this either, most of the `Init` doc comment is about what `has` is which feels disproportionate. Also the error returned from Init is not really actionable for the caller; if they have the gradient (for example) they could have provided it in the first place. The error could be used to select a suitable method from a set that can handle a given problem but then `Needs` could fulfill the same purpose without having to call Init and then inspecting the error. `ProblemContains` would offset this by not passing a fake problem but still I'm not convinced about the benefits.

optimize/types.go
```diff
@@ -132,13 +124,17 @@ type Problem struct {
...
+	// (and return) the provided slice if it is non-nil, or must allocate a new
+	// slice otherwise.
+	Grad func(grad []float64, x []float64) []float64
 
 	// Hess evaluates the Hessian at x and stores the result in-place in hess.
-	// Hess must not modify x.
-	Hess func(hess mat.MutableSymmetric, x []float64)
+	// Hess must not modify x. Hess may use (and return) the provided Symmetric
+	// if it is non-nil, or must allocate a new Symmetric otherwise. Minimize
+	// will 'give back' the returned Symmetric where possible, allowing Hess
```
vladimir-ch: Can we make the statement here stronger than "where possible"?

optimize/minimize.go
```diff
@@ -70,6 +85,13 @@ type Method interface {
 	// composed to specify the valid fields. Methods are free to use or
 	// ignore these values.
 	//
+	// Successful execution of an Operation typically requires the Method to
+	// modify the X field of Location, and MajorIteration operations may use
+	// the information in the remaining fields. Methods are encouraged to leave
+	// the remaining fields untouched so that the memory may be re-used by the
+	// problem. Methods should not allocate memory themselves for a Location,
+	// instead the memory should be left untouched or set to nil.
```
vladimir-ch: Is the repeated "untouched" intentional or is it an editing artifact?

vladimir-ch:
Overall this looks good to me once we figure out the `ProblemContains`/`Init` question. I agree with the memory-reusing form of `Problem`.

optimize/minimize.go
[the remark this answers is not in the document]
btracey: I changed this to remove a lot of the subtlety of documentation, and to go back to the style close to `Needs`. 

I think we need some kind of communication about what the problem contains in the long term. The package as it currently stands does not need it, but imagine if we had a `SolveVecHess` (or equivalent). `Newton` can be smart about using that (or not) depending on which functions are present, but it has to be told what is present. The same is going to be true for constrained problems. Standard QP solvers, for instance, have special algorithms if there are only box constraints, and so they need to know what is present.

I'm not totally convinced we need the reporting, but our tests find it useful, so can't hurt.

optimize/minimize.go
btracey: Made better comment.

optimize/types.go
btracey: I don't think so, if a Method needs to store Hessian information for some reason, if we guarantee memory re-use then the Method needs to copy.

optimize/local_example_test.go
```diff
@@ -27,12 +27,12 @@ func ExampleMinimize() {
...
-	fmt.Printf("result.F: %v\n", result.F)
+	fmt.Printf("result.X: %0.4g\n", result.X)
+	fmt.Printf("result.F: %0.4g\n", result.F)
 	fmt.Printf("result.Stats.FuncEvaluations: %d\n", result.Stats.FuncEvaluations)
 	// Output:
 	// result.Status: GradientThreshold
 	// result.X: [1 1 1 1 1]
-	// result.F: 0
-	// result.Stats.FuncEvaluations: 35
+	// result.F: 4.98e-30
```
btracey: I'm not totally sure why this changed, but the current optimum it's finding has the gradient at 1e-14, so it's pretty close.

btracey:
One thing that changed is that now the optimum location will only contain F, X, and G, but not the Hessian (because otherwise we cannot avoid the N^2). However, apparently we have always explicitly stated that only F, X and G are valid, and so I don't feel bad about making the documentation true.

btracey:
PTAL @vladimir-ch

PR #780, opened 2019-01-04
lapack/gonum,mat: add Dpotri and use it in Cholesky

vladimir-ch:
`Dlauum` has a use after all.

Please take a look.

```
name                        old time/op  new time/op  delta
CholeskyInverseTo/n=10-4    10.8µs ± 2%   3.2µs ± 1%  -70.16%  (p=0.008 n=5+5)
CholeskyInverseTo/n=100-4   1.07ms ± 2%  0.51ms ± 2%  -52.06%  (p=0.008 n=5+5)
CholeskyInverseTo/n=1000-4   713ms ± 1%   315ms ± 1%  -55.83%  (p=0.008 n=5+5)
```


kortschak:
Nice.

PR #781, opened 2019-01-04
stat/distuv: canonicalize stat tests

### LINKED ISSUE #182: stat: improve use of consts for test thresholds
There is a wide variety of approaches to specifying test thresholds in stat, ranging from consts, to vars and numeric literals at the call site. It would be good to have these all as consts in a func preamble.

btracey:
Updates the tests to use a consistent form, and adds tests to a few distributions that are missing them.

Fixes #182.

Please take a look.

### DIFF
--- a/stat/distuv/exponential_test.go
+++ b/stat/distuv/exponential_test.go
@@ -6,7 +6,10 @@ package distuv
 
 import (
 	"math"
+	"sort"
 	"testing"
+
+	"golang.org/x/exp/rand"
 )
 
 func TestExponentialProb(t *testing.T) {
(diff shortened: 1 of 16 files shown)


stat/distuv/exponential_test.go
```diff
@@ -43,6 +46,38 @@ func TestExponentialFitPrior(t *testing.T) {
 	testConjugateUpdate(t, func() ConjugateUpdater { return &Exponential{Rate: 13.7} })
 }
 
+func TestExponential(t *testing.T) {
+	src := rand.New(rand.NewSource(1))
+	for i, dist := range []Exponential{
+		{3, src},
```
kortschak: Field names.

btracey: Done.

Issue #185, opened 2017-08-08
blas/gonum: add complex BLAS Level 3 functions


vladimir-ch:
- [ ] ZGEMM - matrix matrix multiply
- [ ] ZSYMM - symmetric matrix matrix multiply
- [ ] ZHEMM - hermitian matrix matrix multiply
- [ ] ZSYRK - symmetric rank-k update to a matrix
- [ ] ZHERK - hermitian rank-k update to a matrix
- [ ] ZSYR2K - symmetric rank-2k update to a matrix
- [ ] ZHER2K - hermitian rank-2k update to a matrix
- [ ] ZTRMM - triangular matrix matrix multiply
- [ ] ZTRSM - solving triangular matrix with multiple right hand sides

PR #785, opened 2019-01-08
blas/gonum: add Zgemm, Zsyrk, and Zherk

vladimir-ch:
Also fixes a bug in Dsyrk.

Updates #185 

Please take a look.

### DIFF
--- a/blas/gonum/level3cmplx128.go
+++ b/blas/gonum/level3cmplx128.go
@@ -8,6 +8,7 @@ import (
 	"math/cmplx"
 
 	"gonum.org/v1/gonum/blas"
+	"gonum.org/v1/gonum/internal/asm/c128"
 )
 
 var _ blas.Complex128Level3 = Implementation{}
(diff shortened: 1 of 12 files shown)


blas/gonum/level3double.go
```diff
@@ -463,17 +463,31 @@ func (Implementation) Dsyrk(ul blas.Uplo, tA blas.Transpose, n, k int, alpha flo
 			for i := 0; i < n; i++ {
 				ctmp := c[i*ldc+i : i*ldc+n]
 				atmp := a[i*lda : i*lda+k]
-				for jc, vc := range ctmp {
```
kortschak: How did this present?

blas/gonum/level3cmplx128.go
```diff
@@ -254,6 +254,173 @@ func (Implementation) Zgemm(tA, tB blas.Transpose, m, n, k int, alpha complex128
...
+		}
+		return
+	}
+
+	calpha := complex(alpha, 0)
+	if trans == blas.NoTrans {
+		// Form  C = alpha*A*A^H + beta*C.
+		cbeta := complex(beta, 0)
+		if uplo == blas.Upper {
+			for i := 0; i < n; i++ {
```
kortschak: Cases for beta==0?

blas/gonum/level3double.go
```diff
@@ -498,7 +516,11 @@ func (Implementation) Dsyrk(ul blas.Uplo, tA blas.Transpose, n, k int, alpha flo
 	}
 	for i := 0; i < n; i++ {
 		ctmp := c[i*ldc : i*ldc+i+1]
-		if beta != 0 {
```
vladimir-ch: The actual bug was here, the comparison `beta != 0` should have been `beta != 1`. I added the 0 special case (optimization) together with a fix, and for consistency I added it also to the other paths, all lumped into one commit. I'm sorry for taking this shortcut, I can still split it.

blas/gonum/level3double.go
vladimir-ch: As I comment below, the bug was not here.

blas/gonum/level3cmplx128.go
vladimir-ch: I'll add them.

kortschak: This is important because of NaN behaviour.

blas/gonum/level3double.go
kortschak: No need to split it, but can you add a note to the commit message explaining the issue and what was done?

blas/gonum/level3cmplx128.go
vladimir-ch: I split it anyway, I think it's clearer this way. Also added a more descriptive message to the fixing commit.

blas/gonum/level3double.go
vladimir-ch: I split it anyway, I think it's clearer this way. Also added a more descriptive message to the fixing commit.

PR #786, opened 2019-01-10
add Ch. Meessen to AUTHORS and CONTRIBUTORS

chmike:
Please take a look.


sbinet:
needed by https://github.com/gonum/plot/pull/498

kortschak:
It would have been fine to amends and force push here, but I'll approve the other.

chmike:
Sorry. I have the habit to not amend pushed commits. I wasn't sure of the effect on the PR. I never did it before.

kortschak:
On PR branches it's fine. We have protection on the master branch, so we won't break people.

PR #791, opened 2019-01-12
path: use node ID in dominator tree for node retrieval

mewmew:
Updates #31.

Please take a look.

### DIFF
--- a/graph/path/control_flow_lt.go
+++ b/graph/path/control_flow_lt.go
@@ -197,13 +197,13 @@ type DominatorTree struct {
 // Root returns the root of the tree.
 func (d DominatorTree) Root() graph.Node { return d.root }
 
-// DominatorOf returns the immediate dominator of n.
-func (d DominatorTree) DominatorOf(n graph.Node) graph.Node {
-	return d.dominatorOf[n.ID()]
+// DominatorOf returns the immediate dominator of the node with the given ID.
+func (d DominatorTree) DominatorOf(id int64) graph.Node {
+	return d.dominatorOf[id]
 }
 
-// DominatedBy returns a slice of all nodes immediately dominated by n.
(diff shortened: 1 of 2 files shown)


kortschak:
Probably we should have some tests here?

mewmew:
The dom tree has test cases in `control_flow_test.go`. It uses reflect instead of directly invoking `DominatorOf` and `DominatedBy` it seems.

kortschak:
Yeah, I saw that. The accessors are pretty simple, but it would be good to have something that ensures we're being honest in the future.

Replace what exists with this
```
			for q, want := range test.want.dominatorOf {
				node := got.DominatorOf(q)
				if node != want {
					t.Errorf("unexpected dominator tree result from %s dominated of %v: got:%v want:%v",
						alg.name, q, node, want)
				}
			}

			for _, nodes := range got.dominatedBy {
				sort.Sort(ordered.ByID(nodes))
			}
			for q, want := range test.want.dominatedBy {
				nodes := got.DominatedBy(q)
				if !reflect.DeepEqual(nodes, want) {
					t.Errorf("unexpected dominator tree result from %s dominated by %v: got:%v want:%v",
						alg.name, q, nodes, want)
				}
			}
```

mewmew:
> Replace what exists with this

Thanks! I update the test case with the code you wrote.

kortschak:
I should have left the tree comparison in in addition to the new tests; now we confirm that the result we see is a subset of the expectation, but don't confirm that it's not a proper superset. I'll send a fix.

mewmew:
> I should have left the tree comparison in in addition to the new tests; now we confirm that the result we see is a subset of the expectation, but don't confirm that it's not a proper superset. I'll send a fix.

Yeah, that's better. Thanks for submitting the PR.

PR #795, opened 2019-01-14
encoding/dot: ensure that generated DOT is syntactically valid

kortschak:
Please take a look.

I would like to add the same check to the graphql tests, but they currently fail and will do so until the quoting PR is merged. So I'll leave until after that (or you can add it in there - unfortunately we need to just copy/paste this check function).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,6 +34,11 @@ before_install:
  # Required for dot parser checks.
  - ./.travis/install-gocc.sh 0e2cfc030005b281b2e5a2de04fa7fe1d5063722
 
+addons:
+  apt:
+    packages:
+     - graphviz
+
 go_import_path: gonum.org/v1/gonum
 
 # Get deps, build, test, and ensure the code is gofmt'ed.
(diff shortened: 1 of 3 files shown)


kortschak:
I may need to alter `.travis.yml` for this to do anything. Running one set of tests to check.

mewmew:
> I would like to add the same check to the graphql tests, but they currently fail and will do so until the quoting PR is merged. So I'll leave until after that (or you can add it in there - unfortunately we need to just copy/paste this check function).

Lets add it after the first PR is merged. Otherwise these PRs depend on one another.

kortschak:
PTAL

kortschak:
It's worth waiting for the owner to merge (I was bad yesterday with yours); at one point when this looked OK and was approved I noticed an error in it which would have needed a follow-up PR to fix.

mewmew:
>It's worth waiting for the owner to merge (I was bad yesterday with yours); at one point when this looked OK and was approved I noticed an error in it which would have needed a follow-up PR to fix.

Sure. So the idea is that we review PRs, and then let the owner of the PR merge once approved by one or more Gonum members?

kortschak:
Yes, please. Obvious exceptions are when the author has said that someone else can merge or the author is unable to merge.

kortschak:
Also, branch deletion is up to the author.

PR #798, opened 2019-01-14
blas/gonum: add Zsymm and Zhemm

vladimir-ch:
Updates #185 

Please take a look.


vladimir-ch:
Ran the generation to fix the failures, please take a look.

PR #800, opened 2019-01-15
reduce travis CI time

### LINKED ISSUE #799: all: recent (?) increase in time to run tests on Travis
not sure how critical this is (but having an "agile" CI infrastructure is really important to me): it seems each entry in the Travis build matrix is taking about 20mn to build+test.
the total for each PR is then around 5h.

I think it's a bit much:

```
$> go get -t -v ./...        ##  15s
$> go test -a $TAGS -v ./... ## 454s
```

here is the list of top consumers:

```
ok  	gonum.org/v1/gonum/lapack/gonum	159.983s
ok  	gonum.org/v1/gonum/stat/distuv	127.333s
ok  	gonum.org/v1/gonum/mat	90.025s
ok  	gonum.org/v1/gonum/graph/community	88.905s
ok  	gonum.org/v1/gonum/graph/path	46.357s
ok  	gonum.org/v1/gonum/optimize	40.068s
ok  	gonum.org/v1/gonum/stat/distmv	34.287s
ok  	gonum.org/v1/gonum/optimize/convex/lp	33.817s
ok  	gonum.org/v1/gonum/fourier	22.269s
ok  	gonum.org/v1/gonum/graph/graphs/gen	17.038s
ok  	gonum.org/v1/gonum/mathext	6.792s
ok  	gonum.org/v1/gonum/stat/samplemv	5.381s
ok  	gonum.org/v1/gonum/mathext/internal/amos	5.119s
ok  	gonum.org/v1/gonum/stat/sampleuv	3.686s
ok  	gonum.org/v1/gonum/blas/gonum	2.841s
ok  	gonum.org/v1/gonum/integrate	2.339s
ok  	gonum.org/v1/gonum/graph/topo	1.148s
```

extracted (manually -- some errors not unexpected) from https://travis-ci.org/gonum/gonum/jobs/479273645

perhaps something could be done to somewhat reduce the amount of time it takes to run all these tests?

- use a build cache (this will prevent re-running tests across PRs) ([example from Go-HEP](https://github.com/go-hep/hep/blob/master/.travis.yml#L9) where I also had to use this feature to decrease the PR turn-around-time)
- optimize the tests
- reduce the amount of tests

WDYT?

BTW, shouldn't these lines:
```yml
script:
 - go get -d -t -v ./...
 - go build -v ./...
 - go test -a $TAGS -v ./...
```
read instead:
```yml
script:
 - go get -d -t $TAGS -v ./...
 - go build $TAGS -v ./...
 - go test -a $TAGS -v ./...
```
?

sbinet:
Fixes #799.


sbinet:
on my `sbinet-gonum` travis dashboard, a complete build (+tests) went from `2 hrs 24 min 41 sec` to `1 hr 40 min 49 sec`

sbinet:
and, on the `gonum/gonum` travis CI:
from `5 hrs 19 min 32 sec`
to `1 hr 25 min 10 sec`

.travis.yml
```diff
@@ -21,6 +21,9 @@ cache:
    - $HOME/.cache/go-build
    - $HOME/gopath/pkg/mod
 
+git:
+ depth: 10
```
kortschak: Do we need it to be this deep? Isn't 1 enough?

.travis.yml
```diff
@@ -10,12 +10,6 @@ go:
  - 1.11.x
  - master
 
-env:
```
kortschak: Please revert this. While we can get around what this was introduced to fix with the new cache, in cases where there are gonum-wide changes, this will cause a timeout.

.travis.yml
```diff
@@ -45,10 +45,10 @@ script:
  - ${TRAVIS_BUILD_DIR}/.travis/check-formatting.sh
  - go get -d -t -v ./...
  - go build -v ./...
- - go test -v ./...
```
kortschak: Please revert this. This is useful for me when we are debugging failures.

.travis.yml
sbinet: ok

.travis.yml
sbinet: not sure about that (the timeout).
I mean: the cache helps to shrink the build time from PR to PR.
within a PR, adding or removing the `cache:` directive won't help a bit.

also: this very PR was a gonum-wide change as far as it was concerned: everything got rebuilt.
besides, a PR times out when at least one of its build-matrix elements takes more than 1h or when there's no printout for more than 10mn (IIRC.)

leveraging the builtin Go build/test cache and running the 4 `go test -tags=XXX` in a row, in the same PR, should make sure we stay within time budget.

.travis.yml
sbinet: what part do you want reverted? dropping the `-a` ? (but then this defeats the Go builtin cache mechanism, which is now (?) robust)

kortschak: The `-v`. Note that it's difficult to review when the file state go away with force push.

sbinet: running 4 `go test` with `-v` generates way too many lines for Travis (it ends up cutting them in the web display, allowing one to download the raw log file.)

in general, I find that while `-v` for `go (get|install|build)` is fine, for `go test` it usually is wrong after ~30 tests.

kortschak: Which is one of the reason why I want the matrix with TAGS back in https://github.com/gonum/gonum/pull/800#discussion_r248054735.

.travis.yml
kortschak: Divinding the work into tag-based units also helps quickly see what is a cause. If you cannot have this and the cache, then this needs further discussion, but separating the different builds is useful.

sbinet: (moving this discussion to the main PR thread)

sbinet:
> Dividing the work into tag-based units also helps quickly see what is a cause. If you cannot have this and the cache, then this needs further discussion, but separating the different builds is useful.

> Which is one of the reason why I want the matrix with TAGS back in #800 (comment).

the main issue with the `TAGS` matrix is that it multiplies the build variants *and* results in `nbrOf(TAGS) x nbrOf(Go-version)` build slots.
as the number of build slots is fixed on a per-organization basis, people hacking on `gonum/gonum` impact the ability of:
- other people seeing the result of their PR on some other package under `v1/gonum`,
- but _also_ for other people hacking on `v1/plot`.

_e.g._ right now, my PR about https://github.com/gonum/plot/pull/493 is waiting for:
- the builds (16) for `gonum/gonum#803` to complete
- the builds (4) for `gonum/plot#492` (which has been merged into `master`) to complete

_(of course I am not picking on anybody's PR, it just happens `#803` and `#492` are being built :P)_

IIRC, TravisCI allows the `gonum` project to have 5 builds in parallel (but a PR has to be completely "drained" of its "not-allowed-to-fail" slots before Travis allocates build slots to other PRs.)

right now, we are about ~5+/-2 "active" developers.
if somebody has to wait for 5h before her PR is tested, this is getting pretty frustrating quite rapidly and this doesn't really scale.

getting rid of the multiplicity introduced by `$TAGS` completely, reducing the number of `$TAGS` (perhaps merging `noasm` and `appengine` into a single `purego` ?) and introducing TravisCI's build cache are the first low-hanging fruits I could find.

(side note: it seems we don't test with `-race` enabled...)

kortschak:
Perhaps I don't understand how the cache works, but it seems to me from reading the docs (Go, not Travis), that no caching is done when run tests with any `-tag $TAG` args.

```
The rule for a match in the cache is that the run involves the same
test binary and the flags on the command line come entirely from a
restricted set of 'cacheable' test flags, defined as -cpu, -list,
-parallel, -run, -short, and -v. If a run of go test has any test
or non-test flags outside this set, the result is not cached.
```

This means, AFAICS, that the first line, `go test ./...` is cached and there is a potential saving there, but the following three `bounds`, `noasm` and `appengine` are not. (Note here that `noasm` and `appengine` are distinct - `noasm` allows unsafe which we use in places, `safe`/`appengine` do not).

I don't really see how there is a 5h impact. The CI is there to prevent us from merging things where the local testing has fallen down or (as is sometimes the case) where numerical stability problems show up due to hardware differences. It shouldn't be the only testing that's done. This means that it doesn't impact on throughput, but rather only on latency. And human reviewers are a much greater source of PR merge latency than testing ever will or should be.

>(side note: it seems we don't test with `-race` enabled...)

We should, but that will blow out the time budget further.

kortschak:
From my reading of the docs, if you change `go test -a $TAGS -v ./...` to `go test $TAGS -v ./...` and don't do the other changes you have, you will get the same benefit; the cacheable build will be cached and they will finish faster, the uncacheable builds will lag, but you will get the overall release of workers in the same overall time as your current case.

sbinet:
as a rough estimation of the impact of caching (extracting from [this PR](https://travis-ci.org/gonum/gonum/jobs/480313372)):
```
 Go: 1.9.x  33 min 27 sec
 Go: 1.10.x 13 min 46 sec
 Go: 1.11.x 13 min 26 sec
```
(IIRC, caching was introduced with Go1.10)

a somewhat more detailed view:
- this PR, Go1.11.x [here](https://travis-ci.org/gonum/gonum/jobs/480313372)
```
go test ./...                 --> 198.66s
go test -tags=bounds ./...    --> 115.49s
go test -tags=noasm ./...     --> 254.62s
go test -tags=appengine ./... --> 138.00s
```

- compared to [this PR](https://travis-ci.org/gonum/gonum/builds/480299593)
```
go test -a ./...                 --> 446.20s
go test -a -tags=bounds ./...    --> 488.38s
go test -a -tags=noasm ./...     --> 537.11s
go test -a -tags=appengine ./... --> 516.06s
```
(clearly *not* an apple-to-apple comparison because of the `-a` switch, but...)

btracey:
I looked up the Travis pricing plans, and it seems like the number of
parallel builds is the primary distinguisher between price level, so it
doesn't seem like it would be easy to get, say, 20, but we could ask

sbinet:
@btracey alternatively, we could ask the Centre de Calcul IN2P3 (basically my lab's computing resources) to host the CI infrastructure (they're using GitLab-CI...)

kortschak:
BTW What time are you reporting above. I've just run a test set and have "Ran for 19 min 31 sec". The total CPU wall time is not really relevant here.

sbinet:
superseded by #808.

PR #801, opened 2019-01-16
blas/gonum: add Ztrmm and Ztrsm

### LINKED ISSUE #185: blas/gonum: add complex BLAS Level 3 functions

vladimir-ch:
Closes #185 (sound of fanfares in the distance)

Also sneaked in some cleanup changes in Dtrmm and Dtrsm, including fixes to matrix strides in Dtrsm. Unfortunately the tests would require non-small changes to fail on this.

Please take a look.

### DIFF
--- a/blas/gonum/level3double.go
+++ b/blas/gonum/level3double.go
@@ -759,14 +759,11 @@ func (Implementation) Dtrmm(s blas.Side, ul blas.Uplo, tA blas.Transpose, d blas
 						tmp *= a[i*lda+i]
 					}
 					btmp := b[i*ldb : i*ldb+n]
-					for j := range btmp {
-						btmp[j] *= tmp
-					}
+					f64.ScalUnitary(tmp, btmp)
 					for ka, va := range a[i*lda+i+1 : i*lda+m] {
 						k := ka + i + 1
-						tmp := alpha * va
-						if tmp != 0 {
-							f64.AxpyUnitaryTo(btmp, tmp, b[k*ldb:k*ldb+n], btmp)
+						if va != 0 {
+							f64.AxpyUnitary(alpha*va, b[k*ldb:k*ldb+n], btmp)
 						}
 					}
 				}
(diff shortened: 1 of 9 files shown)


blas/gonum/level3double.go
```diff
@@ -172,7 +172,7 @@ func (Implementation) Dtrsm(s blas.Side, ul blas.Uplo, tA blas.Transpose, d blas
 			return
 		}
 		for i := 0; i < m; i++ {
-			btmp := b[i*lda : i*lda+n]
```
kortschak: Oops. How did this not show up before? Can we have an issue to add a test for it if it's not going to be added now.

vladimir-ch: The `blas/gonum` tests have only `lda = ldb` for these two functions, tests in `mat` don't cover them and most tests in `lapack/gonum` that do also have `lda = ldb` except for `Dgetri` which eventually calls `Dtrsm` with `lda < ldb`. So it's not surprising that it didn't show up before although the `Dgetri` makes it a bit mysterious.

I'll see how much work would be to modify the tests to trigger failure without this fix and if not too much, I'll add it to this PR.

Speaking of Dgetri, I'll have to take a closer look at it. Its test uses a very crude tolerance but some quick experiments show that it's necessary only for matrix sizes > 64. Sizes 64 and smaller can bear tolerances in the identity matrix test as low as 1e-14. At 64 the algorithm switches to blocked code but the big jump in the accuracy of the result is disconcerting.

Issue #796, opened 2019-01-14
graph: [proposal] add algorithm for finding intervals in flow graphs


mewmew:
### Background

There exist a number of core concepts in graph theory that are general enough to be useful outside of specific application use cases. Concepts such as strongly connected regions, shortest path, dominance relationships in flow graphs, etc.

Related to flow graphs there is the notion of intervals, which are similar to strongly connected regions, but differ in definition and thus application.

The definition of a an interval is as follows:

From Allen, Frances E. "Control flow analysis." ACM Sigplan Notices. Vol. 5. No. 7. ACM, 1970. (ref: http://www.cs.columbia.edu/~suman/secure_sw_devel/p1-allen.pdf)

> Given a node *h*, an interval *I(h)* is the maximal, single entry subgraph for which *h* is the entry node and in which all closed paths contain *h*. The unique interval node *h* is called the interval head or simply the header node.

![interval](https://user-images.githubusercontent.com/1414531/51094144-de1f8e80-17aa-11e9-96d5-fd5ca68370d1.png)

### Proposal

Implement algorithm for finding intervals in rooted flow graphs. The specific algorithm is outlined in these papers (with slight variations between the papers, but the output is identical).

* Allen, Frances E., and John Cocke. "A program data flow analysis procedure." Communications of the ACM 19.3 (1976): 137.
* Figure 1 "Interval Algorithm" in C. Cifuentes, "A Structuring Algorithm for Decompilation", 1993.
* Figure 6-8 "Interval Algorithm" in C. Cifuentes, "Reverse Comilation Techniques", 1994.

### Potential impact of proposal

It should be enough to introduce the following API to the `graph/path` package, which already contains control flow graph algorithms (e.g. calculation of dominance trees).

Of course, these API additions are open for suggestions. Mainly plotting them down, so we can get a feel for what it may look like.

```go
// Intervals returns the unique set of intervals in the given control flow
// graph rooted at the node with ID eid.
func Intervals(g graph.Directed, eid int64) []*Interval {
}

// An Interval I(h) is the maximal, single entry subgraph for which h
// is the entry node and in which all closed paths contain h. The
// unique interval node h is called the interval head or simply the
// header node.
type Interval struct {
   // Interval head.
   head graph.Node
   // Interval nodes.
   nodes []graph.Node
}

// Head returns the header node of the interval.
func (i *Interval) Head() graph.Node {
}

// Nodes returns all the nodes in the interval.
[...]

PR #803, opened 2019-01-17
graph/flow: add flow package for control flow analysis algorithms

mewmew:
This PR is a *work in progress*. **DO NOT MERGE**.

It is based on the initial discussions in https://github.com/gonum/gonum/issues/796 and the intention is to arrive at an API that control flow analysis algorithms can unify around. Specifically, we define the notion of a `cfa.Graph`, a control flow graph rooted at an entry node.

This PR also adds the initial API scaffolding for finding the unique set of intervals in a control flow graph.

Follow-up PRs (or extensions to this PR) will move other control flow analysis algorithms to the `cfa` package, and update their API to make use of the `cfa.Graph` interface. Specifically, this relates to the `path.DominatorTree` API.

```diff
-func Dominators(root graph.Node, g graph.Directed) DominatorTree
+func Dominators(g Graph) DominatorTree
```

Please take a look.


vladimir-ch:
I like "flow" too.

graph/flow/flow.go
```diff
@@ -0,0 +1,32 @@
+// Package flow provides control flow analysis functions.
```
kortschak: Copyright header.

graph/flow/flow.go
```diff
@@ -0,0 +1,32 @@
...
+package flow
```
kortschak: Import comment when ready.

graph/flow/flow.go
```diff
@@ -0,0 +1,32 @@
...
+
+import "gonum.org/v1/gonum/graph"
+
+// Graph is a control flow graph, a directed graph rooted at an entry node.
+type Graph interface {
+	// Entry returns the entry node of the control flow graph.
+	Entry() graph.Node
```
kortschak: `Root`?

graph/flow/flow.go
```diff
@@ -0,0 +1,32 @@
...
+	graph.Directed
+}
+
+// NewGraph returns a new control flow graph based on the given directed graph
+// and entry node ID.
+func NewGraph(g graph.Directed, entryid int64) Graph {
```
kortschak: s/entryid/rid/

Also, I'm not a fan of returning an interface type from a construction func and given the simplicity of this function, ISTM that just providing a type with the relevant exposed fields would do the job.

graph/flow/flow.go
```diff
@@ -0,0 +1,32 @@
...
+	return &cfg{
```
kortschak: This doesn't need to be a pointer.

graph/flow/interval.go
```diff
@@ -0,0 +1,77 @@
+// Find the unique set of intervals of a control flow graph, as described in
```
kortschak: Copyright header.

graph/flow/interval.go
```diff
@@ -0,0 +1,77 @@
...
+
+// Node returns the node with the given ID if it exists within the interval, and
+// nil otherwise.
+func (i *Interval) Node(id int64) graph.Node {
+	panic("not yet implemented")
+}
+
+// Nodes returns all the nodes of the interval.
+//
+// Nodes must not return nil.
```
kortschak: The requirement statement here is not necessary.

graph/flow/interval.go
```diff
@@ -0,0 +1,77 @@
...
+func (i *Interval) Nodes() graph.Nodes {
+	panic("not yet implemented")
+}
+
+// From returns all nodes that can be reached directly from the node with the
+// given ID within the interval.
+//
+// From must not return nil.
```
kortschak: The requirement statement here is not necessary.

graph/flow/interval.go
```diff
@@ -0,0 +1,77 @@
...
+// HasEdgeFromTo returns whether an edge exists within the interval from u to v
+// with IDs uid and vid.
+func (i *Interval) HasEdgeFromTo(uid, vid int64) bool {
+	panic("not yet implemented")
+}
+
+// To returns all nodes that can reach directly to the node with the given ID
+// within the interval.
+//
+// To must not return nil.
```
kortschak: The requirement statement here is not necessary.

graph/flow/interval_test.go
```diff
@@ -0,0 +1,176 @@
+package flow
```
kortschak: Copyright header.

graph/flow/interval.go
```diff
@@ -0,0 +1,77 @@
...
+// [1]: https://amturing.acm.org/p137-allen.pdf
+
+package flow
+
+import "gonum.org/v1/gonum/graph"
+
+// Intervals returns the unique set of intervals of the given control flow
+// graph. The returned set of intervals share the underlying nodes and edges of
+// the provided control flow graph.
+func Intervals(g Graph) []*Interval {
```
kortschak: Thinking about the API that you need to add above to allow a `graph.Directed` to be used here, ISTM that this should just be `func Intervals(root id, g graph.Directed) []*Interval`.

When `Dominators` is brought into flow, the signature should be updated to take an ID rather than a `graph.Node`.

graph/flow/flow.go
mewmew: Sure, we can go for `Root` instead of `Entry`. In the literature I've mostly taken part of, the term *entry node* or *entry block* is used for control flow graphs. And the graph is a directed graph, rooted at the entry node.

graph/flow/flow.go
mewmew: Fair point, that would also work.

graph/flow/interval.go
mewmew: Yes, that makes sense. It would make the API less restrictive, as users would then not have to create a flow.Graph simply to use the function.

graph/flow/flow.go
mewmew: One question that arise. Do we even need `flow.Graph` at all? The reason I added it was so that functions taking `g graph,Directed, root int64` would instead take `g flow.Graph`. However, with your suggestions, we'd still use the first approach for the API, and then we probably won't have a reason to keep `flow.Graph`.

kortschak: Yes, I did think about whether it should go away, but keeping them bundled once you have them is nice. You have come to the same conclusion.

mewmew:
Two issues remaining since your last review comments @kortschak. Namely,

1. whether to call the entry node/root node `Root` or `Entry`, with `rid`, `eid` for IDs.
2. whether to keep `flow.Graph` at all, and if so, to make it into a concrete type.

```go
// Graph is a control flow graph, a directed graph rooted at an entry node.
type Graph struct {
   // underlying directed graph.
   graph.Directed
   // Root node of the control flow graph.
   Root graph.Node
```

Whether we should use `graph.Node` or just an `int64` ID is not yet clear to me. That is of course if we keep the `flow.Graph` type at all.

kortschak:
Sorry for the delay. I missed this.

1. I prefer root, but I think that's possibly an aesthetic thing.
2. A concrete type as you have there seems pretty reasonable. I think that since we have `Node(int64) graph.Node` we probably don't need the root/entry to be stored as a `graph.Node` (having a single truth holder is a value here). We could though add a helper method to the type, `Root() graph.Node`/`Entry() graph.Node` the just wraps `g.Directed.Node(g.RootID)`/`g.Directed.Node(g.EntryID)` purely for convenience.

mewmew:
> Sorry for the delay. I missed this.

No worries at all. I'm currently on travelling foot so the delay came at a convenient time.

1. Sure, lets go with Root if you so prefer.
2. Ok, concrete struct it is. With `graph.Directed` embedded and a `RootID int64` field. And, yes. It seems reasonable to add a convenience method for `Root() graph.Node` that warps `g.Directed.Node(g.RootID)`.

kortschak:
@mewmew Are there plans to continue this?

mewmew:
> @mewmew Are there plans to continue this?

I'd like to continue working on this in fall. If anything is blocked by this, just disregard this PR for now. I mainly intend to implement the algorithm for locating Intervals.

Yesterday was my last day in Korea, and the week prior was the week for final exams. Now, I'm travelling throughout Mongolia.

Wish you a happy summer Dan!

kortschak:
No worries. I hope you enjoyed Korea and have a good time in Mongolia.

kortschak:
@mewmew Are you still interested in this?

mewmew:
> @mewmew Are you still interested in this?

Definitely! I think it makes more sense though to close this PR and do a new one where we can discuss the API in advance and decide what should go into `graph/flow`.

kortschak:
No worries. I'll leave it to you to close this and open your new approach.

vladimir-ch:

PR #804, opened 2019-01-17
all: add AppVeyor Windows CI

sbinet:
Please take a look.


appveyor.yml
```diff
@@ -0,0 +1,22 @@
...
+
+branches:
+  only:
+    - master
+
+environment:
+  GOPATH: c:\gopath
+  PATH: '%GOPATH%\bin;%PATH%'
+
+stack: go 1.11
```
kortschak: We should be testing all our supported versions.

sbinet:
PTAL

appveyor.yml
```diff
@@ -0,0 +1,26 @@
...
+clone_folder: c:\gopath\src\gonum.org\v1\gonum
+
+cache:
+  - '%LocalAppData%\go-build'
+
+branches:
+  only:
+    - master
+
+environment:
```
kortschak: It's worth adding a note to not move these lines as exists in the [travis config](https://github.com/gonum/gonum/blob/921c8b9ec6c531bca1ab15b67c348869a53625bd/.travis.yml#L5-L6).

kortschak:
What do we want to do about the passing policy for AppVeyor? This should probably be discussed on your original gonum-dev thread. I am happy for it to be mandatory so long as we don't get too many failures on it or someone else is prepared to debug windows problems,

PR #806, opened 2019-01-17
lapack/gonum: fix various bugs in Dgetri

vladimir-ch:
Please take a look.

### DIFF
--- a/lapack/gonum/dgetri.go
+++ b/lapack/gonum/dgetri.go
@@ -91,7 +91,7 @@ func (impl Implementation) Dgetri(n int, a []float64, lda int, ipiv []int, work 
 			jb := min(nb, n-j)
 			// Copy current block column of L to work and replace
 			// with zeros.
-			for jj := j; jj < j+jb-1; jj++ {
+			for jj := j; jj < j+jb; jj++ {
 				for i := jj + 1; i < n; i++ {
 					work[i*ldwork+(jj-j)] = a[i*lda+jj]
 					a[i*lda+jj] = 0
--- a/lapack/gonum/dgetri.go
+++ b/lapack/gonum/dgetri.go
@@ -22,71 +22,95 @@ import (
 // by the temporary space available. If lwork == -1, instead of performing Dgetri,
 // the optimal work length will be stored into work[0].
 func (impl Implementation) Dgetri(n int, a []float64, lda int, ipiv []int, work []float64, lwork int) (ok bool) {
-	checkMatrix(n, n, a, lda)
-	if len(ipiv) < n {
+	switch {
+	case n < 0:
+		panic("lapack: has negative number of columns")
+	case lda < max(1, n):
+		panic("lapack: stride less than number of columns")
+	case lwork < max(1, n) && lwork != -1:
+		panic(badWork)
+	case len(work) < max(1, lwork):
+		panic(shortWork)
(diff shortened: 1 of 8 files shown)


vladimir-ch:
The block size adjustment is not covered by Dgetri's test, so should be added too (later).

lapack/testlapack/dlarfg.go
```diff
@@ -104,8 +104,9 @@ func DlarfgTest(t *testing.T, impl Dlarfger) {
 			Data:   make([]float64, n*n),
 		}
 		blas64.Gemm(blas.Trans, blas.NoTrans, 1, hmat, hmat, 0, eye)
-		if !isIdentity(n, eye.Data, n, 1e-14) {
-			t.Errorf("H^T * H is not I %v", eye)
+		dist := distFromIdentity(n, eye.Data, n)
+		if dist > 1e-14 {
```
kortschak: Make this a const as with `DgetriTest`.

lapack/gonum/dgetri.go
```diff
@@ -49,7 +49,7 @@ func (impl Implementation) Dgetri(n int, a []float64, lda int, ipiv []int, work
 	if nb > 1 && nb < n {
 		iws := max(ldwork*n, 1)
 		if lwork < iws {
-			nb = lwork / ldwork
+			nb = lwork / n
```
kortschak: This doesn't look like a row vs col major issue, so does this affect the NETLIB implementation as well?

lapack/gonum/dgetri.go
```diff
@@ -78,8 +78,8 @@ func (impl Implementation) Dgetri(n int, a []float64, lda int, ipiv []int, work
 			}
 			if j+jb < n {
 				bi.Dgemm(blas.NoTrans, blas.NoTrans, n, jb, n-j-jb, -1, a[(j+jb):], lda, work[(j+jb)*ldwork:], ldwork, 1, a[j:], lda)
-				bi.Dtrsm(blas.Right, blas.Lower, blas.NoTrans, blas.Unit, n, jb, 1, work[j*ldwork:], ldwork, a[j:], lda)
 			}
+			bi.Dtrsm(blas.Right, blas.Lower, blas.NoTrans, blas.Unit, n, jb, 1, work[j*ldwork:], ldwork, a[j:], lda)
```
kortschak: Nasty non-bracketed conditional blocks!

lapack/gonum/dgetri.go
vladimir-ch: It is a row vs column major issue. This line calculates "how many columns fit into the workspace we have". For reference the column size is ldwork, for us it's n.

lapack/gonum/dgetri.go
vladimir-ch: Yeah, the code there is particularly difficult to parse. I had to stare on it for a while and still wasn't completely sure if dtrsm is under the if or not.

lapack/gonum/dgetri.go
btracey: Yep ... sorry.

lapack/testlapack/dlarfg.go
vladimir-ch: Done.

vladimir-ch:
PTAL

PR #808, opened 2019-01-17
travis: use build caching and don't force complete rebuild

kortschak:
This is an alternative to #800. I have confirmed that per matrix-factor-state caching is done by Travis (documented [here](https://docs.travis-ci.com/user/caching/#caches-and-build-matrices) and confirmed by experimentation). This change thus gives us both caching speed up and an informative display of build success states.

Please take a look.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -57,7 +57,8 @@ script:
  - go get -d -t -v ./...
  - go build -v ./...
  - go test $TAGS -v ./...
- - if [[ $TRAVIS_SECURE_ENV_VARS = "true" ]]; then bash ./.travis/test-coverage.sh; fi
+ # Only run test coverage analysis if we are secure and on master.
+ - if [[ "${TRAVIS_BRANCH}" == master && "${TRAVIS_SECURE_ENV_VARS}" == true ]]; then bash ./.travis/test-coverage.sh; fi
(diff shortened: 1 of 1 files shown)


kortschak:
For push builds, this pair of changes gives a significant improvement:
```
Ran for 16 min 19 sec
Total time 1 hr 27 min 10 sec
```

This will improve further when go1.9 is dropped. I estimate the wall clock run time will reduce to about 10 min and the CPU wall time will reduce to about 1 hour. The pr build (this is a build of the PR merged into master) will not see so much of a benefit.

The difference between these two is the cost of running test coverage analysis. It would be nice to not run the test coverage analysis except for merged PRs though I'm not sure that is possible, we can look further into those options.

kortschak:
The last commit here needs discussion about whether we want to not have coverage reports in the PR and only on the front page. I'm just running it to see if it works.

sbinet:
instead of discovering after the facts that a given (merged) PR has tremendously impacted our coverage, perhaps we could only run the coverage analysis for one slot of the build matrix ? (say, Go-1.latest, `TAGS=''` ?)

kortschak:
I don't want to only get coverage for some tags. This was one of the advantages of the tags/env approach.

I would not expect to see a PR go through that had that kind of impact without someone understanding why, and I'd not be particularly worried if it did. I can undo the last commit though if you think it's worth it.

PR #810, opened 2019-01-18
blas/testblas: rewrite DtrsmTest to test various matrix strides

vladimir-ch:
This PR rewrites DtrsmTest so that it would have caught the matrix stride bugs fixed in #801.
Also adds one test case to cover the branch for `alpha == 0`.

Please take a look.

### DIFF
--- a/blas/testblas/dtrsm.go
+++ b/blas/testblas/dtrsm.go
@@ -7,6 +7,8 @@ package testblas
 import (
 	"testing"
 
+	"golang.org/x/exp/rand"
+
 	"gonum.org/v1/gonum/blas"
 	"gonum.org/v1/gonum/floats"
 )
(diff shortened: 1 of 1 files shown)


kortschak:
This is a perfect place to ask about the impact of https://github.com/gonum/gonum/commit/f0573bb67f2975b05e827edfc1e8e070417ddf07. Your PR here is branched from before that, so you are seeing a coverage report. Do you think it's useful to have that here (this would seem to me to be in the single class of PRs that *would* benefit from a coverage report prior to merging)?

vladimir-ch:
Do you mean by "coverage report" what I get after clicking on Details in the coverage/coveralls check here above? I admit I never look there during normal development, I prefer to see the coverage in the editor so I may not be the right person to ask. I would keep it, especially now that you have drawn my attention to it, but I don't have a strong opinion about it.

kortschak:
The details are still reachable from the coverage badge on the front page. I'm referring to the part of the test build summary here that says "coverage/coveralls — Coverage increased (+0.01%..." when you click on the pass tick/fail cross. That doesn't exist since https://github.com/gonum/gonum/pull/808/commits/270fd35e3375620e3c1492e4d1f99252c0ca7bb0 (though updated coverage from README.md does).

PR #812, opened 2019-01-19
travis: run test coverage analysis in PRs for the latest Go release

kortschak:
This is an experiment for using the Go version vs latest Go release to determine when pr builds should run test coverage analysis.

There seems to be a fairly divergent set of views on how important this is, so if you want this please say so here (not just approve). Similarly, if you don't want it, please also explain why.

See [this thread](https://groups.google.com/d/topic/gonum-dev/NV9m2S473Pc/discussion) for background.

Please take a look.


.travis/test-coverage.sh
```diff
@@ -3,6 +3,23 @@
...
 ACC_OUT=$PWD/acc.out
 
+latestRelease=$(
+	git ls-remote --tags https://github.com/golang/go.git 'go*' \
+	| egrep 'go[1-9]+\.[0-9]+\.[0-9]+$' \
+	| sed -e 's|^.*tags/||g' \
+	| sort -V \
+	| tail -n 1
+)
+echo Latest Go releast: $latestRelease
```
sbinet: `s/releast/release/`

.travis/test-coverage.sh
```diff
@@ -3,6 +3,23 @@
...
+
+currentVersion=$(go version | cut -d' ' -f3)
+echo Current Go version: $currentVersion
+
+if [[ "$currentVersion" != "$latestRelease" ]]; then
```
sbinet: perhaps this piece of code should be modified to only test for the 'Major.Minor' version and not the 'Patch' one (as travis may lag a bit before providing the latest patch version to the build bots.) ?

.travis/test-coverage.sh
kortschak: Fixed

.travis/test-coverage.sh
kortschak: Done

btracey:
I'm happy to just run on master

PR #813, opened 2019-01-21
lapack/gonum: fix workspace usage bugs in Dgetri

vladimir-ch:
... and extend its test, which uncovered these issues.

This completes the coverage of Dgetri mentioned in #806 .

This also shows that testing with various workspace sizes definitely makes sense.

Please take a look.


vladimir-ch:
Yeah, if we were starting again from scratch, I would vote for column major.

PR #818, opened 2019-01-26
graph/{multi,simple,testgraph}: add graph mutation tests

### LINKED ISSUE #616: graph/{simple,multi}: clean up implementation organisation and do more rigorous testing

### LINKED ISSUE #822: graph/multi: undirected graphs delete lines too agressively
When deleting lines, the two undirected graph implementations (weighted and unweighted) do not retain other lines between the nodes that are the endpoints of the line that is to be deleted. This is due to an error in the map handling in the `RemoveLine` method which should look more like the methods of the directed implementations.

kortschak:
Please take a look.

Closes  #616.
Fixes #822.

### DIFF
--- a/graph/multi/directed_test.go
+++ b/graph/multi/directed_test.go
@@ -8,8 +8,11 @@ import (
 	"math"
 	"testing"
 
+	"golang.org/x/exp/rand"
+
 	"gonum.org/v1/gonum/graph"
 	"gonum.org/v1/gonum/graph/internal/set"
+	"gonum.org/v1/gonum/graph/iterator"
 	"gonum.org/v1/gonum/graph/multi"
 	"gonum.org/v1/gonum/graph/testgraph"
 )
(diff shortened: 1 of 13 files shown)


kortschak:
This is ready. Please take a look.

PR #819, opened 2019-01-27
mat: Add Zero method to reset values of matrices

### LINKED ISSUE #817: mat: Add Zero() function
Right now we have `Reset` which resets the matrix to zero-sized to be used for operations. Perhaps we should also have `Zero` which sets the data to zero (which is like `Reset`, but preserves the size information)

btracey:
Fixes #817 

Please take a look.

### DIFF
--- a/mat/band.go
+++ b/mat/band.go
@@ -242,3 +242,15 @@ func (b *BandDense) DoColNonZero(j int, fn func(i, j int, v float64)) {
 		}
 	}
 }
+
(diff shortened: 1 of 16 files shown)


btracey:
I think this is a simple enough idea to not need the more complicated suggestion in the issue. If we find we need more things like this we could think about other functions along those lines, but this is similar enough in spirit to `Reset` I think it should behave like it.

btracey:
Done.

PR #820, opened 2019-01-27
all: remove ineffectual assignments

kortschak:
I have left cephes and amos as is.

Please take a look.

### DIFF
--- a/graph/formats/gexf12/gexf_test.go
+++ b/graph/formats/gexf12/gexf_test.go
@@ -437,6 +437,9 @@ func TestUnmarshal(t *testing.T) {
 		}
 		var got Content
 		err = xml.Unmarshal(data, &got)
+		if err != nil {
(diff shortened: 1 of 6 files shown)


lapack/testlapack/dgesvd.go
```diff
@@ -178,7 +178,7 @@ func svdCheckPartial(t *testing.T, impl Dgesvder, job lapack.SVDJob, errStr stri
 	if shortWork {
```
kortschak: Note for this file that there are ineffasigns to `jobU` and `jobV` that I don't understand how to resolve. @btracey do you have suggestions?

vladimir-ch: Please wait with this, I've completely rewritten the Dgesvd test which I will submit soon.

kortschak: I'll drop this commit.

kortschak:
I'll wait for @vladimir-ch as well.

Issue #756, opened 2017-04-30
mat: consider reconciling option passing in factorise methods


kortschak:
At the moment there are three factorisation methods that take options, the methods on `Eigen`, `EigenSym`, `GSVD` and `SVD`. These take different approaches to specifying options for factorisation.

I'm not too sure whether these can be reconciled, or whether doing so would be a net benefit, but it is worth considering.

@btracey @vladimir-ch Views?

btracey:
I agree with @vladimir-ch that ORed bit-fields are out of place. It's not clear to me that reconciling the `Factorize` methods is a goal, but if it is I think the way to do it is to use the approach we took in `samplemv`. There, we moved the options into the struct type, and rectified the signatures. So, it would be

```
type SVD struct {
    UKind SVDKind
    VKind SVDKind 
   // contains filtered or unexported fields
}

func (svd *SVD) Factorize(a Matrix) (ok bool){...}
```
and be used as

` svd := &mat.SVD{UKind: svd.Thin, VKind: svd.None}.Factorize(a)`

In `samplemv` this is clearly a good idea because we wanted the types to all satisfy the same interface. It's not clear that is a useful feature here. I think if we do do this then we should make them satisfy the same interfaces though, in case that turns out to be useful. In particular, `Cholesky` should be `FactorizeSym` rather than just `Factorize`. One upside to this is that we could make the zero-values take the default (UKind and VKind default to Thin, vectors in EigenSym is true, etc.). This way a user who doesn't want to think about anything can use 

`svd := &mat.SVD{}.Factorize(a)`  

and it will work as expected.

kortschak:
There was more to that sentence.

btracey:
I don't follow

btracey:
Assuming you meant @vladimir-ch 's sentence, there was also more to my post?

kortschak:
I was responding to the first sentence, yes.

I don't think that there is a good justification for a common interface, but I do think there is a good reason for a general consistency in the approach taken to passing options. The first would be there to ensure code reusability between the factorising types, which I think is a non-goal since one factorisation is not the same as another factorisation (and the result accessors would be different anyway - so they should be part of the interface too if it existed which means there would not be a common interface). The existence of a consistent user interface (the signature vibe) means that learning one means that the user has the concepts necessary for rapidly learning others (the only difference being the exact semantics of the constants/parameters).

PR #828, opened 2019-01-30
mat: Add SVD option for choosing which vectors to compute

### LINKED ISSUE #487: mat: allow to explicitly specify SVDKind for both jobU, jobVT in Factorize()
I'm implementing Kalman Filter algorithm based on singular value decomposition. The crucial part of this algorithm involves the computation of the singular value of an asymmetric matrix through SVD without explicitly forming its left factor (U). In other words, the algorithm only uses singular values and right singular vectors (Vt). At the moment, I cannot find a way to only get the parts described. I think It would be a significant improvement to `Factorize()` if it used separate arguments for each part, as it is done f.ex in R's `svd()`.

I also tried to write a my own `Factorize()` that wraps `lapack64.Gesvd()`, however private `getFloats()` and `putFloats()` keep me from doing this. Maybe there is an option to make those public?

btracey:
Please take a look.

Fixes #487

### DIFF
--- a/mat/consts.go
+++ b/mat/consts.go
@@ -31,6 +31,12 @@ const (
 	//  A = U * Σ * V^T
 	// where U is of size m×m, Σ is an m×n diagonal matrix, and V is an n×n matrix.
 	SVDFull
+	// SVDNoneThin computes the thin singular vectors as described above, but
+	// only computes Σ and V~^T, not U.
+	SVDNoneThin
+	// SVDThinNone computes the thin singular vectors as described above, but
+	// only computes U~ and Σ, not V~^T.
+	SVDThinNone
 )
 
 // GSVDKind specifies the treatment of singular vectors during a GSVD
--- a/mat/svd.go
+++ b/mat/svd.go
@@ -69,7 +69,7 @@ func (svd *SVD) Factorize(a Matrix, kind SVDKind) (ok bool) {
 	case SVDThin:
 		// TODO(btracey): This code should be modified to have the larger
 		// matrix written in-place into aCopy when the lapack/native/dgesvd
-		// implementation is complete.
+		// implementation is complete (and below also).
 		svd.u = blas64.General{
 			Rows:   m,
 			Cols:   min(m, n),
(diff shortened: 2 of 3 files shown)


mat/consts.go
```diff
@@ -31,6 +31,12 @@ const (
 	//  A = U * Σ * V^T
 	// where U is of size m×m, Σ is an m×n diagonal matrix, and V is an n×n matrix.
 	SVDFull
```
kortschak: This is starting to look to me like it should be a bit flag set.

```
const (
	// SVDU computes the thin singular vectors as described for SVDThin, but
	// only computes U~ and Σ, not V~^T.
	SVDU SVDKind = 1 << iota

	// SVDV computes the thin singular vectors as described for SVDThin, but
	// only computes Σ and V~^T, not U.
	SVDV

	// SVDThin computes the thin singular vectors, that is, it computes
	//  A = U~ * Σ * V~^T
	// where U~ is of size m×min(m,n), Σ is a diagonal matrix of size min(m,n)×min(m,n)
	// and V~ is of size n×min(m,n).
	SVDThin

	// SVDNone specifies that no singular vectors should be computed during
	// the decomposition.
	SVDNone SVDKind = 0
)
```

btracey: We could change the SVD signature to match the original lapack signature. The discussion in the issue concluded supporting all combinations was unnecessary.

kortschak: With the three bits, you get only eight possible combinations with 7 distinct outcomes. Using the bit flags simplifies the API  by making things (largely) orthogonal and the big switch in Factorize become a smaller set of orthogonal conditions with a little bit of preparation before hand.

kortschak: That it true, if your want fine control U-thin/V-full, but that's probably fairly rare. Using three bits gets all the cases above except those two, but for three necessary constants with explicit naming, and one convenience constant for 0 while the approach here requires five constants that use positional implied naming (`SVDNoneThin` really is `SVDNoUThinV`, which would be expressed as `SVDV|SVDThin` in the suggested scheme).

kortschak: See also #756.

vladimir-ch: I think I would be fine with just the current `SVDFull`, `SVDThin`, `SVDNone`, simple, clear and covers a majority of uses. But if we are to refine it, then I slightly prefer the bitfield approach. I agree with @kortschak 's API simplification and orthogonality argument. Also I think the `|` separator makes it easier to see what is being requested, the mixed case constants could look cryptic to an unprepared eye. And if the type were renamed from `SVDKind` to `SVDFlags`, it would better communicate that the parameter is a bitfield. If we keep the way in this PR, then I would include U and V explicitly in the constants, as in `SVDThinUNoneV` or `SVDThinUNoV`.

btracey:
Replaced with #872

PR #835, opened 2019-02-04
graph/path: fix negative cycle handling by AllShortest and FloydWarshall

kortschak:
Over the past ... time ... I've been having doubts about my claim for the approach obtaining a complete set of negative cycles in a graph using iterated Bellman-Ford probing. Today I checked it and my concerns were well founded, so I have removed that.

Wanting to have some way of doing this, I added an example using Floyd-Warshall and found that with negative cycles, `AllShortest` will never complete when a negative cycle exists. So I have fixed that by just not providing a mechanism to obtain the path. This is a bit of a dodge, but we can add it later if we want. The additional cost for the `FloydWashall` is an approximate doubling of time when a negative cycle exists. I think this is OK, but could be convinced otherwise (maybe add an option to retain the data for the call, or add a method to `AllShortest` to sanitise the distance matrix if needed — I think there is an option to do this lazily as well during the weight/paths calls: for each query, check do the work in the k loop in `FloydWarshall` for u and v only, storing the NaN if it's found and returning early if u->v is already NaN).

The use of NaN for undefined path weights reflects the behaviour that already existed in `BellmanFord`/`Shortest`. Other people use `-Inf` for this. I'm open to changing this as well.

Please take a look.

### DIFF
--- a/graph/path/negative_cycles_example_test.go
+++ b/graph/path/negative_cycles_example_test.go
@@ -14,10 +14,7 @@ import (
 
 func ExampleBellmanFordFrom_negativecycles() {
 	// BellmanFordFrom can be used to find a non-exhaustive
-	// set of negative cycles in a graph. Enumerating the
-	// exhaustive list requires iterations of the procedure
-	// here successively omitting links from the new node
-	// to already found negative cycles.
+	// set of negative cycles in a graph.
 
 	// Construct a graph with a negative cycle.
 	edges := []simple.WeightedEdge{
@@ -47,10 +44,10 @@ func ExampleBellmanFordFrom_negativecycles() {
 		fmt.Println("no negative cycle present")
 		return
 	}
-	for _, n := range []simple.Node{'a', 'b', 'c', 'd', 'e', 'f'} {
-		p, w := pt.To(n.ID())
+	for _, id := range []int64{'a', 'b', 'c', 'd', 'e', 'f'} {
+		p, w := pt.To(id)
 		if math.IsNaN(w) {
-			fmt.Printf("negative cycle in path to %c path:%c\n", n, p)
+			fmt.Printf("negative cycle in path to %c path:%c\n", id, p)
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


kortschak:
Do you have any views on `NaN` v. `-Inf` for weights of lowest weight negative cycles? I can see merits in both directions.

graph/path/negative_cycles_example_test.go
```diff
@@ -60,3 +57,75 @@ func ExampleBellmanFordFrom_negativecycles() {
...
+	// c is in a negative cycle
+	// e is in a negative cycle
+	// f is in a negative cycle
+	// negative cycle in path from a to a unique=false
+	// negative cycle in path from a to e unique=false
+	// negative cycle in path from a to f unique=false
+	// negative cycle in path from b to a unique=false
+	// negative cycle in path from b to b unique=false
+	// negative cycle in path from b to c unique=false
+	// negative cycle in path from b to d unique=false
```
vladimir-ch: There is a direct edge from b to d, its weight is 1, d is not reachable in other way, so why is this in the output? Am I missing something?

vladimir-ch:
What merits do you see for each? In the end it doesn't matter too much? We just need some special value to mark the edges. Since we are marking edges on _negative_ cycles, -Inf would seem to me like a more natural choice, but I think your insight is deeper.

graph/path/negative_cycles_example_test.go
kortschak: Hee. This is the joy of negative cycles.

There is a direct path from b->d of cost 1, but there are n paths (b->c->b)_n->d of cost 1-2n. We don't restrict paths to be simple (the b->d path), because if we do and we still retain polynomial time (as we do here), we prove P=NP by example. And we don't want to do that.

kortschak:
I don't know exactly why I picked NaN originally; maybe because the shortest path is undefined. However, the weight of a path with a negative cycle with a simple path weight in the cycle of -1 has a limit of `-Inf` for as n approaches inifinity, so perhaps we can say the path is undefined but the weight is `-Inf`.

As far as the code goes, it gets rid of the `IsNaN` condition check since `-Inf` is negative.

kortschak:
I have changed the weight of negative cycles to `-Inf`.

PTAL

kortschak:
The thing that we lose when we use `-Inf` as I have here is the capacity to distinguish simple paths that have `-Inf` weight and negative cycles. I can imagine that `-Inf` edge weights are sometimes useful to ensure path choice. I can get around that by using `NaN` internally and translating that to `-Inf` in calls to the API; `NaN` weight edges are far less likely to be useful.

I'll think some more, but feel free to chime in.

graph/path/negative_cycles_example_test.go
vladimir-ch: Oh.

PR #836, opened 2019-02-06
num/dualquat: package for dual quaternion maths

kortschak:
I have struggled with getting the trig functions to work with these packages and this has prevented me from sending this. Given that the primary use for these types of numbers is to allow 2D and 3D translations and rotations, something that doesn't require trig in this number space, I have decided that discretion is the better part of valor and have removed them. The problems I was facing were compounded (but not entirely caused) by the odd behaviour of "math/cmplx" for some of these functions. If people want trig functions for these, they can file an issue and then I can look at it again (the almost working code is in dualquat-full).

~There is one issue that I am not entirely sure of yet, and I think stems from the commutativity of complex numbers when quaternions are not commutative (this shows up in the different behaviours of the rotate and translate examples of the two packages). I would like to sort this out, but am sending this now.~

Please take a look.

### DIFF
--- a/num/dualquat/doc.go
+++ b/num/dualquat/doc.go
@@ -1,0 +1,9 @@
+// Copyright ©2018 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Package dualquat provides the dual quaternion numeric type and functions.
+package dualquat // imports "gonum.org/v1/gonum/num/dualquat"
+
+// TODO(kortschak): Handle special cases properly.
+//  - Pow
--- a/num/dualquat/doc.go
+++ b/num/dualquat/doc.go
@@ -3,6 +3,10 @@
 // license that can be found in the LICENSE file.
 
 // Package dualquat provides the dual quaternion numeric type and functions.
+//
+// Dual quaternions provide a system for rigid transformation with interpolation
+// and blending in ℝ³. See https://www.cs.utah.edu/~ladislav/kavan06dual/kavan06dual.pdf and
+// https://en.wikipedia.org/wiki/Dual_quaternion for more details.
 package dualquat // imports "gonum.org/v1/gonum/num/dualquat"
 
 // TODO(kortschak): Handle special cases properly.
--- a/num/dualquat/dual.go
+++ b/num/dualquat/dual.go
@@ -12,7 +12,9 @@ import (
 	"gonum.org/v1/gonum/num/quat"
 )
 
-// Number is a float64 precision dual quaternion.
+// Number is a float64 precision dual quaternion. A dual quaternion
+// is a hypercomplex number composed of two quaternions, q₀+q₂ϵ,
+// where ϵ²=0, but ϵ≠0. Here, q₀ is termed the real and q₂ the dual.
 type Number struct {
 	Real, Dual quat.Number
 }
@@ -104,6 +106,14 @@ func Inv(d Number) Number {
 	return Number{
 		Real: quat.Inv(d.Real),
 		Dual: quat.Scale(-1, quat.Mul(d.Dual, quat.Inv(quat.Mul(d.Real, d.Real)))),
+	}
+}
+
+// Conj returns the dual quaternion conjugate of d₁+d₂ϵ, d̅₁-d̅₂ϵ.
+func Conj(d Number) Number {
+	return Number{
+		Real: quat.Conj(d.Real),
+		Dual: quat.Scale(-1, quat.Conj(d.Dual)),
 	}
 }
 
--- a/num/dualquat/dual.go
+++ b/num/dualquat/dual.go
@@ -1,0 +1,157 @@
+// Copyright ©2018 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package dualquat
+
+import (
+	"bytes"
+	"fmt"
+
+	"gonum.org/v1/gonum/num/dual"
+	"gonum.org/v1/gonum/num/quat"
+)
+
+// Number is a float64 precision dual quaternion.
+type Number struct {
+	Real, Dual quat.Number
+}
+
+var (
+	zero     Number
+	zeroQuat quat.Number
+)
+
+// Format implements fmt.Formatter.
+func (d Number) Format(fs fmt.State, c rune) {
+	prec, pOk := fs.Precision()
+	if !pOk {
+		prec = -1
+	}
+	width, wOk := fs.Width()
+	if !wOk {
+		width = -1
+	}
+	switch c {
+	case 'v':
+		if fs.Flag('#') {
+			fmt.Fprintf(fs, "%T{%#v, %#v}", d, d.Real, d.Dual)
+			return
+		}
+		c = 'g'
+		prec = -1
+		fallthrough
+	case 'e', 'E', 'f', 'F', 'g', 'G':
+		fre := fmtString(fs, c, prec, width, false)
+		fim := fmtString(fs, c, prec, width, true)
+		fmt.Fprintf(fs, fmt.Sprintf("(%s+%[2]sϵ)", fre, fim), d.Real, d.Dual)
+	default:
+		fmt.Fprintf(fs, "%%!%c(%T=%[2]v)", c, d)
+		return
+	}
+}
+
+// This is horrible, but it's what we have.
+func fmtString(fs fmt.State, c rune, prec, width int, wantPlus bool) string {
+	// TODO(kortschak) Replace this with strings.Builder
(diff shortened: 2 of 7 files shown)


kortschak:
I've spent a bit of time messing about trying to get the dual complex examples to work, and I know what it not working, but don't understand how to fix that. The issue is that rotations leave the non-dual as non-unit which they should not. However, using the approach used for quaternion rotation (or dual quaternion rotation/translation does not work).

As a note for myself, read [this](https://arxiv.org/abs/1601.01754v1).

kortschak:
I have backed out the dual complex until I understand it better. It will be a separate PR.

num/dualquat/dual.go
```diff
@@ -0,0 +1,157 @@
...
+		fre := fmtString(fs, c, prec, width, false)
+		fim := fmtString(fs, c, prec, width, true)
+		fmt.Fprintf(fs, fmt.Sprintf("(%s+%[2]sϵ)", fre, fim), d.Real, d.Dual)
+	default:
+		fmt.Fprintf(fs, "%%!%c(%T=%[2]v)", c, d)
+		return
+	}
+}
+
+// This is horrible, but it's what we have.
```
btracey: Almost not though!

num/dualquat/dual.go
```diff
@@ -0,0 +1,157 @@
...
+
+// ConjDual returns the dual conjugate of d₁+d₂ϵ, d₁-d₂ϵ.
+func ConjDual(d Number) Number {
+	return Number{
+		Real: d.Real,
+		Dual: quat.Scale(-1, d.Dual),
+	}
+}
+
+// ConjQuat returns the quaternion conjugate of d₁+d₂ϵ, d̅₁+d̅₂ϵ.
```
btracey: Is there a need to have a `ConjDualQuat` that does both?

num/dualquat/dual_example_test.go
```diff
@@ -0,0 +1,148 @@
...
+	return dualquat.Number{
+		Real: quat.Number{Real: 1},
+		Dual: raise(p),
+	}
+}
+
+// transform performs the quaternion rotation of p by the given quaternion
+// and scaling by the scale factor. The rotations are normalized to unit
+// vectors.
+func transform(p point, by ...dualquat.Number) point {
```
btracey: Is this something that should exported? It's the core of the Example. Or is the idea that some outside package would do this?

num/dualquat/dual_example_test.go
```diff
@@ -0,0 +1,148 @@
...
+		q = dualquat.Mul(o, q)
+	}
+	pp := dualquat.Mul(dualquat.Mul(q, raiseDual(p)), dualquat.ConjDual(dualquat.ConjQuat(q)))
+
+	// Extract the point.
+	return point{x: pp.Dual.Imag, y: pp.Dual.Jmag, z: pp.Dual.Kmag}
+}
+
+func Example() {
+	// Translate a 1×1×1 cube [3, 4, 5] and rotate it 120° around the
```
btracey: s/cube [3,/cube by [3,

num/dualquat/dual_example_test.go
```diff
@@ -0,0 +1,148 @@
...
+
+		// Clean up floating point error for clarity.
+		pp.x = floats.Round(pp.x, 2)
+		pp.y = floats.Round(pp.y, 2)
+		pp.z = floats.Round(pp.z, 2)
+
+		fmt.Printf(" %d %+v -> %+v\n", i, p, pp)
+	}
+
+	// Rotate a line segment from [2, 1, 1] to [2, 1, 2] 120° around
```
btracey: Maybe "Rotate the line segment {[2, 1, 1], [2, 1, 2]}"

num/dualquat/dual_example_test.go
```diff
@@ -0,0 +1,148 @@
...
+
+	// Construct a displacement.
+	displace := dualquat.Number{
+		Real: quat.Number{Real: 1},
+		Dual: quat.Scale(0.5, raise(point{3, 4, 5})),
+	}
+
+	// Construct a rotations.
+	alpha := 2 * math.Pi / 3
+	axis := raise(point{1, 1, 1})
```
btracey: You should add a comment for what it does, as it won't be in the godoc.

num/dualquat/dual_fike.go
```diff
@@ -0,0 +1,152 @@
+// Copyright ©2018 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Derived from code by Jeffrey A. Fike at http://adl.stanford.edu/hyperdual/
+
+// The MIT License (MIT)
```
btracey: Why does this need a license? It seems like this code is yours using the other code that is related to Jeff's

num/dualquat/dual_simple_example_test.go
```diff
@@ -0,0 +1,95 @@
...
+package dualquat_test
+
+import (
+	"fmt"
+
+	"gonum.org/v1/gonum/num/dualquat"
+	"gonum.org/v1/gonum/num/quat"
+)
+
+// Example point, displacement and rotation from Euclidean Space Dual Quaternions page:
```
btracey: It seems that this is worth including in the example text, the website seems nice.

num/dualquat/doc.go
```diff
@@ -0,0 +1,9 @@
+// Copyright ©2018 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Package dualquat provides the dual quaternion numeric type and functions.
```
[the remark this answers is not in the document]
kortschak: Added.

I can add a similar what to dual and hyperdual if you think it's worth it here.

num/dualquat/dual.go
kortschak: ?

num/dualquat/dual.go
kortschak: Yeah. I was thinking about that. For the dual quaternion case, there are descriptions in the literature of both the dual conjugate and the quaternion conjugate, but only the double is used for this case. In the dual complex system, only the one conjugate is described as far as I can see. I am thinking that I could add just `Conj` which does the obviously correct thing and then leave the others as helpers, or even just remove them.

I'll add `Conj` and leave the others for now.

num/dualquat/dual.go
btracey: 1.12 is almost out, which means 1.9 support will be dropped from Gonum

num/dualquat/dual_example_test.go
kortschak: This is factored out to help with the explanation. I wouldn't do it this way.

num/dualquat/dual_example_test.go
kortschak: Done

num/dualquat/dual_example_test.go
kortschak: Done

num/dualquat/dual_example_test.go
kortschak: No, this is in the godoc. See the "Larger Examples" section in https://blog.golang.org/examples.

num/dualquat/dual_fike.go
kortschak: The is very similar to the code from Jeff's for the dual reals. The dual complex is adequately different and I removed the distinction there. I don't mind either way, but if we do remove it, I'd prefer to check with him first. You know him, so do you want to ask?

num/dualquat/dual_simple_example_test.go
kortschak: Added in the package doc.

num/dualquat/dual.go
kortschak: Ah. That's not the horrible thing. The horrible thing is that the `fmt.Formatter` API does not allow a pass through without constructing a new format string. I can just say, format these things with the same format as I already have, so we have to perform at least one more allocation and then reparse the format string.

kortschak:
References are added to the package doc.

Also https://dualquaternions.weebly.com/uploads/1/3/1/5/13150194/dual_quaternions_wscg_paper_pub.pdf

I just remembered the WP page is OK, https://en.wikipedia.org/wiki/Dual_quaternion. I'll add that to the package doc.

num/dualquat/dual_example_test.go
btracey: Thanks, I didn't realize that.

num/dualquat/dual_fike.go
btracey: It's okay to leave in, I looked at the other code and it's more similar than I remembered.

num/dualquat/dual.go
btracey: SGTM

PR #837, opened 2019-02-06
travis: fix gocc installation

kortschak:
It looks like Travis have set `GO111MODULE=on`, breaking our approach to versioning gocc for the code generation checks for graph/format/dot. This gets rid of the use of go get for this pull.

Please take a look.


kortschak:
This is a dogs breakfast! I knew modules would be a pain for us.

Issue #838, opened 2019-02-06
travis: build does not set up GOPATH when go.mod exists


kortschak:
See https://docs.travis-ci.com/user/languages/go/#go-modules

This results is significant breakage of the build. See https://travis-ci.org/gonum/gonum/builds/489341118, https://travis-ci.org/gonum/gonum/builds/489406029 (partial fix) and https://travis-ci.org/gonum/gonum/builds/489413965 (partial fix).

Notable points are that: go get now doesn't give us the source, so we should be using git clone when we want that — this is the case for gocc as we checkout a specific commit to work with; we need to hand gocc information about where the generation is being done; and `gonum.org/v1/gonum/unit`'s code gen needs to find another way to root it's work (there may be others like this).

/cc @btracey @sbinet @mewmew

kortschak:
Work is happening [here](https://github.com/gonum/gonum/tree/travis-modules).

sbinet:
When in 'module' mode, sources are checked out under GOPATH/pkg/mod/import-path.
(IIRC, they are read-only though)

kortschak:
I have all but one of the issue resolved (I think).

The remaining issue is that gocc writes out the wrong import paths.
```
rm -f errors/errors.go
rm -f lexer/acttab.go
rm -f lexer/lexer.go
rm -f lexer/transitiontable.go
rm -f parser/action.go
rm -f parser/actiontable.go
rm -f parser/gototable.go
rm -f parser/parser.go
rm -f parser/productionstable.go
rm -f token/token.go
rm -f util/litconv.go
rm -f util/rune.go
rmdir --ignore-fail-on-non-empty errors
rmdir --ignore-fail-on-non-empty lexer
rmdir --ignore-fail-on-non-empty parser
rmdir --ignore-fail-on-non-empty token
rmdir --ignore-fail-on-non-empty util
rm -f terminals.txt LR1_conflicts.txt LR1_sets.txt first.txt lexer_sets.txt
gocc dot.bnf
go run paste_copyright.go
diff --git a/go.mod b/go.mod
index 4fc1257..2385a9b 100644
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,8 @@
 module gonum.org/v1/gonum
 
 require (
+	github.com/mattn/goveralls v0.0.2 // indirect
 	golang.org/x/exp v0.0.0-20190125153040-c74c464bbbf2
 	golang.org/x/tools v0.0.0-20190206041539-40960b6deb8e
[...]

PR #839, opened 2019-02-07
travis: preparatory changes for Go modules in builds

kortschak:
See #838 for background.

This makes some preparatory changes for using Go modules and for the most part works around the changes that transiently happened to travis' behaviour when a go.mod file existed. It did not completely fix that, but that is not longer necessary.

Please take a look.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,9 +5,9 @@ language: go
 # Do not move these lines; they are referred to by README.md.
 # Versions of go that are explicitly supported by gonum plus go tip.
 go:
- - 1.9.x
- - 1.10.x
  - 1.11.x
+ - 1.10.x
+ - 1.9.x
  - master
 
 env:
--- a/.travis/check-generate.sh
+++ b/.travis/check-generate.sh
@@ -5,6 +5,7 @@ go generate gonum.org/v1/gonum/blas/gonum
 go generate gonum.org/v1/gonum/unit
 go generate gonum.org/v1/gonum/graph/formats/dot
 if [ -n "$(git diff)" ]; then
+	git checkout -- go.mod # Discard changes to go.mod that have been made.
 	git diff
 	exit 1
 fi
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module gonum.org/v1/gonum
 
 require (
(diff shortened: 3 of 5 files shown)


go.mod
```diff
@@ -1,6 +1,6 @@
 module gonum.org/v1/gonum
 
 require (
-	golang.org/x/exp v0.0.0-20180321215751-8460e604b9de
-	golang.org/x/tools v0.0.0-20180525024113-a5b4c53f6e8b
+	golang.org/x/exp v0.0.0-20190125153040-c74c464bbbf2
```
btracey: Is the plan to manually update these as necessary?

.travis.yml
```diff
@@ -5,9 +5,9 @@ language: go
 # Do not move these lines; they are referred to by README.md.
 # Versions of go that are explicitly supported by gonum plus go tip.
 go:
- - 1.9.x
- - 1.10.x
  - 1.11.x
+ - 1.10.x
```
sbinet: why the order inversion?

.travis/check-generate.sh
```diff
@@ -5,6 +5,7 @@ go generate gonum.org/v1/gonum/blas/gonum
 go generate gonum.org/v1/gonum/unit
 go generate gonum.org/v1/gonum/graph/formats/dot
 if [ -n "$(git diff)" ]; then
+	git checkout -- go.mod # Discard changes to go.mod that have been made.
```
sbinet: we possibly might have also to discard `go.sum` ?

go.mod
kortschak: This is done by `go get` in module mode.

.travis.yml
kortschak: More recent are more important for failures.

.travis/check-generate.sh
kortschak: When we commit it, yes. It's not currently in the repo.

go.mod
kortschak: We should probably have a policy on when it's appropriate to bump a dependency's version. I don't know if there are guidelines anywhere.

kortschak:
*ping*

PR #844, opened 2019-02-09
mat: check validity in Cholesky.Cond

### LINKED ISSUE #747: mat: Cholesky should provide a way to query the condition number
There is no way to know the condition number of the matrix without trying to do a solve first.

vladimir-ch:
Please take a look.

### DIFF
--- a/mat/cholesky.go
+++ b/mat/cholesky.go
@@ -60,6 +60,9 @@ func (c *Cholesky) updateCond(norm float64) {
 
 // Cond returns the condition number of the factorized matrix.
 func (c *Cholesky) Cond() float64 {
+	if !c.valid() {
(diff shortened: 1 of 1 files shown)


kortschak:
I'm a bit confused about the reference to the issue. The issue talks about not having this method, but it's here, just not quite right, but I can't find a back reference in the issue to when `Cond` was added.

vladimir-ch:
Yes, it was a bad idea to try and close the issue with this PR. Sorry. I will remove the reference.

`Cholesky.Cond` was added in #104, I guess in order to address #747 but didn't close it. I'll close the issue manually.

kortschak:
Thanks for explaining.

Issue #833, opened 2019-02-02
stat,all: Create Bounds package and type


btracey:
Right now we have a `Bound` type in `stat/distmv`, but really this is a more general concept than distmv. In particular, an equivalent type is needed in `optimize` (for bounded optimization problems), and I could see a type easily being useful in a future `root` package. I think we should have a `gonum.org/v1/gonum/bound` package that contains that type. The signature for `distmv.NewUniform` would then be `func NewUniform(bnds bound.Bound, src rand.Source)`. This allows the same data structure to be used across packages.

sbinet:
SGTM

btracey:
There are more routines to add, see https://github.com/gonum/gonum/pull/845#discussion_r255359176

btracey:
Removing the stability tag because the API changes from this issue are done.

btracey:
Created #939 because it's really a separate issue from this one which is now done.

PR #845, opened 2019-02-10
bound: add Bound package and type

btracey:
Updates #833.

Please take a look.

### DIFF
--- a/bound/bound.go
+++ b/bound/bound.go
@@ -1,0 +1,11 @@
+// Copyright ©2019 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package bound
+
+// Bound represents [Min, Max] bounds.
+type Bound struct {
+	Min float64
+	Max float64
+}
--- a/bound/bound.go
+++ b/bound/bound.go
@@ -6,6 +6,5 @@ package bound
 
 // Bound represents [Min, Max] bounds.
 type Bound struct {
-	Min float64
-	Max float64
+	Min, Max float64
 }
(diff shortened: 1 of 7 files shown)


bound/bound.go
```diff
@@ -0,0 +1,11 @@
+// Copyright ©2019 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package bound
+
+// Bound represents [Min, Max] bounds.
+type Bound struct {
+	Min float64
```
kortschak: Do these need to be on different lines?

bound/doc.go
```diff
@@ -0,0 +1,6 @@
+// Copyright ©2019 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Package bound provides types for bounded data.
```
kortschak: Are there any other things that this package provides? Do we possibly want ℝ² and ℝ³ bounds in the future to go along with ℝ¹?

bound/bound.go
btracey: No

bound/doc.go
btracey: Possibly... `distmv` and `optimize` need arbitrary dimensional `[]bound.Bound` though. We don't have many explicitly dimensioned things in Gonum at the moment. This is a natural type to go in `gonum.org/v1/gonum`, but we've resisted that for a variety of reasons, and we don't want a `util` package, so I wasn't sure where else it could go. I was thinking in the future it could possibly have some intersection/union style helper routines, sort of like in the `distmv.Uniform.Entropy` calculation.

kortschak: Yeah, I think intersection, union, Jacquard are probably enough reason for having this as a seprate package.

Issue #308, opened 2017-11-20
stat/combin: Implement iterator for Cartesian


btracey:
See comment https://github.com/gonum/gonum/pull/303#discussion_r151853074

PR #849, opened 2019-02-11
mat: Change Eigen to use complex matrix representation

### LINKED ISSUE #738: mat: Redesign Eigen
Right now, Eigen stores the eigenvalue matrix as a *Dense. Conceptually, this is wrong, because in the general case, it's really a complex matrix. The code instead uses extra rows and columns to represent complex values, which means that the physical size of the matrix is larger than its conceptual size. It would be better to just have the values just be complex. Acyclic imports mean we can't have mat128 depend on mat64 and vis-versa, but we can have mat64 import blas128. Eigen can return the blas structs, which can then be converted into mat128 matrices. This will take some work to interface with lapack routines, but not doing this work means we just push the problem onto the user (making them do the conversion from 2x2 float64 blocks to complex values). The eigenvalues by blas128.Banded

We should also reconsider the function signatures. First of all, it seems easier to me to have it be
Eigen(a *Dense, epsilon) (eigenvalues blas128.Banded, eigenvectors *Dense)
I'm not sure what the EigenFactors struct helps with
Secondly, we should also have
EigenSym(a *Symmetric, epsilon) (eigenvalues *Diagonal, eigenvectors *Dense)
as the eigenvalues of a symmetric matrix are real.
The asymmetry between the two is unfortunate (Diagonal vs. Banded), however there is no Diagonal matrix in BLAS. We could add a definition both into blas64 and blas128, or we could plausibly have a non-blas package containing the definition of RawDiagonal (name tbd) for both complex and real diagonal matrices. We could also leave as-is, and solve the problem in a different manner (there are a few choices).

btracey:
Fixes #738

Please take a look.


btracey:
I tried to keep this PR small. The plan is to open issues for the TODOs if the scope is loosely acceptable.

I marked as fixing 738 because this fixes the main problem depicted there. The design here may still need to change with #487

btracey:
Sorry, will fix a couple of things I forgot.

mat/cdense.go
```diff
@@ -57,3 +59,91 @@ func NewCDense(r, c int, data []complex128) *CDense {
...
+		m.capCols = c
+		return
+	}
+	if r != m.mat.Rows || c != m.mat.Cols {
+		panic(ErrShape)
+	}
+}
+
+func (m *CDense) reuseAsZeroed(r, c int) {
+	m.reuseAs(r, c)
```
btracey: In Dense (and here currently), it says we must be kept in sync, but I don't see any reason why this kind of implementation can't work. Do you know why @kortschak ?

kortschak: This does extra work. I imagine that's why I had two separate routines.

mat/cmatrix.go
```diff
@@ -71,3 +75,73 @@ type Unconjugator interface {
...
+		}
+	}
+	return true
+}
+
+// CEqualApprox returns whether the matrices a and b have the same size and contain all equal
+// elements with tolerance for element-wise equality specified by epsilon. Matrices
+// with non-equal shapes are not equal. Element-wise equality is compared on
+// the real and imaginary parts separately.
+func CEqualApprox(a, b CMatrix, epsilon float64) bool {
```
kortschak: Do we want a norm-2 distance or what is here? What is normally used for complex numbers? What is here is less strict than norm-2.

mat/cdense.go
btracey: Having two routines makes sense, it's that the Dense code has the reuseAs code copied rather than called as a method

mat/eigen.go
```diff
@@ -183,13 +183,27 @@ func (e *Eigen) Factorize(a Matrix, left, right bool) (ok bool) {
...
+	if left {
+		cvl = *NewCDense(r, r, nil)
+		e.complexEigenTo(&cvl, &vl)
+	}
+	if right {
+		cvr = *NewCDense(c, c, nil)
+		e.complexEigenTo(&cvr, &vr)
+	}
+
+	e.lVectors = &cvl
```
kortschak: Should these be non-conditional?

mat/eigen.go
```diff
@@ -229,18 +242,53 @@ func (e *Eigen) Values(dst []complex128) []complex128 {
 //  u_{j+1} = VL[:,j] - i*VL[:,j+1],
 //  v_j     = VR[:,j] + i*VR[:,j+1],
 //  v_{j+1} = VR[:,j] - i*VR[:,j+1],
-// where i is the imaginary unit. The computed eigenvectors are normalized to
-// have Euclidean norm equal to 1 and largest component real.
+// where i is the imaginary unit.
+func (e *Eigen) complexEigenTo(cd *CDense, d *Dense) {
```
kortschak: s/cd/dst/g

mat/eigen_test.go
```diff
@@ -28,17 +28,39 @@ func TestEigen(t *testing.T) {
...
 		},
+		{
+			// Values compared with numpy.
+			a: NewDense(4, 4, []float64{
+				0.9025, 0.025, 0.475, 0.0475,
+				0.0475, 0.475, 0.475, 0.0025,
+				0.0475, 0.025, 0.025, 0.9025,
+				0.0025, 0.475, 0.025, 0.0475,
+			}),
+			values: []complex128{1, 0.7300317046114154, complex(-0.1400158523057075, 0.452854925738716), complex(-0.1400158523057075, -0.452854925738716)},
```
kortschak: `complex(-0.1400158523057075, 0.452854925738716)` can be written as `-0.1400158523057075+0.452854925738716i`, and the others.

mat/eigen_test.go
```diff
@@ -91,6 +113,18 @@ func cmplxEqual(v1, v2 []complex128) bool {
 	return true
 }
 
+func cmplxEqualTol(v1, v2 []complex128, tol float64) bool {
```
kortschak: norm-2 distance?

mat/hogsvd.go
```diff
@@ -101,7 +101,18 @@ func (gsvd *HOGSVD) Factorize(m ...Matrix) (ok bool) {
 		gsvd.err = errors.New("hogsvd: eigen decomposition failed")
 		return false
 	}
-	v := eig.Vectors()
+	vc := eig.VectorsTo(nil)
+	// vc is guaranteed to have real eigenvalues.
```
kortschak: Do we want to have API for Eigen that will make available real eigenvalues if they exist?

mat/cdense.go
kortschak: That's because the `use` is hardcoded into `reuseAs` and `useZeroed` in `reuseAsZeroed`. If you do `reuseAs` and then `zero`, you may allocate and then re-zero the data field.

btracey: Ahh, yes.

mat/eigen_test.go
btracey: Done

mat/eigen_test.go
btracey: Done.

mat/hogsvd.go
btracey: Not in this PR.

More generally I'm not sure. We should eventually have a `CDense -> Dense` helper routine. I don't know how common it is to know you have real eigenvalues unless your matrix is symmetric (at which point you should use EigenSym). At least in my experience with Matlab it's quite common that matrices that are supposed to be real are complex with some tolerance, and I'm not sure what to do about those cases.

kortschak: I'll file an issue to consider this.

mat/cdense.go
```diff
@@ -57,3 +57,112 @@ func NewCDense(r, c int, data []complex128) *CDense {
...
+	}
+	if r == 0 || c == 0 {
+		panic(ErrZeroLength)
+	}
+	if m.IsZero() {
+		m.mat = cblas128.General{
+			Rows:   r,
+			Cols:   c,
+			Stride: c,
+			Data:   useC(m.mat.Data, r*c),
```
kortschak: This should be `useZeroedC(m.mat.Data, r*c)` and add that:
```
// useZeroed returns a float64 slice with l elements, using f if it
// has the necessary capacity, otherwise creating a new slice. The
// elements of the returned slice are guaranteed to be zero.
func useZeroedC(f []complex128, l int) []complex128 {
	if l <= cap(f) {
		f = f[:l]
		zeroC(f)
		return f
	}
	return make([]complex128, l)
}
```

btracey: Done. Thanks and sorry.

kortschak:
I think there were possibly some comments that should have been changed here. There are references to `#308` which I think is gonum/matrix#308 which is now closed.

- https://github.com/gonum/gonum/blob/ea41d6d4e96a82374e193c0f459b094b6768fd99/mat/eigen.go#L145-L146

- https://github.com/gonum/gonum/blob/ea41d6d4e96a82374e193c0f459b094b6768fd99/mat/eigen.go#L300

Do these need to be updated (and surrounding text)?

btracey:
https://github.com/gonum/gonum/pull/858

PR #851, opened 2019-02-12
blas/gonum: slice arguments to exact length

### LINKED ISSUE #733: blas/gonum: []float64 slices should be sliced to their shortest needed length
Currently we us s[n:] which matches the fortran. We are missing the opportunity for safety here by not saying s[n:m] where we know what m is.

vladimir-ch:
Also, avoid unnecessary use of ...To routines.

Please take a look.

### DIFF
--- a/blas/gonum/dgemm.go
+++ b/blas/gonum/dgemm.go
@@ -259,7 +259,7 @@ func dgemmSerialNotNot(m, n, k int, a []float64, lda int, b []float64, ldb int, 
 		for l, v := range a[i*lda : i*lda+k] {
 			tmp := alpha * v
 			if tmp != 0 {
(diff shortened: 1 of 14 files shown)


btracey:
We should really be slicing to the capacity though, right? Some of the lapack bugs come from missed out-of-bounds access, but they can also come from slicing beyond where it should be (this was at least the major problem I encountered with SVD)

kortschak:
That's a good point. doing `s[:n:n]` would prevent reslicing back to an out-of-bound length.

vladimir-ch:
Slicing to the capacity is too tedious, so I postpone it to a future PR. Can I merge if I remove the "Closes ..." clause?

kortschak:
SGTM

PR #860, opened 2019-02-14
graph/internal/set: improve test coverage and docs

kortschak:
There are issues here that I'm not happy with, but I'm not sure how to deal with them. I will leave them for now since this is internal.

Please take a look.

### DIFF
--- a/graph/internal/set/set.go
+++ b/graph/internal/set/set.go
@@ -129,14 +129,17 @@ func (s *Nodes) clear() {
 	}
 }
 
-// Copy performs a perfect copy from src to dst (meaning the sets will
-// be equal).
+// Copy performs a perfect copy from src to dst returning the result.
+// If len(dst) is not 0, a new Nodes is made and returned.
 func (dst Nodes) Copy(src Nodes) Nodes {
+	// FIXME(kortschak): The behaviour is this method is very surprising.
+	// Whether the return is required or not depends on len(dst).
+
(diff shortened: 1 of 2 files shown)


kortschak:
*ping*

graph/internal/set/set.go
```diff
@@ -129,14 +129,17 @@ func (s *Nodes) clear() {
 	}
 }
 
-// Copy performs a perfect copy from src to dst (meaning the sets will
-// be equal).
+// Copy performs a perfect copy from src to dst returning the result.
+// If len(dst) is not 0, a new Nodes is made and returned.
 func (dst Nodes) Copy(src Nodes) Nodes {
+	// FIXME(kortschak): The behaviour is this method is very surprising.
+	// Whether the return is required or not depends on len(dst).
```
sbinet: shouldn't `Copy` take a `*Nodes` as receiver then?

kortschak: That's one option. They are...

1. Make the methods for `Nodes` be on the pointer and always work on the receiver, returning for convenience.
2. Make them package functions that take a destination and return it potentially modified like `append` does.

They each has pros and cons. The amount of work for 2. is much less and I suspect it will have better performance than 1., but 1. is probably going to be easier to read. So I am going to try both and see how they turn out, but not in this PR.

I also think `Copy` should be called `Clone`.

PR #861, opened 2019-02-14
num: add field names to Go syntax printing of types

kortschak:
Please take a look.


kortschak:
I should also fix `%+v`, but I'll do that later.

PR #862, opened 2019-02-15
num: add %+v support to Format method, extend tests and fix found bug

kortschak:
A nice example of why even trivial code wants tests.

Please take a look.


kortschak:
*ping* This is straightforward `fmt.Formatter` code/testing, so should be open to anyone.

Issue #681, opened 2018-11-23
lapack/cgo,lapack/native: consider early checkMatrix and checkVector


kortschak:
In some cgo and native implementations, checkMatrix and checkVector occur after work length query. I have been writing them very early under the Principle of Repair, but we should be behaving consistently.

This issue is to discuss which way we should change the existing code; either to allow work length queries without checks or require all parameters be correct under all conditions.

Issue #680, opened 2017-01-28
lapack/gonum: generally ld? is not checked for conditions specified in LAPACK when matrix not used


kortschak:
In the LAPACK code, the documentation specifies that `LD? >= max(1, <dim>)`. This is always checked. In `native.checkMatrix` we only check that `lda < n`.

I don't have a particular problem with what we do (the Fortran seems excessively strict), but I just wanted to make sure that what we are doing is what we are wanting to do.

PR #864, opened 2019-02-15

vladimir-ch:
Please take a look.

Updates #680 

This goes against #681 but to me it makes better sense and is consistent with what has been done in blas and in some lapack functions.

Further PRs will follow, I didn't want to make them too big.


mat/eigen.go
```diff
@@ -156,10 +156,15 @@ func (e *Eigen) Factorize(a Matrix, left, right bool) (ok bool) {
 	if left {
 		vl = *NewDense(r, r, nil)
 		jobvl = lapack.LeftEVCompute
+	} else {
+		vl.mat.Stride = 1
```
vladimir-ch: I'm not happy about having to add these to `mat`, on the other hand I'd like to have the strides at least 1 as in the reference. Then `lapack64` seems like the right place to assure this, a connection between blas64 structs with zero default values and lapack with strides at least 1.

kortschak: This seems fair.

PR #865, opened 2019-02-18

vladimir-ch:
Next batch.

Updates #680 

Please take a look.


lapack/gonum/dgetrf.go
```diff
@@ -28,16 +28,31 @@ import (
...
+	switch {
+	case m < 0:
+		panic(mLT0)
+	case n < 0:
+		panic(nLT0)
+	case lda < max(1, n):
+		panic(badLdA)
 	}
-	if m == 0 || n == 0 {
-		return false
```
kortschak: Why was this false?

lapack/gonum/errors.go
```diff
@@ -66,21 +76,35 @@ const (
 	negDimension  = "lapack: negative matrix dimension"
 	negZ          = "lapack: negative z value"
 	nLT0          = "lapack: n < 0"
+	nLT1          = "lapack: n < 1"
```
kortschak: What is the ordering that's being used here?

lapack/gonum/dgetrf.go
vladimir-ch: I vaguely remember a small discussion about this a very long time ago. I suppose it was to indicate that no factorization was done but I think that a no-op should not indicate failure.

lapack/gonum/errors.go
vladimir-ch: This is because of `Dlacn2` which uses reverse communication to compute A*x and A^T*x (similarly to `linsolve`) and I didn't see a sensible way in the algorithm to allow `n == 0`. Looking again now I see that I could allow `n == 0` by terminating the iteration loop by directly `return 0, 0` but that deviates from the reference and from the documented behavior. Unlike with more common lapack functions where it may be convenient to relieve the caller from checking the matrix size, here I think it is ok to enforce non-zero problem size.

lapack/gonum/dgetrf.go
kortschak: Yes, I agree. This is better.

lapack/gonum/errors.go
kortschak: Sorry, I mean the ordering of lines. It seems almost alphabetical.

vladimir-ch: Ah, yes, I will clean and reorder the error messages later.

PR #866, opened 2019-02-18
coveralls: paper over service flakiness

kortschak:
Coveralls has been really flakey for the past few days, failing to collect data from builds and causing otherwise successful builds to fail. There appears to be a long history to this, so rather than losing/hiding success information just warn on Coveralls failure.

Please take a look.

Ameliorates lemurheavy/coveralls-public#1264.


kortschak:
Yes, I was just looking at codecov. Depending on what happens here, I think we might want to move.

kortschak:
So with codecov there is no need for the compex rigmarole that we go through with coveralls to collate all the coverage data. Is that right?

Issue #825, opened 2019-01-30
graph: add Reversed method to Edge


kortschak:
### Background
See  gonum/graph#87 (particularly https://github.com/gonum/graph/issues/87#issuecomment-113000044) and #824.

Currently when an undirected graph edge is returned there is a 0.5 chance that the from node of the edge is not the not the node that was used to query the graph. This can be resolved by the user when they know that this is possible and has code in place to handle this, but this can not be done for example in the case of `Copy` where the the input `src` and `dst` are essentially arbitrary graphs. In the case of `Copy` when an undirected graph is copied to a directed graph (if the `src` edge is to be used - see #824) then is it not possible to faithfully copy the edge according to the documentation (that a u--v becomes a pair of edges {u->v, v->u} which I would say is the only sensible approach).

If a graph were able to conditionally reverse an edge when needed, this problem goes away, and client code dealing with hops becomes simpler. This is only really needed for undirected graphs, but the operation makes sense as a general edge operation.

### Proposal
I propose that the `Edge` type be given an additional method `Reversed` that returns a copy of the receiver with the from and to nodes swapped, but all other characteristics remaining the same under the system the model is intended to be used in.

```
func (e edge) Reversed() graph.Edge {
    return edge{f: e.To(), t: e.From(), otherField: e.otherField}
}
```
By implication, the weighted edges and line types would also have this change.

The undirected interfaces would then be made to specify that edges returned are always returned such that the from and to node corresponds to the nodes used to query the `Graph.Edge` method (this is already the case for directed graphs).

### Potential impact of proposal

This is a breaking change; it requires an additional method to be added to all client code using edges or lines that are not provided by the gonum graph packages.
There also needs to be changes to the testgraph package to ensure that implementations satisfy the new semantic requirements.

The changes required in the gonum packages themselves are relatively minor and I would do them.

/cc @mewmew @vladimir-ch

kortschak:
A working draft of this exists in https://github.com/gonum/gonum/compare/graph/reversed currently at 66a1d5964070f3ef8981265181ec81f2a0901b45. There is no documentation change or additional testing for the change yet.

One issue that arises is the need for `graph.Line` to have the `graph.Edge` return type for `Reversed`. I'm not sure if I'm upset about this; `Reversed` is not really for use outside the graph, so the naming clash is not very apparent, but it still looks ugly. It can be fixed with minor changes to testgraph, but those will conflict with #818, so I will not do that until that PR is merged.

[...]

Issue #824, opened 2019-01-30
graph: Copy creates new edges


sevein:
### What are you trying to do?

Create a copy of a graph with `graph.Copy`.

### What did you expect to happen?

My graph had custom nodes and edges values implementing `encoding.Attributer`.
I expected the copy to use the same values.

### What actually happened?

My nodes were used to build the new graph. However the edges were created from scratch:

    dst.SetEdge(dst.NewEdge(u, v))

Instead of something like the following:

    dst.SetEdge(src.Edge(u.ID(), to.Node.ID()))

I'm not sure if this is a bug, sorry for the noise in advance!

### What version of Go and Gonum are you using?

- go: `go version go1.11.4 linux/amd64`
- gonum: `v0.0.0-20190123113241-dd4cc715c58a` (`go.mod`)

### Does this issue reproduce with the current master?

Yes.

kortschak:
Yes, this is working as intended. Perhaps we should clarify the documentation; really what the copy routines do is copy the topology. The intention (which I now see I actually did not completely capture, is to make sure that the copies are separate, using `dst`'s edges would prevent that because there would be references to nodes in `src` within `dst` (this still may currently happen because nodes copied from `src` to `dst` may be pointer or other types with reference-like characteristics.

Because of the massive flexibility with which graphs can be used, I'm not convinced that we can have a generic does-exactly-what-you-want-for-every-ocassion copy function.

I think the best I can do is break it further (making sure that nodes are not copied either) and document that only topology is copied.

Note that with clever use of interfaces, you can actually get a complete copy using these functions, but it's probably easier to just copy-paste the code there and add in the things that you want done for your types.

kortschak:
There are potential further complications; when a copy is made from an undirected graph to a directed graph with the edges taken directly from the `src`, the documented behaviour of having both directions represented in the `dst` may be broken (it is with all the implementations in simple). The is because, at least in simple, edges for u--v and v--u are both returned as edge(u,v), so when that is written into `dst`, only the u->v edge is retained. This is a corner case of https://github.com/gonum/graph/issues/87.

sevein:
Thanks @kortschak, that's really helpful!

> The user would then be responsible for breaking links between the values if necessary (easily done by wrapping the value in a struct that copies out nodes and edges if necessary).

Out of curiosity, that would require the use of unexported fields?

kortschak:
[...]

PR #867, opened 2019-02-19
graph: make edges and lines reversible

kortschak:
I'm sending this as a candidate implementation addressing #825. After having sat on this for a while, I'm fairly convinced this is the right thing to do; if fixes unfortunate subtleties and an otherwise insoluble bug (~~#417~~#824) at the cost of requiring an extra method on edge and line types.

I am not convinced of my approach to naming the methods because, as it stands (the two versions here), either a type assertion is needed for lines, or edges and lines are not interchangeable without wrapping. I think that probably the method names should be `ReversedEdge` and `ReversedLine` to avoid this collision.

Please take a look.

### DIFF
--- a/graph/testgraph/testgraph.go
+++ b/graph/testgraph/testgraph.go
@@ -21,10 +21,6 @@ import (
 	"gonum.org/v1/gonum/mat"
 )
 
-// TODO(kortschak): Consider making the Builder edges function use a package
-// type Edges that lacks the Reversed method so that we can change the Line
-// interface to have Reversed() Line.
-
 // BUG(kortschak): Edge equality is tested in part with reflect.DeepEqual and
 // direct equality of weight values. This means that edges returned by graphs
 // must not contain NaN values. Weights returned by the Weight method are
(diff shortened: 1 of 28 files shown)


kortschak:
I'm going to leave this until @mewmew gives an opinion.

mewmew:
> I'm going to leave this until @mewmew gives an opinion.

I think my comment is similar to the one I added before. It is unfortunate that we have to extend the number of methods of one of the core interfaces of the graph library. But if it solves a problem that cannot be solved (reasonably well) in another way, then, I'm not going to oppose it.

Still wish we could keep core interfaces to their very minimal, but this may be the purist part of me speaking, and not the pragmatic part.

kortschak:
Yes, I understand the reluctance. Everyother strategy I have thought of is either more brittle or requires more differential method addition (we could have undirected and directed edges, but this massively complicates the API).

We are keeping things as simple as possible. But no simpler. Remember, generic graphs are hard.

I'll make the changes to the method names and then this can be reviewed.

mewmew:
> Remember, generic graphs are hard.

I am starting to get an appreciation for this.

kortschak:
I have not put a great deal of thought into that. However, I'm not convinced that it would impact on this issue one way or another.

kortschak:
PTAL

The focus for review would probably be documentation changes and correctness of implementation additions (I found an error in the `WeightedEdgePair` method that would cause the returned edge to not be a weighted edge. (Maybe there should be a note in the reverse methods' documentation that it MUST return the same type as the receiver).

kortschak:
*ping*

graph/testgraph/testgraph.go
```diff
@@ -48,6 +48,27 @@ func checkEmptyIterator(t *testing.T, it graph.Iterator, useEmpty bool) {
 	}
 }
 
+// Edge is supports the simples edge operations.
```
vladimir-ch: Delete "is", and is it "simple" or "simplest"?

kortschak: `// Edge supports basic edge operations.`

kortschak:
PTAL

kortschak:
*ping*

PR #868, opened 2019-02-20
travis: send coverage profile to codecov.io

kortschak:
Please take a look.

This will have two commits. One with push/pr code coverage turned on so that it can be seen working before merge and one removing that to take us back to the previous set-up.


kortschak:
codecov.io report is [here](https://codecov.io/gh/gonum/gonum/commit/fb3c40052081a2abe840ad296d3220b0c1e972c6).

PR #871, opened 2019-02-20
mat: Make factorization parameters part of the matrix type

### LINKED ISSUE #756: mat: consider reconciling option passing in factorise methods

btracey:
Fixes #756.

Please take a look.

### DIFF
--- a/mat/gsvd.go
+++ b/mat/gsvd.go
@@ -18,7 +18,17 @@ import (
 // variable×sample spaces to reduced and diagonalized "eigenvariable"×"eigensample"
 // spaces.
 type GSVD struct {
-	kind GSVDKind
+	// NoU specifies that the U singular vectors should be computed during
+	// the decomposition.
+	NoU bool
+	// NoV specifies that the V singular vectors should be computed during
+	// the decomposition.
+	NoV bool
+	// NoQ specifies that the Q singular vectors should be computed during
+	// the decomposition.
+	NoQ bool
+
+	completed bool
 
 	r, p, c, k, l int
 	s1, s2        []float64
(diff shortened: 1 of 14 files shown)


btracey:
Partial implementation.

This is a possible fix for 756. I think the upside to this implementation is that it requires knowing very little about the decomposition -- it does the standard thing by default. For all the methods it is
```
var fa mat.Factorization
f.Factorize(a)
```
This gives a reasonable factorization for all the types, and if you want optimizations you can set the appropriate fields of the struct.

kortschak:
This seems reasonable, but thinking.

btracey:
Fixed up the remaining parts to this commit. I think moving the factors in is the way to go. Two outstanding points in my opinion:

1) It would be nice if the default behavior of Eigen be to compute the right eigenvectors and not the left eigenvectors, since that's typically what people want. This could be changed to be `Left` and `NoRight`, but I'm not sure it's worth the confusion.

2) I think this approach is clearer for SVD than the bit-fields approach. However, the bitfields approach is actually quite nice for GSVD, especially in the ease of saying GSVDNone. We could keep the bitfields around, add a Kind field, and then flip the meaning so that the default behavior is "All". I do like the reduction in API surface however, so it's not clear that it's worth it.

PTAL

mat/gsvd.go
```diff
@@ -18,7 +18,17 @@ import (
 // variable×sample spaces to reduced and diagonalized "eigenvariable"×"eigensample"
 // spaces.
 type GSVD struct {
-	kind GSVDKind
+	// NoU specifies that the U singular vectors should be computed during
```
kortschak: "should *not*"

and below.

btracey:
Replaced with #872

Issue #748, opened 2016-04-20
mat: Factorization methods should be consistent about usage 


btracey:
Factorization methods (Cholesky, Eigen) should be consistent about their behavior when factorization fails. For example, they should panic when Solve is called.

PR #872, opened 2019-02-20
mat: change factorization inputs to use bit types

### LINKED ISSUE #756: mat: consider reconciling option passing in factorise methods

btracey:
Fixes #756 and #748 .

Please take a look.

### DIFF
--- a/mat/eigen.go
+++ b/mat/eigen.go
@@ -231,7 +231,7 @@ func (e *Eigen) Factorize(a Matrix, kind EigenKind) (ok bool) {
 	return true
 }
 
-// Kind returns the SVDKind of the decomposition. If no decomposition has been
+// Kind returns the EigenKind of the decomposition. If no decomposition has been
 // computed, Kind returns 0.
 func (e *Eigen) Kind() EigenKind {
 	return e.kind
@@ -328,7 +328,7 @@ func (e *Eigen) LeftVectorsTo(dst *CDense) *CDense {
 	if !e.succFact() {
 		panic(badFact)
 	}
-	if e.kind&EigenRight == 0 {
+	if e.kind&EigenLeft == 0 {
 		panic(badNoVect)
 	}
 	if dst == nil {
--- a/mat/eigen.go
+++ b/mat/eigen.go
@@ -104,6 +104,14 @@ func (m *Dense) EigenvectorsSym(e *EigenSym) {
 	m.Copy(e.vectors)
 }
 
+type EigenFact int
+
+const (
+	EigenNoVectors    = 0
+	EigenLeftVectors  = 1 << 0
+	EigenRightVectors = 1 << 1
+)
+
 // Eigen is a type for creating and using the eigenvalue decomposition of a dense matrix.
 type Eigen struct {
 	n int // The size of the factorized matrix.
@@ -141,7 +149,7 @@ func (e *Eigen) succFact() bool {
 //
 // Factorize returns whether the decomposition succeeded. If the decomposition
 // failed, methods that require a successful factorization will panic.
-func (e *Eigen) Factorize(a Matrix, left, right bool) (ok bool) {
+func (e *Eigen) Factorize(a Matrix, fact EigenFact) (ok bool) {
 	// Copy a because it is modified during the Lapack call.
 	r, c := a.Dims()
 	if r != c {
@@ -149,6 +157,9 @@ func (e *Eigen) Factorize(a Matrix, left, right bool) (ok bool) {
 	}
 	var sd Dense
 	sd.Clone(a)
+
+	left := fact&EigenLeftVectors != 0
+	right := fact&EigenRightVectors != 0
 
 	var vl, vr Dense
 	jobvl := lapack.LeftEVNone
--- a/mat/eigen.go
+++ b/mat/eigen.go
@@ -232,8 +232,11 @@ func (e *Eigen) Factorize(a Matrix, kind EigenKind) (ok bool) {
 }
 
 // Kind returns the EigenKind of the decomposition. If no decomposition has been
-// computed, Kind returns 0.
+// computed, Kind returns -1.
 func (e *Eigen) Kind() EigenKind {
+	if !e.succFact() {
+		return -1
+	}
 	return e.kind
 }
 
--- a/mat/eigen.go
+++ b/mat/eigen.go
@@ -104,21 +104,25 @@ func (m *Dense) EigenvectorsSym(e *EigenSym) {
 	m.Copy(e.vectors)
 }
 
-type EigenFact int
+// EigenKind specifies the computation of eigenvectors during factorization.
+type EigenKind int
 
 const (
-	EigenNone  = 0
-	EigenLeft  = 1 << 0
-	EigenRight = 1 << 1
-	EigenBoth  = EigenLeft | EigenRight
+	// EigenNone specifies to not compute any eigenvectors.
+	EigenNone EigenKind = 0
+	// EigenLeft specifies to compute the left eigenvectors.
+	EigenLeft EigenKind = 1 << iota
+	// EigenRight specifies to compute the right eigenvectors.
+	EigenRight
+	// EigenBoth is a convenience value for computing both eigenvectors.
+	EigenBoth EigenKind = EigenLeft | EigenRight
 )
 
 // Eigen is a type for creating and using the eigenvalue decomposition of a dense matrix.
 type Eigen struct {
 	n int // The size of the factorized matrix.
 
-	right bool // have the right eigenvectors been computed
-	left  bool // have the left eigenvectors been computed
+	kind EigenKind
 
 	values   []complex128
 	rVectors *CDense
@@ -127,7 +131,7 @@ type Eigen struct {
 
 // succFact returns whether the receiver contains a successful factorization.
 func (e *Eigen) succFact() bool {
-	return len(e.values) != 0
+	return e.n != 0
 }
 
 // Factorize computes the eigenvalues of the square matrix a, and optionally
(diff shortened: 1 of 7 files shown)


mat/svd.go
```diff
@@ -20,6 +20,33 @@ type SVD struct {
...
+// factorization.
+type SVDKind int
+
+const (
+	// SVDNone specifies that no singular vectors should be computed during
+	// the decomposition.
+	SVDNone SVDKind = 0
+
+	SVDRightThin SVDKind = 1 << (iota - 1)
+	SVDRightFull
```
kortschak: What happens when someone writes `SVDLeftThin|SVDLeftFull`?

btracey:
This is how I would do the bitfields implementation. It's possible to do something closer to https://github.com/gonum/gonum/issues/487#issuecomment-459124715 if we want to have there be ~7 options rather than 9. Note the names should be something like `SVDU` (and can't just be `u`).

I personally prefer #871 , as the case of "getting the factorization" is easier. Also note that this doesn't qutie rectify the factorization patterns, as `LU` and `QR` and friends don't have parameters, so they are called with `Factorize(a)` while these are called with `Factorize(a, settings)`. 

Happy to develop this further if we want to see more of a comparison.

mat/svd.go
btracey: Yes, that would have to be a panic. The three bit version you suggest avoids that, but also commits to the inflexibility of those settings (not clear this is bad, just a statement).

kortschak: I would suggest instead calling them, e.g. `SVDLeft` and `SVDLeftFull` with `SVDLeftFull` implying `SVDLeft`. With that change and a change to the switch so that it's not on the bit *pattern* but rather on the bits, it works better (simpler).

kortschak: Yeah. This is what I did (I was hoping to factor the switch out, but didn't want to fuss with it (it can be done with some extra logic).

```diff
diff --git a/mat/svd.go b/mat/svd.go
index f6d911c..d6c07ed 100644
--- a/mat/svd.go
+++ b/mat/svd.go
@@ -30,9 +30,9 @@ const (
        SVDNone SVDKind = 0
 
        SVDRightThin SVDKind = 1 << (iota - 1)
-       SVDRightFull
+       SVDRight
        SVDLeftThin
-       SVDLeftFull
+       SVDLeft
 
        // SVDThin computes the thin singular vectors, that is, it computes
        //  A = U~ * Σ * V~^T
@@ -44,7 +44,7 @@ const (
        //  A = U * Σ * V^T
        // where U is of size m×m, Σ is an m×n diagonal matrix, and V is an n×n matrix.
        // Hack for now to keep PR small.
-       SVDFull = SVDLeftFull | SVDRightFull
+       SVDFull = SVDLeft | SVDRight
 )
 
 // Factorize computes the singular value decomposition (SVD) of the input matrix A.
@@ -70,15 +70,11 @@ const (
 func (svd *SVD) Factorize(a Matrix, kind SVDKind) (ok bool) {
        m, n := a.Dims()
        var jobU, jobVT lapack.SVDJob
-       switch kind {
+       switch {
        default:
-               panic("svd: bad input kind")
-       case SVDNone:
                svd.u.Stride = 1
-               svd.vt.Stride = 1
                jobU = lapack.SVDNone
-               jobVT = lapack.SVDNone
-       case SVDFull:
+       case kind&SVDLeft != 0:
                // TODO(btracey): This code should be modified to have the smaller
                // matrix written in-place into aCopy when the lapack/native/dgesvd
                // implementation is complete.
@@ -88,15 +84,8 @@ func (svd *SVD) Factorize(a Matrix, kind SVDKind) (ok bool) {
                        Stride: m,
                        Data:   use(svd.u.Data, m*m),
                }
-               svd.vt = blas64.General{
-                       Rows:   n,
-                       Cols:   n,
-                       Stride: n,
-                       Data:   use(svd.vt.Data, n*n),
-               }
                jobU = lapack.SVDAll
-               jobVT = lapack.SVDAll
-       case SVDThin:
+       case kind&SVDLeftThin != 0:
                // TODO(btracey): This code should be modified to have the larger
                // matrix written in-place into aCopy when the lapack/native/dgesvd
                // implementation is complete.
@@ -106,13 +95,33 @@ func (svd *SVD) Factorize(a Matrix, kind SVDKind) (ok bool) {
                        Stride: min(m, n),
                        Data:   use(svd.u.Data, m*min(m, n)),
                }
+               jobU = lapack.SVDStore
+       }
+       switch {
+       default:
+               svd.vt.Stride = 1
+               jobVT = lapack.SVDNone
+       case kind&SVDRight != 0:
+               // TODO(btracey): This code should be modified to have the smaller
+               // matrix written in-place into aCopy when the lapack/native/dgesvd
+               // implementation is complete.
+               svd.vt = blas64.General{
+                       Rows:   n,
+                       Cols:   n,
+                       Stride: n,
+                       Data:   use(svd.vt.Data, n*n),
+               }
+               jobVT = lapack.SVDAll
+       case kind&SVDRightThin != 0:
+               // TODO(btracey): This code should be modified to have the larger
+               // matrix written in-place into aCopy when the lapack/native/dgesvd
+               // implementation is complete.
                svd.vt = blas64.General{
                        Rows:   min(m, n),
                        Cols:   n,
                        Stride: n,
                        Data:   use(svd.vt.Data, min(m, n)*n),
                }
-               jobU = lapack.SVDStore
                jobVT = lapack.SVDStore
        }
 
```

btracey:
Okay, here's another try that I think makes us both happy. It keeps the common cases easy by providing a couple extra declared constants. People who don't want to do anything clever can select their constant. People who do want to do something clever can fiddle with the bit types.

btracey:
If you agree, I'll fix up some of the documentation and checking.

mat/eigen.go
```diff
@@ -104,6 +104,15 @@ func (m *Dense) EigenvectorsSym(e *EigenSym) {
 	m.Copy(e.vectors)
 }
 
+type EigenFact int
+
+const (
+	EigenNone  = 0
+	EigenLeft  = 1 << 0
```
kortschak: `iota` here.

kortschak:
I'm OK with this approach

mat/eigen.go
```diff
@@ -104,6 +104,15 @@ func (m *Dense) EigenvectorsSym(e *EigenSym) {
 	m.Copy(e.vectors)
 }
 
+type EigenFact int
```
vladimir-ch: I think the type name is not accurate, I would reflect in it that it concerns the eigenvectors. Perhaps `EigenVect` or similar?

mat/svd.go
```diff
@@ -20,6 +20,33 @@ type SVD struct {
...
+	// SVDNone specifies that no singular vectors should be computed during
+	// the decomposition.
+	SVDNone SVDKind = 0
+
+	SVDThinU SVDKind = 1 << (iota - 1)
+	SVDFullU
+	SVDThinV
+	SVDFullV
+
+	// SVDThin computes the thin singular vectors, that is, it computes
```
vladimir-ch: I understand that this is a shortcut but it suggests as if there are two kinds singular vectors, thin and non-thin. I think there is only a single kind of (ok, left and right) singular vectors. In the thin case only the singular vectors are computed while in the full case the set of singular vectors (in the first min(m,n) columns of U and V) is extended to form a complete orthonormal basis of R^n.

btracey: Is that a statement about naming, behavior, or documentation? There is only one (non-unique) set of eigenvalues, but the factorized matrix has rank at most min(m,n). The Full decomposition returns all of the vectors. The thin decomposition only computes the ones in the rank of A. This is pretty commonly called the thin decomposition.

vladimir-ch: About documentation. It would be enough to just replace `thin singular vectors` with `thin singular value decomposition and maybe below add `The columns of U~ and V~ contain the left and right singular vectors, respectively.`

mat/eigen.go
```diff
@@ -212,6 +231,12 @@ func (e *Eigen) Factorize(a Matrix, left, right bool) (ok bool) {
 	return true
 }
 
+// Kind returns the SVDKind of the decomposition. If no decomposition has been
```
kortschak: Is this valid? What if the user called `Factorize(a, EigenNone)`?

Also, s/SVDKind/EigenKind/.

btracey: Then Kind() also returns none?

We can change the behavior to return -1 if the factorization was unsuccessful (or not computed at all)?

kortschak: Fair enough. I read it as indication of success, but that was obviously just me.

btracey: I'll switch it to return -1, so that way it can also be used as an indication of success.

kortschak: No, I think it's right as it is.

btracey:
PTAL

mat/svd.go
```diff
@@ -20,6 +20,35 @@ type SVD struct {
...
+type SVDKind int
+
+const (
+	// SVDNone specifies that no singular vectors should be computed during
+	// the decomposition.
+	SVDNone SVDKind = 0
+
+	// SVDThinU specifies the thin decomposition of U should be computed.
+	SVDThinU SVDKind = 1 << (iota - 1)
+	// SVDThinU specifies the full decomposition of U should be computed.
```
vladimir-ch: SVDFullU

Also strictly speaking it doesn't specify the _decomposition_ of U or V but ok.

btracey: I changed "of" to "for". If you have a compact description that is strictly speaking correct I'm happy to change it... "SVDThinU specifies that when the input matrix is factorized, only the vectors of U associated with the non-rank deficient dimensions of the input matrix will be computed and stored". It's correct but it's about twice as long as probably twice as confusing.

kortschak:
Does this also fix #748? I think so.

PR #874, opened 2019-02-21
travis: check that we work with modules

kortschak:
At the moment, I am doing a nasty (but now correct) git workaround to get rid of changes to `go.mod`. These changes come about because the go tool finds the latest versions of the pre-installed tools specified in the `before_install` yaml stanza. To get these to install at the appropriate version, we need to be in the gonum/gonum directory so the go tool can see go.mod, but I don't think this is where travis is by the time that stanza is run.

I think in the long run (once modules are supported by all the versions of Go we support), we can `cd gonum/gonum && go get ./...` in `before_install` to get our dependencies.

Please take a look.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,10 +11,13 @@ go:
  - master
 
 env:
- - TAGS=""
- - TAGS="-tags bounds"
- - TAGS="-tags noasm"
- - TAGS="-tags appengine"
+ global:
+  - GO111MODULE=on
+ matrix:
+  - TAGS=""
+  - TAGS="-tags bounds"
+  - TAGS="-tags noasm"
+  - TAGS="-tags appengine"
 
 cache:
  directories:
--- a/.travis/check-generate.sh
+++ b/.travis/check-generate.sh
@@ -4,8 +4,13 @@ go generate gonum.org/v1/gonum/blas
 go generate gonum.org/v1/gonum/blas/gonum
 go generate gonum.org/v1/gonum/unit
 go generate gonum.org/v1/gonum/graph/formats/dot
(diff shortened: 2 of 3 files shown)


kortschak:
*ping*

.travis.yml
```diff
@@ -11,10 +11,13 @@ go:
  - master
 
 env:
- - TAGS=""
- - TAGS="-tags bounds"
- - TAGS="-tags noasm"
- - TAGS="-tags appengine"
+ global:
+  - GO111MODULE=on
```
sbinet: this enables the "module-mode" for all variants.
is it really what we want to do?

I'd argue that `GO111MODULE=on` should be set:
- for `master` (it will be the default for 1.13, eventually)
- for a matrix entry for 1.12

otherwise, we might get some impedance mismatch from users who are still using `GOPATH`+`GO111MODULE=auto`.

kortschak: >this enables the "module-mode" for all variants.
> is it really what we want to do?

Yes. It's ignored for go1.10 and tests that go1.11 and go1.12 use modules correctly.

> otherwise, we might get some impedance mismatch from users who are still using `GOPATH`+`GO111MODULE=auto`.

I don't see a failure mode in that direction. Can you given me an example of how that might fail in a meaningful way for a user? The only failures that I can see happening when not in module mode are when the user does not have an adequately recent version of golang.org/x/exp/rand, and that is only a test failure, not a package failure. We don't import anything outside gonum.org/v1/gonum except stdlib and golang.org/x/exp/rand.

kortschak:
*ping*

PR #875, opened 2019-02-25

### LINKED ISSUE #680: lapack/gonum: generally ld? is not checked for conditions specified in LAPACK when matrix not used

vladimir-ch:
Closes #680

Please take a look.


lapack/gonum/dormbr.go
```diff
@@ -45,40 +45,43 @@ import (
...
+		panic(badSide)
+	case trans != blas.NoTrans && trans != blas.Trans:
+		panic(badTrans)
+	case m < 0:
+		panic(mLT0)
+	case n < 0:
+		panic(nLT0)
+	case k < 0:
+		panic(kLT0)
+	case applyQ && lda < max(1, min(nq, k)):
```
kortschak: Just a note; no change necessary.

This could be written as `applyQ && lda < max(1, min(nq, k)), !applyQ && lda < max(1, nq)`. However, I think it is better as you have it since the panic line gives the reason for the failure where the flattened approach would require digging.

lapack/gonum/dormhr.go
```diff
@@ -4,7 +4,9 @@
 
 package gonum
 
-import "gonum.org/v1/gonum/blas"
+import (
```
kortschak: Can you undo this please.

lapack/gonum/dorml2.go
```diff
@@ -23,30 +23,49 @@ import "gonum.org/v1/gonum/blas"
...
+		panic(badTau)
+	case len(c) < (m-1)*ldc+n:
+		panic(shortC)
+	case left && len(work) < n:
+		panic(shortWork)
+	case !left && len(work) < m:
+		panic(shortWork)
+	}
+
+	notran := trans == blas.NoTrans
```
kortschak: s/notran/notrans/G?

lapack/gonum/dormr2.go
```diff
@@ -4,7 +4,9 @@
 
 package gonum
 
-import "gonum.org/v1/gonum/blas"
+import (
```
kortschak: Please undo this.

lapack/gonum/dsytrd.go
```diff
@@ -75,11 +68,16 @@ func (impl Implementation) Dsytrd(uplo blas.Uplo, n int, a []float64, lda int, d
 		panic(shortWork)
 	}
 
+	// Quick return if possible.
 	if n == 0 {
 		work[0] = 1
 		return
 	}
 
+	opts := "U"
```
kortschak: Didn't we change these so that `blas.Lower = 'L'` and `blas.Upper = 'U'`? So `nb := impl.Ilaenv(1, "DSYTRD", string(uplo), n, -1, -1, -1)`. Or is this to save the alloc?

lapack/gonum/dtrexc.go
```diff
@@ -46,31 +46,37 @@ import "gonum.org/v1/gonum/lapack"
...
-	if (ilst < 0 || n <= ilst) && n > 0 {
-		panic("lapack: ilst out of range")
+	switch {
+	case compq != lapack.UpdateSchur && compq != lapack.UpdateSchurNone:
+		panic(badUpdateSchurComp)
+	case n < 0:
+		panic(nLT0)
+	case ldt < max(1, n):
+		panic(badLdT)
+	case ldq < 1 || (compq == lapack.UpdateSchur && ldq < n):
```
kortschak: `case ldq < 1, compq == lapack.UpdateSchur && ldq < n:`

lapack/gonum/dorgbr.go
```diff
@@ -20,37 +20,41 @@ import "gonum.org/v1/gonum/lapack"
...
-			checkMatrix(n, n, a, lda)
-		}
+
+	// Quick return if possible.
+	if m == 0 || n == 0 {
+		work[0] = 1
+		return
 	}
+
 	work[0] = 1
```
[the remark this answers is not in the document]
vladimir-ch: Done.

lapack/gonum/dorgqr.go
```diff
@@ -29,12 +29,20 @@ import (
...
+		panic(nGTM)
 	case k < 0:
 		panic(kLT0)
-	case n < k:
+	case k > n:
 		panic(kGTN)
-	case m < n:
-		panic(mLTN)
+	case lda < max(1, n) && lwork != -1:
+		// Dgesvd calls Dorgqr with a placeholder value for lda (and a)
```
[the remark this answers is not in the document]
vladimir-ch: I changed the wording but not sure if it's what you expected.

lapack/gonum/dorm2r.go
```diff
@@ -23,37 +23,50 @@ import "gonum.org/v1/gonum/blas"
 //
 // Dorm2r is an internal routine. It is exported for testing purposes.
 func (impl Implementation) Dorm2r(side blas.Side, trans blas.Transpose, m, n, k int, a []float64, lda int, tau, c []float64, ldc int, work []float64) {
-	if side != blas.Left && side != blas.Right {
+	left := side == blas.Left
```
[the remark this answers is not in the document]
vladimir-ch: If the variable is used at more than just a couple of places, use the variable, otherwise replace it with `side == blas.Left`. Here I would leave the variable.

lapack/gonum/dormbr.go
vladimir-ch: Ok, I will leave it unchanged.

lapack/gonum/dormhr.go
vladimir-ch: It must be one of the formatting tools that does this. Done.

lapack/gonum/dorml2.go
vladimir-ch: Done here and everywhere else.

lapack/gonum/dsytrd.go
vladimir-ch: I just used what was here before. Now replaced with your suggestion.

lapack/gonum/dtrexc.go
vladimir-ch: Done here and everywhere else.

vladimir-ch:
PTAL

lapack/gonum/dorgqr.go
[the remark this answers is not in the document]
kortschak: That's great. Thanks.

PR #876, opened 2019-02-26
ci: update supported versions

kortschak:
Please take a look.


kortschak:
AppVeyor is slow.

kortschak:
@sbinet Can you restart the AppVeyor tests. Their approach to permissions means that I can't do that. Same for the other repos.

Issue #260, opened 2017-10-11
gonum: consider making gonum.org/v1/gonum a Go package


sbinet:
right now, there is no `.go` file directly under `gonum.org/v1/gonum`.

beginners trying to install `gonum` with a naive `go get` are met with:

```sh
$> go get gonum.org/v1/gonum
can't load package: package gonum.org/v1/gonum: no Go files in /home/binet/dev/go/gocode/src/gonum.org/v1/gonum
```

`glide` even tells a more confusing story:
```sh
$> glide get gonum.org/v1/gonum
[INFO]    Preparing to install 1 package.
[INFO]    Attempting to get package gonum.org/v1/gonum
[INFO]    --> Gathering release information for gonum.org/v1/gonum
[INFO]    --> Adding gonum.org/v1/gonum to your configuration
[INFO]    Downloading dependencies. Please wait...
[INFO]    --> Fetching updates for github.com/chewxy/gorgonia
[INFO]    --> Fetching updates for github.com/spf13/viper
[INFO]    --> Fetching updates for github.com/spf13/cobra
[INFO]    --> Fetching updates for gopkg.in/cheggaaa/pb.v2
[INFO]    --> Fetching gonum.org/v1/gonum
[INFO]    --> Fetching updates for github.com/chewxy/lingo
[INFO]    --> Fetching updates for github.com/pkg/errors
[INFO]    --> Fetching updates for github.com/olekukonko/tablewriter
[WARN]    Unable to checkout gonum.org/v1/gonum
[ERROR]    Update failed for gonum.org/v1/gonum: https://gonum.org/v1/gonum Not Found
[ERROR]    Failed to checkout packages: https://gonum.org/v1/gonum Not Found
```

we probably should just stick a `gonum.go` file with some documentation to make all of this disappear.

kortschak:
This has been discussed somewhere. For there to be a package, there should be a function that it provides, not just documentation. I'd rather this didn't happen. The install docs do say install `gonum.org/v1/gonum/...` (note that `go get gonum.org/v1/gonum` with a placeholder can lead to more confusing results - gocode will not see any of the packages since they must be built, and that `go get` invocation would not do that).

sbinet:
you can put only documentation:

https://github.com/go-hep/hep/blob/master/hep.go
https://godoc.org/go-hep.org/x/hep

the issue you raise about `gocode` might indeed happen, I guess.
[...]

PR #877, opened 2019-02-26
gonum: add Version query function

kortschak:
Please take a look.

Updates #260.

### DIFF
--- a/version.go
+++ b/version.go
@@ -1,0 +1,41 @@
+// Copyright ©2019 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// +build go1.12
+
+package gonum
+
+import (
+	"fmt"
+	"runtime/debug"
+)
+
+const root = "gonum.org/v1/gonum"
+
+// Version returns the version of Gonum and its checksum. The returned
+// values are only valid in binaries built with module support.
+func Version() (version, sum string) {
+	b, ok := debug.ReadBuildInfo()
+	if !ok {
+		return "", ""
+	}
+	for _, m := range b.Deps {
+		if m.Path == root {
+			if m.Replace != nil {
+				switch {
+				case m.Replace.Version != "" && m.Replace.Path != "":
+					return fmt.Sprintf("%s %s", m.Replace.Path, m.Replace.Version), m.Replace.Sum
+				case m.Replace.Version != "":
+					return m.Replace.Version, m.Replace.Sum
+				case m.Replace.Path != "":
+					return m.Replace.Path, m.Replace.Sum
+				default:
+					return m.Version + "*", ""
+				}
+			}
+			return m.Version, m.Sum
+		}
+	}
+	return "", ""
+}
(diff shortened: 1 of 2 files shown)


kortschak:
Are we OK with the format of the versions returned? Do we want to warn that this may change?

kortschak:
*ping*

version.go
```diff
@@ -0,0 +1,41 @@
...
+	for _, m := range b.Deps {
+		if m.Path == root {
+			if m.Replace != nil {
+				switch {
+				case m.Replace.Version != "" && m.Replace.Path != "":
+					return fmt.Sprintf("%s %s", m.Replace.Path, m.Replace.Version), m.Replace.Sum
+				case m.Replace.Version != "":
+					return m.Replace.Version, m.Replace.Sum
+				case m.Replace.Path != "":
+					return m.Replace.Path, m.Replace.Sum
```
sbinet: shouldn't this be `return fmt.Sprintf("%s %s*", m.Replace.Path, m.Version), m.Replace.Sum` instead?

version.go
```diff
@@ -0,0 +1,41 @@
...
+				default:
+					return m.Version + "*", ""
```
sbinet: `s/return m.Version + "*", ""/return m.Version + "*", m.Sum+"*"/` perhaps?

version.go
kortschak: I think so, if `m.Version != ""`.

version.go
kortschak: OK

kortschak:
PTAL

kortschak:
I have a little module in my local called `latest_gonum`. This could be place either as a module in gonum/gonum or could be put in gonum/tools. What do you think?

sbinet:
how would that work? it seems a bit cumbersome to anticipate the checksum of the whole `gonum/gonum` repo...
(hence my suggestion of breaking `Version()` into `Version() + versionOf(b)` so we could write tests using `versionOf(b)` with a manufactured `*debug.BuildInfo` we control...)

kortschak:
It's not for testing. It's a command that can be used to find the latest version. The version doesn't need to be anticipated, the go.mod can be artificially created with `latest` as the version.

I didn't break for testing because the complexity is not in the switch, but rather then `debug.BuildInfo` structure that `debug.ReadBuildInfo` can return. That's outside of our knowledge.

sbinet:
> It's not for testing. It's a command that can be used to find the latest version.

ah. ok. on the basis that it would be a bit silly to install `gonum.org/v1/gonum/cmd/latest_gonum` to get that binary (but then, one could just look into the repo one had just installed), I'd err on putting it under `gonum/tools`.
but that's really just to break the tie.

> I didn't break for testing because the complexity is not in the switch, but rather then debug.BuildInfo structure that debug.ReadBuildInfo can return. That's outside of our knowledge.

yes, but at least this would help understand what's going one if/when `ReadBuildInfo` returns some slightly different output. (under `Go1` compatibility rules that shouldn't happen, of course...)
but ok, this can be left as is.

kortschak:
> on the basis that it would be a bit silly to install gonum.org/v1/gonum/cmd/latest_gonum to get that binary

These would have the same effect. The `latest_gonum` module itself would be small, but would pull in the gonum/gonum module. But this is true also of the gonum/tools module approach.

PR #878, opened 2019-02-27
Convert two files from ISO-8859-1 to UTF-8

tomfuture:
There are just a couple source files that are encoded as  ISO-8859-1 instead of UTF-8.  Let's make them UTF-8, unless there's a reason for them not to be.


kortschak:
Please fix the commit message: 
```
integrate/quad/internal: fix character encoding

Convert from ISO-8859-1 to UTF-8
```

and add another commit adding yourself to A+C with the commit message:
```
A+C: add <your real name>
```

kortschak:
`A+C: add Thomas Berg`

PR #879, opened 2019-02-28
LICENSE: include licenses for third party code

kortschak:
Please take a look.


sbinet:
out of curiosity, where is Boost coming from?

kortschak:
[cephes](https://github.com/gonum/gonum/blob/master/mathext/internal/cephes/lanczos.go) from SciPy.

btracey:
Thanks for doing this, I've been thinking about this recently. Maybe we should add a README to it saying something like "portions of the code are subject to the following additional licenses. They are all either BSD or MIT".

THIRD_PARTY_LICENSES/Cephes-README
```diff
@@ -0,0 +1,17 @@
+Cephes Math Library Release 2.8:  June, 2000
```
btracey: This should copy over the extra note about the license:

Additional copyright information:
Code in this package is adapted from the Cephes library (http://www.netlib.org/cephes/).
" There is no explicit licence on Netlib, but the author has agreed to a BSD release.
See https://github.com/deepmind/torch-cephes/blob/master/LICENSE.txt and
https://lists.debian.org/debian-legal/2004/12/msg00295.html "

kortschak: Yeah, I was looking in to that last night. It's difficult, but I think this probably should be added too.

kortschak: I had a look at the debian-legal list comments and was not comfortable with adding anything more than what I have because of that; note that debian does not include cephes in the debs repo, possibly because of this.

I'd be happy to revisit this, but probably after having written to him with a specific text for a license file for him to approve.

kortschak: I have written to Steve Moshier to ask for his approval to include a BSD 3-clause license with his copyright.

kortschak:
Steve Moshier has granted permission to license under a BSD. I have included the email exchange in the commit message adding the BSD-3 license (removing some mail header lines showing mail paths and machine IPs/addresses).

THIRD_PARTY_LICENSES/Cephes-README
kortschak: Done

PR #882, opened 2019-03-03
num/dualcmplx: package for dual complex maths

kortschak:
Please take a look.

/cc @Armadilloa16

### DIFF
--- a/num/dualcmplx/dual.go
+++ b/num/dualcmplx/dual.go
@@ -9,8 +9,6 @@ import (
 	"math"
 	"math/cmplx"
 	"strings"
-
-	"gonum.org/v1/gonum/num/dual"
 )
 
 // Number is a float64 precision anti-commutative dual complex number.
@@ -86,12 +84,12 @@ func Add(x, y Number) Number {
 // Sub returns the difference of x and y, x-y.
 func Sub(x, y Number) Number {
 	return Number{
-		Real: x.Real + y.Real,
-		Dual: x.Dual + y.Dual,
+		Real: x.Real - y.Real,
+		Dual: x.Dual - y.Dual,
 	}
 }
 
-// Mul returns the dual product of x and y.
+// Mul returns the dual product of x and y, x×y.
 func Mul(x, y Number) Number {
 	return Number{
 		Real: x.Real * y.Real,
@@ -121,8 +119,8 @@ func Scale(f float64, d Number) Number {
 }
 
 // Abs returns the absolute value of d.
-func Abs(d Number) dual.Number {
-	return dual.Number{Real: cmplx.Abs(d.Real)}
+func Abs(d Number) float64 {
+	return cmplx.Abs(d.Real)
 }
 
 // PowReal returns d**p, the base-d exponential of p.
@@ -166,7 +164,7 @@ func PowReal(d Number, p float64) Number {
 		if d.Real == -1 {
 			return Number{Real: 1, Dual: cmplx.NaN()}
 		}
-		if Abs(d).Real > 1 {
+		if Abs(d) > 1 {
 			if d.Dual == 0 {
 				return Number{Real: cmplx.Inf(), Dual: cmplx.NaN()}
 			}
@@ -177,7 +175,7 @@ func PowReal(d Number, p float64) Number {
 		if d.Real == -1 {
 			return Number{Real: 1, Dual: cmplx.NaN()}
 		}
-		if Abs(d).Real > 1 {
+		if Abs(d) > 1 {
 			return Number{Real: 0, Dual: cmplx.NaN()}
 		}
 		if d.Dual == 0 {
--- a/num/dualcmplx/dual.go
+++ b/num/dualcmplx/dual.go
@@ -1,0 +1,269 @@
+// Copyright ©2018 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package dualcmplx
+
+import (
+	"fmt"
+	"math"
+	"math/cmplx"
+	"strings"
+
+	"gonum.org/v1/gonum/num/dual"
+)
+
+// Number is a float64 precision anti-commutative dual complex number.
+type Number struct {
+	Real, Dual complex128
+}
+
+var zero Number
+
+// Format implements fmt.Formatter.
+func (d Number) Format(fs fmt.State, c rune) {
+	prec, pOk := fs.Precision()
+	if !pOk {
+		prec = -1
+	}
+	width, wOk := fs.Width()
+	if !wOk {
+		width = -1
+	}
+	switch c {
+	case 'v':
+		if fs.Flag('#') {
+			fmt.Fprintf(fs, "%T{Real:%#v, Dual:%#v}", d, d.Real, d.Dual)
+			return
+		}
+		if fs.Flag('+') {
+			fmt.Fprintf(fs, "{Real:%+v, Dual:%+v}", d.Real, d.Dual)
+			return
+		}
+		c = 'g'
+		prec = -1
+		fallthrough
+	case 'e', 'E', 'f', 'F', 'g', 'G':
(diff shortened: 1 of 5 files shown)


num/dualcmplx/dual.go
```diff
@@ -0,0 +1,269 @@
...
+	return Number{
+		Real: x.Real + y.Real,
+		Dual: x.Dual + y.Dual,
+	}
+}
+
+// Sub returns the difference of x and y, x-y.
+func Sub(x, y Number) Number {
+	return Number{
+		Real: x.Real + y.Real,
```
Armadilloa16: minus?

Armadilloa16:
```suggestion
		Real: x.Real - y.Real,
```

num/dualcmplx/dual.go
```diff
@@ -0,0 +1,269 @@
...
+		Dual: x.Dual + y.Dual,
+	}
+}
+
+// Mul returns the dual product of x and y.
```
Armadilloa16: , x*y (order or multiplication is important so be explicit)

num/dualcmplx/dual.go
```diff
@@ -0,0 +1,269 @@
...
+func Mul(x, y Number) Number {
+	return Number{
+		Real: x.Real * y.Real,
+		Dual: x.Real*y.Dual + x.Dual*cmplx.Conj(y.Real),
+	}
+}
+
+// Inv returns the dual inverse of d.
+func Inv(d Number) Number {
```
Armadilloa16: Maybe this would be more efficient if d.Real * cmplx.Conj(d.Real) where replaced by |d.Real|^2? Not sure how the complex number implementation works.

num/dualcmplx/dual.go
```diff
@@ -0,0 +1,269 @@
...
+	}
+}
+
+// Scale returns d scaled by f.
+func Scale(f float64, d Number) Number {
+	return Number{Real: complex(f, 0) * d.Real, Dual: complex(f, 0) * d.Dual}
+}
+
+// Abs returns the absolute value of d.
+func Abs(d Number) dual.Number {
```
Armadilloa16: This will always return an actual real number (not complex or dual), is there a good reason for it to return this wrapped as a dual number?

num/dualcmplx/dual.go
```diff
@@ -0,0 +1,269 @@
...
+		if p < 0 {
+			return Number{Real: 0, Dual: cmplx.NaN()}
+		}
+		return Number{Real: cmplx.Inf(), Dual: cmplx.NaN()}
+	}
+	return Pow(d, Number{Real: complex(p, 0)})
+}
+
+// Pow returns d**p, the base-d exponential of p.
+func Pow(d, p Number) Number {
```
Armadilloa16: I don't see why we can do this in the dual numbers.... it might be true, but it is not obvious to me.

num/dualcmplx/dual_example_test.go
```diff
@@ -0,0 +1,134 @@
...
+// transform performs the transformation of p by the given set of dual
+// complex transforms in order. The rotations are normalized to unit
+// vectors.
+func transform(p point, by ...dualcmplx.Number) point {
+	if len(by) == 0 {
+		return p
+	}
+
+	// Ensure the modulus of by is correctly scaled.
+	for i := range by {
```
Armadilloa16: Doesn't this negate any dilations?

num/dualcmplx/dual.go
kortschak: Thanks. Fixed.

kortschak: Also the dual.

num/dualcmplx/dual.go
kortschak: Done.

num/dualcmplx/dual.go
kortschak: This is how the paper expresses it. Defn 2.2.

num/dualcmplx/dual.go
kortschak: None. Simplified.

num/dualcmplx/dual.go
kortschak: This is essentially Defn 5.3.

num/dualcmplx/dual_example_test.go
kortschak: Yes, intentionally. In general cases dilation may be warranted, I didn't want it here.

num/dualcmplx/dual.go
Armadilloa16: Yeah it's equivalent I was just wondering about computational efficiency. It doesn't really matter at this point though, premature optimisation if anything.

num/dualcmplx/dual.go
Armadilloa16: Huh, yeah you're right. Good enough for me, I'll figure out how exactly that works later at some point. Makes sense in principle though, so yeah. I suppose it does work. That's pretty neat.

num/dualcmplx/dual_example_test.go
Armadilloa16: Ok, in that case I'd change 'rotations' to either 'transformations' or some combination or rotations and translations, as this does allow translations. You could mention this is actualy SE(2) as in the paper, which is standard notation in group/ring theory.

kortschak: I need to remember to make the same change to dualquat.

kortschak:
*ping*

kortschak:
There's not a particular hurry.

num/dualcmplx/dual_test.go
```diff
@@ -0,0 +1,456 @@
...
+func TestFormat(t *testing.T) {
+	for _, test := range formatTests {
+		got := fmt.Sprintf(test.format, test.d)
+		if got != test.want {
+			t.Errorf("unexpected result for fmt.Sprintf(%q, %#v): got:%q, want:%q", test.format, test.d, got, test.want)
+		}
+	}
+}
+
+// FIXME(kortschka): See golang/go#29320.
```
vladimir-ch: Your username :-) (in case you grep for it in the future)

num/dualcmplx/dual_example_test.go
```diff
@@ -0,0 +1,132 @@
...
+		z = dualcmplx.Mul(o, z)
+	}
+	pp := dualcmplx.Mul(dualcmplx.Mul(z, raiseDual(p)), dualcmplx.Conj(z))
+
+	// Extract the point.
+	return point{x: real(pp.Dual), y: imag(pp.Dual)}
+}
+
+func Example() {
+	// Translate a 1×1 square [3, 4] and rotate it 90° around the
```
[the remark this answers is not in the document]
kortschak: Done

num/dualcmplx/dual_example_test.go
```diff
@@ -0,0 +1,132 @@
...
+	// origin.
+	fmt.Println("square:")
+
+	// Construct a displacement.
+	displace := dualcmplx.Number{
+		Real: 1,
+		Dual: 0.5 * raise(point{3, 4}),
+	}
+
+	// Construct a rotations.
```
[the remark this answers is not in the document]
kortschak: Done

num/dualcmplx/dual_example_test.go
```diff
@@ -0,0 +1,132 @@
...
+		pp.x = floats.Round(pp.x, 2)
+		pp.y = floats.Round(pp.y, 2)
+
+		fmt.Printf(" %d %+v -> %+v\n", i, p, pp)
+	}
+
+	// Rotate a line segment 90° around its lower end [2, 2].
+	fmt.Println("\nline segment:")
+
+	// Construct an displacement to the origin from the lower end...
```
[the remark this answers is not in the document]
kortschak: Done

num/dualcmplx/dual_test.go
kortschak: Done

PR #885, opened 2019-03-05
A+C: add Kirill Motkov

sosiska:
needed by gonum/plot#533


sbinet:
needed by https://github.com/gonum/plot/pull/533

PR #887, opened 2019-03-06
graph/formats/dot,travis: update gocc version

kortschak:
Please take a look.


kortschak:
I broke gocc. When the fix is merged, I'll ping here with an update.

kortschak:
PTAL

PR #888, opened 2019-03-07
graph/formats/dot/fuzz: add fuzz corpus

kortschak:
Please take a look.

Note that fuzzing with this fuzz function did find two inputs that failed with the Go parser but were accepted by `dot`. These were (originally) "" and "\\x00". I think we can happily ignore those.


graph/formats/dot/fuzz/fuzz.go
```diff
@@ -0,0 +1,33 @@
...
+package fuzz
+
+import (
+	"bytes"
+	"os/exec"
+
+	"gonum.org/v1/gonum/graph/formats/dot"
+)
+
+func Fuzz(data []byte) int {
```
mewmew: Perhaps not necessary, but add a doc comment just describing what `Fuzz` does with the given data, and what is returned.

graph/formats/dot/internal/lexer/lexer_test.go
```diff
@@ -52,3 +54,35 @@ func TestParseFile(t *testing.T) {
 		}
 	}
 }
+
+func TestParseFuzz(t *testing.T) {
+	r, err := zip.OpenReader("../../fuzz/corpus.zip")
+	if err != nil {
+		if os.IsNotExist(err) {
+			t.Skip("no corpus")
```
mewmew: How does `t.Skip` work? Would you not have to add a `return` statement after this?

graph/formats/dot/internal/parser/parser_test.go
```diff
@@ -86,6 +88,38 @@ func TestParseFile(t *testing.T) {
 	}
 }
 
+func TestParseFuzz(t *testing.T) {
+	r, err := zip.OpenReader("../../fuzz/corpus.zip")
+	if err != nil {
+		if os.IsNotExist(err) {
+			t.Skip("no corpus")
```
mewmew: Likewise here, is a `return` statement needed?

graph/formats/dot/internal/parser/parser_test.go
```diff
@@ -86,6 +88,38 @@ func TestParseFile(t *testing.T) {
...
+		t.Fatalf("failed to open corpus: %v", err)
+	}
+	defer r.Close()
+
+	for _, f := range r.File {
+		rc, err := f.Open()
+		if err != nil {
+			t.Fatalf("failed to open %q: %v", f.Name, err)
+		}
+		func() {
```
mewmew: What is the purpose of this anonymous function? Is it just to capture the scope of the local variables? It does not seem to be used in either a Go-routine statement or a defer statement.

graph/formats/dot/fuzz/fuzz.go
kortschak: I'll just add a link to github.com/dvyukov/go-fuzz.

graph/formats/dot/internal/lexer/lexer_test.go
kortschak: The return is not necessary. `Skip` works like `Fatal` it finishes the running goroutine.

graph/formats/dot/internal/parser/parser_test.go
kortschak: Same.

graph/formats/dot/internal/parser/parser_test.go
kortschak: This is to allow recovery of panics for each individual test case. Otherwise only the first panic would be found.

kortschak:
PTAL

graph/formats/dot/internal/parser/parser_test.go
mewmew: Ok, thanks. I learnt something new :)

PR #889, opened 2019-03-07
lapack/gonum: clean up panic strings and their usage

vladimir-ch:
Please take a look.


lapack/gonum/errors.go
```diff
@@ -6,134 +6,163 @@ package gonum
...
-	shortWR       = "lapack: insufficient length of wr"
-	shortWV       = "lapack: insufficient length of wv"
-	shortX        = "lapack: insufficient length of x"
-	shortY        = "lapack: insufficient length of y"
-	shortZ        = "lapack: insufficient length of z"
-	shortIWork    = "lapack: insufficient length of iwork"
-	shortWork     = "lapack: working array shorter than declared"
-	zeroDiv       = "lapack: zero divisor"
-
+	badApplyOrtho      = "lapack: bad ApplyOrtho"
```
kortschak: What is the ordering used here?

vladimir-ch: The first block are bad enum values, then bad general parameter values, bad vector increments, bad slice lengths, insufficient slice lengths, and bad leading dimensions. The order of the blocks is arbitrary.

Within a block the strings are ordered alphabetically.

kortschak: We should probably have a comments saying what each block is so that someone coming later doesn't just place new things wherever.

vladimir-ch: Actually I had comments for each block but then decided against them. I've put them back now.

PR #891, opened 2019-03-09
unit/constant: new package for dimensioned fundamental constants

kortschak:
Please take a look.

Adding constants is trivial, but this seemed like a reasonable starting set.


kortschak:
Just a note on the choice of unexported types. The use of unexported types does unofortunately prevent the use of the types in signatures, however I don't believe this is a huge cost; a function is not likely to take a constant as a parameter (but will take variables as parameters). The upside of using unexported types is that there is not large quantities of apparently duplicated documentation/method signatury.

unit/constant/generate_constants.go
```diff
@@ -0,0 +1,327 @@
...
+	"math"
+	"os"
+	"strings"
+	"text/template"
+
+	"gonum.org/v1/gonum/unit"
+)
+
+const (
+	elementaryCharge = 1.6021766208e-19
```
btracey: For instance, in about 2 months this value will be defined to be 1.602176634e-19

unit/constant/generate_constants.go
```diff
@@ -0,0 +1,327 @@
+// Copyright ©2014 The Gonum Authors. All rights reserved.
```
btracey: I think we should make these values accurate to the May 2019 redefinition (even if we're a bit early). We should think about what sort of compatibility guarantees we have. I would rather us update as changes happen / science gets better, but it's not clear what the right behavior is.

Also, did you have a standard practice for defining the constants? Would be nice to cite a reference or some canonical source. This question is especially related to the "uncertain" digits, where dropping them is effectively the same as setting them to zero.

unit/constant/generate_constants.go
```diff
@@ -0,0 +1,327 @@
...
+var constants = []Constant{
+	{
+		Name: "AtomicMass", Value: 1.660539040e-27,
+		Dimensions: []Dimension{{massName, 1}},
+		Comment:    "AtomicMass is the atomic mass constant (mᵤ), one twelfth of the mass of an unbound atom of carbon-12 at rest and in its ground state.",
+	},
+	{
+		Name: "Avogadro", Value: 6.02214076e23,
+		Dimensions: []Dimension{{moleName, -1}},
+		Comment:    "Avogadro is the Avogadro constant (A), the number of constituent particles contained in one mole of a substance.",
```
btracey: I think the constants that are defined should have a note that they are so (i.e. actually constant, at least until the next SI revision in 50 years)

kortschak:
I'll fix these up in a bit. I would like the constants that are here that use derived units defined in unit to use those. This makes them more flexible, e.g. `const StandardGravity = unit.Acceleration(9.80665)`. I need to figure out how to do that logic without too much configuration.

unit/constant/generate_constants.go
kortschak: I trawled WP. Is there a good standard document?

btracey: NIST releases a guide: https://physics.nist.gov/cuu/index.html , but it's not up to date with the SI redefinition

kortschak: What would you prefer wrt uncertainty?

I'll add some notes to the doc.go talking about stability of values.

unit/constant/generate_constants.go
kortschak: I will add this for defined constants.

unit/constant/generate_constants.go
kortschak: I think for the purposes of compatibility it makes sense to say in the docs that we use the most recent NIST values (and not precede them). This gives us a clear reason for changes.

kortschak:
PTAL

kortschak:
*ping*

unit/constant/generate_constants.go
```diff
@@ -0,0 +1,334 @@
...
+	fineStructure    = 7.2973525664e-3
+	lightSpeed       = 2.99792458e8
+	planck           = 6.626070040e-34
+)
+
+var constants = []Constant{
+	{
+		Name: "AtomicMass", Value: 1.660539040e-27,
+		Dimensions:  []Dimension{{massName, 1}},
+		Comment:     "AtomicMass is the atomic mass constant (mᵤ), one twelfth of the mass of an unbound atom of carbon-12 at rest and in its ground state.",
```
btracey: Oh man. I know I'm a huge nerd, but I'm actually excited to redo these definitions in May! Grand-K free SI!

unit/constant/generate_constants.go
```diff
@@ -0,0 +1,334 @@
...
+		Dimensions: []Dimension{{lengthName, 1}, {timeName, -1}},
+		Comment:    "LightSpeedInVacuum is the c constant, the speed of light in a vacuum.",
+	},
+	{
+		Name: "MagneticConstant", Value: 2 * fineStructure * planck / (elementaryCharge * elementaryCharge * lightSpeed),
+		Dimensions: []Dimension{{currentName, 2}, {timeName, 4}, {massName, -1}, {lengthName, -3}},
+		Comment:    "MagneticConstant is the magnetic constant (μ₀), the magnetic permeability in a classical vacuum.",
+	},
+	{
+		Name: "Planck", Value: planck,
```
btracey: We should also have ReducedPlank (hbar). Doesn't have to happen this PR.

kortschak: Can happen in the update.

unit/constant/generate_constants.go
kortschak: Do you want to file an issue to remind us, or will you have an alarm already set?

Issue #730, opened 2015-09-04
unit: Package documentation has unimplemented features


scott-maddox:
Many of the examples in `doc.go` do not work. For example, `unit.Temperature`, `unit.Pressure`, `unit.Density`, `unit.Bar`, `unit.Energy`, and `unit.Area` do not exist.

Currently, it is unclear how this package is meant to be used. Are the features in the documentation features that used to exist that are now deprecated, or are they features that are still desired? I would be happy to help improve the package and documentation, if the authors could explain their vision.

Thanks!

jonshao:
Seems like this is still an issue. Have there been any developments?

kortschak:
A more [complete example](https://godoc.org/gonum.org/v1/gonum/unit#ex-package--Horsepower) was recently added. Can you point to specific issues in the docs and/or examples that you would like to see?

kortschak:
OK. Just taken a look. The examples in the doc.go comment should be removed and replaced with similar runnable examples. The truth of some of that documentation is clearly rotted. However because of the design of the package user-provided concrete types can be used to illustrate the same behaviours. I think that we probably should not be providing derived types, but that could change. I may be able to find some time to do this over the next few days.

@btracey Do you have any views on this?

kortschak:
To help with this, I'd like to propose a ".../unit/constant" package that includes `float64`-derived types satisfying `unit.Uniter`. This allows us to provide physical constants in a straightforward way.

btracey:
I agree we should clean it up. I think we probably should follow the existing documentation rather than the existing code and split out the SI prefixes. This would have been clearly better from the start except for the fact that `Kilogram` being the base unit messes everything up. Maybe we provide prefixes for everything except Kilogram where we provide the full suite? 

I think we should provide derived types. I don't know how to draw the line on where given that there are an infinite number of them. I would say that most things with an SI name (Joules, Pascals, etc.) is good to add, plus a couple common ones without names (Acceleration, Area, Speed). 

I like the idea of a `constant` package in Unit.

kortschak:
I have added the base SI units to the package in #890. I don't think that it's necessary to add derived types after putting together the contant package; user types that satisfy `unit.Uniter` are enough. I would like to merge #890 and then write some examples that show how compile-time type safe types can be written. This makes a base to work from. If a limited set of derived types is seen to be needed then those can be added later.

[...]

PR #893, opened 2019-03-10
unit: add named derived units

kortschak:
Please take a look.

I have not added L/dm^3 since I don't know how to handle that. It might just be necessary to us L rather than the correct unit.

Updates #730.


unit/generate_unit.go
```diff
@@ -250,29 +251,267 @@ var Units = []Unit{
...
+		TypeComment: "Velocity represents a velocity in meters per second",
+		Dimensions: []Dimension{
+			{Name: LengthName, Power: 1},
+			{Name: TimeName, Power: -1},
+		},
+	},
+	{
+		Name:        "Voltage",
+		Receiver:    "v",
+		PrintString: "Pa",
```
btracey: Volts, not pascals

unit/generate_unit.go
```diff
@@ -250,29 +251,267 @@ var Units = []Unit{
...
+		Suffix:      "volt",
+		Singular:    "Volt",
+		TypeComment: "Voltage represents a voltage in Volts",
+		Dimensions: []Dimension{
+			{Name: CurrentName, Power: -1},
+			{Name: LengthName, Power: 2},
+			{Name: MassName, Power: 1},
+			{Name: TimeName, Power: -2},
```
btracey: This should be -3, right?

unit/generate_unit.go
```diff
@@ -250,29 +251,267 @@ var Units = []Unit{
...
+	},
+	{
+		Name:        "Inductance",
+		Receiver:    "i",
+		PrintString: "H",
+		Suffix:      "henry",
+		Singular:    "Henry",
+		TypeComment: "Inductance represents an electrical inductance in Henry",
+		Dimensions: []Dimension{
+			{Name: CurrentName, Power: -1},
```
btracey: This should be -2?

unit/generate_unit.go
kortschak: I'm inventing a new system of unified units.

kortschak: ... but fixed.

btracey: Go is about making things simpler, afterall. I would probably have tried to start with Candela, but we can always get rid of that later.

unit/generate_unit.go
kortschak: Fixed

unit/generate_unit.go
kortschak: Fixed

kortschak:
PTAL

kortschak:
Queries:

1. Are you happy with the capitalisation of the units? There are many an varied ways to do this.
2. What should be do with volume? Just use m^3 or use L as the rendering and have liter-based constants with m^3 the underlying unit (I blame the chemists for this).

btracey:
Assuming you're referring to things like "EquivalentRadioactiveDose", then I think the current choices are reasonable.

I think provide Volume in m^3, and then define L relative to that. I see cL and mL frequently, but I've never heard someone talk about megameters^3. It's sad that it's messy, but at least it's not imperial. There are aspects of imperial I really like, but volume is a total disaster.

kortschak:
> Assuming you're referring to things like "EquivalentRadioactiveDose", then I think the current choices are reasonable.

No, I was thinking about initial capitals of Watt etc in the docs (some people write watt other Watt). Also pluralisation of units in docs. I have also made the Kelvin suffix "Kelvin" rather than "kelvin" (I don't know why  I should be consistent). Maybe we sort out this in the #90 handling of prefix constants?

kortschak:
Sorry. PTAL

btracey:
Fully sorting it out with 90 seems reasonable. I prefer megawatt to megaWatt, so I like the current way.

kortschak:
Thanks for the review.

PR #895, opened 2019-03-11
graph/flow: new package for (control) flow analysis

kortschak:
This PR does the initial movement from #796. It is happening here so we can get the API breakage done now, allowing the additions to come later at a more relaxed pace.

Please take a look.


kortschak:
@sbinet Can yoy make it so that I can restart AppVeyor jobs here please?

kortschak:
Thanks

kortschak:
Have fun in Korea.

Issue #483, opened 2018-05-08
optimize[?]: Add root finding subpackage


btracey:
We should have a root-finding package, either in `gonum/root` or in `gonum/optimize/root`.

I'm not exactly sure how this package should look. An idea is to roughly copy the design of `optimize`, using the reverse call technique. This could allow for the same controls as in optimize (and could allow for really nice integration with `optimize` and `root`), but might be a bit overkill. The other tricky factor with `root` is that, unlike optimize, there are a lot of one-dimensional root finding problems and algorithms specifically designed with one-dimensional functions. That is, we may want to have `root` and `rootmv`. `rootmv` could possibly live in `optimize` itself, but this may make documentation hard, since some of the methods will belong to Minimize and some to Root.

Thoughts @vladimir-ch @chewyx ?

chewxy:
Personally I've never ran into needing to use multivariate root finding algorithms... so from a purely YAGNI point of view, that's essentially something that doesn't exist to me.

But I like very much the idea of the root finding packge

vladimir-ch:
I prefer `gonum/root` over `gonum/optimize/root`. The connection between them is mathematical and can be considered an implementation detail that does not need to be reflected in how packages are organized. What we may find useful is to take the `Linesearcher` implementations out of optimize and put them into their own package imported both by optimize and root.

If at all possible I would also prefer to avoid having `rootmv`. In optimize we default to multidimensional problems, the same will be true for `ode` which will eventually be a heavy user of `root`. We do have integrate.Trapezoidal but I'm not really happy about it. If a method works only with 1d problems, it can panic if used on an nd problem.

btracey:
Agreed on `gonum/root`.

I'm not sure I agree that only having a multi-variate solution is the right one.

`Integrate`, for example, does admittedly need much help as a package, but it seems clear to me that having a unidimensional package is a good idea. There are clever algorithms for integration that either only work in 1-D or have multi-dimensional extensions that are significantly more complex. The scaling for integration is typically terrible (i.e. exponential) with dimension, and so in high dimensions you switch classes of algorithms entirely (`samplemv` territory).

Root seems somewhere in the middle of optimize and integrate. Root finding methods often start with a bounding box that gets shrunken. There are some techniques you can do that will scale poorly, but others that won't.

[...]

PR #896, opened 2019-03-12
optimize: add Brent implementation

sbinet:
Please take a look.

### DIFF
--- a/optimize/brent.go
+++ b/optimize/brent.go
@@ -1,0 +1,266 @@
+package optimize
+
+import "math"
+
+const (
+	phi           = 1.618043 // phi is the golden ratio
+	tiny          = 1e-21
+	limit float64 = 110
+)
+
+type brentIterType int
+
+var (
+	_ Method = &Brent{}
+)
+
+const (
+	brentGo brentIterType = iota
+	brentBrakA
+	brentBrakB
+	brentBrakC
+	brentBrak1
+	brentBrak2
+	brentBrak3
+	brentBrak4
+	brentBrak5 // noop
+)
+
+// Brent is an optimization method from Richard Brent's "Algorithms for Minimization without Derivatives" (page 79)
+type Brent struct {
+	Min, Max, Limit float64 // brackets of absiccas
+
+	x              float64
+	a, b, c        float64
+	fa, fb, fc, fw float64
+	denom, w, wlim float64
+	iter           brentIterType
+}
+
+func (b *Brent) Init(loc *Location) (Operation, error) {
+	if len(loc.X) != 1 {
+		panic("Expect only 1 parameter")
+	}
+	if b.Max <= b.Min && b.Max == 0 {
+		b.Max = 1
+	}
+
+	b.a = b.Min
+	b.b = b.Max
+	b.x = loc.X[0]
+
+	loc.X[0] = b.Min
+	b.iter = brentBrakA
+	return FuncEvaluation, nil
+}
+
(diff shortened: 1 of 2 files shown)


sbinet:
@chewxy, @btracey  could you have a look at this?
I've adapted (I think) what you had for the new `optimize` design.

but it seems there are a few things that aren't working...

btracey:
Sorry for the delay. We didn't add this before because the discussion was that we should have a `root` package instead of putting the root-finding methods into optimize. There is some discussion on organization and design: https://github.com/gonum/gonum/issues/483

vladimir-ch:
Why couldn't go this into optimize? Variants of the Brent's method can be used both for minimization and root finding of functions of one variable. What I'm not sure is whether this should implement `Method` or whether it should have its own entry function.

btracey:
If it goes in here, I don't see why it should have its own entry function. Are there that many optimization algorithms that are specific to one-D that aren't root-finding algorithms? My experience is that the standard algorithms work pretty well in one-d (either local linesearch, or ListSearch, or some kind of global algorithm). It seems like if you want a bisection-like algorithm, it would be better to have those implemented in `root` and have a wrapper there which satisfies `optimize.Problem`.

vladimir-ch:
I don't insist on Brent having its own entry function. The suggestion comes from the fact that `Method` was made for general n-dimensional problems while Brent works only for 1-dimensional problems. Perhaps a standalone function could offer a better and simpler API.

I'm confused about using an implementation of Brent in `root` for 1-d minimization. How would that work without differentiation?

btracey:
Sorry, I meant to suggest that there could be a wrapper in either `root` or
`optimize` that takes a root-finding method and turns it into an
`optimize.Method`. You basically just shift one level of derivative
(Meaning, in `root` the user needs to supply Function and Gradient, the
wrapper would need Gradient and Hessian, which is fine in 1-d).

My main point is not wanting duplicate implementations of Bisection,
GoldenSearch, etc. in both `root` and `optimize` when the algorithms are
virtually identical, just one derivative shifted.

Both of the above points are orthogonal to the discussion of having a
second entry function for 1-D problems. I think we would only have to add a
new entry function and a new Problem type, so it's not very much API
surface. I'm not sure in gains very much though. Brent (for example) would
implement Method (in my hypothetical design), so it would have to check
it's not used for a multi-D function. I probably would still choose LBFGS
as a default method, so you don't gain that much there either. Seems like
the main benefit is a better matching Problem type. That could be enough
benefit to justify the API surface, unclear.

One idea is to have `root.Bounded` which is 1-d, and `root.Find` which is
many-D. We could use the same paradigm in optimize if we think it is worth
it, `optimize.Bounded` would be One-D. It would require passing a bound,
but we could enable methods to support inf, and so practically could
support unbounded choices.

vladimir-ch:
About the wrapper: I understand what you mean but I'm still confused. How the "shift one level of derivative" part would be done? I can imagine going the other way around: in `root` take `f` and pass `|f|^2` to `optimize` in order to find the roots of `f` but even in that case I'm not convinced. I would just have two implementations, one in `root` and one in `optimize`.

About the second entry function for 1-d-only methods: it seems to me that having them all implement `Method` and checking that the dim is 1 is the best. Otherwise I'm afraid we would end up replicating `Method` and `Minimize` only this time with scalars instead of `[]float64`.

vladimir-ch:
@sbinet Would you like to keep this open? I would eventually be able to take a look at it.

PR #897, opened 2019-03-12
graph: use src edges in dst graph

### LINKED ISSUE #824: graph: Copy creates new edges

kortschak:
Please take a look.

Fixes #824.
Depends #867.

### DIFF
--- a/graph/graph.go
+++ b/graph/graph.go
@@ -244,10 +244,11 @@ func Copy(dst Builder, src Graph) {
 	nodes.Reset()
 	for nodes.Next() {
 		u := nodes.Node()
-		to := src.From(u.ID())
+		uid := u.ID()
+		to := src.From(uid)
 		for to.Next() {
 			v := to.Node()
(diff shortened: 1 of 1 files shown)


kortschak:
This will need to be re-approved after I rebase onto master to get it passing.

kortschak:
Thanks

PR #900, opened 2019-03-15
graph/internal/set: put Nodes methods on the pointer type

### LINKED ISSUE #863: graph/internal/set: fix mutation and return behaviour of set types
There are surprising behaviours of `Nodes.Copy` and the methods dependent on `Nodes.Copy` resulting from the fact that the method on for `Nodes` are not on a pointer, so when a `Nodes` needs to be cleared (it has a non-zero length), the receiver is no longer the recipient of mutations.

I think the solution of this is to make `Copy`, `Union` and `Intersect` package functions `CopyNodes` (maybe `CloneNodes`?), `UnionNodes` and `IntersectNodes` rather than methods on `Nodes` and note that the returned value should always be assigned, as is the case with `append`. The alternative is to put the methods on `*Nodes`, but I think this is too much indirection.

The package is internal to graph, so this should not cause widespread problems.

kortschak:
Please take a look.

Fixes #863.

### DIFF
--- a/graph/community/k_communities.go
+++ b/graph/community/k_communities.go
@@ -48,20 +48,20 @@ func KCliqueCommunities(k int, g graph.Undirected) [][]graph.Node {
 					nodes.Add(n)
 				}
 			}
-			if len(nodes) < k {
-				for _, n := range nodes {
+			if nodes.Count() < k {
+				for _, n := range *nodes {
 					single.Add(n)
 				}
 				continue
 			}
 			var kc []graph.Node
-			for _, n := range nodes {
+			for _, n := range *nodes {
 				inCommunity.Add(n)
 				kc = append(kc, n)
 			}
 			kcc = append(kcc, kc)
 		}
-		for _, n := range single {
+		for _, n := range *single {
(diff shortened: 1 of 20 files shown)


kortschak:
This has a fairly significant impact on at least one set-intense routine.

<table class='benchstat'>
<tr><th>name</th><th>old time/op</th><th>new time/op</th><th>delta</th>
<tr><td>KCliqueCommunities/simple-k=3-8</td><td>76.1µs ± 2%</td><td>88.2µs ± 2%</td><td>+15.80%</td><td>(p=0.000 n=17+20)</td>
<tr><td>KCliqueCommunities/Batagelj-Zaversnik_Graph-k=3-8</td><td>230µs ± 3%</td><td>258µs ± 3%</td><td>+12.47%</td><td>(p=0.000 n=20+20)</td>
<tr><td>KCliqueCommunities/Batagelj-Zaversnik_Graph-k=4-8</td><td>229µs ± 2%</td><td>258µs ± 2%</td><td>+12.71%</td><td>(p=0.000 n=19+20)</td>
</table>

PR #901, opened 2019-03-15
opimize/nlls: first implementation of Levenberg-Marquardt.

maorshutman:
Please take a look. 

@vladimir-ch


optimize/nlls/lm_test.go
```diff
@@ -0,0 +1,242 @@
+package lm
```
sbinet: please add a license header.

also: the package is named `lm` while the directory holding the code is `nlls`.
this is a bit confusing. they should somewhat match.

optimize/nlls/lm_test.go
```diff
@@ -0,0 +1,242 @@
...
+
+import (
+	"math"
+	"testing"
+
+	"gonum.org/v1/gonum/floats"
```
sbinet: please regroup the `gonum.org` imports together.

optimize/nlls/lmopt.go
```diff
@@ -0,0 +1,221 @@
+/*
```
sbinet: please add a license header.

optimize/nlls/lmopt.go
```diff
@@ -0,0 +1,221 @@
...
+			}
+		}
+	}
+	return &Result{
+		X:      parameters,
+		Status: status,
+	}, nil
+}
+
+// LMProblem is used for running LM optomization. The objective function is
```
sbinet: `s/optomization/optimization/`

btracey:
@sbinet @vladimir-ch , thoughts on location? I don't think it should be in optimize proper, and so I suggested this. Optimize is for black-box objective functions, while L-M assumes that you're solving least-squares and that you can evaluate each point separately. I have a bunch of things it would be nice to bring in at github.com/btracey/sgd . SGD has the same property that you have a bunch of different pieces that you assume you can evaluate separately. However, I can't think of any reasonable package name that unifies that and this, and I've been thinking about it on and off for quite some time. It seems like `nlls` is as good as any.

btracey:
At a high level, we need to decide what this package looks like. In optimize, we have `optimize.Optimize`, which abstracts what you're trying to do (minimize the function) with how you want to do it (the method). I believe we will want a similar approach here. Thinking about it here, I suspect we could make that function use optimize, which would enable a huge class of methods easily.

Rather than do all of that here, could we simplify this PR to be just the implementation of the core of LM? That is, have a type that is `LevenbergMarquardt` that has a method `Step(dst, theta, y []float64, x, jac mat.Matrix)`. This method computes the next set of parameters theta and stores the result in dst. A test for this would then have a simple implementation which just calls step for a few iterations.  I'm not totally sure that function name and/or signature is right, but it'll be a much smaller set of things to review and decide for now.

optimize/nlls/lm_test.go
maorshutman: Changed to nlls.

optimize/nlls/lmopt.go
maorshutman: Fixed.

maorshutman:
So the Idea to have a type `LevenbergMarquardt` that stores all the variables that define the current state of the computation. I can add a method `Solve` that will call `Step` until convergence and then the test file will call this `Solve` method instead of `LM` for each test problem. @btracey Do I understand your proposal correctly?

vladimir-ch:
A couple of quick comments.

* I think this should go into `exp` first.
* L-M is a particular method in the trust region class of methods. I think it would be nice to have `TrustRegionMethod` (similarly to `LinesearchMethod`) that uses something like `TrustRegionStrategy` (like `NextDirectioner`) of which `LevenbergMarquardt` would be a concrete implementation.
* We should decide what NLLS model we want to minimize. Saying "sum of functions" is not enough. How the objective function looks like exactly? What is in this PR assumes that each residual depends on all parameters, leading to a dense Jacobian. However, most practical problems have some structure and the Jacobians will be sparse. How to put all of this together - the modeling of the objective function and how this structure will be communicated to the solver, a storage format for the Jacobian, and a linear solver (various factorizations, iterative)? I don't know how to disentangle this so that we could start with something small and isolated.
* I can imagine that this could all live in `optimize` by extending `Problem` and having another entry point routine (`optimize.LeastSquares`). Or the objective-function modeling could be in its own package and the solver in `optimize`.
* What are the expectations of SGD on the objective function? In my understanding it is "evaluate a given term in the sum" and "evaluate the gradient of a given term in the sum". I think that SGD and related methods don't care about the Jacobian, so it is less than what L-M needs. Apart from these considerations I wouldn't bring the stochastic methods into the discussion about L-M/trust region methods.

btracey:
I think those are mostly good ideas by @vladimir-ch . The thing I don't like is `optimize.LeastSquares`. That itself is fine, but is the plan to also have `optimize.LSProblem`, etc.? At that point, I think I'd rather have `nlls.Minimize` (or `lsq.` or `nlsq.`). It separates what and how to solve pretty cleanly. I don't think there's that much to be gained in having them in the same package.

vladimir-ch:
> I think I'd rather have nlls.Minimize (or lsq. or nlsq.). It separates what and how to solve pretty cleanly. I don't think there's that much to be gained in having them in the same package.

That sounds good to me.

@maorshutman Do you want to re-submit this to `gonum/exp` where we would have more freedom to experiment and sort out the API?

maorshutman:
Done.

PR #903, opened 2019-03-15
unit: add dimension and value accessor methods

kortschak:
I found the absence of these irritating when I was working on the recent changes.

Please take a look.

### DIFF
--- a/unit/unittype.go
+++ b/unit/unittype.go
@@ -130,6 +130,31 @@ var (
 // dimension is zero. Dimensions is used in conjunction with New.
 type Dimensions map[Dimension]int
 
+func (d Dimensions) clone() Dimensions {
+	if len(d) == 0 {
+		return nil
+	}
+	c := make(Dimensions, len(d))
+	for dim, pow := range d {
+		if pow != 0 {
+			c[dim] = pow
+		}
+	}
+	return c
+}
+
+func (d Dimensions) matches(o Dimensions) bool {
+	if len(d) != len(o) {
+		return false
+	}
+	for dim, pow := range d {
+		if o[dim] != pow {
+			return false
+		}
+	}
+	return true
+}
+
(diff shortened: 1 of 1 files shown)


unit/unittype.go
```diff
@@ -130,6 +130,31 @@ var (
...
+	for dim, pow := range d {
+		if pow != 0 {
+			c[dim] = pow
+		}
+	}
+	return c
+}
+
+func (d Dimensions) matches(o Dimensions) bool {
+	if len(d) != len(o) {
```
btracey: There should be a comment somewhere (maybe in Unit.DimensionsMatch) that we require zero values to be deleted as keys. Just something so that if we change implementations in future there's a warning that we have some counter-intuitive behavior.

unit/unittype.go
```diff
@@ -130,6 +130,31 @@ var (
 // dimension is zero. Dimensions is used in conjunction with New.
 type Dimensions map[Dimension]int
 
+func (d Dimensions) clone() Dimensions {
+	if len(d) == 0 {
+		return nil
+	}
+	c := make(Dimensions, len(d))
+	for dim, pow := range d {
+		if pow != 0 {
```
btracey: (wrt below) this shouldn't be necessary because all of the 0 dimensions shouldn't have keys. On the one hand, it's not bad to check, on the other hand if we have some failure of this invariant we disguise it. Not sure what the best plan is.

unit/unittype.go
kortschak: Done

unit/unittype.go
kortschak: It is necessary to have it here, because `Dimensions` is an exported and user-mutable type. We must remove any zeros prior to use. The zeroing on handing back to the user is a side-effect.

kortschak:
PTAL

PR #904, opened 2019-03-15
unit: make printing of units concurrently safe and unicode-aware

kortschak:
Please take a look.

Updates #90.

### DIFF
--- a/unit/unittype.go
+++ b/unit/unittype.go
@@ -222,9 +222,6 @@ func (u unitPrinters) Swap(i, j int) {
 type Unit struct {
 	dimensions Dimensions
 	value      float64
-
-	mu        sync.RWMutex
-	formatted string
 }
 
 // New creates a new variable of type Unit which has the value and dimensions
@@ -285,7 +282,6 @@ func (u *Unit) Mul(uniter Uniter) *Unit {
 			u.dimensions[key] = d + val
 		}
 	}
-	u.setFormattedUnits("")
 	u.value *= a.value
 	return u
 }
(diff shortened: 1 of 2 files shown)


unit/unittype.go
```diff
@@ -212,8 +214,10 @@ func SymbolExists(symbol string) bool {
 // mass to get a force. See the package documentation for further explanation.
 type Unit struct {
 	dimensions Dimensions
-	formatted  string
 	value      float64
+
+	mu        sync.RWMutex
```
btracey: We should just ditch this and recompute the formatting string on the fly. It's not that expensive an operation, it's way simpler to avoid the mutex, and it simplifies other methods because we don't track if we've already computed it or not.

unit/unittype.go
```diff
@@ -322,25 +326,40 @@ func (u *Unit) Format(fs fmt.State, c rune) {
 	case 'e', 'E', 'f', 'F', 'g', 'G':
 		p, pOk := fs.Precision()
 		w, wOk := fs.Width()
-		if u.formatted == "" && len(u.dimensions) > 0 {
-			u.formatted = u.dimensions.String()
+		formatted := u.formattedUnits()
+		if formatted == "" && len(u.dimensions) > 0 {
+			formatted = u.dimensions.String()
```
btracey: Note also that we still aren't concurrency safe because `NewDimension` can be called during formatting. See my comment in #902

unit/unittype.go
kortschak: I'm not convinced, but we can add it back if needed.

OK

unit/unittype.go
kortschak: I missed that. You are right. This complicates things significantly.

kortschak: I have fixed this. It requires locking `symbols` and `dimensions` and only accessing those through a small set of functions.

kortschak:
PTAL

PR #906, opened 2019-03-17
 graph/encoding/{,di}graph6: add support for digraph6- and graph6-encoded graphs

kortschak:
Please take a look.

Note that the copyright date is last year because that's when it was written.

### DIFF
--- a/graph/encoding/digraph6/digraph6.go
+++ b/graph/encoding/digraph6/digraph6.go
@@ -1,0 +1,338 @@
+// Copyright ©2018 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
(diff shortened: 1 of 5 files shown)


graph/encoding/digraph6/digraph6.go
```diff
@@ -0,0 +1,338 @@
...
+	}
+	if size%6 != 0 {
+		buf.WriteByte(c + 63)
+	}
+
+	return Graph(buf.String())
+}
+
+// IsValid returns whether the graph is a valid digraph6 encoding. An invalid Graph
+// is behaves a the null graph.
```
vladimir-ch: s/is/ and s/a/as?
Also for graph6.

kortschak: Done

PR #907, opened 2019-03-19
 graph/internal/set: make complex set operations pure functions 

### LINKED ISSUE #863: graph/internal/set: fix mutation and return behaviour of set types

kortschak:
Looking at the benchmarks for this approach, we have a better outcome than the original code for `community.KCliqueCommunities` and a better outcome than the pointer approach in #900 for `topo.BronKerbosch` and `topo.CliqueGraph`.

### Original versus this approach

<table class='benchstat'>
<tr><th>name</th><th>original time/op</th><th>pure function time/op</th><th>delta</th>
<tr><td>KCliqueCommunities/simple-k=3-8</td><td>76.1µs ± 2%</td><td>61.1µs ± 3%</td><td>-19.81%</td><td>(p=0.000 n=17+20)</td>
<tr><td>KCliqueCommunities/Batagelj-Zaversnik_Graph-k=3-8</td><td>230µs ± 3%</td><td>184µs ± 1%</td><td>-19.83%</td><td>(p=0.000 n=20+17)</td>
<tr><td>KCliqueCommunities/Batagelj-Zaversnik_Graph-k=4-8</td><td>229µs ± 2%</td><td>183µs ± 1%</td><td>-20.02%</td><td>(p=0.000 n=19+18)</td>
<tr><td>BronKerbosch/wikipedia_example-8</td><td>19.0µs ± 2%</td><td>21.4µs ± 3%</td><td>+12.39%</td><td>(p=0.000 n=20+20)</td>
<tr><td>BronKerbosch/Batagelj-Zaversnik_Graph-8</td><td>106µs ± 6%</td><td>111µs ± 6%</td><td>+4.62%</td><td>(p=0.000 n=20+20)</td>
<tr><td>CliqueGraph/simple-8</td><td>54.5µs ± 3%</td><td>58.1µs ± 2%</td><td>+6.52%</td><td>(p=0.000 n=20+20)</td>
<tr><td>CliqueGraph/Batagelj-Zaversnik_Graph-8</td><td>149µs ± 6%</td><td>156µs ± 1%</td><td>+4.75%</td><td>(p=0.000 n=20+16)</td>
</table>

### Pointer method approach versus this approach

<table class='benchstat'>
<tr><th>name</th><th>pointer method time/op</th><th>pure function time/op</th><th>delta</th>
<tr><td>BronKerbosch/wikipedia_example-8</td><td>24.7µs ± 6%</td><td>21.4µs ± 3%</td><td>-13.22%</td><td>(p=0.000 n=20+20)</td>
<tr><td>BronKerbosch/Batagelj-Zaversnik_Graph-8</td><td>123µs ± 5%</td><td>111µs ± 6%</td><td>-9.78%</td><td>(p=0.000 n=20+20)</td>
<tr><td>CliqueGraph/simple-8</td><td>62.1µs ± 1%</td><td>58.1µs ± 2%</td><td>-6.53%</td><td>(p=0.000 n=18+20)</td>
<tr><td>CliqueGraph/Batagelj-Zaversnik_Graph-8</td><td>167µs ± 4%</td><td>156µs ± 1%</td><td>-6.18%</td><td>(p=0.000 n=20+16)</td>
</table>

I also think the clarity of the calls is not bad (surprisingly to me).

Finally, the purity of the functions means that other potentially surprising behaviours that I did not notice previously are now not possible.

All up I think this is the correct approach.

Please take a look.

Fixes #863.
Closes #900.

### DIFF
--- a/graph/community/k_communities_test.go
+++ b/graph/community/k_communities_test.go
@@ -5,6 +5,7 @@
 package community
 
 import (
+	"fmt"
 	"reflect"
 	"sort"
 	"testing"
@@ -43,11 +44,13 @@ var batageljZaversnikGraph = []intset{
 }
 
 var kCliqueCommunitiesTests = []struct {
+	name string
 	g    []intset
 	k    int
 	want [][]graph.Node
 }{
 	{
+		name: "simple",
 		g: []intset{
 			0: linksTo(1, 2, 4, 6),
 			1: linksTo(2, 4, 6),
@@ -65,8 +68,9 @@ var kCliqueCommunitiesTests = []struct {
 		},
 	},
 	{
-		g: batageljZaversnikGraph,
-		k: 3,
+		name: "Batagelj-Zaversnik Graph",
+		g:    batageljZaversnikGraph,
+		k:    3,
 		want: [][]graph.Node{
 			{simple.Node(0)},
 			{simple.Node(1)},
@@ -84,8 +88,9 @@ var kCliqueCommunitiesTests = []struct {
 		},
 	},
 	{
-		g: batageljZaversnikGraph,
-		k: 4,
+		name: "Batagelj-Zaversnik Graph",
+		g:    batageljZaversnikGraph,
+		k:    4,
 		want: [][]graph.Node{
 			{simple.Node(0)},
 			{simple.Node(1)},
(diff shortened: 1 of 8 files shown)


kortschak:
I'm happy to make that change.

kortschak:
PTAL

PR #911, opened 2019-03-21
github: ask for module version information

kortschak:
Please take a look.


kortschak:
For the record, I have used grep on the go.sum rather than asking for `go list -m gonum.org/v1/gonum` or similar because I don't see a way to also get the sum for the gonum module.

sbinet:
I figured as much :)

PR #914, opened 2019-03-23
graph/traverse: reduce edge confusion

### LINKED ISSUE #913: graph/traverse: {DeptFirst,BreadthFirst}.Visit is is potentially misleading
Use `DepthFirst.Walk` to copy edges in a dependency graph (see discussion at https://groups.google.com/d/msg/gonum-dev/xupu8gEmuIs/NUn9Zvo8CAAJ).

### What did you do?

Reproducer is [here](https://groups.google.com/group/gonum-dev/attach/4e8683b7ed58/deps.go?part=0.1&authuser=0&view=1).

All edges on the walk be replicated.

### What actually happened?
Edges on walk with To nodes already visited are missing.

This is exactly as the code says. However, the presence of the From node in the `Visit` signature may lead the user to believe that it can be used for such a task (as I did). There is a way to capture the edges that are being walked; the `EdgeFilter` closure is called for all edge hops, independent of whether the To node has already been visited. The user *can* check whether the To node has been visited by using the `Visited` method. So the functionality remains, but the subtlety that is this issue goes away. I would also argue that visited-conditional hops on edges are incorrect given that traversal order is not defined beyond depth-first and breadth-first.

kortschak:
Please take a look.

It could be argued that the `Visit` could go away completely, since the `until` parameter can now be used to do the same thing. However, `until` differs between `BreadthFirst` and `DepthFirst` so that BF can stop after a certain depth. This means that a visitors for `BreadthFirst` and `DepthFirst` would not be interchangeable. For this reason I have kept it.

Fixes #913.

### DIFF
--- a/graph/traverse/traverse.go
+++ b/graph/traverse/traverse.go
@@ -27,22 +27,34 @@ type Graph interface {
 
 // BreadthFirst implements stateful breadth-first graph traversal.
 type BreadthFirst struct {
+	// Visit is called on all nodes on their first visit.
+	Visit func(graph.Node)
+
+	// EdgeFilter is called on all edges that may be traversed
+	// during the walk. This includes edges that would hop to
+	// an already visited node.
+	//
+	// The value returned by EdgeFulter determines whether an
+	// edge can be traverse during the walk.
 	EdgeFilter func(graph.Edge) bool
-	Visit      func(u, v graph.Node)
-	queue      linear.NodeQueue
-	visited    set.Int64s
+
+	queue   linear.NodeQueue
+	visited set.Int64s
 }
 
 // Walk performs a breadth-first traversal of the graph g starting from the given node,
 // depending on the the EdgeFilter field and the until parameter if they are non-nil. The
 // traversal follows edges for which EdgeFilter(edge) is true and returns the first node
 // for which until(node, depth) is true. During the traversal, if the Visit field is
-// non-nil, it is called with the nodes joined by each followed edge.
+// non-nil, it is called with each node the first time it is visited.
 func (b *BreadthFirst) Walk(g Graph, from graph.Node, until func(n graph.Node, d int) bool) graph.Node {
 	if b.visited == nil {
 		b.visited = make(set.Int64s)
 	}
 	b.queue.Enqueue(from)
+	if b.Visit != nil && !b.visited.Has(from.ID()) {
+		b.Visit(from)
+	}
 	b.visited.Add(from.ID())
 
 	var (
@@ -67,7 +79,7 @@ func (b *BreadthFirst) Walk(g Graph, from graph.Node, until func(n graph.Node, d
 				continue
 			}
 			if b.Visit != nil {
-				b.Visit(t, n)
+				b.Visit(n)
 			}
 			b.visited.Add(nid)
 			children++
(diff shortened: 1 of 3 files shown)


kortschak:
While I'm breaking the API, I will change the `EdgeFilter` field to `Traverse`. This name is less restrictive in it's connotation; with the filter name, there is a suggestion that it is purely for filtering the edges during the walk. With `Traverse`, it can mean either "Should I traverse to the other side?" or "Traverse this edge."

/cc @umarcor because this will break his code.

umarcor:
Thanks for warning me.

A little suggestion: since you are changing the signature so that `Visit` provides a `Node` and `Traverse` provides an `Edge`, why not make it more explicit by naming them `OnNode` and `OnEdge` instead? I think that it avoids the ambiguity of 'filter', 'traverse', 'visit', etc. The user just does whatever **on each** whatever, during the walk.

kortschak:
`Visit` is a Go convention (see https://golang.org/pkg/go/ast/#Walk and https://golang.org/pkg/path/filepath/#example_Walk) and also within graph terminology. The documentation now makes it clear what each does, and I'm not convinced that `On<Object>` indicates any more clearly what is done.

graph/traverse/traverse.go
```diff
@@ -27,22 +27,34 @@ type Graph interface {
...
-	queue      linear.NodeQueue
-	visited    set.Int64s
+	// Visit is called on all nodes on their first visit.
+	Visit func(graph.Node)
+
+	// Traverse is called on all edges that may be traversed
+	// during the walk. This includes edges that would hop to
+	// an already visited node.
+	//
+	// The value returned by EdgeFulter determines whether an
```
umarcor: Typo. `EdgeFulter` should be `EdgeFilter` and now `Traverse`.
```suggestion
	// The value returned by Traverse determines whether an
```

graph/traverse/traverse.go
```diff
@@ -27,22 +27,34 @@ type Graph interface {
...
+	// edge can be traverse during the walk.
```
umarcor:
```suggestion
	// edge can be traversed during the walk.
```

graph/traverse/traverse.go
```diff
@@ -123,22 +135,34 @@ func (b *BreadthFirst) Reset() {
...
-	stack      linear.NodeStack
-	visited    set.Int64s
+	// Visit is called on all nodes on their first visit.
+	Visit func(graph.Node)
+
+	// Traverse is called on all edges that may be traversed
+	// during the walk. This includes edges that would hop to
+	// an already visited node.
+	//
+	// The value returned by EdgeFulter determines whether an
```
umarcor:
```suggestion
	// The value returned by Traverse determines whether an
```

graph/traverse/traverse.go
```diff
@@ -123,22 +135,34 @@ func (b *BreadthFirst) Reset() {
...
+	// edge can be traverse during the walk.
```
umarcor:
```suggestion
	// edge can be traversed during the walk.
```

graph/traverse/traverse.go
kortschak: Thanks. Fixed.

graph/traverse/traverse.go
kortschak: Fixed

graph/traverse/traverse.go
kortschak: Fixed

graph/traverse/traverse.go
kortschak: Fixed

kortschak:
*ping*

PR #917, opened 2019-03-23
unit: add Torque type

### LINKED ISSUE #915: unit: add type for torque
We should probably add a unit for torque.

However, I don't see a nice way to do this. In order to prevent the length dim in Nm from interacting with `LengthDim` we would need to make a new length dimension (`OrthoLengthDim`). This would have the same symbol (probably — or maybe something else, "₥"?) which would make things impossible.

Thoughts?

/cc @btracey

kortschak:
Please take a look.

Fixes #915.


unit/generate_unit.go
```diff
@@ -480,6 +481,23 @@ var Units = []Unit{
...
 		ErForm: "Pressurer",
 	},
+	{
+		Name:        "Torque",
+		Receiver:    "t",
+		PrintString: "N m",
+		Suffix:      "newtonmeter",
+		Singular:    "Newtonmeter",
+		TypeComment: "Torque represents a torque in Newton meters",
+		Bug: `Torque is measured in N·m, which is dimensionally identical to energy.
```
btracey: I still think this is a bit overblown, but there's no harm in having a longer comment on the type warning people about use. This is not a bug though.

kortschak: It's not a bug in the code here, it is a bug in the system. See https://golang.org/pkg/encoding/xml/#pkg-note-BUG which is also not a bug in the same way this isn't.

kortschak:
PTAL

kortschak:
*ping*

unit/torque.go
```diff
@@ -0,0 +1,94 @@
...
+// Unit converts the Torque to a *Unit
+func (t Torque) Unit() *Unit {
+	return New(float64(t), Dimensions{
+		LengthDim: 2,
+		MassDim:   1,
+		TimeDim:   -2,
+	})
+}
+
+// Torque allows Torque to implement a Torqueer interface
```
btracey: This isn't the doc we want, but this fix can happen in a different PR (it may be elsewhere, I haven't checked).

unit/generate_unit.go
```diff
@@ -480,6 +480,19 @@ var Units = []Unit{
...
+		Receiver:    "t",
+		PrintString: "N m",
+		Suffix:      "newtonmeter",
+		Singular:    "Newtonmeter",
+		TypeComment: "Torque represents a torque in Newton meters",
+		Dimensions: []Dimension{
+			{Name: LengthName, Power: 2},
+			{Name: MassName, Power: 1},
+			{Name: TimeName, Power: -2},
+		},
```
btracey: Wait, I lied, you can use the `ErForm` here to fix the `Torqueer` issue.

unit/torque.go
kortschak: Fixed.

PR #922, opened 2019-03-25
mat: Rename Solve(Vec) to Solve(Vec)To

### LINKED ISSUE #830: mat: Rename SolveVec to SolveVecTo
A number of our factorization types have a `SolveVec` method, but this should be `SolveVecTo` to be in line with `VTo`, etc.

btracey:
Fix #830.

Please take a look.

### DIFF
--- a/mat/cholesky.go
+++ b/mat/cholesky.go
@@ -162,7 +162,7 @@ func (c *Cholesky) LogDet() float64 {
 	return det
 }
 
-// Solve finds the matrix X that solves A * X = B where A is represented
+// SolveTo finds the matrix X that solves A * X = B where A is represented
 // by the Cholesky decomposition. The result is stored in-place into dst.
 func (c *Cholesky) SolveTo(dst *Dense, b Matrix) error {
 	if !c.valid() {
@@ -185,7 +185,7 @@ func (c *Cholesky) SolveTo(dst *Dense, b Matrix) error {
 	return nil
 }
 
-// SolveChol finds the matrix X that solves A * X = B where A and B are represented
+// SolveCholTo finds the matrix X that solves A * X = B where A and B are represented
 // by their Cholesky decompositions a and b. The result is stored in-place into
 // dst.
 func (a *Cholesky) SolveCholTo(dst *Dense, b *Cholesky) error {
@@ -208,10 +208,10 @@ func (a *Cholesky) SolveCholTo(dst *Dense, b *Cholesky) error {
 	return nil
 }
 
-// SolveVec finds the vector X that solves A * x = b where A is represented
+// SolveVecTo finds the vector X that solves A * x = b where A is represented
 // by the Cholesky decomposition. The result is stored in-place into
 // dst.
-func (c *Cholesky) SolveVec(dst *VecDense, b Vector) error {
+func (c *Cholesky) SolveVecTo(dst *VecDense, b Vector) error {
 	if !c.valid() {
 		panic(badCholesky)
 	}
--- a/mat/cholesky.go
+++ b/mat/cholesky.go
@@ -162,9 +162,9 @@ func (c *Cholesky) LogDet() float64 {
 	return det
 }
 
-// Solve finds the matrix x that solves A * X = B where A is represented
-// by the Cholesky decomposition, placing the result in x.
-func (c *Cholesky) Solve(x *Dense, b Matrix) error {
+// Solve finds the matrix X that solves A * X = B where A is represented
+// by the Cholesky decomposition. The result is stored in-place into dst.
+func (c *Cholesky) SolveTo(dst *Dense, b Matrix) error {
 	if !c.valid() {
 		panic(badCholesky)
 	}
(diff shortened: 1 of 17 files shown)


btracey:
The outstanding question is if we should allow a nil value as `dst` and return `*(Vec)Dense, error`. Thoughts @kortschak @vladimir-ch ? It's more consistent with the rest of the Gonum suite (things like `CovarianceMat` behave that way). We also found that behavior useful in `optimize`, so it seems likely it could be similarly useful here.

btracey:
Actually, interestingly, `QR` apparently allows pointers to zero-valued matrices. 
```
var ans Dense
err := qr.SolveTo(&ans, b)
```

That's a decent approach actually, maybe we should port that to `stat.CovarianceMatrix` and make things consistent? Matrix destinations can be zero-valued to be appropriately sized.

kortschak:
I thought that was how everything was supposed to work. This is part of the reason I did not understand the need to return the value. You can always

```
var d Type
for condition {
    d.Reset()
    f.ResultTo(&d)
    use(&d)
}
```

btracey:
Yes, I probably understood that at one point, but now I do. Should we also be changing the other destination matrices, i.e. in stat?

kortschak:
I think consistency is a good thing, so yes, that would be reasonable.

mat/cholesky.go
```diff
@@ -174,20 +174,21 @@ func (c *Cholesky) Solve(x *Dense, b Matrix) error {
...
 	if c.cond > ConditionTolerance {
 		return Condition(c.cond)
 	}
 	return nil
 }
 
-// SolveChol finds the matrix x that solves A * X = B where A and B are represented
-// by their Cholesky decompositions a and b, placing the result in x.
-func (a *Cholesky) SolveChol(x *Dense, b *Cholesky) error {
+// SolveChol finds the matrix X that solves A * X = B where A and B are represented
```
kortschak: `SolveCholTo`

mat/cholesky.go
```diff
@@ -162,9 +162,9 @@ func (c *Cholesky) LogDet() float64 {
 	return det
 }
 
-// Solve finds the matrix x that solves A * X = B where A is represented
-// by the Cholesky decomposition, placing the result in x.
-func (c *Cholesky) Solve(x *Dense, b Matrix) error {
+// Solve finds the matrix X that solves A * X = B where A is represented
```
kortschak: `SolveTo`

mat/cholesky.go
```diff
@@ -196,20 +197,21 @@ func (a *Cholesky) SolveChol(x *Dense, b *Cholesky) error {
...
 	return nil
 }
 
-// SolveVec finds the vector x that solves A * x = b where A is represented
-// by the Cholesky decomposition, placing the result in x.
-func (c *Cholesky) SolveVec(x *VecDense, b Vector) error {
+// SolveVec finds the vector X that solves A * x = b where A is represented
+// by the Cholesky decomposition. The result is stored in-place into
+// dst.
+func (c *Cholesky) SolveVec(dst *VecDense, b Vector) error {
```
kortschak: `SolveVecTo`? and above?

mat/lq.go
```diff
@@ -148,8 +148,8 @@ func (lq *LQ) QTo(dst *Dense) *Dense {
 // The minimization problem solved depends on the input parameters.
```
kortschak: `SolveTo` above.

mat/lq.go
```diff
@@ -209,7 +209,7 @@ func (lq *LQ) Solve(x *Dense, trans bool, b Matrix) error {
 
 // SolveVec finds a minimum-norm solution to a system of linear equations.
 // See LQ.Solve for the full documentation.
-func (lq *LQ) SolveVec(x *VecDense, trans bool, b Vector) error {
+func (lq *LQ) SolveVec(dst *VecDense, trans bool, b Vector) error {
```
kortschak: `SolveVecTo`? and above?

mat/lu.go
```diff
@@ -286,11 +286,11 @@ func (m *Dense) Permutation(r int, swaps []int) {
 //  A * X = B if trans == false
```
kortschak: `SolveTo` above.

mat/lu.go
```diff
@@ -329,41 +329,41 @@ func (lu *LU) Solve(x *Dense, trans bool, b Matrix) error {
 //  A * x = b if trans == false
```
kortschak: `SolveVecTo` above.

mat/qr.go
```diff
@@ -144,8 +144,8 @@ func (qr *QR) QTo(dst *Dense) *Dense {
 // The minimization problem solved depends on the input parameters.
```
kortschak: `SolveTo` above

mat/qr.go
```diff
@@ -206,7 +206,7 @@ func (qr *QR) Solve(x *Dense, trans bool, b Matrix) error {
 // SolveVec finds a minimum-norm solution to a system of linear equations,
```
kortschak: `SolveVecTo`

mat/cholesky.go

mat/cholesky.go

mat/cholesky.go

mat/lq.go
btracey: Done

mat/lq.go
btracey: Done

mat/lu.go
btracey: Done

mat/lu.go
btracey: Done

mat/qr.go
btracey: Done

mat/cholesky.go
```diff
@@ -174,20 +174,21 @@ func (c *Cholesky) Solve(x *Dense, b Matrix) error {
...
 	return nil
 }
 
-// SolveChol finds the matrix x that solves A * X = B where A and B are represented
-// by their Cholesky decompositions a and b, placing the result in x.
-func (a *Cholesky) SolveChol(x *Dense, b *Cholesky) error {
+// SolveCholTo finds the matrix X that solves A * X = B where A and B are represented
+// by their Cholesky decompositions a and b. The result is stored in-place into
+// dst.
+func (a *Cholesky) SolveCholTo(dst *Dense, b *Cholesky) error {
```
kortschak: `golint` doesn't like this. We can deal with that in another PR.

PR #924, opened 2019-03-26
blas/gonum: generate complex64 implementation

vladimir-ch:
Please take a look (per commit).


vladimir-ch:
What do you envision for the tests? Write them manually or try to generate them from `float64/complex128` tests?

kortschak:
I was hoping to do it mechanically. If we square root the tolerances we probably have a pretty good approximation of what will work. I did this for some testing of OpenBLAS a while back and it was OK. The problem is how much stuff there is.

PR #927, opened 2019-03-27
mat: conform receiver name to convention

kortschak:
Please take a look.


kortschak:
I left the other one because I'm not sure what to do with it. This one was obvious.

PR #946, opened 2019-04-02
Add fast paths to matrix sum function

jonreiter:
Add fast paths to matrix sum function for triangular, symmetric and vector.

There was a TODO in there and I call this a lot on symmetric matrices.

As this is partially a performance-related change I added benchmarks for sum.
The new version is faster for triangular and symmetric -- but as the
benchmark is new I'm not sure a clean way to benchstat this without two
pull requests.

Also add a few function comments to triangular.go to appease golint.

Please take a look.

### DIFF
--- a/lapack/testlapack/dgesvd.go
+++ b/lapack/testlapack/dgesvd.go
@@ -164,7 +164,7 @@ func dgesvdTest(t *testing.T, impl Dgesvder, m, n, mtype int, tol float64) {
 
 		// Check that uAll, sAll, and vtAll multiply back to A by computing a residual
 		//  |A - U*S*VT| / (n*aNorm)
-		if resid := svdFullResidual(m, n, aNorm, aCopy, lda, uAll, ldu, sAll, vtAll, ldvt); resid > 1e-13 {
+		if resid := svdFullResidual(m, n, aNorm, aCopy, lda, uAll, ldu, sAll, vtAll, ldvt); resid > tol {
 			t.Errorf("Case %v: original matrix not recovered for full SVD, |A - U*D*VT|=%v", prefix, resid)
 		}
 		if minmn > 0 {
@@ -238,14 +238,14 @@ func dgesvdTest(t *testing.T, impl Dgesvder, m, n, mtype int, tol float64) {
 					if !hasOrthonormalColumns(blas64.General{Rows: m, Cols: minmn, Data: u, Stride: ldu}) {
 						t.Errorf("Case %v: columns of U are not orthogonal", prefix)
 					}
-					if res := svdPartialUResidual(m, minmn, u, uAll, ldu); res > 1e-13 {
+					if res := svdPartialUResidual(m, minmn, u, uAll, ldu); res > tol {
 						t.Errorf("Case %v: columns of U do not match UAll", prefix)
 					}
 				case lapack.SVDAll:
 					if !hasOrthonormalColumns(blas64.General{Rows: m, Cols: m, Data: u, Stride: ldu}) {
 						t.Errorf("Case %v: columns of U are not orthogonal", prefix)
 					}
-					if res := svdPartialUResidual(m, m, u, uAll, ldu); res > 1e-13 {
+					if res := svdPartialUResidual(m, m, u, uAll, ldu); res > tol {
 						t.Errorf("Case %v: columns of U do not match UAll", prefix)
 					}
 				}
@@ -255,14 +255,14 @@ func dgesvdTest(t *testing.T, impl Dgesvder, m, n, mtype int, tol float64) {
 					if !hasOrthonormalRows(blas64.General{Rows: minmn, Cols: n, Data: vtAll, Stride: ldvt}) {
 						t.Errorf("Case %v: rows of VT are not orthogonal", prefix)
 					}
-					if res := svdPartialVTResidual(minmn, n, vt, vtAll, ldvt); res > 1e-13 {
+					if res := svdPartialVTResidual(minmn, n, vt, vtAll, ldvt); res > tol {
 						t.Errorf("Case %v: rows of VT do not match VTAll", prefix)
 					}
 				case lapack.SVDAll:
 					if !hasOrthonormalRows(blas64.General{Rows: n, Cols: n, Data: vtAll, Stride: ldvt}) {
 						t.Errorf("Case %v: rows of VT are not orthogonal", prefix)
 					}
-					if res := svdPartialVTResidual(n, n, vt, vtAll, ldvt); res > 1e-13 {
+					if res := svdPartialVTResidual(n, n, vt, vtAll, ldvt); res > tol {
 						t.Errorf("Case %v: rows of VT do not match VTAll", prefix)
 					}
 				}
@@ -274,7 +274,7 @@ func dgesvdTest(t *testing.T, impl Dgesvder, m, n, mtype int, tol float64) {
 				if floats.Min(s) < 0 {
 					t.Errorf("Case %v: some singular values from full SVD are negative", prefix)
 				}
-				if !floats.EqualApprox(s, sAll, 1e-14) {
+				if !floats.EqualApprox(s, sAll, tol/10) {
 					t.Errorf("Case %v: singular values differ between full and partial SVD\n%v\n%v", prefix, s, sAll)
 				}
 			}


mat/dense_test.go
```diff
@@ -2093,3 +2093,14 @@ func denseMulTransSymBench(b *testing.B, size int, rho float64) {
 		wd = &n
 	}
 }
+
+func BenchmarkDenseSum1000(b *testing.B) { denseSumBench(b, 1000, 100) }
```
kortschak: This looks like a job for sub-benchmarks.

mat/dense_test.go
```diff
@@ -2093,3 +2093,14 @@ func denseMulTransSymBench(b *testing.B, size int, rho float64) {
...
+
+func denseSumBench(b *testing.B, size, iters int) {
+	b.StopTimer()
+	a, _ := randDense(size, 1.0, rand.NormFloat64)
+	b.StartTimer()
+	for i := 0; i < iters; i++ {
```
kortschak: This is not correct. The form of a benchmark loop is `for i := 0; i < b.N; i++ {`

mat/dense_test.go
```diff
@@ -2093,3 +2093,14 @@ func denseMulTransSymBench(b *testing.B, size int, rho float64) {
...
+		_ = Sum(a)
```
kortschak: This should be assigned to a value that is used or a global.

mat/matrix.go
```diff
@@ -848,20 +848,57 @@ func normLapack(norm float64, aTrans bool) lapack.MatrixNorm {
...
 	aU, _ := untranspose(a)
-	if rma, ok := aU.(RawMatrixer); ok {
+	switch rma := aU.(type) {
+	case RawSymmetricer:
+		rm := rma.RawSymmetric()
+		for i := 0; i < rm.N; i++ {
+			// Diagonals count once while off-diagonals count twice.
+			sum += rm.Data[i*rm.Stride+i]
+			for _, v := range rm.Data[i*rm.Stride+i+1 : i*rm.Stride+rm.N] {
+				sum += 2 * v
```
kortschak: Do a single doubling at the end.

mat/matrix.go
```diff
@@ -848,20 +848,57 @@ func normLapack(norm float64, aTrans bool) lapack.MatrixNorm {
...
+		return sum
+	case RawMatrixer:
 		rm := rma.RawMatrix()
 		for i := 0; i < rm.Rows; i++ {
 			for _, v := range rm.Data[i*rm.Stride : i*rm.Stride+rm.Cols] {
 				sum += v
 			}
 		}
 		return sum
+	case *VecDense:
```
kortschak: This isn't mentioned in the commit message. Please do.

mat/matrix.go
```diff
@@ -848,20 +848,57 @@ func normLapack(norm float64, aTrans bool) lapack.MatrixNorm {
...
+		rm := rma.RawVector()
+		for i := 0; i < rm.N; i++ {
+			sum += rm.Data[i*rm.Inc]
+		}
+		return sum
 	}
+	// Finally the generic case.
```
kortschak: Maybe rather than this comment, just make a `default:` case in the switch.

mat/symmetric_test.go
```diff
@@ -735,3 +735,28 @@ func TestPowPSD(t *testing.T) {
 		}
 	}
 }
+
+func BenchmarkSymSum1000(b *testing.B) { symSumBench(b, 1000, 100) }
```
kortschak: See previous comments about sub-benchmarks and benchmark iterations.

mat/triangular.go
```diff
@@ -105,6 +105,7 @@ func (t TransposeTri) Untranspose() Matrix {
 	return t.Triangular
 }
 
+// UntransposeTri and return specifically a Triangular object.
```
kortschak: These documentation changes are incidental and should be in a separate PR.

mat/triangular_test.go
```diff
@@ -512,3 +512,25 @@ func TestCopySymIntoTriangle(t *testing.T) {
 		}
 	}
 }
+
+func BenchmarkTriSum1000(b *testing.B) { triSumBench(b, 1000, 100) }
```
kortschak: See previous benchmark comments.

mat/dense_test.go
sbinet: one should probably store the result in a global variable, to prevent the clever optimizer from doing its job.

mat/dense_test.go
sbinet: shouldn't this be: `s/iters/b.N/` instead?

mat/dense_test.go
sbinet: ie:
```go
var denseSumBenchResult float64

func denseSumBench(b *testing.B, size int) {
    a, _ := randDense(size, 1.0, rand.NormFloat64)
    b.ResetTimer()
    for i := 0; i < b.N; i++ {
        denseSumBenchResult = Sum(a)
    }
}
```

mat/symmetric_test.go
```diff
@@ -735,3 +735,28 @@ func TestPowPSD(t *testing.T) {
...
+
+func symSumBench(b *testing.B, size, iters int) {
+	b.StopTimer()
+	a := randSymDense(size)
+	b.StartTimer()
+	for i := 0; i < iters; i++ {
```
sbinet: ditto. (`b.N` + global variable)

mat/triangular_test.go
```diff
@@ -512,3 +512,25 @@ func TestCopySymIntoTriangle(t *testing.T) {
...
+
+func triSumBench(b *testing.B, size, iters int) {
+	b.StopTimer()
+	a := randTriDense(size)
+	b.StartTimer()
+	for i := 0; i < iters; i++ {
```
sbinet: ditto.

jonreiter:
benchstat:

```
name                old time/op  new time/op   delta
VectorSum100000-4   515µs ± 1%   106µs ± 1%   -79.44%  (p=0.029 n=4+4)
TriSum1000-4        5.55ms ± 1%  0.61ms ± 1%  -88.96%  (p=0.029 n=4+4)
SymSum1000-4        6.71ms ± 9%  0.61ms ± 1%  -90.85%  (p=0.029 n=4+4)
DenseSum1000-4      1.09ms ± 0%  1.09ms ± 1%     ~     (p=0.343 n=4+4)

```

mat/triangular_test.go
```diff
@@ -512,3 +512,27 @@ func TestCopySymIntoTriangle(t *testing.T) {
...
+	a := randTriDense(size)
+	b.StartTimer()
+	for i := 0; i < b.N; i++ {
+		triSumForBench = Sum(a)
+	}
+}
+
+func randTriDense(size int) *TriDense {
+	t := NewTriDense(size, Upper, nil)
+	return t
```
sbinet: debug statement leftover?

mat/vector_test.go
```diff
@@ -561,3 +561,16 @@ func randVecDense(size, inc int, rho float64, rnd func() float64) *VecDense {
...
 }
+
+func BenchmarkVectorSum100000(b *testing.B) { vectorSumBench(b, 100000) }
+
+var vectorSumForBench float64
+
+func vectorSumBench(b *testing.B, size int) {
+	b.StopTimer()
+	a := randVecDense(size, 1, 1.0, rand.NormFloat64)
+	b.StartTimer()
```
sbinet: I still find using `b.RestartTimer()` easier (and dropping `b.StopTimer()` entirely) but ok.

mat/dense_test.go
jonreiter: i think i've addressed all the benchmark issues.

mat/matrix.go
jonreiter: now breaks out a single doubling and does the diagonal.  neither approach looks totally clean to me honestly.

mat/triangular.go
jonreiter: ive just removed them.

mat/triangular_test.go
jonreiter: indeed. sorted now. thanks spot.

mat/matrix.go
```diff
@@ -848,26 +848,65 @@ func normLapack(norm float64, aTrans bool) lapack.MatrixNorm {
...
 			}
 		}
 		return sum
-	}
-	for i := 0; i < r; i++ {
-		for j := 0; j < c; j++ {
-			sum += a.At(i, j)
+	case *VecDense:
+		rm := rma.RawVector()
+		for i := 0; i < rm.N; i++ {
```
kortschak: This is not correct. `rm.Inc` is not guaranteed to be 1.

mat/matrix.go
```diff
@@ -848,26 +848,65 @@ func normLapack(norm float64, aTrans bool) lapack.MatrixNorm {
...
 	var sum float64
 	aU, _ := untranspose(a)
-	if rma, ok := aU.(RawMatrixer); ok {
+	switch rma := aU.(type) {
+	case RawSymmetricer:
+		rm := rma.RawSymmetric()
+		for i := 0; i < rm.N; i++ {
+			// Diagonals count once while off-diagonals count twice.
+			sum += rm.Data[i*rm.Stride+i]
+			subSum := 0.0
```
kortschak: `var subSum float64` (though I'd just call this `s`).

mat/matrix.go
jonreiter: isn't rm.Inc the spacing between elements per http://www.netlib.org/lapack/explore-html/de/d05/dasum_8f_source.html (for example)?

jonreiter: i think i have rebased etc now per your request (i have never done that before so hopefully correct).

kortschak:
What I was suggest was the following:

Start an interactive rebase with 8 commits.
```
$ git rebase -i HEAD~8 # 8 is the number of commits in the PR plus one because we are modifying the first one.
```


This will start an editor (whichever editor is you default - usually vim) that will show you this (with some additional commentary/instructions below it thanks to `git`)
```
pick 3d26580 lapack/testlapack: replace all literal constants in dgesvdTest with tol expressions
pick 9f02923 Add fast paths to matrix sum function for triangular, symmetric and vector. There was a TODO in there and I call this a lot on symmetric matrices.
pick 3f5a65c Revert "Add fast paths to matrix sum function for triangular, symmetric and vector."
pick 3208802 mat: add benchmarks for Sum with Dense, SymDense, TriDense and VecDense
pick 80d1b4c mat: add fast paths for Sum with SymDense, TriDense and VecDense
pick 32192d5 A+C: add Jonathan Reiter
pick 20ad6d0 remove mistaken debug statement
pick f35ed94 small cleanups
```

You then change it to look like this (the commented stuff below it is not impotrant)
```
pick 3d26580 lapack/testlapack: replace all literal constants in dgesvdTest with tol expressions


pick 3208802 mat: add benchmarks for Sum with Dense, SymDense, TriDense and VecDense
squash 20ad6d0 remove mistaken debug statement
pick 80d1b4c mat: add fast paths for Sum with SymDense, TriDense and VecDense
pick 32192d5 A+C: add Jonathan Reiter
pick f35ed94 small cleanups
```
and then write out the file and close it.

It will come back with a file buffer that has this text.
```
# This is a combination of 2 commits.
# The first commit's message is:

mat: add benchmarks for Sum with Dense, SymDense, TriDense and VecDense

# This is the 2nd commit message:

remove mistaken debug statement
```

Edit it to look like thie
```
mat: add benchmarks for Sum with Dense, SymDense, TriDense and VecDense
```

and then writ it out and close it.

This will leave you with a new history which looks like this:
```
d0ac098 small cleanups
c928e9b A+C: add Jonathan Reiter
9045fee mat: add fast paths for Sum with SymDense, TriDense and VecDense
c920e60 mat: add benchmarks for Sum with Dense, SymDense, TriDense and VecDense
```

(with different shas).

Ideally, I would like the "cleanups" commit to be squashed as wel, but it has changes corresponding to two different commits (the benchmarks and the substantive changes).
This can be fixed though; if you `git reset --mixed HEAD~1` after the procedure above, you'll have that commit back as a set of changes that you can merge as two separate commits, and then do the same kind of interactive rebase as above, squashing each to the appropriate place.

If anything goes wrong during this, you can always just `git reset --hard f35ed94` to take it back to the state it's at now.

When it's all good (you can check by doing `git diff master origin/master` — assuming your remotes is origin — and confirm no difference), just `git push -f origin master`.

mat/matrix.go
kortschak: No, but this has fewer multiplies.

jonreiter:
I think correct now.  Reread a few times...again sorry if still not quite there.

jonreiter:
@kortschak 

I think correct now. Reread a few times...again sorry if still not quite there.

mat/matrix.go
kortschak: Sorry, you are right.

kortschak:
This is almost there, but I don't see an easy rebase to get to where we want to be (it's possible, but it's much easier to do a mixed reset to the commit before you started and do the commits again). We want changes to the benchmarks in the first commit, changes to the `Sum` code in the second (no changes to the benchmarks here) and then your addition to A+C in the third. The final state is OK.

Thank you for bearing with me on this.

PR #950, opened 2019-04-09
unit: remove explicitly prefixed constants

### LINKED ISSUE #90: unit: clean up unit package

kortschak:
Please take a look.

Closes #90.


unit/generate_unit.go
```diff
@@ -18,16 +18,15 @@ import (
 )
 
 type Unit struct {
-	Name          string
+	DimensionName string
 	Receiver      string
-	Offset        int    // From normal (for example, mass base unit is kg, not kg)
+	PowerOffset   int    // From normal (for example, mass base unit is kg, not kg)
```
vladimir-ch: is kg, not kg?

unit/generate_unit.go
```diff
@@ -262,54 +169,51 @@ var Units = []Unit{
...
 	},
 	{
-		Name:        "Area",
-		Receiver:    "a",
-		PrintString: "m^2",
-		TypeComment: "Area represents and area in square metres",
+		DimensionName: "Area",
+		Receiver:      "a",
+		PrintString:   "m^2",
+		TypeComment:   "Area represents and area in square metres",
```
vladimir-ch: an area

unit/generate_unit.go
kortschak: Ha!

Fixed

unit/generate_unit.go
kortschak: Fixed

kortschak:
PTAL

kortschak:
I'll wait.

btracey:
That's such an obvious solution to the Kg problem ... can't believe I didn't think of that.

kortschak:
Which part, the extra constant or the 1e-3?

btracey:
Setting the gram to be 1e-3.

kortschak:
When you remove all the constants, that's what was left. It seemed obvious.

btracey:
I agree, it does seem obvious

PR #951, opened 2019-04-10
optimize/convex/lp: refactoring and affine-scaling solver

jonreiter:
this has a (mostly) working basic-but-rugged affine-scaling solver.

to borrow the existing simplex tests required some cleanup which addresses some TODOs in that code.  and similarly raises design issues in this package.

i don't think this code is near done yet.  but am curious thoughts on how this should all be organized.  for example simplex cares only about basic while affine only needs feasible.  so there is a simplexPresolve() wrapper returning a basicidx + values to share called by both solvers now.  it's a mild presolve.

convert could be more general as well.  i'd like it to transform an initial feasible for the affine solver alongside converting the problem.  using 0s is fine when simplex is bootstraping from the presolve but doesn't work when you have a non-basic starting point and need to convert.  there is a new wrapper for convert-and-solve with simplex which is straightforward but doesn't extend to affine easily.  there are a few ways to wrangle that.

i could put through the simplex/convert changes on their own as they address some TODOs.  but i suspect the straightforward approach isn't the long term solution here.

Please take a look.


btracey:
Hi, thanks for the PRs and the comments on the issues. I've been busy lately and haven't had a chance to look into it, but I want to acknowledge I'm listening.

jonreiter:
thanks and understood.  am slowly groping my way towards understanding the codebase through all of this...

jonreiter:
any further thoughts on how to organize this?  i am just getting back to this work now and see a fair amount of discussion on organizing the optimize package.

at a minimum i'd like to:
- do a convert-and-solve wrapper, which also addresses a TODO
- split up the simplex code to pull out the code which finds an initial basic
so it is clean to write other methods (which i'd then do).

i think even without a settled higher level interface this would be useful in the lp space.  i know it would be for me.

vladimir-ch:

Issue #939, opened 2019-03-30
bound: Helper issues for bound


btracey:
See:
https://github.com/gonum/gonum/pull/845#discussion_r255359176

corona10:
@btracey Can I take a look this issue? cc @sbinet

btracey:
Please do!

corona10:
@btracey 

Thanks, I'd like to try to work on creating a new helper API for this package.
Since I am not familiar with the use case of this helper API there's some question for this issue.

For example, when we implement intersection helpers for this package.
What API we are going to provide `Intersection([]bound)` or `Intersection(b1 ,b2 bound)`?

Or is this issue is an open status to provide a helper?
So do I have to find some use case of this package?
For example, I can find some use case of `boundMax-bound.Min` so we can provide `Distance API` for this use case.

Thanks for the guide :)

btracey:
I think `Intersection` and `Union` might be best as a variadic input actually, `Intersection(b ...Bound)`

corona10:
@btracey @kortschak

Next step is implementing `Union`
I 'd like to propose the signature of `Union` to be `func Union(b ...Bound) []Bound`.
What do you think?

btracey:
I think it should match Intersection, with Union(b …Bound) Bound.

If there’s a vector version you think is useful, feel free to file a feature request issue

vladimir-ch:
What would be the result of `Union([0,1],[2,3])` with a single returned Bound?

kortschak:
I'd say that's not possible here.

corona10:
@vladimir-ch 
My expected result was `[[0,1], [2,3]]` but according to @kortschak this input is not possible usage of gonum usage right?

btracey:
My intent (and I believe @kortschak ) is to make some standard things easy. Intersection is basically "We have a bunch of bounds on the same variable, and the constraint is that all of them can be met". This would come up in an optimization context. For union it is "Here are a bunch of semi-overlapping things, bring them all together". In that case, since it can't be represented by a single Bound, it should return the NaN bound.

This isn't to say that other functions aren't useful, just that Intersection and Union will likely be used commonly, and will work with other packages (`distmv.Uniform` and eventually `optimize. Problem`)

kortschak:
[...]

PR #954, opened 2019-04-13
bound: Add Intersection helper for bound package

corona10:
Add Intersection helper API for bound package

Updates: #939

### DIFF
--- a/bound/bound.go
+++ b/bound/bound.go
@@ -4,7 +4,35 @@
 
 package bound
 
+import "math"
+
 // Bound represents [Min, Max] bounds.
 type Bound struct {
 	Min, Max float64
+}
+
+// IsValid returns whether the bound is valid
+func (b Bound) IsValid() bool {
+	return b.Min <= b.Max
+}
+
+// Intersection returns a Bound that is the intersection of the input bounds.
+// If the intersection is empty or if the input length is zero,
+// then the NaN Bound will be returned.
+func Intersection(bounds ...Bound) Bound {
+	if len(bounds) == 0 {
+		return Bound{Min: math.NaN(), Max: math.NaN()}
+	}
+
+	ret := Bound{Min: bounds[0].Min, Max: bounds[0].Max}
+	for _, b := range bounds[1:] {
+		ret.Min = math.Max(ret.Min, b.Min)
+		ret.Max = math.Min(ret.Max, b.Max)
+	}
+
+	if !ret.IsValid() {
+		return Bound{Min: math.NaN(), Max: math.NaN()}
+	}
+
+	return ret
 }
--- a/bound/bound_test.go
+++ b/bound/bound_test.go
@@ -1,0 +1,77 @@
+// Copyright ©2019 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package bound
+
+import (
+	"math"
+	"testing"
+)
+
+func TestIntersection(t *testing.T) {
+	b0 := Bound{Min: 0.0, Max: 5.0}
+	b1 := Bound{Min: 3.0, Max: 4.0}
+	b2 := Bound{Min: -1.0, Max: 1.0}
+	b3 := Bound{Min: 6.0, Max: 8.0}
+
+	ret := Intersection(b0)
+	expected := Bound{Min: 0.0, Max: 5.0}
+	if ret != expected {
+		t.Errorf("Intersection(b0) should be %v but got %v", expected, ret)
+	}
+
+	ret = Intersection(b0, b1)
+	expected = Bound{Min: 3.0, Max: 4.0}
+	if ret != expected {
+		t.Errorf("Intersection(b0, b1) should be %v but got %v", expected, ret)
+	}
+
+	ret = Intersection(b0, b2)
+	expected = Bound{Min: 0.0, Max: 1.0}
+	if ret != expected {
+		t.Errorf("Intersection(b0, b2) should be %v but got %v", expected, ret)
+	}
+
+	ret = Intersection(b0, b3)
+	if ret.IsValid() {
+		t.Error("Intersection(b0, b3) should not be valid")
+	}
+
+	ret = Intersection(b0, b1, b2)
+	if ret.IsValid() {
+		t.Error("Intersection(b0, b1, b2) should not be valid")
+	}
+
+	ret = Intersection()
+	if ret.IsValid() {
+		t.Error("Intersection() with zero input length should not be valid")
+	}
+}
+
+func TestIsValid(t *testing.T) {
+	b0 := Bound{Min: 0.0, Max: 5.0}
+	if !b0.IsValid() {
+		t.Error("b0 is valid")
+	}
+
+	b1 := Bound{Min: 5.0, Max: 0.0}
+	if b1.IsValid() {
+		t.Error("b1 is invalid")
+	}
+
+	b2 := Bound{Min: math.NaN(), Max: 5.0}
+	if b2.IsValid() {
+		t.Error("b2 is invalid")
+	}
+
+	b3 := Bound{Min: 5.0, Max: math.NaN()}
+	if b3.IsValid() {
+		t.Error("b3 is invalid")
+	}
+
+	b4 := Bound{Min: math.NaN(), Max: math.NaN()}
+	if b4.IsValid() {
+		t.Error("b4 is invalid")
+	}
+}
--- a/AUTHORS
+++ b/AUTHORS
@@ -24,6 +24,7 @@ David Samborski <bloggingarrow@gmail.com>
 Davor Kapsa <davor.kapsa@gmail.com>
 DeepMind Technologies
 Dezmond Goff <goff.dezmond@gmail.com>
+Dong-hee Na <donghee.na92@gmail.com>
 Egon Elbre <egonelbre@gmail.com>
 Ekaterina Efimova <katerina.efimova@gmail.com>
 Ethan Burns <burns.ethan@gmail.com>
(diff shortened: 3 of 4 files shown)


bound/bound.go
```diff
@@ -4,7 +4,41 @@
...
+	"sort"
+)
+
 // Bound represents [Min, Max] bounds.
 type Bound struct {
 	Min, Max float64
 }
+
+// Intersection returns intersected bounds between bounds.
+func Intersection(bounds ...Bound) []Bound {
```
btracey: I think the idea was to just get the single intersection.

`func Intersection(bounds ...Bound) Bound`

Then there's no need to sort or copy. 

You'll also need to add documentation on what happens when len(bound) == 0, where I would say the zero `Bound` should be returned.

bound/bound_test.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+
+	ret = Intersection(b1, b4)
+	expected = []Bound{}
+
+	if !isIdentical(ret, expected) {
+		t.Errorf("Intersection should be %v but got %v", expected, ret)
+	}
+
+	ret = Intersection(b0, b1, b3)
```
corona10: @btracey 
Thanks for the fast review.
If we update the signature of this API into `func Intersection(bounds ...Bound) Bound`
What is the expected result of this API?

bound/bound.go
corona10: If we intend to get the single intersection.
what about the update the signature into `func Intersection(b0, b1Bound) Bound`?
IMHO, variadic input intends multiple bounds so the number of intersected bounds can be more than 1.

WDYT?

bound/bound_test.go
corona10: or did you intend `Intersect(Intersect(b0, b1), b3))`?

btracey: My intended behavior is `Intersect(Intersect(b0, b1), b2))`, but then made easy with the variadic input. This way you can write `Intersection(b0, b1)`, which as you say is the common case, but also could write `Intersection(bounds...)` if you have a list of them.

corona10:
@btracey PTAL!

bound/bound.go
```diff
@@ -4,7 +4,37 @@
...
 package bound
 
+import "math"
+
 // Bound represents [Min, Max] bounds.
 type Bound struct {
 	Min, Max float64
 }
+
+// Intersection returns a intersected bound between bounds.
```
btracey: Intersection returns a Bound that is the intersection of the input bounds. If the intersection is empty or if the input length is zero, then the NaN Bound will be returned.

bound/bound.go
```diff
@@ -4,7 +4,37 @@
...
+	if len(bounds) == 0 {
+		return Bound{Min: math.NaN(), Max: math.NaN()}
+	}
+
+	if len(bounds) == 1 {
+		return bounds[0]
+	}
+
+	ret := Bound{Min: bounds[0].Min, Max: bounds[0].Max}
+	for _, b := range bounds {
```
btracey: This would be cleaner just sweeping over the bounds to do the min and max, with one final check at the end for the bounds matching. It's probably better too, as it'll be faster in the common case of the set actually having a non-empty intersection.

bound/bound_test.go
```diff
@@ -0,0 +1,40 @@
...
+	b1 := Bound{Min: 3.0, Max: 4.0}
+	b2 := Bound{Min: -1.0, Max: 1.0}
+
+	ret := Intersection(b0)
+	expected := Bound{Min: 0.0, Max: 5.0}
+	if ret != expected {
+		t.Errorf("Intersection should be %v but got %v", expected, ret)
+	}
+
+	ret = Intersection(b0, b1)
```
btracey: Also add a test for b0 and b2, where the bounds come from two different ones.

bound/bound.go
```diff
@@ -4,7 +4,37 @@
...
 type Bound struct {
 	Min, Max float64
 }
+
+// Intersection returns a intersected bound between bounds.
+// if len(bounds) == 0 then it will return NaN value bound.
+// Or if there is no intersection bound then it will return Nan value bound.
+func Intersection(bounds ...Bound) Bound {
+	if len(bounds) == 0 {
+		return Bound{Min: math.NaN(), Max: math.NaN()}
```
kortschak: If NaN is being used we should have a helper to test whether the `Bound` is valid.

kortschak: BTW Is there a reason to not just use the zero value? It is an empty set and so expresses what this would.

corona10: @kortschak I would like to distinguish between empty bound and intersected bound which actually has zero value.

I 've updated it to return an empty value.

kortschak: We have a problem here that stems from the use of closed intervals for `Bound`. My comment above is not correct, [x, x] is not an empty set, and since [0, 0] is both in that set of sets, and 0 is a valid number, the zero value for `Bound` can't be used as a semaphore. So this answers my question.

This means that we *do* need to use NaN, but we should also have a way to test validity. Even if only to signal to users that `b1 == b2` may not necessarily be meaningful for all `Bound`s.

kortschak: Names for the helper (I don't have firm favourites): `IsEmpty`, `IsValid`, `IsZero`. Obviously the semantics of `IsEmpty`/`IsZero` and `IsValid` are inverted. (I think I prefer `IsEmpty(b Bound) bool { return math.IsNaN(b.Min) || math.IsNaN(b.Max) }`.

corona10:
@btracey @kortschak 

I 've updated the CL which is reflected reviews.
PTAL

bound/bound.go
```diff
@@ -4,7 +4,30 @@
...
+// Intersection returns a Bound that is the intersection of the input bounds.
+// If the intersection is empty or if the input length is zero,
+// then the zero Bound will be returned.
+func Intersection(bounds ...Bound) Bound {
+	if len(bounds) == 0 {
+		return Bound{}
+	}
+
+	ret := Bound{Min: bounds[0].Min, Max: bounds[0].Max}
+	for _, b := range bounds {
```
kortschak: Since `bounds[0]` has already been considered, this should be `bounds[1:]`.

bound/bound.go
```diff
@@ -4,7 +4,30 @@
...
 type Bound struct {
 	Min, Max float64
 }
+
+// Intersection returns a Bound that is the intersection of the input bounds.
+// If the intersection is empty or if the input length is zero,
+// then the zero Bound will be returned.
+func Intersection(bounds ...Bound) Bound {
+	if len(bounds) == 0 {
+		return Bound{}
```
kortschak: See https://github.com/gonum/gonum/pull/954#discussion_r275136928.

bound/bound.go
```diff
@@ -4,7 +4,30 @@
...
+	}
+
+	ret := Bound{Min: bounds[0].Min, Max: bounds[0].Max}
+	for _, b := range bounds {
+		ret.Min = math.Max(ret.Min, b.Min)
+		ret.Max = math.Min(ret.Max, b.Max)
+	}
+
+	if ret.Min > ret.Max {
+		return Bound{}
```
kortschak: See https://github.com/gonum/gonum/pull/954#discussion_r275136928.

corona10:
@kortschak @btracey

I've updated the CL as you comment
two APIs are added also.

- Bound.IsEmpty()
- Bound.IsValid()

Please take a look.

bound/bound.go
```diff
@@ -11,22 +11,32 @@ type Bound struct {
...
 }
 
+// IsEmpty returns whether the bound is empty
+func (b *Bound) IsEmpty() bool {
+	return math.IsNaN(b.Min) || math.IsNaN(b.Max)
+}
+
+// IsValid returns whether the bound is valid
+func (b *Bound) IsValid() bool {
+	return !math.IsNaN(b.Min) && !math.IsNaN(b.Max) && b.Min <= b.Max
```
kortschak: You don't need the NaN checks here since any equality test except `!=` always returns false if there is at least one NaN.

bound/bound.go
```diff
@@ -4,7 +4,40 @@
...
 package bound
 
+import "math"
+
 // Bound represents [Min, Max] bounds.
 type Bound struct {
 	Min, Max float64
 }
+
+// IsEmpty returns whether the bound is empty
```
btracey: Can we just call this IsNaN? If we have `IsNaN` and `IsValid` it's clear from just the names what they do. With `IsEmpty` and `IsValid` I can't tell the difference from the name alone.

@kortschak ?

kortschak: Nice. `IsNaN` SGTM.

kortschak: Do we need it though? If we have `IsValid`, is that not enough?

btracey: IsValid is probably enough

bound/bound.go
vladimir-ch: Should this be a package-level function?
What about intervals with inverted min max?

bound/bound_test.go
```diff
@@ -0,0 +1,97 @@
...
+		t.Error("Intersection should be empty")
+	}
+
+	ret = Intersection()
+	if !ret.IsEmpty() {
+		t.Error("Intersection should be empty")
+	}
+}
+
+func TestIsMethodOfBound(t *testing.T) {
```
vladimir-ch: What does the test name mean? `TestIsValid`?

bound/bound_test.go
```diff
@@ -0,0 +1,97 @@
...
+// license that can be found in the LICENSE file.
+
+package bound
+
+import (
+	"math"
+	"testing"
+)
+
+func TestBoundInterSect(t *testing.T) {
```
vladimir-ch: `TestIntersection`?

bound/bound.go
vladimir-ch: Ah, sorry, I'm on my phone and I didn't see the end of the line.

btracey: If it stays a method, it doesn't need a pointer receiver

corona10:
@kortschak I 've rebased commits
@btracey I remove IsEmpty method and update receiver of IsValid
@vladimir-ch I 've updated unit test names

Please take a look!

kortschak:
This probably should have been a rebase rather than a squash since it was nicely formatted to have them as separate commits.

btracey:
Oh, shoot, sorry!

PR #955, opened 2019-04-13
diff/fd,optimize: don't allocate matrix on nil input to fd.Hessian and use mat.SymDense in optimize

### LINKED ISSUE #223: optimize,diff/fd: impedance mismatch in Hessian and mat.MutableSymmetric

kortschak:
Please take a look.

Closes  #223.
Updates #593.

### DIFF
--- a/diff/fd/hessian.go
+++ b/diff/fd/hessian.go
@@ -11,30 +11,24 @@ import (
 	"gonum.org/v1/gonum/mat"
 )
 
-// Hessian approximates the Hessian matrix of the multivariate function f
-// at the location x. That is
+// Hessian approximates the Hessian matrix of the multivariate function f at
+// the location x. That is
 //  H_{i,j} = ∂^2 f(x)/∂x_i ∂x_j
-// If dst is not nil, the resulting H will be stored in-place into dst
-// and returned, otherwise a new matrix will be allocated first. Finite difference
-// formula and other options are specified by settings. If settings is nil,
-// the Hessian will be estimated using the Forward formula and a default step size.
+// The resulting H will be stored in dst. Finite difference formula and other
+// options are specified by settings. If settings is nil, the Hessian will be
+// estimated using the Forward formula and a default step size.
 //
-// Hessian panics if the size of dst and x is not equal, or if the derivative
-// order of the formula is not 1.
-func Hessian(dst *mat.SymDense, f func(x []float64) float64, x []float64, settings *Settings) *mat.SymDense {
+// If the dst matrix is zero-sized it will be resized to the correct dimensions,
+// otherwise the dimensions of dst must match the length of x or Hessian will panic.
+// Hessian will panic if the derivative order of the formula is not 1.
+func Hessian(dst *mat.SymDense, f func(x []float64) float64, x []float64, settings *Settings) {
 	n := len(x)
-	if dst == nil {
-		dst = mat.NewSymDense(n, nil)
-	} else {
-		if n2 := dst.Symmetric(); n2 != n {
-			panic("hessian: dst size mismatch")
-		}
-		for i := 0; i < n; i++ {
-			for j := i; j < n; j++ {
-				dst.SetSym(i, j, 0)
-			}
-		}
+	if dst.IsZero() {
+		*dst = *(dst.GrowSym(n).(*mat.SymDense))
+	} else if dst.Symmetric() != n {
+		panic("hessian: dst size mismatch")
 	}
+	dst.Zero()
 
 	// Default settings.
 	formula := Forward
@@ -74,10 +68,9 @@ func Hessian(dst *mat.SymDense, f func(x []float64) float64, x []float64, settin
 	nWorkers := computeWorkers(concurrent, evals)
 	if nWorkers == 1 {
 		hessianSerial(dst, f, x, formula.Stencil, step, originKnown, originValue)
-		return dst
+		return
 	}
 	hessianConcurrent(dst, nWorkers, evals, f, x, formula.Stencil, step, originKnown, originValue)
-	return dst
 }
 
 func hessianSerial(dst *mat.SymDense, f func(x []float64) float64, x []float64, stencil []Point, step float64, originKnown bool, originValue float64) {
--- a/optimize/functions/functions.go
+++ b/optimize/functions/functions.go
@@ -36,12 +36,9 @@ func (Beale) Func(x []float64) float64 {
 	return f1*f1 + f2*f2 + f3*f3
 }
 
-func (Beale) Grad(grad, x []float64) []float64 {
+func (Beale) Grad(grad, x []float64) {
 	if len(x) != 2 {
 		panic("dimension of the problem must be 2")
-	}
-	if grad == nil {
-		grad = make([]float64, len(x))
 	}
 	if len(x) != len(grad) {
 		panic("incorrect size of the gradient")
@@ -57,16 +54,13 @@ func (Beale) Grad(grad, x []float64) []float64 {
 
 	grad[0] = -2 * (f1*t1 + f2*t2 + f3*t3)
 	grad[1] = 2 * x[0] * (f1 + 2*f2*x[1] + 3*f3*x[1]*x[1])
-	return grad
 }
 
 func (Beale) Hess(dst *mat.SymDense, x []float64) {
 	if len(x) != 2 {
 		panic("dimension of the problem must be 2")
 	}
-	if dst.IsZero() {
-		*dst = *(dst.GrowSym(len(x)).(*mat.SymDense))
-	} else if len(x) != dst.Symmetric() {
+	if len(x) != dst.Symmetric() {
 		panic("incorrect size of the Hessian")
 	}
 
@@ -119,12 +113,9 @@ func (BiggsEXP2) Func(x []float64) (sum float64) {
 	return sum
 }
 
-func (BiggsEXP2) Grad(grad, x []float64) []float64 {
+func (BiggsEXP2) Grad(grad, x []float64) {
 	if len(x) != 2 {
 		panic("dimension of the problem must be 2")
-	}
-	if grad == nil {
-		grad = make([]float64, len(x))
 	}
 	if len(x) != len(grad) {
 		panic("incorrect size of the gradient")
@@ -144,7 +135,6 @@ func (BiggsEXP2) Grad(grad, x []float64) []float64 {
 		grad[0] += 2 * f * dfdx0
 		grad[1] += 2 * f * dfdx1
 	}
-	return grad
 }
 
 func (BiggsEXP2) Minima() []Minimum {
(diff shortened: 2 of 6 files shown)


optimize/minimize.go
```diff
@@ -480,7 +480,11 @@ func evaluate(p *Problem, loc *Location, op Operation, x []float64) {
 		loc.Gradient = p.Grad(loc.Gradient, x)
 	}
 	if op&HessEvaluation != 0 {
-		loc.Hessian = p.Hess(loc.Hessian, x)
+		// Make sure we have a destination in which to place the Hessian.
```
kortschak: This is here because it makes the initialisation vastly more convenient at the cost of some simple pixie dust. There is no documentation for this, though it could be added to `Location`, saying that `Hessian` will be allocated if needed.

diff/fd/hessian.go
```diff
@@ -11,28 +11,26 @@ import (
...
-	} else {
-		if n2 := dst.Symmetric(); n2 != n {
-			panic("hessian: dst size mismatch")
-		}
-		for i := 0; i < n; i++ {
-			for j := i; j < n; j++ {
-				dst.SetSym(i, j, 0)
-			}
+	if dst.IsZero() {
+		*dst = *(dst.GrowSym(n).(*mat.SymDense))
```
btracey: Outside the scope of this PR, but we should think about making an easier version of this. I'm thinking basically an exported version of `ReuseAs`. Then this can all be
```
n := len(x)
dst.ReuseAs(n)
dst.Zero()
``` 
(or ReuseAsSym). This would simplify code and make it easier for outside implementers.

diff/fd/hessian.go
```diff
@@ -11,28 +11,26 @@ import (
...
+	} else if dst.Symmetric() != n {
+		panic("hessian: dst size mismatch")
+	}
+	for i := 0; i < n; i++ {
```
btracey: While here, just dst.Zero()?

optimize/minimize.go
btracey: SGTM.

diff/fd/hessian.go
btracey: Though, perhaps the public `ReuseAs` should always zero, because otherwise it will be a very convenient source of bugs.

optimize/types.go
```diff
@@ -129,11 +129,9 @@ type Problem struct {
 	Grad func(grad []float64, x []float64) []float64
 
 	// Hess evaluates the Hessian at x and stores the result in-place in hess.
-	// Hess must not modify x. Hess may use (and return) the provided Symmetric
-	// if it is non-nil, or must allocate a new Symmetric otherwise. Minimize
-	// will 'give back' the returned Symmetric where possible, allowing Hess
-	// to use a type assertion on the provided matrix.
-	Hess func(hess mat.Symmetric, x []float64) mat.Symmetric
+	// Hess must not modify x. Hess may use the provided mat.SymDense, and will
```
btracey: In this case, `Hess` must use the provided `mat.SymDense`, and must resize it if zero-valued. With a TODO to document ReuseAs if/when it is added.

diff/fd/hessian.go
kortschak: I think just a local helper would be appropriate.

diff/fd/hessian.go
kortschak: I forgot we had that. I was wishing for one, but didn't bother looking.

optimize/types.go
kortschak: Done the comment here, but I'm not sure what you want with the TODO.

kortschak:
PTAL

I was also wanting broad agreement for this given how infrequently I come here.

optimize/functions/functions.go
```diff
@@ -60,14 +60,13 @@ func (Beale) Grad(grad, x []float64) []float64 {
 	return grad
 }
 
-func (Beale) Hess(hess mat.Symmetric, x []float64) mat.Symmetric {
+func (Beale) Hess(hess *mat.SymDense, x []float64) {
```
vladimir-ch: Should these be `dst` instead of `hess`?

kortschak: Yes. I think they should.

Done

kortschak:
@btracey and @vladimir-ch Are you both OK with this?

diff/fd/hessian.go
btracey: I think it would be nice to make it easy, especially to encourage outside users to follow the API conventions

kortschak: Can you file a proposal issue for that?

optimize/types.go
```diff
@@ -129,11 +129,9 @@ type Problem struct {
...
 
 	// Hess evaluates the Hessian at x and stores the result in-place in hess.
-	// Hess must not modify x. Hess may use (and return) the provided Symmetric
-	// if it is non-nil, or must allocate a new Symmetric otherwise. Minimize
-	// will 'give back' the returned Symmetric where possible, allowing Hess
-	// to use a type assertion on the provided matrix.
-	Hess func(hess mat.Symmetric, x []float64) mat.Symmetric
+	// Hess must not modify x. Hess must use the provided mat.SymDense, and
+	// must resize it if it is zero-sized.
+	Hess func(hess *mat.SymDense, x []float64)
```
vladimir-ch: I'm slightly concerned about the discord between `Grad` and `Hess` but I don't have an alternative other than Grad taking a non-nil pointer to correctly sized or zero sized slice.

diff/fd/hessian.go
```diff
@@ -11,30 +11,24 @@ import (
...
-// the Hessian will be estimated using the Forward formula and a default step size.
+// The resulting H will be stored in dst. Finite difference formula and other
+// options are specified by settings. If settings is nil, the Hessian will be
+// estimated using the Forward formula and a default step size.
 //
-// Hessian panics if the size of dst and x is not equal, or if the derivative
-// order of the formula is not 1.
-func Hessian(dst *mat.SymDense, f func(x []float64) float64, x []float64, settings *Settings) *mat.SymDense {
+// If the dst matrix is zero-sized it will be resized to the correct dimensions,
+// otherwise dst must match the dimension of the receiver or CovarianceMatrix
```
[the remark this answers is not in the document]
kortschak: Yes. Sorry about that.

btracey:
Dan, before submitting could you file a bug to fix up `optimize.Result`?

optimize/types.go
kortschak: I've been thinking about that. The difference and justification is that `Hess` takes a pointer value destination and `Grad` takes a non-pointer destination. This means that in order to resize the `Grad` destination in a way that the caller can see, there needs to be a returned value. This doesn't need to happen with the `Hess` destination. We could have them always returning the destination, but then `Hess` has two potential sources of truth, the passed in parameter and the returned value. Since these are potentially user-implemented functions, this is a crack for bugs to live in.

In short, I'm not wildly happy with it, but I think it's the least worst.

kortschak:
I'm not sure what it is that you want to change there.

BTW. I'm happy to fix up the copy/paste error that @vladimir-ch found. Given that you all have approved the substance of the PR, I'd only need a single approver after that.

kortschak:
PTAL

kortschak:
@vladimir-ch Sorry, I had thought that you had approved. I'll wait for your approval.

optimize/types.go
btracey: If we are doing it this way, we could go back to the old `optimize` way, where the user didn't need to care about it (and had to store in-place). We do need the ability for the `Method` to set the value to `nil`, but that case could checked and allocated in `Minimize`.  This was not possible for an arbitrary `Symmetric` but is very easy with a fixed `*SymDense`.

btracey: This would be simpler for users and would still mesh with `fd`, since the matrix would always be allocated.

kortschak: I'm unclear what the old `optimize` way was. What was the signature?

btracey:
```
Grad: func(dst, x []float64)
Hess: func(dst *mat.SymDense, x []float64)
```

The difference being that `Minimize` guarantees that `len(dst) == len(x)` in the call to grad, and `dst.Symmetric() == len(x)` in the call to `Hess`. In other words, `Minimize` guarantees they are the right size and the user doesn't have to worry about checking if they are nil or not.

kortschak: Ah. OK. Would you like me to update this for that behaviour?

btracey: I think so, yes. That check should probably happen in `evaluate`, minimize.go ~ line 470 (and there would be corresponding documentation changes in `Problem`)

kortschak: There's a TODO for consideration. I can do that in this PR if you want; whether the allocation of the gradient slice is when it is nil, or when it is zero-length.

btracey: We can leave as a TODO. I suspect the correct behavior is to only check if the relevant entry is `nil`. These values are never allocated by the provider of `Problem`, this is only for communicating between `Minimize` and `Method`. The reason to allow nil values is so the `Method` can safely save the values. It seems like allowing zero values as well could be ripe for bugs, in that it's mid-way between explicitly keeping the data and explicitly destroying it.

kortschak: That would suggest that the Hessian should also only be a nil check then.

btracey:
Agreed. Sorry, I should have been more clear

optimize/types.go
kortschak: Done

kortschak:
PTAL

kortschak:
It feels a lot cleaner now.

I have to say, doing this in, say python, would have been horrible.

optimize/types.go
vladimir-ch: This is better.

kortschak:
@sbinet Are you happy with this?

PR #962, opened 2019-04-16
set Stride in mat.Dense.Slice

pa-m:
Hi,
using result of Slice casted as (*mat.Dense) currently fails because Stride isn't set

Please take a look.


mat/dense.go
```diff
@@ -305,6 +305,7 @@ func (m *Dense) Slice(i, k, j, l int) Matrix {
 	t.mat.Data = t.mat.Data[i*t.mat.Stride+j : (k-1)*t.mat.Stride+l]
 	t.mat.Rows = k - i
 	t.mat.Cols = l - j
+	t.mat.Stride = m.mat.Stride
```
sbinet: I don't understand why this is needed.
as you can see, line 305, `t.mat.Stride` is used twice...

could you add a test that exercizes the issue you are talking about?

pa-m: You're right, it's not needed.

PR #966, opened 2019-04-18
bound: Add Union helper for bound package

corona10:
Add Union helper for bound package

Updates: #939

### DIFF
--- a/bound/bound.go
+++ b/bound/bound.go
@@ -16,23 +16,45 @@ func (b Bound) IsValid() bool {
 	return b.Min <= b.Max
 }
 
-// Intersection returns a Bound that is the intersection of the input bounds.
-// If the intersection is empty or if the input length is zero,
-// then the NaN Bound will be returned.
+// Intersection returns the intersection of the input bounds if possible.
+// Otherwise a NaN Bound is returned.
 func Intersection(bounds ...Bound) Bound {
 	if len(bounds) == 0 {
 		return Bound{Min: math.NaN(), Max: math.NaN()}
 	}
 
-	ret := Bound{Min: bounds[0].Min, Max: bounds[0].Max}
+	intersection := Bound{Min: bounds[0].Min, Max: bounds[0].Max}
 	for _, b := range bounds[1:] {
-		ret.Min = math.Max(ret.Min, b.Min)
-		ret.Max = math.Min(ret.Max, b.Max)
+		intersection.Min = math.Max(intersection.Min, b.Min)
+		intersection.Max = math.Min(intersection.Max, b.Max)
 	}
 
-	if !ret.IsValid() {
+	if !intersection.IsValid() {
 		return Bound{Min: math.NaN(), Max: math.NaN()}
 	}
 
-	return ret
+	return intersection
+}
+
+// Union returns the contiguous union of the input bounds if possible.
+// Otherwise a NaN Bound is returned.
+func Union(bounds ...Bound) Bound {
+	if len(bounds) == 0 {
+		return Bound{Min: math.NaN(), Max: math.NaN()}
+	}
+
+	union := Bound{Min: bounds[0].Min, Max: bounds[0].Max}
+	for _, b := range bounds[1:] {
+		if b.Max < union.Min || union.Max < b.Min {
+			return Bound{Min: math.NaN(), Max: math.NaN()}
+		}
+		union.Min = math.Min(union.Min, b.Min)
+		union.Max = math.Max(union.Max, b.Max)
+	}
+
+	if !union.IsValid() {
+		return Bound{Min: math.NaN(), Max: math.NaN()}
+	}
+
+	return union
 }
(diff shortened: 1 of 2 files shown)


bound/bound.go
```diff
@@ -36,3 +35,22 @@ func Intersection(bounds ...Bound) Bound {
...
+// Union returns a contiguous union of the input bounds if possible.
+// Otherwise a NaN Bound is returned.
+func Union(bounds ...Bound) Bound {
+	if len(bounds) == 0 {
+		return Bound{Min: math.NaN(), Max: math.NaN()}
+	}
+
+	ret := Bound{Min: bounds[0].Min, Max: bounds[0].Max}
+	for _, b := range bounds[1:] {
+		if ret.Max >= b.Min && b.Max >= ret.Min {
```
kortschak: If you invert this test, then the return takes the execution out of the function and we don't need an else block. It will also be clearer.
```
if b.Max < ret.Min || ret.Max < b.Min {
	return Bound{Min: math.NaN(), Max: math.NaN()}
}
ret.Min = math.Min(ret.Min, b.Min)
ret.Max = math.Max(ret.Max, b.Max)
```

bound/bound_test.go
```diff
@@ -75,3 +75,45 @@ func TestIsValid(t *testing.T) {
 		t.Error("b4 is invalid")
 	}
 }
+
+func TestUnion(t *testing.T) {
```
kortschak: Please add a TODO at the top of the file to rewrite these tests as table-driven.

bound/bound.go
```diff
@@ -36,3 +35,22 @@ func Intersection(bounds ...Bound) Bound {
...
 }
+
+// Union returns a contiguous union of the input bounds if possible.
+// Otherwise a NaN Bound is returned.
+func Union(bounds ...Bound) Bound {
+	if len(bounds) == 0 {
+		return Bound{Min: math.NaN(), Max: math.NaN()}
+	}
+
+	ret := Bound{Min: bounds[0].Min, Max: bounds[0].Max}
```
kortschak: We don't conventionally name variable based on whether they'll be returned. This might be, `u` or `union`.

bound/bound.go
```diff
@@ -16,9 +16,8 @@ func (b Bound) IsValid() bool {
 	return b.Min <= b.Max
 }
 
-// Intersection returns a Bound that is the intersection of the input bounds.
-// If the intersection is empty or if the input length is zero,
-// then the NaN Bound will be returned.
+// Intersection returns a contiguous intersection of the input bounds if possible.
```
kortschak: You can remove "contiguous" since all possible results from the intersection of contiguous intervals will be contiguous or null. Also change "a" to "the" since it returns the only possible answer.

bound/bound.go
```diff
@@ -36,3 +35,22 @@ func Intersection(bounds ...Bound) Bound {
 
 	return ret
 }
+
+// Union returns a contiguous union of the input bounds if possible.
```
kortschak: s/a/the/ for the same reason as above.

bound/bound_test.go
```diff
@@ -75,3 +77,45 @@ func TestIsValid(t *testing.T) {
...
+	b0 := Bound{Min: 0.0, Max: 5.0}
+	b1 := Bound{Min: 3.0, Max: 4.0}
+	b2 := Bound{Min: -1.0, Max: 1.0}
+	b3 := Bound{Min: 1.0, Max: 2.0}
+	b4 := Bound{Min: 1.1, Max: 2.0}
+
+	ret := Union(b0)
```
corona10: @kortschak 

Shall I update this one also into `union := Union(b0)`?
My intent was the `return value VS expected value`  for testing.

kortschak: You can leave the tests. We usually use `got` for test results, but this will be rewritten to use tables.

corona10: Okay!

corona10:
@kortschak Please take a look!

bound/bound.go
```diff
@@ -16,23 +16,40 @@ func (b Bound) IsValid() bool {
...
+	}
+
+	union := Bound{Min: bounds[0].Min, Max: bounds[0].Max}
+	for _, b := range bounds[1:] {
+		if b.Max < union.Min || union.Max < b.Min {
+			return Bound{Min: math.NaN(), Max: math.NaN()}
+		}
+		union.Min = math.Min(union.Min, b.Min)
+		union.Max = math.Max(union.Max, b.Max)
+	}
```
kortschak: For consistency with the behaviour of `Intersection`, we should probably convert invalid values at this point to NaNs. At the moment, if one of the inputs happens to have one NaN end and one non-NaN end, the output here will be the same unless it's otherwise invalid.

corona10:
@kortschak I've updated the CL. Please take a look

btracey:
Is this function right? If we have [[0, 1], [2, 3], [0.5, 2.5]] this has a valid union of [0,3] but this algorithm will return NaN as far as I can see

kortschak:
I'll add that to the tests I'm going to write in the next couple of hours and fix it up. It needs to sort the bounds first and then walk them. Thanks for pointing it out. This is why I always use trees for this.

PR #968, opened 2019-04-18
bound: rewrite tests to use tables and fix unsorted input handling for Union

kortschak:
Please take a look.

### DIFF
--- a/bound/bound.go
+++ b/bound/bound.go
@@ -4,20 +4,24 @@
 
 package bound
 
-import "math"
+import (
+	"math"
+	"sort"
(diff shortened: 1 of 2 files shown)


bound/bound.go
```diff
@@ -58,3 +67,9 @@ func Union(bounds ...Bound) Bound {
 
 	return union
 }
+
+type byStart []Bound
```
corona10: Is there any reason not to use `byMin`

kortschak: Done. Thanks

PR #969, opened 2019-04-19
bound: Add JaccardIndex helper for bound package

corona10:
bound: Add Jaccard helper for bound package

closes: #939

### DIFF
--- a/bound/bound.go
+++ b/bound/bound.go
@@ -14,6 +14,14 @@ type Bound struct {
 	Min, Max float64
 }
 
+// Len returns the length of the bound
+func (b Bound) Len() float64 {
+	if math.IsNaN(b.Min) && math.IsNaN(b.Max) {
+		return 0
+	}
+	return b.Max - b.Min
+}
+
 // IsValid returns whether the bound is valid. A valid bound will have
 // the minimum less than or equal to the maximum.
 func (b Bound) IsValid() bool {
@@ -32,8 +40,8 @@ func JaccardIndex(bounds ...Bound) float64 {
 		return math.NaN()
 	}
 	i := Intersection(bounds...)
-	di := i.Max - i.Min
-	du := u.Max - u.Min
+	di := i.Len()
+	du := u.Len()
 
 	if di == 0 {
 		// If the union is empty, the intersection must
--- a/bound/bound.go
+++ b/bound/bound.go
@@ -20,6 +20,37 @@ func (b Bound) IsValid() bool {
 	return b.Min <= b.Max
 }
 
+// JaccardIndex returns a jaccard index between bounds. The result is only valid for
+// sets of bounds that have a contiguous union.
+// If all sets of inputs bounds are empty, the result is 1.
+func JaccardIndex(bounds ...Bound) float64 {
+	if len(bounds) == 0 {
+		return math.NaN()
+	}
+	u := Union(bounds...)
+	if !u.IsValid() {
+		return math.NaN()
+	}
+	i := Intersection(bounds...)
+	di := i.Max - i.Min
+	du := u.Max - u.Min
+
+	if di == 0 {
+		// If the union is empty, the intersection must
+		// also be empty. J(Ø, Ø) is 1 by definition.
+		if du == 0 {
+			return 1
+		}
+
+		if !math.IsNaN(du) {
+			return 0
+		}
+		return math.NaN()
+	}
+
+	return di / du
+}
+
 // Intersection returns the intersection of the input bounds. If the
 // intersection is empty an invalid Bound is returned.
 func Intersection(bounds ...Bound) Bound {
--- a/bound/bound_test.go
+++ b/bound/bound_test.go
@@ -145,7 +145,7 @@ var jaccardTests = []struct {
 	},
 	{
 		bounds: []Bound{{Min: -1, Max: 1}, {Min: 1, Max: 2}},
-		want:   0.0,
+		want:   0,
 	},
 	{
 		bounds: []Bound{{Min: -1, Max: 1}, {Min: 1.1, Max: 2}},
@@ -153,11 +153,11 @@ var jaccardTests = []struct {
 	},
 	{
 		bounds: []Bound{{Min: 0, Max: 1}, {Min: 2, Max: 3}, {Min: 0.5, Max: 2.5}},
-		want:   math.NaN(),
+		want:   0,
 	},
 	{
 		bounds: []Bound{{Min: 0, Max: 1}, {Min: 1, Max: 1}},
-		want:   0.0,
+		want:   0,
 	},
 	{
 		bounds: []Bound{},
--- a/bound/bound_test.go
+++ b/bound/bound_test.go
@@ -126,3 +126,56 @@ func TestUnion(t *testing.T) {
 func same(a, b Bound) bool {
 	return floats.Same([]float64{a.Min, a.Max}, []float64{b.Min, b.Max})
 }
+
+var jaccardTests = []struct {
+	bounds []Bound
+	want   float64
+}{
+	{
+		bounds: []Bound{{Min: 0, Max: 5}},
+		want:   1,
+	},
+	{
+		bounds: []Bound{{Min: 0, Max: 5}, {Min: 3, Max: 4}},
+		want:   0.2,
+	},
+	{
+		bounds: []Bound{{Min: 0, Max: 5}, {Min: -1, Max: 1}},
+		want:   0.16666666666666666,
+	},
+	{
+		bounds: []Bound{{Min: -1, Max: 1}, {Min: 1, Max: 2}},
+		want:   0.0,
+	},
+	{
+		bounds: []Bound{{Min: -1, Max: 1}, {Min: 1.1, Max: 2}},
+		want:   math.NaN(),
+	},
+	{
+		bounds: []Bound{{Min: 0, Max: 1}, {Min: 2, Max: 3}, {Min: 0.5, Max: 2.5}},
+		want:   math.NaN(),
+	},
+	{
+		bounds: []Bound{{Min: 0, Max: 1}, {Min: 1, Max: 1}},
+		want:   0.0,
+	},
+	{
+		bounds: []Bound{},
+		want:   math.NaN(),
+	},
+	{
+		bounds: nil,
+		want:   math.NaN(),
+	},
+}
+
+func TestJaccardIndex(t *testing.T) {
+	for _, test := range jaccardTests {
+		got := JaccardIndex(test.bounds...)
+		if got != test.want {
+			if !math.IsNaN(test.want) || !math.IsNaN(got) {
+				t.Errorf("unexpected result from Jaccard(%#v...): got:%+v want:%+v", test.bounds, got, test.want)
+			}
+		}
+	}
+}


bound/bound_test.go
```diff
@@ -126,3 +126,50 @@ func TestUnion(t *testing.T) {
...
+		bounds: []Bound{{Min: -1, Max: 1}, {Min: 1.1, Max: 2}},
+		want:   0.0,
+	},
+	{
+		bounds: []Bound{{Min: 0, Max: 1}, {Min: 2, Max: 3}, {Min: 0.5, Max: 2.5}},
+		want:   0.0,
+	},
+	{
+		bounds: []Bound{},
+		want:   0.0,
```
kortschak: I would have thought this would be NaN. Same for nil.

bound/bound.go
```diff
@@ -20,6 +20,26 @@ func (b Bound) IsValid() bool {
...
 }
 
+// Jaccard returns a jaccard index between bounds
+func Jaccard(bounds ...Bound) float64 {
+	if len(bounds) == 0 {
+		return 0.0
+	}
+
+	union := Union(bounds...)
+	if !union.IsValid() || union.Max-union.Min == 0.0 {
```
kortschak: If the union is invalid, return NaN. If the size of the bound is zero, let this fall through. This can be done simply, by writing:
```
u := Union(bounds...)
i : = Intersection(bounds...)
return (i.Max - i.Min) / (u.Max - u.Min)
```

Though we cost a little more in cases where the union is invalid since we don't short circuit the intersection call.

bound/bound.go
```diff
@@ -20,6 +20,26 @@ func (b Bound) IsValid() bool {
...
+		return 0.0
+	}
+
+	inter := Intersection(bounds...)
+	if !inter.IsValid() {
+		return 0.0
```
kortschak: If it's invalid, return NaN.

bound/bound.go
```diff
@@ -20,6 +20,26 @@ func (b Bound) IsValid() bool {
 	return b.Min <= b.Max
 }
 
+// Jaccard returns a jaccard index between bounds
+func Jaccard(bounds ...Bound) float64 {
+	if len(bounds) == 0 {
+		return 0.0
```
kortschak: Return NaN.

corona10:
@kortschak Done! Thanks

bound/bound.go
```diff
@@ -20,6 +20,13 @@ func (b Bound) IsValid() bool {
 	return b.Min <= b.Max
 }
 
+// Jaccard returns a jaccard index between bounds
```
kortschak: Final full stop.

I would give a caveat here that the JAccard result is only correct for sets of bounds that have a contiguous union.

bound/bound_test.go
```diff
@@ -126,3 +126,52 @@ func TestUnion(t *testing.T) {
...
+	},
+	{
+		bounds: []Bound{{Min: -1, Max: 1}, {Min: 1, Max: 2}},
+		want:   0.0,
+	},
+	{
+		bounds: []Bound{{Min: -1, Max: 1}, {Min: 1.1, Max: 2}},
+		want:   math.NaN(),
+	},
+	{
```
kortschak: I need to rethink the approach. I would expect this to return 0. Maybe, rather than what I suggested, get the lengths of each of the union and intersection and convert to 0 if NaN, then do the Jaccard calculation.

bound/bound.go
```diff
@@ -20,6 +20,13 @@ func (b Bound) IsValid() bool {
...
+func Jaccard(bounds ...Bound) float64 {
```
kortschak: Sorry I missed this before. This should only have two input params, `Jaccard(a, b Bound) float64`. A distance between more than that is not meaningful.

btracey: Are you sure about this? The Jaccard wikipedia article defines it as "The intersection of a set divided by the union of a set". That seems pretty well defined, and this package already has a definition of intersection and union.

bound/bound.go
btracey: Do you think this should be called JaccardIndex since there is also a well-defined JaccardDistance?

bound/bound.go
kortschak: You're right. Thinking about it formally, I could see a meaning, as soon as I thought about it visually, it makes sense.

bound/bound.go
corona10: @btracey 
Okay, I will update it into JaccardDistance :)

kortschak: `JaccardIndex`

corona10: @btracey @kortschak Oh.. I mistranslated... I 've now understood what you are saying!

corona10:
@kortschak Please take a look!

bound/bound.go
```diff
@@ -20,6 +20,25 @@ func (b Bound) IsValid() bool {
...
+	}
+	u := Union(bounds...)
+	if !u.IsValid() {
+		return 0.0
+	}
+	i := Intersection(bounds...)
+	if !i.IsValid() {
+		return 0.0
+	}
+	return (i.Max - i.Min) / (u.Max - u.Min)
```
kortschak:
```
	di := i.Max - i.Min
	du := u.Max - u.Min
	if du == 0 {
		// If the union is empty, the intersection must
		// also be empty. J(Ø, Ø) is 1 by definition.
		return 1
	}
	return di/du
```

bound/bound_test.go
```diff
@@ -126,3 +126,50 @@ func TestUnion(t *testing.T) {
...
+		bounds: []Bound{},
+		want:   1.0,
+	},
+	{
+		bounds: nil,
+		want:   1.0,
+	},
+}
+
+func TestJaccard(t *testing.T) {
```
kortschak: `TestJaccardIndex`

bound/bound_test.go
```diff
@@ -126,3 +126,50 @@ func TestUnion(t *testing.T) {
...
 	return floats.Same([]float64{a.Min, a.Max}, []float64{b.Min, b.Max})
 }
+
+var jaccardTests = []struct {
+	bounds []Bound
+	want   float64
+}{
+	{
+		bounds: []Bound{{Min: 0, Max: 5}},
+		want:   1.0,
```
kortschak: Note that the decimal and trailing zero is not needed since `want` is already declared to be a `float64`.

corona10:
@kortschak 

Done! Thanks

bound/bound_test.go
kortschak: This still doesn't seem right. The intersection of this is empty, but the union is [0, 3]. So the index should be 0.

kortschak:
The confusion here stems from the fact that NaN is used to indicate empty. [x, x] is not empty though, it is x. I think the special cases need to be enumerated and put in tests.

corona10:
@kortschak 
I'd like to add this one. What do you think?
```go
// IsEmpty returns whether the bound is empty.
// Note that IsEmpty does not indicate whether the bound is valid or not
func (b Bound) IsEmpty() bool {
	return math.IsNaN(b.Min) || math.IsNaN(b.Max)
}
```

corona10:
Umm situation is more confused than I think. 
 I'll give it some thought

corona10:
@kortschak 
> The confusion here stems from the fact that NaN is used to indicate empty. [x, x] is not empty though, it is x. I think the special cases need to be enumerated and put in tests.

So if the union is [0, 3] and the intersection is [1,1]
The expected result should be (1-1+1)/(3-0+1)?

kortschak:
This is the problem. [1, 1] is infinitesimal. The size of the set [0, 3] is 3 plus an infinitesimal. `float64` doesn't represent infinitesimals. For the most part, the approach we have makes sense, so here we just need to make special cases, I would say that J([1, 1], [0, 3]) is 0 (it certainly limits to that).

The contingencies are (I think):

|\|intersection\||\|union\||result|notes|
|-|-|-|-|
|x|y|x/y|x and y are non-zero, non-NaN|
|0|0|1|by definition|
|~~x~~|~~0~~|~~x/0~~|not possible unless x is zero|
|0|y|0|y is non-NaN|
|~~NaN~~|~~0~~|~~NaN~~|not possible|
|0|NaN|NaN|happens with completely discontiguous sets which we don't handle|
|~~NaN~~|~~y~~|~~NaN~~|not possible|
|x|NaN|NaN|happens with partially discontiguous sets which we don't handle|
|NaN|NaN|NaN|empty/invalid input|

kortschak:
I'd rather not have an `IsEmpty`.

bound/bound_test.go
corona10: @kortschak 
Current intersection implementation returns `{NaN NaN}`for `[{0.5 2.5} {0 1} {2 3}]`.
According to https://github.com/gonum/gonum/pull/969#issuecomment-485382802
NaN/3.0 == NaN.
Should we deal `{NaN NaN}` as empty bound?

kortschak: Yes, `{NaN NaN}` should be treated as empty. This is the problems we have, there is invalid empty and validly empty and we don't distinguish between them.

corona10:
@kortschak 
I've added `Len()` for handling empty bound in a consistent way.
Please left feedbacks for API name and signature.

PTAL

kortschak:
I don't know that that is what we want to do. The NaN signalling is more subtle than that method indicates. I'd like others' input here though. @btracey?

btracey:
I think that table mostly good. The size of [0, 3] U [1, 1] is just 3, it's not 3 plus an infintesimal. The intersection is just the point 1, which has measure 0, and so Jaccard should clearly return 0.

You say intersection = Nan, Union = Non Nan cannot happen, but wouldn't [0, 2] [1, 3] [2.5, 3.5] have Nan intersection but finite union?

I'm not sure we should have `Len` the way it is though. I feel like an invalid bound should have `NaN` length (though obviously the current behvaior is nice for saying 0 intersection/union.

This is way more complicated than it first seemed.

kortschak:
> I think that table mostly good. The size of [0, 3] U [1, 1] is just 3, it's not 3 plus an infintesimal. The intersection is just the point 1, which has measure 0, and so Jaccard should clearly return 0.

What I was saying was that |[x, y]| < |[x, y)| for x < y. We can calculate |[x, y)| trivially, but the open interval has the right end point and so it infinitesimially longer. Is there convention that |[1, 1]| is 0? (I understand that there are reasons why this would be the case, but I want to clarify).
 
> You say intersection = Nan, Union = Non Nan cannot happen, but wouldn't [0, 2] [1, 3] [2.5, 3.5] have Nan intersection but finite union?

If we are using NaN to signal empty. Yes. I have corrected that.

btracey:
>  Is there convention that |[1, 1]| is 0?

Yes. In countable sets you can say that {1} has a cardinality of 1, but you need a different idea with uncountable sets. This is where measure theory comes in. The most common measure is the Lebesgue measure, where the size of (a, b), [a, b), and [a, b] are all the same and equal to b - a.  https://en.wikipedia.org/wiki/Lebesgue_measure  . This is used, for example, in Lebesgue integration (an alternate approach than Riemannian integration). The single element under this definition has zero measure. A bit mathy, but you can see http://math.uchicago.edu/~may/REU2013/MeasureZero.pdf for a definition. If you look at exercise 1.4, a special case is that a point has no measure compared to the line, and so [1,1] is 0.

kortschak:
Thanks. Yes, Lyron and I have discussed these kinds of things in the past, so should have known better.

vladimir-ch:
I think this is complicated because we've made it so by using an invalid interval to denote both a) valid empty returns from Intersection, and b) invalid returns from Union in response to input that we don't handle, which is what causes the trouble here.

I still think that current Union is not very useful. It requires that the input intervals are overlapping. If the user knows that they are overlapping, then this function just replaces a simple for loop. If they don't know whether they are overlapping, then this function is as useful as `bounds.AreOverlapping` (which avoids the invalid interval), for `true` return we're at the simple loop case, for `false` return the current `Union` doesn't provide any additional information either.

Apart from that, having to check the return from something called `Union` for validity is surprising and not right, although it could be remedied by renaming it to something else.

In my opinion there should be only `IsEmpty` for which the measure is 0 and build from that.

kortschak:
>I think this is complicated because we've made it so by using an invalid interval to denote both a) valid empty returns from Intersection, and b) invalid returns from Union in response to input that we don't handle, which is what causes the trouble here.

Agreed.

I agree with the approach of having a function to return whether a set of bounds are overlapping.

Another possible solution is to put everything on the pointer and use `nil` for invalid.

A bigger question is whether we drop all the helpers that were added here? I suggested them, but do we need them?

vladimir-ch:
> Another possible solution is to put everything on the pointer and use nil for invalid.

I'd rather avoid having to return an invalid interval in the first place. Why do you think it is inevitable?

kortschak:
If a user insists on calling `JaccardIndex` on a set of bounds after getting a `false` result from `AreOverlapping` (or more likely does not call `AreOverlapping` at all), what do we do?

I think we need to think more carefully about what we actually want to provide.

vladimir-ch:
I don't think we should add `AreOverlapping`, it was just an example of how to get the same utility as from Union but without having to deal with the invalid interval.

> I think we need to think more carefully about what we actually want to provide.

Exactly.

> If a user insists on calling JaccardIndex on a set of bounds after getting a false result from AreOverlapping ...

If they are not overlapping, the intersection is the empty set of measure 0 and JaccardIndex returns 0. This leads to the question whether it makes sense to ask about similarity of multiple intervals because one dis-similar can spoil the similarity of the rest without indicating which one the culprit is. Perhaps it would be better to provide only pair-wise JaccardIndex?

kortschak:
OK, that approach to Jaccard index seems reasonable. Can we come to some consensus on what the features that we want to implement are since it seems that `Union` should go away? I suggest that this should happen at #939.

sbinet:
ping @corona10 ?

corona10:
@sbinet  cc @kortschak 
AFAIK, commits related  #939 should be rollbacked? 
https://github.com/gonum/gonum/issues/939#issuecomment-493803700

I think that we can discuss after how bound package will be changed

kortschak:
Closing in line with #1002. If we come up with a sensible approach we can revisit this. Thank you for putting in the time to do this and I'm sorry I didn't have a clearer plan for it.

PR #970, opened 2019-04-22
travis: run multi-os testing

kortschak:
Please take a look.


kortschak:
The failure here is due to Travis-CI helpfully setting `core.autcrlf` in the global gitconfig in a way that is not user configurable. This causes some of our parse tests to fail.

kortschak:
>(the alternative would be to have the `deps.d/windows/xxx` scripts detect the Go version and bail out. This would still use slots and some resources...)

Yes, I considered this and came to the same conclusion.

The thing that will prevent this from being a (huge) problem is that if someone gets it wrong it will be very obvious. I can think about writing a check to look at whether versions that are not intended to be tested have been tested and then fail early, but not right now.

kortschak:
Oh. The other option was to run all tags cases on the most recent release for all GOARCHs. This would reduce the exclude text at the cost of more test time. We can revisit this later.

PR #972, opened 2019-04-24
spatial/kdtree: new package implementing k-d tree storage

kortschak:
Please take a look.

This is a reworking of the bíogo kdtree implemented in 2012. I was the only author on the files in that package, so it can be brought into Gonum without any additional licensing.

I have polished up some of the more glaring gotchas (sentinels are now removed from sets filled by the nearest neighbour search for example) and removed gocheck from the tests. One issue that needs consideration is the iteration termination signal. I use a boolean to indicate when the iteration has been completed. This has the advantage that an empty return means continue, but I'm open to discussion on this (an error return could be a possibility - similar to `filepath.Walk`).

See https://github.com/gonum/exp/pull/27#issuecomment-484435947

### DIFF
--- a/spatial/kdtree/doc.go
+++ b/spatial/kdtree/doc.go
@@ -4,6 +4,5 @@
 
 // Package kdtree implements a k-d tree.
 //
-// See https://en.wikipedia.org/w/index.php?title=K-d_tree for more details of
-// k-d tree functionality.
+// See https://en.wikipedia.org/wiki/K-d_tree for more details of k-d tree functionality.
 package kdtree // import "gonum.org/v1/gonum/spatial/kdtree"
--- a/spatial/kdtree/doc.go
+++ b/spatial/kdtree/doc.go
@@ -1,0 +1,9 @@
+// Copyright ©2019 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Package kdtree implements a k-d tree.
+//
+// See https://en.wikipedia.org/w/index.php?title=K-d_tree for more details of
+// k-d tree functionality.
+package kdtree // import "gonum.org/v1/gonum/spatial/kdtree"
--- a/spatial/kdtree/kdtree.go
+++ b/spatial/kdtree/kdtree.go
@@ -148,10 +148,7 @@ func buildBounded(p bounder, plane Dim, bounding bool) *Node {
 	d := p.Index(piv)
 	np := (plane + 1) % Dim(d.Dims())
 
-	var b *Bounding
-	if bounding {
-		b = p.Bounds()
-	}
+	b := p.Bounds()
 	return &Node{
 		Point:    d,
 		Plane:    plane,
--- a/spatial/kdtree/kdtree.go
+++ b/spatial/kdtree/kdtree.go
@@ -1,0 +1,468 @@
+// Copyright ©2019 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package kdtree
+
+import (
+	"container/heap"
+	"fmt"
+	"math"
+	"sort"
+)
+
+type Interface interface {
+	// Index returns the ith element of the list of points.
+	Index(i int) Comparable
+
+	// Len returns the length of the list.
+	Len() int
+
+	// Pivot partitions the list based on the dimension specified.
+	Pivot(Dim) int
+
+	// Slice returns a slice of the list.
+	Slice(start, end int) Interface
+}
+
+// Bounder returns a bounding volume containing the list of points. Bounds may return nil.
+type Bounder interface {
+	Bounds() *Bounding
+}
+
+type bounder interface {
+	Interface
+	Bounder
+}
+
+// Dim is an index into a point's coordinates.
+type Dim int
+
+// Comparable is the element interface for values stored in a k-d tree.
+type Comparable interface {
+	// Compare returns the shortest translation of the plane through b with
+	// normal vector along dimension d to the parallel plane through a.
+	//
+	// Given c = a.Compare(b, d):
+	//  c = a_d - b_d
+	//
+	Compare(Comparable, Dim) float64
+
+	// Dims returns the number of dimensions described in the Comparable.
+	Dims() int
+
+	// Distance returns the squared Euclidean distance between the receiver and
+	// the parameter.
+	Distance(Comparable) float64
+}
+
+// Extender is a Comparable that can increase a bounding volume to include the
+// point represented by the Comparable.
+type Extender interface {
+	Comparable
+
+	// Extend returns a bounding box that has been extended to include the
+	// receiver. Extend may return nil.
+	Extend(*Bounding) *Bounding
+}
+
+// Bounding represents a volume bounding box.
+type Bounding struct {
+	Min, Max Comparable
+}
+
+// Contains returns whether c is within the volume of the Bounding. A nil Bounding
+// returns true.
+func (b *Bounding) Contains(c Comparable) bool {
+	if b == nil {
+		return true
+	}
+	for d := Dim(0); d < Dim(c.Dims()); d++ {
+		if c.Compare(b.Min, d) < 0 || 0 < c.Compare(b.Max, d) {
+			return false
+		}
+	}
+	return true
+}
+
+// Node holds a single point value in a k-d tree.
+type Node struct {
+	Point       Comparable
+	Plane       Dim
+	Left, Right *Node
+	*Bounding
+}
+
+func (n *Node) String() string {
+	if n == nil {
+		return "<nil>"
+	}
+	return fmt.Sprintf("%.3f %d", n.Point, n.Plane)
+}
+
+// Tree implements a k-d tree creation and nearest neighbor search.
+type Tree struct {
+	Root  *Node
+	Count int
+}
+
+// New returns a k-d tree constructed from the values in p. If p is a Bounder and
+// bounding is true, bounds are determined for each node.
+func New(p Interface, bounding bool) *Tree {
+	if p, ok := p.(bounder); ok && bounding {
+		return &Tree{
+			Root:  buildBounded(p, 0, bounding),
+			Count: p.Len(),
+		}
+	}
+	return &Tree{
+		Root:  build(p, 0),
+		Count: p.Len(),
+	}
+}
+
+func build(p Interface, plane Dim) *Node {
+	if p.Len() == 0 {
+		return nil
+	}
+
+	piv := p.Pivot(plane)
+	d := p.Index(piv)
+	np := (plane + 1) % Dim(d.Dims())
+
+	return &Node{
+		Point:    d,
+		Plane:    plane,
+		Left:     build(p.Slice(0, piv), np),
+		Right:    build(p.Slice(piv+1, p.Len()), np),
+		Bounding: nil,
+	}
+}
+
+func buildBounded(p bounder, plane Dim, bounding bool) *Node {
+	if p.Len() == 0 {
+		return nil
+	}
+
+	piv := p.Pivot(plane)
+	d := p.Index(piv)
+	np := (plane + 1) % Dim(d.Dims())
+
+	var b *Bounding
+	if bounding {
+		b = p.Bounds()
+	}
+	return &Node{
+		Point:    d,
+		Plane:    plane,
+		Left:     buildBounded(p.Slice(0, piv).(bounder), np, bounding),
+		Right:    buildBounded(p.Slice(piv+1, p.Len()).(bounder), np, bounding),
+		Bounding: b,
+	}
+}
+
+// Insert adds a point to the tree, updating the bounding volumes if bounding is
+// true, and the tree is empty or the tree already has bounding volumes stored,
+// and c is an Extender. No rebalancing of the tree is performed.
+func (t *Tree) Insert(c Comparable, bounding bool) {
+	t.Count++
+	if t.Root != nil {
+		bounding = t.Root.Bounding != nil
+	}
+	if c, ok := c.(Extender); ok && bounding {
+		t.Root = t.Root.insertBounded(c, 0, bounding)
+		return
+	} else if !ok && t.Root != nil {
+		// If we are not rebounding, mark the tree as non-bounded.
+		t.Root.Bounding = nil
+	}
+	t.Root = t.Root.insert(c, 0)
+}
+
+func (n *Node) insert(c Comparable, d Dim) *Node {
+	if n == nil {
+		return &Node{
+			Point:    c,
+			Plane:    d,
+			Bounding: nil,
+		}
+	}
+
+	d = (n.Plane + 1) % Dim(c.Dims())
+	if c.Compare(n.Point, n.Plane) <= 0 {
+		n.Left = n.Left.insert(c, d)
+	} else {
+		n.Right = n.Right.insert(c, d)
+	}
+
+	return n
+}
+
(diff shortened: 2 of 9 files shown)


spatial/kdtree/doc.go
```diff
@@ -0,0 +1,9 @@
+// Copyright ©2019 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Package kdtree implements a k-d tree.
+//
+// See https://en.wikipedia.org/w/index.php?title=K-d_tree for more details of
```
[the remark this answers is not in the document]
kortschak: Thanks. Done.

spatial/kdtree/kdtree.go
```diff
@@ -0,0 +1,469 @@
...
+// Tree implements a k-d tree creation and nearest neighbor search.
+type Tree struct {
+	Root  *Node
+	Count int
+}
+
+// New returns a k-d tree constructed from the values in p. If p is a Bounder and
+// bounding is true, bounds are determined for each node.
+func New(p Interface, bounding bool) *Tree {
+	if p, ok := p.(bounder); ok && bounding {
```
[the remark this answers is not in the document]
kortschak: This happens only once at the beginning of the construction, which is vastly more expensive than a type assertion. To do the bounding check first reduces readability, so I'd rather leave it like this.

spatial/kdtree/kdtree.go
```diff
@@ -0,0 +1,469 @@
...
+	if p.Len() == 0 {
+		return nil
+	}
+
+	piv := p.Pivot(plane)
+	d := p.Index(piv)
+	np := (plane + 1) % Dim(d.Dims())
+
+	var b *Bounding
+	if bounding {
```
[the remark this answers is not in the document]
kortschak: Indeed it is. Removed.

spatial/kdtree/kdtree_simple_example_test.go
```diff
@@ -0,0 +1,77 @@
...
+
+import (
+	"fmt"
+	"math"
+
+	"gonum.org/v1/gonum/spatial/kdtree"
+)
+
+func ExampleTree() {
+	// Example data from https://en.wikipedia.org/w/index.php?title=K-d_tree
```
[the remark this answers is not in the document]
kortschak: Done

spatial/kdtree/kdtree_simple_example_test.go
```diff
@@ -0,0 +1,77 @@
...
+	t := kdtree.New(points, false)
+	q := kdtree.Point{8, 7}
+	p, d := t.Nearest(q)
+	fmt.Printf("%v is closest point to %v, d=%f\n", p, q, math.Sqrt(d))
+	// Output:
+	// [9 6] is closest point to [8 7], d=1.414214
+}
+
+func ExampleTree_bounds() {
+	// Example data from https://en.wikipedia.org/w/index.php?title=K-d_tree
```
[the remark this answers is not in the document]
kortschak: Done

spatial/kdtree/medians_test.go
```diff
@@ -0,0 +1,187 @@
...
+	rnd := rand.New(rand.NewSource(1))
+
+	for i := 0; i < b.N; i++ {
+		b.StopTimer()
+		list := make(ints, 1e4)
+		for i := range list {
+			list[i] = rnd.Int()
+		}
+		b.StartTimer()
+		_ = MedianOfMedians(list)
```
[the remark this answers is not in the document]
kortschak: Done

spatial/kdtree/medians_test.go
```diff
@@ -0,0 +1,187 @@
...
+	rnd := rand.New(rand.NewSource(1))
+
+	b.StopTimer()
+	list := make(ints, 1e4)
+	for i := range list {
+		list[i] = rnd.Int()
+	}
+	b.StartTimer()
+	for i := 0; i < b.N; i++ {
+		_ = MedianOfRandoms(list, list.Len()/1e3)
```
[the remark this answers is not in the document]
kortschak: Done

kortschak:
I can't really see a reason for other that true/false, and if there is a need, the decision is made in a closure, so any additional information can be retained.

kortschak:
PTAL

spatial/kdtree/kdtree.go
```diff
@@ -0,0 +1,465 @@
...
+	return &Node{
+		Point:    d,
+		Plane:    plane,
+		Left:     buildBounded(p.Slice(0, piv).(bounder), np, bounding),
+		Right:    buildBounded(p.Slice(piv+1, p.Len()).(bounder), np, bounding),
+		Bounding: b,
+	}
+}
+
+// Insert adds a point to the tree, updating the bounding volumes if bounding is
```
vladimir-ch: After several readings I can match the comment to the code but it did take me some time. I don't have an alternative suggestion but does `Insert` have to take `bounding bool`? Can't the `bounding` passed to `New` determine the behavior of the whole tree?

spatial/kdtree/kdtree.go
```diff
@@ -0,0 +1,465 @@
...
+type Heap []ComparableDist
+
+func (h *Heap) Max() ComparableDist  { return (*h)[0] }
+func (h *Heap) Len() int             { return len(*h) }
+func (h *Heap) Less(i, j int) bool   { return (*h)[i].Comparable == nil || (*h)[i].Dist > (*h)[j].Dist }
+func (h *Heap) Swap(i, j int)        { (*h)[i], (*h)[j] = (*h)[j], (*h)[i] }
+func (h *Heap) Push(x interface{})   { (*h) = append(*h, x.(ComparableDist)) }
+func (h *Heap) Pop() (i interface{}) { i, *h = (*h)[len(*h)-1], (*h)[:len(*h)-1]; return i }
+
+// NKeeper is a Keeper that retains the n best ComparableDists that it is called to Keep.
```
vladimir-ch: There's something wrong with `... that it is called to Keep`. Or is it a play on the capital `K`?

spatial/kdtree/kdtree.go
```diff
@@ -0,0 +1,465 @@
...
+
+// NewNKeeper returns an NKeeper with the max value of the heap set to infinite distance. The
+// returned NKeeper is able to retain at most n values.
+func NewNKeeper(n int) *NKeeper {
+	k := NKeeper{make(Heap, 1, n)}
+	k.Heap[0].Dist = inf
+	return &k
+}
+
+// Keep add c to the heap if its distance is less than the maximum value of the heap. If adding
```
vladimir-ch: `Keep adds c ...`

spatial/kdtree/kdtree.go
```diff
@@ -0,0 +1,465 @@
...
+	Keep(ComparableDist) // Keep conditionally pushes the provided ComparableDist onto the heap.
+	Max() ComparableDist // Max returns the maximum element of the Keeper.
+	heap.Interface
+}
+
+// NearestSet finds the nearest values to the query accepted by the provided Keeper, k.
+// k must be able to return a ComparableDist specifying the maximum acceptable distance
+// when Max() is called, and retains the results of the search in min sorted order after
+// the call to NearestSet returns.
+// If a sentinel ComparableDist with a nil Comparable is used by the Keeper to mark the
```
vladimir-ch: It could be me but I haven't yet fully grasped this concept with a nil Comparable as a sentinel. Either I'm missing something or it is not explained sufficiently.

spatial/kdtree/kdtree_simple_example_test.go
```diff
@@ -0,0 +1,77 @@
...
+func ExampleTree_Do() {
+	// Example data from https://en.wikipedia.org/wiki/K-d_tree
+	points := kdtree.Points{{2, 3}, {5, 4}, {9, 6}, {4, 7}, {8, 1}, {7, 2}}
+
+	// Print all points in the data set within 3 of (3, 5).
+	t := kdtree.New(points, false)
+	q := kdtree.Point{3, 5}
+	t.Do(func(c kdtree.Comparable, _ *kdtree.Bounding, _ int) (done bool) {
+		// Compare each distance and output points
+		// with a distance less than 9. Distance
```
vladimir-ch: `less than 3`, maybe also write `9` below as `3*3`?

spatial/kdtree/kdtree_user_type_example_test.go
```diff
@@ -0,0 +1,163 @@
...
+	q := place{lat: 51.501476, lon: -0.140634}
+
+	var keep kdtree.Keeper
+
+	// Find all stations within 0.75 of the residence.
+	keep = kdtree.NewDistKeeper(0.75 * 0.75) // Distances are squared.
+	t.NearestSet(keep, q)
+
+	fmt.Println(`Stations within 750 m of 51.501476N 0.140634W.`)
+	for _, c := range keep.(*kdtree.DistKeeper).Heap {
```
vladimir-ch: Could the type assertion be avoided somehow, e.g. by extending the `Keeper` interface?

spatial/kdtree/kdtree_user_type_example_test.go
```diff
@@ -0,0 +1,163 @@
...
+	{name: "Westminster", lat: 51.501, lon: -0.1254},
+}
+
+// place is a kdtree.Comparable implementations.
+type place struct {
+	name     string
+	lat, lon float64
+}
+
+// Compare satisfies the axis comparions method of the kdtree.Comparable interface.
```
vladimir-ch: s/comparions/comparison

spatial/kdtree/kdtree_user_type_example_test.go
```diff
@@ -0,0 +1,163 @@
...
+
+// places is a collection of the place type that satisfies kdtree.Interface.
+type places []place
+
+func (p places) Index(i int) kdtree.Comparable         { return p[i] }
+func (p places) Len() int                              { return len(p) }
+func (p places) Pivot(d kdtree.Dim) int                { return plane{places: p, Dim: d}.Pivot() }
+func (p places) Slice(start, end int) kdtree.Interface { return p[start:end] }
+
+// plane is required to help places.
```
vladimir-ch: help?

spatial/kdtree/kdtree.go
kortschak: 'called' in this sense is 'demand', however, it is grammatically slightly off; it should be 'that has been called to Keep'. It is a small play on word. I can change it to more normal phrasing:
```
// NKeeper is a Keeper that retains the n best ComparableDists that have been passed to Keep.
```

spatial/kdtree/kdtree.go
kortschak: It's something that fits in with the concept that "You don't need to understand this," just don't insert one.

The heap has an intrinsic property of a maximum value. I could have a maximum that is stored outside the heap to initially configure the keeper (and I did look into doing this), but it significantly complicates the code. The nil sentinel merely provides an upper bound on the distance for items that the keeper will bother to try to retain. The fact that it's nil allows the keeper to remove the sentinel after it has gone past its use (the bíogo version leaves the sentinel in, which is irritating).

spatial/kdtree/kdtree_user_type_example_test.go
kortschak: There is no intrinsic reason that a `Keeper` should have a heap.

spatial/kdtree/kdtree_user_type_example_test.go
kortschak: Yes, `plane` is a helper type for `places`. "Se requiere `plane` para ayudar a `places`".

spatial/kdtree/kdtree.go
```diff
@@ -0,0 +1,465 @@
...
+package kdtree
+
+import (
+	"container/heap"
+	"fmt"
+	"math"
+	"sort"
+)
+
+type Interface interface {
```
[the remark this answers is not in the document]
kortschak: Done

spatial/kdtree/kdtree.go
```diff
@@ -0,0 +1,465 @@
...
+	Interface
+	Bounder
+}
+
+// Dim is an index into a point's coordinates.
+type Dim int
+
+// Comparable is the element interface for values stored in a k-d tree.
+type Comparable interface {
+	// Compare returns the shortest translation of the plane through b with
```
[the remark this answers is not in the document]
kortschak: Nice. I think the wording I used came from a formal definition.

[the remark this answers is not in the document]
kortschak: Nice. I think the wording I used came from a formal definition.

spatial/kdtree/kdtree.go
```diff
@@ -0,0 +1,465 @@
...
+	// Index returns the ith element of the list of points.
+	Index(i int) Comparable
+
+	// Len returns the length of the list.
+	Len() int
+
+	// Pivot partitions the list based on the dimension specified.
+	Pivot(Dim) int
+
+	// Slice returns a slice of the list.
```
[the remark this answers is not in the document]
kortschak: Done

spatial/kdtree/kdtree.go
kortschak: It's possible to build a tree without using `New`, so unfortunately it's not possible to know what the intention is for the first point being added without having this parameter.

spatial/kdtree/kdtree.go
kortschak: Done

spatial/kdtree/kdtree_simple_example_test.go
kortschak: I've said "Euclidean distance less than 3" now. Formally, it's still a distance less than 9, but this is clearer now.

spatial/kdtree/kdtree_user_type_example_test.go
kortschak: Done

spatial/kdtree/kdtree_user_type_example_test.go
vladimir-ch: I meant rather whether `Keeper` should have a method that returns the collected points but now I see that the assertion is done here only because you reuse `keep` for both `DistKeeper` and `NKeeper`.

spatial/kdtree/kdtree_user_type_example_test.go
vladimir-ch: Ok :-D

spatial/kdtree/kdtree_simple_example_test.go
vladimir-ch: Yes, I think it is clearer now.

spatial/kdtree/kdtree.go
vladimir-ch: No, I think it's fine to keep as it is.

spatial/kdtree/kdtree.go
vladimir-ch: Ok, thanks for the explanation.

spatial/kdtree/kdtree.go
kortschak: I changed it with the rest.

spatial/kdtree/kdtree.go
```diff
@@ -0,0 +1,469 @@
...
+	}
+
+	c := q.Compare(n.Point, n.Plane)
+	dist = math.Min(dist, q.Distance(n.Point))
+
+	bn := n
+	if c <= 0 {
+		ln, ld := n.Left.search(q, dist)
+		if ld < dist {
+			dist = ld
```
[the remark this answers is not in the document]
kortschak: I'll do that now.

kortschak:
PTAL

PR #976, opened 2019-04-30
spatial/barneshut: add Barnes-Hut force approximation package and R2/R3 helpers

kortschak:
The initial commit is essentially the same as the code that was merged into exp (barring grammatical fixes). The two later commits address the [comment](https://github.com/gonum/exp/pull/27#discussion_r279701191) that the vectors could go into their own packages.

Please take a look.

### DIFF
--- a/spatial/barneshut/barneshut2.go
+++ b/spatial/barneshut/barneshut2.go
@@ -11,59 +11,6 @@ import (
 	"gonum.org/v1/gonum/spatial/r2"
 )
 
-const (
-	ne = iota
-	se
-	sw
-	nw
-)
-
-// quadrantOf returns which quadrant of b that p should be placed in.
-func quadrantOf(b r2.Box, p Particle2) int {
-	center := r2.Vec{
-		X: (b.Min.X + b.Max.X) / 2,
-		Y: (b.Min.Y + b.Max.Y) / 2,
-	}
-	c := p.Coord2()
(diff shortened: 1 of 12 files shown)


spatial/barneshut/barneshut2.go
```diff
@@ -0,0 +1,249 @@
...
+
+const (
+	ne = iota
+	se
+	sw
+	nw
+)
+
+// quadrantOf returns which quadrant of b that p should be placed in.
+func quadrantOf(b r2.Box, p Particle2) int {
```
vladimir-ch: This could take `p` as `r2.Vec` at the price of less informative panic message.

spatial/barneshut/galaxy.go
```diff
@@ -0,0 +1,107 @@
+// Copyright ©2019 The Gonum Authors. All rights reserved.
```
vladimir-ch: Isn't this file committed by mistake?

kortschak: Yes. Thanks, I've removed it now.

spatial/barneshut/barneshut2.go
kortschak: That doesn't really save anything and I think the panic message will likely help users. `passDown` needs to hold a `Particle[23]` in order to insert it, so all we can do is hoist the `Coord[23]` call up one stack level.

vladimir-ch: No problem in leaving as is, I just wanted to read your comment.

PR #977, opened 2019-05-03
spatial/vptree: new package for vantage point tree NN search

kortschak:
Please take a look.

### DIFF
--- a/spatial/vptree/vptree.go
+++ b/spatial/vptree/vptree.go
@@ -115,14 +115,14 @@ func (b *builder) selectVantage(s []Comparable, effort int) Comparable {
 	var best Comparable
 	var bestVar float64
 	b.work = b.work[:effort]
-	for _, p := range b.random(effort, s) {
-		for i, q := range b.random(effort, s) {
+	choices := b.random(effort, s)
+	for _, p := range choices {
+		for i, q := range choices {
 			b.work[i] = p.Distance(q)
 		}
 		variance := stat.Variance(b.work, nil)
 		if variance > bestVar {
-			bestVar = variance
-			best = p
+			best, bestVar = p, variance
 		}
 	}
 	return best
--- a/spatial/vptree/vptree.go
+++ b/spatial/vptree/vptree.go
@@ -75,7 +75,7 @@ func New(p []Comparable, effort int, src rand.Source) *Tree {
 		intn = rnd.Intn
 		shuf = rnd.Shuffle
 	}
-	b := builder{work: make([]float64, max(len(p), effort)), intn: intn, shuf: shuf}
+	b := builder{work: make([]float64, len(p)), intn: intn, shuf: shuf}
 	return &Tree{
 		Root:  b.build(p, effort),
 		Count: len(p),
@@ -171,13 +171,6 @@ func (c byDist) Less(i, j int) bool { return c.dists[i] < c.dists[j] }
 func (c byDist) Swap(i, j int) {
 	c.dists[i], c.dists[j] = c.dists[j], c.dists[i]
 	c.points[i], c.points[j] = c.points[j], c.points[i]
-}
-
-func max(a, b int) int {
-	if a > b {
-		return a
-	}
-	return b
 }
 
 // Len returns the number of elements in the tree.
--- a/spatial/vptree/vptree.go
+++ b/spatial/vptree/vptree.go
@@ -1,0 +1,372 @@
+// Copyright ©2019 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package vptree
+
+import (
+	"container/heap"
+	"math"
+	"sort"
+
+	"golang.org/x/exp/rand"
+
+	"gonum.org/v1/gonum/stat"
+)
+
+// Comparable is the element interface for values stored in a vp tree.
+type Comparable interface {
+	// Distance returns the Euclidean distance between the receiver and
+	// the parameter.
+	Distance(Comparable) float64
+}
+
+// Point represents a point in a k-d space that satisfies the Comparable interface.
+type Point []float64
+
+// Distance returns the Euclidean distance between c and the receiver. The concrete
+// type of c must be Point.
+func (p Point) Distance(c Comparable) float64 {
+	q := c.(Point)
+	var sum float64
+	for dim, c := range p {
+		d := c - q[dim]
+		sum += d * d
+	}
+	return math.Sqrt(sum)
+}
+
+// Node holds a single point value in a vantage point tree.
+type Node struct {
+	Point   Comparable
+	Radius  float64
+	Closer  *Node
+	Further *Node
+}
+
+// Tree implements a vantage point tree creation and nearest neighbor search.
+type Tree struct {
+	Root  *Node
+	Count int
+}
+
+// New returns a vantage point tree constructed from the values in p. The effort
+// parameter specifies how much work should be put into optimizing the choice of
+// vantage point. If effort is one or less, random vantage points are chosen.
+// The order of elements in p will be altered after New returns.
+func New(p []Comparable, effort int) *Tree {
+	b := builder{work: make([]float64, max(len(p), effort*effort))}
+	return &Tree{
+		Root:  b.build(p, effort),
+		Count: len(p),
+	}
+}
+
+// builder performs vp-tree construction as described for the simple vp-tree
+// algorithm in http://pnylab.com/papers/vptree/vptree.pdf.
+type builder struct {
+	work []float64
+}
+
+func (b *builder) build(s []Comparable, effort int) *Node {
+	if len(s) <= 1 {
+		if len(s) == 0 {
+			return nil
+		}
+		return &Node{Point: s[0]}
+	}
+	n := Node{Point: b.selectVantage(s, effort)}
+	radius, closer, further := b.partition(n.Point, s)
+	n.Radius = radius
+	n.Closer = b.build(closer, effort)
+	n.Further = b.build(further, effort)
+	return &n
+}
+
+func (b *builder) selectVantage(s []Comparable, effort int) Comparable {
+	if effort <= 1 {
+		return s[rand.Intn(len(s))]
+	}
+	if effort > len(s) {
+		effort = len(s)
+	}
+	var best Comparable
+	var bestVar float64
+	b.work = b.work[:effort*effort]
+	for i, p := range random(effort, s) {
+		for j, q := range random(effort, s) {
+			b.work[i*effort+j] = p.Distance(q)
+		}
+		variance := stat.Variance(b.work, nil)
+		if variance > bestVar {
+			bestVar = variance
+			best = p
+		}
+	}
+	return best
+}
+
+func random(n int, s []Comparable) []Comparable {
+	if n >= len(s) {
+		return s
+	}
+	rand.Shuffle(len(s), func(i, j int) { s[i], s[j] = s[j], s[i] })
+	return s[:n]
+}
+
+func (b *builder) partition(v Comparable, s []Comparable) (radius float64, closer, further []Comparable) {
+	b.work = b.work[:len(s)]
+	for i, p := range s {
+		b.work[i] = v.Distance(p)
+	}
+	sort.Sort(byDist{dists: b.work, points: s})
+
+	// Note that this does not conform exactly to the description
+	// in the paper which specifies d(p, s) < mu for L; in cases
+	// where the median element has a lower indexed element with
+	// the same distance from the vantage point, L will include a
+	// d(p, s) == mu.
+	// The additional work required to satisfy the algorithm is
+	// not worth doing as it has no effect on the correctness or
+	// performance of the algorithm.
+	radius = b.work[len(b.work)/2]
+
+	if len(b.work) > 1 {
+		// Remove vantage if it is present.
+		closer = s[1 : len(b.work)/2]
+	}
+	further = s[len(b.work)/2:]
+	return radius, closer, further
+}
+
+type byDist struct {
+	dists  []float64
+	points []Comparable
+}
+
+func (c byDist) Len() int           { return len(c.dists) }
+func (c byDist) Less(i, j int) bool { return c.dists[i] < c.dists[j] }
+func (c byDist) Swap(i, j int) {
+	c.dists[i], c.dists[j] = c.dists[j], c.dists[i]
+	c.points[i], c.points[j] = c.points[j], c.points[i]
+}
+
+func max(a, b int) int {
+	if a > b {
+		return a
+	}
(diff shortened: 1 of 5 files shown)


kortschak:
This implements the optimised vp-tree algorithm from Yianilos, which says "An optimized tree results because function Select_vp endevours to pick better than random vantage points. Replacing it with a simple random selection generates a valid and effective tree; but experiments have shown that some effort towards making better choices, results in non-trivial search-time savings."

I don't see that.

```
Benchmark/NearestBruteN10:0-8         	     200	   6999116 ns/op
Benchmark/NearestBrute:0-8            	     300	   5947496 ns/op

Benchmark/Nearest:0-8                 	  300000	      5216 ns/op
Benchmark/Nearest:3-8                 	  300000	      5241 ns/op
Benchmark/Nearest:10-8                	  300000	      5173 ns/op
Benchmark/Nearest:30-8                	  300000	      5233 ns/op
Benchmark/Nearest:100-8               	  300000	      5258 ns/op
Benchmark/Nearest:300-8               	  300000	      5220 ns/op
Benchmark/NearestSetN10:0-8           	 1000000	      2207 ns/op
Benchmark/NearestSetN10:3-8           	 1000000	      2183 ns/op
Benchmark/NearestSetN10:10-8          	 1000000	      2151 ns/op
Benchmark/NearestSetN10:30-8          	 1000000	      2220 ns/op
Benchmark/NearestSetN10:100-8         	 1000000	      2176 ns/op
Benchmark/NearestSetN10:300-8         	 1000000	      2172 ns/op
```

With significant upfront costs for the additional work.
```
BenchmarkNew/New:0-8         	       5	 283258373 ns/op
BenchmarkNew/New:10-8        	       2	 875709497 ns/op
BenchmarkNew/New:100-8       	       1	7654383061 ns/op
```

Maybe the optimisation should be removed.

kortschak:
The other question is whether there should be a user hook for injecting the random seed; should `New` also take a `src *rand.Source`?

spatial/vptree/vptree.go
```diff
@@ -0,0 +1,372 @@
...
+	"sort"
+
+	"golang.org/x/exp/rand"
+
+	"gonum.org/v1/gonum/stat"
+)
+
+// Comparable is the element interface for values stored in a vp tree.
+type Comparable interface {
+	// Distance returns the Euclidean distance between the receiver and
```
vladimir-ch: If it's for general metric spaces, then the distance does not have to be Euclidean.

spatial/vptree/vptree.go
```diff
@@ -0,0 +1,372 @@
...
+		effort = len(s)
+	}
+	var best Comparable
+	var bestVar float64
+	b.work = b.work[:effort*effort]
+	for i, p := range random(effort, s) {
+		for j, q := range random(effort, s) {
+			b.work[i*effort+j] = p.Distance(q)
+		}
+		variance := stat.Variance(b.work, nil)
```
vladimir-ch: I think this should use the i-th row slice of `work`.

spatial/vptree/vptree.go
```diff
@@ -0,0 +1,372 @@
...
+	if n == nil {
+		return nil, inf
+	}
+
+	d := q.Distance(n.Point)
+	dist = math.Min(dist, d)
+
+	bn := n
+	if d < n.Radius {
+		if d-dist <= n.Radius {
```
vladimir-ch: This is equivalent to `if -dist <= n.Radius - d` but since `d < n.Radius` here, we are testing `if non-positive <= positive` which holds always. I find this suspicious.

spatial/vptree/vptree.go
```diff
@@ -0,0 +1,372 @@
...
+		}
+		if d+dist >= n.Radius {
+			fn, fd := n.Further.search(q, dist)
+			if fd < dist {
+				dist = fd
+				bn = fn
+			}
+		}
+	} else {
+		if d+dist >= n.Radius {
```
vladimir-ch: Similarly here.

spatial/vptree/vptree.go
```diff
@@ -0,0 +1,372 @@
...
+
+// Keep adds c to the heap if its distance is less than or equal to the max value of the heap.
+func (k *DistKeeper) Keep(c ComparableDist) {
+	if c.Dist <= k.Heap[0].Dist {
+		heap.Push(k, c)
+	}
+}
+
+// Keeper implements a conditional max heap sorted on the Dist field of the ComparableDist type.
+// kd search is guided by the distance stored in the max value of the heap.
```
vladimir-ch: What is `kd search` here? Should it be `vp search`?

spatial/vptree/vptree.go
```diff
@@ -0,0 +1,372 @@
...
+	if n == nil {
+		return
+	}
+
+	k.Keep(ComparableDist{Comparable: n.Point, Dist: q.Distance(n.Point)})
+
+	d := q.Distance(n.Point)
+	dist := k.Max().Dist
+	if d < n.Radius {
+		if d-dist <= n.Radius {
```
vladimir-ch: Also here and below.

spatial/vptree/vptree_simple_example_test.go
```diff
@@ -0,0 +1,60 @@
...
+		vptree.Point{7, 2},
+	}
+
+	// Print all points in the data set within 3 of (3, 5).
+	t := vptree.New(points, 0)
+	q := vptree.Point{3, 5}
+	t.Do(func(c vptree.Comparable, _ int) (done bool) {
+		// Compare each distance and output points
+		// with a Euclidean distance less than 3.
+		// Distance returns the square of the
```
vladimir-ch: Copy/paste from `kdtree`? There is no square here.

spatial/vptree/vptree.go
kortschak: Fixed

spatial/vptree/vptree.go
kortschak: I'm an idiot.

spatial/vptree/vptree.go
kortschak: Fixed

spatial/vptree/vptree_simple_example_test.go
kortschak: Fixed

spatial/vptree/vptree.go
kortschak: I'll look into this further

kortschak:
Now with the corrected selection calculation.
```
BenchmarkNew/New:0-8         	       5	 247633792 ns/op
BenchmarkNew/New:10-8        	       2	 770851644 ns/op
BenchmarkNew/New:100-8       	       1	3911120178 ns/op

Benchmark/NearestBruteN10:0-8         	     300	   5913136 ns/op
Benchmark/NearestBrute:0-8            	     300	   4935473 ns/op

Benchmark/Nearest:0-8                 	  300000	      4802 ns/op
Benchmark/Nearest:3-8                 	  300000	      4774 ns/op
Benchmark/Nearest:10-8                	  300000	      4590 ns/op
Benchmark/Nearest:30-8                	  300000	      4444 ns/op
Benchmark/Nearest:100-8               	  300000	      4366 ns/op
Benchmark/Nearest:300-8               	  300000	      4303 ns/op

Benchmark/NearestSetN10:0-8           	 1000000	      1971 ns/op
Benchmark/NearestSetN10:3-8           	 1000000	      1949 ns/op
Benchmark/NearestSetN10:10-8          	 1000000	      1971 ns/op
Benchmark/NearestSetN10:30-8          	 1000000	      1949 ns/op
Benchmark/NearestSetN10:100-8         	 1000000	      1988 ns/op
Benchmark/NearestSetN10:300-8         	 1000000	      1951 ns/op
```

spatial/vptree/vptree.go
kortschak: Yes. The code came from algorithm 3 which includes the bounds which provide a better control of which branches to explore. The condition on the first option for each case can be omitted.

kortschak: Interestingly this condition is in all the reproductions of this algorithm.

1. [Yianilos](http://pnylab.com/papers/vptree/vptree.pdf)
2. [Fu _et al_](https://citeseerx.ist.psu.edu/viewdoc/download;jsessionid=477FE81DBEDC30B6A9240512C9CCF76D?doi=10.1.1.115.8253&rep=rep1&type=pdf).

and [this blog](http://stevehanov.ca/blog/index.php?id=130) and the remarkably similar [code here](https://github.com/lvdmaaten/bhtsne/blob/master/vptree.h#L229-L270).

spatial/vptree/vptree.go
```diff
@@ -0,0 +1,385 @@
...
+// Comparable is the element interface for values stored in a vp-tree.
+type Comparable interface {
+	// Distance returns the distance between the receiver and the
+	// parameter. The returned distance must satisfy the properties
+	// of distances in a metric space.
+	//
+	// - a.Distance(a) == 0
+	// - a.Distance(b) >= 0
+	// - a.Distance(b) == b.Distance(a)
+	// - a.Distamce(b) <= a.Distamce(c)+c.Distance(b)
```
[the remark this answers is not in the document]
kortschak: Fixed

kortschak:
Yeah. I guess it depends on how many times the tree will be used. I'm going to do some experiments to see when/if it pays off. Preliminarily, tree size makes makes a difference, but the benchmarks timeouted before I could see how much.

kortschak:
It looks to me like it is extremely unlikely to pay off.

Here is a set of benchmarks with 1e4, 1e5 and 1e6 elements in the tree. For the 1e4 and 1e5 I have made unoptimised selection and perfectly optimised selection, for the 1e6 I have made unoptimised and optimisation with 100 samples because perfect optimisation is unfeasible at that scale.

We can see that for the 1e6 case, we would need to make ~5e6 queries to break even with the time investment, for 1e4 and 1e5 it is 15e4 to 14e6 and 6e5 to 5e8 queries respectively. In each case you're making more queries than there are elements in the tree.

There are some optimisations to make, but not really in the optimisation parts of the vp selection routines.

```
1e4 elements

BenchmarkNew/Newx10000:0-8                  	     100	  17065619 ns/op
BenchmarkNew/Newx10000:100-8                	      20	  60613299 ns/op
BenchmarkNew/Newx10000:10000-8              	       1	3153260301 ns/op
Benchmark/Nearestx10000:0-8                 	  500000	      2152 ns/op
Benchmark/Nearestx10000:100-8               	 1000000	      1870 ns/op
Benchmark/Nearestx10000:10000-8             	 1000000	      1926 ns/op
Benchmark/NearestSetN10x10000:0-8           	  100000	     15988 ns/op
Benchmark/NearestSetN10x10000:100-8         	  100000	     13808 ns/op
Benchmark/NearestSetN10x10000:10000-8       	  100000	     14567 ns/op

1e5 elements

BenchmarkNew/Newx100000:0-8                 	       5	 259904726 ns/op
BenchmarkNew/Newx100000:100-8               	       2	 682755984 ns/op
BenchmarkNew/Newx100000:100000-8                       1	595475559603 ns/op
Benchmark/Nearestx100000:0-8                	  200000	      5580 ns/op
Benchmark/Nearestx100000:100-8              	  300000	      4822 ns/op
Benchmark/Nearestx100000:100000-8           	  300000	      4489 ns/op
Benchmark/NearestSetN10x100000:0-8          	   50000	     36974 ns/op
Benchmark/NearestSetN10x100000:100-8        	   50000	     34467 ns/op
Benchmark/NearestSetN10x100000:100000-8     	   50000	     28500 ns/op

1e6 elements

BenchmarkNew/Newx1000000:0-8                       1	4480506641 ns/op
BenchmarkNew/Newx1000000:100-8                     1	8981545826 ns/op
Benchmark/Nearestx1000000:0-8                 200000	      9194 ns/op
Benchmark/Nearestx1000000:100-8               200000	      8378 ns/op
Benchmark/NearestSetN10x1000000:0-8            30000	     51336 ns/op
Benchmark/NearestSetN10x1000000:100-8          30000	     44197 ns/op
```

It's difficult to know whether this is expected, since Yianilos doesn't give any evidence in the paper for the performance benefits and with one exception (UnixJunkie/vp-tree) there is no implementation I can find that does the extra work (that is an OCaml implementation that also does the bounds in a way that is similar to vps). I have written to @UnixJunkie to see if he sees a worthwhile benefit.

I think that we should probably merge this as it is to keep the code in the history, but then back out the optimisation code, unless there is something I'm completely missing (also the rand source, since I think we can just always seed with 1 since there is no real benefit beyond repeatable execution to being able to specify the PRNG seed).

UnixJunkie:
If you construct trees often, choosing vantage points randomly should be fine.
If you construct the tree once, but then query it many times, it might be worth choosing the VPs
in a smarter way.

Recently, I use bisector-trees instead of VP-trees in production, for static spatial indexing.
CF. https://github.com/UnixJunkie/bisec-tree
The "vantage points" selection heuristic is called two_bands in the code.
I think I saw it in a paper by Sergei Brin and called "greedy algorithm". It looks like a heuristic way
to find a double band of the point cloud (a diameter is always a double band, but the reverse is not true).
https://github.com/UnixJunkie/bisec-tree/blob/master/bisec_tree.ml
Choose one point randomly (call it A), then find the farthest from it (call it B), then the farthest from B (call it C).
Use B as your "vantage point" for the left sub-tree and C for the right sub-tree.

Note that benchmarks are just benchmarks, you would need point clouds from an actual problem
to be really sure (number of dimensions of the points, number of points, number of queries per constructed tree, actual distance function being used, etc).

kortschak:
Thanks for the input. The trouble that I'm having is seeing room for an improvement (heeding caveat on benchmarks) that would pay for the cost of calculation of optimalish vantage points. The pay off (from random data) looks like it's query count of about 10x the number of points stored in the tree. This is an unlikely use case.

UnixJunkie:
Some machine learning methods use nearest-neighbor searches a lot.
Indexing the training set well is really worthy in some situations.

kortschak:
Thanks. What has your experience been with the bisector trees cf vp trees?

UnixJunkie:
I don't have benchmarks, but:
i) my BST implementation is bucketized; so it is more advanced than my VPT implementation.
Bucket size higher than one allows the tree to use less memory, and buckets could
be fetched from disk on demand, for very large datasets.
Note that having buckets of size one is equivalent to a non bucketized BST.
ii) I did not understand in Yianilos' paper how to do that with VPTs (despites he explains and draws
an example).
iii) BSTs have a theoretical advantage in high dimensions: they split the space more equally at each step. Sorry, I don't remember where I read about this.

vladimir-ch:
Any conclusion from this, @kortschak ?

kortschak:
The conclusion in am email was that it can make a difference in some machine learning contexts and that we should leave it up to the user. On the basis of that, this is ready.

Issue #768, opened 2018-12-15
stat/mds: ~0 Eigenvalues returned by TorgersonScaling are problematic


kortschak:
### What are you trying to do?
Run `mds.TorgersonScaling` on the R eurodist sample dataset and get reasonable agreement with the R output.

### What did you do?

Run the code at https://play.golang.org/p/z7kPeRXDuAI both with `-tags noasm` and without tags.

### What did you expect to happen?
Output of Eigenvalue similar to the output given by R.

```
$eig
 [1]  1.953838e+07  1.185656e+07  1.528844e+06  1.118742e+06  7.893472e+05
 [6]  5.816552e+05  2.623192e+05  1.925976e+05  1.450845e+05  1.079673e+05
[11]  5.139484e+04 -3.259629e-09 -9.496124e+03 -5.305820e+04 -1.322166e+05
[16] -2.573360e+05 -3.326719e+05 -5.162523e+05 -9.191491e+05 -1.006504e+06
[21] -2.251844e+06
```

### What actually happened?
Two different values for Eigenvalue 12, both of which are greater than 0 where the R-returned value is less than 0.
```
~ $ go run main.go 

  1.953838e+07  1.185656e+07  1.528844e+06  1.118742e+06  7.893472e+05
  5.816552e+05  2.623192e+05  1.925976e+05  1.450845e+05  1.079673e+05
  5.139484e+04  7.598369e-11 -9.496124e+03 -5.305820e+04 -1.322166e+05
 -2.573360e+05 -3.326719e+05 -5.162523e+05 -9.191491e+05 -1.006504e+06
 -2.251844e+06
~ $ go run -tags noasm main.go 

  1.953838e+07  1.185656e+07  1.528844e+06  1.118742e+06  7.893472e+05
  5.816552e+05  2.623192e+05  1.925976e+05  1.450845e+05  1.079673e+05
  5.139484e+04  1.348610e-10 -9.496124e+03 -5.305820e+04 -1.322166e+05
 -2.573360e+05 -3.326719e+05 -5.162523e+05 -9.191491e+05 -1.006504e+06
 -2.251844e+06
```

This is pretty clearly an issue with small values, but there is also an issue of difference between the ASM and pure-Go implementations.

While the issue appears trivial, with MDS using Torgerson Scaling the choice of k, the maximum dimension of the space which the data are to be can represented in, is based on the number of Eigenvalues above zero, so in this case we estimate k one greater than does R (R's result looks correct when the coordinates are examined).

[...]

PR #981, opened 2019-05-12
mat: add Rank function

yuichiro12:
Discussions are here. Main reason to add this is that rank is often used in linear algebra.
https://groups.google.com/forum/#!topic/gonum-dev/m1bE98FYv24

@vladimir-ch @btracey
Would you mind reviewing this?


.gitignore
```diff
@@ -0,0 +1 @@
+.idea
```
kortschak: Please remove this file.

yuichiro12: I removed it. Is there any policy not to have .gitignore in this repository.

kortschak: Thanks

If there are things that get constructed by tests that should not be committed then we use them (this happens in plot), otherwise no. They would collect all the various editer and build environment cruft from contributors and we don't want that.

mat/matrix_test.go
```diff
@@ -528,6 +528,31 @@ func TestNormZero(t *testing.T) {
 	}
 }
 
+func TestRank(t *testing.T) {
```
vladimir-ch: Please make this a table-driven test like
```
tests := []struct{
    a    *Dense
    eps  float64
    want int
} {
    // ...
}

for i, test := range tests {
    // ...
}
```
and add more test cases: wide, tall, square matrices of various sizes including trivial 1x1, 1x2, ... sizes, with full and deficient rank, diagonal matrices with a small element and tolerance that is smaller and larger that this element (so that the rank is different depending on epsilon), .... Each test case can also additionally test whether `Rank(a, .) == Rank(a.T(), .)`.

.gitignore
yuichiro12: That make sense.

kortschak:
Do we want to look into this approach? https://math.berkeley.edu/~mgu/MA273/Strong_RRQR.pdf

mat/matrix.go
```diff
@@ -846,6 +846,21 @@ func normLapack(norm float64, aTrans bool) lapack.MatrixNorm {
...
 }
 
+// Rank returns rank of input matrix using singular value decomposition.
+// Give epsilon to determine singular values as zero.
+func Rank(a Matrix, epsilon float64) int {
+	var svd SVD
+	svd.Factorize(a, SVDNone)
+	sv := svd.Values(nil)
+	for i, v := range sv {
+		if v < epsilon {
```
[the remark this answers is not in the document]
kortschak: Yes, sorry, this was my error.

vladimir-ch:
We already have Algorithm 1 of https://math.berkeley.edu/~mgu/MA273/Strong_RRQR.pdf as `lapack/gonum.Dgeqp3` but according to this paper it is known to fail for some matrices. Algorithm 5 should remedy this but without a reference implementation I don't feel like going there. Compared to SVD we could gain speed but with this numerical rank stuff (for a given matrix mapping from _continuous_ epsilon to _discrete_ rank) I think it's safer to use something reliable, that is, SVD. Matlab is doing the same.

kortschak:
Fair enough. Maybe add a TODO to keep this as an option in the future?

vladimir-ch:
@yuichiro12 , could you please add the TODO comment with the link that @kortschak mentioned? Let us know if there's anything unclear about my comments or if you need help. The additional test cases can also be postponed to another PR.

mat/matrix.go
```diff
@@ -846,6 +846,21 @@ func normLapack(norm float64, aTrans bool) lapack.MatrixNorm {
 	}
 }
 
+// Rank returns rank of input matrix using singular value decomposition.
```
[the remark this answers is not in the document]
yuichiro12: Thank you for reviewing. Forgive me for slow working.

> If epsilon is negative, Rank will just return min(m, n). This will behave as documented but it will not return the rank of A (if different from min(m,n)). Do we want to allow this or panic instead?

I added panic for negative epsilon for safety. Thank you for catching.

> Do we want to replace zero epsilon with some default value? I guess it is difficult to have an intuition for choosing a proper value of epsilon, so it could be a convenient tool which on the other hand adds (small) complexity that would have to be documented.

Replacing default value is a good idea (in my original implementation, epsilon was global variable to have default value). Since epsilon is often used as positive (and small) number, replacing zero-epsilon with default value may not be counterintuitive.
Could you please review additional commit?

mat/matrix_test.go
yuichiro12: I make it table driven. Please wait for additional test cases.

mat/matrix.go
```diff
@@ -846,6 +846,21 @@ func normLapack(norm float64, aTrans bool) lapack.MatrixNorm {
...
+func Rank(a Matrix, epsilon float64) int {
+	var svd SVD
+	svd.Factorize(a, SVDNone)
+	sv := svd.Values(nil)
+	for i, v := range sv {
+		if v < epsilon {
+			return i
+		}
+	}
+	_, c := a.Dims()
```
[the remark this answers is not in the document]
yuichiro12: 😅

mat/matrix.go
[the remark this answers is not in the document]
yuichiro12: May be my epsilon is too large?

yuichiro12:
I added TODO comment.

yuichiro12:
- [ ] 1x1
- [ ] 1x2
- [ ] wide
- [ ] tall (transpose of wide)
- [ ] square-full
- [ ] square-deficient
- [ ] diagonal with a small element (for epsilon)

mat/matrix.go
```diff
@@ -846,6 +846,28 @@ func normLapack(norm float64, aTrans bool) lapack.MatrixNorm {
...
 }
 
+// Rank returns rank of input matrix using singular value decomposition.
+// Give non-negative epsilon to determine singular values as zero.
+// If input epsilon is exactly 0, it is replaced with 1e-10, the default value.
+func Rank(a Matrix, epsilon float64) int {
+	if epsilon < 0 {
+		panic(ErrNegativeEpsilon)
+	}
+	if epsilon == 0 {
```
kortschak: I would think dlamchE or 2*dlamchE would be appropriate if this approach is being used (this I'm not convinced of though).

yuichiro12: Is that related to SVD implementation or something? (I'm not quite into it yet...)

kortschak: https://github.com/gonum/gonum/blob/a5f8f3a4840b76492571a7beed7958b420800458/lapack/gonum/lapack.go#L38-L39

However, I'm not still not convinced that doing this magically is the correct thing to do.

vladimir-ch: I put the suggestion of a default value for the zero epsilon as a question, indicating that I also wasn't sure whether it's a good idea. Seeing it now in the code I'm not convinced either.

mat/matrix_test.go
```diff
@@ -528,6 +528,111 @@ func TestNormZero(t *testing.T) {
...
+			want:    4,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if got := Rank(tt.a, tt.epsilon); got != tt.want {
+				t.Errorf("Rank() = %v, want %v", got, tt.want)
+			}
+			if got := Rank(tt.a.T(), tt.epsilon); got != tt.want {
+				t.Errorf("Rank() = %v, want %v", got, tt.want)
```
vladimir-ch: Please indicate somehow in the error message that it is the comparison with `a.T()` that fails.

yuichiro12: fixed

mat/matrix.go
yuichiro12: I personally tested a little more: https://play.golang.org/p/V7HW_RAZxUr
It generates random float value matrix with given rank and apply this Rank() function.
Seeing returned singular value vectors, `1.0 / (1 << 53)` is too small and `1e-10` is enough for most cases. (you can change const values to test various cases)

yuichiro12: @vladimir-ch where can I see your suggestion?

kortschak: Please remove this.

kortschak: I should add, in [stat/mds](https://godoc.org/gonum.org/v1/gonum/stat/mds#TorgersonScaling) we have a similar problem [from a similar cause](https://github.com/gonum/gonum/issues/768). The approach there is to return the slice of Eigenvalues so the user can make an informed choice.

We already have the sigma vector, so perhaps we should just return that as well. @vladimir-ch @btracey?

glaslos:
@yuichiro12 do you have plans to continue on this feature?

vladimir-ch:
I'm closing this PR because it has gone stale and because in its current form it merely duplicates the functionality of `SVD.Rank`. We're still interested in contributions, so please create a new PR if you still intend to submit this change.

Issue #956, opened 2019-04-14
bound: consider renaming Bound to Interval


kortschak:
Merely to reduce stutter.

vladimir-ch:
Or rename it to `r1` and move to `spatial`.

kortschak:
SGTM

btracey:
I guess I don't like `r1`, but r2 and r3 are there, as is https://godoc.org/github.com/golang/geo/r1  , so I'm probably wrong (i.e. I'm not going to veto it). It's not clear that `r1.Bound` makes sense though. `r1.Interval` seems to be better, but it's a lot of characters. Maybe `r1.Range`?

kortschak:
`Range` has more semantic content than `Interval`. Considering https://github.com/gonum/gonum/pull/982#issuecomment-491822888, I'd push for `Interval`.

PR #982, opened 2019-05-13
spatial/r1: import and rename from bound

sbinet:
Mechanically generated with:

```
$> gomvpkg -from gonum.org/v1/gonum/bound -to gonum.org/v1/gonum/spatial/r1 -vcs_mv_cmd "git mv {{.Src}} {{.Dst}}"
```

Fixes gonum/gonum#956.

Please take a look.

### DIFF
--- a/spatial/r1/bound.go
+++ b/spatial/r1/bound.go
@@ -4,7 +4,7 @@
 
 package r1
 
-// Bound represents [Min, Max] bounds.
-type Bound struct {
+// Interval represents [Min, Max] bounds.
+type Interval struct {
 	Min, Max float64
 }
--- a/spatial/r1/bound.go
+++ b/spatial/r1/bound.go
@@ -2,5 +2,9 @@
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
-// Package bound provides types for bounded data.
-package bound // import "gonum.org/v1/gonum/bound"
+package r1
+
+// Bound represents [Min, Max] bounds.
+type Bound struct {
+	Min, Max float64
+}
--- a/spatial/r1/doc.go
+++ b/spatial/r1/doc.go
@@ -2,5 +2,5 @@
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
-// Package r1 provides 1D vectors and boxes and operations on them.
+// Package r1 provides 1D vectors and intervals and operations on them.
 package r1 // import "gonum.org/v1/gonum/spatial/r1"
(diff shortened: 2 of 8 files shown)


sbinet:
ah... perhaps hold this one off before https://github.com/gonum/gonum/pull/969 has been sorted out and/or merged in?

kortschak:
We should probably wait. Also, what do you think of `r1.Interval`? The reasoning for the question is that a bound is a semantic layer on top of an interval.

sbinet:
I am fine with `s/Bound/Interval/` (probably with a `gorename`.)
I'll wait for `#969` to be collected.

vladimir-ch:
> ... bound is a semantic layer on top of an interval

I also thought about this and this may also be where the confusion about `bound.Union` comes from. I like how `r2.Box` and `r3.Box` provide only the type for storing two points without implying any semantics (is the boundary included, what is an empty box, does it bound anything, ...).

kortschak:
Yes, I'm beginning to regret https://github.com/gonum/gonum/pull/845#discussion_r255359176. I think perhaps those additions should be rolled back.

kortschak:
@sbinet The PRs arising from #969 have been reverted, so this can go ahead now. Do you want to do that?

sbinet:
Sure. I'll update this PR shortly

sbinet:
PTAL

spatial/r1/doc.go
```diff
@@ -2,5 +2,5 @@
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
-// Package bound provides types for bounded data.
-package bound // import "gonum.org/v1/gonum/bound"
+// Package r1 provides 1D vectors and boxes and operations on them.
```
kortschak: s/boxes/intervals/.

spatial/r1/bound.go
```diff
@@ -2,9 +2,9 @@
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
-package bound
+package r1
 
-// Bound represents [Min, Max] bounds.
-type Bound struct {
+// Interval represents [Min, Max] bounds.
```
kortschak: @vladimir-ch Do we want to be this definitive? We use them as [min, max] bounds, but they could be any type of real interval.

vladimir-ch: I would prefer leaving it unspecified, it was the main motivation behind these changes. But what will we put there instead? `Interval is an interval`? :-)

kortschak: Simple, direct and to the point. SGTM ("Interval is an interval in R1" seems like over cooking.)

sbinet:
PTAL

PR #983, opened 2019-05-15
github: add note about using play.golang.org

kortschak:
Please take a look.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -11,7 +11,10 @@ of the problem, prefixed by the primary affected package.
 
 
 ### What did you do?
-<!-- Please include a link to a minimal reproducer here. -->
+<!--
+Please include a link to a minimal reproducer here. Preferably this should
+be a complete runnable example at https://play.golang.org/.
+-->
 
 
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -11,7 +11,10 @@ of the problem, prefixed by the primary affected package.
 
 
 ### What did you do?
-<!-- Please include a link to a minimal reproducer here. -->
+<!--
+Please include a link to a minimal reproducer here. Preferably this should
+be a complete runnable example at https://play.golang.org/.
```
sbinet: we should also perhaps require they display the result of `gonum.Version()`. (perhaps with a link to a pre-shared template?)

kortschak: I'm not sure how stable that will be in the current implementation. I note though that in the issue associated with the CL that added foreign imports that there is discussion about versioning. Probably it would be best to wait until that is decided.

sbinet: I was merely advocating for replacing the link to `https://play.golang.org` with that kind of link:
https://play.golang.org/p/OqLpL0_kyV5

but ok.

PR #984, opened 2019-05-15
ci: use GOPROXY from proxy.golang.org

sbinet:
Using this proxy allows to push and then use the latest Gonum version
from the playground.

Please take a look.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,7 @@ os:
 env:
  global:
   - GO111MODULE=on
+  - GOPROXY="https://proxy.golang.org"
   - GOFLAGS="-mod=readonly"
  matrix:
   - TAGS=""


kortschak:
Can you explain (link to?) how this works?

sbinet:
this is an implementation of the proxy for Go modules (like PyPI for python, more or less.)
some details:
 https://proxy.golang.org/

the part about the playground may have probably been just me being a bit fooled by the caching mechanism of that service... (I was trying to use `plot.Version()` but the playground wasn't picking a version of `gonum/plot` with that new API in. I thought that was because perhaps only Go-HEP was pulling an old version of `gonum/plot` inside the cache of proxy.golang.org...)

I'll amend the commit message.

testing `GOPROXY` seems like a good idea, though.
right?

kortschak:
It looks to me like this is a consumer-related environment variable. I don't see how it's relevant here.

sbinet:
"""
The Go team is providing the following services run by Google: a module mirror for accelerating Go module downloads, an index for discovering new modules, and a global go.sum database for authenticating module content.
"""

ISTM like an important part of the (still nascent) Go modules ecosystem.
Gonum should engage with that.

being able to list modules (and versions) on a global scale sounds as useful as godoc+search:

- https://proxy.golang.org/gonum.org/v1/plot/@v/list
- https://proxy.golang.org/gonum.org/v1/gonum/@v/list
- https://proxy.golang.org/go-hep.org/x/hep/@v/list

that said, GoHEP is already using proxy.golang.org as its `GOPROXY`, so all the Gonum and Gonum/Plot versions that GoHEP is using (or has been using and will be using) will be in the proxy.golang.org instance, so I can close this PR if you prefer.

kortschak:
Sure, I just don't understand how setting the var here does anything; there's nothing in the page at proxy.golang.org that says that setting the var will cause an artifact to be cached there. Maybe that does happen, but the docs don't mention it.

kortschak:
ISTM that if the behaviour that this change seems to intend is wanted, then the approach should be to run `go get gonum.org/v1/gonum@version` from outside the repo when a PR has been merged in to master (so this would just be added on to the build tasks that run code coverage). However, this just makes sure the version gets in to the cache earlier; it should get in eventually anyway.

From the proxy.golang.org FAQ:

>In order to improve our services' caching and serving latencies, new versions may not show up right away. If you want new code to be immediately available in the mirror, then first make sure there is a semantically versioned tag for this revision in the underlying source repository. Then explicitly request that version via `go get module@version`. After one minute for caches to expire, the go command will see that tagged version.

sbinet:
Go 1.14 has this setting by default (IIRC).

Abandoned.

PR #985, opened 2019-05-16
mat: relax restrictions on vector orientation

### LINKED ISSUE #388: mat: as a default, don't require vectors to conform to a specific shape
See discussion in https://github.com/gonum/gonum/pull/378

kortschak:
Please take a look.

If I've missed any please let me know.

Fixes #388.

### DIFF
--- a/mat/dense_arithmetic.go
+++ b/mat/dense_arithmetic.go
@@ -760,10 +760,10 @@ func (m *Dense) Apply(fn func(i, j int, v float64) float64, a Matrix) {
 	}
 }
 
-// RankOne performs a rank-one update to the matrix a with the vectors x
-// and y, where x is treated as a column vector and y is treated as a row
-// vector. The result is stored in the receiver. If a is zero, see Outer.
-//  m = a + alpha * x * y
+// RankOne performs a rank-one update to the matrix a with the vectors x and
+// y, where x and y are treated as column vectors. The result is stored in the
+// receiver. If a is zero, see Outer.
+//  m = a + alpha * x * y^T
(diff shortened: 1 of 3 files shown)


mat/dense_arithmetic.go
```diff
@@ -758,17 +758,16 @@ func (m *Dense) Apply(fn func(i, j int, v float64) float64, a Matrix) {
 	}
 }
 
-// RankOne performs a rank-one update to the matrix a and stores the result
-// in the receiver. If a is zero, see Outer.
-//  m = a + alpha * x * y'
+// RankOne performs a rank-one update to the matrix a with the vectors x
+// and y, where x is treated as a column vector and y is treated as a row
+// vector. The result is stored in the receiver. If a is zero, see Outer.
+//  m = a + alpha * x * y
```
vladimir-ch: For the mathematician in me this unadorned notation is confusing. I would preserve the mathematical notation here (only perhaps written with `^T` instead of `'`) and instead document that both vectors are considered to be (or treated as) column vectors regardless of their actual orientation.

kortschak: I don't have strong opinions either way, but I'll wait for @btracey to chime in.

kortschak:
There are additional changes here that are needed to make shadow detection correct; striding is not correctly handled as it stands.

This is a note to myself.

kortschak:
PTAL

btracey:
Sorry that I've been absent, I've been traveling the last few weeks. I'm
almost caught up on the backlog

PR #932, opened 2019-03-28
mat: implement helper routines for type extraction and update Trace to use an interface

btracey:
… an interface.

Updates #929.
Updates #692 

Please take a look.

PR #989, opened 2019-05-17
mat: generalise vector mul fast path

kortschak:
Please take a look.

### DIFF
--- a/mat/dense_arithmetic.go
+++ b/mat/dense_arithmetic.go
@@ -347,29 +347,6 @@ func (m *Dense) Mul(a, b Matrix) {
 			blas64.Trmm(blas.Right, bT, 1, bmat, m.mat)
 			return
 
-		case *VecDense:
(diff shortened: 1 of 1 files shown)


kortschak:
I'll wait also for @btracey since I've removed the `*VecDense` special case. This reduced some near exact code repetition but will make that type a tiny smidge slower.

btracey:
I think instead of this we want to be using `untransposeExtract` a la https://github.com/gonum/gonum/pull/932 . The main idea being that there are a lot of `Raw` types that don't conform to the assumptions of our Dense types (negative strides, Diag representations, etc.). `untransposeExtract` allows us to switch over the Dense types while also allowing external representations that implement the `RawFoo` types

kortschak:
Do you want to send that?

kortschak:
GitHub should have closed this, but Closes comments don't seem to work except in the first comment.

PR #995, opened 2019-06-06
graph/layout: new package for calculating graph layouts

kortschak:
Please take a look.

This is a first pass at doing something like this. The Eades algorithm is simple and shows how the framework can be exercised. I'm not at all weded to the overall approach I have here, so any input is welcome.


graph/layout/eades_test.go
```diff
@@ -0,0 +1,205 @@
...
+package layout
+
+import (
+	"bytes"
+	"os/exec"
+	"path/filepath"
+	"reflect"
+	"testing"
+
+	"gonum.org/v1/gonum/graph/encoding/dot"
```
sbinet: regroup with the other `gonum/graph` imports?

graph/layout/layout_test.go
```diff
@@ -0,0 +1,120 @@
...
+	"image"
+	"image/png"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"sort"
+	"strings"
+	"testing"
+
+	"gonum.org/v1/plot/cmpimg"
```
sbinet: regroup with the other `gonum` imports?

graph/layout/eades_test.go
```diff
@@ -0,0 +1,205 @@
...
+				}
+				layout.SetEdge(simple.Edge{
+					F: positionNode{id: u.ID(), pos: got[u.ID()].Scale(scaling)},
+					T: positionNode{id: v.ID(), pos: got[v.ID()].Scale(scaling)},
+				})
+			}
+		}
+		b, _ := dot.Marshal(layout, test.name, "", "  ")
+		path := filepath.Join("testdata", test.name+".png")
+		cmd := exec.Command("neato", "-n", "-Tpng", "-o", path)
```
sbinet: perhaps document somehow (and somewhere) that `neato` comes with graphviz?
(and perhaps also test for that command and disable the test if it's not available?)

graph/layout/eades_test.go
```diff
@@ -0,0 +1,205 @@
...
+		for nodes.Next() {
+			id := nodes.Node().ID()
+			got[id] = o.Location(id)
+		}
+		if !reflect.DeepEqual(got, test.want) {
+			t.Errorf("unexpected result:\ngot: %#v\nwant:%#v", got, test.want)
+		}
+
+		// TODO(kortschak): Replace this with Go rendering code.
+		if _, err := exec.LookPath("neato"); err != nil {
```
sbinet: perhaps document somehow (and somewhere) that `neato` comes with graphviz?

graph/layout/eades_test.go
```diff
@@ -0,0 +1,205 @@
...
+				})
+			}
+		}
+		b, _ := dot.Marshal(layout, test.name, "", "  ")
+		path := filepath.Join("testdata", test.name+".png")
+		cmd := exec.Command("neato", "-n", "-Tpng", "-o", path)
+		cmd.Stdin = bytes.NewReader(b)
+		out, err := cmd.CombinedOutput()
+		if err != nil {
+			t.Errorf("unexpected error: %v\n%s", err, out)
```
sbinet: `s/t.Errorf/t.Fatalf/`

go.sum
```diff
@@ -1,6 +1,20 @@
...
+golang.org/x/exp v0.0.0-20180321215751-8460e604b9de/go.mod h1:CJ0aWSM057203Lf6IL+f9T1iT9GByDxfZKAQTCR3kQA=
+golang.org/x/exp v0.0.0-20180807140117-3d87b88a115f/go.mod h1:CJ0aWSM057203Lf6IL+f9T1iT9GByDxfZKAQTCR3kQA=
 golang.org/x/exp v0.0.0-20190125153040-c74c464bbbf2 h1:y102fOLFqhV41b+4GPiJoa0k/x+pJcEi2/HB1Y5T6fU=
 golang.org/x/exp v0.0.0-20190125153040-c74c464bbbf2/go.mod h1:CJ0aWSM057203Lf6IL+f9T1iT9GByDxfZKAQTCR3kQA=
+golang.org/x/image v0.0.0-20180708004352-c73c2afc3b81 h1:00VmoueYNlNz/aHIilyyQz/MHSqGoWJzpFv/HW8xpzI=
+golang.org/x/image v0.0.0-20180708004352-c73c2afc3b81/go.mod h1:ux5Hcp/YLpHSI86hEcLt0YII63i6oz57MZXIpbrjZUs=
+golang.org/x/tools v0.0.0-20180525024113-a5b4c53f6e8b/go.mod h1:n7NCudcB/nEzxVGmLbDWY5pfWTLqBcC2KZ6jyYvM4mQ=
 golang.org/x/tools v0.0.0-20190206041539-40960b6deb8e h1:Io7mpb+aUAGF0MKxbyQ7HQl1VgB+cL6ZJZUFaFNqVV4=
 golang.org/x/tools v0.0.0-20190206041539-40960b6deb8e/go.mod h1:n7NCudcB/nEzxVGmLbDWY5pfWTLqBcC2KZ6jyYvM4mQ=
+gonum.org/v1/gonum v0.0.0-20180816165407-929014505bf4/go.mod h1:Y+Yx5eoAFn32cQvJDxZx5Dpnq+c3wtXuadVZAcxbbBo=
```
sbinet: this will probably be a bit confusing...
perhaps we could just extract `gonum.org/v1/plot/cmpimg` out of `gonum/plot` and put it somewhere else?
or just vendor it?

graph/layout/eades_test.go
kortschak: Done

graph/layout/layout_test.go
kortschak: Done

graph/layout/eades_test.go
kortschak: This is already disabled if `neato` is not in `$PATH`. It's only temporary until I implement a plotter for graphs anyway.

graph/layout/eades_test.go
kortschak: This is only temporary.

graph/layout/eades_test.go
kortschak: Not fatal, but I have added a `continue`.

go.sum
kortschak: The tests will fully depend on gonum/plot when I'm done.

kortschak:
*ping*

vladimir-ch:
This has not slipped under my radar, give me a couple of days.

graph/layout/eades.go
```diff
@@ -0,0 +1,155 @@
...
+	for i, p := range u.particles {
+		f := plane.ForceOn(p, u.Theta, barneshut.Gravity2).Scale(-u.C3)
+		// Prevent marginal updates that can be caused by
+		// floating point error when nodes are very far apart.
+		if math.Hypot(f.X, f.Y) > 1e-12 {
+			updated = true
+		}
+		u.forces[i] = f
+	}
+	if !updated {
```
vladimir-ch: What is the justification for this early return? If the nodes are too far apart, there could be attractive forces computed below. And why not having this test in the attraction loop below - if the nodes have reached an equilibrium before M runs out, Update should return false.

graph/layout/eades.go
```diff
@@ -0,0 +1,155 @@
...
+	u.nodes.Reset()
+
+	// Apply global repulsion.
+	plane := barneshut.NewPlane(u.particles)
+	var updated bool
+	for i, p := range u.particles {
+		f := plane.ForceOn(p, u.Theta, barneshut.Gravity2).Scale(-u.C3)
+		// Prevent marginal updates that can be caused by
+		// floating point error when nodes are very far apart.
+		if math.Hypot(f.X, f.Y) > 1e-12 {
```
vladimir-ch: Just a comment: a threshold of what a "steady configuration" is will most likely depend of the values of `C?` parameters but it seems to me that for practical plotting purposes an equilibrium can be declared for forces several orders greater than `1e-12`. Maybe make the threshold relative to the layout's current extent? Later.

graph/layout/eades.go
```diff
@@ -0,0 +1,155 @@
...
+			yid := to.Node().ID()
+			if seen[[2]int64{xid, yid}] {
+				continue
+			}
+			seen[[2]int64{yid, xid}] = true
+			yidx := u.indexOf[yid]
+
+			// Apply adjacent node attraction.
+			v := u.particles[yidx].Coord2().Sub(u.particles[xidx].Coord2())
+			f := v.Scale(weight(xid, yid) / 2 * u.C1 * math.Log(math.Hypot(v.X, v.Y)))
```
vladimir-ch: This does not apply `C2`.

graph/layout/eades.go
kortschak: I have moved this below the adjacent nodes force calculations and added conditional mutation of `updated` to that loop.

graph/layout/eades.go
kortschak: Ack.

graph/layout/eades.go
kortschak: Fixed.

kortschak:
PTAL

kortschak:
A couple of questions come out of the Eades implementation.

1. Should we give the constants better names or just leave them as they are in the paper?
2. Should we provide R2 and R3 implementations? How do we do the namespacing?

vladimir-ch:
> Should we give the constants better names or just leave them as they are in the paper?

I would reduce the number of constants to 2 and then given them better names. Let me explain.

The layout algorithm is a minimization of an objective function (a potential energy of a spring system). The algorithm is a kind of gradient descent where the computed force points in the direction of the negative gradient. Stepping in the direction of the force we step in the direction of a local minimum. The force in the paper has the form
```
C1*log(d/C2) - C3/sqrt(d)
```
This notation is of course very crude, the terms are actually sums over the right nodes and the direction vectors are not in there.

There is also C4 which is a step size for the gradient descent algorithm.

I would say that for graph layout/plotting we don't care about the scale of the layout as long as the relative positions of nodes are the same. Actually, in the end some rescaling will have to take place anyway to fit the layout into the canvas extent. In other words, we don't care about about the ideal edge length. Therefore we can set `C2=1`. If we already had C1 and C3 given, then by setting `C3'=C3/sqrt(C2)` we would get the exact same layout (only rescaled). So we are at
```
C1*log(d) - C3'/sqrt(d)
```
Then we factor out C1 (or C3) to get `C1*[ log(d) - C3/(C1*sqrt(d)) ]`. Then C4 can absorb C1: `C4'=C4*C1`, and `C3''=C3'/C1`. So we end up with
```
log(d) - C3''/sqrt(d)
```
and with the step size `C4'`. I think we could come up with a reasonable default value for the algorithmic parameter C4, leaving C3 as the only modeling parameter, balancing the attracting and repulsing terms and so the overall layout of nodes.

For the repulsion the code uses the Barnes-Hut approximation but I think it should not change anything.

What do you think?

kortschak:
I think this sounds pretty reasonable. When I was implementing it, I was surprised by the presence of C4.

I'll have a go at this later today.

kortschak:
If I remove C1, C2 and C4 (keeping the scaling at 0.1) then I get exactly the same output as with the current code, so it seems reasonably robust to C1 and C2. If I don't scale `f` then the thing explodes. I suspect that we want to keep a gradient decent rate parameter here in some form, perhaps as a power so that zero generally just works.

kortschak:
I have just sent a PR that adds a recoverable panic to barneshut that means that we can recover a panic by the `NewPlane` call and then return false if the points are too far apart.

vladimir-ch:
>  If I don't scale f then the thing explodes. I suspect that we want to keep a gradient decent rate parameter here in some form, perhaps as a power so that zero generally just works.

Yes, I agree we need to keep `Rate` tunable. We could also start reusing here all the work done on adaptive stepsize/learning rate selection for stochastic gradient descent because this is basically what we have here except that instead of choosing the components randomly they are picked consecutively. Even though this is "just" graph layout, it's apparently not straightforward to do it robustly.

graph/layout/eades.go
```diff
@@ -0,0 +1,171 @@
...
+}
+
+// Update is the EadesR2 spatial graph update function.
+func (u *EadesR2) Update(g graph.Graph, layout LayoutR2) bool {
+	if u.Updates <= 0 {
+		return false
+	}
+	u.Updates--
+
+	if !layout.IsInitialized() {
```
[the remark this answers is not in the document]
kortschak: Yes. It's pretty common for initial layouts to be one of random in a unit space (as is done here - gephi does something similar, but for reasons of java it doesn't run here at the moment though it did last week with no changes between then and now except maybe the jdk I have installed being updated), or layout out the nodes on a grid randomly with respect to connections. It would be possible to prearrange using Torgerson scaling, but that would make this particularly expensive for no real gain.

I think I'll just do the spreading more. Gephi uses 50x50 for random layouts, so maybe just start with that.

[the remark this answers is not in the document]
kortschak: I've just tried that and I don't think it's a good idea. Choosing a variety of other starting spreads gives bad layouts. This is not a justification for using this arbitrary unit random layout for this restricted set of input graphs, but rather a justification for not putting any real effort into making this layout work for all cases. The Eades layout here is really just to exercise the framework to see if it handles the kinds of things we need to do and worksish. Also note that with the current design it is possible to perform an initial random layout by writing an update function that assigns positions to nodes and then replacing the `Updater` value after that initial conditioning. I don't know if this is expecting too much of people.

I'm going to leave it as is for the moment.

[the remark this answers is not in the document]
vladimir-ch: > I'm going to leave it as is for the moment.

Sure, I just wanted to finish commenting on this branch of the discussion.

> I don't know if this is expecting too much of people.

That's quite a hidden capability and not very elegant either, I'll think more about it when I've finished reading optimizer.go.

graph/layout/eades.go
```diff
@@ -0,0 +1,171 @@
...
+	for u.nodes.Next() {
+		xid := u.nodes.Node().ID()
+		xidx := u.indexOf[xid]
+		to := g.From(xid)
+		for to.Next() {
+			yid := to.Node().ID()
+			if seen[[2]int64{xid, yid}] {
+				continue
+			}
+			seen[[2]int64{yid, xid}] = true
```
vladimir-ch: Should this have `{xid, yid}` or is it indended?

graph/layout/eades.go
```diff
@@ -0,0 +1,171 @@
...
+
+// EadesR2 implements the graph layout algorithm essentially as
+// described in "A heuristic for graph drawing", Congressus
+// numerantium 42:149-160.
+// The implementation here uses the Barnes-Hut approximation for
+// global repulsion calculation, and edge weights are considered
+// when calculating adjacent node attraction.
+type EadesR2 struct {
+	// Updates is the number of updates to perform.
+	Updates int
```
vladimir-ch: For me this does not fit here, the caller will be calling Update in a loop anyway, so they can control how many updates are performed. Optimizer looks like a better place? Or are you including it here because you don't want to assume all layout algorithms being iterative? But in that case why pass them through Optimizer? I don't want to overthink this and replicate what `optimize` does (although on the other hand, it almost looks like this could be adapted to use `optimize` :-) ).

graph/layout/optimizer.go
```diff
@@ -0,0 +1,106 @@
...
+
+import (
+	"gonum.org/v1/gonum/graph"
+	"gonum.org/v1/gonum/spatial/r2"
+)
+
+// LayoutR2 implements graph layout updates and representations.
+type LayoutR2 interface {
+	// IsInitialized returns whether the Layout is initialized.
+	IsInitialized() bool
```
vladimir-ch: Why the need for this?

graph/layout/eades.go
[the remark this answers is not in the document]
kortschak: >That's quite a hidden capability and not very elegant either, I'll think more about it when I've finished reading optimizer.go.

Mmmm. It has precedence in the standard library and it handles nicely a variety of layout problems that arise where an initial conditioning layout is needed and then a polishing layout is applied. I can't see a way of achieving that without a functionality like this.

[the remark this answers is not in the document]
vladimir-ch: Why not initialize the layout before starting the Optimizer.Update loop? I'm afraid I'm looking at it from a wrong angle.

graph/layout/eades.go
kortschak: This is intended. We mark the next thing orientation we will see since we won't see this orientation again.

graph/layout/eades.go
kortschak: This is indeed here because there are layout algorithms that do not do iterative updates. A layout that I want to include is minumum edge crossing circular layout (e.g. [here](https://www.sciencedirect.com/science/article/pii/S1570866705000031/pdfft?md5=9012fb96c356e2a2611e6f9d32bb6ece&pid=1-s2.0-S1570866705000031-main.pdf). The reason to pass them through optimiser is to allow a general API for constructing layouts. The user does not need to know the type approach taken.

graph/layout/optimizer.go
kortschak: This allows the updater to know whether layout needs to be started. Without it we need to perform the intialisation prior to the loop.

graph/layout/plotter_test.go
```diff
@@ -0,0 +1,132 @@
...
+
+	"gonum.org/v1/plot"
+	"gonum.org/v1/plot/plotter"
+	"gonum.org/v1/plot/vg"
+	"gonum.org/v1/plot/vg/draw"
+)
+
+const radius = vg.Length(15)
+
+// render implements the plot.Plotter interface for graphs.
```
vladimir-ch: Do you plan to export this in some form later?

kortschak: Yes, but it needs to have more user control over the rendering options. That needs a fair amount of thought.

PR #998, opened 2019-06-12
modules: fix expected language version directive

kortschak:
Please take a look.

After discussion [here](https://groups.google.com/d/topic/golang-nuts/JNVpE6ONzPU/discussion) this seems to be correct.


kortschak:
Thanks. It doesn't make a great deal of difference at this stage. When language changes become more substantial it will.

PR #999, opened 2019-06-12
lapack/gonum: add Dpbtrf

vladimir-ch:
Please take a look.

### DIFF
--- a/lapack/gonum/dpbtf2.go
+++ b/lapack/gonum/dpbtf2.go
@@ -47,9 +47,9 @@ import (
 // version.
 //
 // Dpbtf2 is an internal routine, exported for testing purposes.
-func (Implementation) Dpbtf2(ul blas.Uplo, n, kd int, ab []float64, ldab int) (ok bool) {
+func (Implementation) Dpbtf2(uplo blas.Uplo, n, kd int, ab []float64, ldab int) (ok bool) {
 	switch {
-	case ul != blas.Upper && ul != blas.Lower:
+	case uplo != blas.Upper && uplo != blas.Lower:
 		panic(badUplo)
 	case n < 0:
 		panic(nLT0)
@@ -59,6 +59,7 @@ func (Implementation) Dpbtf2(ul blas.Uplo, n, kd int, ab []float64, ldab int) (o
 		panic(badLdA)
 	}
 
+	// Quick return if possible.
 	if n == 0 {
 		return
 	}
@@ -70,16 +71,17 @@ func (Implementation) Dpbtf2(ul blas.Uplo, n, kd int, ab []float64, ldab int) (o
 	bi := blas64.Implementation()
 
 	kld := max(1, ldab-1)
-	if ul == blas.Upper {
+	if uplo == blas.Upper {
+		// Compute the Cholesky factorization A = U^T * U.
 		for j := 0; j < n; j++ {
-			// Compute U(J,J) and test for non positive-definiteness.
+			// Compute U(j,j) and test for non-positive-definiteness.
 			ajj := ab[j*ldab]
 			if ajj <= 0 {
 				return false
 			}
 			ajj = math.Sqrt(ajj)
 			ab[j*ldab] = ajj
-			// Compute elements j+1:j+kn of row J and update the trailing submatrix
+			// Compute elements j+1:j+kn of row j and update the trailing submatrix
 			// within the band.
 			kn := min(kd, n-j-1)
 			if kn > 0 {
@@ -89,16 +91,16 @@ func (Implementation) Dpbtf2(ul blas.Uplo, n, kd int, ab []float64, ldab int) (o
 		}
 		return true
 	}
+	// Compute the Cholesky factorization A = L * L^T.
 	for j := 0; j < n; j++ {
-		// Compute L(J,J) and test for non positive-definiteness.
+		// Compute L(j,j) and test for non-positive-definiteness.
 		ajj := ab[j*ldab+kd]
 		if ajj <= 0 {
 			return false
 		}
 		ajj = math.Sqrt(ajj)
 		ab[j*ldab+kd] = ajj
-
-		// Compute elements J+1:J+KN of column J and update the trailing submatrix
+		// Compute elements j+1:j+kn of column j and update the trailing submatrix
 		// within the band.
 		kn := min(kd, n-j-1)
 		if kn > 0 {
--- a/lapack/gonum/dpbtrf.go
+++ b/lapack/gonum/dpbtrf.go
@@ -1,0 +1,214 @@
+// Copyright ©2019 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package gonum
+
+import (
+	"gonum.org/v1/gonum/blas"
+	"gonum.org/v1/gonum/blas/blas64"
+)
+
+// Dpbtrf computes the Cholesky factorization of an n×n symmetric positive
+// definite band matrix
+//  A = U^T * U  if uplo == blas.Upper
+//  A = L * L^T  if uplo == blas.Lower
+// where U is an upper triangular band matrix and L is lower triangular. kd is
+// the number of super- or sub-diagonals of A.
+//
+// The band storage scheme is illustrated below when n = 6 and kd = 2. Elements
+// marked * are not used by the function.
+//
+//  uplo == blas.Upper
+//  On entry:         On return:
+//   a00  a01  a02     u00  u01  u02
+//   a11  a12  a13     u11  u12  u13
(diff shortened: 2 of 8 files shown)


lapack/gonum/dpbtf2.go
```diff
@@ -89,16 +91,16 @@ func (Implementation) Dpbtf2(ul blas.Uplo, n, kd int, ab []float64, ldab int) (o
 		}
 		return true
 	}
+	// Compute the Cholesky factorization A = L * L^T.
 	for j := 0; j < n; j++ {
-		// Compute L(J,J) and test for non positive-definiteness.
+		// Compute L(j,j) and test for non-positive-definiteness.
```
kortschak: We don't conventionally write hyphenation like this. I think we are wrong, so please leave it as is. (Just by way of a note).

lapack/gonum/ilaenv.go
```diff
@@ -165,12 +165,12 @@ func (impl Implementation) Ilaenv(ispec int, name string, opts string, n1, n2, n
 				panic(badName)
 			case "TRF":
 				if sname {
-					if n4 <= 64 {
+					if n2 <= 64 {
```
kortschak: How did you find this?

lapack/testlapack/general.go
```diff
@@ -1030,21 +1030,22 @@ func equalApproxSymmetric(a, b blas64.Symmetric, tol float64) bool {
...
 	a := make([]float64, n*n)
 	for i := 0; i < n; i++ {
-		for j := max(0, i-kb); j <= i; j++ {
+		for j := max(0, i-kd); j < i; j++ {
 			a[i*n+j] = rnd.NormFloat64()
 		}
-		a[i*n+i] = math.Abs(a[i*n+i])
-		// Add an extra amound to the diagonal in order to improve the condition number.
-		a[i*n+i] += 1.5 * rnd.Float64()
+		a[i*n+i] = 1 + math.Abs(rnd.NormFloat64())
```
kortschak: Why this change?

lapack/gonum/dpbtrf.go
```diff
@@ -0,0 +1,188 @@
...
+// license that can be found in the LICENSE file.
+
+package gonum
+
+import (
+	"gonum.org/v1/gonum/blas"
+	"gonum.org/v1/gonum/blas/blas64"
+)
+
+func (impl Implementation) Dpbtrf(uplo blas.Uplo, n, kd int, ab []float64, ldab int) (ok bool) {
```
kortschak: Doc comment?

lapack/gonum/dpbtrf.go
```diff
@@ -0,0 +1,188 @@
...
+		// storage.
+
+		// Process the band matrix one diagonal block at a time.
+		for i := 0; i < n; i += nb {
+			ib := min(nb, n-i)
+			// Factorize the diagonal block.
+			ok := impl.Dpotf2(uplo, ib, ab[i*ldab:], ldab-1)
+			if !ok {
+				return false
+
```
kortschak: Delete blank line.

lapack/gonum/dpbtrf.go
```diff
@@ -0,0 +1,188 @@
...
+			}
+			if i+ib >= n {
```
kortschak: `>`? The Fortran has `i+ib.LE.n`. Is this a 1 versus 0?

lapack/gonum/dpbtrf.go
```diff
@@ -0,0 +1,188 @@
...
+			//
+			// The numbers of rows and columns in the partitioning
+			// are ib, i2, i3 respectively. The blocks A12, A22 and
+			// A23 are empty if ib = kd. The upper triangle of A13
+			// lies outside the band.
+			i2 := min(kd-ib, n-i-ib)
+			if i2 > 0 {
+				// Update A12.
+				bi.Dtrsm(blas.Left, blas.Upper, blas.Trans, blas.NonUnit, ib, i2,
+					1, ab[i*ldab:], ldab-1, ab[i*ldab+ib:], ldab-1)
```
kortschak: The absence of `kd` is row versus column major?

lapack/gonum/dpbtrf.go
```diff
@@ -0,0 +1,188 @@
...
+	} else {
+		// Compute the Cholesky factorization of a symmetric band
+		// matrix, given the lower triangle of the matrix in band
+		// storage.
+
+		// Process the band matrix one diagonal block at a time.
+		for i := 0; i < n; i += nb {
+			ib := min(nb, n-i)
+			// Factorize the diagonal block.
+			ok := impl.Dpotf2(uplo, ib, ab[i*ldab+kd:], ldab-1)
```
kortschak: And presence here and in the equivalent above for the same reason.

lapack/testlapack/dpbtrf.go
```diff
@@ -0,0 +1,123 @@
...
+					case blas.Upper:
+						for i := 0; i < n; i++ {
+							for j := 0; j < min(kd+1, n-i); j++ {
+								diff = math.Max(diff, math.Abs(abFac[i*ldab+j]-ab[i*ldab+j]))
+							}
+						}
+					case blas.Lower:
+						for i := 0; i < n; i++ {
+							for j := max(0, i-kd); j <= i; j++ {
+								// diff = math.Max(diff, math.Abs(got[i*n+j]-abCopy[i*ldab+kd-i+j]))
```
kortschak: Is there here for a reason?

lapack/gonum/dpbtf2.go
vladimir-ch: Ok. The changes in this file are unnecessary but since I was going through it anyway, I thought I could do some minimal cleanup.

lapack/gonum/ilaenv.go
vladimir-ch: Because the test was passing on the first try. Impossible, I thought. Then I found out that the blocked path was not taken because `Ilaenv` was always returning block size of 1 no matter how large the matrix/bandwidth was.

lapack/testlapack/general.go
vladimir-ch: For no particular reason. I was playing with different ways of generating the matrix and this was what stayed. I'll put the original code back although it doesn't really matter. This is again the problem of generating the test matrices. It would be nice to have the complete implementation of `DLATMS` which can also generate band matrices but `DLATMS` is ugly and complex and a lot of work.

lapack/gonum/dpbtf2.go
kortschak: Sorry. That was really unclear. Leave it how you have it.

lapack/gonum/dpbtrf.go
vladimir-ch: Yes, 1 vs 0 and the inverted condition. Another way to see it, `i+ib` is the _index_ where the next block would begin. Maybe it would be clearer to break out of the loop here and remove the end test in the for loop above. I just didn't want to have two apparent places to end the loop.

lapack/gonum/dpbtrf.go
vladimir-ch: Yes. There are two ways to translate Fortran banded to Go banded code:
1. The manual one: In Fortran, in case of Upper, the first index select the diagonal. `kd+1` is the main diagonal, `kd+1-1` is the first super-diagonal, etc. In general, `kd+1-d` is the d-th super-diagonal. So if we rewrite the row index in this form, we get our column index. For example, `1+IB = KD+1+IB-KD = KD+1-(KD-IB)`, so our column index would be `d = kd-ib`. To get the row index, we must subtract `d` from the Fortran column index. For example, `I+KD-d = I+KD-(KD-IB) = I+IB`. So the final index is `(i+ib)*ldab+kd-ib`.
2. The semi-manual one: In case of Upper, take Fortran code for Lower, swap row/column indices and convert to 0-based.

2 is of course less work but you have take care of sizes, sidedness, and transpose, which must match the actual uplo, not the uplo from which you are taking the indices. Overall, it's quite unpleasant and annoying.

lapack/testlapack/dpbtrf.go
vladimir-ch: I don't know how this got here, maybe a github glitch? I had it in the code at some point but fixed before pushing. And I don't remember pushing to the branch after creating the PR and yet it is already marked as Outdated. Either way, it's not in the code now.

lapack/gonum/dpbtrf.go
vladimir-ch: Omg, how could I have forgotten this? :-) Added now.

lapack/testlapack/general.go
vladimir-ch: Done.

vladimir-ch:
PTAL

PR #1001, opened 2019-06-16
lapack/gonum: add Dpbtrs

vladimir-ch:
Please take a look.

### DIFF
--- a/lapack/gonum/dpbtrs.go
+++ b/lapack/gonum/dpbtrs.go
@@ -1,0 +1,62 @@
+// Copyright ©2019 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package gonum
+
+import (
+	"gonum.org/v1/gonum/blas"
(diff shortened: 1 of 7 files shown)


lapack/gonum/dpbtrs.go
```diff
@@ -0,0 +1,62 @@
...
+
+import (
+	"gonum.org/v1/gonum/blas"
+	"gonum.org/v1/gonum/blas/blas64"
+)
+
+// Dpbtrs solves a system of linear equations A*X = B with a symmetric positive
+// definite band matrix A using the Cholesky factorization A = U^T * U or
+// A = L * L^T  computed by Dpbtrf.
+func (Implementation) Dpbtrs(uplo blas.Uplo, n, kd, nrhs int, ab []float64, ldab int, b []float64, ldb int) {
```
kortschak: Is this intended for external use? If it is, don't we normally have a little more documentation on valid entry conditions?

vladimir-ch: I've extended the documentation a bit. PTAL

PR #1009, opened 2019-06-27
spatial/barneshut: catch finitely too far points

kortschak:
It turns out the previous check was not strict enough. When the quadrants and octants are being subdeivided, if the box is large enough the midpoint will be numerically identical to one of the bounds. This results in an infinite recursion down the tree, overflowing the stack.

The approach here is to restrict the area/volume to within sqrt(maxfloat) of the origin as a heuristic and then check for non-progression during tree construction.

Please take a look.

### DIFF
--- a/spatial/barneshut/barneshut2.go
+++ b/spatial/barneshut/barneshut2.go
@@ -5,6 +5,7 @@
 package barneshut
 
 import (
+	"errors"
 	"fmt"
 	"math"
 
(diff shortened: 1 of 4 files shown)


kortschak:
This is still not right; we can get to tree construction and fail. I'll look further.

spatial/barneshut/barneshut2.go
```diff
@@ -97,15 +94,28 @@ func (q *Plane) Reset() {
 		}
 	}
 
+	defer func() {
+		switch r := recover(); r {
+		case nil:
+		case volumeTooBig:
```
kortschak: Oops. This should have been `planeTooBig`. I'll fix it in the layout PR.

Issue #1019, opened 2019-07-20
graph/iterator: add reflect.MapIter iterator implementation


kortschak:
See https://github.com/gonum/gonum/pull/567#issuecomment-420810893 when go1.11 is no longer supported.

Also an unsafe equivalent.

PR #1020, opened 2019-07-20
graph/iterator: add unordered node iterator

kortschak:
Please take a look.

This is mainly a reminder for me to ensure that I implement the remainder and the unsafe version when I can. In the meantime it can be used by clients using go1.12+ in their implementations.

Updates #1019.

### DIFF
--- a/graph/iterator/nodes_unordered.go
+++ b/graph/iterator/nodes_unordered.go
@@ -1,0 +1,63 @@
+// Copyright ©2018 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// +build go1.12
+
+// TODO(kortschak): Merge this into nodes.go when go1.11 is no longer supported.
+
+package iterator
+
+import (
+	"reflect"
+
+	"gonum.org/v1/gonum/graph"
+)
+
+// Nodes implements the graph.Nodes interfaces.
+// The iteration order of Nodes is randomized.
+type Nodes struct {
+	nodes reflect.Value
+	iter  *reflect.MapIter
+	pos   int
+	curr  graph.Node
+}
+
+// NewNodes returns a Nodes initialized with the provided nodes, a
+// map of node IDs to graph.Nodes. No check is made that the keys
+// match the graph.Node IDs, and the map keys are not used.
+//
+// Behavior of the Nodes is unspecified if nodes is mutated after
+// the call the NewNodes.
+func NewNodes(nodes map[int64]graph.Node) *Nodes {
+	rv := reflect.ValueOf(nodes)
+	return &Nodes{nodes: rv, iter: rv.MapRange()}
+}
+
+// Len returns the remaining number of nodes to be iterated over.
+func (n *Nodes) Len() int {
+	return n.nodes.Len() - n.pos
+}
+
+// Next returns whether the next call of Node will return a valid node.
+func (n *Nodes) Next() bool {
+	ok := n.iter.Next()
+	if ok {
+		n.pos++
+		n.curr = n.iter.Value().Interface().(graph.Node)
+	}
+	return ok
+}
+
+// Node returns the current node of the iterator. Next must have been
+// called prior to a call to Node.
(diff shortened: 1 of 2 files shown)


kortschak:
This is the unsafe version demonstrated. I don't know how portable it is (it won't run on the playground).

```
package main

import (
	"fmt"
	"unsafe"

	_ "reflect" // For map iteration helpers.

	"gonum.org/v1/gonum/graph"
)

// unpackEface converts the empty interface i to an emptyInterface.
func unpackEface(i interface{}) *emptyInterface {
	return (*emptyInterface)(unsafe.Pointer(&i))
}

// emptyInterface is the header for an interface{} value.
type emptyInterface struct {
	typ  *rtype
	word unsafe.Pointer
}

// rtype is the common implementation of most values.
// It is embedded in other struct types.
//
// rtype must be kept in sync with ../runtime/type.go:/^type._type.
type rtype struct {
	size       uintptr
	ptrdata    uintptr  // number of bytes in the type that can contain pointers
	hash       uint32   // hash of type; avoids computation in hash tables
	tflag      tflag    // extra type information flags
	align      uint8    // alignment of variable with this type
	fieldAlign uint8    // alignment of struct field with this type
	kind       uint8    // enumeration for C
	alg        *typeAlg // algorithm table
	gcdata     *byte    // garbage collection data
	str        nameOff  // string form
	ptrToThis  typeOff  // type for pointer to this type, may be zero
}

// tflag is used by an rtype to signal what extra type information is
// available in the memory directly following the rtype value.
//
// tflag values must be kept in sync with copies in:
//	cmd/compile/internal/gc/reflect.go
//	cmd/link/internal/ld/decodesym.go
//	runtime/type.go
type tflag uint8

// a copy of runtime.typeAlg
type typeAlg struct {
	// function for hashing objects of this type
	// (ptr to object, seed) -> hash
	hash func(unsafe.Pointer, uintptr) uintptr
	// function for comparing objects of this type
	// (ptr to object A, ptr to object B) -> ==?
	equal func(unsafe.Pointer, unsafe.Pointer) bool
}

type nameOff int32 // offset to a name
type typeOff int32 // offset to an *rtype

// m escapes into the return value, but the caller of mapiterinit
// doesn't let the return value escape.
//go:linkname mapiterinit reflect.mapiterinit
//go:noescape
func mapiterinit(t *rtype, m unsafe.Pointer) unsafe.Pointer

//go:linkname mapiterkey reflect.mapiterkey
//go:noescape
func mapiterkey(it unsafe.Pointer) (key unsafe.Pointer)

//go:linkname mapiterelem reflect.mapiterelem
//go:noescape
func mapiterelem(it unsafe.Pointer) (elem unsafe.Pointer)

//go:linkname mapiternext reflect.mapiternext
//go:noescape
func mapiternext(it unsafe.Pointer)

type mapIter struct {
	m  *emptyInterface
	it unsafe.Pointer
}

func newMapIter(m map[int64]graph.Node) *mapIter {
	return &mapIter{m: unpackEface(m)}
}

// Next advances the map iterator and reports whether there is another
// entry. It returns false when the iterator is exhausted; subsequent
// calls to Value, or Next will panic.
func (it *mapIter) Next() bool {
	if it.it == nil {
		it.it = mapiterinit(it.m.typ, it.m.word)
	} else {
		if mapiterkey(it.it) == nil {
			panic("mapIter.Next called on exhausted iterator")
		}
		mapiternext(it.it)
	}
	return mapiterkey(it.it) != nil
}

// Value returns the value of the iterator's current map entry.
func (it *mapIter) Value() graph.Node {
	if it.it == nil {
		panic("mapIter.Value called before Next")
	}
	if mapiterkey(it.it) == nil {
		panic("mapIter.Value called on exhausted iterator")
	}
	return *(*graph.Node)(mapiterelem(it.it))
}

func main() {
	m := map[int64]graph.Node{1: node{"one", 1}, 3: node{"three", 3}}
	it := newMapIter(m)
	for it.Next() {
		fmt.Printf("%T %#[1]v\n", it.Value())
	}
}

type node struct {
	name string
	id   int64
}

func (n node) ID() int64 { return n.id }
```

PR #1023, opened 2019-07-20
travis: allow coverage scripts to run

kortschak:
We have not been running test coverage analysis for the past few month due to missing `x` bits from when I made the multi-OS testing system. This fixes that.

Please take a look.


kortschak:
Note that nothing will show up in the push or pr logs. The effect is only on the post-merge master branch.

PR #1025, opened 2019-07-22
travis: only run coverage when secure and on master

kortschak:
This got lost in the multi-os testing system.

Please take a look.

TBR


kortschak:
This was merged prior to review because of the absence of the secure vars test. It doesn't work on master (though it did previously). I will look into that further.

PR #1026, opened 2019-07-22
graph/simple: fix a bug

yinyanghu:
Please take a look.


kortschak:
Ooop. That's embarrassing. This exists in a number of other places as well. I'll send a fix for them all.

PR #1031, opened 2019-07-23
graph/testgraph: test for missing From and To paths

### LINKED ISSUE #1030: graph/testgraph: tests do not adequately cover To methods
See #1026 and #1027.

In the case of a node with no inbound edges, the optimisation that was present prior to merging #1027 was merely ineffective, but in the case of  a node with no outbound but some inbound, the `To` method would incorrectly return `graph.Empty`. Since the tests previously passed, this case is presumably not properly covered. I assume the the original bug in #1026 was found through this behaviour, though that was not reported and no issue was filed.

The test that is defective is `testgraph.ReturnAdjacentNodes`. Is does test for over reporting nodes, but does not test for missed nodes. There should be an iteration over the complete list of edges and confirmation that they are properly reported by the graph. This is not currently done.

kortschak:
It turns out the implementations were correctly behaving, but I think this was more fortuitous than by design.

I'm interested to know how @yinyanghu found the bug. I don't think it was from the behaviour, but if it was, can you show me the case that was failing and I will add it here.

Please take a look.

Fixes #1030.

### DIFF
--- a/graph/testgraph/testcases.go
+++ b/graph/testgraph/testcases.go
@@ -160,6 +160,31 @@ var testCases = []struct {
 	},
 
 	{
+		name:  "three in only",
+		nodes: []graph.Node{node(0), node(1), node(2), node(3)},
+		edges: []WeightedLine{
+			line{F: node(1), T: node(0), UID: 0, W: 0.5},
+			line{F: node(2), T: node(0), UID: 1, W: 0.5},
+			line{F: node(3), T: node(0), UID: 2, W: 0.5},
+		},
+		nonexist: []graph.Node{node(-1), node(4)},
+		self:     0,
+		absent:   math.Inf(1),
+	},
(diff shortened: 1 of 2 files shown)


yinyanghu:
FYI, I found this bug by looking at the code. I'm trying to use gonum/graph as a general graph framework in my own repository [watermill](https://github.com/yinyanghu/watermill). It's good to see that you are adding more unit tests here. I think I would like to look at the code under gonum/graph a lot, and I would let you know if I find more issues.

kortschak:
Thanks. It turns out that when an edge is added, both the `from` and `to` edge maps are made for both nodes. So even if a node has no edges leading to it, that condition will be true (making it currently non-useful). I will look into improving that - at least to see if it is a benefit. It looks like there is, and a new bug.

PR #1035, opened 2019-07-23
lapack/gonum: add Dlatbs

vladimir-ch:
This is one of the uglier lapack routines, especially in combination with the band storage.
The coverage is almost complete, so hopefully this translation is also correct.

There are some differences from the Fortran version that I'll be happy to explain. Almost all are done to make the code clearer. I left one `goto` to not increase the indentation further.

Please take a look.

### DIFF
--- a/lapack/gonum/dlatbs.go
+++ b/lapack/gonum/dlatbs.go
@@ -31,7 +31,7 @@ import (
 // non-trivial solution to A*x = 0 is returned.
 //
 // Dlatbs is an internal routine. It is exported for testing purposes.
-func (Implementation) Dlatbs(uplo blas.Uplo, trans blas.Transpose, diag blas.Diag, normin bool, n, kd int, ab []float64, ldab int, x []float64, cnorm []float64) (scale float64) {
+func (Implementation) Dlatbs(uplo blas.Uplo, trans blas.Transpose, diag blas.Diag, normin bool, n, kd int, ab []float64, ldab int, x, cnorm []float64) (scale float64) {
 	noTran := trans == blas.NoTrans
 	switch {
 	case uplo != blas.Upper && uplo != blas.Lower:
@@ -298,7 +298,8 @@ skipComputeGrow:
 
 			// Scale x if necessary to avoid overflow when adding a multiple of
 			// column j of A.
-			if xj > 1 {
+			switch {
+			case xj > 1:
 				rec := 1 / xj
 				if cnorm[j] > (bignum-xMax)*rec {
 					// Scale x by 1/(2*abs(x[j])).
@@ -306,7 +307,7 @@ skipComputeGrow:
 					bi.Dscal(n, rec, x, 1)
 					scale *= rec
 				}
-			} else if xj*cnorm[j] > bignum-xMax {
+			case xj*cnorm[j] > bignum-xMax:
 				// Scale x by 1/2.
 				bi.Dscal(n, 0.5, x, 1)
 				scale *= 0.5
--- a/lapack/gonum/dlatbs.go
+++ b/lapack/gonum/dlatbs.go
@@ -398,39 +398,41 @@ skipComputeGrow:
 				// if 1/A[j,j] was not used to scale the dot product.
 				x[j] -= sumj
 				xj = math.Abs(x[j])
-				if diag == blas.NonUnit {
-					// Compute x[j] = x[j] / A[j,j], scaling if necessary.
-					switch {
-					case tjj > smlnum:
-						// smlnum < abs(A[j,j]):
-						if tjj < 1 && xj > tjj*bignum {
-							// Scale x by 1/abs(x[j]).
-							rec := 1 / xj
-							bi.Dscal(n, rec, x, 1)
-							scale *= rec
-							xMax *= rec
-						}
-						x[j] /= tjjs
-					case tjj > 0:
-						// 0 < abs(A[j,j]) <= smlnum:
-						if xj > tjj*bignum {
-							// Scale x by (1/abs(x[j]))*abs(A[j,j])*bignum.
-							rec := (tjj * bignum) / xj
-							bi.Dscal(n, rec, x, 1)
-							scale *= rec
-							xMax *= rec
-						}
-						x[j] /= tjjs
-					default:
-						// A[j,j] == 0: Set x[0:n] = 0, x[j] = 1, and scale = 0, and
-						// compute a solution A^T * x = 0.
-						for i := range x[:n] {
-							x[i] = 0
-						}
-						x[j] = 1
-						scale = 0
-						xMax = 0
+				// Compute x[j] = x[j] / A[j,j], scaling if necessary.
+				// Note: the reference implementation skips this step for blas.Unit matrices
+				// when tscal is equal to 1 but it complicates the logic and only saves
+				// the comparison and division in the first switch-case. Not skipping it
+				// is also consistent with the NoTrans case above.
+				switch {
+				case tjj > smlnum:
+					// smlnum < abs(A[j,j]):
+					if tjj < 1 && xj > tjj*bignum {
+						// Scale x by 1/abs(x[j]).
+						rec := 1 / xj
+						bi.Dscal(n, rec, x, 1)
+						scale *= rec
+						xMax *= rec
 					}
+					x[j] /= tjjs
+				case tjj > 0:
+					// 0 < abs(A[j,j]) <= smlnum:
+					if xj > tjj*bignum {
+						// Scale x by (1/abs(x[j]))*abs(A[j,j])*bignum.
+						rec := (tjj * bignum) / xj
+						bi.Dscal(n, rec, x, 1)
+						scale *= rec
+						xMax *= rec
+					}
+					x[j] /= tjjs
+				default:
+					// A[j,j] == 0: Set x[0:n] = 0, x[j] = 1, and scale = 0, and
+					// compute a solution A^T * x = 0.
+					for i := range x[:n] {
+						x[i] = 0
+					}
+					x[j] = 1
+					scale = 0
+					xMax = 0
 				}
 			} else {
 				// Compute x[j] := x[j] / A[j,j] - sumj
--- a/lapack/gonum/dlatbs.go
+++ b/lapack/gonum/dlatbs.go
@@ -1,0 +1,449 @@
+// Copyright ©2019 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package gonum
+
+import (
+	"math"
+
+	"gonum.org/v1/gonum/blas"
+	"gonum.org/v1/gonum/blas/blas64"
+)
+
+// Dlatbs solves a triangular banded system of equations
+//  A * x = s*b    if trans == blas.NoTrans
+//  A^T * x = s*b  if trans == blas.Trans or blas.ConjTrans
+// where A is an upper or lower triangular band matrix, x and b are n-element
+// vectors, and s is a scaling factor chosen so that the components of x will be
+// less than the overflow threshold.
+//
+// On entry, x contains the right-hand side b of the triangular system.
+// On return, x is overwritten by the solution vector x.
+//
+// normin specifies whether the cnorm parameter contains the column norms of A on
+// entry. If it is true, cnorm[j] contains the norm of the off-diagonal part of
+// the j-th column of A. If it is false, the norms will be computed and stored
+// in cnorm.
+//
+// Dlatbs returns the scaling factor s for the triangular system. If the matrix
+// A is singular (A[j,j]==0 for some j), then scale is set to 0 and a
+// non-trivial solution to A*x = 0 is returned.
+//
+// Dlatbs is an internal routine. It is exported for testing purposes.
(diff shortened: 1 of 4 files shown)


lapack/gonum/dlatbs.go
```diff
@@ -0,0 +1,449 @@
...
+		xMax = bignum
+	}
+
+	if noTran {
+		// Solve A * x = b.
+		for j := jFirst; j != jLast; j += jInc {
+			// Compute x[j] = b[j] / A[j,j], scaling x if necessary.
+			xj := math.Abs(x[j])
+			tjjs := tscal
+			if diag == blas.NonUnit {
```
kortschak: I think I got lost here. There is a condition in the Fortran for `tscal.EQ.one` dropping through to the scaling of `x` that doesn't seem to happen here.

lapack/gonum/dlatbs.go
```diff
@@ -0,0 +1,449 @@
...
+				}
+				x[j] = 1
+				xj = 1
+				scale = 0
+				xMax = 0
+			}
+
+			// Scale x if necessary to avoid overflow when adding a multiple of
+			// column j of A.
+			if xj > 1 {
```
kortschak: `switch` since this seems to be a switch-prone function?

lapack/gonum/dlatbs.go
```diff
@@ -0,0 +1,449 @@
...
+				bi.Dscal(n, 0.5, x, 1)
+				scale *= 0.5
+			}
+
+			if uplo == blas.Upper {
+				if j > 0 {
+					// Compute the update
+					//  x[max(0,j-kd):j] := x[max(0,j-kd):j] - x[j] * A[max(0,j-kd):j,j]
+					jlen := min(j, kd)
+					if jlen > 0 {
```
kortschak: Optimising conditional?

lapack/gonum/dlatbs.go
```diff
@@ -0,0 +1,449 @@
...
+				xMax = math.Abs(x[i])
+			}
+		}
+	} else {
+		// Solve A^T * x = b.
+		for j := jFirst; j != jLast; j += jInc {
+			// Compute x[j] = b[j] - sum A[k,j]*x[k].
+			//                       k!=j
+			xj := math.Abs(x[j])
+			tjjs := tscal
```
kortschak: This few lines is conditional in the Fortran. I think I see, it's calculated conditionally there but in two places. This is a hoist. Is that correct?

lapack/testlapack/dlatbs.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+}
+
+// dlatbsResidual returns the residual for the solution to a scaled triangular
+// system of equations  A*x = s*b  or  A^T*x = s*b  when A is an n×n triangular
+// band matrix with kd super- or sub-diagonals. The residual is computed as
+//  norm( op(A)*x - scale*b ) / ( norm(op(A)) * norm(x) ).
+//
+// This function corresponds to DTBT03 in Reference LAPACK.
+func dlatbsResidual(uplo blas.Uplo, trans blas.Transpose, diag blas.Diag, n, kd int, ab []float64, ldab int, scale float64, cnorm, b, x []float64) float64 {
```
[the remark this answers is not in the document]
vladimir-ch: No, it's just making my life easier and the code simpler. DTBT03 is called in LAPACK only from dchktb.f and only with NRHS=1, so I didn't want to make it more general than necessary.

lapack/gonum/dlatbs.go
```diff
@@ -0,0 +1,449 @@
...
+	kld := max(1, ldab-1)
+	if !normin {
+		// Compute the 1-norm of each column, not including the diagonal.
+		if uplo == blas.Upper {
+			for j := 0; j < n; j++ {
+				jlen := min(j, kd)
+				if jlen > 0 {
+					cnorm[j] = bi.Dasum(jlen, ab[(j-jlen)*ldab+jlen:], kld)
+				} else {
+					cnorm[j] = 0
```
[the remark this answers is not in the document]
vladimir-ch: Nothing, I just wanted both uplo branches to look the same. I can change to rely on Dasum with the zeroing if you want.

lapack/testlapack/dlatbs.go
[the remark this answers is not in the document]
kortschak: Fair enough.

lapack/gonum/dlatbs.go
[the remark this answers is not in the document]
kortschak: No, this is fine.

lapack/gonum/dlatbs.go
vladimir-ch: Yes, this one is a bit tricky. I omitted the `goto` because: if `tscal` is 1, then `tjjs` and `tjj` are 1 and so the first `case` of the `switch` will trigger. In that `case` the `if` branch will not be taken, so essentially what the `goto` is skipping is only the division by `tjjs` and the absolute value. I thought that doing these two operations is a fair price for a clearer code flow because for example the declaration of tjj would have to happen before the goto and tjj would be initialized from tjjs only conditionally.

lapack/gonum/dlatbs.go
vladimir-ch: Ok, will do.

lapack/gonum/dlatbs.go
```diff
@@ -0,0 +1,449 @@
...
+// entry. If it is true, cnorm[j] contains the norm of the off-diagonal part of
+// the j-th column of A. If it is false, the norms will be computed and stored
+// in cnorm.
+//
+// Dlatbs returns the scaling factor s for the triangular system. If the matrix
+// A is singular (A[j,j]==0 for some j), then scale is set to 0 and a
+// non-trivial solution to A*x = 0 is returned.
+//
+// Dlatbs is an internal routine. It is exported for testing purposes.
+func (Implementation) Dlatbs(uplo blas.Uplo, trans blas.Transpose, diag blas.Diag, normin bool, n, kd int, ab []float64, ldab int, x []float64, cnorm []float64) (scale float64) {
```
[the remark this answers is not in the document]
vladimir-ch: Will do.

lapack/gonum/dlatbs.go
vladimir-ch: Yes (that's a lot of argument preparation for a no-op) and also to have the same logic as in the Lower case.

lapack/gonum/dlatbs.go
vladimir-ch: Correct, this is a hoist and hopefully is also clearer, otherwise the logic is quite ugly:
```
DO 160 J = JFIRST, JLAST, JINC
    USCAL = TSCAL
    IF( CNORM( J ).GT.( BIGNUM-XJ )*REC ) THEN
*       Initialize and use TJJS
*       Sometimes modify USCAL
    END IF
* ...
    IF( USCAL.EQ.TSCAL ) THEN
*       Initialize and use TJJS
    ELSE
*       Use TJJS - understanding that TJJS has indeed been initialized here
*       doesn't come immediately.
    END
160 CONTINUE
```
I see no benefit of the logic in the reference.

lapack/gonum/dlatbs.go
```diff
@@ -0,0 +1,449 @@
...
+					}
+				}
+			}
+
+			if uscal == tscal {
+				// Compute x[j] := ( x[j] - sumj ) / A[j,j]
+				// if 1/A[j,j] was not used to scale the dot product.
+				x[j] -= sumj
+				xj = math.Abs(x[j])
+				if diag == blas.NonUnit {
```
vladimir-ch: I _think_ this should be `if diag == blas.NonUnit || tjjs != 1 {` but could you please also give it a thought to confirm?

vladimir-ch: Could it even be only `if tjjs != 1 {`?

lapack/gonum/dlatbs.go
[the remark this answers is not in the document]
vladimir-ch: Done.

lapack/gonum/dlatbs.go
vladimir-ch: Done.

lapack/gonum/dlatbs.go
kortschak: Thanks

lapack/gonum/dlatbs.go
kortschak: Thanks

lapack/gonum/dlatbs.go
kortschak: If the required scaling is 1, the first path of the switch is taken, of which the conditional is not run and the elements of x are scaled by 1. In the case of a unit matrix, this still has not impact. For `tjjs != 1` then there is scaling. In the unit case, this looks to me like it will also scale the diagonal. If the user is storing information in the diagonal, this is probably not what they want.

So I think the `diag == blas.NonUnit` is required, but you could add `&& tjjs != 1` to elide the pointless scaling by 1.

Sorry to muddy the waters further.

lapack/testlapack/matgen.go
```diff
@@ -738,3 +738,505 @@ func applyReflector(qh blas64.General, q blas64.General, v []float64) {
...
+		// CALL DLATMS( N, N, DIST, ISEED, TYPE, B, MODE, CNDNUM, ANORM,
+		// $            KL, KU, PACKIT, AB( IOFF, 1 ), LDAB, WORK, INFO )
+		panic("test matrix type not implemented")
+
+	case 6:
+		// Matrix is the identity.
+		if uplo == blas.Upper {
+			for i := 0; i < n; i++ {
+				// Fill the diagonal with random numbers.
+				ab[i*ldab] = rnd.NormFloat64()
```
kortschak: Is there a reason to use random rather than the systematic non-unitary elements in the original?

lapack/testlapack/matgen.go
```diff
@@ -738,3 +738,505 @@ func applyReflector(qh blas64.General, q blas64.General, v []float64) {
...
+				rexp := uniformM11()
+				if rexp < 0 {
+					star1 = -math.Pow(sfac, 1-rexp)
+				} else {
+					star1 = math.Pow(sfac, 1+rexp)
+				}
+			}
+			// Copy the diagonal to AB.
+			if uplo == blas.Upper {
+				bi.Dcopy(n-1, work1, 1, ab[1:], ldab)
```
kortschak: `n` can be 1 by here, no? Is it worth this being made conditional?

lapack/testlapack/matgen.go
```diff
@@ -738,3 +738,505 @@ func applyReflector(qh blas64.General, q blas64.General, v []float64) {
...
+				if i != iy {
+					ab[i*ldab+kd] = math.Copysign(2, ab[i*ldab+kd])
+				} else {
+					ab[i*ldab+kd] = 0
+				}
+			}
+		}
+		bi.Dscal(n, 2, b, 1)
+
+		// case 16:
```
kortschak: TODO?

lapack/gonum/dlatbs.go
vladimir-ch: >  In the case of a unit matrix, this still has not impact. 

The problem that I see is that even in the case of a unit matrix there is the `tscal` factor which may cause `tjjs` to be non-unit. The reference executes the switch always for `diag == NonUnit` matrices _and_ for `Unit` matrices when `tscal != 1`. My code here executes the switch only for `NonUnit` matrices. So I still think it should have `|| tjjs != 1`. However, effectively what matters is not whether the matrix is Unit or NonUnit but rather what is the value that divides `x[j]` and therefore the simpler test `if tjjs != 1 {` should be enough?

> If the user is storing information in the diagonal, this is probably not what they want.
The diagonal elements in `ab` are accessed only for `NonUnit` matrices:
```
tjjs := tscal
if diag == blas.NonUnit {
	tjjs *= ab[j*ldab+maind]
}
```

vladimir-ch: Or simply don't skip the switch at all. If `tjjs` is 1, so be it, there will be some negligible extra calculations. It's not being skipped in the `noTran` case either.

kortschak: Yes, sorry, I completely misread that.

I don't know. I will leave it up to you. Maybe @btracey wants to have a read?

vladimir-ch: I've removed the skipping of the switch. I'm fairly confident that it is correct and the current tests pass ...

lapack/testlapack/matgen.go
vladimir-ch: Not really. Maybe I was not consistent in what I put on the diagonal of Unit matrices, I'll check again.

lapack/testlapack/matgen.go
vladimir-ch: Yes, it can. I had it conditional but then decided it's not worth it. I think it's ok to leave as is.

lapack/testlapack/matgen.go
vladimir-ch: Yes, I'll add it.

lapack/testlapack/matgen.go
vladimir-ch: Done.

lapack/testlapack/matgen.go
vladimir-ch: Done.

vladimir-ch:
It's not as bad as it looks, except for 16 the test matrices are relatively easy to understand. And even if there are mistakes, it doesn't really matter: `dlattb` does create a matrix and a right hand side, so as long as the test for Dlatbs passes and Dlatbs is covered, it fulfills its purpose. But feel free to ask if you want to understand.

PR #1036, opened 2019-07-23
mat: replace untranspose with untransposeExtract

btracey:
This leaves one remaining use in TriDense but that case is a bit more complicated

Please take a look.

### DIFF
--- a/mat/dense.go
+++ b/mat/dense.go
@@ -389,10 +389,10 @@ func (m *Dense) CloneFrom(a Matrix) {
 	}
 	m.capRows, m.capCols = r, c
 
-	aU, trans := untranspose(a)
+	aU, trans := untransposeExtract(a)
 	switch aU := aU.(type) {
-	case RawMatrixer:
-		amat := aU.RawMatrix()
+	case *Dense:
+		amat := aU.mat
 		mat.Data = make([]float64, r*c)
 		if trans {
 			for i := 0; i < r; i++ {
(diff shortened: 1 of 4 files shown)


btracey:
Now that I've done this, I wonder if untransposeExtract should be more complicated (or replaced with a more complicated version). The challenge with writing special case code is that it's annoying because there are so many of them. Direct versions, Raw versions, implements-interface-but-not-raw, etc. This makes contributors (or at least me) less interested in writing fast paths. untransposeExtract was step to at least eliminate the direct/raw distinction. The suggestion is to go further by doing something like 

func untransposeToDense(m Matrix) (Matrix, bool){
     a, trans := untranspose(m)
     switch a.(type){
          case *VecDense:
               return a, trans
          case Vectorer:
               var t VecDense
               // copy a into VecDense
               return &t, trans
           case *DiagDense:
               return a, trans
           case Diagonal:
              var t DiagDense
              // copy a into t
             return t, trans
      // etc. etc.
           case *Dense
               return a, trans
           default:
               // copy into *Dense
     }
}
```

This is moderately slower in the case that the Matrix implements the interface but not the RawFoo, as the elements will be looped over twice rather than once. However, it 1) drastically simplifies code like SymDense.RankTwo where we have a bunch of different code paths depending on the exact combination of inputs. 2) Means its much less likely the fast-path for that interface type will actually exist in the first place rather than as a TODO, which can often improve the Big-O time 3) Code going through the interface directly is probably slow anyway. It may be that we don't want the Matrix case to copy into *Dense, as for functions like VecDense.Mul this could lead to a large increase in memory that is otherwise unavoidable. But for the other cases (especially Diagonal and Vector), I don't see any major downsides.

kortschak:
Can you file that suggestion as an issue?

PR #1040, opened 2019-08-02
lapack/gonum: add Dpbcon

vladimir-ch:
Please take a look.

### DIFF
--- a/lapack/gonum/dpbcon.go
+++ b/lapack/gonum/dpbcon.go
@@ -1,0 +1,108 @@
+// Copyright ©2019 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package gonum
+
+import (
+	"math"
+
+	"gonum.org/v1/gonum/blas"
+	"gonum.org/v1/gonum/blas/blas64"
+)
+
+// Dpbcon returns an estimate of the reciprocal of the condition number (in the
+// 1-norm) of an n×n symmetric positive definite band matrix using the Cholesky
+// factorization
+//  A = U^T*U  if uplo == blas.Upper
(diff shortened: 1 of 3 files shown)


lapack/gonum/dpbcon.go
```diff
@@ -0,0 +1,108 @@
...
+		cnorm = work[2*n : 3*n]
+	)
+	// Estimate the 1-norm of the inverse.
+	bi := blas64.Implementation()
+	for {
+		ainvnm, kase = impl.Dlacn2(n, v, x, iwork, ainvnm, kase, &isave)
+		if kase == 0 {
+			break
+		}
+		if uplo == blas.Upper {
```
kortschak: I'm torn by the factoring in this block. On the one hand it would be nice to factor out the work completely and just set the first half and second half `blas.Trans`/`blas.NoTrans` values in the conditional block, but on the other it's nice to have the explicit expressions. I don't know.

Whay I'm thinking for the first case is
```
var op1, op2 blas blas.Transpose
if uplo == blas.Upper {
    // Multiply x by inv(U^T),
    op1 = blas.Trans
    // then by inv(U).
    op2 = blas.NoTrans
} else {
    // Multiply x by inv(L),
    op1 = blas.NoTrans
    // then by inv(L^T).
    op2 = blas.Trans
}
scaleL = impl.Dlatbs(uplo, op1, blas.NonUnit, normin, n, kd, ab, ldab, x, cnorm)
normin = true
scaleU = impl.Dlatbs(uplo, op2, blas.NonUnit, normin, n, kd, ab, ldab, x, cnorm)
```

Alternatively, explicitly show the `uplo`.

vladimir-ch: I changed it to your version, I like how it highlights the difference between the two uplo.

PR #1046, opened 2019-08-04
mat: clarify LQ

kortschak:
Fix comment and make code format consistent with QR.

Please take a look.

### DIFF
--- a/mat/lq.go
+++ b/mat/lq.go
@@ -40,12 +40,12 @@ func (lq *LQ) updateCond(norm lapack.MatrixNorm) {
 	putInts(iwork)
 }
 
-// Factorize computes the LQ factorization of an m×n matrix a where n <= m. The LQ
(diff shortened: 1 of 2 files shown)


mat/lq.go
```diff
@@ -40,12 +40,12 @@ func (lq *LQ) updateCond(norm lapack.MatrixNorm) {
 	putInts(iwork)
 }
 
-// Factorize computes the LQ factorization of an m×n matrix a where n <= m. The LQ
+// Factorize computes the LQ factorization of an m×n matrix a where m <= n. The LQ
 // factorization always exists even if A is singular.
 //
 // The LQ decomposition is a factorization of the matrix A such that A = L * Q.
 // The matrix Q is an orthonormal n×n matrix, and L is an m×n upper triangular matrix.
```
vladimir-ch: L is _lower_ triangular. Can you please fix also this?

kortschak: Done. Now we know for sure that this was a copy paste error.

PR #1047, opened 2019-08-04
mat: add minimal shadow detection for complex matrices

kortschak:
This is initial bootstrapping work to allow generation of CDense from Dense.

I'm not sure of `RawCMatrix`. Is this what we want?

Please take a look.

### DIFF
--- a/mat/offset.go
+++ b/mat/offset.go
@@ -18,3 +18,14 @@ func offset(a, b []float64) int {
 	// move. See https://golang.org/issue/12445.
 	return int(uintptr(unsafe.Pointer(&b[0]))-uintptr(unsafe.Pointer(&a[0]))) / int(unsafe.Sizeof(float64(0)))
 }
+
+// offsetComplex returns the number of complex128 values b[0] is after a[0].
+func offsetComplex(a, b []complex128) int {
+	if &a[0] == &b[0] {
+		return 0
+	}
+	// This expression must be atomic with respect to GC moves.
+	// At this stage this is true, because the GC does not
+	// move. See https://golang.org/issue/12445.
+	return int(uintptr(unsafe.Pointer(&b[0]))-uintptr(unsafe.Pointer(&a[0]))) / int(unsafe.Sizeof(float64(0)))
+}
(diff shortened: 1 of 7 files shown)


mat/offset.go
```diff
@@ -18,3 +18,14 @@ func offset(a, b []float64) int {
...
+
+// offsetComplex returns the number of complex128 values b[0] is after a[0].
+func offsetComplex(a, b []complex128) int {
+	if &a[0] == &b[0] {
+		return 0
+	}
+	// This expression must be atomic with respect to GC moves.
+	// At this stage this is true, because the GC does not
+	// move. See https://golang.org/issue/12445.
+	return int(uintptr(unsafe.Pointer(&b[0]))-uintptr(unsafe.Pointer(&a[0]))) / int(unsafe.Sizeof(float64(0)))
```
vladimir-ch: Shouldn't it be `complex128`?

mat/offset_appengine.go
```diff
@@ -22,3 +22,16 @@ func offset(a, b []float64) int {
...
+func offsetComplex(a, b []complex128) int {
+	va0 := reflect.ValueOf(a).Index(0)
+	vb0 := reflect.ValueOf(b).Index(0)
+	if va0.Addr() == vb0.Addr() {
+		return 0
+	}
+	// This expression must be atomic with respect to GC moves.
+	// At this stage this is true, because the GC does not
+	// move. See https://golang.org/issue/12445.
+	return int(vb0.UnsafeAddr()-va0.UnsafeAddr()) / sizeOfFloat64
```
vladimir-ch: Also here?

mat/offset.go
kortschak: Yes. Thanks

mat/offset_appengine.go
kortschak: Fixed.

btracey:
I haven't fully finished reviewing the shadow_complex bit but that's all I noticed as well.

kortschak:
Yeah. It's essentially copy/paste, with all the errors associated with that approach. There is no testing because that depends on `Slice`. It will be very much easier when there is a reasonable set of generated code for complex.

mat/offset.go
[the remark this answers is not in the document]
btracey: Whoops, didn't realize these were still hanging around

Issue #1049, opened 2019-08-06
r2/r3: flesh out Vec API


odsod:
### What are you trying to do?

We want to do some simple vector math on the r3.Vec type (cross product, dot product, etc): https://github.com/golang/geo/blob/master/r3/vector.go

### What did you try?

We've added common operations as helper functions. It works, but might be useful to more people (and simplify reuse) to include them as member functions on the Vec type.

### How does Gonum not allow you to achieve your goal?

Gonum helps us, but we are proposing to add some of the most common vector operations to as member functions. 

### What version of Go and Gonum are you using?

go version go1.12.7 linux/amd64
gonum: 3ce8ac2a2fb11ca0240b45098d1f4393d3912a71

### Is this feature absent from the current master?

Yes.

### Are you able to help contribute the feature?

Yes.

odsod:
@kortschak I can see you recently added the r2/r3 packages. Do you think additions like this are in scope for these packages?

kortschak:
I think this is where they would go if they do. I'd like to hear others' opinions though; we rushed into the set operations on `r1` and then decided it was too complicated for what we intended and then had to back them out. I'd like to avoid that happening again.

/cc @btracey @vladimir-ch @sbinet

odsod:
@kortschak Understood!

If you give the go-ahead, I can submit PRs for the most common operations like:

* Mul
* Cross (included in the example PR #1050)
* Dot
* Norm
* anything else you suggest

sbinet:
I guess that's a good set of minimal operations one would want to apply on r3.Vec.

I'd be fine with it.

btracey:
I think that's a good set as well. 

In the past I've also implemented `[2][2]float64` matrices and `[3][3]float64` matrices to do common operations (i.e. determinant, splitting in various ways). Would those kinds of operations belong here as well? If that is true, would we want Vector to be a `[2]float64`? It's easy to have an `func (v Vector) X() float64` method, and would be more consistent.

kortschak:
I could be convinced.

sbinet:
what's the status of this PR?

(I am asking b/c I'll probably have to add some more R^3 capabilities to one of my packages, soon)
playing a bit with `geo/r3` (and my very limited `Vec3` type), I think I'd be w/ @btracey on making `r3.Vec` a 3-float64 array. (that way, creating a `mat.VecDense` is just sub-slicing the `r3.Vec` 3-array, possibly modifying it in-place)

[...]

PR #1050, opened 2019-08-06
 spatial/r3: add (Vec).Cross

odsod:
https://github.com/gonum/gonum/issues/1049

Submitting this change as an example of operations to add. Happy to provide more common operations in subsequent PRs if this one gets accepted.

### DIFF
--- a/spatial/r3/vector.go
+++ b/spatial/r3/vector.go
@@ -17,6 +17,15 @@ func (p Vec) Add(q Vec) Vec {
 	return p
 }
 
+// Cross returns the vector cross product of p and q.
+func (p Vec) Cross(q Vec) Vec {
+	return Vec{
+		X: p.Y*q.Z - p.Z*q.Y,
+		Y: p.Z*q.X - p.X*q.Z,
+		Z: p.X*q.Y - p.Y*q.X,
+	}
+}
+
 // Sub returns the vector sum of p and -q.
 func (p Vec) Sub(q Vec) Vec {
 	p.X -= q.X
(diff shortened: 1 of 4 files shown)


spatial/r3/vector_test.go
```diff
@@ -0,0 +1,13 @@
+package r3
+
+import "testing"
+
+func TestVector_Cross(t *testing.T) {
+	a := Vec{X: 2, Y: 3, Z: 4}
+	b := Vec{X: 5, Y: 6, Z: 7}
+	expected := Vec{X: -3, Y: 6, Z: -3}
+	actual := a.Cross(b)
```
sbinet: and that `got`:
```suggestion
	got := a.Cross(b)
```

spatial/r3/vector_test.go
```diff
@@ -0,0 +1,13 @@
+package r3
```
sbinet: please add the missing copyright header.

spatial/r3/vector.go
```diff
@@ -17,6 +17,15 @@ func (p Vec) Add(q Vec) Vec {
 	return p
 }
 
+// Cross returns the vector cross product of p and q.
```
sbinet: we could perhaps replace `and` with the cross symbol (a simple `x` would do, but we could also reach for UTF8's `×`)

odsod:
@sbinet Thanks for the feedback, fixes coming shortly!

spatial/r3/vector.go
kortschak: The UTF-8 symbol is what we usually use where we use a symbol, though we often don't use a symbol.

odsod:
Update - we moved in a different direction and are not needing this functionality - dropping this PR. Cheers

PR #1051, opened 2019-08-06
travis: make sure checks exit non-zero if they fail

kortschak:
At the moment, it's printing file names for missing headers, but not failing.

Please take a look.

### DIFF
--- a/.travis/run-parts
+++ b/.travis/run-parts
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+set -e
+
 if [ $# -ne 1 ]; then
     echo "run-parts <directory>"
     exit 1


kortschak:
This doesn't fix the problem, and I don't understand now 1. why it doesn't work as it is in master and 2. why this doesn't work either.

sbinet:
wouldn't this because the `run-parts` script has no `set -e` line?

kortschak:
Could be, except that the tests are run by `run-parts` and they successfully fail when needed.

kortschak:
I think you are right. The tests are last and so would be returned.

kortschak:
PTAL (experimentally confirmed)

kortschak:
I don't understand these failures. They are replicable on my machine locally, but have never appeared before. I will look into what is going on.

kortschak:
OK. I now understand. The tests previously were passing exactly because of the loop here quietening the failures. I never saw the failures that we have above because I generally don't run `-tags=noasm` locally. The issue comes down now to numerical stability of ED used in `mds.Torgerson`, which is something that we've already seen.

kortschak:
I've rebased this to ensure that it now passes and changed the commit message since the fix was more general than previously though.

PTAL

PR #1052, opened 2019-08-08
graph/layout: handle difference between pure go and asm kernels

kortschak:
This fixes the failures seen in #1051.

Please take a look.


kortschak:
Sorry. Not right yet. I didn't commit the var decl change.

kortschak:
PTAL

PR #1053, opened 2019-08-11
stat/combin: Implement functions for helping with Permutations

### LINKED ISSUE #1033: stat/combin: Add Permutations and PermutationsGenerator
Could consider Product as well (as in CartesianProduct)

btracey:
Fixes #1033.

Please take a look.

### DIFF
--- a/stat/combin/combin.go
+++ b/stat/combin/combin.go
@@ -185,6 +185,98 @@ func nextCombination(s []int, n, k int) {
 	}
 }
 
+// PermutationIndex returns the index of the given permutation.
+//
+// The functions PermutationIndex and IndexToPermutation define a bijection
+// between the integers and the permutations of the integers [0, len(perm)-1].
+// PermutationIndex returns the inverse of IndexToPermutation.
+//
+// PermutationIndex panics if perm is not a permutation of the first len(perm)
+// non-negative integers.
+func PermutationIndex(perm []int) int {
+	// Note(btracey): This is an n^2 algorithm, but factorial increases
+	// very quickly (25! overflows int64) so this is not a problem in
+	// practice.
+
+	check := map[int]struct{}{}
+	for _, v := range perm {
+		if 0 <= v || v >= len(perm) {
+			panic("combin: perm value out of range")
+		}
+		check[v] = struct{}{}
+	}
+	if len(check) != len(perm) {
+		panic("combin: perm contains duplicate elements")
+	}
+
+	idx := 0
+	for i, u := range perm {
+		less := 0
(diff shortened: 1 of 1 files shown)


kortschak:
I think this wants some tests, and probably also some examples to help out with the documentation.

btracey:
This is a partial commit, because implemented the work and realized the issue.

Permutations are normally P(n, k), while often times in iteration you just want the iterations of the first k objects (like in the python itertools.permutations). I assume we want this to be the fully general form, so `IndexToPermutation(dst []int, idx, n, k int) int`? I'm pretty sure this is possible, because there's also a function `IndexToCombination`, and we can basically select the combination first and then the permutation. I'll make that the next PR.

This current implementation uses that `n = k`, and just uses the length of the appropriate slice to get both values. If we go to the more general version, would we rather keep inferring `k` from the length of the slice, and just require the user to specify `n`, or require the user to specify both? I lean toward the latter given that they're both integers and would be confusing from context otherwise.

btracey:
Sorry @kortschak didn't get my note out fast enough!

stat/combin/combin.go
```diff
@@ -185,6 +185,98 @@ func nextCombination(s []int, n, k int) {
...
+	p := 1
+	for i := n - k + 1; i <= n; i++ {
+		p *= i
+	}
+	return p
+}
+
+// factorial returns a!.
+func factorial(a int) int {
+	a := 1
```
kortschak: `f`?

stat/combin/combin.go
```diff
@@ -185,6 +185,98 @@ func nextCombination(s []int, n, k int) {
...
+// PermutationIndex panics if perm is not a permutation of the first len(perm)
+// non-negative integers.
+func PermutationIndex(perm []int) int {
+	// Note(btracey): This is an n^2 algorithm, but factorial increases
+	// very quickly (25! overflows int64) so this is not a problem in
+	// practice.
+
+	check := map[int]struct{}{}
+	for _, v := range perm {
+		if 0 <= v || v >= len(perm) {
```
vladimir-ch: `0 <= v` should be `v < 0`?

kortschak: Also, `len(perm) <= v`.

sbinet: I guess an easier way to spell it might be:
```go
if !(0 <= v && v < len(perm)) {
```

kortschak: We normally write value out of range as `v < 0 || len(perm) <= v`.

btracey:
Closing in favor of #1076

PR #1054, opened 2019-08-11
stat/combin: Add CombinationToIndex and IndexToCombination functions

btracey:
Please take a look.


btracey:
Couple of (meta) comments

1) @kortschak suggested examples would be good in #1033 . I don't disagree, but I'm wondering what a helpful example would contain. I can write one converting it to integers and back, is that helpful? It's a bit more difficult to explain the use case (which is basically random access into CombinationGenerator)

2) I just realized this order does not match the order of the other iterators, though I believe that can be fixed. Is having a consistent order something we want to document?

3) Scipy (at least) defines Binomial to be zero when n < k. That would also simplify the algorithms here. Is this a change we want to make? (Can be an issue / different PR)

btracey:
Yea, the problem with the original algorithm is that it treats the "ones place" as the first index, and larger numbers as later (and thus an ending order like [2 3 5], [0, 4, 5], [1, 4, 5] etc.) while the other two algorithms treat the last index as the "ones place" and earlier numbers sooner, so you get [0, 3, 4], [0, 3, 5] [0, 4, 5], [1, 2, 3].

I implemented the "quick fix", which uses the old algorithms, but just swaps the sign in a few places. If this isn't too egregious, I can add comments as to what is going on. I think the alternative involves performing the same operations on the micro scale rather than the macro scale, which seems like it would be more confusing, but I could give it a try.

kortschak:
1. An example for show a short loop that iterates over the combination space with a non-unit step. This can't be efficiently done with the generator that exists. The reverse would just be a lookup.
2. I do think it's important that the orders match.
3. Wolfram alpha (and so probably mathematica) also treat C(n, k) = 0 when n < k.
4. Yes.

btracey:
Fixed import orders and added an example.

PTAL

stat/combin/combinations_example_test.go
```diff
@@ -0,0 +1,82 @@
+package combin_test
```
kortschak: Copyright header.

btracey: Thanks.

stat/combin/combin.go
```diff
@@ -185,6 +186,96 @@ func nextCombination(s []int, n, k int) {
...
+		panic(badNegInput)
+	}
+	if n < k {
+		panic(badSetSize)
+	}
+	if len(comb) != k {
+		panic("combin: bad length combination")
+	}
+	if !sort.IntsAreSorted(comb) {
+		panic("comb: input combination is not sorted")
```
kortschak: s/comb/combin/

stat/combin/combin.go
```diff
@@ -185,6 +186,96 @@ func nextCombination(s []int, n, k int) {
...
+// IndexToCombination panics if n or k are non-negative, if k is greater than n,
+// or if idx is not in [0, Binomial(n,k)-1]. IndexToCombination will also panic
+// if dst is non-nil and len(dst) is not k.
+func IndexToCombination(dst []int, idx, n, k int) []int {
+	if idx < 0 || idx >= Binomial(n, k) {
+		panic("combin: invalid index")
+	}
+	if dst == nil {
+		dst = make([]int, k)
+	} else {
```
kortschak: `} else if len(dst) != k {`

stat/combin/combin.go
```diff
@@ -185,6 +186,96 @@ func nextCombination(s []int, n, k int) {
...
+		// Find the largest number m such that Binomial(m, k-i) <= idx.
+		// This is one less than the first number such that it is larger.
+		m := sort.Search(n, func(m int) bool {
+			if m < k-i {
+				return false
+			}
+			return Binomial(m, k-i) > idx
+		})
+		m--
+		// Normally this is put m into the last free spot, but we
```
kortschak: "Normally this puts m into the last free spot..."

stat/combin/combinations_example_test.go
```diff
@@ -0,0 +1,86 @@
...
+
+package combin_test
+
+import (
+	"fmt"
+
+	"gonum.org/v1/gonum/stat/combin"
+)
+
+func Example() {
```
kortschak: I would split these into three separate package-level examples.

stat/combin/combin.go
btracey: Done.

stat/combin/combin.go
btracey: Done.

stat/combin/combinations_example_test.go
btracey: Done.

PR #1058, opened 2019-08-14
lapack/testlapack: rewrite and extend test for Dgecon

vladimir-ch:
* Now that we understand that a mismatch between the exact and estimated condition number is normal, the logging of the mismatch can be removed.
* In general in lapack I prefer randomized tests.
* I don't know what I was thinking when I wrote the Dpbcon test.

Please take a look.

### DIFF
--- a/lapack/testlapack/dgecon.go
+++ b/lapack/testlapack/dgecon.go
@@ -5,92 +5,87 @@
 package testlapack
 
 import (
+	"fmt"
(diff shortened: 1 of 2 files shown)


kortschak:
>Now that we understand that a mismatch between the exact and estimated condition number is normal, the logging of the mismatch can be removed.

Is there a reference for this?

PR #1066, opened 2019-08-20
stat/running: Implement package for computing stats of a stream of data

### LINKED ISSUE #1037: stat: "running" subpackage for generator-like statistics?
Just a thought -- logging as an issue so I don't forget / can solicit feedback.

At the moment we have `stat.Mean(x, weights []float64) float64` which computes the mean of a set of data. The thought is to also provide something like stat/running

```
package running

type Mean struct {
    count float64
    mean float64
}

func (m *Mean) Mean() float64 {
    return m.mean
}

func (m *Mean) Count() float64 {
    return m.count 
}

func (m *Mean) Add(x float64) {
      m.mean = x + m.mean * m.count / (m.count + 1)
      m.count++
}

func (m *Mean) AddWeighted(x, weight float64){ .... }

func (m *Mean) Reset() { ... }
```

and there could be similar for Variance/StdDev, DiscountedMean (where the count is discounted by a constant, etc.). These could be relatively easily combined with a shim:

```
type MeanStd struct {
    running.Mean()
    running.StdDev()
}

func (m *MeanStd) Add(x float64) {
    m.Mean.Add(x)
    m.StdDev.Add(x)
}
```

This would help both with streams that are too large to want to put in a `[]float64` and data that comes from some source (say, a channel) and so cannot be put into a `float64`. 

I don't think `Add` is the right verb, but I'm not sure what the right verb is (`Accumulate` seems too heavy, `Append` also seems wrong ...)

btracey:
Please take a look.

Fixes #1037


btracey:
Sorry, this was branched from the wrong place.

PR #1067, opened 2019-08-20
stat/running: Add running package for computing stats of a running st…

### LINKED ISSUE #1037: stat: "running" subpackage for generator-like statistics?

btracey:
…ream of data

Fixes #1037.

Please take a look.

### DIFF
--- a/stat/running/running.go
+++ b/stat/running/running.go
@@ -1,0 +1,74 @@
+// Copyright ©2019 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package running
+
+import "math"
+
+// Mean helps compute running mean
+type Mean struct {
+	// Decay sets the decay rate. A value of 0.99 means that the effective count
+	// is multiplied by 0.99 for every added value. If Decay is zero, a default
+	// value of 1 is used (the count is not decayed).
+	Decay     float64
+	InitCount float64
+	InitMean  float64
+
+	count float64
+	mean  float64
+}
+
+// Reset resets the counter. The next accumulate will use the
+// initial mean and count.
+func (m *Mean) Reset() {
+	m.count = 0
+	m.mean = 0
+}
+
+func (m *Mean) isZero() bool {
+	return m.count == 0
+}
+
+// Mean returns the current estimate of the running mean.
+func (m *Mean) Mean() float64 {
+	if m.isZero() {
+		return m.InitMean
+	}
+	return m.mean
+}
+
+// Count returns the current count of values.
+func (m *Mean) Count() float64 {
+	if m.isZero() {
+		return m.InitCount
+	}
+	return m.count
+}
+
+// Accum adds the value to the running total.
+func (m *Mean) Accum(v float64) {
+	m.AccumWeighted(v, 1)
+}
+
+// AccumWeighted adds the weighted value to the running total.
+func (m *Mean) AccumWeighted(v, weight float64) {
+	if weight < 0 {
+		panic("running: negative weight")
+	}
+	if m.isZero() && weight == 0 {
+		return
+	}
+	decay := m.decay()
+	decay = math.Pow(decay, weight)
+	m.count *= decay
+	m.mean = (m.mean*m.count + v*weight) / (m.count + weight)
+	m.count += weight
+}
+
+func (m *Mean) decay() float64 {
+	if m.Decay == 0 {
+		return 1
+	}
+	return m.Decay
+}
--- a/stat/running/doc.go
+++ b/stat/running/doc.go
@@ -1,0 +1,7 @@
+// Copyright ©2019 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Package running implements routines for computing statistics on a stream
+// of values.
+package running // import "gonum.org/v1/gonum/stat/running"


btracey:
Currently a mostly complete sketch of how it could look (but needs better comments and tests).

I think a good idea would be to have a `running.Stats` that computes the kitchen sink (i.e. some decent set of things that can be computed online). Seems reasonable to also have a MeanStd . Not sure what other scope.

stat/running/running.go
```diff
@@ -0,0 +1,74 @@
...
+
+import "math"
+
+// Mean helps compute running mean
+type Mean struct {
+	// Decay sets the decay rate. A value of 0.99 means that the effective count
+	// is multiplied by 0.99 for every added value. If Decay is zero, a default
+	// value of 1 is used (the count is not decayed).
+	Decay     float64
+	InitCount float64
```
kortschak: Please add comments for these fields.

stat/running/running.go
```diff
@@ -0,0 +1,74 @@
...
+	InitMean  float64
+
+	count float64
+	mean  float64
+}
+
+// Reset resets the counter. The next accumulate will use the
+// initial mean and count.
+func (m *Mean) Reset() {
+	m.count = 0
```
kortschak: I don't think I understand how these are being used. I was going to suggest

```
m.count = m.InitCount
m.mean = m.InitMean
```

but then, the code in `AccumWeighted` doesn't make sense.

stat/running/running.go
```diff
@@ -0,0 +1,74 @@
+// Copyright ©2019 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package running
+
+import "math"
+
+// Mean helps compute running mean
```
kortschak: `// Mean is a running mean accumulator.`? The "helps" seems odd.

kortschak:
*ping*

kortschak:

stat/running/running.go
btracey: Done.

stat/running/running.go
btracey: Done.

stat/running/running.go
btracey: Done.

stat/running/running.go
```diff
@@ -52,17 +55,23 @@ func (m *Mean) Accum(v float64) {
...
+	}
+	if m.count == 0 && weight == 0 {
+		// Avoid NaN.
 		return
 	}
 	decay := m.decay()
 	decay = math.Pow(decay, weight)
 	m.count *= decay
-	m.mean = (m.mean*m.count + v*weight) / (m.count + weight)
+	//m.mean = (m.mean*m.count + v*weight) / (m.count + weight)
```
kortschak: Delete this commented out code?

stat/running/running.go
```diff
@@ -72,3 +81,11 @@ func (m *Mean) decay() float64 {
 	}
 	return m.Decay
 }
+
+func updateMean(mean, count, v, weight float64) float64 {
```
kortschak: Will this be used elsewhere?

stat/running/running.go
```diff
@@ -72,3 +81,11 @@ func (m *Mean) decay() float64 {
...
+	return (mean*count + v*weight) / (count + weight)
+}
+
+// Stats is an accumulator for running statistics.
+type Stats struct {
```
kortschak: More to come?

stat/running/running.go
```diff
@@ -6,42 +6,45 @@ package running
...
-	Decay     float64
+	Decay float64
+	// InitCount sets the initial count before accumulating.
 	InitCount float64
-	InitMean  float64
+	// InitMean sets the initial value for the mean before accumulating.
+	// Note that InitMean has no effect if the InitCount is zero.
+	InitMean float64
 
+	set   bool
```
kortschak: This needs a comment.

btracey:
I'm sorry, the last thing I pushed up was not what I meant to push. I pushed up what I intended now. Doing a refactor which will make it simpler.

btracey:
Yep, this is way better, thanks for the comments.

kortschak:
What about exporting the fields?

stat/running/running.go
```diff
@@ -0,0 +1,69 @@
...
+}
+
+// Count returns the current count of values.
+func (m *Mean) Count() float64 {
+	return m.count
+}
+
+// Accum adds the value to the running total.
+func (m *Mean) Accum(v float64) {
+	m.AccumWeighted(v, 1)
```
kortschak: I'd do this explicitly here. There are elidable checks and `math.Pow` is not cheap.

vladimir-ch:
@btracey Do you want to keep this open?

stat/running/running.go
```diff
@@ -0,0 +1,69 @@
+// Copyright ©2019 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package running
+
+import "math"
+
+// Mean is a running mean accumulator.
+type Mean struct {
```
soypat, line 10: Should we maybe name this type "MeanExpDecay" to leave the floor open for future running mean implementations?

PR #1068, opened 2019-08-26
stat: Add normal test

seppo0010:


stat/stat.go
```diff
@@ -1293,3 +1294,76 @@ func MeanVariance(x, weights []float64) (mean, variance float64) {
 	variance = (ss - compensation*compensation/sumWeights) / (sumWeights - 1)
 	return mean, variance
 }
+
+// NormalSkewTest test whether the skew is different from the normal distribution.
+// This function tests the null hypothesis that the skewness of population that the
+// sample was drawn from is the same as that of a corresponding normal distribution.
+// Implementation based on scipy's: https://github.com/scipy/scipy/blob/v1.3.0/scipy/stats/stats.py#L1481-L1549
```
kortschak: Implementation details should not be in user-facing docs unless it is important for them to understand this.

stat/stat.go
```diff
@@ -1293,3 +1294,76 @@ func MeanVariance(x, weights []float64) (mean, variance float64) {
...
+		return 1
+	}
+	return 0
+}
+
+// NormalKurtosisTest test whether a dataset has normal kurtosis.
+// This function tests the null hypothesis that the kurtosis of the population from
+// which the sample was drawn is that of the normal distribution:
+// `kurtosis = 3(n-1)/(n+1)`.
+// Implementation based on scipy's: https://github.com/scipy/scipy/blob/v1.3.0/scipy/stats/stats.py#L1481-L1549
```
kortschak: Implementation doc.

stat/stat.go
```diff
@@ -1293,3 +1294,76 @@ func MeanVariance(x, weights []float64) (mean, variance float64) {
...
+	}
+	term2 := sign(denom) * math.Pow((1-2.0/A)/math.Abs(denom), 1/3.0)
+	return (term1 - term2) / math.Sqrt(2/(9.0*A))
+}
+
+// NormalTest test whether a sample differs from a normal distribution.
+// This function tests the null hypothesis that a sample comes from a normal distribution.
+// It is based on D'Agostino and Pearson's [1]_, [2]_ test that combines skew and kurtosis to
+// produce an omnibus test of normality.
+// Implementation based on scipy's: https://github.com/scipy/scipy/blob/v1.3.0/scipy/stats/stats.py#L1481-L1549
```
kortschak: Implementation doc.

stat/stat_test.go
```diff
@@ -16,6 +16,118 @@ import (
 	"gonum.org/v1/gonum/floats"
 )
 
+// It'd be better if we could use distuv.Normal but that would create a cyclic import
```
kortschak: `package stat_test` would fix this, however see global comment.

stat/stat_test.go
```diff
@@ -1772,3 +1884,107 @@ func TestStdScore(t *testing.T) {
 		}
 	}
 }
+
+func TestNormalSkewTest(t *testing.T) {
+	// even though the skew test and kurtosis test are a copy from the scipy's the results
```
kortschak: Please add necessary leading capitals and final full-stops.

stat/stat.go
```diff
@@ -1293,3 +1294,76 @@ func MeanVariance(x, weights []float64) (mean, variance float64) {
...
+	W2 := -1 + math.Sqrt(2*(beta2-1))
+	delta := 1 / math.Sqrt(0.5*math.Log(W2))
+	alpha := math.Sqrt(2.0 / (W2 - 1))
+	if y == 0 {
+		y = 1
+	}
+	return delta * math.Log(y/alpha+math.Sqrt(math.Pow(y/alpha, 2)+1))
+}
+
+func sign(v float64) float64 {
```
kortschak: `math.CopySign`, but helpers get defined after use, not before.

stat/stat.go
```diff
@@ -1293,3 +1294,76 @@ func MeanVariance(x, weights []float64) (mean, variance float64) {
...
+
+// NormalSkewTest test whether the skew is different from the normal distribution.
+// This function tests the null hypothesis that the skewness of population that the
+// sample was drawn from is the same as that of a corresponding normal distribution.
+// Implementation based on scipy's: https://github.com/scipy/scipy/blob/v1.3.0/scipy/stats/stats.py#L1481-L1549
+func NormalSkewTest(a []float64) float64 {
+	b2 := Skew(a, nil)
+	n := float64(len(a))
+	if n < 8 {
+		panic(fmt.Sprintf("skewtest is not valid with less than 8 samples, %d samples were given", int(n)))
```
kortschak: Please document that this panic condition exists.

stat/stat.go
```diff
@@ -1293,3 +1294,76 @@ func MeanVariance(x, weights []float64) (mean, variance float64) {
...
+
+// NormalKurtosisTest test whether a dataset has normal kurtosis.
+// This function tests the null hypothesis that the kurtosis of the population from
+// which the sample was drawn is that of the normal distribution:
+// `kurtosis = 3(n-1)/(n+1)`.
+// Implementation based on scipy's: https://github.com/scipy/scipy/blob/v1.3.0/scipy/stats/stats.py#L1481-L1549
+func NormalKurtosisTest(a []float64) float64 {
+	n := float64(len(a))
+	if n < 5 {
+		panic(fmt.Sprintf("kurtosistest requires at least 5 observations; %d observations were given", int(n)))
```
kortschak: Please document panic condition.

stat/stat.go
```diff
@@ -1293,3 +1294,76 @@ func MeanVariance(x, weights []float64) (mean, variance float64) {
...
 	return mean, variance
 }
+
+// NormalSkewTest test whether the skew is different from the normal distribution.
+// This function tests the null hypothesis that the skewness of population that the
+// sample was drawn from is the same as that of a corresponding normal distribution.
+// Implementation based on scipy's: https://github.com/scipy/scipy/blob/v1.3.0/scipy/stats/stats.py#L1481-L1549
+func NormalSkewTest(a []float64) float64 {
+	b2 := Skew(a, nil)
+	n := float64(len(a))
```
kortschak: Perform this conversion after the check.

stat/stat.go
```diff
@@ -1293,3 +1294,76 @@ func MeanVariance(x, weights []float64) (mean, variance float64) {
...
+	return 0
+}
+
+// NormalKurtosisTest test whether a dataset has normal kurtosis.
+// This function tests the null hypothesis that the kurtosis of the population from
+// which the sample was drawn is that of the normal distribution:
+// `kurtosis = 3(n-1)/(n+1)`.
+// Implementation based on scipy's: https://github.com/scipy/scipy/blob/v1.3.0/scipy/stats/stats.py#L1481-L1549
+func NormalKurtosisTest(a []float64) float64 {
+	n := float64(len(a))
```
kortschak: After the check please.

stat/stat.go
```diff
@@ -1293,3 +1294,76 @@ func MeanVariance(x, weights []float64) (mean, variance float64) {
...
+// Implementation based on scipy's: https://github.com/scipy/scipy/blob/v1.3.0/scipy/stats/stats.py#L1481-L1549
+func NormalSkewTest(a []float64) float64 {
+	b2 := Skew(a, nil)
+	n := float64(len(a))
+	if n < 8 {
+		panic(fmt.Sprintf("skewtest is not valid with less than 8 samples, %d samples were given", int(n)))
+	}
+	y := b2 * math.Sqrt(((n+1)*(n+3))/(6.0*(n-2)))
+	beta2 := (3.0 * (math.Pow(n, 2) + 27*n - 70) * (n + 1) * (n + 3) / ((n - 2.0) * (n + 5) * (n + 7) * (n + 9)))
+	W2 := -1 + math.Sqrt(2*(beta2-1))
```
kortschak: Lower case labels.

stat/stat.go
```diff
@@ -1293,3 +1294,76 @@ func MeanVariance(x, weights []float64) (mean, variance float64) {
...
+// This function tests the null hypothesis that the skewness of population that the
+// sample was drawn from is the same as that of a corresponding normal distribution.
+// Implementation based on scipy's: https://github.com/scipy/scipy/blob/v1.3.0/scipy/stats/stats.py#L1481-L1549
+func NormalSkewTest(a []float64) float64 {
+	b2 := Skew(a, nil)
+	n := float64(len(a))
+	if n < 8 {
+		panic(fmt.Sprintf("skewtest is not valid with less than 8 samples, %d samples were given", int(n)))
+	}
+	y := b2 * math.Sqrt(((n+1)*(n+3))/(6.0*(n-2)))
```
kortschak: The `X.0` syntax is not required since the expression is in a float context, throughout.

kortschak:
This addition will require that the Scipy license be added to THIRD_PARTY_LICENSES.

stat/testuv/norm.go
```diff
@@ -0,0 +1,77 @@
+package testuv
```
kortschak: Please add Gonum copyright headers to new files.

stat/testuv/norm.go
```diff
@@ -0,0 +1,77 @@
...
+	"math"
+
+	"gonum.org/v1/gonum/stat"
+)
+
+// NormalSkewTest test whether the skew is different from the normal distribution.
+// This function tests the null hypothesis that the skewness of population that the
+// sample was drawn from is the same as that of a corresponding normal distribution.
+// The length of a must be at least 8.
+func NormalSkewTest(a []float64) float64 {
```
kortschak: I think these can be renamed to remove the `Test` suffix now (also below).

stat/testuv/norm.go
```diff
@@ -0,0 +1,77 @@
...
+	// Implementation based on scipy's: https://github.com/scipy/scipy/blob/v1.3.0/scipy/stats/stats.py#L1481-L1549
+	b2 := stat.Skew(a, nil)
+	n := len(a)
+	if n < 8 {
```
kortschak: I would write this a `len(a) < 8` rather than the intermediate variable. `len(x)` is a constant time operation as costly as `n`. Then you can do the `n := float64(len(a))` below and not need an `nf`.

stat/testuv/norm.go
```diff
@@ -0,0 +1,77 @@
...
+		panic(fmt.Sprintf("testuv: skewtest is not valid with less than 8 samples, %d samples were given", n))
```
kortschak: `panic("testuv: not enough samples for normal skew test")`

stat/testuv/norm.go
```diff
@@ -0,0 +1,77 @@
...
+// NormalKurtosisTest test whether a dataset has normal kurtosis.
+// This function tests the null hypothesis that the kurtosis of the population from
+// which the sample was drawn is that of the normal distribution:
+// `kurtosis = 3(n-1)/(n+1)`.
+// The length of a must be at least 5.
+func NormalKurtosisTest(a []float64) float64 {
+	// Implementation based on scipy's: https://github.com/scipy/scipy/blob/v1.3.0/scipy/stats/stats.py#L1481-L1549
+	n := len(a)
+	if n < 5 {
+		panic(fmt.Sprintf("testuv: kurtosistest requires at least 5 observations; %d observations were given", n))
```
kortschak: panic("testuv: not enough samples for normal kurtosis test")

stat/testuv/norm_test.go
```diff
@@ -0,0 +1,119 @@
+package testuv
```
kortschak: Please add the copyright header.

stat/testuv/norm.go
```diff
@@ -0,0 +1,77 @@
...
+	return (term1 - term2) / math.Sqrt(2/(9*a2))
+}
+
+// NormalTest test whether a sample differs from a normal distribution.
+// This function tests the null hypothesis that a sample comes from a normal distribution.
+// It is based on D'Agostino and Pearson's [1]_, [2]_ test that combines skew and kurtosis to
+// produce an omnibus test of normality.
+// The length of a must be at least 8.
+//
+// References
```
kortschak: We're usually more light weight with references.

Suggestion (but please wait for others' input on this). (Note the change in function name).

```
// Normality tests whether a sample differs from a normal distribution. It returns the
// 2-sided chi squared probability for the hypothesis test.
//
// This function tests the null hypothesis that a sample comes from a normal distribution.
// It is based on D'Agostino and Pearson's test that combines skew and kurtosis to
// produce an omnibus test of normality. See https://doi.org/10.1093/biomet/58.2.341 and
// https://doi.org/10.1093/biomet/60.3.613 for details.
//
// The length of a must be at least 8 otherwise Normality will panic.
```

stat/testuv/norm.go
```diff
@@ -0,0 +1,77 @@
...
+
+// NormalKurtosisTest test whether a dataset has normal kurtosis.
+// This function tests the null hypothesis that the kurtosis of the population from
+// which the sample was drawn is that of the normal distribution:
+// `kurtosis = 3(n-1)/(n+1)`.
+// The length of a must be at least 5.
+func NormalKurtosisTest(a []float64) float64 {
+	// Implementation based on scipy's: https://github.com/scipy/scipy/blob/v1.3.0/scipy/stats/stats.py#L1481-L1549
+	n := len(a)
+	if n < 5 {
```
kortschak: Same as above.

stat/testuv/norm.go
```diff
@@ -0,0 +1,77 @@
...
+	w2 := -1 + math.Sqrt(2*(beta2-1))
+	delta := 1 / math.Sqrt(0.5*math.Log(w2))
+	alpha := math.Sqrt(2 / (w2 - 1))
+	if y == 0 {
+		y = 1
+	}
+	return delta * math.Log(y/alpha+math.Sqrt(math.Pow(y/alpha, 2)+1))
+}
+
+// NormalKurtosisTest test whether a dataset has normal kurtosis.
```
kortschak:
```
// NormalKurtosis tests whether a dataset has normal kurtosis. It returns the
// 2-sided p-value for the hypothesis test.
//
// This function tests the null hypothesis that the kurtosis of the population from
// which the sample was drawn is that of the normal distribution:
//  kurtosis = 3(n-1)/(n+1).
//
// The length of a must be at least 5 otherwise NormalKurtosis will panic.
```

stat/testuv/norm.go
```diff
@@ -0,0 +1,77 @@
+package testuv
+
+import (
+	"fmt"
+	"math"
+
+	"gonum.org/v1/gonum/stat"
+)
+
+// NormalSkewTest test whether the skew is different from the normal distribution.
```
kortschak:
```
// NormalSkew tests whether the skew is different from the normal distribution.
// It returns the 2-sided p-value for the hypothesis test.
//
// This function tests the null hypothesis that the skewness of population that the
// sample was drawn from is the same as that of a corresponding normal distribution.
//
// The length of a must be at least 8 otherwise NormalSkew will panic.
```

stat/testuv/norm.go
```diff
@@ -0,0 +1,77 @@
...
+// This function tests the null hypothesis that the skewness of population that the
+// sample was drawn from is the same as that of a corresponding normal distribution.
+// The length of a must be at least 8.
+func NormalSkewTest(a []float64) float64 {
+	// Implementation based on scipy's: https://github.com/scipy/scipy/blob/v1.3.0/scipy/stats/stats.py#L1481-L1549
```
kortschak: I think the lines should be L1307-L1382.

stat/testuv/norm.go
```diff
@@ -0,0 +1,77 @@
...
+	return delta * math.Log(y/alpha+math.Sqrt(math.Pow(y/alpha, 2)+1))
+}
+
+// NormalKurtosisTest test whether a dataset has normal kurtosis.
+// This function tests the null hypothesis that the kurtosis of the population from
+// which the sample was drawn is that of the normal distribution:
+// `kurtosis = 3(n-1)/(n+1)`.
+// The length of a must be at least 5.
+func NormalKurtosisTest(a []float64) float64 {
+	// Implementation based on scipy's: https://github.com/scipy/scipy/blob/v1.3.0/scipy/stats/stats.py#L1481-L1549
```
kortschak: I think the lines should be L1388-L1475.

stat/testuv/norm_test.go
```diff
@@ -0,0 +1,119 @@
...
+
+import (
+	"math"
+	"testing"
+
+	"golang.org/x/exp/rand"
+	"gonum.org/v1/gonum/stat/distuv"
+)
+
+func getNormalValues() []float64 {
```
kortschak: `func rnorm(n int) []float64` and use `n` as the slice length.

stat/testuv/norm_test.go
```diff
@@ -0,0 +1,119 @@
...
+			values:   []float64{1, 2, 3, 4, 5, 6, 7, 8000},
+			expected: 4.319816401673864,
+		},
+		{
+			values:   []float64{100, 100, 100, 100, 100, 100, 100, 101},
+			expected: 4.319820025201098,
+		},
+	} {
+		z := NormalSkewTest(test.values)
+		if math.Abs(z-test.expected) > 1e-7 {
```
kortschak: `if floats.EqualWithinAbsOrRel(z, test.want, tol, tol) ...` and define a reasonable `const tol`.

stat/testuv/norm_test.go
```diff
@@ -0,0 +1,119 @@
...
+	return nums
+}
+
+func TestNormalSkewTest(t *testing.T) {
+	// Even though the skew test and kurtosis test are a copy from the scipy's the results
+	// are not exactly the same because the skew/kurtosis corrections are not exactly the same
+	// but the statistic value is close enough.
+	for i, test := range []struct {
+		values   []float64
+		expected float64
```
kortschak: s/expected/want/g

seppo0010:
Great. Thanks for the feedback. I'm waiting on other's input about the documentation to apply the changes as requested.

btracey:
Couple of points
1) I'd be okay with calling this `test` rather than `testuv`. Most commonly used statistical tests are univariate, and the ones that aren't often have their own names. (for instance d versus MahalanobisD). I don't do a lot of significance testing, so perhaps I'm missing a bunch of obvious clashes. I am thinking that `test.T` reads a lot better than `testuv.T`. 

2) Maybe I'm overreacting here, but I don't like the naming. `testuv.Normal` seems extremely generic, given that there are many different tests for normality under many different conditions. For instance, other than the fact that scipy does it, is there a good reason to use a K-squared test for normality? There are other possibilities, such as the Anderson-Darling or Shapiro-Wilk. Similarly, this assumes unknown mean and variance, but there are other possibilities. I think it's much better for the functions to be explicit about what they're computing, i.e. `AndersonDarlingNormal` or `DAgonstinoK`. Then, to provide better direction, we could have an example that is TestNormal, which shows someone how to use them

3) I think it is correct that these functions return the measure and not the interpretation of it. I think we should think of how to transform these measures into a boolean or probability estimate where appropriate

4) Many tests, such as Anderson-Darling work for a variety of distributions. We should think on good ways to keep the namespace clean. One option is 
```
type AndersonDarling struct{}

func (AndersonDarling) Normal(x []float64) float64
```

then in use

```
a := test.AndersonDarling{}.Normal(x)
```

I'm not sure though (on how to keep a clean and consistent interface across many tests).

kortschak:
1. SGTM
2. I'll defer to others (@Armadilloa16 do you have any views?)
3. "Interpretation" needs to be more clearly defined. In my mind, conversion to a `bool` is an interpretation, conversion to a P-value is more arguable.
4. Looks like this needs discussion.

seppo0010:
I'm OK with all the comments. I'll wait for more input on 2, 3, and 4.

scipy's version returns, besides the statistic, the p-value. I think it is useful but at the same time it is computation that may not be required. Maybe adding a function to transform the statistic to its pvalue?

romanwerpachowski:
I think the most common use case is that you need the p-value. Using a different function to compute it introduces a possibility for error, by using the wrong conversion. I would default to returning the p-value.

kortschak:
It depends on the API.

romanwerpachowski:
How about:

- XXXTest returns the statistic and p-value
- XXXTestStatistic returns just the statistic, and is called by XXXTest

kortschak:
It will be in a `test` package, so the `Test` infix in not necessary. A proper sketch is needed to see how it looks though.

PR #1074, opened 2019-09-03
ci: update testing for go1.13

kortschak:
Please take a look.


kortschak:
Those declarations are still weird. I will leave it.

kortschak:
For clarity, the go version directive will change if we introduce language feature uses above go1.12 since this is what it's for. How this interacts with multiple version support is still something that has not been properly clarified by the Go team.

chewxy:
shouldn't the go version be switched to 1.13 given 1.13 is the first official version to support modules

kortschak:
Go1.11 does with the experiment turned on, which is how we use them. When there is no requirement for `GO111MODULES`, then I suppose we should update the version directive.

PR #1082, opened 2019-09-09
mat: Expose ReuseAs method to help resizing

### LINKED ISSUE #963: mat: Helper routine(s) for resizing zero matrices
We have a pattern that Zero matrices can be resized. This is both receivers of methods, but also as inputs to functions, for example in `stat` and `optimize`. It would be nice to implement methods necessary to make this pattern easy to implement. This would encourage outside packages to follow the same pattern.

btracey:
Fixes #963

Please take a look.

### DIFF
--- a/mat/dense.go
+++ b/mat/dense.go
@@ -47,7 +47,7 @@ func NewDense(r, c int, data []float64) *Dense {
 		if r == 0 || c == 0 {
 			panic(ErrZeroLength)
 		}
-		panic("mat: negative dimension")
+		panic(ErrNegativeDimension)
 	}
 	if data != nil && r*c != len(data) {
 		panic(ErrShape)
@@ -65,6 +65,27 @@ func NewDense(r, c int, data []float64) *Dense {
 		capRows: r,
 		capCols: c,
 	}
+}
+
+// ReuseAs changes the receiver if it IsZero() to be of size r×c.
+//
+// ReuseAs re-uses the backing data slice if it has sufficient capacity,
+// otherwise a new slice is allocated. The data is then zeroed.
+//
+// ReuseAs panics if the receiver is not zero-sized, and panics if
+// the input sizes are less than one. To zero a matrix for re-use,
+// Reset should be used.
+func (m *Dense) ReuseAs(r, c int) {
+	if r <= 0 || c <= 0 {
+		if r == 0 || c == 0 {
+			panic(ErrZeroLength)
+		}
+		panic(ErrNegativeDimension)
+	}
+	if !m.IsZero() {
+		panic("mat: ReuseAs called on non-zero matrix")
+	}
+	m.reuseAsZeroed(r, c)
 }
 
 // reuseAs resizes an empty matrix to a r×c matrix,
--- a/mat/dense.go
+++ b/mat/dense.go
@@ -12,8 +12,10 @@ import (
 var (
 	dense *Dense
 
-	_ Matrix  = dense
-	_ Mutable = dense
+	_ Matrix      = dense
+	_ allMatrix   = dense
+	_ denseMatrix = dense
+	_ Mutable     = dense
 
 	_ ClonerFrom   = dense
 	_ RowViewer    = dense
@@ -73,7 +75,7 @@ func NewDense(r, c int, data []float64) *Dense {
 // otherwise a new slice is allocated. The data is then zeroed.
 //
 // ReuseAs panics if the receiver is not zero-sized, and panics if
-// the input sizes are less than one. To zero a matrix for re-use,
+// the input sizes are less than one. To zero the receiver for re-use,
 // Reset should be used.
 func (m *Dense) ReuseAs(r, c int) {
 	if r <= 0 || c <= 0 {
@@ -91,9 +93,8 @@ func (m *Dense) ReuseAs(r, c int) {
 // reuseAsNonZeroed resizes an empty matrix to a r×c matrix,
 // or checks that a non-empty matrix is r×c. It does not zero
 // the data in the receiver.
-//
-// reuseAs must be kept in sync with reuseAsZeroed.
 func (m *Dense) reuseAsNonZeroed(r, c int) {
+	// reuseAs must be kept in sync with reuseAsZeroed.
 	if m.mat.Rows > m.capRows || m.mat.Cols > m.capCols {
 		// Panic as a string, not a mat.Error.
 		panic("mat: caps not correctly set")
@@ -120,9 +121,8 @@ func (m *Dense) reuseAsNonZeroed(r, c int) {
 // reuseAsZeroed resizes an empty matrix to a r×c matrix,
 // or checks that a non-empty matrix is r×c. It zeroes
 // all the elements of the matrix.
-//
-// reuseAsZeroed must be kept in sync with reuseAs.
 func (m *Dense) reuseAsZeroed(r, c int) {
+	// reuseAsZeroed must be kept in sync with reuseAsNonZeroed.
 	if m.mat.Rows > m.capRows || m.mat.Cols > m.capCols {
 		// Panic as a string, not a mat.Error.
 		panic("mat: caps not correctly set")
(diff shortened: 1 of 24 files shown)


btracey:
I started the implementation, but wanted to discuss the singular case (documentation and behavior) before doing the work on all of the matrix types

1) I think ReuseAs should be Zeroed. This is going to be far safer in the general case, and I fear not having it be that way will eventually be a source of bugs. If we need it in the future we could add a ReuseAsNonzerod which at least is explicit about the behavior

2) If we're agreed on 1), this is going to get confusing with the unexported code. I propose to make the name shift reuseAs -> reuseAsNZ in this PR (and so all uses will be caught with a compile time error)

mat/dense.go
```diff
@@ -67,6 +67,27 @@ func NewDense(r, c int, data []float64) *Dense {
...
+// Reset should be used.
+func (m *Dense) ReuseAs(r, c int) {
+	if r <= 0 || c <= 0 {
+		if r == 0 || c == 0 {
+			panic(ErrZeroLength)
+		}
+		panic(ErrNegativeDimension)
+	}
+	if !m.IsZero() {
+		panic("mat: ReuseAs called on non-zero matrix")
```
btracey: This should be made one of the Error types

kortschak:
1. Agreed.
2. I don't understand the basis for this possible confusion.

mat/dense.go
kortschak: Yes, with the text "mat: reuse non-zero matrix".

btracey:
The basis is that `ReuseAs` would zero while `reuseAs` would not (and instead `reuseAsZeroed` gets the same behavior)

kortschak:
Ah, yes. Thanks. Agreed.

btracey:
PTAL 

This implements the method for Dense

mat/symmetric.go
```diff
@@ -176,6 +197,11 @@ func (s *SymDense) reuseAs(n int) {
 	}
 }
 
+func (s *SymDense) reuseAsZeroed(n int) {
```
kortschak: This should be fully spelled out to avoid double work on the backing slice.

mat/errors.go
```diff
@@ -114,7 +114,9 @@ type Error struct{ string }
 func (err Error) Error() string { return err.string }
 
 var (
+	ErrNegativeDimension   = Error{"matrix: negative dimension"}
```
kortschak: Should we fix the prefixes here or in another PR?

btracey: Filed https://github.com/gonum/gonum/issues/1084

btracey:
`reuseAs` also allows the matrix to be size rxc. In internal code, this allows  `m.reuseAs(r, c)` without the extra check to the size of the matrix. Do we want to replicate this behavior?

```
func (r *Foo) Bar(dst *mat.Dense){
    if dst.IsZero() {
         dst.ReuseAs(r, c)
    }else {
        if r2, c2 := dst.Dims(); r != r2 || c != c2 {
              panic("receiver shape mismatch")
        }
    }
    // do stuff
}
```

The alternative is just

```
func (f *Foo) Bar(dst *mat.Dense){
     dst.ReuseAs(r, c)
     // do stuff
}
```

The former is clearly more code and so slightly harder to use. The upside is it makes the `ReuseAs` method easier to understand, and will make panic traces easier to parse. 

Thoughts?

mat/symmetric.go
```diff
@@ -176,6 +197,11 @@ func (s *SymDense) reuseAs(n int) {
...
+	s.reuseAs(n)
```
btracey: This needs to change to reuseAsNonZeroed

kortschak:
In the first examples, should that be `dst.reuseAs(r, c)` (unexported)?

mat/symmetric.go
kortschak: I'd rather we don't do it that way but just spell out this completely.

kortschak: Actually, `useZeroed` is incorrect for the general case. So leave this as is and we'll fix `useZeroed` later. The issue is that the `zero` function is unaware of striding and so will clobber other data outside the matrix if they share a backing slice.

kortschak: Updating this, it's not fixable in itself, so this should be updated. We need some additional documentation as a warning on the `ReuseAs` methods. They are a fairly substantial footgun that cannot have a safety.

mat/dense.go
```diff
@@ -67,11 +67,33 @@ func NewDense(r, c int, data []float64) *Dense {
...
-// reuseAs resizes an empty matrix to a r×c matrix,
-// or checks that a non-empty matrix is r×c.
+// ReuseAs changes the receiver if it IsZero() to be of size r×c.
+//
+// ReuseAs re-uses the backing data slice if it has sufficient capacity,
+// otherwise a new slice is allocated. The data is then zeroed.
+//
+// ReuseAs panics if the receiver is not zero-sized, and panics if
+// the input sizes are less than one. To zero a matrix for re-use,
+// Reset should be used.
```
kortschak:
```
//
// ReuseAs is not safe to use if the receiver shares backing data with
// another Matrix.
```

mat/symmetric.go
```diff
@@ -134,6 +134,27 @@ func (s *SymDense) Reset() {
...
 }
 
+// ReuseAsSym changes the receiver if it IsZero() to be of size n×n.
+//
+// ReuseAsSym re-uses the backing data slice if it has sufficient capacity,
+// otherwise a new slice is allocated. The data is then zeroed.
+//
+// ReuseAsSym panics if the receiver is not zero-sized, and panics if
+// the input size is less than one. To zero a matrix for re-use,
+// Reset should be used.
```
kortschak:
```
//
// ReuseAsSym is not safe to use if the receiver shares backing data with
// another Matrix.
```

btracey:
No, I meant the exported function. The question is how do outside packages interact with this function. The goal is that that if I'm implemented `mypgk.Foo`, and want to resize a destination matrix, what is the pattern? (For example, in `stat`, how do we want the checking code to look?)

mat/symmetric.go
btracey: Sorry, that was a note to myself that `SymDense.reuseAs` needs to also be changed to `SymDense.reuseAsNonZeroed`

mat/dense.go
btracey: Isn't this obvious from the behavior? The same is true with `Reset`, `Augment`, `Copy`, `CloneFrom`, etc. Any operation that can change the backing slice is not safe to use when sharing backing data.

kortschak: It is also true for those. I think that given that people mess up `append` and this is more complex, it's worth noting. Probably in the package doc as well that these are possible. The issue is not about shared data though, it's about potentially clearing data that's not obviously shared.

kortschak:
Externally (outside gonum), it should look like the bottom one. Inside gonum, probably the upper one to make the panics clearer.

mat/dense.go
btracey: I mean, it is about shared data, in the sense that if you don't share data there are no problems. There also aren't any problems if you aren't tricky about it (splitting rows and columns of matrices, which is actually reasonably hard to do with the package).

I don't think this function itself presents problems. The dangerous one is `Reset`, which enables `ReuseAs` and 85% of the potential for misuse of shared data (i.e. reshaping receivers and destination matrices).

kortschak: The second paragraph is the issue, yes. Perhaps the warning should be on reset. There the warning would be that by resetting matrix guards against leaky sharing are lost.

mat/dense.go
btracey: Done.

mat/symmetric.go
btracey: Done.

btracey:
Okay. Done.

1) Added some extra warnings to Reset
2) Exposed ReuseAs on the types where we internally check for sizes (Dense, Tri, Sym, Vec). Leaving Diag and the bandeds for later.
3) Added some helper types for making these kinds of consistent interfaces to check in the future. This was made as simple as possible, and not intended to be comprehensive. In particular I did not add denseMatrix to VecDense because VecDense does not implement DiagView or Tracer. I figured it was outside the scope of this PR to talk about that here.

mat/symmetric.go
btracey: Added warnings to Reset instead.

kortschak:
Fixes #1085.

Issue #1081, opened 2019-09-08
mat/all: Matrix receiver inputs should respond to IsZero and not nil


btracey:
stat.Covariance used to check if the input matrix was `nil`, and allocated a new matrix if it was. #705 changed that to instead check if the input matrix is zero and allocate otherwise. This means the usage pattern is 
```
var cov mat.SymDense
stat.CoviarianceMatrix(&cov, data, weights)
```
which is aligned with the similar pattern we've had since the beginning in `mat` for matrix receivers

```
var c mat.Dense
c.Mul(a, b)
```

This issue proposes to change the remaining places we have the `nil` pattern to instead use the `IsZero` pattern. From a quick glance, I think this is mostly in the `FooTo` patterns in `Cholesky`, `SVD`, etc. It is also present in `distmat`, i.e `Wishart.MeanSym`

Note that while this is a change for some of the functions, we do already use the proposed pattern in a number of places, `distmv.Normal.CovarianceMatrix` for example. `Cholesky.SolveVec` is not documented what happens when `dst` is nil, but the current behavior is to resize.

I think implementing this change would make it consistent across the board. #963 can also be implemented to make it easy to help others conform to this style.

btracey:
RFC @kortschak @vladimir-ch

kortschak:
SGTM

Thank you for finding where that was discussed.

vladimir-ch:
SGTM, too.

kortschak:
Yay!

PR #1088, opened 2019-09-15
mat,all: Rename IsZero to IsEmpty

### LINKED ISSUE #1083: mat: Zero() is does not make IsZero() return true.
Right now we have `IsZero` which checks if the receiver is zero-sized, and `Zero`, which sets the matrix elements to zero. It is confusing that calling `Zero` does not make `IsZero` return true. I suggest we rename `IsZero` to `IsEmpty`. This removes the implicit connection, and `IsReset` seems linguistically awkward.

btracey:
This avoids the confusion between Zero() and IsZero() which sounds like they should be related
to one another but are not. This makes IsEmpty the counterpart to Reset. Add check for Zero in allMatrix

Fixes #1083.
Updates #1081.

Please take a look.

### DIFF
--- a/mat/matrix.go
+++ b/mat/matrix.go
@@ -35,7 +35,8 @@ type Matrix interface {
 // Dense types, especially helpful when adding new features.
 type allMatrix interface {
 	Reseter
-	IsZero() bool
+	IsEmpty() bool
+	Zero()
 }
 
 // denseMatrix represents the extra set of methods that all Dense Matrix types
--- a/diff/fd/hessian.go
+++ b/diff/fd/hessian.go
@@ -18,12 +18,12 @@ import (
 // options are specified by settings. If settings is nil, the Hessian will be
 // estimated using the Forward formula and a default step size.
 //
-// If the dst matrix is zero-sized it will be resized to the correct dimensions,
+// If the dst matrix is empty it will be resized to the correct dimensions,
 // otherwise the dimensions of dst must match the length of x or Hessian will panic.
 // Hessian will panic if the derivative order of the formula is not 1.
 func Hessian(dst *mat.SymDense, f func(x []float64) float64, x []float64, settings *Settings) {
 	n := len(x)
-	if dst.IsZero() {
+	if dst.IsEmpty() {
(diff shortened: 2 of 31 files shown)


btracey:
I went through mat and manually checked all of the instances of "zero" and "Zero" and made them consistent with the new usage (or at least tried). Outside of mat, I only fixed the places needed to compile. I'll go through the rest in fixing #1081

mat/matrix.go
```diff
@@ -35,7 +35,8 @@ type Matrix interface {
 // Dense types, especially helpful when adding new features.
 type allMatrix interface {
 	Reseter
-	IsZero() bool
+	IsEmpty() bool
+	Zero()
```
kortschak: Why was this added?

btracey: Because all of the types implement it and it's relevant to the update?

kortschak: They all implement it, but it's not mentioned in the commit message.

btracey: Done.

kortschak: It seems not entirely related, but I'll let it go.

btracey:
I thought it highlighted the confusion between Zero and IsZero

ggggggggg:
This is a breaking change. Semantic versioning would require a version bump for this change.

btracey:
Yes, sorry. We have https://github.com/gonum/gonum/issues/1089 to make this easier in the future, which will be done soon.

kortschak:
@ggggggggg no. We are pre-1.0 and we note on the first line of the readme that we are unstable as shown by this badge [stability-unstable](https://github.com/emersion/stability-badges#unstable).

ggggggggg:
I see. I misunderstood the import convention. I thought because ‘v1’ was in
the import path it was 1.0. I’m guessing this is part of the transition to
go modules, the docs say that if ‘v2’ is in the path is has to be version 2
but nothing for lower versions.

Sorry for the noise.

kortschak:
No worries. The v1 actually predates modules and was our attempt to mitigate version problems in the future. It was presumptive and unfortunately a little temporally optimistic.

PR #1095, opened 2019-09-23
stat/distmat: change methods to accept empty instead of nil matrices

kortschak:
Please take a look.

At the moment, `RandCholTo` doesn't care about sizes. Do we want it to?

Updates #1081.

### DIFF
--- a/stat/distmat/wishart.go
+++ b/stat/distmat/wishart.go
@@ -69,7 +69,7 @@ func NewWishart(v mat.Symmetric, nu float64, src rand.Source) (*Wishart, bool) {
 	return w, true
 }
 
-// MeanSymTo calculates the mean matrix of the distribution in and stored it in dst.
+// MeanSymTo calculates the mean matrix of the distribution in and stores it in dst.
 // If dst is empty, it is resized to be an d×d symmetric matrix where d is the order
 // of the receiver. When dst is non-empty, MeanSymTo panics if dst is not d×d.
 func (w *Wishart) MeanSymTo(dst *mat.SymDense) {
@@ -150,8 +150,10 @@ func (w *Wishart) RandSymTo(dst *mat.SymDense) {
 	c.ToSym(dst)
 }
 
-// RandChol generates the Cholesky decomposition of a random matrix from the distribution.
+// RandCholTo generates the Cholesky decomposition of a random matrix from the distribution.
 func (w *Wishart) RandCholTo(dst *mat.Cholesky) {
+	// TODO(kortschak): Make RandCholTo care about the size of dst.
+
 	// TODO(btracey): Modify the code if the underlying data from c is exposed
 	// to avoid the dim^2 allocation here.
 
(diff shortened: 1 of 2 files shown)


stat/distmat/wishart.go
```diff
@@ -69,22 +69,18 @@ func NewWishart(v mat.Symmetric, nu float64, src rand.Source) (*Wishart, bool) {
...
-// If x is nil, a new matrix is allocated and returned. If x is not nil, the
-// result is stored in-place into x and MeanSym will panic if the order of x
-// is not equal to the order of the receiver.
-func (w *Wishart) MeanSym(x *mat.SymDense) *mat.SymDense {
-	if x == nil {
-		x = mat.NewSymDense(w.dim, nil)
-	}
-	d := x.Symmetric()
-	if d != w.dim {
+// MeanSymTo calculates the mean matrix of the distribution in and stored it in dst.
```
vladimir-ch: "... of the distribution and stores it in dst."?

stat/distmat/wishart.go
```diff
@@ -145,19 +141,17 @@ func (w *Wishart) logProbSymChol(cholX *mat.Cholesky) float64 {
...
+func (w *Wishart) RandSymTo(dst *mat.SymDense) {
 	var c mat.Cholesky
-	w.RandChol(&c)
-	c.ToSym(x)
-	return x
+	w.RandCholTo(&c)
+	c.ToSym(dst)
 }
 
 // RandChol generates the Cholesky decomposition of a random matrix from the distribution.
```
vladimir-ch: RandCholTo

Also s/mat64/mat in the comment below.

stat/distmat/wishart.go
kortschak: Thanks

stat/distmat/wishart.go
kortschak: Done

kortschak:
Making `RandCholTo` care about sizes requires that we can tell if a `mat.Cholesky` is empty. We can't do that right now, so I have placed a TODO.

kortschak:
The other option is to make `Cholesky.SetFromU` care, which is easier.

btracey:
`Cholesky.SetFromU` definitely should care. Filed #1097

stat/distmat/wishart.go
```diff
@@ -145,19 +141,19 @@ func (w *Wishart) logProbSymChol(cholX *mat.Cholesky) float64 {
...
+	c.ToSym(dst)
 }
 
-// RandChol generates the Cholesky decomposition of a random matrix from the distribution.
-func (w *Wishart) RandChol(c *mat.Cholesky) *mat.Cholesky {
+// RandCholTo generates the Cholesky decomposition of a random matrix from the distribution.
+func (w *Wishart) RandCholTo(dst *mat.Cholesky) {
+	// TODO(kortschak): Make RandCholTo care about the size of dst.
+
 	// TODO(btracey): Modify the code if the underlying data from c is exposed
```
btracey: s/`c`/`dst`

kortschak: Should I change this now, or when I remove the TODO above with the change at Cholesky?

btracey:
If you're going to make that PR now then doesn't really matter. It's better here because that's when `dst` changed names (and the TODO isn't going away), but I'm fine either way.

kortschak:
I need to think about the Cholesky change, so it won't happen straight away. I'll fix this now.

btracey:
Sounds good. There are two questions in the Cholesky change. The first is
if you should be able to resize an empty matrix, the second is if you
should be able to SetFromU with the wrong size. I'm not sure about the
first, but I'm pretty sure we shouldn't allow the second.

PR #1099, opened 2019-09-23
mat: add fastpath for SymBandDense to VecDense.MulVec

vladimir-ch:
Please take a look.

### DIFF
--- a/mat/shadow.go
+++ b/mat/shadow.go
@@ -67,6 +67,8 @@ func (m *Dense) checkOverlapMatrix(a Matrix) bool {
 		amat = ar.RawMatrix()
 	case RawSymmetricer:
 		amat = generalFromSymmetric(ar.RawSymmetric())
+	case RawSymBander:
+		amat = generalFromSymmetricBand(ar.RawSymBand())
 	case RawTriangular:
 		amat = generalFromTriangular(ar.RawTriangular())
 	case RawVectorer:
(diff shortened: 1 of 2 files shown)


mat/vector.go
```diff
@@ -595,6 +595,11 @@ func (v *VecDense) MulVec(a Matrix, b Vector) {
 		}
 		v.setVec(0, sum)
 		return
+	case *SymBandDense:
+		if fast {
```
btracey: Don't we need to check overlap? It seems like in general it would be best to do this before the switch though.

mat/shadow.go
```diff
@@ -196,3 +202,41 @@ func generalFromVector(a blas64.Vector, r, c int) blas64.General {
 		Data:   a.Data,
 	}
 }
+
+func (s *SymBandDense) checkOverlap(a blas64.General) bool {
+	return checkOverlap(generalFromSymmetricBand(s.RawSymBand()), a)
+}
+
+func (s *SymBandDense) checkOverlapMatrix(a Matrix) bool {
```
golangcibot: func `(*SymBandDense).checkOverlapMatrix` is unused (from `unused`)

mat/vector.go
vladimir-ch: I've now added overlap detection for SymBandDense (@kortschak could you please take a look?) and at the moment use it here inside the switch case.

mat/shadow.go
vladimir-ch: Yea, maybe this can be added later when it's actually needed, right now we don't have any arithmetic operations on `SymBandDense`.

kortschak: We can ignore the bot here.

mat/vector.go
kortschak: This looks fine.

Issue #400, opened 2018-01-28
Proposal: implement simpson's rule


millerick:
It's been a long while since I've implemented numerical methods, and I wanted to get back to my undergrad roots.  I was looking at the `integrate` portion of the project, and found it odd that there is a very basic trapezoidal rule implemented, robust quadratures, and nothing in between.  Would it be acceptable to add simpson's rule to the integrate package?

sbinet:
Sure, SGTM :)

kortschak:
Closed by #1107.

PR #1100, opened 2019-09-24
integrate: add Simpsons Rule

nisfornature:
This change provides an extended Simpson's rule implementation in the num/integrate package.

 #400.

### DIFF
--- a/integrate/simpsons.go
+++ b/integrate/simpsons.go
@@ -1,0 +1,104 @@
+// Copyright ©2019 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package integrate
+
+import (
+	"sort"
+)
+
+// Simpson's method approximates the integral of a function f(x)
+// by means of subdividing the interval of integration into segments
+// and applying a method that fits a polynomial to each subinterval.
+// this implementation makes use of the following composite Simpson's method
+// to estimate \int_a^b f(x) dx:
+//  \sum_{i=0}^{(N/2)-1} {a_0}_i * f_{2i-1} + {a_1}_i * f_{2i} + {a_2}_i * f_{2i+1}
+// where N is the number of subintervals and the number of subintervals
+// is even and {a_0}_i, {a_1}_i, and {a_2}_i are constants at index i given by:
+//  {a_0}_i = 2 * h^{3}_{2i} - h^{3}_{2i+1} + 3 * h_{2i+1} * h^{2}_{2i} /
+//            6 * h_{2i} * (h_{2i+1} + h_{2i})
+//  {a_1}_i = h^{3}_{2i+1} + h^{3}_{2i} +
+//            3 * h_{2i+1} * (h_{2i+1} + h_{2i}) /
+//            6 * h_{2i+1} * h_{2i}
+//  {a_2}_i = 2 * h^{3}_{2i+1} - h^{3}_{2i} + 3 * h_{2i} * h^{2}_{2i+1} /
+//            6 * h_{2i+1} * (h_{2i+1} + h_{2i})
+// where h_{k} is the length of the subinterval at k.
+//
+// The formula above approximates the integral of function of f
+// if N is an even number of subintervals. If the number of subintervals are odd,
+// the subintervals up i=0..n-2 are given by the above
+// and the approximations over the second to last and last subintervals are given by:
+//  {a_0}_i * f_{N-2} + {a_1}_i * f_{N-1} + {a_2}_i * f_{N}
+// where the coefficients are:
+//  {a_0}_i = -1 * h^{3}_{N-1} /
+//             6 * h_{N-2} * (h_{N-2} + h_{N-1})
+//  {a_1}_i = h^{2}_{N-1} + 3 * h^{3}_{N-1} * h_{N-2} /
+//            6 * h_{N-2}
+//  {a_2}_i = 2 * h^{2}_{N-1} + 3 * h^{2}_{N-1} * h_{N-2} /
+//            6 * (h_{N-2} + h_{N-1})
+// More information is available at:
+// https://en.wikipedia.org/wiki/Simpson%27s_rule#Composite_Simpson's_rule_for_irregularly_spaced_data
+//
+// The (x,f) input data points must be sorted along x. One can use stat.SortWeighted to do that.
+// The x and f slices must be of equal length and have length > 1.
+func Simpsons(x, f []float64) float64 {
+	validateInterval(x, f)
+
+	var integral float64
+
+	// If only three points are supplied, approximate the integral
+	// as a midpoint and interpolate with Simpson's method.
+	if len(x) == 3 {
+		// Panic if either subinterval coordinate points match
+		if x[0] == x[1] && f[0] == f[1] || x[1] == x[2] && f[0] == f[1] {
+			panic("integrate: at least three unique points are required")
+		}
+		return (x[2] - x[0]) * (f[0]/6.0 + 4.0*f[1]/6.0 + f[2]/6.0)
+	}
+
+	for i := 1; i < len(x)-1; i += 2 {
+		h0 := x[i] - x[i-1]
+		h0p2 := h0 * h0
+		h0p3 := h0 * h0 * h0
+		h1 := x[i+1] - x[i]
+		h1p2 := h1 * h1
+		h1p3 := h1 * h1 * h1
+		hph := h0 + h1
+		f0 := f[i-1]
+		f1 := f[i]
+		f2 := f[i+1]
+		a0 := (2*h0p3 - h1p3 + 3*h1*h0p2) / (6 * h0 * hph)
+		a1 := (h1p3 + h0p3 + 3*h1*h0*hph) / (6 * h1 * h0)
+		a2 := (2*h1p3 - h0p3 + 3*h0*h1p2) / (6 * h1 * hph)
+		integral += a0 * f0
+		integral += a1 * f1
+		integral += a2 * f2
+	}
+
+	if (len(x)-1)%2 != 0 {
+		h0 := x[len(x)-2] - x[len(x)-3]
+		h1 := x[len(x)-1] - x[len(x)-2]
+		h1p2 := h1 * h1
+		h1p3 := h1 * h1 * h1
+		hph := h1 + h0
+		a0 := -1 * h1p3 / (6 * h0 * hph)
+		a1 := (h1p2 + 3*h1*h0) / (6 * h0)
+		a2 := (2*h1p2 + 3*h1*h0) / (6 * hph)
+		integral += a0 * f[len(x)-3]
+		integral += a1 * f[len(x)-2]
+		integral += a2 * f[len(x)-1]
+	}
+	return integral
+}
+
+func validateInterval(x, f []float64) {
+	switch {
+	case len(x) != len(f):
+		panic("integrate: slice length mismatch")
+	case len(x) <= 2:
+		panic("integrate: input data too small")
+	case !sort.Float64sAreSorted(x):
+		panic("integrate: must be sorted")
+	}
+}
--- a/integrate/simpsons_test.go
+++ b/integrate/simpsons_test.go
@@ -1,0 +1,217 @@
+// Copyright ©2019 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package integrate
+
+import (
+	"math"
+	"testing"
+
+	"gonum.org/v1/gonum/floats"
+)
+
+func TestSimpsons(t *testing.T) {
+	const (
+		n   = 1e6
+		tol = 1e-12
+	)
+	x := floats.Span(make([]float64, n), 0, 1)
+	for i, test := range []struct {
+		x         []float64
+		f         func(x float64) float64
+		want      float64
+		tolerance float64
+	}{
+		{
+			x:         floats.Span(make([]float64, n), 0, 1),
+			f:         func(x float64) float64 { return math.Pi },
+			want:      math.Pi,
+			tolerance: 1e-10,
+		},
+		{
+			x:         floats.Span(make([]float64, n), 0, 1),
+			f:         func(x float64) float64 { return 1.0 },
+			want:      1,
+			tolerance: 1e-10,
+		},
+		{
+			x: []float64{0.0, 0.1, 1.0},
+			f: func(x float64) float64 {
+				switch {
+				case 1e-31 <= x || x <= 1:
+					return 1.0
+				default:
+					return 0
+				}
+			},
+			want:      1,
+			tolerance: tol,
+		},
+		{
+			x:         floats.Span(make([]float64, 3), 0, 2),
+			f:         func(x float64) float64 { return 2*x + 0.5 },
+			want:      5,
+			tolerance: tol,
+		},
+		{
+			x:         floats.Span(make([]float64, 10), 0, 2),
+			f:         func(x float64) float64 { return 2*x + 0.5 },
+			want:      5,
+			tolerance: tol,
+		},
+		{
+			x:         floats.Span(make([]float64, n), 0, 2),
+			f:         func(x float64) float64 { return 2*x + 0.5 },
+			want:      5,
+			tolerance: tol,
+		},
+		{
+			x:         x,
+			f:         func(x float64) float64 { return x },
+			want:      0.5,
+			tolerance: tol,
+		},
+		{
+			x:         floats.Span(make([]float64, n), -1, 1),
+			f:         func(x float64) float64 { return x },
+			want:      0,
+			tolerance: tol,
+		},
+		{
+			x:         x,
+			f:         func(x float64) float64 { return x + 10 },
+			want:      10.5,
+			tolerance: tol,
+		},
+		{
+			x:         x,
+			f:         func(x float64) float64 { return 3*x*x + 10 },
+			want:      11,
+			tolerance: tol,
+		},
+		{
+			x:         x,
+			f:         func(x float64) float64 { return math.Exp(x) },
+			want:      1.7182818284591876,
+			tolerance: tol,
+		},
+		{
+			x:         floats.Span(make([]float64, n), 0, math.Pi),
+			f:         func(x float64) float64 { return math.Cos(x) },
+			want:      0,
+			tolerance: tol,
+		},
+		{
+			x:         floats.Span(make([]float64, n), 0, 2*math.Pi),
+			f:         func(x float64) float64 { return math.Cos(x) },
+			want:      0,
+			tolerance: tol,
+		},
+		{
+			x:         floats.Span(make([]float64, n*10), 0, math.Pi),
+			f:         func(x float64) float64 { return math.Sin(x) },
+			want:      2,
+			tolerance: tol,
+		},
+		{
+			x:         floats.Span(make([]float64, n*10), 0, 0.5*math.Pi),
+			f:         func(x float64) float64 { return math.Sin(x) },
+			want:      1,
+			tolerance: tol,
+		},
+		{
+			x:         floats.Span(make([]float64, n), 0, 2*math.Pi),
+			f:         func(x float64) float64 { return math.Sin(x) },
+			want:      0,
+			tolerance: tol,
+		},
+		{
+			x: join(floats.Span(make([]float64, 3), 0, math.Pi/3),
+				[]float64{4 * math.Pi / 10},
+				floats.Span(make([]float64, 4), 3*math.Pi/4, math.Pi)),
+			f:         func(x float64) float64 { return math.Sin(x) },
+			want:      2,
+			tolerance: 1e-2,
+		},
+		{
+			x: join(floats.Span(make([]float64, 30), 0, 5*math.Pi/16),
+				floats.Span(make([]float64, 100), 3*math.Pi/8, math.Pi)),
+			f:         func(x float64) float64 { return math.Sin(x) },
+			want:      2,
+			tolerance: 1e-4,
+		},
+		{
+			x: join(floats.Span(make([]float64, 1e5), 0, 15),
+				floats.Span(make([]float64, 1e5), 23, 40),
+				floats.Span(make([]float64, 1e5), 50, 80),
+				floats.Span(make([]float64, 1e5), 90, 100)),
+			f:         func(x float64) float64 { return 2 * x },
+			want:      10000,
+			tolerance: 1e-9,
+		},
+	} {
+		y := make([]float64, len(test.x))
+		for i, v := range test.x {
+			y[i] = test.f(v)
+		}
+		v := Simpsons(test.x, y)
+		if !floats.EqualWithinAbs(v, test.want, test.tolerance) {
+			t.Errorf("test #%d: got=%v want=%f\n", i, v, float64(test.want))
+		}
+	}
+}
+
+func TestSimpsonsHandlesNaN(t *testing.T) {
+	x := floats.Span(make([]float64, 1e3), 0, 1)
+	y := floats.Span(make([]float64, 1e3), 0, math.NaN())
+
+	v := Simpsons(x, y)
+	if !math.IsNaN(v) {
+		t.Error("integrate: test expects a NaN from the calling function")
+	}
+}
+
+func TestSimpsonsPanics(t *testing.T) {
+	for i, test := range []struct {
+		x    []float64
+		y    []float64
+		want string
+	}{
+		{
+			x:    floats.Span(make([]float64, 100), 0, 2*math.E),
+			y:    []float64{0, 1, 2, 3},
+			want: "integrate: slice length mismatch",
+		},
+		{
+			x:    []float64{0},
+			y:    []float64{1e2},
+			want: "integrate: input data too small",
+		},
+		{
+			x:    []float64{1.0, 0.9, 0.8, 0.7, 0.6, 0.5, 0.4, 0.2, 0.3, 0.1},
+			y:    []float64{0, 0, 0, 0, 0, 0, 0, 0, 0, 0},
+			want: "integrate: must be sorted",
+		},
+		{
+			x:    []float64{0, 0, 1},
+			y:    []float64{0, 0, 1},
+			want: "integrate: at least three unique points are required",
+		},
+	} {
+		defer func() {
+			if r := recover(); r == nil {
+				t.Errorf("test #%d did not panic: expectedPanic=\"%s\"", i, test.want)
+			}
+		}()
+		Simpsons(test.x, test.y)
+	}
+}
+
+func join(slices ...[]float64) []float64 {
+	var c []float64
+	for _, s := range slices {
+		c = append(c, s...)
+	}
+	return c
+}


integrate/simpsons.go
```diff
@@ -0,0 +1,62 @@
...
+// https://en.wikipedia.org/wiki/Simpson's_rule
+//
+// The (x,f) input data points must be sorted along x.
+// One can use github.com/gonum/stat.SortWeighted to do that.
+// The x and f slices must be of equal length and have length > 1.
+
+func Simpsons(x, f []float64) float64 {
+	validateInterval(x, f)
+
+	var coefficient    float64 = 0.0
```
golangcibot: ineffectual assignment to `coefficient` (from `ineffassign`)

btracey:
Sorry, i'll try to get to this soon, I've been sick this week.

integrate/simpsons.go
```diff
@@ -0,0 +1,64 @@
...
+// to estimate
+//  \int_x[1]^x[n] f(x) dx
+// where f[i] is the value of the function at x[i].
+// The form and mathematical formulae of extended Simpson's rule can be found at:
+// https://en.wikipedia.org/wiki/Simpson's_rule
+//
+// The (x,f) input data points must be sorted along x.
+// One can use github.com/gonum/stat.SortWeighted to do that.
+// The x and f slices must be of equal length and have length > 1.
+func Simpsons(x, f []float64) float64 {
```
vladimir-ch: The signature makes it appear as if the interior elements of `x` matter but they do not. I think the function as it is, if we decide to accept this, should take the endpoints of the integration interval, the number of subintervals and the function.

integrate/simpsons.go
```diff
@@ -0,0 +1,64 @@
...
+			coefficient = 59
+		case i == 3 || i == len(x)-3:
+			coefficient = 43
+		case i == 4 || i == len(x)-3:
+			coefficient = 49
+		default:
+			coefficient = 48
+		}
+
+		integral += coefficient * f[i]
```
vladimir-ch: I'm concerned about the numerical accuracy of this approach; sum many potentially large numbers and in the end divide them by something large.

integrate/simpsons_test.go
```diff
@@ -0,0 +1,82 @@
...
+			f:    func(x float64) float64 { return math.Sin(x) },
+			want: 0,
+		},
+	} {
+		y := make([]float64, len(test.x))
+		for i, v := range test.x {
+			y[i] = test.f(v)
+		}
+		v := Simpsons(test.x, y)
+		if !floats.EqualWithinAbs(v, test.want, 1e-4) {
```
vladimir-ch: The same test for `Trapezoidal` uses `1e-12`. Why such a loose tolerance here? Maybe it's the inaccuracy I suspect above? This does not seem to accurately integrate even a linear function.

integrate/simpsons.go
nisfornature: f should be ordered. Ordering is determined by the ordering of values in x. f is not a function. The subintervals are each element of f.

integrate/simpsons_test.go
nisfornature: Simpson's rule does not yield exact values for linear functions. We are making approximations with quadratic polynomials (parabolas). To describe f as points on a plane, this model best fits when the formed path/trajectory is curved and smooth (e.g. has arcs and not ridges, is quadratic and not linear).

I feel num:integrate models should be tested against the same corpus of functions. That being said, Simpson's rule will likely yield greater error than methods of numerical integration like Trapezium and Gaussian Quadrature.

If we were fuzzing these functions under test, I'd argue tolerance should be adjusted to appropriately fit each model, for example.

integrate/simpsons.go
vladimir-ch: Regarding `f`, you're right, I was stuck in thinking in terms of `integrate.Fixed` API. Regarding `x`, I'm not convinced, it's too easy to make a mistake. Either the algorithm should use the elements of `x` or the signature should be different.

integrate/simpsons_test.go
vladimir-ch: If the formula can integrate exactly quadratic functions, then it can also integrate exactly linear and constant functions. How could it deal with the linear and constant terms otherwise? Linear and constant functions _are_ smooth.

For this reason, I agree that each test case should have it's own tolerance. Then we could set very stringent tolerances for functions that should be integrated exactly.

However, the reason for needing `1e-4` here is much simpler: there are several bugs in the algorithm. I'll comment on corresponding lines. When those are fixed, then with N=9 and tolerance `1e-12` only test cases 4, 6, 7, 8 fail. With N=100 only 6, 7, 8 fail. And with N=1000 all succeed.

integrate/simpsons.go
```diff
@@ -0,0 +1,64 @@
...
+	var coefficient, integral float64
+
+	xInitial := x[0]
+	xFinal := x[len(x)-1]
+
+	for i := 0; i < len(x); i++ {
+		switch {
+		case i == 0 || i == len(x)-1:
+			coefficient = 17
+		case i == 2 || i == len(x)-2:
```
vladimir-ch: `i == 1`

integrate/simpsons.go
```diff
@@ -0,0 +1,64 @@
...
+			coefficient = 59
+		case i == 3 || i == len(x)-3:
```
vladimir-ch: `i == 2`

integrate/simpsons.go
```diff
@@ -0,0 +1,64 @@
...
+			coefficient = 43
+		case i == 4 || i == len(x)-3:
```
vladimir-ch: `i == 3 || i == len(x)-4`

integrate/simpsons.go
```diff
@@ -0,0 +1,64 @@
...
+			coefficient = 49
+		default:
+			coefficient = 48
+		}
+
+		integral += coefficient * f[i]
+	}
+
+	return (xFinal - xInitial) / float64(len(x)) * (1.0 / 48.0) * integral
```
vladimir-ch: `float64(len(x)-1)`

integrate/simpsons_test.go
nisfornature: Yes, but not all functions we are interested in attempting to integrate are smooth. Please keep in mind those coefficients are not 'arbitrary constants'. 1) I feel one should never modify functions according to test cases. 2) By changing this implementation in that spirit I am concerned we are at risk of fitting an already extended Simpson's rule implementation to a group of equations.

vladimir-ch: > Yes, but not all functions we are interested in attempting to integrate are smooth.

My comment said something different. Please read the discussion again. But now that you bring this up, I think we can say that all functions we can hope to integrate accurately with these simple methods have to be smooth and well-behaved. Simpson's rule and other similar methods are based on interpolating the sample points by polynomials, therefore they cannot approximate the underlying function well if it is not smooth or is even discontinuous. If there is such a point, then one should split the integration interval.

I don't understand the rest of your comment. I'm not saying you should change the weights, I'm saying that you should fix the bugs that I pointed out.

integrate/simpsons.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+func validateInterval(x, f []float64) {
+	switch {
+	case len(x) != len(f):
+		panic("integrate: slice length mismatch")
+	case len(x) < 2:
```
vladimir-ch: What if the input has less than 8 points? Then the weights of the formula don't sum to 1 and the current implementation of `Simpsons` cannot integrate exactly even constant functions.

vladimir-ch:
Apart from the implementation issues in this PR, do we want to pursue the direction of adding a new function for each method that exists? The functionality is basically the same as `integrate.Fixed`, only that the input is function samples instead of the function itself. Maybe we could have something like `integrated.Sampled`, @btracey ?

btracey:
Sorry, coming out of the disease-haze. I agree with @vladimir-ch that it is not good for this function to 'assume' that the x values are in the correct locations. I would propose that instead function implement the irregularly-spaced Simpson's rule: https://en.wikipedia.org/wiki/Simpson%27s_rule#Composite_Simpson's_rule_for_irregularly_spaced_data

If this suggestion is adopted, then I think the difference is clear between this and `quad.Fixed`. The base `integrate` (implicit) interface says "Here are some x and f locations that I've calculated, please estimate the integral". The `quad.Fixed` methods say "you give me a function, and I'm going to tell you both where to evaluate and estimate the integral". Something like Gaussian quadrature only works if you have samples at specific locations. Simpson's rule can work in both, if you allow for irregularly-spaced data.

vladimir-ch:
> ... I think the difference is clear between this and quad.Fixed.

In other words, as entry points to `integrate` the plan is to have `integrate.Fixed`, maybe in a far future `integrate.Adaptive`, and then one function for every method that takes function sample instead of a function, like `integrate.Trapezoidal` and `integrate.Simpsons`?

btracey:
I think that could be plausible, but I don't know if there's a need to duplicate everything. Do you really gain that much from an `intergrate.Simpsons` (or at least non-adaptive)? Note that it doesn't mean two separate implementations of the algorithm, since `integrate` can generate the `x` and `f` values and then call `quad`.

nisfornature:
Thanks for providing points and clarification @kortschak and @vladimir-ch . My apologies.

@btracey, as you suggested, I have implemented the irregularly spaced composite Simpson's method for this PR. If only three points are provided, I opted to perform interpolation. 

If two points are supplied, I opted to error. I'm not sure if linear interpolation should be performed or a stationary point should be approximated in the event only two points are supplied.

integrate/simpsons.go
nisfornature: I've made updates to make use of elements of x

integrate/simpsons.go
nisfornature: Thank you

integrate/simpsons.go
```diff
@@ -0,0 +1,124 @@
...
+		f1 := f[i]
+		f2 := f[i+1]
+
+		a0 := (2*math.Pow(h1, 3) - math.Pow(h0, 3) + 3*h0*math.Pow(h1, 2)) /
+			(6 * h1 * hph)
+		a1 := (math.Pow(h1, 3) + math.Pow(h1, 3) + 3*h1*h0*hph) /
+			(6 * h1 * h0)
+		a2 := (2*math.Pow(h0, 3) - math.Pow(h1, 3) + 3*h1*math.Pow(h0, 2)) /
+			(6 * h0 * hph)
+		integral += a0 * f0
```
vladimir-ch: The current code doesn't match the formula in the Wikipedia article. According to that formula your `a0` should multiply the rightmost value `f2` and `a2` the leftmost value `f0`.

integrate/simpsons.go
```diff
@@ -0,0 +1,124 @@
...
+	// If only three points are supplied,
+	// approximate the integral
+	// assuming the seccond coordinate points
+	// suffices as a midpoint and interpolate.
+	if len(x) == 3 {
+		return (x[2] - x[0]) * (f[0]/6 + 4/6*f[1] + 4*f[2]/6)
+	}
+
+	// Compute difference as coordinates are traversed
+	diff := []float64{}
```
vladimir-ch: This slice is not necessary, you ever need only it's last two elements. If you don't want to compute `h0` and `h1` anew in every iteration, just hoist them out of the loop and update them.

integrate/simpsons.go
```diff
@@ -0,0 +1,124 @@
...
+	validateInterval(x, f)
+
+	var integral float64
+
+	// If only three points are supplied,
+	// approximate the integral
+	// assuming the seccond coordinate points
+	// suffices as a midpoint and interpolate.
+	if len(x) == 3 {
+		return (x[2] - x[0]) * (f[0]/6 + 4/6*f[1] + 4*f[2]/6)
```
vladimir-ch: `4/6` evaluates to 0, and the last term shouldn't have `4*`. These bugs are not catched by the tests because this line is not covered by them. It should. In addition, this again ignores the location of the middle sample, it should not.

integrate/simpsons_test.go
```diff
@@ -0,0 +1,82 @@
...
+import (
+	"math"
+	"testing"
+
+	"gonum.org/v1/gonum/floats"
+)
+
+func TestSimpsons(t *testing.T) {
+	const N = 1e6
+	x := floats.Span(make([]float64, N), 0, 1)
```
vladimir-ch: This should create unevenly spaced locations, it would have revealed the mistakes in the formulas which do not manifest themselves with uniform subdivison.

integrate/simpsons_test.go
```diff
@@ -0,0 +1,82 @@
...
+
+import (
+	"math"
+	"testing"
+
+	"gonum.org/v1/gonum/floats"
+)
+
+func TestSimpsons(t *testing.T) {
+	const N = 1e6
```
vladimir-ch: This is unnecessarily large.

integrate/simpsons_test.go
```diff
@@ -0,0 +1,82 @@
...
+	x := floats.Span(make([]float64, N), 0, 1)
+	for i, test := range []struct {
```
vladimir-ch: I would include in the tests:
* test cases that have a small number of locations, say `3 <= len(x) <= 12`
* a constant function
* test-case specific tolerance. A global tolerance must accommodate the least accurate result which might hide bugs.

integrate/simpsons.go
```diff
@@ -0,0 +1,124 @@
...
+
+		h0 := diff[i-1]
+		h1 := diff[i]
+		hph := h1 + h0
+
+		f0 := f[i-1]
+		f1 := f[i]
+		f2 := f[i+1]
+
+		a0 := (2*math.Pow(h1, 3) - math.Pow(h0, 3) + 3*h0*math.Pow(h1, 2)) /
```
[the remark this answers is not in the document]
nisfornature: Sure one sec. I have time today.

integrate/simpsons.go
nisfornature: Thanks for catching this.

I'm abusing Newton-Cotes as a general rule here, but I feel like this equation isn't ignoring the midpoint, so to speak?

I've added a condition to ensure uniqueness between the points if only three are provided.

When performing Lagrangian interpolation for weight values to construct a polynomial with three points a=x[0], m=x[1], and b=x[2] where a < m <  b given:
```
L_i (x) =   f(a) * ((x-m) * (x-b)) / ((a-m) * (a-b))
         + f(m) * ((x-a) * (x-b)) / ((m-a) * (m-b))
         + f(b) * ((x-a) * (x-m)) / ((b-m)*(b-a))
```
i=1 is not valid. Additionally the endpoints in the three point form should produce a whole weight, to the best of my understanding. For example at i=0,1
```
// at L_1
integral += f[0] * ((x[0] - x[1]) * (x[0] - x[2]))
	         / ((x[0] - x[1]) * (x[0] - x[2]))
```
and
```
// at L_2
integral += f[1] * (x[2] - x[0]) * (x[2] - x[1])
	         / (x[2] - x[0]) * (x[2] - x[1])
```

I feel the form f[0]/6 + 4*f[1]/6 + f[2]  does account for the midpoint when evaluating the integral of a three point approximation with Simpson's method.

integrate/simpsons_test.go
nisfornature: Thanks.

The issue I'm struggling with is supplying a passable test that jointly illustrates these points and also serves as a 'true' representation of the algorithm. It's my understanding that all common methods used to perform numerical integrations make the assumption that points are evenly distributed.

My take is that users should be responsible for passing values that are suitable for the mathematical model and form (e.g Newton-Cotes) over implementing additional computational arithmetic.

That being said, I'm not sure if segmentation over irregularly spaced data (http://cs-people.bu.edu/evimaria/cs565/TT06.pdf) where one could segment intervals and then evaluate the integral, is a possible solution? Evaluating x in any condition is costly in and of itself. 

```
// Apply Simpson's to the k-segment produced from ...
{
			x:    append(floats.Span(make([]float64, N), 0, 1),
                                            floats.Span(make([]float64, N), 100, 101)...),
			f:    func(x float64) float64 { return x },
			want: 50.5,
		},
```

integrate/simpsons_test.go
nisfornature: I don't think so. I think that's a suitable range.

integrate/simpsons.go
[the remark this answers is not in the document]
nisfornature: Performing explicit multiplication instead with the supplied benchmark resulting in the following improvement:

```
type Segment struct {
	x []float64
	y []float64
}

var benchmarkData = []Segment{
	{
		x: floats.Span(make([]float64, ^uint32(0)>>7), math.MaxFloat64-1, math.MaxFloat64),
		y: floats.Span(make([]float64, ^uint32(0)>>7), math.MaxFloat64-1, math.MaxFloat64),
	},
}

func BenchmarkSimpsons(b *testing.B) {
	for i := 0; i < len(benchmarkData); i++ {
		Simpsons(benchmarkData[i].x, benchmarkData[i].y)
	}
}

func BenchmarkSimpsonsMathPow(b *testing.B) {

	for i := 0; i < len(benchmarkData); i++ {
		SimpsonsMathPow(benchmarkData[i].x, benchmarkData[i].y)
	}
}
```

> goos: linux
> goarch: amd64
> pkg: 
>  gonum.org/v1/gonum/integrate
> 
> BenchmarkSimpsons-6  000000000 .462 ns/op 0 B/op 0 allocs/op
> BenchmarkSimpsonsMathPow-6  1 3775877519 ns/op 1616782624 B/op 57 allocs/op
> PASS
> ok  	gonum.org/v1/gonum/integrate	15.396s

integrate/simpsons.go
```diff
@@ -0,0 +1,124 @@
...
+
+	for i := 1; i < len(x)-1; i += 2 {
+		diff = append(diff, x[i]-x[i-1])
+		diff = append(diff, x[i+1]-x[i])
+
+		h0 := diff[i-1]
+		h1 := diff[i]
+		hph := h1 + h0
+
+		f0 := f[i-1]
```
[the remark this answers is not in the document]
nisfornature: Alrighty

integrate/simpsons.go
nisfornature: Sorry about that:

I've updated line 17 to:
`\\ \sum_{i=0}^{(N/2)-1} {a_0}_i * f_{2i+1} + {a_1}_i * f_{2i} + {a_2}_i * f_{2i}`

Lines 91-92 are now:
```
f0 := f[i+1]
f1 := f[i]
f2 := f[i-1]
```
https://en.wikipedia.org/wiki/Simpson%27s_rule#Composite_Simpson's_rule_for_irregularly_spaced_data

integrate/simpsons.go
nisfornature: Removed the slice

integrate/simpsons.go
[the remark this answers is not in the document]
kortschak: This output looks wrong to me; I believe the improvement, but the  `000000000` and `.462` are not normal benchmark output.

For future reference, you should keep the name the same, run about 10 rounds of benchmarks and then use [golang.org/x/perf/cmd/benchstat](https://golang.org/x/perf/cmd/benchstat) to perform the comparisons.

integrate/simpsons.go
```diff
@@ -0,0 +1,133 @@
...
+	"sort"
+)
+
+// Simpson's method approximates the integral of a function f(x)
+// by means of subdividing the interval of integration into segments
+// and applying a method that fits a polynomial to each subinterval.
+// this implementation makes use of the following composite Simpson's method
+// to estimate \int_a^b f(x) dx:
+//
+// \sum_{i=0}^{(N/2)-1} {a_0}_i * f_{2i+1} + {a_1}_i * f_{2i+1} + {a_2}_i * f_{2i}
```
kortschak: Indent these expressions by an extra space (U+20).

integrate/simpsons.go
```diff
@@ -0,0 +1,133 @@
...
+//
+// where N is the number of subintervals and the number of subintervals
+// is even and {a_0}_i, {a_1}_i, and {a_2}_i are constants at index i given by:
+//
+// {a_0}_i = 2 * h^{3}_{2i+1} - h^{3}_{2i} + 3 * h_{2i} * h^{2}_{2i+1} /
```
kortschak: Indent.

integrate/simpsons.go
```diff
@@ -0,0 +1,133 @@
...
+//
+// The formula above approximates the integral of function of f
+// if N is an even number of subintervals.
+//
+// If the number of subintervals are odd,
+// the subintervals up i=0..n-2 are given by the above
+// and the approximations over the second to last and last subintervals
+// are given by:
+//
+// {a_0}_i * f_{N} + {a_1}_i * f_{N-1} + {a_2}_i * f_{N-2}
```
kortschak: Indent.

integrate/simpsons.go
```diff
@@ -0,0 +1,133 @@
...
+// {a_1}_i = h^{2}_{N-1} + 3 * h^{3}_{N-1} * h_{N-2} /
+//           6 * h_{N-2}
+//
+// {a_2}_i = -1 * h^{3}_{N-1} /
+//            6 * h_{N-2} * (h_{N-2} + h_{N-1})
+// More information is available at:
+// https://en.wikipedia.org/wiki/Simpson%27s_rule#Composite_Simpson's_rule_for_irregularly_spaced_data
+//
+// The (x,f) input data points must be sorted along x.
+// One can use github.com/gonum/stat.SortWeighted to do that.
```
kortschak: s|github.com/gonum/stat.SortWeighted|stat.SortWeighted|

Also there is no need for the line breaks between these three sentences.

I will file an issue to fix this in `Trapezoidal`.

integrate/simpsons.go
```diff
@@ -0,0 +1,133 @@
...
+// The x and f slices must be of equal length and have length > 1.
+func Simpsons(x, f []float64) float64 {
+	validateInterval(x, f)
+
+	var integral float64
+
+	// If only three points are supplied,
```
kortschak: Please justify this text at about 60 columns.

```
	// If only three points are supplied, approximate the integral
	// assuming the second coordinate points suffices as a midpoint
	// and interpolate with Simpson's method.
```

integrate/simpsons.go
```diff
@@ -0,0 +1,133 @@
...
+			panic("integrate: at least three unique points are required")
+		}
+		return (x[2] - x[0]) * (f[0]/6.0 + 4.0*f[1]/6.0 + f[2]/6.0)
+	}
+
+	for i := 1; i < len(x)-1; i += 2 {
+
+		// h0 is the subinterval that is at the position two left of the point x[i]
+		// h1 is the subinterval whose right endpoint is x[i]
+		// hph is the addition of the two subintervals which represents
```
kortschak: This comment is incomplete and probably redundant. If you believe the sentence when complete is helpful, please move it down to the line above the `hph` decl and add a blank line above it.

integrate/simpsons.go
```diff
@@ -0,0 +1,133 @@
...
+		if x[0] == x[1] && f[0] == f[1] || x[1] == x[2] && f[0] == f[1] {
+			panic("integrate: at least three unique points are required")
+		}
+		return (x[2] - x[0]) * (f[0]/6.0 + 4.0*f[1]/6.0 + f[2]/6.0)
+	}
+
+	for i := 1; i < len(x)-1; i += 2 {
+
+		// h0 is the subinterval that is at the position two left of the point x[i]
+		// h1 is the subinterval whose right endpoint is x[i]
```
kortschak: Can you move this down to be the line above where `h1` is declared and have a blank line above it. Also finish the sentence with a full stop.

integrate/simpsons.go
```diff
@@ -0,0 +1,133 @@
...
+		// Panic if either subinterval coordinate points match
+		if x[0] == x[1] && f[0] == f[1] || x[1] == x[2] && f[0] == f[1] {
+			panic("integrate: at least three unique points are required")
+		}
+		return (x[2] - x[0]) * (f[0]/6.0 + 4.0*f[1]/6.0 + f[2]/6.0)
+	}
+
+	for i := 1; i < len(x)-1; i += 2 {
+
+		// h0 is the subinterval that is at the position two left of the point x[i]
```
kortschak: Final full stop please.

integrate/simpsons.go
```diff
@@ -0,0 +1,133 @@
...
+
+		// Panic if either subinterval coordinate points match
+		if x[0] == x[1] && f[0] == f[1] || x[1] == x[2] && f[0] == f[1] {
+			panic("integrate: at least three unique points are required")
+		}
+		return (x[2] - x[0]) * (f[0]/6.0 + 4.0*f[1]/6.0 + f[2]/6.0)
+	}
+
+	for i := 1; i < len(x)-1; i += 2 {
+
```
kortschak: Please delete this blank line.

integrate/simpsons.go
```diff
@@ -0,0 +1,133 @@
...
+	var integral float64
+
+	// If only three points are supplied,
+	// approximate the integral
+	// assuming the second coordinate points
+	// suffices as a midpoint and interpolate with Simpson's method.
+	if len(x) == 3 {
+
+		// Panic if either subinterval coordinate points match
```
kortschak: Final full stop.

integrate/simpsons.go
```diff
@@ -0,0 +1,133 @@
...
+	validateInterval(x, f)
+
+	var integral float64
+
+	// If only three points are supplied,
+	// approximate the integral
+	// assuming the second coordinate points
+	// suffices as a midpoint and interpolate with Simpson's method.
+	if len(x) == 3 {
+
```
kortschak: Please delete this blank line.

integrate/simpsons.go
```diff
@@ -0,0 +1,133 @@
...
+		// h0 is the subinterval that is at the position two left of the point x[i]
+		// h1 is the subinterval whose right endpoint is x[i]
+		// hph is the addition of the two subintervals which represents
+		h0 := x[i] - x[i-1]
+		h0p2 := h0 * h0
+		h0p3 := h0 * h0 * h0
+		h1 := x[i+1] - x[i]
+		h1p2 := h1 * h1
+		h1p3 := h1 * h1 * h1
+		hph := h1 + h0
```
kortschak: `hph := h0 + h1`

integrate/simpsons_test.go
```diff
@@ -0,0 +1,130 @@
...
+
+import (
+	"math"
+	"testing"
+
+	"gonum.org/v1/gonum/floats"
+)
+
+func TestSimpsons(t *testing.T) {
+	const N = 1e6
```
kortschak: Please make this label lower case.

integrate/simpsons.go
vladimir-ch: It ignores `x[1]`. It assumes that the interpolating quadratic passes through `f[1]` located at `0.5*(x[0]+x[2])`.

integrate/simpsons_test.go
vladimir-ch: This code allows non-uniform spacing, so the test should generate such input. You can either randomly jitter values from `floats.Span` or generate `x` algorithmically.

integrate/simpsons_test.go
vladimir-ch: Why? Using less points makes it harder to pass the test and so makes it more reliable. Also with less points the test finishes faster.

integrate/simpsons.go
```diff
@@ -0,0 +1,133 @@
...
+	var integral float64
+
+	// If only three points are supplied,
+	// approximate the integral
+	// assuming the second coordinate points
+	// suffices as a midpoint and interpolate with Simpson's method.
+	if len(x) == 3 {
+
+		// Panic if either subinterval coordinate points match
+		if x[0] == x[1] && f[0] == f[1] || x[1] == x[2] && f[0] == f[1] {
```
vladimir-ch: What does this check? Why this special case is more stringent than the general case? This does not compensate for the fact that the formula below ignores `x[1]`. I would remove this.

integrate/simpsons.go
```diff
@@ -0,0 +1,133 @@
...
+		// hph is the addition of the two subintervals which represents
+		h0 := x[i] - x[i-1]
+		h0p2 := h0 * h0
+		h0p3 := h0 * h0 * h0
+		h1 := x[i+1] - x[i]
+		h1p2 := h1 * h1
+		h1p3 := h1 * h1 * h1
+		hph := h1 + h0
+
+		f0 := f[i+1]
```
vladimir-ch: It's confusing that the suffix for `h` increases with `i` but for `f` and `a` it's the opposite.

integrate/simpsons.go
```diff
@@ -0,0 +1,133 @@
...
+		a0 := (2*h1p3 - h0p3 + 3*h0*h1p2) / (6 * h1 * hph)
+		a1 := (h1p3 + h0p3 + 3*h1*h0*hph) / (6 * h1 * h0)
+		a2 := (2*h0p3 - h1p3 + 3*h1*h0p2) / (6 * h0 * hph)
+
+		integral += a0 * f0
+		integral += a1 * f1
+		integral += a2 * f2
+	}
+
+	if (len(x)-1)%2 != 0 {
```
vladimir-ch: `len(x)%2 == 0`

integrate/simpsons.go
```diff
@@ -0,0 +1,133 @@
...
+		h0 := x[len(x)-2] - x[len(x)-3]
```
vladimir-ch: I would do `n := len(x)` where `len(x)` is first used.

integrate/simpsons_test.go
```diff
@@ -0,0 +1,130 @@
...
+	for i, test := range []struct {
+		x    []float64
+		f    func(x float64) float64
+		want float64
+	}{
+		{
+			x: []float64{0.0, 0.1, 1.0},
+			f: func(x float64) float64 {
+				switch {
+				case x >= 1e-31 || x <= 1.0:
```
vladimir-ch: What is this test case testing? I think you wanted to put here `&&` instead of `||`. When you fix it, the test will fail.

integrate/simpsons_test.go
```diff
@@ -0,0 +1,130 @@
...
+			y[i] = test.f(v)
+		}
+		v := Simpsons(test.x, y)
+		if !floats.EqualWithinAbs(v, test.want, 1e-12) {
+			t.Errorf("test #%d: got=%v want=%f\n", i, v, float64(test.want))
+		}
+	}
+}
+
+func TestSimpsonsPanics(t *testing.T) {
```
vladimir-ch: I don't think we need to test this. We don't do such tests anywhere else in Gonum.

integrate/simpsons_test.go
kortschak: Also on this, please use the convention that intervals are written in number line order, this will be (with @vladimir-ch's correction) as `1e-31 <= x && x <= 1`. This eases identification of this kind of error.

kortschak:
For the sake of our build load, please try to push a complete set of changes rather than one small change at a time. Each travis push build takes just over 2 hours of cpu time and potentially up to about 40 minutes wall time, during which time other builds will have to wait.

integrate/simpsons_test.go
nisfornature: - I added a few tests with small number ranges

I tried adding a tolerance value, but it looks like floats.EqualWithinAbs behaves differently when I pass the tolerance through a struct:
```
floats.EqualWithinAbs(v, test.want, 1e-12)
```
```
// test.tol := 1e-6
floats.EqualWithinAbs(v, test.want, test.tol)
```

integrate/simpsons.go
nisfornature: Added indentation

integrate/simpsons_test.go
nisfornature: I think tests for panics are really important. If I mix-up upper and lower bounds for example, I feel like those sorts of things should be introduced.

integrate/simpsons.go
nisfornature: The existing checks guarantee ordering and that three points are supplied, but if I pass the points (0,1), (0,1), (1,1) or (0,0), (1,1), (1,1) there's no middle point.

vladimir-ch: Why is (0,1)(0,2,)(1,1) allowed? Why is this check for repeated locations done only in the 3-length case? Does it mean that `validateInterval` does not actually validate the input?

integrate/simpsons_test.go
vladimir-ch: And in case my message was not clear, the test will fail as it should because this is a bad test case.

integrate/simpsons_test.go
vladimir-ch: I think the cost of this outweighs any potential benefits. It clutters the CI output and is more code to maintain and keep in sync with the implementation for practically no benefit. @kortschak may have more argument for why we don't need this.

integrate/simpsons_test.go
```diff
@@ -0,0 +1,155 @@
...
+			y:    []float64{0, 1, 2, 3},
+			want: "integrate: slice length mismatch",
+		},
+		{
+			x:    []float64{0},
+			y:    []float64{1e2},
+			want: "integrate: input data too small",
+		},
+		{
+			x:    []float64{1.0, 0.9, 0.8, 0.7, 0.6, 0.5, 0.4, 0.2, 0.3, 0.1},
```
vladimir-ch: Come to think of it, why should such inputs not be allowed? Why shouldn't they return -1 times the result with the reversed input?

integrate/simpsons_test.go
vladimir-ch: Must be a bug in your code.

integrate/simpsons_test.go
kortschak: The only time we check for panics is when we are testing non-obvious logic. This happens in mat, graph and lapack. We don't check for this kind of thing in stat in other place nor in floats.

In general adding tests is a way to write in a potentially clearer way an assertion about behaviour. The `validateInterval` helper is already simple enough (possibly simpler if the switch were placed in the function proper) that it is obviously correct.

integrate/simpsons.go
btracey: I think we should just disallow repeated points altogether. It's easy to check as you loop through the slice. The algorithm doesn't really make sense with repeated x locations.

integrate/simpsons_test.go
btracey: Because it's adding to the list of complications of the function. It's clear to say we test if the data is sorted. We do this in other places at the moment without similar checks for reverse sorted.

integrate/simpsons.go
vladimir-ch: > we should just disallow repeated points altogether

That's exactly where I'm trying to get with the questions. But then `sort.Float64sAreSorted` cannot be used which means adding a helper here and this PR is already taking too long. That's why I asked for removing the check here.

integrate/simpsons_test.go
vladimir-ch: Yea, it's probably complicating things too much.

integrate/simpsons.go
nisfornature: I'm iterating over the sub-intervals not the number of points supplied.

integrate/simpsons.go
nisfornature: Square waves and Haar wavelets and noisy data sets may produce conditions where x maps to multiple values?

integrate/simpsons_test.go
nisfornature: I'm using a piecewise function.

integrate/simpsons.go
nisfornature: Thanks

integrate/simpsons.go
kortschak: @vladimir-ch's point is that these expressions are equivalent, but the proposed expression is simpler.

nisfornature:
Ah sorry for spamming updates on this PR, just trying to tidy up these two files with the editor I'm using and submit something relatively valid without complicating things!

integrate/simpsons_test.go
```diff
@@ -0,0 +1,217 @@
...
+		defer func() {
+			if r := recover(); r == nil {
+				t.Errorf("test #%d did not panic: expectedPanic=\"%s\"", i, test.want)
+			}
+		}()
+		Simpsons(test.x, test.y)
+	}
+}
+
+func appendMany(elems ...[]float64) []float64 {
```
kortschak: Please rewrite this as:

```
func cat(slices ...[]float64) []float64 {
	var c []float64
	for _, s := range slices {
		c = append(c, s...)
	}
	return c
}
```

integrate/simpsons_test.go
nisfornature: I had been looking at the stats package. I was hoping to approximate functions having more points at an endpoint or middle given a piecewise functions that plots points along a polynomial and a distribution chosen at random. I'd like to leave this bit to someone else if possible.

I still need to change the interval.

integrate/simpsons.go
nisfornature: Without knowing the rate of change at all values of x and only having 3 points to work with, I think using the Simpson's method is most appropriate for the Simpson's formula.

`(x[2] - x[0]) * (f[0]/6.0 + 4.0*f[1]/6.0 + f[2]/6.0)`

We are only given three points and no derivative values. I think we run into the same issues with methods that don't assume equidistantly spaced points (e.g divided difference) due to the fact that evaluating the derived polynomial at the same points of x will cancel out like terms and evaluating the derived polynomial at an offset (e.g. 1e-4) introduces additional error.

```
var dividedDifferenceInterpolation = func(p float64) float64 {
            return f[0] +
                   (f[1]-f[0])/(x[1]-x[0]) +
                   (p-x[0])*(p-x[1])*((f[2]-f[1])/(x[2]-x[1])-(f[1]-f[0])/(x[1]-x[0]))/(x[2]-x[1])
        }

return (x[1]-x[0])*(dividedDifferenceInterpolation(x[1])-dividedDifferenceInterpolation(x[2]))*0.5 +
       (x[2]-x[1])*(dividedDifferenceInterpolation(x[2])-dividedDifferenceInterpolation(x[1]))*0.5
```

integrate/simpsons_test.go
vladimir-ch: It's not a piecewise constant function unless you put `&&` instead of `||` here.

Apart from that, it's still a bad test case. You are supplying a discontinuous function to a method that assumes smooth input. And you are sampling this function at only three locations, two fall into the 1 region, one falls into the 0 region. `Simpsons` integrates the quadratic function that interpolates these three values and returns the result which has nothing to the with the original piecewise constant function.

integrate/simpsons.go
vladimir-ch: If you have multiple values for one x, then such a mapping is not a function. If your data points come from some noisy measurements and may contain repeated locations, then you must either clean the input first or approximate the data with a smooth function. All of this is out of scope for this package.

integrate/simpsons.go
vladimir-ch: The formula
```
(x[2] - x[0]) * (f[0]/6.0 + 4.0*f[1]/6.0 + f[2]/6.0)
```
ignores the value of `x[1]`. But why are you special-casing this at all? Why the general case cannot handle it?

vladimir-ch:
The discussion has reached 100 comments and it's become difficult to navigate. I suggest that you close this PR, clean up the code and address _all_ our comments and suggestions, and then resubmit it in a new PR.

vladimir-ch:
This PR has reached 100 comments and the progress is rather slow. My free time is limited and unfortunately I cannot dedicate more of it to reviewing this. Maybe others can help you get it to a mergeable state, otherwise

vladimir-ch:
Sorry, the close and the last comment were not intentional.

nisfornature:
Closing PR.

PR #1102, opened 2019-09-27
stat/*: Update functions to take empty matrices

btracey:
Please take a look.

### DIFF
--- a/stat/cca_example_test.go
+++ b/stat/cca_example_test.go
@@ -85,16 +85,17 @@ func ExampleCC() {
 	}
 
 	// Unpack cc.
+	var pVecs, qVecs, phiVs, psiVs mat.Dense
 	ccors := cc.CorrsTo(nil)
-	pVecs := cc.LeftTo(nil, true)
-	qVecs := cc.RightTo(nil, true)
-	phiVs := cc.LeftTo(nil, false)
-	psiVs := cc.RightTo(nil, false)
+	cc.LeftTo(&pVecs, true)
+	cc.RightTo(&qVecs, true)
+	cc.LeftTo(&phiVs, false)
+	cc.RightTo(&psiVs, false)
 
 	// Canonical Correlation Matrix, or the correlations between the sphered
 	// data.
 	var corSph mat.Dense
-	corSph.CloneFrom(pVecs)
+	corSph.CloneFrom(&pVecs)
 	col := make([]float64, xd)
 	for j := 0; j < yd; j++ {
 		mat.Col(col, j, &corSph)
@@ -110,13 +111,13 @@ func ExampleCC() {
 	fmt.Printf("\n\nccors = %.4f", ccors)
 
 	// Left and right eigenvectors of the canonical correlation matrix.
-	fmt.Printf("\n\npVecs = %.4f", mat.Formatted(pVecs, mat.Prefix("        ")))
-	fmt.Printf("\n\nqVecs = %.4f", mat.Formatted(qVecs, mat.Prefix("        ")))
+	fmt.Printf("\n\npVecs = %.4f", mat.Formatted(&pVecs, mat.Prefix("        ")))
+	fmt.Printf("\n\nqVecs = %.4f", mat.Formatted(&qVecs, mat.Prefix("        ")))
 
 	// Canonical Correlation Transforms. These can be useful as they represent
 	// the canonical variables as linear combinations of the original variables.
-	fmt.Printf("\n\nphiVs = %.4f", mat.Formatted(phiVs, mat.Prefix("        ")))
-	fmt.Printf("\n\npsiVs = %.4f", mat.Formatted(psiVs, mat.Prefix("        ")))
+	fmt.Printf("\n\nphiVs = %.4f", mat.Formatted(&phiVs, mat.Prefix("        ")))
+	fmt.Printf("\n\npsiVs = %.4f", mat.Formatted(&psiVs, mat.Prefix("        ")))
 
 	// Output:
 	// corRaw = ⎡-0.2192   0.3527   0.5828  -0.3883⎤
--- a/stat/mds/mds.go
+++ b/stat/mds/mds.go
@@ -73,8 +73,7 @@ func TorgersonScaling(dst *mat.Dense, eigdst []float64, dis mat.Symmetric) (k in
 
 	var tmp mat.Dense
 	tmp.Mul(dst, mat.NewDiagonalRect(n, k, vals[:k]))
-	tmp2 := dst.Slice(0, n, 0, k).(*mat.Dense)
-	*dst = *tmp2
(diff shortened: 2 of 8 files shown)


btracey:
I believe this fixes the issue. It would be nice if someone could double check I didn't miss anything. I did not look at the `graph` packages, so there may be some holdouts there.

kortschak:
I can confirm that there is nothing in graph that needs this update; given that we have an assumption of sparsity for values there we don't use matrices as destinations.

stat/mds/mds.go
```diff
@@ -14,21 +14,19 @@ import (
...
+// If the scaling is not successful, dst will be empty upon return.
+// When the scaling is successful, dst will be resized to k columns wide.
 // Eigenvalues will be copied into eigdst and returned as eig if it is provided.
 //
-// If dst is nil, a new mat.Dense is allocated. If dst is not empty,
-// the dimensions of dst and dis must match otherwise TorgersonScaling will panic.
-// The dis matrix must be square or TorgersonScaling will panic.
-func TorgersonScaling(dst *mat.Dense, eigdst []float64, dis mat.Symmetric) (k int, mds *mat.Dense, eig []float64) {
+// TorgersonScaling will panic if dst is not empty.
+func TorgersonScaling(dst *mat.Dense, eigdst []float64, dis mat.Symmetric) (k int, eig []float64) {
```
kortschak: Rename this to `TorgersonScalingTo`?

stat/cca_example_test.go
```diff
@@ -85,16 +85,17 @@ func ExampleCC() {
 	}
 
 	// Unpack cc.
+	var pVecs, qVecs, phiVs, psiVs mat.Dense
 	ccors := cc.CorrsTo(nil)
```
kortschak: I vaguely recall that we discussed the slice case and decided that the utility of the `range fn(nil)` case was worth keeping. Do I remember correctly?

btracey: I do not remember that at all (though it could have happened). We talked about it in `floats`, but that's a bit different because `floats` does not currently allow `nil` inputs. I think it's good to keep the nil case for slices, since we do that in a lot of places, and it's not that common to use pointers to slices. I think the default should be "you can either supply the exact size, or nil". Matrices end up (should be) the special case, because we have to use receivers, and we want those to be able to be Empty. Supplying empty matrices as destinations is just a follow-on case to that

stat/mds/mds.go
btracey: If you want ... I personally don't think it's better. In `mat`, things are confusing because input matrices can be sources or destinations. There it's useful to have `To` in the signature, for example to distinguish between `SetFromU` and `UTo`. Or perhaps more importantly, many of the operations modify the receiver, so it's important to distinguish those that don't.

On the flip side, if you look at somewhere like `distmv`, I feel like if you see the signature of `Normal.CovarianceMatrix(dst *mat.Dense)`, I don't think there's any confusion that the behavior stores the covariance matrix in `dst`. The `To` then just adds clutter to the names.

stat/cca_example_test.go
btracey: As an addendum, this is going to be used as

```
dst := &mat.Dense{} or dst.Reset()
a,b := mds.TorgersonScaling(dst, stuff)
```
and so either way it seems pretty obvious to me that `dst` can't be supplying any useful information to the function.

stat/mds/mds.go
kortschak: Fair enough.

stat/cca_example_test.go
kortschak: I think this second comment belonged above.

btracey: yes it did, sorry.

stat/mds/mds.go
```diff
@@ -75,10 +73,11 @@ func TorgersonScaling(dst *mat.Dense, eigdst []float64, dis mat.Symmetric) (k in
 
 	var tmp mat.Dense
 	tmp.Mul(dst, mat.NewDiagonalRect(n, k, vals[:k]))
-	dst = dst.Slice(0, n, 0, k).(*mat.Dense)
+	tmp2 := dst.Slice(0, n, 0, k).(*mat.Dense)
```
kortschak: Does `*dst = *dst.Slice(0, n, 0, k).(*mat.Dense)` not work?

stat/mds/mds.go
```diff
@@ -14,21 +14,19 @@ import (
 // TorgersonScaling converts a dissimilarity matrix to a matrix containing
 // Euclidean coordinates. TorgersonScaling places the coordinates in dst and
 // returns it and the number of positive Eigenvalues if successful.
-// If the scaling is not successful, dst is returned, but will not be a valid scaling.
-// When the scaling is successful, mds will be resized to k columns wide.
+// If the scaling is not successful, dst will be empty upon return.
+// When the scaling is successful, dst will be resized to k columns wide.
 // Eigenvalues will be copied into eigdst and returned as eig if it is provided.
 //
-// If dst is nil, a new mat.Dense is allocated. If dst is not empty,
```
kortschak: Is there a reason to make the constraints tighter here?

btracey: Sorry, I put it in the commit message, which I now notice doesn't show. Reproducing here:

Change TorgersonScaling to require an empty matrix. Users who want to reuse data can call Reset now that it is exposed. This function is different than others because the return size is unknown. Forcing the input matrix to be empty makes it clear that the dst matrix will be dynamically resized

kortschak: That seems reasonable. It's pretty much what I thought you were thinking.

stat/mds/mds_test.go
```diff
@@ -5,6 +5,7 @@
 package mds
 
 import (
+	"fmt"
```
kortschak: Please delete this.

stat/mds/mds_test.go
```diff
@@ -112,17 +113,20 @@ var torgersonScalingTests = []struct {
...
 		_, c := test.dis.Dims()
-		gotK, got, gotVals := TorgersonScaling(nil, make([]float64, c), test.dis)
+		var got mat.Dense
+		gotK, gotVals := TorgersonScaling(&got, make([]float64, c), test.dis)
 		if gotK == 0 {
 			t.Error("unexpected scaling failure")
 			continue
 		}
+		r, c := got.Dims()
+		fmt.Println("got dims", r, c)
```
kortschak: Delete.

stat/mds/mds.go
kortschak: It does. Can you please make that change too.

kortschak:
*ping*

btracey:
Sorry. I guess I dreamed I pushed this or something.

kortschak:
No worries.

Issue #1089, opened 2019-09-17
proposal: tag a 0.5 release


kortschak:
### Background
We don't have a release. This impacts on distro packagers, so we should have one.

### Proposal
Make a non-stable release. I propose 0.5 since we are probably half way there (in the Zeno sense).

### Potential impact of proposal

I will do this.

btracey:
This is a good idea. I think we should fix #1081 , and then I think we're in a good place for a point release. Thanks for pushing this forward.

sbinet:
I am +1 on this.
`0.5` is as good a number as any.
I'd go with `v0.5.0` for the tag, though.

vladimir-ch:
SGTM

btracey:
Maybe we could make it `0.6` in honor of 6 years since creation?

kortschak:
`vV.Y.M` for less than `v1`?

sbinet:
having a 3-digit version number plays well with the whole semver/go.mod ecosystem.
but I think it would be less confusing to be less cute and just go with `v0.X.0` for versions less than `v1`.

(`v0.6.0` is fine by me.)

kortschak:
I'm looking at how long we are going to be pre-v1. Having `vV.Y.M` for pre-v1 makes it easy to choose reasonably meaningful version numbers.

laxatives:
Please pin this repo. Its already useful but its hard to justify putting this into build pipelines if it is liable to break builds randomly in the future.

kortschak:
If you use go modules this will not break you, but yes, we are working towards this.

kortschak:
Tag v0.6.0 has been pushed.

PR #1106, opened 2019-10-02
gonum: add release schedule

sbinet:
Please take a look.

Fixes gonum/gonum#1089.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,17 +14,14 @@ Gonum supports and tests on the three most recent minor versions of Go on [Linux
 
 ## Release schedule
 
-The Gonum modules are released on a quaterly-based time schedule, aligned with the Go releases.
+The Gonum modules are released on a six-month release schedule, aligned with the Go releases.
 _i.e.:_ when `Go-1.x` is released, `Gonum-v0.n.0` is released around the same time.
-Three months later, `Gonum-v0.n+1.0` is released.
-Three months after, `Go-1.x+1` is released, and `Gonum-v0.n+2.0` as well.
+Six months after, `Go-1.x+1` is released, and `Gonum-v0.n+1.0` as well.
 
 The release schedule, based on the current Go release schedule is thus:
 
 - `Gonum-v0.n.0`: February
-- `Gonum-v0.n+1.0`: May
-- `Gonum-v0.n+2.0`: August
-- `Gonum-v0.n+3.0`: November
+- `Gonum-v0.n+1.0`: August
 
 ## Build tags
 
--- a/README.md
+++ b/README.md
@@ -12,6 +12,19 @@ go get -u gonum.org/v1/gonum/...
 
 Gonum supports and tests on the three most recent minor versions of Go on [Linux](https://github.com/gonum/gonum/blob/master/.travis.yml#L6-L11) and [Windows](https://github.com/gonum/gonum/blob/master/appveyor.yml#L13-L18).
 
+## Release schedule
+
+The Gonum modules are released on a quaterly-based time schedule, aligned with the Go releases.
+_i.e.:_ when `Go-1.x` is released, `Gonum-v0.n.0` is released around the same time.
+Three months later, `Gonum-v0.n+1.0` is released.
+Three months after, `Go-1.x+1` is released, and `Gonum-v0.n+2.0` as well.
+
+The release schedule, based on the current Go release schedule is thus:
+
+- `Gonum-v0.n.0`: February
+- `Gonum-v0.n+1.0`: May
+- `Gonum-v0.n+2.0`: August
+- `Gonum-v0.n+3.0`: November
 
 ## Build tags
 
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 ## Installation
 
-The core packages of the gonum suite are written in pure Go with some assembly.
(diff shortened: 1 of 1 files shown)


kortschak:
Is there a reason for the quarterly release?

If we are going to do this, I'd like to drop the oldest supported Go version and only cover 2 releases.

sbinet:
> Is there a reason for the quarterly release?

no earth-shattering reason, no.
I think the major driver here is to be aligned w/ Go releases.
I'll amend and update the PR for a 2-releases per year release schedule.

> I'd like to drop the oldest supported Go version and only cover 2 releases.

according to the [Go release policy](https://golang.org/doc/devel/release.html#policy), the version support is N-2 *excluded* (I must admit I thought it was N-2 included.)
so I guess we could also piggy-back on that policy.
(I suspect this would mostly only impact (ie: reduce) our travis-ci build times)

one thing I didn't mention in the PR is what we want to do with `go.mod` dependencies.
I think we should `go get -u foo@latest` for each new Gonum release (at least the `golang.org/x/...` ones) to cut on the accumulated cruft.
thoughts?

kortschak:
The other reason for doing N and N-1 is that it allows us to take up recent additions to the standard library and language quicker.

I'm not sure of the need to update deps unless there are bugs identified in them; MVS will already pick the more recent version in client code and we don't actually have that many deps (by design).

sbinet:
IMHO, updating deps is the same mechanism than doing N and N-1: it makes sure we advance the global ecosystem together towards N,N-1 versions. (and prevents or limits version skew)

for example, in Go-HEP, I get multiple versions of `golang.org/x/...` repos from a bunch of deps (some through gonum and gonum/plot).

kortschak:
>for example, in Go-HEP, I get multiple versions of `golang.org/x/...` repos from a bunch of deps (some through gonum and gonum/plot).

What do you mean by this? The compiled code should have only one version of any given dependency.

sbinet:
no, you can have many versions of a given module in a given binary.
I myself may be using mod-v0.2.0 while some of my dependencies may be using mod-v0.1.1.
when one looks at the resulting `go.sum` file, one should see all of those.

_e.g.:_ with current `gonum/gonum`:
```
$> cat gonum.org/v1/gonum/go.sum
[...]
golang.org/x/exp v0.0.0-20180321215751-8460e604b9de
golang.org/x/exp v0.0.0-20180807140117-3d87b88a115f
golang.org/x/exp v0.0.0-20190125153040-c74c464bbbf2
golang.org/x/image v0.0.0-20180708004352-c73c2afc3b81
golang.org/x/tools v0.0.0-20180525024113-a5b4c53f6e8b
golang.org/x/tools v0.0.0-20190206041539-40960b6deb8e
[...]
```

3 different versions of `x/exp`, 2 different versions of `x/tools`.

kortschak:
That's not what MVS is doing. There are sums for every possible dep, but only one of them will be included in the actual binary.

sbinet:
~~I beg to differ.~~
ok. re-reading (vgo-mvs)[https://research.swtch.com/vgo-mvs], I may have generalized the rule for `M.N` (in `v.M.N.P`) to `N=0` and `P`.

ok. let's not add this clause about upgrading module deps, and stick with "explicitly updating them when a bug is encountered (or a new feature needed)".

kortschak:
Yeah, the definitive test is to query an executable. If you run `goversion -m` on some executable built with modules it shows which modules and versions have gone into its build.

sbinet:
for archives' sake, `goversion` is [rsc.io/goversion](https://godoc.org/rsc.io/goversion)

btracey:
I'm in favor of N and N-1. It aligns with the Go ecosystem. I guess we may have more users stuck on older hardware, but at least if we have version tags it'll be easier to stick with older versions of us.

6 month releases SGTM as an experiment.

README.md
```diff
@@ -38,4 +48,4 @@ https://groups.google.com/forum/#!forum/gonum-dev
 
 Original code is licensed under the Gonum License found in the LICENSE file. Portions of the code are subject to the additional licenses found in THIRD_PARTY_LICENSES. All third party code is licensed either under a BSD or MIT license.
 
-Code in graph/formats/dot is dual licensed [Public Domain Dedication](https://creativecommons.org/publicdomain/zero/1.0/) and Gonum License, and users are free to choose the license which suits their needs for this code.
\ No newline at end of file
+Code in graph/formats/dot is dual licensed [Public Domain Dedication](https://creativecommons.org/publicdomain/zero/1.0/) and Gonum License, and users are free to choose the license which suits their needs for this code.
```
kortschak: I take it that this is just an EOL addition.

sbinet: yes.
I can remove this if you prefer (I figured the amount of noise would be rather minimal).

kortschak:
Do we have agreement on this? When we do, we can move forward and make an initial release.

DrGo:
support for Go version N and N-1 makes sense. A regular release schedule is a good motivator.

btracey:
SGTM

sbinet:
Sounds like a quorum? :)

(I am about to be off the grid until end of October. Feel free to move forward without waiting for me.)

kortschak:
@sbinet Please go ahead and merge this.

PR #1111, opened 2019-10-04
stats/combin: add CartesianGenerator

### LINKED ISSUE #308: stat/combin: Implement iterator for Cartesian

Oppodelldog:
Please take a look.

this fixes #308 according to the existing generator API.

Fixes #308.

### DIFF
--- a/stat/combin/combin.go
+++ b/stat/combin/combin.go
@@ -293,10 +293,10 @@ func Cartesian(lens []int) [][]int {
 }
 
 // Card computes the cardinality of the multi-dimensional space whose dimensions have size specified by dims
-// At least one length must be provided and all length values must be positive, otherwise this will panic.
+// All length values must be positive, otherwise this will panic.
 func Card(dims []int) int {
 	if len(dims) == 0 {
-		panic("combin: empty lengths")
+		return 0
 	}
 	card := 1
 	for _, v := range dims {
--- a/stat/combin/combin.go
+++ b/stat/combin/combin.go
@@ -312,7 +312,7 @@ func Card(dims []int) int {
 }
 
 // NewCartesianGenerator returns a CartesianGenerator for iterating over cartesian products which are generated on the fly.
-// At least one length must be provided and all length values must be positive, otherwise this will panic.
+// All values in lens must be positive, otherwise this will panic.
 func NewCartesianGenerator(lens []int) *CartesianGenerator {
 	return &CartesianGenerator{
 		lens: lens,
--- a/stat/combin/combin.go
+++ b/stat/combin/combin.go
@@ -284,6 +284,16 @@ func IndexToCombination(dst []int, idx, n, k int) []int {
 //  [ 1 2 0 ]
 // Cartesian panics if any of the provided lengths are less than 1.
 func Cartesian(lens []int) [][]int {
+	rows := NumCartesianProducts(lens)
+	out := make([][]int, rows)
+	for i := 0; i < rows; i++ {
+		out[i] = SubFor(nil, i, lens)
+	}
+	return out
+}
+
+// NumCartesianProducts calculates the number of products of a cartesian product set for the given lengths.
+func NumCartesianProducts(lens []int) int {
 	if len(lens) == 0 {
 		panic("combin: empty lengths")
 	}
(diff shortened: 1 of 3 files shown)


stat/combin/combin.go
```diff
@@ -294,11 +304,40 @@ func Cartesian(lens []int) [][]int {
...
+	lens []int
+	rows int
+	idx  int
+}
+
+// Next moves to the next product of the cartesian set.
+// It returns true if a product can be generated and false if the generator reached the cartesian sets end.
+func (g *CartesianGenerator) Next() bool {
+	g.idx++
+	if g.idx == g.rows {
```
golangcibot: S1008: should use 'return <expr>' instead of 'if <expr> { return <bool> }; return <bool>' (from `gosimple`)

stat/combin/combin.go
```diff
@@ -294,11 +304,40 @@ func Cartesian(lens []int) [][]int {
...
+
+// CartesianGenerator iterates over a cartesian product set.
+type CartesianGenerator struct {
+	lens []int
+	rows int
+	idx  int
+}
+
+// Next moves to the next product of the cartesian set.
+// It returns true if a product can be generated and false if the generator reached the cartesian sets end.
```
sbinet:
```suggestion
// It returns false when there are no cartesian product to iterate over anymore.
```

(just a suggestion from a non-native english speaker.)

stat/combin/combin.go
kortschak: The bot is asking for this to be rewritten as `return g.idx == g.rows`.

Oppodelldog: finally I applied your suggestion, thanks for that

stat/combin/combin.go
Oppodelldog: Thanks for your input.
I am also no native english speaker - I shortened the sentence a bit to focus on the false case, but kept the words I find to be important.

Oppodelldog:
> just a couple of nitpicks.
> 
> I may have one bikeshedding aspect, though: `CartesianGenerator` or `CartesianIterator` ?

Nice one.

I'd say it is both. It's not a plain iterator since it also generates values, so I would go for the "bigger" name. Also naming it generator aligns to the existing API.

Let's see if one of the other reviewers can give some input fort his one.

kortschak:
@Oppodelldog Can you use the phrasing "Closes \#\<nnn\>." or "Fixes \#\<nnn\>." so that GitHub pickes up the closure of that issue.

@btracey used the name `CartesionGenerator` in his comment, so please wait for him.

stat/combin/combin.go
```diff
@@ -284,6 +284,17 @@ func IndexToCombination(dst []int, idx, n, k int) []int {
...
+	out := make([][]int, rows)
+	for i := 0; i < rows; i++ {
+		out[i] = SubFor(nil, i, lens)
+	}
+	return out
+}
+
+// NumCartesianProducts calculates the number of products of a cartesian product set for the given lengths.
+// At least one length must be provided and all length values must be positive, otherwise this will panic.
+func NumCartesianProducts(lens []int) int {
```
btracey: I think we can find a better name for this. I'm thinking `Card`. It's just the multiplier of all the dimensions, which is the cardinality. `Card` is really best matched with `IdxFor` and `SubFor`, so it's better to have a shorter name I think. That way it's something like 
```
dims := make([]int, 10)
for i := range dims {
     dims[i] = 3
}
card := combin.Card(dims)
```

Then the documentation is just "computes the cardinality of the multi-dimensional space whose dimensions have size specified by dims"

(and then change the input parameter to be `dims` to match `IdxFor`. I think at that point we can return 0 if len(dims) = 0, and panic if any are less than 1.

Oppodelldog: Thanks for the suggestions.

The Naming has been applied in 148d721.

The second point, ```return 0 if len(dims)==0``` has been fixed in 53ff12d
Since that panic originally came from ```func Cartesian(lens []int) [][]int``` I added the panic there again to not change the behavior of that existing API. 3c15556 

I am not sure about the last one regarding keeping the existing behavior.

stat/combin/combin.go
```diff
@@ -284,23 +284,67 @@ func IndexToCombination(dst []int, idx, n, k int) []int {
...
+		if v < 1 {
+			panic("combin: length less than zero")
+		}
+		card *= v
+	}
+	return card
+}
+
+// NewCartesianGenerator returns a CartesianGenerator for iterating over cartesian products which are generated on the fly.
+// At least one length must be provided and all length values must be positive, otherwise this will panic.
```
btracey: This doesn't panic if one length isn't provided. I think for now add the same check in Product.

btracey:
LGTM after the one change.

stat/combin/combin.go
Oppodelldog: solved with 5a4529e

btracey:
Thanks for the PR!

PR #1117, opened 2019-10-11
dynamic: clarify use of inequality with float-based keys

### LINKED ISSUE #1116: graph/path/dynamic: annotate nolint on poisoned key tests

Read the D* lite algorithm.

### What did you do?

Looking over the code of the D* lite algorithm I noticed this:

https://github.com/gonum/gonum/blob/5d5638e6749a344a69611ea36989c7dcbdd2dc8d/graph/path/dynamic/dstarlite.go#L416

Unless I'm missing something, this doesn't look quite right :-)

I was trying to work out what this failure case was expected to be. Perhaps

```
	if k == badKey || other == badKey {
```

### What actually happened?

N/A

Master branch of Gonum.

Yes

kortschak:
Please take a look.

Fixes #1116.

### DIFF
--- a/graph/path/dynamic/dstarlite.go
+++ b/graph/path/dynamic/dstarlite.go
@@ -406,14 +406,18 @@ queue:
 // key is a D* Lite priority queue key.
 type key [2]float64
 
+// badKey is a poisoned key. Testing for a bad key uses NaN inequality.
 var badKey = key{math.NaN(), math.NaN()}
+
+//nolint:staticcheck
+func (k key) isBadKey() bool { return k != k }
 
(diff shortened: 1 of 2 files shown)


graph/path/dynamic/dstarlite.go
```diff
@@ -406,14 +406,18 @@ queue:
 // key is a D* Lite priority queue key.
 type key [2]float64
 
+// badKey is a poisoned key. Testing for a bad key uses NaN inequality.
 var badKey = key{math.NaN(), math.NaN()}
 
+//nolint:staticcheck
```
kortschak: This can be removed when dominikh/go-tools#623 is fixed.

PR #1119, opened 2019-10-12
mathext/amos/internal/amos: clean up linter warnings

kortschak:
Please take a look.

This also suggests that there should be consideration about the error status returns here.


btracey:
That's definitely code it would be nice to have a clean implementation of, if we knew the algorithm

PR #1129, opened 2019-10-13
Integrate: Add Romberg integration

davidkleiven:
This PR implements the Romberg integration scheme. There are two methods: `Romberg` and `RombergHL`. The latter is simply a convenience function, which can be useful for cases where users don't want to allocate a `work` array themselves.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -21,6 +21,7 @@ Clayton Northey <clayton.northey@gmail.com>
 Dan Kortschak <dan.kortschak@adelaide.edu.au> <dan@kortschak.io>
 Daniel Fireman <danielfireman@gmail.com>
 Dario Heinisch <dario.heinisch@gmail.com>
+David Kleiven <davidkleiven446@gmail.com>
 David Samborski <bloggingarrow@gmail.com>
 Davor Kapsa <davor.kapsa@gmail.com>
 DeepMind Technologies
--- a/integrate/romberg.go
+++ b/integrate/romberg.go
@@ -1,0 +1,65 @@
+// Copyright ©2019 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package integrate
+
+import (
+	"math"
+	"math/bits"
+)
+
+// Romberg returns an approximate value of the integral
+//  \int_a^b f(x)dx
+// computed using the Romberg's method. The function f is given
+// as a slice of equally-spaced samples, that is,
+//  f[i] = f(a + i*dx)
+// and dx is the spacing between the samples.
+//
+// The length of f must be 2^k + 1, where k is a positive integer,
+// and dx must be positive.
+//
+// See https://en.wikipedia.org/wiki/Romberg%27s_method for a description of
+// the algorithm
+func Romberg(f []float64, dx float64) float64 {
+	if len(f) < 3 {
+		panic("integral: invalid slice length: must be at least 3")
+	}
+
+	if dx <= 0 {
+		panic("integral: invalid spacing: must be larger than 0")
+	}
+
+	n := len(f) - 1
+	k := bits.Len(uint(n - 1))
+
+	if len(f) != 1<<uint(k)+1 {
+		panic("integral: invalid slice length: must be 2^k + 1")
+	}
+
+	work := make([]float64, 2*(k+1))
+	prev := work[:k+1]
+	curr := work[k+1:]
+
+	h := dx * float64(n)
+	prev[0] = (f[0] + f[len(f)-1]) * 0.5 * h
+
+	step := n
+	for i := 1; i <= k; i++ {
+		h /= 2
+		step /= 2
+		var estimate float64
+		for j := 0; j < n/2; j += step {
+			estimate += f[2*j+step]
+		}
+
+		curr[0] = estimate*h + 0.5*prev[0]
+		for j := 1; j <= i; j++ {
+			factor := math.Pow(4, float64(j))
+			curr[j] = (factor*curr[j-1] - prev[j-1]) / (factor - 1)
+		}
+
+		prev, curr = curr, prev
+	}
+	return prev[k]
+}
--- a/integrate/romberg_test.go
+++ b/integrate/romberg_test.go
@@ -1,0 +1,86 @@
+// Copyright ©2019 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package integrate
+
+import (
+	"math"
+	"testing"
+
+	"gonum.org/v1/gonum/floats"
+)
+
+func TestRomberg(t *testing.T) {
+	const n = 1<<8 + 1
+	x := floats.Span(make([]float64, n), 0, 1)
+
+	for i, test := range []struct {
+		x    []float64
+		f    func(x float64) float64
+		want float64
+		tol  float64
+	}{
+		{
+			x:    x,
+			f:    func(x float64) float64 { return x },
+			want: 0.5,
+			tol:  1e-12,
+		},
+		{
+			x:    x,
+			f:    func(x float64) float64 { return x * x },
+			want: 1.0 / 3.0,
+			tol:  1e-12,
+		},
+		{
+			x:    x,
+			f:    func(x float64) float64 { return x * x * x },
+			want: 1.0 / 4.0,
+			tol:  1e-12,
+		},
+		{
+			x:    x,
+			f:    func(x float64) float64 { return math.Sqrt(x) },
+			want: 2.0 / 3.0,
+			tol:  1e-4,
+		},
+		{
+			x:    x,
+			f:    func(x float64) float64 { return math.Sin(math.Pi * x) },
+			want: 2.0 / math.Pi,
+			tol:  1e-12,
+		},
+		{
+			x:    floats.Span(make([]float64, 3), 0, 1),
+			f:    func(x float64) float64 { return x * x },
+			want: 1.0 / 3.0,
+			tol:  1e-12,
+		},
+		{
+			x:    floats.Span(make([]float64, 3), 0, 1),
+			f:    func(x float64) float64 { return x * x * x },
+			want: 1.0 / 4.0,
+			tol:  1e-12,
+		},
+		{
+			x:    x,
+			f:    func(x float64) float64 { return x * math.Exp(-x) },
+			want: (math.Exp(1) - 2) / math.Exp(1),
+			tol:  1e-12,
+		},
+	} {
+		n := len(test.x)
+		y := make([]float64, n)
+		for i, v := range test.x {
+			y[i] = test.f(v)
+		}
+
+		dx := (test.x[n-1] - test.x[0]) / float64(n-1)
+		v := Romberg(y, dx)
+		diff := math.Abs(v - test.want)
+		if diff > test.tol {
+			t.Errorf("test #%d: got=%v want=%v diff=%v\n", i, v, test.want, diff)
+		}
+	}
+}
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -29,6 +29,7 @@ Clayton Northey <clayton.northey@gmail.com>
 Dan Kortschak <dan.kortschak@adelaide.edu.au> <dan@kortschak.io>
 Daniel Fireman <danielfireman@gmail.com>
 Dario Heinisch <dario.heinisch@gmail.com>
+David Kleiven <davidkleiven446@gmail.com>
 David Samborski <bloggingarrow@gmail.com>
 Davor Kapsa <davor.kapsa@gmail.com>
 Delaney Gillilan <delaneygillilan@gmail.com>


integrate/romberg.go
```diff
@@ -0,0 +1,98 @@
...
+// See https://en.wikipedia.org/wiki/Romberg%27s_method for a thorough description
+// of the algorithm.
+//
+// The length of the length of the f array must be 2**k + 1, where k is a positive integer
+// The minimum length of the work array is 2*k
+// The method returns the value of the integral, and the number of levels needed to reach
+// the desired accuracy
+func Romberg(f []float64, acc float64, work []float64) (float64, int) {
+	ninterv := len(f) - 1
+	// Check that number of intervals is equal to a power of 2
```
kortschak: This is declared too early and the comment looks well out of place. Lower?

integrate/romberg.go
```diff
@@ -0,0 +1,98 @@
...
+
+		// Swap buffers
+		tmp := Rprev
+		Rprev = Rcur
+		Rcur = tmp
+	}
+	return Rprev[k-1], k
+}
+
+//
```
kortschak: Please don't decorate comments.

I'm not sure about this function. We do have similar kind of help elsewhere, but it usually a whole level of abstraction above, not just a slight rename in the same package. I'd rather leave this out and we can add it later if we need (with a better name).

integrate/romberg_test.go
```diff
@@ -0,0 +1,102 @@
...
+package integrate
+
+import (
+	"math"
+	"testing"
+
+	"gonum.org/v1/gonum/floats"
+)
+
+func TestRomberg(t *testing.T) {
```
kortschak: Can you make this table-driven please.

integrate/romberg.go
```diff
@@ -0,0 +1,98 @@
...
+// increase the number of evaluation points until the desired accuracy
+// is obtained or the maximum number of evaluation points has been used.
+// See https://en.wikipedia.org/wiki/Romberg%27s_method for a thorough description
+// of the algorithm.
+//
+// The length of the length of the f array must be 2**k + 1, where k is a positive integer
+// The minimum length of the work array is 2*k
+// The method returns the value of the integral, and the number of levels needed to reach
+// the desired accuracy
+func Romberg(f []float64, acc float64, work []float64) (float64, int) {
```
vladimir-ch: Specifying the accuracy tolerance `acc` makes sense for an automatic algorithm so that it can evaluate the integrand only as many times as necessary to reach it. However, here we have _samples_ of the function (the function has already been evaluated) and there is no benefit in not using all of them. Also, once this change is made it makes no sense to return the number of levels used (because all are used).

On the other hand, the function could take the spacing between samples, which would enable arbitrary (finite) integration intervals, although even this parameter can be avoided - the function can assume spacing of 1 but the user then has to rescale the result.

Finally, I'm not sure about the `work` slice. In Gonum we do try to reuse memory but here it spoils what could otherwise be a very clear and simple API:
```
func Romberg(f []float64, dx float64) float64
```
or even
```
func Romberg(f []float64) float64
```
Let's see what others think but I would remove it.

btracey: I don't think we should include a work slice. AFAIK there are only two places we use it 1) Lapack, but that's different and we have an API around it  2) FFT, and that is also avoidable. 

More generally though, this seems like it's much more suited to `quad` than here. The idea is to be an adaptive algorithm, which is what quad is about.

You could make it a non-adaptive rule (and just evaluate the triangles), but then it's just a special case of the Trapezoidal rule, isn't it? (been a while since I looked at Romberg)

kortschak: I'd be happy for the work slice to go away.

integrate/romberg.go
```diff
@@ -0,0 +1,98 @@
...
+package integrate
+
+import (
+	"math"
+)
+
+//
+// Helper method for getting the base 2 exponent of an integer
+//
+func getBase2Exponent(x int) int {
```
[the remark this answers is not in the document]
davidkleiven: Replaced by `bits.Len`

integrate/romberg.go
```diff
@@ -0,0 +1,98 @@
...
+// is obtained or the maximum number of evaluation points has been used.
+// See https://en.wikipedia.org/wiki/Romberg%27s_method for a thorough description
+// of the algorithm.
+//
+// The length of the length of the f array must be 2**k + 1, where k is a positive integer
+// The minimum length of the work array is 2*k
+// The method returns the value of the integral, and the number of levels needed to reach
+// the desired accuracy
+func Romberg(f []float64, acc float64, work []float64) (float64, int) {
+	ninterv := len(f) - 1
```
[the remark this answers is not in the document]
davidkleiven: Check that the length is larger or equal to 3 before passing the length to the `bits.Len` function.

integrate/romberg.go
```diff
@@ -0,0 +1,98 @@
...
+	// Check that number of intervals is equal to a power of 2
+	dx := 1.0
+
+	k := getBase2Exponent(ninterv)
+
+	n := 1 << uint(k)
+	if n != ninterv {
+		panic("The length of the function array passed has to be 1 + 2^k. For some integer k.")
```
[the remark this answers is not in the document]
davidkleiven: Changed message according to the suggestion

integrate/romberg.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+
+	if len(work) < 2*k {
+		panic("The length of the work array has to be at least 2*k, when len(f) = 2^k + 1")
```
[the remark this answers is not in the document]
davidkleiven: Work array is not passed anymore, so this check is removed

integrate/romberg.go
```diff
@@ -0,0 +1,98 @@
...
+		for j := 1; j <= i; j++ {
+			factor := math.Pow(4, float64(j))
+			Rcur[j] = (factor*Rcur[j-1] - Rprev[j-1]) / (factor - 1)
+		}
+
+		if math.Abs(Rprev[i-1]-Rcur[i]) < acc {
+			return Rcur[i], i
+		}
+
+		// Swap buffers
```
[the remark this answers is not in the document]
davidkleiven: It is done like this now

integrate/romberg.go
```diff
@@ -0,0 +1,98 @@
...
+	n := 1 << uint(k)
+	if n != ninterv {
+		panic("The length of the function array passed has to be 1 + 2^k. For some integer k.")
+	}
+
+	if len(work) < 2*k {
+		panic("The length of the work array has to be at least 2*k, when len(f) = 2^k + 1")
+	}
+
+	Rprev := work[:k]
```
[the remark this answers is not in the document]
davidkleiven: Names are changed according to the suggestion

integrate/romberg.go
davidkleiven: This comment is removed

integrate/romberg.go
davidkleiven: I removed the function

integrate/romberg_test.go
davidkleiven: Tests are now on the same format as **trapezoidal_test**

davidkleiven:
Good points, I think all your suggestions have been addressed.

integrate/romberg_test.go
```diff
@@ -0,0 +1,66 @@
...
+		x    []float64
+		f    func(x float64) float64
+		want float64
+		tol  float64
+	}{
+		{
+			x:    x,
+			f:    func(x float64) float64 { return x },
+			want: 0.5,
+			tol:  1E-12,
```
kortschak: Please lower-case the floats.

integrate/romberg_test.go
```diff
@@ -0,0 +1,66 @@
...
+
+import (
+	"math"
+	"testing"
+
+	"gonum.org/v1/gonum/floats"
+)
+
+func TestRomberg(t *testing.T) {
+	const N = 1<<16 + 1
```
kortschak: s/N/n/g

integrate/romberg.go
```diff
@@ -0,0 +1,59 @@
...
+	}
+	ninterv := len(f) - 1
+	k := bits.Len(uint(ninterv - 1))
+
+	n := 1 << uint(k)
+	if n != ninterv {
+		panic("integral: invalid slice length: must be 2^k + 1")
+	}
+
+	prev := make([]float64, k)
```
kortschak: These allocations can be done as a single alloc and then sliced in half.

```
work := make([]float64, 2*k)
prev := work[:k:k]
curr := work[k:]
```

integrate/romberg.go
```diff
@@ -0,0 +1,59 @@
...
+	curr := make([]float64, k)
+
+	h := dx * float64(len(f))
+	prev[0] = (f[0] + f[len(f)-1]) * 0.5 * h
+
+	for i := 1; i < k; i++ {
+		h /= 2.0
+		estimate := 0.0
```
kortschak: `var estimate float64`

integrate/romberg.go
```diff
@@ -0,0 +1,59 @@
...
+	}
+
+	prev := make([]float64, k)
+	curr := make([]float64, k)
+
+	h := dx * float64(len(f))
+	prev[0] = (f[0] + f[len(f)-1]) * 0.5 * h
+
+	for i := 1; i < k; i++ {
+		h /= 2.0
```
kortschak: `h /= 2` or `h *= 0.5`

integrate/romberg.go
```diff
@@ -0,0 +1,59 @@
...
+		estimate := 0.0
+		step := ninterv >> uint32(i)
```
kortschak: Just `uint` is fine.

integrate/romberg.go
```diff
@@ -0,0 +1,60 @@
...
+// license that can be found in the LICENSE file.
+
+package integrate
+
+import (
+	"math"
+	"math/bits"
+)
+
+// Romberg estimates the integral of the function f
```
vladimir-ch: I'd write something like:
```
// Romberg returns an approximate value of the integral
//  \int_a^b f(x)dx
// computed using the Romberg's method. The function f is given
// as a slice of equally-spaced samples, that is,
//  f[i] = f(a + i*dx)
// and dx is the spacing between the samples.
//
// The length of f must be 2^k + 1, where k is a positive integer,
// and dx must be positive.
//
// See https://en.wikipedia.org/wiki/Romberg%27s_method for a description of the algorithm.
```

integrate/romberg.go
```diff
@@ -0,0 +1,60 @@
...
+// Romberg's method approximates the integral of the function by
+// applying Richardson extrapolation.
+// See https://en.wikipedia.org/wiki/Romberg%27s_method for a thorough description
+// of the algorithm.
+//
+// The length of the length of the f array must be 2**k + 1, where k is a positive integer
+// The method returns the value of the integral
+func Romberg(f []float64, dx float64) float64 {
+	if len(f) < 3 {
+		panic("integral: invalid slice length: must be 2^k + 1")
```
vladimir-ch: The message does not match the condition.

I'd also add a check that `dx` is positive.

integrate/romberg.go
```diff
@@ -0,0 +1,60 @@
...
+	work := make([]float64, 2*k)
+	prev := work[:k]
+	curr := work[k:]
+
+	h := dx * float64(len(f))
+	prev[0] = (f[0] + f[len(f)-1]) * 0.5 * h
+
+	for i := 1; i < k; i++ {
+		h /= 2
+		var estimate float64
```
vladimir-ch: Please move this just before the loop.

integrate/romberg.go
```diff
@@ -0,0 +1,60 @@
...
+		step := ninterv >> uint(i)
```
vladimir-ch: Can't this be implemented as a repeated division by 2 instead? I think it could be clearer than the bit shifting.

integrate/romberg.go
```diff
@@ -0,0 +1,60 @@
...
+	n := 1 << uint(k)
+	if n != ninterv {
+		panic("integral: invalid slice length: must be 2^k + 1")
+	}
+
+	work := make([]float64, 2*k)
+	prev := work[:k]
+	curr := work[k:]
+
+	h := dx * float64(len(f))
```
vladimir-ch: Is `h` supposed to be equal to `b - a`? If that's the case, this should have `len(f)-1`. It would be nice to have a test that fails due to this; it will need a test with few samples to make a difference of one `dx` significant.

integrate/romberg.go
```diff
@@ -0,0 +1,60 @@
...
+// The length of the length of the f array must be 2**k + 1, where k is a positive integer
+// The method returns the value of the integral
+func Romberg(f []float64, dx float64) float64 {
+	if len(f) < 3 {
+		panic("integral: invalid slice length: must be 2^k + 1")
+	}
+	ninterv := len(f) - 1
+	k := bits.Len(uint(ninterv - 1))
+
+	n := 1 << uint(k)
```
vladimir-ch: Since `n` is used only on the next line, I would compute the power of 2 inline in the condition. That would free the variable name `n`, so I would rename `ninterv` to `n`. Something like this:
```
	n := len(f) - 1
	k := bits.Len(uint(n)) - 1
	if 1 << k + 1 != n {
		panic("integral: invalid slice length: must be 2^k + 1")
	}
```
The panic message then nicely matches the condition.

integrate/romberg_test.go
```diff
@@ -0,0 +1,66 @@
...
+
+import (
+	"math"
+	"testing"
+
+	"gonum.org/v1/gonum/floats"
+)
+
+func TestRomberg(t *testing.T) {
+	const n = 1<<16 + 1
```
vladimir-ch: Too many points. I'd use 1000 at most. See also my comment about `h`.

integrate/romberg_test.go
```diff
@@ -0,0 +1,66 @@
...
+			x:    x,
+			f:    func(x float64) float64 { return math.Sqrt(x) },
+			want: 2.0 / 3.0,
+			tol:  1e-4,
+		},
+		{
+			x:    x,
+			f:    func(x float64) float64 { return math.Sin(math.Pi * x) },
+			want: 2.0 / math.Pi,
+			tol:  1e-4,
```
vladimir-ch: Why is such a high tolerance necessary with so many points and a smooth function?

integrate/romberg_test.go
```diff
@@ -0,0 +1,66 @@
...
+		},
+	} {
+		y := make([]float64, len(test.x))
+		for i, v := range test.x {
+			y[i] = test.f(v)
+		}
+
+		dx := (test.x[n-1] - test.x[0]) / n
```
vladimir-ch: `n` is the number of samples but we need to divide by the number of intervals, that is, `n-1`.

integrate/romberg_test.go
```diff
@@ -0,0 +1,66 @@
...
+		v := Romberg(y, dx)
+		if !floats.EqualWithinAbs(v, test.want, test.tol) {
```
vladimir-ch: When a test fails, I like to see the actual difference explicitly. Therefore I'd write this as:
```
diff := math.Abs(v - test.want)
if diff > tol {
    t.Errorf(" ... diff=%v", ..., diff)
}
```

integrate/romberg_test.go
vladimir-ch: With `1<<8 + 1` and after fixing the bugs all test cases pass with `tol = 1e-14` except for the Sqrt one which needs `1e-4`. The polynomial tests would of course pass already with n=3 or 5.

integrate/romberg_test.go
```diff
@@ -0,0 +1,66 @@
...
+		},
+		{
+			x:    x,
+			f:    func(x float64) float64 { return x * x * x },
+			want: 1.0 / 4.0,
+			tol:  1e-12,
+		},
+		{
+			x:    x,
+			f:    func(x float64) float64 { return math.Sqrt(x) },
```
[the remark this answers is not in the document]
vladimir-ch: It's ok to keep this function and you can add the suggested one as another test case.

integrate/romberg.go
vladimir-ch: The +1 and -1 in my example have to be adjusted for it to be correct but the idea is there.

integrate/romberg.go
vladimir-ch: With `n = 1<<8 + 1` this bug demonstrates itself.

integrate/romberg.go
davidkleiven: Good

integrate/romberg_test.go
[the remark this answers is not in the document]
davidkleiven: I added a variant of this test (mainly to have a simple exact answer). There is a test case for `x*math.Exp(x)`

integrate/romberg_test.go
davidkleiven: No, it is not nessecary. It is fine with `1e-12`

integrate/romberg_test.go
davidkleiven: Done

davidkleiven:
@vladimir-ch 
> Is h supposed to be equal to b - a?

Yes, it is the case. And it is fixed, together with all your other suggestions.

integrate/romberg_test.go
davidkleiven: Done

integrate/romberg.go
davidkleiven: There is now a test with only 5 points. However, it tests a polynomial function as then we now the answer should be exact.

integrate/romberg.go
davidkleiven: Yes, it could.

integrate/romberg.go
```diff
@@ -0,0 +1,66 @@
...
+//  \int_a^b f(x)dx
+// computed using the Romberg's method. The function f is given
+// as a slice of equally-spaced samples, that is,
+//  f[i] = f(a + i*dx)
+// and dx is the spacing between the samples.
+//
+// The length of f must be 2^k + 1, where k is a positive integer,
+// and dx must be positive.
+//
+// See https://en.wikipedia.org/wiki/Romberg%27s_method for a description of t
```
kortschak: Is this sentence truncated?

integrate/romberg.go
```diff
@@ -0,0 +1,66 @@
...
+func Romberg(f []float64, dx float64) float64 {
+	if len(f) < 3 {
+		panic("integral: invalid slice length: has to be at least 3")
```
kortschak: s/has to/must/

integrate/romberg.go
davidkleiven: Yes, you're right. Sorry about that.

integrate/romberg.go
davidkleiven: Done

integrate/romberg.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+
+	work := make([]float64, 2*k)
+	prev := work[:k]
+	curr := work[k:]
+
+	h := dx * float64(len(f)-1)
+	prev[0] = (f[0] + f[len(f)-1]) * 0.5 * h
+
+	var estimate float64
```
vladimir-ch: No need to have this here, move it into the loop body.

integrate/romberg_test.go
```diff
@@ -0,0 +1,79 @@
...
+			tol:  1e-4,
+		},
+		{
+			x:    x,
+			f:    func(x float64) float64 { return math.Sin(math.Pi * x) },
+			want: 2.0 / math.Pi,
+			tol:  1e-12,
+		},
+		{
+			x:    floats.Span(make([]float64, 5), 0, 1),
```
vladimir-ch: I would expect to get a correct result even with 3 points here but that's not the case. I get 0.5 which suggests that only the trapezoidal result is returned. Better check whether the main algorithm loop is correct, especially indices and for loop bounds, maybe by printing the table and comparing it with output from scipy.romb?

integrate/romberg.go
```diff
@@ -0,0 +1,67 @@
...
+	for i := 1; i < k; i++ {
+		h /= 2
+		step /= 2
+		estimate = 0
+		for j := 0; j < n/2; j += step {
+			estimate += f[2*j+step]
+		}
+
+		curr[0] = estimate*h + 0.5*prev[0]
+
```
vladimir-ch: Delete this line.

integrate/romberg.go
vladimir-ch: Now I see it's here because of my suggestion, sorry :-)

integrate/romberg_test.go
```diff
@@ -0,0 +1,79 @@
...
+		y := make([]float64, len(test.x))
+		for i, v := range test.x {
+			y[i] = test.f(v)
+		}
+
+		dx := (test.x[len(test.x)-1] - test.x[0]) / float64(len(test.x)-1)
+		v := Romberg(y, dx)
+		diff := math.Abs(v - test.want)
+		if diff > test.tol {
+			t.Errorf("test #%d: got=%v want=%v diff%v\n", i, v, test.want, diff)
```
vladimir-ch: `diff=`

integrate/romberg.go
davidkleiven: I'll take set it back

integrate/romberg_test.go
davidkleiven: You are absolutely right, for 3 points it should return Simpson. Loop should be `i<=k` not `i<k`

integrate/romberg.go
davidkleiven: I assume you are the empty line. Is that right?

integrate/romberg_test.go
davidkleiven: Done

davidkleiven:
Thanks for your thorogh checking. Many of the things I should have spotted myself of course, but I missed them. @vladimir-ch Your latest suggestion should now be fixed.

integrate/romberg.go
vladimir-ch: Yes, the empty line so that the lines computing `curr` are together.

integrate/romberg.go
vladimir-ch: Ah, I remember. When I wrote to move this line just before the loop, I meant the inner `j` loop because that's the only place where it's used.

integrate/romberg.go
```diff
@@ -0,0 +1,65 @@
...
+// computed using the Romberg's method. The function f is given
+// as a slice of equally-spaced samples, that is,
+//  f[i] = f(a + i*dx)
+// and dx is the spacing between the samples.
+//
+// The length of f must be 2^k + 1, where k is a positive integer,
+// and dx must be positive.
+//
+// See https://en.wikipedia.org/wiki/Romberg%27s_method for a description of
+// the algorithm
```
vladimir-ch: Full stop.

integrate/romberg.go
```diff
@@ -0,0 +1,65 @@
...
+
+	if len(f) != 1<<uint(k)+1 {
+		panic("integral: invalid slice length: must be 2^k + 1")
+	}
+
+	work := make([]float64, 2*(k+1))
+	prev := work[:k+1]
+	curr := work[k+1:]
+
+	h := dx * float64(len(f)-1)
```
vladimir-ch: Replace `len(f)-1` with `n` here and on the next line.

integrate/romberg_test.go
```diff
@@ -0,0 +1,79 @@
...
+			want: 2.0 / 3.0,
+			tol:  1e-4,
+		},
+		{
+			x:    x,
+			f:    func(x float64) float64 { return math.Sin(math.Pi * x) },
+			want: 2.0 / math.Pi,
+			tol:  1e-12,
+		},
+		{
```
vladimir-ch: Add the same test case for `x^3`, should be exact as well.

integrate/romberg_test.go
```diff
@@ -0,0 +1,79 @@
...
+			x:    x,
+			f:    func(x float64) float64 { return x * math.Exp(-x) },
+			want: (math.Exp(1) - 2) / math.Exp(1),
+			tol:  1e-12,
+		},
+	} {
+		y := make([]float64, len(test.x))
```
vladimir-ch: Add line with `n := len(test.x)` and use `n` here and below on the `dx :=` line.

kortschak:
There are two approaches. Either squash all these commits into one and add another commit, "A+C: add David Kleiven" and then we merge is as a single PR. Otherwise, we can squash this in the merge and then another PR to add you to A+C.

AUTHORS
```diff
@@ -21,6 +21,7 @@ Clayton Northey <clayton.northey@gmail.com>
 Dan Kortschak <dan.kortschak@adelaide.edu.au> <dan@kortschak.io>
 Daniel Fireman <danielfireman@gmail.com>
 Dario Heinisch <dario.heinisch@gmail.com>
+David Kleivem <davidkleiven446@gmail.com>
```
vladimir-ch: Kleiven, maybe? :-)

integrate/romberg.go
vladimir-ch: Has this been done?

AUTHORS
kortschak: Also needs to be added to CONTRIBUTORS.

davidkleiven: Yes, done.

integrate/romberg.go
vladimir-ch: I saw that but thought, yeah, we'll fix later. I planned to revisit the other two methods anyway, the tests take too long.

PR #1131, opened 2019-10-15
floats: new norm2 approach to have better performance

neojou:
Would you mind reviewing it? Thanks.

### DIFF
--- a/floats/floats.go
+++ b/floats/floats.go
@@ -143,6 +143,50 @@ func CumSum(dst, s []float64) []float64 {
 	return f64.CumSum(dst, s)
 }
 
+// Norm2 retruns the 2 norm of the slice S, defined as 
+// sqrt(s[0]^2 + s[1]^2 + ... s[i]^2)
+//
+// refered to the approach in math.Hypot() to avoid unnecessary overflow
+// and underflow
+func Norm2(s []float64) float64 {
+	items := len(s)
+
+	if items == 0 {
+		return math.NaN()
+	}
+
+	s0 := s[0]
+
+	switch {
+	case math.IsInf(s0, 0):
+		return math.Inf(1)
+	case math.IsNaN(s0):
+		return math.NaN()
+	}
+
+	twoNorm := math.Abs(s0)
+
+	for i := 1; i < len(s); i++ {
+		switch {
+		case math.IsInf(twoNorm, 0):
+			return math.Inf(1)
+		case math.IsNaN(twoNorm):
+			return math.NaN()
+		}
+		q := math.Abs(s[i])
+
+		if twoNorm < q {
+			twoNorm, q = q, twoNorm
+		}
+		if twoNorm == 0 {
+			continue
+		}
+		q = q / twoNorm
+		twoNorm *= math.Sqrt(1+q*q)
+	}
+	return twoNorm
+}
+
 // Distance computes the L-norm of s - t. See Norm for special cases.
 // A panic will occur if the lengths of s and t do not match.
 func Distance(s, t []float64, L float64) float64 {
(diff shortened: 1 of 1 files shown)


kortschak:
This is more likely to overflow.

vladimir-ch:
`diff * diff` may also underflow leading potentially to an inaccurate result.

rcoreilly:
I was wondering about this too and wrote my own version due to the performance diff.  Seems like you might want to add an option or a "Fast" version or something -- when computing large distance matricies etc the performance can really matter.

kortschak:
Making that kind of addition is very unlikely. First, because it is a doubling of development and maintenance effort and second because providing a fast way and a good way invariably leads to some users not being adequately informed about the risks of the fast way (no matter how much documentation and how doom laden it is). My view is that if people what to do things the wrong way, the amount of code they have to write in this case is minimal and we should be encouraging it.

kortschak:
Very obviously, when I talk about a doubling or work, I'm talking about ours. This is my primary concern when it comes to effort. It is an increase in the amount of work not just for the very small loop that others would have to write, knowing the limitations of the approach, while for us it is the loop, the tests and additional documentation telling people they should not use it, and then answering issues filed by people that ignored the documentation and find that they code is now unsafe.

The reasonable way to approach this would be to add a `Norm2` to "internal/asm/f64" and call that. If someone wants to write it. We would certainly welcome the PR.

rcoreilly:
well if you didn't want a bunch of extra work, why did you make gonum in the first place!?  The whole point of making such a wonderful package is to save everyone *else* all the trouble of doing so, right? :)  And to suffer all their annoying comments :)

Anyway, maybe I'm missing something, but we can't access an `internal` package from outside right?  you mean just to optimize the existing sqrt-based computation?  how much faster would that likely be?

kortschak:
>well if you didn't want a bunch of extra work, why did you make gonum in the first place!? The whole point of making such a wonderful package is to save everyone else all the trouble of doing so, right? :) And to suffer all their annoying comments :)

Work and extra work are different things. Work that is counter to the philosophy of the project is unlikely to be done. I believe that providing a fast and potentially dangerously incorrect implementation is counter to the philosophy of the project.

>you mean just to optimize the existing sqrt-based computation?

Yes.

>how much faster would that likely be?

I don't know. You could get an idea of that by looking at the speed of the current implementation and comparing it to OpenBLAS' `dnrm2`. It is written in assembler.

rcoreilly:
It looks like they use a different algorithm that should be quite fast and just uses one sqrt at the end -- maybe could Go-ify that and make everyone happy?

http://www.netlib.org/lapack/explore-html/de/da4/group__double__blas__level1_ga4b25c539b862414e6f91ebb30b391d7c.html#ga4b25c539b862414e6f91ebb30b391d7c

kortschak:
https://godoc.org/gonum.org/v1/gonum/blas/gonum#Implementation.Dnrm2

neojou:
Thank you everyone for the comments. I learn a lot from this. 
I just notice why math.Hypot() does some to avoid overflow and underflow, and I rewrote it by using similar approaches as math.Hypot() without doing sqrt/square for each sum up of the slice, It is 30% faster then the previous one. I didn't change Distance function yet, since may need a better way to recalculate the float array (s[i] - v), will study further about blas dnrm2 algorithm and rewrite it. (or just follow rcoreilly's change. ^_^) Thanks.

rcoreilly:
@neojou you are unlikely to do better than BLAS, so #1132 is probably the best way to go at this point.  If you could add your benchmark code for Distance to the floats_test file, and run the comparisons, that would be great!

neojou:
@rcoreilly Yes, I agree with you. will use the way  [#1132](https://github.com/gonum/gonum/pull/1132) and do some benchmark as you said. ^^

neojou:
[1134](https://github.com/gonum/gonum/pull/1134) looks great. close this one.

PR #1132, opened 2019-10-16
use blas dnrm2 algorithm for floats.Distance L == 2 case instead of much more expensive Hypot for each

rcoreilly:
test result is off by 1 extra digit -- use approx to fix?

Fixes issues with #1131 per discussion there.


kortschak:
Thanks. Would you post bench mark results comparing the new and old code (please use `benchstat`) and adjust the tolerance in the test.

rcoreilly:
I couldn't find any existing benchmark for the Distance function.  I'm going to bed now so maybe someone else can benchmark it..

rcoreilly:
should Norm just call blas.Dnrm2 or should it have its own local copy of the code?  one way or another it should also be updated.

neojou:
Hi, @rcoreilly , 

Thanks for kind help. I tested the program you wrote, and it looks fine also. 
It is about 4x faster than before ^^ 
https://github.com/neojou/go-ameba/tree/master/tGOpl/ch11

Distance           14028379      80.6 ns/op
Your Distance  51248080       21.8 ns/op

#1134 sounds great. Will study further. Thank you all.

PR #1133, opened 2019-10-17
integrate: rework tests, clean up docs and code

vladimir-ch:
Please take a look.

cc @davidkleiven


davidkleiven:
Agree, this is nice.

PR #1134, opened 2019-10-18
internal/asm,blas,floats: move level 2 norm to asm

kortschak:
Please take a look.

This brings a significant improvement to `floats.Norm(x, 2)` call performance at the cost of a disagreement with `floats.Distance(s, t, 2)` beyond 1e-15. This disagreement could be reconciled by adding a two-parameter `L2Norm` to asm that matches the one-parameter version here. If we want to go ahead with this, I'll do that.

<details>
<table class='benchstat oldnew'>
<tr class='configs'><th><th>old.bench<th>new.bench

<tbody>
<tr><th><th colspan='2' class='metric'>time/op<th>delta
<tr class='better'><td>Norm2Small-8<td>106ns ± 1%<td>29ns ± 4%<td class='delta'>−73.08%<td class='note'>(p=0.000 n=10&#43;10)
<tr class='better'><td>Norm2Medium-8<td>18.1µs ± 1%<td>2.1µs ± 3%<td class='delta'>−88.40%<td class='note'>(p=0.000 n=10&#43;10)
<tr class='better'><td>Norm2Large-8<td>1.81ms ± 0%<td>0.20ms ± 1%<td class='delta'>−88.69%<td class='note'>(p=0.000 n=10&#43;10)
<tr class='better'><td>Norm2Huge-8<td>182ms ± 0%<td>21ms ± 2%<td class='delta'>−88.38%<td class='note'>(p=0.000 n=9&#43;10)
</tbody>

</table>
</details>

Closes #1131.
Updates #1132.

### DIFF
--- a/internal/asm/f64/l2norm.go
+++ b/internal/asm/f64/l2norm.go
@@ -1,0 +1,62 @@
+// Copyright ©2019 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package f64
+
+import "math"
+
+// L2NormUnitary is the level 2 norm of x.
+func L2NormUnitary(x []float64) (sum float64) {
+	var scale float64
+	sumSquares := 1.0
+	for _, v := range x {
+		if v == 0 {
+			continue
+		}
+		absxi := math.Abs(v)
+		if math.IsNaN(absxi) {
+			return math.NaN()
+		}
+		if scale < absxi {
+			s := scale / absxi
+			sumSquares = 1 + sumSquares*s*s
+			scale = absxi
+		} else {
+			s := absxi / scale
+			sumSquares += s * s
+		}
+	}
+	if math.IsInf(scale, 1) {
+		return math.Inf(1)
+	}
+	return scale * math.Sqrt(sumSquares)
+}
+
+// L2NormInc is the level 2 norm of x.
+func L2NormInc(x []float64, n, incX uintptr) (sum float64) {
+	var scale float64
+	sumSquares := 1.0
+	for ix := uintptr(0); ix < n*incX; ix += incX {
+		val := x[ix]
+		if val == 0 {
+			continue
+		}
+		absxi := math.Abs(val)
+		if math.IsNaN(absxi) {
+			return math.NaN()
+		}
+		if scale < absxi {
+			s := scale / absxi
+			sumSquares = 1 + sumSquares*s*s
+			scale = absxi
+		} else {
+			s := absxi / scale
+			sumSquares += s * s
+		}
+	}
+	if math.IsInf(scale, 1) {
+		return math.Inf(1)
+	}
+	return scale * math.Sqrt(sumSquares)
+}
--- a/internal/asm/f64/l2norm.go
+++ b/internal/asm/f64/l2norm.go
@@ -60,3 +60,31 @@ func L2NormInc(x []float64, n, incX uintptr) (sum float64) {
 	}
 	return scale * math.Sqrt(sumSquares)
 }
+
+// L2DistanceUnitary is the L2 norm of x-y.
+func L2DistanceUnitary(x, y []float64) (sum float64) {
+	var scale float64
+	var sumSquares float64 = 1
+	for i, v := range x {
+		v -= y[i]
+		if v == 0 {
+			continue
+		}
+		absxi := math.Abs(v)
+		if math.IsNaN(absxi) {
+			return math.NaN()
+		}
+		if scale < absxi {
+			s := scale / absxi
+			sumSquares = 1 + sumSquares*s*s
+			scale = absxi
+		} else {
+			s := absxi / scale
+			sumSquares += s * s
+		}
+	}
+	if math.IsInf(scale, 1) {
+		return math.Inf(1)
+	}
+	return scale * math.Sqrt(sumSquares)
+}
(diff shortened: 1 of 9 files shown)


rcoreilly:
impressive speedups!  adding the two-parameter L2Norm seems like a good way to go to keep it consistent.

internal/asm/f64/l2norm.go
```diff
@@ -0,0 +1,62 @@
+// Copyright ©2019 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package f64
+
+import "math"
+
+// L2NormUnitary is the level 2 norm of x.
+func L2NormUnitary(x []float64) (sum float64) {
```
kortschak: The alternative that I'm considering is
```
// L2NormUnitary is
//  sumSquares := 1.0
//  for _, v := range x {
//  	if v == 0 {
//  		continue
//  	}
//  	absxi := math.Abs(v)
//  	if math.IsNaN(absxi) {
//  		return math.NaN(), math.NaN()
//  	}
//  	if scale < absxi {
//  		s := scale / absxi
//  		sumSquares = 1 + sumSquares*s*s
//  		scale = absxi
//  	} else {
//  		s := absxi / scale
//  		sumSquares += s * s
//  	}
//  }
//  return scale, math.Sqrt(sumSquares)
func L2NormUnitary(x []float64) (scale, sum float64) {
	sumSquares := 1.0
	for _, v := range x {
		if v == 0 {
			continue
		}
		absxi := math.Abs(v)
		if math.IsNaN(absxi) {
			return math.NaN(), math.NaN()
		}
		if scale < absxi {
			s := scale / absxi
			sumSquares = 1 + sumSquares*s*s
			scale = absxi
		} else {
			s := absxi / scale
			sumSquares += s * s
		}
	}
	return scale, math.Sqrt(sumSquares)
}
```
Where the rescaling and infinity check is performed by the caller. It's also possible to omit the NaN check and just let that propagate through naturally.

Kunde21: I'm not sure how this helps, because `X * +Inf = +Inf` by rule.  Separating the scaling makes the API less friendly and I don't know that the benefit is worth it.

kortschak: Yeah, I think you're right. There's stuff in the current code that's probably there because it comes from code that is decades old and predates the current numerical standards. We can probably elide the inf check and then this becomes as simple as the original as far as the API goes.

vladimir-ch: > The other question is whether the zero check is necessary...

I was wondering the same... I guess it depends on the number of zeros but since these are dense routines I would expect that typically not many. Perhaps a benchmark?

vladimir-ch: Regarding the inf check, scale can become inf only if a vector element is inf. Then if any following element is also inf, a nan will be added to sumSquares and without the final check for inf scale a nan will be returned. This would be problematic: without the check one inf element returns inf, more inf elements return nan.

Returning nan if _any_ inf element occurs could actually be reasonable. The mathematical definition of a norm does not count with vectors with inf elements because of the property: `||av|| = |a| ||v||` for all scalars `a` and vectors `v`. When a is 0 and v has inf elements `av` is undefined because `0xInf`. `a=0` case can be excluded but then it's not the standard norm strictly speaking. Returning nan would be consistent with other functions like Scale where scaling with 0 also turns any inf elements into nan but may be less expected than Inf?

Whatever the decision, there should be test cases that cover it.

kortschak: I think to most users, return NaN for an Inf-containing vector would be surprising. Both because the blas code that we currently have (and from history, the BLAS and cblas code) have the behaviour that is here, and the a length of a vector with an infinite component feels like it should be infinite.

vladimir-ch: Sure but as I wrote, the final inf check should stay or the code should return quickly.

kortschak: Yes, the suggestion here was to have the check for `math.IsInf(scale)` be done by the caller on return from this.

Kunde21: I don't think we're gaining much with the zero check, tbh.  Unless we expect a lot of zeros in the input slice, the cost of a branch there isn't going to be measurably faster in the common case.

Kunde21: MIght be better to do an initial loop to discard any leading zeros in the slice, then start calculating from the first non-zero value.

kortschak: I think leaving the code as is is the best first iteration. If we see that there are possible optimisations we can add them in if they pay back enough.

The reason for my suggestion here was entirely to simplify the assembly if it came. If that's not worth doing, then we should use the form that exists in the PR.

kortschak:
By way of additional comment. This PR is primarily a refactor. I don't want to do any major algorithmic changes here.

internal/asm/f64/l2norm.go
Kunde21: I'm good to leave it as-is.  Readable and faithfully implements the netlib algorithm.
I'll likely use a different approach for the asm, because this algorithm is very difficult to make concurrent.

kortschak: Locally, I have an `L2DistanceUnitary` in floats that matches the algorithm here. Should I include that in asm so that you can attack that at the same time?

Kunde21: Yeah, please do.

kortschak: Done.

PR #1135, opened 2019-10-18
integrate: small cleanup in Simpsons

vladimir-ch:
A follow-up to #1133 

Please take a look.

### DIFF
--- a/integrate/simpsons.go
+++ b/integrate/simpsons.go
@@ -27,31 +27,6 @@ func Simpsons(x, f []float64) float64 {
 		panic("integrate: must be sorted")
 	}
 
-	// Simpson's method approximates the integral of a function f by subdividing
-	// the interval of integration into segments and applying a method that fits
-	// a polynomial to each subinterval.
-	//
-	// This implementation makes use of the following composite Simpson's method
-	// to estimate \int_a^b f(x) dx where a, b are x at x[0] and x[len(x)-1]
-	// respectively:
-	//  \sum_{i=1}^{N} {a_0}_i * f_{i-1} + {a_1}_i * f_{i} + {a_2}_i * f_{i+1}
-	// where N is the count of subintervals and {a_0}_i, {a_1}_i, and {a_2}_i
-	// are constants at index i given by:
-	//  {a_0}_i = 2 * h^3_i - h^3_{i+1} + 3 * h_{i+1} * h^2_i / 6 * h_i * (h_i + h_{i+1})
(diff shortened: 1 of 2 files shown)


integrate/simpsons.go
```diff
@@ -27,31 +27,6 @@ func Simpsons(x, f []float64) float64 {
 		panic("integrate: must be sorted")
 	}
 
-	// Simpson's method approximates the integral of a function f by subdividing
```
vladimir-ch: In the end I decided to delete this comment. It just repeats what's already in the code and moreover the code is probably clearer.

integrate/simpsons_test.go
```diff
@@ -21,20 +21,24 @@ func TestSimpsons(t *testing.T) {
 		tol      float64
 	}{
 		{integral: testquad.Constant(0), n: 3, tol: 0},
+		{integral: testquad.Constant(0), n: 4, tol: 0},
```
vladimir-ch: The intention is to have a test with the minimum even number of nodes.

integrate/simpsons_test.go
```diff
@@ -21,20 +21,24 @@ func TestSimpsons(t *testing.T) {
...
-		{integral: testquad.Poly(3), n: 1e3, tol: 1e-8},
-		{integral: testquad.Poly(4), n: 1e3, tol: 1e-8},
-		{integral: testquad.Poly(5), n: 1e3, tol: 1e-7},
-		{integral: testquad.Sin(), n: 1e2, tol: 1e-8},
-		{integral: testquad.XExpMinusX(), n: 1e2, tol: 1e-8},
+		{integral: testquad.Poly(3), n: 1000, tol: 1e-8},
+		{integral: testquad.Poly(4), n: 1000, tol: 1e-8},
+		{integral: testquad.Poly(5), n: 1000, tol: 1e-7},
+		{integral: testquad.Sin(), n: 100, tol: 1e-8},
+		{integral: testquad.XExpMinusX(), n: 101, tol: 1e-8},
```
vladimir-ch: This test was failing in my virtual machine, plus there will be a test case with a large odd number of nodes.

kortschak: Also failing on travis and appveyor:
```
--- FAIL: TestSimpsons (0.00s)
    simpsons_test.go:58: Test #16: ∫_0^1 x*exp(-x)dx, n=101: unexpected result; got=0.26424112862624427 want=0.26424111765711533 diff=1.096912893849833e-08
```

vladimir-ch: Yes, I noticed it too late. Let's see if it passes now.

PR #1146, opened 2019-10-31
graph/product: add extended modular product function

kortschak:
Please take a look.

ISTM this is too open to test in a more specific way.

### DIFF
--- a/graph/product/product_ext_example_test.go
+++ b/graph/product/product_ext_example_test.go
@@ -1,0 +1,131 @@
+// Copyright ©2019 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package product_test
+
+import (
+	"fmt"
+
+	"gonum.org/v1/gonum/graph"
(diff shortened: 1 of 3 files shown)


graph/product/product_ext_example_test.go
```diff
@@ -0,0 +1,131 @@
...
+	p := simple.NewDirectedGraph()
+	product.ModularExt(p, people, pattern, func(a, b graph.Edge) bool {
+		return a.(simple.WeightedEdge).Weight() == b.(simple.WeightedEdge).Weight()
+	})
+
+	// Find the maximal cliques in the undirected induction
+	// of the modular product.
+	mc := topo.BronKerbosch(undirected{p})
+
+	// Report the the cliques that are identical in order to the pattern.
```
vladimir-ch: Double "the".

kortschak: Fixed

kortschak:
I think I'm actually OK with how the testing works. The extended function agrees with the plain modular product with this accept all edges agreement. The function is pretty trivial.

vladimir-ch:
Sure, no problem.

PR #1148, opened 2019-11-01
graph: clean up lint

kortschak:
These were all identified by golangci-lint. The remaining two lint issues are false positives:
```
encoding/dot/decode.go:152:2: `graphAttr` is unused (structcheck)
	graphAttr, nodeAttr, edgeAttr encoding.AttributeSetter
	^
topo/2sat_example_test.go:177:1: tests: ExampleTarjanSCC_2sat refers to unknown field or method: TarjanSCC.2sat (govet)
func ExampleTarjanSCC_2sat() {
^
```
The first is golangci/golangci-lint#826 and the second is golang/go#23864.

Please take a look.

### DIFF
--- a/graph/simple/weighted_directed_test.go
+++ b/graph/simple/weighted_directed_test.go
@@ -160,7 +160,7 @@ func TestWeightedDirected(t *testing.T) {
 
 // Tests Issue #27
 func TestWeightedEdgeOvercounting(t *testing.T) {
-	g := generateDummyGraph()
+	g := generateDummyWeightedGraph()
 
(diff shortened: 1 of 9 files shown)


graph/simple/weighted_directed_test.go
```diff
@@ -160,7 +160,7 @@ func TestWeightedDirected(t *testing.T) {
 
 // Tests Issue #27
 func TestWeightedEdgeOvercounting(t *testing.T) {
-	g := generateDummyGraph()
+	g := generateDummyWeightedGraph()
```
kortschak: This was a bug.

PR #1153, opened 2019-11-04
optimize: clean up lint

kortschak:
Please take a look.

### DIFF
--- a/optimize/newton.go
+++ b/optimize/newton.go
@@ -152,9 +152,11 @@ func (n *Newton) NextDirection(loc *Location, dir []float64) (stepSize float64) 
 		pd := n.chol.Factorize(n.hess)
 		if pd {
 			// Store the solution in d's backing array, dir.
-			_ = n.chol.SolveVecTo(d, grad)
-			d.ScaleVec(-1, d)
-			return 1
+			err := n.chol.SolveVecTo(d, grad)
+			if err == nil {
+				d.ScaleVec(-1, d)
+				return 1
+			}
 		}
 		// Modified Hessian is not PD, so increase tau.
 		n.tau = math.Max(n.Increase*n.tau, 0.001)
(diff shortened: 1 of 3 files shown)


optimize/newton.go
```diff
@@ -152,7 +152,7 @@ func (n *Newton) NextDirection(loc *Location, dir []float64) (stepSize float64)
 		pd := n.chol.Factorize(n.hess)
 		if pd {
 			// Store the solution in d's backing array, dir.
-			n.chol.SolveVecTo(d, grad)
+			_ = n.chol.SolveVecTo(d, grad)
```
kortschak: Are we comfortable with this? Does it deserved a comment?

sbinet: perhaps introduce a `must` function?
```go
func must(err error) {
    if err != nil {
        panic(err)
    }
}
```
and use that? (should be inlined by the compiler.)

kortschak: The issue is not that, but whether this is guaranteed to not return an error. If it can, then we need to consider what to do. Whether panicking is OK is sort of the question there; if it is, then I'd just do a conditional panic.

vladimir-ch: I'd say this should check the error:
```
err := n.chol.SolveVecTo(d, grad)
if err == nil {
    d.ScaleVec(-1, d)
    return 1
}
```
`Cholesky.SolveVecTo` returns an error only if the modified Hessian is badly conditioned. It will compute something but  it may not be a descent direction, leading to an `ErrNonDescentDirection`. This modification increases the chance of overcoming this error at the expense of an extra Cholesky factorization.

kortschak: Done

kortschak:
Travis appears to be broken in how it reports back to Github; the pr build has succeeded, but is not being reflected here. So I am rebasing anyway.

PR #1154, opened 2019-11-04
stat: clean up lint

kortschak:
Please take a look.

### DIFF
--- a/stat/distmv/normal.go
+++ b/stat/distmv/normal.go
@@ -110,7 +110,10 @@ func NewNormalPrecision(mu []float64, prec *mat.SymDense, src rand.Source) (norm
 		return nil, false
 	}
 	var sigma mat.SymDense
-	chol.InverseTo(&sigma)
+	err := chol.InverseTo(&sigma)
+	if err != nil {
+		return nil, false
+	}
 	return NewNormal(mu, &sigma, src)
 }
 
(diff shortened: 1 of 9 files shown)


stat/distmv/normal.go
```diff
@@ -341,7 +344,7 @@ func (n *Normal) ScoreInput(score, x []float64) []float64 {
 	copy(tmp, x)
 	floats.Sub(tmp, n.mu)
 
-	n.chol.SolveVecTo(mat.NewVecDense(len(score), score), mat.NewVecDense(len(tmp), tmp))
+	_ = n.chol.SolveVecTo(mat.NewVecDense(len(score), score), mat.NewVecDense(len(tmp), tmp))
```
kortschak: Are we comfortable with this? Does it deserved a comment?

stat/distmv/statdist.go
```diff
@@ -335,15 +335,15 @@ func (Wasserstein) DistNormal(l, r *Normal) float64 {
 
 	// Compute Σ_l^(1/2)
 	var ssl mat.SymDense
-	ssl.PowPSD(&l.sigma, 0.5)
+	_ = ssl.PowPSD(&l.sigma, 0.5)
```
kortschak: Are we comfortable with this and the `PowPSD` call below? Do they deserved comments?

kortschak:
*ping* @btracey

stat/distmv/normal.go
btracey: Maybe grab the err and panic if it's not okay? Right now this is fine because it cannot err if the factorization is okay. The check is at least defensive against changes.

stat/distmv/statdist.go
btracey: Maybe same comment as above?

stat/distmv/normal.go
kortschak: Done

stat/distmv/statdist.go
kortschak: Done

PR #1155, opened 2019-11-04
translate netlib l2norm algorithm to asm and remove branches

Kunde21:
Netlib algorithm reduces overflow while calculating the l2norm of a
vector.

Translated to asm while reducing branches in NaN and Inf checks.
Overflow protection is equivalent to the Netlib standard implementation.

Builds upon #1134 to further speed up L2Norm operations.

<details>
<summary>Speedup ~-33%</summary>

```
name                                 old time/op    new time/op    delta
L2NormNetlib/L2NormUnitary-1-8         5.08ns ± 2%    5.20ns ± 3%   +2.36%  (p=0.019 n=10+10)
L2NormNetlib/L2NormUnitary-3-8         9.00ns ± 2%    8.64ns ± 3%   -3.95%  (p=0.000 n=10+10)
L2NormNetlib/L2NormUnitary-10-8        27.6ns ± 4%    22.1ns ± 1%  -19.84%  (p=0.000 n=10+9)
L2NormNetlib/L2NormUnitary-30-8        89.5ns ± 6%    61.8ns ± 1%  -31.00%  (p=0.000 n=10+10)
L2NormNetlib/L2NormUnitary-100-8        314ns ± 1%     213ns ± 1%  -32.06%  (p=0.000 n=9+9)
L2NormNetlib/L2NormUnitary-300-8        977ns ± 3%     650ns ± 2%  -33.47%  (p=0.000 n=10+10)
L2NormNetlib/L2NormUnitary-1000-8      3.28µs ± 2%    2.17µs ± 2%  -33.83%  (p=0.000 n=10+10)
L2NormNetlib/L2NormUnitary-3000-8      9.89µs ± 1%    6.47µs ± 0%  -34.58%  (p=0.000 n=9+9)
L2NormNetlib/L2NormUnitary-10000-8     33.0µs ± 1%    21.6µs ± 0%  -34.62%  (p=0.000 n=8+8)
L2NormNetlib/L2NormUnitary-30000-8     97.9µs ± 1%    64.9µs ± 1%  -33.71%  (p=0.000 n=10+10)
L2NormNetlib/L2NormUnitary-100000-8     323µs ± 0%     216µs ± 0%  -33.27%  (p=0.000 n=9+8)
```
</details> 

Please take a look.

edit: 

Though this doesn't directly change any floats code, updates via #1134:
<details>
<summary>Floats</summary>

```
name           old time/op  new time/op  delta
Norm2Small-8   27.6ns ± 7%  21.1ns ± 8%  -23.66%  (p=0.000 n=10+10)
Norm2Medium-8  2.12µs ± 3%  1.24µs ± 2%  -41.39%  (p=0.000 n=10+9)
Norm2Large-8    205µs ± 2%   122µs ± 2%  -40.27%  (p=0.000 n=10+10)
Norm2Huge-8    21.1ms ± 2%  13.0ms ± 3%  -38.74%  (p=0.000 n=10+10)
```
</details>

### DIFF
--- a/internal/asm/f64/l2norm_amd64.s
+++ b/internal/asm/f64/l2norm_amd64.s
@@ -1,0 +1,109 @@
+// Copyright ©2019 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// +build !noasm,!appengine,!safe
+
+#include "textflag.h"
+
+#define SUMSQ X0
+#define ABSX X1
+#define SCALE X2
+#define ZERO X3
+#define TMP X4
+#define ABSMASK X5
+#define INF X7
+#define INFMASK X11
+#define NANMASK X12
+#define IDX AX
+#define LEN SI
+#define X_ DI
+
+#define ABSMASK_DATA l2nrodata<>+0(SB)
+#define INF_DATA l2nrodata<>+8(SB)
+#define NAN_DATA l2nrodata<>+16(SB)
+// AbsMask
+DATA l2nrodata<>+0(SB)/8, $0x7FFFFFFFFFFFFFFF
+// Inf
+DATA l2nrodata<>+8(SB)/8, $0x7FF0000000000000
+// NaN
+DATA l2nrodata<>+16(SB)/8, $0xFFF8000000000000
+GLOBL l2nrodata<>+0(SB), RODATA, $24
+
+// L2NormUnitary returns the L2-norm of x.
+// func L2NormUnitary(x []float64) (sum float64)
+TEXT ·L2NormUnitary(SB), NOSPLIT, $0
+	MOVQ x_len+8(FP), LEN
+	MOVQ x_base+0(FP), X_
+	PXOR ZERO, ZERO
+	CMPQ LEN, $0          // if len== 0 {return 0}
+	JZ   retZero
+
+	PXOR  INFMASK, INFMASK
+	PXOR  NANMASK, NANMASK
+	MOVSD $1.0, SUMSQ                // ssq = 1
+	XORPS SCALE, SCALE
+	MOVSD l2nrodata<>+0(SB), ABSMASK
+	MOVSD l2nrodata<>+8(SB), INF
+	XORQ  IDX, IDX                   // idx == 0
+
+initZero:  // for ;x[i]==0; i++ {}
+	// Skip all leading zeros, to avoid divide by zero NaN
+	MOVSD   (X_)(IDX*8), ABSX // absxi = x[i]
+	UCOMISD ABSX, ZERO
+	JP      retNaN            // if isNaN(x[i]) { return NaN }
+	JNE     loop              // if x[i] != 0 { goto loop }
+	INCQ    IDX
+	CMPQ    IDX, LEN
+	JE      retZero           // if ++i == len(x) { return 0 }
+	JMP     initZero
+
+loop:
+	MOVSD   (X_)(IDX*8), ABSX // absxi = x[i]
+	MOVUPS  ABSX, TMP
+	CMPSD   ABSX, TMP, $3
+	ORPD    TMP, NANMASK      // NANMASK = NANMASK | IsNaN(absxi)
+	MOVSD   INF, TMP
+	ANDPD   ABSMASK, ABSX     // absxi == Abs(absxi)
+	CMPSD   ABSX, TMP, $0
+	ORPD    TMP, INFMASK      // INFMASK =  INFMASK | IsInf(absxi)
+	UCOMISD SCALE, ABSX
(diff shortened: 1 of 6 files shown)


internal/asm/f64/l2norm_amd64.s
```diff
@@ -0,0 +1,108 @@
...
+
+#define ABSMASK_DATA l2nrodata<>+0(SB)
+#define INF_DATA l2nrodata<>+8(SB)
+#define NAN_DATA l2nrodata<>+16(SB)
+// AbsMask
+DATA l2nrodata<>+0(SB)/8, $0x7FFFFFFFFFFFFFFF
+// Inf
+DATA l2nrodata<>+8(SB)/8, $0x7FF0000000000000
+// NaN
+DATA l2nrodata<>+16(SB)/8, $0xFFF8000000000000
```
kortschak: Can you add checks that other NaN values also work correctly? We provide `floats.NaNWith(payload uint64)`, so there is possibility that we will see those.

internal/asm/f64/stubs_amd64.go
```diff
@@ -170,3 +170,29 @@ func ScalIncTo(dst []float64, incDst uintptr, alpha float64, x []float64, n, inc
 //      sum += x[i]
 //  }
 func Sum(x []float64) float64
+
+// L2NormUnitary is
```
kortschak: Have this first line be `// L2NormUnitary is the level 2 norm of x:`. This is the first function we have here where the Go is non-trivial, so I think it needs to be spelled out. Can you also reflect this in the doc comment for the pure-Go version.

internal/asm/f64/l2norm_noasm.go
```diff
@@ -0,0 +1,36 @@
...
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// +build !amd64 noasm appengine safe
+
+package f64
+
+import "math"
+
+// L2NormUnitary is the level 2 norm of x.
```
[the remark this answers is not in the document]
kortschak: Please also make the doc comment match the comment in the asm stub.

internal/asm/f64/l2norm_amd64.s
```diff
@@ -0,0 +1,108 @@
...
+	ADDSD  $1.0, SUMSQ  // sumsq += 1
+	MOVUPS ABSX, SCALE  // scale = absxi
+	INCQ   IDX          // ++i
+	CMPQ   IDX, LEN
+	JNE    loop         // if i < len(x) { continue }
+
+retSum:  // Calculate return value
+	SQRTSD  SUMSQ, SUMSQ    // sumsq = sqrt(sumsq)
+	MULSD   SCALE, SUMSQ    // sumsq += scale
+	MOVQ    SUMSQ, R10      // tmp = sumsq
```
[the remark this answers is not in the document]
Kunde21: `TMP` is an `XMM` register, which cannot be operated upon by a `CMOV` instruction.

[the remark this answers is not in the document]
vladimir-ch: (⊙_☉)

internal/asm/f64/l2norm_test.go
```diff
@@ -4,7 +4,22 @@
 
 package f64
 
-import "testing"
+import (
+	"fmt"
+	"math"
+	"math/big"
+	"math/rand"
```
kortschak: I don't think these should be here.

Kunde21: Cleaned out

internal/asm/f64/l2norm_amd64.s
Kunde21: Added a test case with inf and nan w/payload

internal/asm/f64/l2norm_noasm.go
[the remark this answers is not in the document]
kortschak: This isn't done, but we can do it later.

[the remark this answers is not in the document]
Kunde21: I'll update the inc/dist doc strings in a separate PR

internal/asm/f64/l2norm_amd64.s
```diff
@@ -0,0 +1,109 @@
...
+	MOVQ x_base+0(FP), X_
+	PXOR ZERO, ZERO
+	CMPQ LEN, $0          // if len== 0 {return 0}
+	JZ   retZero
+
+	PXOR  INFMASK, INFMASK
+	PXOR  NANMASK, NANMASK
+	MOVSD $1.0, SUMSQ                // ssq = 1
+	XORPS SCALE, SCALE
+	MOVSD l2nrodata<>+0(SB), ABSMASK
```
vladimir-ch: I think these two lines could use the `ABSMASK_DATA` and `INF_DATA` defines but it can wait.

PR #1163, opened 2019-11-07
testlapack: clean up lint

kortschak:
Please take a look.


lapack/testlapack/dlahr2.go
```diff
@@ -203,7 +203,10 @@ func Dlahr2Test(t *testing.T, impl Dlahr2er) {
 	defer r.Close()
 
 	var tests []Dlahr2test
-	json.NewDecoder(r).Decode(&tests)
+	err = json.NewDecoder(r).Decode(&tests)
```
vladimir-ch: Eventually I'd like to get rid of these json-based tests.

kortschak: Yes. That would be good.

PR #1167, opened 2019-11-14
all: fix capitalisation of Gonum

kortschak:
This fixes the capitalisation of Gonum where it refers to the project rather
than the GitHub organisation or repository. The text of CONTRIBUTORS also
is fixed to reflect the reality that contributors may have contributed to
other repositories within the project.

Please take a look.


mathext/internal/amos/amoslib/fortran.go
```diff
@@ -3,7 +3,7 @@
 // license that can be found in the LICENSE file.
 
 // +build fortran
-// TODO(jonlawlor): remove fortran build tag when gonum only supports go 1.7+.
+// TODO(jonlawlor): remove fortran build tag when Gonum only supports go 1.7+.
```
kortschak: Leaving the build tag here since we don't ask travis to install gfortran. If we want to do this TODO we will need to do that. I'm not convinced it's worthwhile.

PR #1176, opened 2019-11-20
mat: add MulVecTo method to BandDense and SymBandDense

vladimir-ch:
This is adding the `MulVecTo` method to two matrix types which I think that it makes most sense to support natively in `linsolve`. However, seeing the function signature in the context of `mat` it doesn't seem to fit because of the slices, so I'm wondering whether it should take `*mat.VecDense` instead. This would of course require revisiting `linsolve` and changing the implementation to be more mat-like.

Just a draft at the moment.

Please take a look.

### DIFF
--- a/mat/band.go
+++ b/mat/band.go
@@ -5,6 +5,7 @@
 package mat
 
 import (
+	"gonum.org/v1/gonum/blas"
 	"gonum.org/v1/gonum/blas/blas64"
 )
 
@@ -296,4 +297,39 @@ func (b *BandDense) Trace() float64 {
 		tr += rb.Data[rb.KL+i*rb.Stride]
 	}
 	return tr
+}
+
+// MulVecTo computes B⋅x or Bᵀ⋅x storing the result into dst.
+func (b *BandDense) MulVecTo(dst *VecDense, trans bool, x Vector) {
+	m, n := b.Dims()
+	if trans {
+		m, n = n, m
+	}
+	if x.Len() != n {
+		panic(ErrShape)
+	}
+	dst.reuseAsNonZeroed(m)
+
+	t := blas.NoTrans
+	if trans {
+		t = blas.Trans
+	}
+
+	xMat, _ := untransposeExtract(x)
+	if xVec, ok := xMat.(*VecDense); ok {
+		if dst != xVec {
+			dst.checkOverlap(xVec.mat)
+			blas64.Gbmv(t, 1, b.mat, xVec.mat, 0, dst.mat)
+		} else {
+			xCopy := getWorkspaceVec(n, false)
+			xCopy.CloneVec(xVec)
+			blas64.Gbmv(t, 1, b.mat, xCopy.mat, 0, dst.mat)
+			putWorkspaceVec(xCopy)
+		}
+	} else {
+		xCopy := getWorkspaceVec(n, false)
+		xCopy.CloneVec(x)
+		blas64.Gbmv(t, 1, b.mat, xCopy.mat, 0, dst.mat)
+		putWorkspaceVec(xCopy)
+	}
 }
--- a/mat/matrix_test.go
+++ b/mat/matrix_test.go
@@ -10,6 +10,7 @@ import (
 	"reflect"
 	"testing"
 
+	"golang.org/x/exp/rand"
 	"gonum.org/v1/gonum/blas"
 	"gonum.org/v1/gonum/blas/blas64"
 	"gonum.org/v1/gonum/floats"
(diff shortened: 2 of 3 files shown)


kortschak:
Yes, I think a `*mat.VecDense` would be better. Also it seems that floating point differences between native Go and the assembly implementations impact here.

vladimir-ch:
> Yes, I think a *mat.VecDense would be better.

That will have implications in the linsolve package. I'll take look what needs to be changed and the merging into this repo will have to be postponed.

> Also it seems that floating point differences between native Go and the assembly implementations impact here.

Fixed.

kortschak:
>That will have implications in the linsolve package. I'll take look what needs to be changed and the merging into this repo will have to be postponed.

Yes. I imagine this will be reasonably invasive.

btracey:
I think it should take in a `mat.Vector` and go to a `*mat.VecDense`.

vladimir-ch:
Makes sense. I'll change it.

vladimir-ch:
I've changed it according to your suggestion, @btracey. If it looks good, I'll remove the draft status.

mat/matrix_test.go
```diff
@@ -693,3 +707,87 @@ func TestDoer(t *testing.T) {
...
+
+		r, c := a.Dims()
+
+		for _, trans := range []bool{false, true} {
+			m, n := r, c
+			if trans {
+				m, n = c, r
+			}
+
+			for xType := 0; xType < 4; xType++ {
```
btracey: I wonder if you could use one of the functions in list_test.go instead of this (testOneInputFunc may be the most appropriate?). testTwoInput assumes that the receiver is being modified, so I think you have to use testTwoInput, and then loop over the different receiver cases. Let me know if you have questions, as I know it's a bit of a doozy, but there are actually more like 6 different cases to test

mat/band.go
```diff
@@ -297,3 +298,36 @@ func (b *BandDense) Trace() float64 {
...
+	}
+	if x.Len() != n {
+		panic(ErrShape)
+	}
+	dst.reuseAsNonZeroed(m)
+	t := blas.NoTrans
+	if trans {
+		t = blas.Trans
+	}
+	switch x := x.(type) {
```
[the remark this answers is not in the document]
vladimir-ch: Should `VecDense` have `SetRawVector`? And do we need `untransposeVector(Vector) (Vector,bool)`?

And from looking at existing function for vectors we don't care about the proper vector orientation, right? For example matrix times row vector is ok, only the length matters.

mat/matrix_test.go
vladimir-ch: Yes, I planned to tackle the test after finishing `MulVecTo`.

Btw, "doozy", nice word, never seen it before.

vladimir-ch: I've tried `testTwoInput` but for me it's simpler and faster to write the test manually.

vladimir-ch:
I've removed the Draft status, PTAL.

mat/band.go
```diff
@@ -297,3 +298,36 @@ func (b *BandDense) Trace() float64 {
...
+		} else {
+			var xCopy VecDense
+			xCopy.CloneVec(xVec)
+			blas64.Gbmv(t, 1, b.mat, xCopy.mat, 0, dst.mat)
+		}
+	} else {
+		var xCopy VecDense
+		xCopy.CloneVec(x)
+		blas64.Gbmv(t, 1, b.mat, xCopy.mat, 0, dst.mat)
+	}
```
kortschak: i'm wondering if these would benefit from something like this.
```
		} else {
			xCopy := getWorkspaceVec(n, false)
			xCopy.CloneVec(xVec)
			blas64.Gbmv(t, 1, b.mat, xCopy.mat, 0, dst.mat)
			putWorkspaceVec(xCopy)
		}
	} else {
		xCopy := getWorkspaceVec(n, false)
		xCopy.CloneVec(x)
		blas64.Gbmv(t, 1, b.mat, xCopy.mat, 0, dst.mat)
		putWorkspaceVec(xCopy)
	}
```

kortschak:
The failure on arm64 goes away if tol is set to 1e-14.

mat/band.go
vladimir-ch: Done. AFAIU, the more the workspace is used, the more useful it is?

vladimir-ch:
With this arm64 thing we've made a whip for our own back (ok, we did find a couple of bugs thanks to it but still). I've changed the tolerance to 1e-14.

mat/band.go
kortschak: Yes. A benchmark might be worthwhile, but that can happen later.

kortschak:
> With this arm64 thing we've made a whip for our own back

Indeed.

axkirillov:
Is it possible to do the same for Dense?

kortschak:
@axkirillov Can you file a feature request please?

PR #1178, opened 2019-11-21
stat/card: fix rho functions

kortschak:
Of course there was an error.

Please take a look.

### DIFF
--- a/stat/card/card_test.go
+++ b/stat/card/card_test.go
@@ -45,19 +45,19 @@ var counterTests = []struct {
 }{
 	{name: "exact-1e5", count: 1e5, counter: func() counter { return make(exact) }, tol: 0},
 
-	{name: "HyperLogLog32-0-10-FNV-1a", count: 0, counter: func() counter { return mustCounter(NewHyperLogLog32(10, fnv.New32a())) }, tol: 0.02},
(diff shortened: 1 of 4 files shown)


stat/card/card_test.go
```diff
@@ -135,7 +135,7 @@ func TestUnion(t *testing.T) {
 			t.Errorf("unexpected error from Union call: %v", err)
 		}
 		if got := u.Count(); !floats.EqualWithinRel(got, 2*test.count, 2*test.tol) {
```
sbinet:
```suggestion
		if got, want := u.Count(), 2*test.count; !floats.EqualWithinRel(got, want, 2*test.tol) {
```

stat/card/card_test.go
```diff
@@ -135,7 +135,7 @@ func TestUnion(t *testing.T) {
...
-			t.Errorf("unexpected count for %s: got:%.0f want:%.0f", test.name, got, test.count)
+			t.Errorf("unexpected count for %s: got:%.0f want:%.0f", test.name, got, 2*test.count)
```
sbinet:
```suggestion
			t.Errorf("unexpected count for %s: got:%.0f want:%.0f", test.name, got, want)
```

stat/card/card_test.go
kortschak: I don't think these are necessary.

kortschak:
It pops out when the more complex estimations are used.

Issue #1177, opened 2019-11-21
stat/card: extend implementation for serialisation, and ml and co-cardinality estimates


kortschak:
The current HLL implementation is very simple and is there to get the package started. To get to feature completeness we should have

- [ ] Sketch serialisation (simple, though hash serialisation is a consideration)
- [ ] Maximum likelihood cardinality estimation (see https://oertl.github.io/hyperloglog-sketch-estimation-paper/)
- [ ] Maximum likelihood co-cardinality estimation (ibid - and significantly harder)

PR #1179, opened 2019-11-23
stat/card: add sketch serialisation

kortschak:
Please take a look.

Updates #1177.

### DIFF
--- a/stat/card/card_test.go
+++ b/stat/card/card_test.go
@@ -5,6 +5,7 @@
 package card
 
 import (
+	"encoding"
 	"fmt"
 	"hash/fnv"
 	"io"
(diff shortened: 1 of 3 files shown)


stat/card/hll32.go
```diff
@@ -142,3 +146,61 @@ func (h *HyperLogLog32) Reset() {
...
+	}
+	t := reflect.TypeOf(h.hash)
+	var prefix string
+	if t.Kind() == reflect.Ptr {
+		t = t.Elem()
+		prefix = "*"
+	}
+	var buf bytes.Buffer
+	enc := gob.NewEncoder(&buf)
+	enc.Encode(prefix + path.Join(t.PkgPath(), t.Name()))
```
golangcibot: Error return value of `enc.Encode` is not checked (from `errcheck`)

stat/card/hll32.go
```diff
@@ -142,3 +146,61 @@ func (h *HyperLogLog32) Reset() {
...
+	enc.Encode(h.p)
```
golangcibot: Error return value of `enc.Encode` is not checked (from `errcheck`)

stat/card/hll32.go
```diff
@@ -142,3 +146,61 @@ func (h *HyperLogLog32) Reset() {
...
+	enc.Encode(h.register)
```
golangcibot: Error return value of `enc.Encode` is not checked (from `errcheck`)

stat/card/hll32.go
```diff
@@ -142,3 +146,82 @@ func (h *HyperLogLog32) Reset() {
...
+		return nil, errors.New("card: hash function not set")
+	}
+	t := reflect.TypeOf(h.hash)
+	var prefix string
+	if t.Kind() == reflect.Ptr {
+		t = t.Elem()
+		prefix = "*"
+	}
+	var buf bytes.Buffer
+	enc := gob.NewEncoder(&buf)
```
btracey: Given that this is basically gob, why not call this GobEncoder? I would have thought that a BinaryEncoder would maybe store the states directly to make a more easily readable format.

kortschak: Both are picked up by gob, but other binary encoding paths don't pick up `gob.Gob{Encoder,Decoder}`. If we only provide one serialisation hook, then it makes sense to have it work with all binary encoding paths, `encoding.Binary{Marshaler,Unmarshaler}` gets us this.

kortschak:
*ping*

stat/card/card_test.go
```diff
@@ -195,6 +196,96 @@ func TestResetCounters(t *testing.T) {
...
+	{
+		name: "HyperLogLog64-8-4-FNV-1a", count: 1e3,
+		src: func() counterEncoder { return mustCounterEncoder(NewHyperLogLog64(8, fnv.New64a())) },
+		dst: func() counterEncoder { return mustCounterEncoder(NewHyperLogLog64(4, fnv.New64a())) },
+	},
+}
+
+func mustCounterEncoder(c counterEncoder, err error) counterEncoder {
+	if err != nil {
+		panic(fmt.Sprintf("bad test: %v", err))
```
oleg-balunenko: Suggest not using panics.
You can use https://godoc.org/testing#T.Fatal

kortschak: Not possible here; these calls happen in package-level vars where there is no `*testing.T` available. The errors returned by the constructors are solely dependent on the precision parameter and are essentially a compile time error, so panics are fine here.

stat/card/hll32.go
btracey: Okay. I'm not sure that this is right, but I'll admit that I can't think through the long-term ramifications of the decision.

kortschak: The only issue that I can see is that it gets in the way of foreign language interoperability. I don't think this is a huge issue, particularly since the implementation is diverging from what people normally do anyway.

PR #1180, opened 2019-11-24
prng: non-stdlib PRNG implementations

kortschak:
Please take a look.

The intention here is to provide PRNGs that we will see in C/C++ and other language implementations.


kortschak:
I don't know. Putting it in mathext hides it away, but it does seem to fit there.

I want to add some docs to this explaining the various generators/their uses. I'll think about this while I do that. I also want to wait for #1181 to be fixed and rebase onto the master after that.

kortschak:
The other thing that I think is worth considering is adding support for serialising PRNG state. This would have significant advantages for long running simulations, allowing check pointing and continuation. If we do that, we need to duplicate the PCGSource from exp/rand, so that it can also be serialised; we can automate that duplication/extension.

kortschak:
I have added serialisation, and this has also been added to `x/exp/rand.PCGSource`, so we don't need to do anything here for that.

PTAL

kortschak:
I'm confident on the basis of the tests passing the series produced by the original C code in all cases.

PR #1182, opened 2019-11-25
lapack/testlapack: remove JSON-based part of Dlahr2Test

vladimir-ch:
Please take a look.


btracey:
Sorry, I forget, do we have an alternate test or are we just removing this?

vladimir-ch:
IIRC, Dlahr2 was my second lapack routine that I ported to Gonum and it had initially only the JSON-based tests with data generated by the Reference implementation. Later I added testing that follows our usual styles which should test Dlahr2 completely. This PR removes just the JSON part.

PR #1186, opened 2019-11-26
mat: add VecDense and RawVectorer cases to untransposeExtract

vladimir-ch:
Please take a look.

### DIFF
--- a/mat/vector.go
+++ b/mat/vector.go
@@ -219,6 +219,10 @@ func (v *VecDense) RawVector() blas64.Vector {
 	return v.mat
 }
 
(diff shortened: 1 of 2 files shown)


mat/vector.go
```diff
@@ -219,6 +219,10 @@ func (v *VecDense) RawVector() blas64.Vector {
 	return v.mat
 }
 
+func (v *VecDense) SetRawVector(a blas64.Vector) {
```
kortschak: Doc comment? (also add for the already existing `RawVector` method).

vladimir-ch: Done.

PR #1188, opened 2019-11-29
graph/encoding/dot: add example for decoding a weighted graph

kortschak:
This adds the example (cleaned up) I recently posted to gonum-dev.

Please take a look.


kortschak:
I'm thinking of adding extra parts that show how the edge shims can be dropped on accessing them from the deserialised graph, but I'll do that later.

PR #1191, opened 2019-11-30
lapack/internal: remove Fortran code for generating Dsterf test case

vladimir-ch:
Please take a look.


lapack/testlapack/dsterf.go
```diff
@@ -20,15 +20,14 @@ type Dsterfer interface {
...
-	// Hand coded tests.
+	// Tests with precomputed eigenvalues.
 	for cas, test := range []struct {
 		d []float64
 		e []float64
 		n int
 
 		ans []float64
 	}{
-		// Computed from Fortran code.
```
kortschak: I'd leave this, or change it to "Computed from original Fortran code."

vladimir-ch: Ok, done. My intention was to remove it to not give a wrong impression that Fortran is necessary to get the expected result. The same can be obtained for example by Octave:
```
a = [1, 2, 0, 0; 2, 3, 4, 0; 0, 4, 4, 5; 0, 0, 5, 6]

a =

   1   2   0   0
   2   3   4   0
   0   4   4   5
   0   0   5   6

eig(a)

ans =

   -2.54638
    0.70423
    4.79592
   11.04623
```

kortschak: I suspect the octave code is invoking fortran, but yes. The point here is to let people know approximately the provenance.

PR #1193, opened 2019-12-02
all: remove redundant bad prefix and make panic strings const

kortschak:
This was discussed in a PR in stat a fair while back.

Please take a look.

### DIFF
--- a/mat/eigen.go
+++ b/mat/eigen.go
@@ -11,7 +11,7 @@ import (
 
 const (
 	badFact   = "mat: use without successful factorization"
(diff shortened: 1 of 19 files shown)


kortschak:
The basis for the difference between `badThing` and `errSomeCondition` is pretty much just that; `badThing` means the thing (value/parameter) that is being panicked over is bad, while `errSomeCondition` is the condition describing state is an error (e.g. `errParamFunctionSingularity` - `badParamFunctionSingularity` would not make sense since the singularity is an error, it's not a bad singularity).

I hope that clarifies the situation.

PR #1199, opened 2019-12-15
lapack/testlapack: rework Dlasq2 test

vladimir-ch:
... so that it doesn't use precomputed data.

As for `Dlasq3`, `Dlasq4` and `Dlasq5`, I'd be fine with removing their tests altogether because:
* without using data precomputed by the reference implementation they are untestable. These routines are not standalone but rather a tightly coupled set of workers called from a "high" level driver `Dlasq2`.
* the current tests for `Dlasq1` and `Dlasq2` cover them almost completely, except for some error and degenerate cases.
`Dlasq1` gives coverage Dlasq3 83.3%, Dlasq4 88.2%, Dlasq5 92.5%
`Dlasq2` gives coverage Dlasq3 85.4%, Dlasq4 90.7%, Dlasq5 92.5%
This means that the trouble of using precomputed data is not really justified.

Let me know what you think. If we removed the tests, we could then remove also `internal/testdata/dlasqtest/`.

Note that there is currently no test for `Dlasq6` and also the current tests don't cover it (it's called only in an underflow path of `Dlasq3`).

Please take a look.


lapack/testlapack/dlasq2.go
```diff
@@ -5,711 +5,118 @@
...
-			n:    21,
-			zOut: []float64{0.2309990065237698e+01, 0.2261037149894265e+01, 0.1946075797700671e+01, 0.1868880050926239e+01, 0.1838164096767709e+01, 0.1792848619787804e+01, 0.1601503570675395e+01, 0.1526383850720226e+01, 0.1395433670210245e+01, 0.1010109387319946e+01, 0.8644978203099237e+00, 0.7632668975571170e+00, 0.4552388959212355e+00, 0.4364144823342648e+00, 0.2696980874891977e+00, 0.1302983706884531e+00, 0.1278582769210706e+00, 0.7956979307866878e-01, 0.7389266586143965e-01, 0.8102308371415692e-02, 0.2690335013870282e-02, 0.5752609692135112e-13, 0.2842400234743069e-06, 0.1045980487121886e-05, 0.1601503570675395e+01, 0.1668093198756260e-18, 0.5190756964721459e-05, 0.2367180586416983e-05, 0.1526383850720226e+01, 0.1262177448353619e-27, 0.1974011051185169e-03, 0.4728862593194215e-03, 0.1395433670210245e+01, 0.2337388813815525e-19, 0.1632969150852556e-06, 0.4026698297028720e-06, 0.1010109387319946e+01, 0.3802651360818291e-21, 0.1380334146666910e-03, 0.1568641113059460e-03, 0.8644978203099237e+00, 0.2788260408793746e-19, 0.2076195419278686e+02, 0.2076195419278686e+02, 0.1070000000000000e+03, 0.3183673469387755e+01, 0.2803738317757009e+01, 0.12735788373933
```
vladimir-ch: It'd be better to start with the desired eigenvalues and generate the slice `z` from them but I don't immediately see how to do that (going from the eigenvalues to a tridiagonal symmetric matrix is work but straightforward, it's the next steps of going to T and then L*U=T that are not so clear). I'd say that for now we have what we have.

PR #1207, opened 2019-12-20
lapack/testlapack: don't test unsupported matrix sizes in DlantrTest

vladimir-ch:
The current Dlantr failures in our netlib repository are caused by out-of-bounds write in the Fortran reference LAPACK when computing the inf-norm of wide lower-triangular matrices with "tight" `work`. I will submit a PR to the reference and OpenBLAS to lift this restriction (which is documented but unnecessary in my opinion and to top it off DLANTR doesn't check the input parameters) but it cannot be relied upon anyway.

Please take a look.

### DIFF
--- a/lapack/testlapack/dlantr.go
+++ b/lapack/testlapack/dlantr.go
@@ -26,6 +26,12 @@ func DlantrTest(t *testing.T, impl Dlantrer) {
 	for _, m := range []int{0, 1, 2, 3, 4, 5, 10} {
 		for _, n := range []int{0, 1, 2, 3, 4, 5, 10} {
 			for _, uplo := range []blas.Uplo{blas.Lower, blas.Upper} {
+				if uplo == blas.Upper && m > n {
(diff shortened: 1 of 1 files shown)


vladimir-ch:
Once this is merged I think we'll need 0.6.2 to use it in the `netlib` repo, won't we?

kortschak:
> Once this is merged I think we'll need 0.6.2 to use it in the `netlib` repo, won't we?

Yes. We can make a v0.6.2 milestone now to add this to.

kortschak:
Actually, could you add some text to the commit message explaining the issue.

vladimir-ch:
I've added a paragraph explaining the issue and reason for modifying the test. Let me know if the explanation is not clear or not well written.

kortschak:
Thanks

kortschak:
v0.6.2 is released, so the changes can now be made to netlib.

Issue #916, opened 2019-03-23
unit/constant: Update to new NIST constant values


btracey:
Not a bug currently, but will be a bug in a few months (very strange).

There was a revision to SI, which will update values of certain fundamental constants. When NIST updates their reference values, we need to update the constant values

PR #1208, opened 2019-12-23
unit/constant: update the values and uncertainties for 2018 CODATA

kortschak:
Please take a look.

Complete table of values at https://physics.nist.gov/cuu/Constants/Table/allascii.txt.

Updates #916.


kortschak:
That's the joy of having a unit system that doesn't include things like "rods per hogshead" and "libraries of congress".

btracey:
To be fair, we could pull an avogadro and precisely define the number of books required to make one Library of Congress

kortschak:
Want to file an issue?

PR #1209, opened 2019-12-24
mat: add Kronecker product function

kortschak:
Please take a look.

### DIFF
--- a/mat/dense_arithmetic.go
+++ b/mat/dense_arithmetic.go
@@ -673,6 +673,20 @@ func (m *Dense) Pow(a Matrix, n int) {
 	putWorkspace(x)
 }
 
+// Kronecker calculates the Kronecker product of a and b, placing the result in
+// the receiver.
+func (m *Dense) Kronecker(a, b Matrix) {
+	ra, ca := a.Dims()
+	rb, cb := b.Dims()
+
+	m.reuseAsNonZeroed(ra*rb, ca*cb)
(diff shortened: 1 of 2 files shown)


kortschak:
Specialised cases to come in other PRs as needed.

- generalised x diagonal and vice versa
- generalised x band and vice versa
- generalised x triangular and vice versa

The combinatoric explosion is irritating and so not dealt with here.

mat/dense_arithmetic.go
```diff
@@ -673,6 +673,20 @@ func (m *Dense) Pow(a Matrix, n int) {
...
+// Kronecker calculates the Kronecker product of a and b, placing the result in
+// the receiver.
+func (m *Dense) Kronecker(a, b Matrix) {
+	ra, ca := a.Dims()
+	rb, cb := b.Dims()
+
+	m.reuseAsNonZeroed(ra*rb, ca*cb)
+	for i := 0; i < ra; i++ {
+		for j := 0; j < ca; j++ {
+			m.Slice(i*rb, (i+1)*rb, j*cb, (j+1)*cb).(*Dense).Scale(a.At(i, j), b)
```
sbinet: it'd be great if we had a `Slice` version that would return the underlying `*Dense` matrix directly (possibly one only available from inside `mat`, of course).

kortschak: #1211

Issue #1200, opened 2019-12-16
Add arm64 hardware acceleration


blackrez:
### Background

On onnx-go, we have poor performance on running some models on arm64. After profiling, we identify that we spend a lot of time on gonum.
More informations on this issue :
https://github.com/owulveryck/onnx-go/issues/169

### Proposal

Add arm64 asm instructions.

Kunde21:
@blackrez I haven't started studying arm64 asm and I still have yet to finish x64 (s|d)gemm kernel.  This is on my roadmap and we're looking for help on it, but it may not be right away.

blackrez:
I'm volunteer to help on this issue. I'm not an expert on arm64 but I want to learn. If you have some resources to point me I will be glad to help

Kunde21:
I always start with the manuals, so [arm intro](https://developer.arm.com/common-tasks/writing-arm-assembly-code/learning-about-the-instruction-set) and the [ISA manual](https://developer.arm.com/docs/ddi0596/latest) are the best references.  Reading through the kernels in OpenBLAS can help, though they usually have a lot of C preprocessor to them.

PR #1215, opened 2020-01-09
math32: Add basic support for ARM64

blackrez:
I add a basic instructions for ARM64, a little step for the big issue #1200.
I also remove TODO in AMD64.


internal/math32/sqrt.go
```diff
@@ -17,7 +17,9 @@ import (
 //	Sqrt(±0) = ±0
 //	Sqrt(x < 0) = NaN
 //	Sqrt(NaN) = NaN
-func Sqrt(x float32) float32 {
```
Kunde21: This is the version that is used when asm isn't available.  
Use build tags to avoid duplicate definitions.  `!amd64 !arm64 noasm appengine safe`

kortschak:
While I think that having `GOARCH=arm64` asm is valuable, I am concerned about the lack of our ability to integrate this work with continuous integration. Are there services that provide ARM64 architecture testing?

internal/math32/sqrt_amd64.s
```diff
@@ -8,8 +8,7 @@
 
 //+build !noasm,!appengine,!safe
 
-// TODO(kortschak): use textflag.h after we drop Go 1.3 support
```
kortschak: Please revert this. It needs to be done, but it should be in a separate PR.

internal/math32/sqrt_arm64.s
```diff
@@ -6,15 +6,14 @@
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
-// +build !noasm,!appengine,!safe
+//+build !noasm,!appengine,!safe
 
-package math32
+#include "textflag.h"
+// Don't insert stack check preamble.
```
kortschak: Please delete this comment.

internal/math32/sqrt_arm64.s
```diff
@@ -6,15 +6,14 @@
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
-// +build !noasm,!appengine,!safe
+//+build !noasm,!appengine,!safe
```
kortschak: Please revert this, the conventional syntax is `// +build <tags>`.

kortschak:
There are: https://blog.travis-ci.com/2019-10-07-multi-cpu-architecture-support.

See #1217.

There is a non-trivial number of failures there, all to do with float precision differences between amd64 and arm64.

kortschak:
When you resolve the conflicts here, please rebase onto master where we are now testing on arm64.

PR #1217, opened 2020-01-10
travis: build with arm64 on linux

kortschak:
Please take a look.

Updates #1200.


kortschak:
I'm not sure what to do with the failures here. There are two possible approaches that immediately come to mine:

1. Relax the stringency of the test tolerances and re-engineer the graph layout failure.
2. Only test internal on arm64.

I'm inclined towards 2., though I think I should probably make the `layout.EadesR2` test safer.

vladimir-ch:
I think that instead of choosing looser tolerances it would be worth to review the failing tests and see if we can improve them. Our tests are actually quite strict, using mostly absolute (and quite tight) tolerances, while for example Lapack in general uses relative tolerance often multiplied by the size of the problem. I'll try to go through the log output and see what can be done.

kortschak:
I have relaxed the FFT failure and worked around the Eades test failure (in a way that I'm not entirely happy with, but can't debug until I have set up some hardware to look into it - I have this, but it needs a setup).

kortschak:
Well, I think the layout failures are not workable around.

kortschak:
Isomap layout golden images are generated on a Librem5.

kortschak:
I'd like to merge this so that the work on arm64 assembly can be continued with reasonably safety. I don't want the failures to get in the way of other development though, so I'm thinking of putting arm64 into allowed failures. If this is OK, I'll do that and file an issue to make arm64 green and move it out of allowed failure.

WDYT?

kortschak:
As a note, qemu seems to do a good job now of the emulation (at least on linux), so debugging arm64 should be doable on other architectures. I have the gnueabi gcc toolchain installed and qemu-user.

vladimir-ch:
> I'm thinking of putting arm64 into allowed failures. If this is OK, I'll do that and file an issue to make arm64 green and move it out of allowed failure.

SGTM

kortschak:
I have moved arm64 into allowed failures. I'm fine with the graph/layout fixes (they are either things that should have been there in the past or approaches we already take) but the some of the other fixes are a bit blunt. Would you like me to remove them, or shall we just go over them again (e.g. the asm within function should probably be a relative error rather than an absolute)?

vladimir-ch:
I'd merge and go over them again later (agreed on `within`).

Issue #1218, opened 2020-01-12
travis: builds on GOARCH=arm64 are broken


kortschak:
Adding the arm64 architecture to the build matrix shows up a number of issues in our testing. These all come down to numerical precision differences between amd64 and arm64 (some show up in arm as well, 386 shows no differences with amd64).

The arm64 arch will be in allowed failures when #1217 is merged which is not ideal, so we need to get the failures resolved so it will show up in the build report on PRs.

This is the list of failure (checked indicates I'm happy with the fix approach - I will add the change sha when #1217 is merged).

- [x] gonum.org/v1/gonum/blas/blas64 `TestDrotmg` "tolerance drotm y_0 mismatch RD1_Med_RD2_Big_Flag_0: expected 0, found 5.1194107541670215e-11" 86e8186781d9a1bb0d5594cda002ef6ba7c5b06e
- [x] gonum.org/v1/gonum/blas/gonum `TestDrotmg` "tolerance drotm y_0 mismatch RD1_Med_RD2_Big_Flag_0: expected 0, found 5.1194107541670215e-11" 86e8186781d9a1bb0d5594cda002ef6ba7c5b06e
- [x] gonum.org/v1/gonum/fourier tolerance `ExampleFFT_Coefficients_tone` 0c0a3fef116df8c4a19bf956d9c1a536b34a08e0
- [x] gonum.org/v1/gonum/graph/layout bug `TestEadesR2` infinite value b5cc9a985978f3b32ea1066f8722cf8a3dc8495e
- [x] gonum.org/v1/gonum/graph/layout tolerance/ED `TestIsomapR2` d8ec179f2bfc07f4e1af26de87896d672c76a332
- [x] gonum.org/v1/gonum/lapack/gonum tolerance `TestDlansb` "got 3.9533942447863906, want 3.9533942447863923" a84dbb2e4f19f947a3a20094acda433ea06ee75a
- [x] gonum.org/v1/gonum/lapack/gonum tolerance `TestDlascl` "unexpected A[0,0]=26.358078839820152, want 26.358078839820152" a84dbb2e4f19f947a3a20094acda433ea06ee75a
- [x] gonum.org/v1/gonum/lapack/gonum tolerance `TestDlaln2` "Want (-0.9316138629711816+0i), got (-0.9316138629728812+0i)" a84dbb2e4f19f947a3a20094acda433ea06ee75a
[...]

PR #1219, opened 2020-01-13
stat: fix failures on arm64

kortschak:
Please take a look (check test log since these are currently allowed failures).

Updates #1218.

### DIFF
--- a/stat/roc.go
+++ b/stat/roc.go
@@ -107,10 +107,10 @@ func ROC(cutoffs, y []float64, classes []bool, weights []float64) (tpr, fpr, thr
 	invNeg := 1 / nNeg
 	invPos := 1 / nPos
 	for i := range tpr {
-		tpr[i] *= invPos
-		tpr[i] = 1 - tpr[i]
-		fpr[i] *= invNeg
-		fpr[i] = 1 - fpr[i]
+		// Prevent fused float operations by
+		// making explicit float64 conversions.
+		tpr[i] = 1 - float64(tpr[i]*invPos)
+		fpr[i] = 1 - float64(fpr[i]*invNeg)
 	}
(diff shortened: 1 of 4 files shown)


kortschak:
push tests:
https://travis-ci.org/gonum/gonum/jobs/636171232#L9034
https://travis-ci.org/gonum/gonum/jobs/636171233#L9034
https://travis-ci.org/gonum/gonum/jobs/636171235#L9040

stat/roc.go
```diff
@@ -107,10 +107,10 @@ func ROC(cutoffs, y []float64, classes []bool, weights []float64) (tpr, fpr, thr
 	invNeg := 1 / nNeg
 	invPos := 1 / nPos
 	for i := range tpr {
-		tpr[i] *= invPos
-		tpr[i] = 1 - tpr[i]
-		fpr[i] *= invNeg
-		fpr[i] = 1 - fpr[i]
+		// Prevent fused float operations by
+		// making explicit float64 conversions.
+		tpr[i] = 1 - float64(tpr[i]*invPos)
```
vladimir-ch: Is there a reason to prevent fused float operations other than making the results consistent between amd64 and arm64?

kortschak: Yeah, this is the thing that troubles me. The reason here is that using the FMA here results in a negative value. We know that ROC values must always be in [0, 1], so this is wrong. The question then is why the additional precision gives us this error and what we should do about it.

1. Either we're lucky with the examples that we have and the zeros are fortuitous or there is an interaction between the FMA and other parts the routine that give a discordance in the precisions and so these errors.
2. We make the code the same as it was and hope the second answer to 1 is correct or we round small negative values.

The approach I've taken seems simpler.

kortschak: *ping*

kortschak: I'll file an issue to look at this again. #1222

PR #1221, opened 2020-01-14
blas/gonum: make drotmg test pass

kortschak:
This change prevents the compiler from emitting FMA instructions during Drotmg and Drotm which would otherwise cause the arm64 build to fail.

Please take a look.

Updates #1218.

### DIFF
--- a/blas/gonum/level1float32.go
+++ b/blas/gonum/level1float32.go
@@ -336,7 +336,7 @@ func (Implementation) Srotmg(d1, d2, x1, y1 float32) (p blas.SrotmParams, rd1, r
 			h22 = 1
 			h21 = -y1 / x1
 			h12 = p2 / p1
-			u := 1 - h12*h21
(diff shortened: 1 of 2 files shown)


kortschak:
Which do you think are unnecessary?

kortschak:
&*ping*

btracey:
Sorry, I didn't mean unnecessary in the sense of "they aren't needed to prevent FMA", but in the sense of "Maybe our tests are too tight if this matters". For instance, where we have 
`u := 1 + float64(h11*h22)`, that seems like the kind of optimization we do want to use FMA for, as it's not about getting some improperly rounded value (in the way `x*x -y*y` can). It's good to have tests pass, but maybe we shouldn't be demanding floating point agreement and letting some of these optimizations happen.

This looks good for now, but just a thought.

kortschak:
Do you want to file an issue about that? It's independent of #1218.

PR #1224, opened 2020-01-26
testlapack: adjust some tolerances

vladimir-ch:
Updates #1218


vladimir-ch:
I don't know how to test locally for arm64, so I may need to adjust the tolerances based on results from Travis.

kortschak:
To run alternative arch code locally, install qemu-user, then build/test the code with `GOARCH=arm64`.

vladimir-ch:
I'll have to revisit the test for Dgeev again, the increase in eigenvector tolerance for Gear matrix for even sizes was not enough.

PR #1229, opened 2020-02-02
testlapack: rework eigenvector checks in DgeevTest and adjust tolerances

vladimir-ch:
Please take a look.

### DIFF
--- a/lapack/testlapack/dgeev.go
+++ b/lapack/testlapack/dgeev.go
@@ -15,7 +15,6 @@ import (
 
 	"gonum.org/v1/gonum/blas"
 	"gonum.org/v1/gonum/blas/blas64"
-	"gonum.org/v1/gonum/floats"
 	"gonum.org/v1/gonum/lapack"
 )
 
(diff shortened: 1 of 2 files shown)


lapack/testlapack/dgeev.go
```diff
@@ -658,80 +643,79 @@ func testDgeev(t *testing.T, impl Dgeever, tc string, test dgeevTest, jobvl lapa
...
-				}
+	// Check that the columns of VL and VR are eigenvectors that:
+	//  - correspond to the computed eigenvalues
+	//  - have Euclidean norm equal to 1
+	//  - have the largest component real
+	bi := blas64.Implementation()
+	if jobvr == lapack.RightEVCompute {
+		resid := residualRightEV(test.a, vr, wr, wi)
+		if resid > vecTol {
+			t.Errorf("%v: unexpected right eigenvectors; residual=%v, want<=%v", prefix, resid, vecTol)
```
kortschak: Currently prefix uses the `%v` verb for the `jobvr` and `jobvl` values. This should probably be `%c` for ease of reading, but that would be another PR since there are a bunch of them and they should all change together.

vladimir-ch: Ok, I'll put it on my todo list.

PR #1232, opened 2020-02-10
lapack/gonum,testlapack: fix bug in Dtrevc3 and rework its test

vladimir-ch:
Please take a look.

### DIFF
--- a/lapack/gonum/dtrevc3.go
+++ b/lapack/gonum/dtrevc3.go
@@ -62,7 +62,7 @@ import (
 //
 // On entry, if howmny is lapack.EVAllMulQ, it is assumed that VL (if side
 // is lapack.EVLeft or lapack.EVBoth) contains an n×n matrix QL,
-// and that VR (if side is lapack.EVLeft or lapack.EVBoth) contains
+// and that VR (if side is lapack.EVRight or lapack.EVBoth) contains
 // an n×n matrix QR. QL and QR are typically the orthogonal matrix Q of Schur
 // vectors returned by Dhseqr.
 //
--- a/lapack/gonum/dtrevc3.go
+++ b/lapack/gonum/dtrevc3.go
@@ -647,7 +647,7 @@ leftev:
 					// when forming the right-hand side.
 					beta := math.Max(norms[j], norms[j+1])
 					if beta > vcrit {
-						bi.Dscal(n-ki+1, 1/vmax, b[ki*ldb+iv:], 1)
+						bi.Dscal(n-ki, 1/vmax, b[ki*ldb+iv:], ldb)
 						vmax = 1
 					}
 					b[j*ldb+iv] -= bi.Ddot(j-ki-1, t[(ki+1)*ldt+j:], ldt, b[(ki+1)*ldb+iv:], ldb)
--- a/lapack/testlapack/dtrexc.go
+++ b/lapack/testlapack/dtrexc.go
@@ -27,7 +27,7 @@ func DtrexcTest(t *testing.T, impl Dtrexcer) {
 		for _, n := range []int{1, 2, 3, 4, 5, 6, 10, 18, 31, 53} {
 			for _, extra := range []int{0, 1, 11} {
 				for cas := 0; cas < 100; cas++ {
-					tmat := randomSchurCanonical(n, n+extra, rnd)
+					tmat, _, _ := randomSchurCanonical(n, n+extra, rnd)
 					ifst := rnd.Intn(n)
 					ilst := rnd.Intn(n)
 					testDtrexc(t, impl, compq, tmat, ifst, ilst, extra, rnd)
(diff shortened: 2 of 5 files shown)


lapack/gonum/dtrevc3.go
```diff
@@ -647,7 +647,7 @@ leftev:
 					// when forming the right-hand side.
 					beta := math.Max(norms[j], norms[j+1])
 					if beta > vcrit {
-						bi.Dscal(n-ki+1, 1/vmax, b[ki*ldb+iv:], 1)
+						bi.Dscal(n-ki, 1/vmax, b[ki*ldb+iv:], ldb)
```
vladimir-ch: This shows the importance of using also bad matrices for testing. This was started to be covered only after I added zero/tiny eigenvalues to `randomSchurCanonical`.

kortschak:
This appears to have flowed on to many `Dtrexc` failures. Is this something that we should be worried about, or should we enable conditional return of zero/tiny eigenvalues from `randomSchurCanonical`?

vladimir-ch:
Sorry, I was running only the Dtrevc3 test locally. I had thought about enabling zero/tiny values conditionally but then decided against it for simplicity. I'll take a look today but at least the "unexpected eigenvalue of 2x2 block" failure looks like an issue with the test.

vladimir-ch:
In the end I've added a flag to `randomSchurCanonical` because checking Dtrexc is another rabbit hole and I need time for exploring it. I've checked, reworked and (hopefully) simplified the test for Dtrexc but for bad matrices the test consistently fails. For good matrices the coverage is only 50% while with bad matrices it's practically complete, so it's still possible there's a bug.

lapack/testlapack/dtrexc.go
```diff
@@ -205,16 +199,126 @@ func testDtrexc(t *testing.T, impl Dtrexcer, compq lapack.UpdateSchurComp, tmat
...
+	qtq := cloneGeneral(tmat1)
+	blas64.Gemm(blas.NoTrans, blas.NoTrans, -1, qt, q, 1, qtq)
+	resid = dlange(lapack.MaxColumnSum, n, n, qtq.Data, qtq.Stride)
+	if resid > float64(n)*tol {
+		t.Errorf("%v: mismatch between Qᵀ*(initial T)*Q and (final T); resid=%v, want<=%v",
+			name, resid, float64(n)*tol)
+	}
+}
+
+func residualOrthogonal(q blas64.General, rowwise bool) float64 {
```
vladimir-ch: Yes, this is yet another way in `testlapack` to check orthogonality of matrices. I'll remove the others and change the test to use this one.

lapack/testlapack/dtrexc.go
```diff
@@ -205,16 +199,126 @@ func testDtrexc(t *testing.T, impl Dtrexcer, compq lapack.UpdateSchurComp, tmat
...
+
+	// Compute
+	//  work = work - Q * Qᵀ = I - Q * Qᵀ
+	// or
+	//  work = work - Qᵀ * Q = I - Qᵀ * Q
+	blas64.Syrk(transq, -1, q, 1, work)
+	return dlansy(lapack.MaxColumnSum, blas.Upper, work.N, work.Data, work.Stride)
+}
+
+func dlansy(norm lapack.MatrixNorm, uplo blas.Uplo, n int, a []float64, lda int) float64 {
```
vladimir-ch: As with `dlange` I'm not sure if it's better to add these methods to the tested interface or use these duplicated but local, independent implementations. I slightly prefer this but you may have a different opinion.

kortschak: I'm fine with this.

lapack/testlapack/dtrexc.go
kortschak: ok

lapack/testlapack/general.go
```diff
@@ -1158,29 +1158,51 @@ func extract2x2Block(t []float64, ldt int) (a, b, c, d float64) {
 // isSchurCanonical returns whether the 2×2 matrix [a b; c d] is in Schur
 // canonical form.
 func isSchurCanonical(a, b, c, d float64) bool {
-	return c == 0 || (a == d && math.Signbit(b) != math.Signbit(c))
```
kortschak: I actually find the original easier to read.

`return c == 0 || (b != 0 && a == d && math.Signbit(b) != math.Signbit(c))`

vladimir-ch: You're right. Done.

lapack/testlapack/general.go
```diff
@@ -131,8 +131,8 @@ func randomHessenberg(n, stride int, rnd *rand.Rand) blas64.General {
 // form, that is, block upper triangular with 1×1 and 2×2 diagonal blocks where
 // each 2×2 diagonal block has its diagonal elements equal and its off-diagonal
 // elements of opposite sign.
-func randomSchurCanonical(n, stride int, rnd *rand.Rand) blas64.General {
-	t := randomGeneral(n, n, stride, rnd)
+func randomSchurCanonical(n, stride int, rnd *rand.Rand) (t blas64.General, wr, wi []float64) {
```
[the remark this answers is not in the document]
vladimir-ch: Here I'm following the lapack namin for these slices/arrays but I don't know why lapack calls them like this.

vladimir-ch:
I tested with netlib and I get only errors for unexpected eigenvalues of 2x2 blocks, not for the matrix Q. This confirms that there will be a bug somewhere in the updates to Q but yesterday I didn't see any mistake in Dtrexc quickly. I'll check again and also the call dependencies.

kortschak:
I tested with netlib earlier (a29e678437e40d1d2c47ed1171b760d192089c5a) and got no errors (and now again with 520dc6057072ee5d1d9e40dec6bfcabb29f13884). When I checkout 04cb64a26ccf96b185ca61788c0f6189abe35abc, and run the netlib test I get many "unexpected {first,second} eigenvalue of 2×2 block".

PR #1233, opened 2020-02-14
lapack/gonum: fix Dlanv2

vladimir-ch:
Please take a look.


kortschak:
Nice work. That's fixed the arm64 `Dgeev` failures. The only remaining failures are in gonum.org/v1/gonum/optimize `BGFS`, `LBFGS` and `Newton`.

Updates  #1218.

vladimir-ch:
That's good. I still have some local changes to Dlaexc test and I'd like to try using `bad=true` in all `randomSchurCanonical`.

PR #1236, opened 2020-02-16
testlapack: rework test for Dlaexc and clean up test for Dtrexc

vladimir-ch:
The tests now use "bad" matrices.

Please take a look.

### DIFF
--- a/lapack/testlapack/dlaexc.go
+++ b/lapack/testlapack/dlaexc.go
@@ -14,6 +14,7 @@ import (
 
 	"gonum.org/v1/gonum/blas"
 	"gonum.org/v1/gonum/blas/blas64"
+	"gonum.org/v1/gonum/lapack"
 )
 
 type Dlaexcer interface {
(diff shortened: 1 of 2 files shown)


lapack/testlapack/dlaexc.go
```diff
@@ -129,77 +149,21 @@ func testDlaexc(t *testing.T, impl Dlaexcer, wantq bool, n, j1, n1, n2, extra in
 			}
 			diff := tmat.Data[i*tmat.Stride+j] - tmatCopy.Data[i*tmatCopy.Stride+j]
 			if diff != 0 {
-				t.Errorf("%v: unexpected modification of T[%v,%v]", prefix, i, j)
+				t.Errorf("%v: unexpected modification of T[%v,%v]", name, i, j)
 			}
 		}
 	}
 
-	if n1 == 1 {
```
vladimir-ch: The alternative to deleting these blocks would be to increase tolerance to `1e-7` or to have a separate tolerance only for these eigenvalue checks. Since I already check that T is Schur canonical, that it is similar to the original T and that it is not modified outside of the row/column band, these checks are in principle redundant and removing them also simplifies the test.

lapack/testlapack/dtrexc.go
```diff
@@ -152,46 +157,19 @@ func dtrexcTest(t *testing.T, impl Dtrexcer, rnd *rand.Rand, n, ifst, ilst, extr
...
 			}
-			diff := tmat1.Data[i*tmat1.Stride+j] - tmatCopy.Data[i*tmatCopy.Stride+j]
+			diff := tmat.Data[i*tmat.Stride+j] - tmatCopy.Data[i*tmatCopy.Stride+j]
 			if diff != 0 {
 				t.Errorf("%v: unexpected modification at T[%v,%v]", name, i, j)
 			}
 		}
 	}
 
-	// Check that the block at ifstGot was delivered to ilstGot correctly.
```
vladimir-ch: Same here.

Issue #1235, opened 2020-02-15
optimize: add documentation to Status constants and Location fields


kortschak:
These are reasonably clear in their intention, but documentation would help.

The `Status` constants should probably have comments explaining under what conditions they are returned and how they should be interpreted in the context of the API.

The `Location` fields should explain what they are.

PR #1237, opened 2020-02-16
optimize: add doc comments for Location fields

kortschak:
Please take a look.

Updates #1235.

### DIFF
--- a/optimize/minimize.go
+++ b/optimize/minimize.go
@@ -35,10 +35,14 @@ type Task struct {
 
 // Location represents a location in the optimization procedure.
 type Location struct {
-	X        []float64
-	F        float64
+	// X is the function input for the location.
+	X []float64
+	// F is the result of evaluating the function at X.
+	F float64
+	// Gradient holds the gradient of the function at X.
 	Gradient []float64
-	Hessian  *mat.SymDense
+	// Hessian holds the curvature of the function at X.
+	Hessian *mat.SymDense
 }
 
 // Method is a type which can search for an optimum of an objective function.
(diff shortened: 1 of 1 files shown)


kortschak:
I have added documentation about size restrictions that reflects the allocation changes from #1238.

PTAL

optimize/minimize.go
```diff
@@ -35,10 +35,21 @@ type Task struct {
...
+	X []float64
+	// F is the result of evaluating the function at X.
+	F float64
+	// Gradient holds the gradient of the function at X.
+	// The length of Gradient must match the length of X
+	// or be zero. If the capacity of Gradient is less
+	// than the length of X, a new slice will be allocated.
 	Gradient []float64
-	Hessian  *mat.SymDense
+	// Hessian holds the curvature of the function at X.
```
vladimir-ch: I would avoid mentioning curvature, there is a relation but it is not that the Hessian is the curvature.

kortschak: What would you prefer?

vladimir-ch: I would just say "Hessian holds the Hessian of the function at X". That's 5 Hessians on 3 lines but maybe it doesn't matter.

kortschak: Mmm. I was hoping for something a little more descriptive. "Hessian holds the second-order partial derivatives of the function at X."?

vladimir-ch: Yeah, that's fine too and better than curvature. Then also gradient could be equally documented as "Gradient holds the first-order partial derivatives of the function at X".

kortschak: Done. PTAL

PR #1242, opened 2020-02-19
dsp/window: use half offset to exclude flanking zeros

### LINKED ISSUE #1240: dsp/window: revisit window function definitions' use of N-1 vs N
See https://github.com/gonum/gonum/pull/1230#pullrequestreview-360730303

kortschak:
Please take a look.

Closes #1240.


kortschak:
cc @goofinator

dsp/window/doc.go
```diff
@@ -40,4 +40,9 @@
 //
 // The ɣ_max parameter is the maximum level of the side lobes of the
 // normalized spectrum, in decibels.
+//
+// Window interval
+//
+// The window intervals used in the window function are offset by a half
```
sbinet:
```suggestion
// The window intervals used in the window functions are offset by a half
```

dsp/window/window_test.go
```diff
@@ -5,82 +5,111 @@
...
-		name: "Lanczos", fn: Lanczos, winLen: 20,
-		want: []float64{0.000000, 0.115514, 0.247646, 0.389468, 0.532984, 0.669692, 0.791213, 0.889915, 0.959492, 0.995450,
-			0.995450, 0.959492, 0.889915, 0.791213, 0.669692, 0.532984, 0.389468, 0.247646, 0.115514, 0.000000},
+		name: "Lanczos", fn: Lanczos, fnCmplx: LanczosComplex, winLen: 20,
+		want: []float64{
+			0.052415, 0.170011, 0.300105, 0.436333, 0.57162, 0.698647, 0.810332, 0.900316, 0.963398, 0.995893,
+			0.995893, 0.963398, 0.900316, 0.810332, 0.698647, 0.57162, 0.436333, 0.300105, 0.170011, 0.052415,
+		},
 	},
 	// This case tests Lanczos for a NaN condition. The Lanczos NaN condition is k=(N-1)/2, that is when N is odd.
```
goofinator: // This case tests Lanczos for a NaN condition. The Lanczos NaN condition is k=(N-2)/2, 
	// that is when N is even and N>=2.
So, test with winLen=21 not checks NaN condition now.

dsp/window/window.go
```diff
@@ -30,14 +30,14 @@ func Rectangular(seq []float64) []float64 {
 // Sine window is a high-resolution window.
 //
 // The sequence weights are
-//  w[k] = sin(π*k/(N-1)),
+//  w[k] = sin(π*k/N),
```
vladimir-ch: This description (and others) misses the `+ 1/2` offset.

dsp/window/window_test.go
```diff
@@ -5,82 +5,111 @@
...
 
 var windowTests = []struct {
-	name   string
-	fn     func([]float64) []float64
-	want   []float64
-	winLen int
+	name    string
+	fn      func([]float64) []float64
+	fnCmplx func([]complex128) []complex128
+	winLen  int
```
vladimir-ch: I think `winLen` can be taken from the length of `want`.

dsp/window/window_test.go
```diff
@@ -119,7 +154,7 @@ func TestWindows(t *testing.T) {
 			dst := test.fn(src)
 
 			if !floats.EqualApprox(dst, test.want, tol) {
```
vladimir-ch: It's not so critical here but in general when comparing two slices/vectors/matrices I prefer to get a measure of their distance and then compare this value to a tolerance. It's then easy to log the distance and see if a failure is due to too strict tolerance or an actual bug.

dsp/window/window_test.go
```diff
@@ -134,15 +169,72 @@ func TestGausWindows(t *testing.T) {
...
+	const tol = 1e-6
+
+	src := make([]complex128, 20)
+	for i := range src {
+		src[i] = complex(1, 1)
+	}
+
+	for _, test := range gausWindowTests {
+		t.Run(fmt.Sprintf("%sComplex (sigma=%.1f)", test.name, test.sigma), func(t *testing.T) {
+			// Copy the input since we are mutating it.
```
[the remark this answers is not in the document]
kortschak: Done.

dsp/window/window_test.go
```diff
@@ -111,6 +145,7 @@ func TestWindows(t *testing.T) {
 
 	for _, test := range windowTests {
 		t.Run(test.name, func(t *testing.T) {
+			// Copy the input since we are mutating it.
```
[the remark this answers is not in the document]
kortschak: Done.

dsp/window/window_test.go
kortschak: Done.

dsp/window/doc.go
kortschak: Done.

dsp/window/window.go
kortschak: I had intended to show the general case and implement the specific case, but I'll fix this.

dsp/window/window_test.go
kortschak: Yeah. The slices here are short and we show them, so I'm OK with doing it this way.

dsp/window/window.go
vladimir-ch: If you did it on purpose, then it's fine as it is.

kortschak: Done.

kortschak:
PTAL

dsp/window/window.go
kortschak: I think you are right though. Having it explicit helps explain the intervals comment and makes it easier for reviewers to see errors.

PR #1248, opened 2020-02-21
optimize/functions: fix ExtendedRosenbrock for BFGS test

kortschak:
Please take a look.

Updates #1218.

### DIFF
--- a/optimize/functions/functions.go
+++ b/optimize/functions/functions.go
@@ -789,12 +789,13 @@ func (ExtendedRosenbrock) Grad(grad, x []float64) {
 	for i := range grad {
 		grad[i] = 0
 	}
+	// Prevent fused multiply add and fused multiply subtract.
 	for i := 0; i < dim-1; i++ {
-		grad[i] -= 2 * (1 - x[i])
-		grad[i] -= 400 * (x[i+1] - x[i]*x[i]) * x[i]
+		grad[i] -= float64(2 * (1 - x[i]))
+		grad[i] -= float64(400 * (x[i+1] - float64(x[i]*x[i])) * x[i])
 	}
 	for i := 1; i < dim; i++ {
-		grad[i] += 200 * (x[i] - x[i-1]*x[i-1])
+		grad[i] += float64(200 * (x[i] - float64(x[i-1]*x[i-1])))
 	}
 }
 


kortschak:
From what I can see, the higher precision allows results that are calculated in the gradient to not underflow to zero. This means that when the step is taken, because the gradients are so small, the step *does* actually underflow, thus leaving a non-zero gradient and a zero progression. This is one of the manifestations, there are others that I'm not entirely clear on; I added print statements into the tail of the function being changed here and at places in `LinesearchMethod.Iterate` and compared the amd64 and arm64 runs side by side. There are catastrophic deviations that happen when either the gradients are very small (<1e-30) or when elements of the gradient are significantly different in magnitude (>1e16).

My intention is to leave #1218 open and only check off the fixes that we're happy with (this is why I'm tagging them with commits, so we can easily go back to the fixes to see if we are). When we're happy with all of them, we can close it.

vladimir-ch:
Yes, BrownBadlyScaled function is indeed badly scaled and is difficult to optimize it with a method that doesn't take it into account. So basically even with this "fix" it works by chance and I doubt the utility of the test case. It doesn't verify that the implementation is neither correct nor numerically robust.

Interestingly, if I comment out the check that causes the failure (in `LinesearchMethod.initNextLinesearch` and relax the gradient tolerance to 1e-9, the test passes. This happens because even if the method does not advance from the current location, the history of the method is updated by incorporating the current gradient. At every retry the generated direction points more in the direction of the gradient so eventually the current location may change and move towards a local minimum. I commented on this already in the past but I'm still unsure if we should do anything about it. Not all methods have this "autocorrecting" ability, gradient descent does not, for example. We could have a retry mechanism and give up only after a certain number of retries but that increases complexity slightly and adds another magic parameter (number of retries). In this particular case one retry is enough to move the location. The current way is simple but we sometimes fail to find the minimum. Seeing that the alternative is to semi-randomly prevent fused multiply-add operations, I wouldn't be opposed to the retry mechanism.

vladimir-ch:
> Yes, BrownBadlyScaled function is indeed badly scaled ...
Oh, wrong function but ExtendedRosenbrock is not pretty either.

I tried to implement the retries but the logic is not so simple there and it would be an ugly hack too.

LGTM

PR #1252, opened 2020-02-21
testlapack: use residualOrthogonal helper in Dorg2lTest and DgesvdTest

vladimir-ch:
Please take a look.


vladimir-ch:
Yes, I'll add it. There is still `isOrthogonal` that I want to replace.

PR #1256, opened 2020-02-22
math32: add arm64 assembly implementation for Sqrt

kortschak:
Please take a look.

Closes #1215.

### DIFF
--- a/internal/math32/sqrt_arm64.go
+++ b/internal/math32/sqrt_arm64.go
@@ -1,14 +1,14 @@
+// Copyright 2009 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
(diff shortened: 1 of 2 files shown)


internal/math32/sqrt_arm64.go
```diff
@@ -0,0 +1,20 @@
+// Copyright 2009 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Copyright ©2015 The Gonum Authors. All rights reserved.
```
sbinet: 2020?

kortschak: It's a copy of the sqrt_amd64.go file without changes, so I think not.

kortschak:
See comments https://github.com/chewxy/math32/pull/11#pullrequestreview-363018766

PR #1259, opened 2020-02-25
all: update for go1.14 release

kortschak:
Please take a look.


sbinet:
Seems like we need to also update 'go.mod' to get the -lang right (c.f. the appveyor failure)

kortschak:
Done, PTAL

PR #1266, opened 2020-03-03
num/dual: fix Number example

sbinet:
Please take a look.


kortschak:
I made the same mistake in num/hyperdual, would you fix that too please.

sbinet:
done.

sbinet:
(I wish there was a linter for that... I regularly fall in that trap as well...)

PR #1267, opened 2020-03-03
dsp/window: add hamming plot example

sbinet:
Please take a look.


sbinet:
it's true godoc not displaying images is annoying at times (it's clearly one those...)
so clearly, that kind of code is not meant to be completely useful from a godoc standpoint.

that said, we could integrate these into a `README.md` file that could be displayed by `pkg.go.dev`, like that one:
- https://github.com/go-hep/hep/tree/master/hplot#hplot

right now, `pkg.go.dev` doesn't display the readme pages of "sub-packages":
- https://pkg.go.dev/go-hep.org/x/hep/hplot?tab=overview
(it always shows the "main", root one)
but I suppose it could be changed (I've sent a request about that, we'll see...)

I must say I am a bit more confident in the approach I took with the `hplot` README (automatic extraction of example code (with [campoy/embedmd](https://github.com/campoy/embedmd)) + insertion of images (checked via the `testing.T` infrastructure)) than the one I initially followed based off a Jupyter notebook (that is a bit more cumbersome to maintain: `defer` handling doesn't expose the same semantics than "regular" Go code, and maintenance over time (API changes, output changes, ...) generates a lot of friction.)

that said, I agree this could be the basis for a blog post on Gonum (or a very small seed for a bigger article on signal processing w/ Gonum), but I don't think it should be an example in gonum/plot.
this example doesn't show *how* to use gonum/plot, but rather what this Hamming window looks like (w/ a couple of plots).

but if the consensus is to not have this example in gonum/gonum, then feel free to close that PR.

kortschak:
>that said, I agree this could be the basis for a blog post on Gonum (or a very small seed for a bigger article on signal processing w/ Gonum), but I don't think it should be an example in gonum/plot.

I think this is the right thing to do, and agree that this is not an example for gonum/plot (we already cover how to use these functions in plot examples).

sbinet:
the issue w/ blog posts that are outside of the repository they try to "document" is bit rot.

kortschak:
If we have the blog post import the code snippets somehow (this is the rub) from actual runnable code, then that ceases to be an issue since the code can be tested. The https://godoc.org/golang.org/x/tools/blog package does this, though it doesn't generate an html output other than serving it. It doesn't look like a big job though to fork that package and make a static site generator from it (which would also run the code to generate images and so on).

sbinet:
ok.

PR #1270, opened 2020-03-05
Autodiff

### LINKED ISSUE #1268: diff/autofd: package to generate derivative code from a Go func
### Background

see #1265 

### Proposal

introduce a `diff/autofd` package to automatically generate the derivative's code from a `func` declaration.

```go
package autofd

type Func struct {
	Path  string // import path of the package holding the function
	Name  string // function or method name
	Deriv string // name of the output derivative function
}

// Derivative generates the derivative code from the given function declaration.
func Derivative(w io.Writer, f Func) error { ... }
```

we could also consider creating an `autofd` command leveraging that function to generate code from the command line:
```
$> autofd -pkg gonum.org/v1/gonum/diff/autofd/internal/testfunc -fct F1
func DerivF1(x float64) float64 {
	return dual.Mul(dual.Number{Real:x, Emag:1}, dual.Number{Real:x, Emag:1}).Emag
}
```

another possible work item would be to leverage [expreduce](https://github.com/corywalker/expreduce).

sbinet:
Please take a look.

Fixes #1268


diff/autofd/autofd.go
```diff
@@ -0,0 +1,298 @@
...
+		if fct == nil {
+			return nil, fmt.Errorf("could not find %q in package %q", name, path)
+		}
+
+	default:
+		obj := scope.Lookup(name)
+		if obj == nil {
+			return nil, fmt.Errorf("could not lookup %q in package %q", name, path)
+		}
+		var ok = false
```
golangcibot: ineffectual assignment to `ok` (from `ineffassign`)

kortschak: `var ok bool`

diff/autofd/autofd.go
```diff
@@ -0,0 +1,298 @@
...
+	"go/ast"
+	"go/token"
+	"go/types"
+	"io"
+	"strings"
+
+	"golang.org/x/tools/go/packages"
+)
+
+type Func struct {
```
kortschak: This needs a doc comment.

diff/autofd/autofd.go
```diff
@@ -0,0 +1,298 @@
...
+	Path  string // import path of the package holding the function
```
kortschak: Leading capitals and full stops.

diff/autofd/autofd.go
```diff
@@ -0,0 +1,298 @@
...
+		Mode: packages.NeedName |
+			packages.NeedFiles |
+			packages.NeedCompiledGoFiles |
+			packages.NeedSyntax |
+			packages.NeedTypes |
+			packages.NeedTypesInfo,
+	}
+	pkgs, err := packages.Load(cfg, path)
+	if err != nil {
+		return nil, fmt.Errorf("could not load package of %q: %w", f, err)
```
kortschak: I'd destructure `f` more for this; `("could not load package of %q %s: %w", f.Path, f.Name, err)`.

diff/autofd/autofd.go
```diff
@@ -0,0 +1,298 @@
...
+	}
+
+	pkg := func() *packages.Package {
+		for _, p := range pkgs {
+			if p.PkgPath == path {
+				return p
+			}
+		}
+		return nil
+	}()
```
kortschak:
```
	var pkg *packages.Package
	for _, p := range pkgs {
		if p.PkgPath == path {
			pkg = p
			break
		}
	}
```

diff/autofd/autofd.go
```diff
@@ -0,0 +1,298 @@
...
+		}
+		fct = func() *types.Func {
+			for i := 0; i < typ.NumMethods(); i++ {
+				m := typ.Method(i)
+				if m.Name() == name[idx+1:] {
+					return m
+				}
+			}
+			return nil
+		}()
```
kortschak:
```
	for i := 0; i < typ.NumMethods(); i++ {
		m := typ.Method(i)
		if m.Name() == name[idx+1:] {
			fct = m
			break
		}
	}
```

diff/autofd/autofd.go
```diff
@@ -0,0 +1,298 @@
...
+	}()
+
+	if pkg == nil || len(pkg.Errors) > 0 {
+		return nil, fmt.Errorf("could not find package %q", path)
+	}
+
+	var (
+		fct   *types.Func
+		scope = pkg.Types.Scope()
+	)
```
kortschak:
```
var fct *types.Func
scope := pkg.Types.Scope()
```

diff/autofd/autofd.go
```diff
@@ -0,0 +1,298 @@
...
+			for i := 0; i < typ.NumMethods(); i++ {
+				m := typ.Method(i)
+				if m.Name() == name[idx+1:] {
+					return m
+				}
+			}
+			return nil
+		}()
+		if fct == nil {
+			return nil, fmt.Errorf("could not find %q in package %q", name, path)
```
kortschak: `"could not find %s in package %q", name, path`

diff/autofd/autofd.go
```diff
@@ -0,0 +1,298 @@
...
+		}
+
+	default:
+		obj := scope.Lookup(name)
+		if obj == nil {
+			return nil, fmt.Errorf("could not lookup %q in package %q", name, path)
```
kortschak: `"could not find %s in package %q", name, path`

diff/autofd/autofd.go
```diff
@@ -0,0 +1,298 @@
...
+		}
+		var ok bool
+		fct, ok = obj.(*types.Func)
+		if !ok {
+			return nil, fmt.Errorf("object %q in package %q is not a func (%T)", name, path, obj)
```
kortschak: `"object %s in package %q is not a func (%T)", name, path, obj`

diff/autofd/autofd.go
```diff
@@ -0,0 +1,298 @@
...
+					continue
+				}
+
+				if decl.Name.Name == g.fct.Name() {
+					return decl
+				}
+			}
+		}
+		return nil
+	}()
```
kortschak:
```
	for _, f := range g.pkg.Syntax {
		for i := range f.Decls {
			decl, ok := f.Decls[i].(*ast.FuncDecl)
			if !ok {
				continue
			}

			if decl.Name.Name == g.fct.Name() {
				fct = decl
				break
			}
		}
	}
```

diff/autofd/autofd.go
```diff
@@ -0,0 +1,298 @@
...
+		}
+		return true
+	})
+
+	if returns > 1 {
+		return fmt.Errorf("can not handle functions with multiple return statements")
+	}
+	if returns == 0 {
+		return fmt.Errorf("could not find a return statement")
+	}
```
kortschak:
```
switch returns {
case 0:
    return fmt.Errorf("could not find a return statement")
case 1:
    // ok
default:
   return fmt.Errorf("can not handle functions with multiple return statements")
}
```
?

diff/autofd/autofd.go
```diff
@@ -0,0 +1,298 @@
...
+					return decl
+				}
+			}
+		}
+		return nil
+	}()
+
+	var (
+		ret     *ast.ReturnStmt
+		returns = 0
```
kortschak: `returns int`?

diff/autofd/autofd.go
```diff
@@ -0,0 +1,298 @@
...
+	}
+}
+
+func (g *generator) printf(format string, args ...interface{}) {
+	fmt.Fprintf(g.w, format, args...)
+}
+
+var (
+	f1x *types.Func
+)
```
kortschak: `var f1x *types.Func`

I'd say it needs a comment as well.

diff/autofd/autofd.go
```diff
@@ -0,0 +1,298 @@
...
+	g.printf("\treturn ")
+	g.expr(ret.Results[0])
+	g.printf(".Emag\n")
+	g.printf("}\n")
+
+	return g.err
+}
+
+func (g *generator) expr(expr ast.Expr) {
+	if g.err != nil {
```
kortschak: I can't see anywhere that this gets set. Am I missing something?

diff/autofd/autofd.go
```diff
@@ -0,0 +1,298 @@
...
+
+	args := g.fct.Type().Underlying().(*types.Signature).Params()
+	g.printf("func %s(%s float64) float64 {\n",
+		g.der,
+		args.At(0).Name(),
+	)
+	g.printf("\treturn ")
+	g.expr(ret.Results[0])
+	g.printf(".Emag\n")
+	g.printf("}\n")
```
kortschak: Can we make this do:
```
d := <dual.Func>()
return d.Emag
```
This would allow people to look at the result when debugging a little more easily if they need to.

diff/autofd/autofd_test.go
```diff
@@ -0,0 +1,182 @@
...
+import (
+	"fmt"
+	"strings"
+	"testing"
+
+	"gonum.org/v1/gonum/diff/autofd"
+)
+
+func TestDerivative(t *testing.T) {
+	for _, tc := range []struct {
```
kortschak: Please pull the table out of `TestDerivative` as `var derivativeTests` and s/tc/test/g.

diff/autofd/internal/testfunc/funcs.go
```diff
@@ -0,0 +1,100 @@
+// Copyright ©2020 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Package testfunc holds various functions to test autofd features.
+package testfunc // import "gonum.org/v1/gonum/diff/autofd/internal/testfunc"
+
+import (
+	"math"
+)
```
kortschak: `import "math"`

sbinet:
PTAL.

regarding the second derivative, one could:
- `s/Derivative/FirstDerivative/g` + introduce `SecondDerivative`
- `s/Derivative/D1x/g` + introduce `D2x`

or add another parameter to `Derivative`:
```go
type Kind byte
const (
    D1 Kind = iota
    D2
)

func Derivative(w io.Writer, f Func, kind Kind) error { ... }
```

regarding whether this should be in `gonum.org/v1/gonum` or in `gonum/tools`, I am agnostic.
(but I'd argue we should let the whole codereview process happen here, and then send another PR against `gonum/tools` if needed.)

sbinet:
(seems like the darwin build bots are in a wacky state...)

sbinet:
> `func Derivative(w io.Writer, f Func, d2 bool) error`

IMHO, boolean arguments look weird from the call site perspective:
```go
err = autofd.Derivative(w, fct, true)
```
what about:
```go
func Deriv1(w io.Writer, f Func) error { ... }
func Deriv2(w io.Writer, f Func) error { ... }
```

or:
```go
package autofd

type Option func(g *generator) error

func Derivative(w io.Writer, f Func, opts ...Option) error { ... }

func Rank(i int) Option { ...}

// ---
package user

err = autofd.Derivative(w, fct, autofd.Rank(1))
err = autofd.Derivative(w, fct, autofd.Rank(2))
```

kortschak:
I agree, but I think it's the best of a bad lot.

For the case `func Deriv2(w io.Writer, f Func) error { ... }`, this is misleading since it's not giving only the second derivative, it's giving both. I think functional options are overkill and share a problem with having a plain integer argument, e.g. `func Derivative(io.Writer, Func, int) error`, that we don't have any integer types (except for `bool`) that are constrained to have only 2 values.

Ideally, we could have the `func Derivative(w io.Writer, fn Func, maxDerivative int) error`, although that is also ugly, and has the requirement for checking the range of `maxDerivative`.

sbinet:
here is another stab, with the ability to generate second derivatives.

diff/autofd/autofd_test.go
```diff
@@ -0,0 +1,350 @@
...
+		name: autofd.Func{Path: "gonum.org/v1/gonum/diff/autofd/internal/testfunc", Name: "T1.Fxxx"},
+		err:  fmt.Errorf(`could not create first derivative generator: could not find T1.Fxxx in package "gonum.org/v1/gonum/diff/autofd/internal/testfunc"`),
+	},
+	{
+		name: autofd.Func{Path: "gonum.org/v1/gonum/diff/autofd/internal/testfunc", Name: "ErrT1.F"},
+		err:  fmt.Errorf(`could not create first derivative generator: could not find ErrT1.F in package "gonum.org/v1/gonum/diff/autofd/internal/testfunc"`),
+	},
+}
+
+var d2xTests = []struct {
```
golangcibot: `d2xTests` is unused (from `varcheck`)

diff/autofd/autofd.go
```diff
@@ -0,0 +1,373 @@
...
+		g.der,
+		args.At(0).Name(),
+	)
+	g.printf("\tv := ")
+	g.expr(ret.Results[0])
+	switch g.rank {
+	case d1xKind:
+		g.printf("\n\treturn v.Emag\n")
+	case d2xKind:
+		g.printf("\n\treturn v.E1E2mag\n")
```
kortschak: I think this is a waste. The hyperdual returns both the first and second derivatives of the function at x; `g.printf("\n\treturn v.E1mag, v.E1E2mag\n")`. This is why I think that the parameterisation should be a `bool` indicating that the second derivative is invuded.

diff/autofd/autofd.go
```diff
@@ -0,0 +1,373 @@
...
+		// ok
+	default:
+		return fmt.Errorf("too many return values")
+	}
+
+	args := g.fct.Type().Underlying().(*types.Signature).Params()
+	g.printf("func %s(%s float64) float64 {\n",
+		g.der,
+		args.At(0).Name(),
+	)
```
kortschak:
```
switch g.rank {
case d1xKind:
	g.printf("func %s(%s float64) float64 {\n",
		g.der,
		args.At(0).Name(),
	)
case d2xKind:
	g.printf("func %s(%s float64) (d1, d2 float64) {\n",
		g.der,
		args.At(0).Name(),
	)
}
```

diff/autofd/cmd/autofd/main.go
```diff
@@ -0,0 +1,71 @@
...
+ex:
+ $> autofd -pkg gonum.org/v1/gonum/diff/autofd/internal/testfunc -fct F1
+ func DerivF1(x float64) float64 {
+ 	return dual.Mul(dual.Number{Real:x, Emag:1}, dual.Number{Real:x, Emag:1}).Emag
+ }
+
+ $> autofd -pkg gonum.org/v1/gonum/diff/autofd/internal/testfunc -fct F1 -der DxF1
+ func DxF1(x float64) float64 {
+ 	return dual.Mul(dual.Number{Real:x, Emag:1}, dual.Number{Real:x, Emag:1}).Emag
+ }
```
kortschak: These should probably be updated to match the form of the code as it now is.

diff/autofd/autofd.go
sbinet: my reasonning (but I admit I haven't benchmarked it) was that it would be CPU-wasteful to use `hyperdual` when only `dual` was needed.

if the overhead of `hyperdual` wrt `dual` is indeed minimal, then I guess that indeed we can indeed just generated `func Derivative(x float64) (d1, d2 float64) { ... }` and be done with it.
(and let users possibly wrap that function to only return one of the two values)

diff/autofd/cmd/autofd/main.go
sbinet: right.

diff/autofd/autofd.go
kortschak: I agree, but if you need the second derivative, you get the first for free. If you don't, you only use the dual. With the approach that's here, if you need both the first and second derivatives, you need to generate two functions, and you need to call both functions.

kortschak:
I don't think I explained correctly, sorry. We should have `func Derivative(w io.Writer, f Func, d2 bool) error` and generate the hyperdual case when `d2` is true, the dual otherwise.

sbinet:
yes (your last comment and my PR crossed each others)

diff/autofd/autofd.go
```diff
@@ -22,9 +22,11 @@ type Func struct {
 	Deriv string // Name of the output derivative function.
 }
 
-// Derivative generates code for both the first and second derivatives from the given function declaration.
-func Derivative(w io.Writer, f Func) error {
-	gen, err := newGenerator(w, f)
+// Derivative generates code for derivatives from the given function declaration.
+// It generates only the first derivative when d2 is false and generates
+// both first and second derivatives otherwise.
```
kortschak:
```
// If d2 is true, both the generated function returns both the first and second
// derivatives. Otherwise only the first derivative function is generated.
```

sbinet:
ok.

sbinet:
see https://github.com/gonum/tools/pull/17

Issue #1260, opened 2020-02-26
mathext/internal/amos: revisit fixes for Zseri and Zacai failures on arm64


kortschak:
See #1218 and 08cd598920a727fc67106acff7e04c1ffde895db, we just skip on arm64.

vladimir-ch:
If I comment out the `t.Skipf` switch in `TestZacai` and run
```
env GOARCH=arm64 go test -v -exec qemu-aarch64 -run Zacai
```
I don't get any errors.

kortschak:
It appears to be contextual. Running on real arm64 hardware:

```
~/src/gonum.org/v1/gonum/mathext/internal/amos [master*]$ go test -run Zacai
PASS
ok  	gonum.org/v1/gonum/mathext/internal/amos	31.254s
~/src/gonum.org/v1/gonum/mathext/internal/amos [master*]$ go test -run '(Airy|Zacai)'
--- FAIL: TestZacai (31.03s)
    amos_test.go:468: Case zacai yi_idx_1: Float64 mismatch. c = 2.4275995843707387e+72, native = 3.269273576136148e+72
         cb: 5689730552559693195, nb: 5691875935004640452
FAIL
exit status 1
```

I'll look into it.

kortschak:
So, it's contextual because the `randInput` function uses `rand.Intn` instead of `rnd.Intn`. This is the good news. The bad news is that fixing this makes the arm64 (but not amd64) build fail consistently. Changing the order of calls to `rnd.Intn`/`rnd.Float64` (via `randnum`) can make `Zacai` pass consistently.

I'll send a PR fixing the rand calls.

vladimir-ch:
I compared `Zacai` to the Fortran original and didn't spot any issue. Since `Zacai` calls `Zseri`, I think the root cause will be in `Zseri`. The computations in `zseriOrig` and `Zseri` start to substantially differ in the loop at line 2029 but only after many iterations. Since `Zseri` is the only "Go-ified" routine that uses `complex128` instead of two `float64` for complex numbers, I suspect the differences will stem from different floating point operations being used in the two implementations.

kortschak:
That's certainly possible. The arm64 fused operations are added in to complex multiplication, so they could be having an effect there that we can't easily remove.

PR #1275, opened 2020-03-08
amoslib: remove stale TODO and document package

kortschak:
Please take a look.


kortschak:
@btracey I'd like to check that you're happy with 1. keeping the fortran (I think it may be useful to address/debug #1260) and 2. that the failures are not a significant problem.

kortschak:
*ping* @btracey

kortschak:
I'm going to merge this and file an issue to revisit the failures that are papered over. I'll also send a PR to note that the tests in question should not be deleted until that issue has been addressed.

PR #1277, opened 2020-03-09
Add cfloats package

whipstein:
Added Matt Braunstein to contributors list for future contributions.

Please take a look.


kortschak:
Normally we add a person to A+C together after they have made their first contribution. The commit message would then be, "A+C: add Matt Braunstein". If you add yourself to AUTHORS as well, then we can have this PR ready for your first contrib.

PR #1287, opened 2020-03-14
.travis: experiment with faster build

egonelbre:
This should reduce the overhead of starting and running tests. 

This is a quick draft the final solution might look different.

This is something similar I did in https://github.com/mattn/go-sqlite3/pull/693, which reduced the total running time 2x.

Initial run demonstrates the time difference https://travis-ci.org/github/gonum/gonum/builds/662330315

### DIFF
--- a/.travis/run-parts
+++ b/.travis/run-parts
@@ -13,9 +13,25 @@ if [[ "${dir}" = "" || ! -d "${dir}" ]]; then
     exit 1
 fi
 
-for f in $(echo ${dir%/}/*); do
-	[ -d $f ] && continue
-	[ ! -x $f ] && continue
-	echo "+$f"
-	$f
+fold_start() {
+    echo -e "travis_fold:start:$1\033[33;1m$2\033[0m"
+}
+
+fold_end() {
+    echo -e "\ntravis_fold:end:$1\r"
+}
+
+for tagset in '' 'bounds' 'noasm' 'appengine' ;
+do
+    export TAGS="-tags $tagset"
+    fold_start tag "${TAGS}"
+    for f in $(echo ${dir%/}/*); do
+        [ -d $f ] && continue
+        [ ! -x $f ] && continue
+
+        fold_start part "$(basename $f)"
+        $f
+        fold_end part
+    done
+    fold_end tag
 done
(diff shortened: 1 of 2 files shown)


.travis/run-parts
```diff
@@ -13,9 +13,25 @@ if [[ "${dir}" = "" || ! -d "${dir}" ]]; then
...
+    echo -e "travis_fold:start:$1\033[33;1m$2\033[0m"
+}
+
+fold_end() {
+    echo -e "\ntravis_fold:end:$1\r"
+}
+
+for tagset in '' 'bounds' 'noasm' 'appengine' ;
+do
+    export TAGS=$tagset
```
kortschak: This needs to include the `-tags=` prefix if `$tagset` is set.

egonelbre: Ah, right, still trying to get everything to work nicely :). Thanks.

kortschak:
I suspect that this was fail out due to overtime. The approach we use now was actually done to reduce the time in running tests.

egonelbre:
@kortschak what's the limit on time for each job?

AFAIK, travis does have some large limit (50min), but the only criteria is that the job must be outputting something in a decent interval.

PS: the current approach definitely isn't optimal, since it's running some things multiple times that don't make a difference with different tags.

egonelbre:
@kortschak, indeed, it looks like there isn't as big of a win here as I hoped.

kortschak:
Yes, 50 minutes is the limit and 10 for silence (we don't hit that ever unless something terrible has happened).

I agree that it's not optimal. What it your plan to only run tagged tests where they are impinging on the package being tested? Do we collate a list of packages that use or have a transitive dependency on a package that uses the tag in the tag set?

egonelbre:
I think most of the tests end up transitively depending on internal/math, internal/asm. So I don't think that approach would help significantly.

If you mean only test the packages that have specific tags. I think having other packages test, some other package behavior is a really good way to flush out corner cases.

kortschak:
The only one that it really likely to save any thing would be bounds; that tightens the bounds checking in calls that are internal in mat, so it would only affect packages that import mat.

egonelbre:
So one thing I noticed, is that none of the tests have been marked with "t.Parallel()". It probably wouldn't help test speed on travis, however would help locally. A quick local experiment with mat package made it go from 30s -> 8s.

kortschak:
That sounds like a worthwhile thing to do.

egonelbre:
I'll close this one and start afresh.

PR #1289, opened 2020-03-14
travis: replace golangci.com service with travis-local golangci-lint

kortschak:
Please take a look.

The golangci.com linting service is [going away next month](https://medium.com/golangci/golangci-com-is-closing-d1fc1bd30e0e), so this replaces that with work done on travis. It is a little harsher since it will actually fail a build that fails a lint check, but we can adjust that if needed, and lint checks can be tuned or turned off with [`//nolint` directives](https://github.com/golangci/golangci-lint#nolint).

[Currently linting cannot be done on arm64](https://travis-ci.community/t/git-fetch-shallow-exclude-does-not-work-on-arm64-architecture/7583), so asm lint won't be checked on that arch for now.

### DIFF
--- a/.travis/run.d/linux/02-check-lint.sh
+++ b/.travis/run.d/linux/02-check-lint.sh
@@ -1,0 +1,1 @@
+../../script.d/check-lint.sh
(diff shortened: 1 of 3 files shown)


kortschak:
*ping*

`golangci.com` is going away in a couple of weeks.

kortschak:
Agreed. I use a source-based install locally. With modules, hopefully they will come around at some point.

I've filed  golangci/golangci-lint#1006.

PR #1291, opened 2020-03-17
mat: don't use global rand state

kortschak:
Please take a look.

See https://github.com/gonum/gonum/pull/1288#issuecomment-599844471

### DIFF
--- a/mat/cholesky_test.go
+++ b/mat/cholesky_test.go
@@ -331,10 +331,11 @@ func TestCloneCholesky(t *testing.T) {
 
 func TestCholeskyInverseTo(t *testing.T) {
 	t.Parallel()
+	rnd := rand.New(rand.NewSource(1))
 	for _, n := range []int{1, 3, 5, 9} {
 		data := make([]float64, n*n)
 		for i := range data {
-			data[i] = rand.NormFloat64()
+			data[i] = rnd.NormFloat64()
 		}
 		var s SymDense
 		s.SymOuterK(1, NewDense(n, n, data))
@@ -363,13 +364,13 @@ func TestCholeskyInverseTo(t *testing.T) {
 
 func TestCholeskySymRankOne(t *testing.T) {
 	t.Parallel()
-	rand.Seed(1)
+	rnd := rand.New(rand.NewSource(1))
 	for _, n := range []int{1, 2, 3, 4, 5, 7, 10, 20, 50, 100} {
 		for k := 0; k < 50; k++ {
 			// Construct a random positive definite matrix.
 			data := make([]float64, n*n)
 			for i := range data {
-				data[i] = rand.NormFloat64()
+				data[i] = rnd.NormFloat64()
 			}
 			var a SymDense
 			a.SymOuterK(1, NewDense(n, n, data))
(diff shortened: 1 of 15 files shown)


egonelbre:
I do agree that they shouldn't set seed from tests. However, by hardcoding the seed to 1 the tests will point out less bugs. Most of the tests should be able to deal with any input.

kortschak:
Setting the seeds to 1 is equivalent to using the global rand with only one test. This was the rationale for doing it this way. Do you have an alternative suggestion?

egonelbre:
Randomize the seed, unless there's a good reason not to.

kortschak:
I think this would be OK if there is a flag for that, defaulting to 1, and outputting the seeds if not 1.

egonelbre:
Why would you want to default it to 1?

kortschak:
The tests are primarily regression tests, so having a consistent build is very helpful. The value of 1 is used because that reflects the seed of the global rand and so makes the behaviour of a test the same as if the global rand had been used when running only a single test.

I agree that being able to fuzz the tests is a benefit, but in the general case it's less important than the regression testing.

egonelbre:
I think we in this regard have different opinions. I highly value the fuzzing part since it will keep finding existing bugs, either in the code or tests. It does mean however for some time builds would end up inconsistent. I do think, it's better for these to come from a test environment rather than from miscalculation from production.

With regards to consistent builds, the current approach wasn't too consistent either, since an addition or removal of a test could change the values returned by all the random tests following it.

In some sense the bug https://github.com/gonum/gonum/issues/1293 could've stayed hidden and caused silent problems upstream for quite a bit longer as a result.

There of course could be a gradual move towards completely random tests. e.g. there's one builder that is allowed to fail with random. But, I think that would lengthen the bug fixing period.

However, I'm definitely not a primary maintainer so the decision will be up to you (primary maintainers).

egonelbre:
Note, for similar reasons I would prefer `rand.Float64()` (with a slight chance to return NaN-s and other weird values) over `rand.NormFloat64()`. `NormFloat64` can return the same range of values, however it's going to be less likely to hit those extremes.

kortschak:
There is a tension here. I agree, fuzzing has a value, but it's something that should be in addition to the consistent builds.

You're also right that the previous case was not consistent. This was a bug; none of the tests should be using shared random number generators, but the changes to remove those were missed here.

I think there are two approaches that could be used here that gain the benefit of both consistent builds and randomised testing. The first is to add on randomised testing that is run optionally and the second is to run randomised testing with the seed being emitted on failure and being able to set the seed for investigation of bugs that are found in testing (in the case of optional randomised testing this would also be useful).

egonelbre:
I often have helpers for testing things, so in this case it could be all contained as:

```
package example_test

import "gonum.org/v1/gonum/internal/testrand"

// Usually a test would use:
func TestCommon(t *testing.T) {
    src := testrand.NewSource(t)
    rnd := testrand.New(t)
}

// When some test depends on a particular seed being used:
func TestDependOnSeed(t *testing.T) {
    src := testrand.NewSourceSeeded(t, 1)
    rnd := testrand.NewSeeded(t, 1)
 
    // alternatively have `SetSeed` method
    src := testrand.NewSource(t)
    src.SetSeed(1)
    rnd := testrand.New(t)
    rnd.SetSeed(1)
}
```

kortschak:
What does the testrand package do with the seeds? Presumably it logs the seed to the output.

If we go to that kind of an approach, this is a consistent step towards it.

egonelbre:
@kortschak I'll make a draft pr and we can discuss details there.

kortschak:
SGTM

PR #1292, opened 2020-03-17
lapack: add Dpb* functions to Float64 interface

vladimir-ch:
Also remove the dependence of DpbconTest on Dlansber (and Dlanger) interfaces because the netlib package doesn't provide Dlansb.

Please take a look.


kortschak:
The failure here is #1293.

PR #1294, opened 2020-03-18
blas/gonum,mat: handle NaN in destination matrices

### LINKED ISSUE #1293: blas/gonum: [DS]symm and [ZC]hemm do not zero out NaN when beta = 0 and alpha != 0
See https://github.com/gonum/gonum/pull/1288#issuecomment-599844471

This showed up because the running of parallel tests allowed workspace matrices containing NaN to be visible to `Dsymm` as part of `Dense.Mul` when a non-symetric parameter was handed in as the transpose (thus needing a temporary workspace).

The [code currently](https://github.com/gonum/gonum/blob/3efb0d11eefea4bb09aebd92e3febb8b78c836fb/blas/gonum/level3float64.go#L270-L287) only zeroes out C by beta when alpha is 0 as well.

For example in `Dsymm`:

```
	if alpha == 0 {
		if beta == 0 {
			for i := 0; i < m; i++ {
				ctmp := c[i*ldc : i*ldc+n]
				for j := range ctmp {
					ctmp[j] = 0
				}
			}
			return
		}
		for i := 0; i < m; i++ {
			ctmp := c[i*ldc : i*ldc+n]
			for j := 0; j < n; j++ {
				ctmp[j] *= beta
			}
		}
		return
	}
```

The same issue exists in the complex implementations. Possibly also others.

cc @vladimir-ch

kortschak:
Please take a look.

Updates #1293 (there may be other cases of this that the mat tests have not found).

### DIFF
--- a/blas/gonum/level3cmplx128.go
+++ b/blas/gonum/level3cmplx128.go
@@ -1027,17 +1027,30 @@ func (Implementation) Zsyrk(uplo blas.Uplo, trans blas.Transpose, n, k int, alph
 			for i := 0; i < n; i++ {
 				ci := c[i*ldc+i : i*ldc+n]
 				ai := a[i*lda : i*lda+k]
-				for jc, cij := range ci {
-					j := i + jc
-					ci[jc] = beta*cij + alpha*c128.DotuUnitary(ai, a[j*lda:j*lda+k])
+				if beta == 0 {
+					for jc := range ci {
+						j := i + jc
+						ci[jc] = alpha * c128.DotuUnitary(ai, a[j*lda:j*lda+k])
+					}
+				} else {
+					for jc, cij := range ci {
+						j := i + jc
+						ci[jc] = beta*cij + alpha*c128.DotuUnitary(ai, a[j*lda:j*lda+k])
+					}
 				}
 			}
 		} else {
 			for i := 0; i < n; i++ {
 				ci := c[i*ldc : i*ldc+i+1]
 				ai := a[i*lda : i*lda+k]
-				for j, cij := range ci {
-					ci[j] = beta*cij + alpha*c128.DotuUnitary(ai, a[j*lda:j*lda+k])
+				if beta == 0 {
+					for j := range ci {
+						ci[j] = alpha * c128.DotuUnitary(ai, a[j*lda:j*lda+k])
+					}
+				} else {
+					for j, cij := range ci {
+						ci[j] = beta*cij + alpha*c128.DotuUnitary(ai, a[j*lda:j*lda+k])
+					}
 				}
 			}
 		}
--- a/blas/testblas/zhemm.go
+++ b/blas/testblas/zhemm.go
@@ -10,6 +10,7 @@ import (
 	"testing"
 
 	"golang.org/x/exp/rand"
+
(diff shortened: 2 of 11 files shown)


kortschak:
With the last edition, I think this fixes #1293 now. I have also tested the tests against netlib/blas and they still pass.

blas/gonum/level3cmplx128.go
```diff
@@ -306,15 +306,17 @@ func (Implementation) Zhemm(side blas.Side, uplo blas.Uplo, m, n int, alpha comp
...
 	}
 
-	if alpha == 0 {
-		if beta == 0 {
-			for i := 0; i < m; i++ {
-				ci := c[i*ldc : i*ldc+n]
-				for j := range ci {
-					ci[j] = 0
-				}
+	if beta == 0 {
```
vladimir-ch: Can you please summarize what is the essence of the change? If alpha is zero, the behavior doesn't change, for any value of beta the function sets C and returns. For non-zero alpha and zero beta C is first zeroed out but then overwritten with alpha*A*B so the zeroing out was unnecessary? However, the issues has been fixed, so I must be missing something?

blas/testblas/zhemm.go
```diff
@@ -46,81 +47,90 @@ func zhemmTest(t *testing.T, impl Zhemmer, side blas.Side, uplo blas.Uplo, m, n
...
-						// the expected result using zmm.
-						aHem := make([]complex128, len(a))
-						copy(aHem, a)
-						if uplo == blas.Upper {
-							for i := 0; i < nA; i++ {
-								aHem[i*lda+i] = complex(real(aHem[i*lda+i]), 0)
-								for j := i + 1; j < nA; j++ {
-									aHem[j*lda+i] = cmplx.Conj(aHem[i*lda+j])
+						for _, nanC := range []bool{false, true} {
+							if nanC && beta != 0 {
```
vladimir-ch: Can you please add a comment explaining why we are not interested in this test case? Also at the other places.

mat/list_test.go
```diff
@@ -540,6 +540,204 @@ func makeRandOf(a Matrix, m, n int, src rand.Source) Matrix {
 	return rMatrix
 }
 
+// makeNaNOf returns a new randomly filled m×n matrix of the underlying matrix type.
```
vladimir-ch: "m×n matrix filled with NaN values"?

mat/list_test.go
```diff
@@ -540,6 +540,204 @@ func makeRandOf(a Matrix, m, n int, src rand.Source) Matrix {
...
+func makeNaNOf(a Matrix, m, n int) Matrix {
+	var rMatrix Matrix
+	switch t := a.(type) {
+	default:
+		panic("unknown type for make rand of")
```
vladimir-ch: s/make rand of/makeNaNOf

mat/list_test.go
kortschak: Done

mat/list_test.go
kortschak: Done

blas/testblas/zhemm.go
kortschak: Done

PR #1295, opened 2020-03-19
internal/testrand: random package for testing

egonelbre:
Testing package that allows seed randomization and setting a bias for extreme and nan values.

### DIFF
--- a/.travis/script.d/test-coverage.sh
+++ b/.travis/script.d/test-coverage.sh
@@ -49,8 +49,8 @@ testCover() {
 # Init coverage.txt
 echo "mode: ${MODE}" > $ACC_OUT
 
-# Run test coverage on all directories containing go files except testlapack testblas and testgraph.
-find . -type d -not -path '*testlapack*' -and -not -path '*testblas*' -and -not -path '*testgraph*' | while read d; do testCover $d || exit; done
+# Run test coverage on all directories containing go files except testlapack, testblas, testgraph and testrand.
+find . -type d -not -path '*testlapack*' -and -not -path '*testblas*' -and -not -path '*testgraph*' -and -not -path '*testrand*' | while read d; do testCover $d || exit; done
 
 # Upload the coverage profile to coveralls.io
 [ -n "$COVERALLS_TOKEN" ] && ( goveralls -coverprofile=$ACC_OUT || echo -e '\n\e[31mCoveralls failed.\n' )
--- a/internal/testrand/extreme.go
+++ b/internal/testrand/extreme.go
@@ -1,0 +1,210 @@
+// Copyright ©2020 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package testrand
+
+import (
+	"math"
+
+	"golang.org/x/exp/rand"
+)
+
+// extreme is a pseudo-random number generator that has high probability of returning extreme values.
+type extreme struct {
+	probability    float64
+	nanProbability float64
+
+	rnd Rand
+}
+
+// newExtreme creates a new extreme pseudo-random generator.
+// p is the probability of returning an extreme value.
+// nan is the probability of returning a NaN.
+func newExtreme(p, nan float64, rnd Rand) *extreme {
+	return &extreme{p, nan, rnd}
+}
+
+// Perm returns a permutation of integers [0, n).
+func (e *extreme) Perm(n int) []int { return e.rnd.Perm(n) }
+
+// Read generates len(p) pseudo-random bytes.
+func (e *extreme) Read(p []byte) (n int, err error) { return e.rnd.Read(p) }
+
+// Seed reseeds the pseudo-random generator.
+func (e *extreme) Seed(seed uint64) { e.rnd.Seed(seed) }
+
+// Shuffle shuffles n items using the swap callback.
+func (e *extreme) Shuffle(n int, swap func(i, j int)) { e.rnd.Shuffle(n, swap) }
+
+// p returns true when the generator should output an extreme value.
+func (e *extreme) p() bool {
+	if e.probability <= 0 {
+		return false
+	}
+	return e.rnd.Float64() < e.probability
+}
+
+// nan returns true when the generator should output nan.
+func (e *extreme) nan() bool {
+	if e.nanProbability <= 0 {
+		return false
+	}
+	return e.rnd.Float64() < e.nanProbability
+}
+
+// ExpFloat64 returns an exponentialy distributed pseudo-random float64 in range (0, math.MaxFloat64].
+func (e *extreme) ExpFloat64() float64 {
+	switch {
+	case e.p():
+		return extremeFloat64Exp[e.rnd.Intn(len(extremeFloat64Exp))]
+	case e.nan():
+		return math.NaN()
+	}
+
+	return e.rnd.ExpFloat64()
+}
+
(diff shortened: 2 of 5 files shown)


egonelbre:
@kortschak this is what I had in mind. Any improvements/reductions etc. are welcome.

internal/testrand/extremes.go
```diff
@@ -0,0 +1,111 @@
...
+		-1,
+		-math.SmallestNonzeroFloat32,
+		0,
+		math.SmallestNonzeroFloat32,
+		1,
+		math.MaxFloat32 / 2,
+		math.MaxFloat32,
+	}
+
+	extremeFloat32Exp = [...]float32{
```
golangcibot: `extremeFloat32Exp` is unused (from `varcheck`)

internal/testrand/extremes.go
```diff
@@ -0,0 +1,111 @@
...
+// license that can be found in the LICENSE file.
+
+package testrand
+
+import "math"
+
+const (
+	maxUint = ^uint(0)
+	maxInt  = int(maxUint >> 1)
+	minInt  = -maxInt - 1
```
golangcibot: `minInt` is unused (from `varcheck`)

internal/testrand/extremes.go
```diff
@@ -0,0 +1,111 @@
...
+
+	extremeFloat32Unit = [...]float32{
+		0,
+		math.SmallestNonzeroFloat32,
+		0.5,
+		1 - math.SmallestNonzeroFloat32,
+		1,
+	}
+
+	extremeFloat32Norm = [...]float32{
```
golangcibot: `extremeFloat32Norm` is unused (from `varcheck`)

kortschak:
This looks pretty reasonable. The only suggestion that I'd make is to not emit the seed to the log if it's 1, or the extreme rates if they are zero.

kortschak:
The travis failure here is because there is an import of math/rand, which is a prohibited import in Gonum.

vladimir-ch:
Agreed, this is nice.

egonelbre:
@kortschak fixed the tests, added docs, wasn't sure what's the best way to test it, or whether it's worth it.

internal/testrand/rand.go
```diff
@@ -0,0 +1,89 @@
...
+	nanFlag     = flag.Float64("testrand.nan", 0, "probability of returning nan")
+)
+
+// TB is an interfaces that corresponds to *testing.T and *testing.B
+type TB interface {
+	Logf(format string, args ...interface{})
+}
+
+// Source corresponds to the interface in golang.org/x/exp/rand.Source.
+type Source interface {
```
kortschak: Should we just use `type Source = golang.org/x/exp/rand.Source`?

internal/testrand/rand.go
```diff
@@ -0,0 +1,89 @@
...
+	Uint64() uint64
+	Seed(seed uint64)
+}
+
+// Rand corresponds to golang.org/x/exp/rand.Rand.
+type Rand interface {
```
kortschak: Same here.

internal/testrand/rand.go
egonelbre: Yeah, I was wondering whether to be explicit or not. I don't have a good argument for either approach.

internal/testrand/rand.go
egonelbre: Here we need the interface since `rand.Rand` is a concrete struct and to support the extreme value generation we need a different implementation.

kortschak: Sorry, you're right.

internal/testrand/rand.go
kortschak: Here I think the type alias is a nice documentation of the reason for the type.

egonelbre: Done

.travis/script.d/test-coverage.sh
```diff
@@ -49,8 +49,8 @@ testCover() {
 # Init coverage.txt
 echo "mode: ${MODE}" > $ACC_OUT
 
-# Run test coverage on all directories containing go files except testlapack testblas and testgraph.
-find . -type d -not -path '*testlapack*' -and -not -path '*testblas*' -and -not -path '*testgraph*' | while read d; do testCover $d || exit; done
+# Run test coverage on all directories containing go files except testlapack testblas and testgraph and testrand.
```
sbinet:
```suggestion
# Run test coverage on all directories containing go files except testlapack, testblas, testgraph and testrand.
```

internal/testrand/extreme.go
```diff
@@ -0,0 +1,205 @@
...
+	} else if e.nan() {
+		return float32(math.NaN())
+	}
+
+	return e.rnd.Float32()
+}
+
+// Float64 returns a pseudo-random float64 in range [0.0, 1.0).
+func (e *extreme) Float64() float64 {
+	if e.p() {
```
sbinet: ditto

internal/testrand/extreme.go
```diff
@@ -0,0 +1,205 @@
...
+func (e *extreme) nan() bool {
+	if e.nanProbability <= 0 {
+		return false
+	}
+	return e.rnd.Float64() < e.nanProbability
+}
+
+// ExpFloat64 returns an exponentialy distributed pseudo-random float64 in range (0, math.MaxFloat64].
+func (e *extreme) ExpFloat64() float64 {
+	if e.p() {
```
sbinet: according to https://golang.org/doc/effective_go.html#switch, it's idiomatic to write `if/else-if` with a `switch {}`.
```go
switch {
case e.p():
    // ...
case e.nan():
    // ...
}
```

internal/testrand/extreme.go
```diff
@@ -0,0 +1,205 @@
...
+	} else if e.nan() {
+		return math.NaN()
+	}
+
+	return e.rnd.ExpFloat64()
+}
+
+// Float32 returns a pseudo-random float32 in range [0.0, 1.0).
+func (e *extreme) Float32() float32 {
+	if e.p() {
```
sbinet: ditto

internal/testrand/extreme.go
```diff
@@ -0,0 +1,205 @@
...
+		case 3:
+			return n - 1
+		}
+	}
+	return e.Intn(n)
+}
+
+// NormFloat64 returns a normally distributed pseudo-random float64 in range [-math.MaxFloat64, math.MaxFloat64].
+func (e *extreme) NormFloat64() float64 {
+	if e.p() {
```
sbinet: ditto

internal/testrand/rand.go
```diff
@@ -0,0 +1,86 @@
...
+	"golang.org/x/exp/rand"
+)
+
+var (
+	seedFlag    = flag.Uint64("testrand.seed", 1, "random seed for tests (0=randomize)")
+	extremeFlag = flag.Float64("testrand.extreme", 0, "probability of returning extreme values")
+	nanFlag     = flag.Float64("testrand.nan", 0, "probability of returning nan")
+)
+
+// TB is an interfaces that corresponds to *testing.T and *testing.B
```
sbinet:
```suggestion
// TB is an interface that corresponds to *testing.T and *testing.B
```
(also, perhaps it's more correct to say `it corresponds to a subset of`...)

egonelbre:
@sbinet done.

egonelbre:
Not sure what's happening, but ADM64 + OSX on travis doesn't seem to want to start.

kortschak:
The macos builders on travis have been flakey for the past several days.

PR #1297, opened 2020-03-24
r3.Vec as [3]float64

sbinet:
Please take a look.

### DIFF
--- a/spatial/barneshut/barneshut3.go
+++ b/spatial/barneshut/barneshut3.go
@@ -33,7 +33,7 @@ type Force3 func(p1, p2 Particle3, m1, m2 float64, v r3.Vec) r3.Vec
 // particles and returns a zero vector when the two particles are
 // coincident, but performs no other sanity checks.
 func Gravity3(_, _ Particle3, m1, m2 float64, v r3.Vec) r3.Vec {
-	d2 := v.X*v.X + v.Y*v.Y + v.Z*v.Z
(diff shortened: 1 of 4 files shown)


egonelbre:
How does this affect performance?

Small arrays can be significantly slower: https://github.com/egonelbre/exp/blob/master/bench/vector_fusing/vector_test.go

```
# array
BenchmarkAddMul-32                      70589480                17.3 ns/op
# struct
BenchmarkStructAddMul-32                1000000000               0.740 ns/op
```

sbinet:
hum... that's unfortunate.
do you know the underlying reason for this? (it doesn't seem to be bounds-check0

egonelbre:
SSA doesn't handle arrays nicely, see https://github.com/golang/go/issues/15925 for more information.

kortschak:
I'd really rather this didn't happen. It breaks symmetry with r1 and r2, and if symmetry were retained , would break nice ergonomics (that I believe has been discussed elsewhere, but can't find) in allowing direct `r2.Point` interconversion to `vg.Point`.

I'm happy for additional operators to be added, though the API should be discussed first.

sbinet:
the `struct{x,y,z} -> [3]float64` was discussed in the issue #1049.

as the SSA issue seem a bit over my head (and been lingering for a couple of years, now) I suspect changing to an array is out of question (x20 in performances is not something we want to leave on the table)

I'll start a discussion on `gonum-dev`.

PR #1301, opened 2020-03-30
R2 vec api

sbinet:
Please take a look.

### DIFF
--- a/spatial/r2/vector.go
+++ b/spatial/r2/vector.go
@@ -35,6 +35,11 @@ func (p Vec) Dot(q Vec) float64 {
 	return p.X*q.X + p.Y*q.Y
 }
 
+// Cross returns the cross product p×q.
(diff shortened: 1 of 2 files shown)


sbinet:
same PR than #1298 (but for `r2.Vec`)

kortschak:
Ugh, this identified an issue with the approach I'm taking for doing the linting checks; it cannot extend the git history when the PR comes from another user's repo.

I'll look into it.

kortschak:
The issue here is `git fetch --shallow-exclude=master origin r2-vec-api` uses `origin` and it should determine the repo from Travis variables (I hope this is possible). It's not.

If you send a PR that outputs the result of `git remote` during a Travis build, what is that output? I think this is the only possible way to get this information.

kortschak:
Can you rebase this on to master please.

sbinet:
done.

PR #1315, opened 2020-05-03
mat: address two small TODOs

jonreiter:
use CopyVec per a TODO, references long-resolved issue #167
in Cholesky.SymRankOne don't modify receiver if singular

### DIFF
--- a/mat/cholesky.go
+++ b/mat/cholesky.go
@@ -494,11 +494,8 @@ func (c *Cholesky) ExtendVecSym(a *Cholesky, v Vector) (ok bool) {
 	//  3) c'*c + d'*d = k  =>  d = sqrt(k-c'*c)
 
 	// First, compute c = U'^-1 a
-	// TODO(btracey): Replace this with CopyVec when issue 167 is fixed.
 	w := NewVecDense(n, nil)
-	for i := 0; i < n; i++ {
-		w.SetVec(i, v.At(i, 0))
-	}
+	w.CopyVec(v)
 	k := v.At(n, 0)
 
 	var t VecDense
--- a/mat/cholesky.go
+++ b/mat/cholesky.go
@@ -528,7 +528,8 @@ func (c *Cholesky) ExtendVecSym(a *Cholesky, v Vector) (ok bool) {
 // Note that when alpha is negative, the updating problem may be ill-conditioned
 // and the results may be inaccurate, or the updated matrix A' may not be
 // positive definite and not have a Cholesky factorization. SymRankOne returns
-// whether the updated matrix A' is positive definite.
+// whether the updated matrix A' is positive definite. If the update fails
+// the receiver is left unchanged.
 //
 // SymRankOne updates a Cholesky factorization in O(n²) time. The Cholesky
 // factorization computation from scratch is O(n³).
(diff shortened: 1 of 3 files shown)


mat/cholesky.go
```diff
@@ -666,31 +663,31 @@ func (c *Cholesky) SymRankOne(orig *Cholesky, alpha float64, x Vector) (ok bool)
 	}
 	umat := c.chol.mat
 	stride := umat.Stride
+	workData := getFloats(umat.N*umat.N, false)
+	defer putFloats(workData)
+	copy(workData, umat.Data)
```
vladimir-ch, line 668: This relies on `umat.Stride == umat.N`. I think that currently this is always true, I hope it will not change in the future because this assumption is hard to spot. An alternative would be to use `workData` to create a `TriDense` and then use its `Copy` here and .

It would be also good to say in the doc comment that the receiver won't be modified if the update fails.

kortschak: This reliance on stride was my concern.The `chol` matrix is never touched by others so I think it's probably safe, but it would be worth putting a comment on the field to this effect, if we want to do this. The `TriDense.Copy` approach is safe.

jonreiter: fair enough, and reworked to be a smaller change as well

mat/cholesky.go
```diff
@@ -664,13 +663,13 @@ func (c *Cholesky) SymRankOne(orig *Cholesky, alpha float64, x Vector) (ok bool)
 			sin[i] *= -1
 		}
 	}
-	umat := c.chol.mat
-	stride := umat.Stride
+	workMat := NewTriDense(c.chol.mat.N, c.chol.triKind(), nil)
```
vladimir-ch, line 666: I've just noticed that there is a `getWorkspaceTri`/`putWorkspaceTri` function pair that will help avoid the allocation here.

mat/cholesky.go
```diff
@@ -688,9 +687,8 @@ func (c *Cholesky) SymRankOne(orig *Cholesky, alpha float64, x Vector) (ok bool)
 		}
 	}
 	if ok {
+		copy(c.chol.mat.Data, umat.Data)
```
vladimir-ch, line 690: This should be
```suggestion
		c.chol.Copy(workMat)
```

mat/cholesky.go
```diff
@@ -46,6 +46,7 @@ type Cholesky struct {
 }
 
 // updateCond updates the condition number of the Cholesky decomposition. If
+
```
kortschak, line 49: Please revert this blank line insertion.

jonreiter:
i think i've pushed a version with everything cleaned up.  1 of the required travis jobs failed but the error message doesn't mean anything to me...

kortschak:
I gave it a kick.

mat/cholesky.go
```diff
@@ -531,7 +528,8 @@ func (c *Cholesky) ExtendVecSym(a *Cholesky, v Vector) (ok bool) {
 // Note that when alpha is negative, the updating problem may be ill-conditioned
 // and the results may be inaccurate, or the updated matrix A' may not be
 // positive definite and not have a Cholesky factorization. SymRankOne returns
-// whether the updated matrix A' is positive definite.
+// whether the updated matrix A' is positive definite.  If the update fails
```
kortschak, line 531: Sorry. I missed this. The convention is single space after period.

kortschak:
We generally rebase onto master, so merging master back into working branches is unnecessary.

Can you please interactively rebase this to be two commits, one for the copy and one for the conditional update. If you want that can be rebased onto master, but that's not necessary.

jonreiter:
ive done a fresh request with a fresh clean branch

Issue #1205, opened 2019-12-20
how to  use eigen solve svd  qr pca  lstsq method ,could you give me some example ?


mullerhai:
hi ,I want to use these method ,but i find these method is struct ,not like python numpy ,I am  confuse how to use it,would you like give us some detail use toturial for these.thx

sbinet:
Some examples are on the godoc page:

https://godoc.org/gonum.org/v1/gonum/mat#example-Eigen
https://godoc.org/gonum.org/v1/gonum/mat#example-GSVD
https://godoc.org/gonum.org/v1/gonum/mat#example-Cholesky
https://godoc.org/gonum.org/v1/gonum/stat#example-PC
https://godoc.org/gonum.org/v1/gonum/stat#example-LinearRegression

hth.

kortschak:
Please take questions to gonum-dev.

mullerhai:
> https://godoc.org/gonum.org/v1/gonum/mat#example-Eigen

thx ,I check the gonum package find some engen_test.go gsvd_test.go *** file

alexandru-balan:
So there is no way of computing lstsq in go? I tried with the SVD, but the SVD.Values returned are different than what python lstsq is returning. The SVD.Values can be used indeed to construct a solution of a A*x = b system, but unlike the solution returned by lstsq in python, the go solution I calculated is not minimum norm.

Sorry for necrobumping this, but a lstsq would really be nice in GO

kortschak:
Please take a look at https://godoc.org/gonum.org/v1/gonum/mat#VecDense.SolveVec, but also note that questions should be taken to gonum-dev.

alexandru-balan:
Alright, I'll post there, but from the description it seems that SolveVec is not that different from Solve. The matrix in my system is near-singular so both would return garbage data. However the lstsq should be computed using the SVD; I tried to mimick the python implementation of lstsq, but the values returned from SVD.Values are different in GO than what python found for the same data.

kortschak:
If you post there with your Go and python code I can have a look.

kortschak:
Here is an Ax = b implementation (the AX = B implementation that the python also has is left as an exercise for the reader).
```
package main

import (
	"fmt"

	"gonum.org/v1/gonum/mat"
)

// lstsq calculates the least-squares solution to a linear matrix equation, Ax = b.
func lstsq(a *mat.Dense, b *mat.VecDense) (x *mat.VecDense, res float64, s []float64) {
[...]

PR #1316, opened 2020-05-03
mat: add SolveTo and SolveVecTo to SVD

kortschak:
WIP: needs tests and likely some optimisations.

See https://github.com/gonum/gonum/issues/1205#issuecomment-623091024.

Please take a look.


mat/solve.go
```diff
@@ -103,6 +103,51 @@ func (m *Dense) Solve(a, b Matrix) error {
 	}
 }
 
+// SolveLeastSquares calculates the least-squares solution to a linear matrix equation,
+// AV = B, placing the result in the received. It returns the residuals and the singular
```
sbinet, line 107:
```suggestion
// AV = B, placing the result in the receiver. It returns the residuals and the singular
```

mat/solve.go
```diff
@@ -138,3 +183,28 @@ func (v *VecDense) SolveVec(a Matrix, b Vector) error {
 	m := v.asDense()
 	return m.Solve(a, b)
 }
+
+// SolveLeastSquaresVec calculates the least-squares solution to a linear matrix equation,
+// Ax = b, placing the result in the received. It returns the residual and the singular
```
sbinet, line 188:
```suggestion
// Ax = b, placing the result in the receiver. It returns the residual and the singular
```

btracey:
I'm sorry for the drive-by, but it's not clear from the thread why this needs to exist. This is what `Solve` already does. Is there a particular case where this returns a different answer? If there's a need to explicitly use SVD, then maybe this should be SolveSVD instead?

btracey:
For instance:

```
func main() {
	a := mat.NewDense(4, 2, []float64{
		0, 1,
		1, 1,
		2, 1,
		3, 1,
	})
	b := mat.NewVecDense(4, []float64{-1, 0.2, 0.9, 2.1})

	c := &mat.VecDense{}
	c.SolveVec(a, b)
	fmt.Printf("x=%.4v\n", mat.Formatted(c, mat.Prefix("  ")))
}
```

```
x=⎡    1⎤
  ⎣-0.95⎦
```

kortschak:
@btracey Please see the discussion at #1205.

mat/svd.go
```diff
@@ -249,3 +259,117 @@ func (svd *SVD) VTo(dst *Dense) {
...
 	dst.Copy(tmp.T())
 }
+
+// SolveTo calculates the least-squares solution to a linear matrix equation,
+// AX = B, using a rank-truncated SVD based on cond. The result is placed in in dst.
+// It returns the residuals calculated from the complete SVD. For this value to be
+// valid the factorization must have been performed with at least SVDFullU.
+// The decomposition must have been factorized computing both the U and V singular
+// vectors, and dst must be either empty or have length equal to svd.Rank(cond).
+func (svd *SVD) SolveTo(dst *Dense, b Matrix, cond float64) []float64 {
```
kortschak, line 269: Wondering whether this should take `rank int` calculated from `svd.Rank(cond)`?

umarcor: I think so.

Also, note that it is possible to compute an approximate solution by taking a band of singular values. With full-rank matrices, it is a low-band. With rank-deficient matrices, it lies in the middle.

mat/svd.go
```diff
@@ -147,6 +147,16 @@ func (svd *SVD) Kind() SVDKind {
 	return svd.kind
 }
 
+// Rank returns the rank of A based on the count of singular values greater than cond.
+func (svd *SVD) Rank(cond float64) int {
```
umarcor, line 151: It should be possible to return the rank as part of the SVD. Depending on the underlying algorithm, this is likely to be `m*n*epsilon`, where `m` and `n` are the dimensions of A, or `s(1)*epsilon`. See http://www.netlib.no/netlib/lapack/lug/node96.html.

Naturally, if input data was truncated/quantized with a smaller precision, the value based on the machine epsilon will not be valid. In this context, this `Rank` function is useful, but I think that `cond` should be the quantization precision, and the function should take the dimensions of the problem into account.

mat/svd.go
```diff
@@ -249,3 +259,117 @@ func (svd *SVD) VTo(dst *Dense) {
...
+func (m repVector) At(i, j int) float64 {
+	if i < 0 || len(m.vec) <= i || j < 0 || m.cols <= j {
+		panic(ErrIndexOutOfRange)
+	}
+	return m.vec[i]
+}
+func (m repVector) T() Matrix { return Transpose{m} }
+
+// SolveVecTo calculates the least-squares solution to a linear matrix equation,
+// Ax = b, using a rank-truncated SVD based on cond. The result is placed in in dst.
```
sbinet, line 330:
```suggestion
// Ax = b, using a rank-truncated SVD based on cond. The result is placed in dst.
```

mat/svd.go
```diff
@@ -249,3 +259,117 @@ func (svd *SVD) VTo(dst *Dense) {
 	}
 	dst.Copy(tmp.T())
 }
+
+// SolveTo calculates the least-squares solution to a linear matrix equation,
+// AX = B, using a rank-truncated SVD based on cond. The result is placed in in dst.
```
sbinet, line 264:
```suggestion
// AX = B, using a rank-truncated SVD based on cond. The result is placed in dst.
```

mat/svd.go
kortschak: Done

mat/svd.go
kortschak: Done

mat/svd.go
kortschak: In the case of rank 0, what do we leave in `dst`?

kortschak: I will just reset the `dst`.

umarcor: I believe that any matrix with content must have rank>=1. Rank 0 would be a matrix with all zeros.

The rank-1 approximation of the SVD is composed of the largest SV, and the rank-1 approximation of the LSE is composed of the smallest non-zero SV.

kortschak: Yes, that's true, but we still need to handle it. We can return doing nothing, panic, or something else. I've done something else, but I'm not sure if it's the right thing.

umarcor: I think a panic is fair. Any user computing the SVD of a matrix should first ensure that it contains information that can be extracted. For example, computing the variance of each column and removing the ones equal zero is a sensible pre-processing. By doing so, the user would find that the matrix is zero before calling SVD.

Doing something else might involve doing the check explicitly, but I think it would add overhead for users that have done their homework.

kortschak: Done

umarcor: FTR, Matlab returns a 0. Not a matrix. By a vector of singular values with a single element, a zero.

kortschak:
Please take a look

vladimir-ch:
I think that a better solution would be to have `DGELSD` or `DGELSY` and use it in `mat`. These functions take the matrix directly instead of the SVD factorization, so then the functions shouldn't be hanged on `mat.SVD`. A good thing is that we already have a lot of dependencies of DGELSD in `lapack/gonum` but still the implementation would require some effort.

If you want, then this can exist as an interim solution.

kortschak:
Would you expect the API to look the same?

vladimir-ch:
I thought that it could look something like
```
// simplest but does not expose all computed information by Dgelsd
func SolveLeastSquares(dst *Dense, a, b Matrix) (ok bool)

// provides singular values and computed rank assuming the default rcond; allocates values
func SolveLeastSquares(dst *Dense, a, b Matrix) (values []float64, rank int, ok bool)

// provides singular values and computed rank assuming the given rcond; allocates values
func SolveLeastSquares(dst *Dense, a, b Matrix, cutoff float64) (values []float64, rank int, ok bool)
```
It could also have a dedicated type:
```
package mat

type LeastSquares struct {
    Cutoff float64 // or cutoff float64

    values []float64
    rank int
}

func (ls *LeastSquares) SolveTo(dst *Dense, a, b Matrix) (ok bool) {
or
func (ls *LeastSquares) SolveTo(dst *Dense, a, b Matrix, cutoff float64) (ok bool) {
    ls.cutoff = cutoff
    ls.values = reuse(ls.values)
    // ...
    ls.rank = ...
}
func (ls *LeastSquares) Rank() int { ... }
func (ls *LeastSquares) SingularValues(dst []float64) []float64 { ... }
```
but I'm not sure it's the right fit here.

kortschak:
It looks like there is indeed a fair bit to do to add either of those (about half the dependencies for `DGELSD` and more than half (though fewer) for `DGELSY`. Given this, it's probably worth having this go in now (when it's ready). Also, given that this approach is used in a other places, there's likely no reason to remove it when they are added (and for symmetry with the other decompositions that are used as solvers).

We can leave #1205 open or open a new issue to add (probably both of) `DGELSD`/`DGELSY` and add some functions like you have above, with the discussion about the exact form in that issue (I'm inclined towards the second on the basis that it works out of the box, but users can examine the returned values if they want to use a different rank condition threshold.

vladimir-ch:
Agreed. I'll take a look at this PR now.

kortschak:
http://www.math.clemson.edu/~petersj/m860/LeastSquares/lecture001.html is one that I saw. NumPy uses dgelsd.

mat/svd_test.go
```diff
@@ -212,3 +212,448 @@ func extractSVD(svd *SVD) (s []float64, u, v *Dense) {
...
+
+func TestSVDSolveTo(t *testing.T) {
+	t.Parallel()
+	rnd := rand.New(rand.NewSource(1))
+	// Hand-coded cases.
+	for i, test := range []struct {
+		a    [][]float64
+		b    [][]float64
+		cond float64
+		ans  [][]float64
```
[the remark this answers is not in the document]
kortschak: We should probably go through and fix these elsewhere as well.

mat/svd_test.go
```diff
@@ -212,3 +212,448 @@ func extractSVD(svd *SVD) (s []float64, u, v *Dense) {
 	svd.VTo(v)
 	return svd.Values(nil), u, v
 }
+
+func TestSVDSolveTo(t *testing.T) {
+	t.Parallel()
+	rnd := rand.New(rand.NewSource(1))
+	// Hand-coded cases.
+	for i, test := range []struct {
+		a    [][]float64
```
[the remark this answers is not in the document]
kortschak: We should make an issue to remove these throughout. They are a hold over from very early Gonum.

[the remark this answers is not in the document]
kortschak: I just remembered by I haven't done this in the past. The convenience of using shape to represent shape is too alluring.

mat/svd.go
```diff
@@ -249,3 +259,123 @@ func (svd *SVD) VTo(dst *Dense) {
 	}
 	dst.Copy(tmp.T())
 }
+
+// SolveTo calculates the least-squares solution to a linear matrix equation,
```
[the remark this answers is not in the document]
kortschak: Do you have suggested words for this?

mat/svd.go
kortschak: 2. Seems reasonable.

kortschak:
PTAL

mat/svd.go
```diff
@@ -147,6 +149,26 @@ func (svd *SVD) Kind() SVDKind {
 	return svd.kind
 }
 
+// Rank returns the rank of A based on the count of singular values greater than
+// rcond scaled by the first singular value.
```
vladimir-ch, line 153: I'd say "the largest" instead of "the first".

mat/svd.go
```diff
@@ -147,6 +149,26 @@ func (svd *SVD) Kind() SVDKind {
...
+// Rank will panic if the receiver does not contain a successful factorization or
+// rcond is negative
```
vladimir-ch, line 155: Full stop.

mat/svd_test.go
```diff
@@ -212,3 +212,451 @@ func extractSVD(svd *SVD) (s []float64, u, v *Dense) {
...
+func TestSVDSolveTo(t *testing.T) {
+	t.Parallel()
+	rnd := rand.New(rand.NewSource(1))
+	// Hand-coded cases.
+	for i, test := range []struct {
+		a      []float64
+		m, n   int
+		b      []float64
+		bc     int
+		cond   float64
```
vladimir-ch, line 225: `rcond`

mat/svd_test.go
```diff
@@ -212,3 +212,451 @@ func extractSVD(svd *SVD) (s []float64, u, v *Dense) {
...
+		svd.SolveTo(&x, b, rank)
+		if !EqualApprox(&x, want, 1e-12) {
+			t.Errorf("Solve answer mismatch. Want %v, got %v", want, x)
+		}
+	}
+
+	// Random Cases.
+	for i, test := range []struct {
+		m, n, bc int
+		cond     float64
```
vladimir-ch, line 425: `rcond`

mat/svd_test.go
```diff
@@ -212,3 +212,451 @@ func extractSVD(svd *SVD) (s []float64, u, v *Dense) {
...
+		ok := svd.Factorize(a, SVDFull)
+		if !ok {
+			t.Errorf("unexpected factorization failure for test %d", i)
+			continue
+		}
+
+		var x Dense
+		rank := svd.Rank(test.cond)
+		if rank == 0 {
+			continue
```
vladimir-ch, line 414: I don't know if this is triggered but add `t.Log`? If done, then also below.

mat/svd_test.go
```diff
@@ -212,3 +212,451 @@ func extractSVD(svd *SVD) (s []float64, u, v *Dense) {
...
+}
+
+func TestSVDSolveVecTo(t *testing.T) {
+	t.Parallel()
+	rnd := rand.New(rand.NewSource(1))
+	// Hand-coded cases.
+	for i, test := range []struct {
+		a    [][]float64
+		b    []float64
+		cond float64
```
vladimir-ch, line 487: `rcond`

mat/svd_test.go
```diff
@@ -212,3 +212,451 @@ func extractSVD(svd *SVD) (s []float64, u, v *Dense) {
...
+		ans  []float64
```
vladimir-ch, line 488: `want`

mat/svd_test.go
```diff
@@ -212,3 +212,451 @@ func extractSVD(svd *SVD) (s []float64, u, v *Dense) {
...
+		}
+	}
+}
+
+func TestSVDSolveVecTo(t *testing.T) {
+	t.Parallel()
+	rnd := rand.New(rand.NewSource(1))
+	// Hand-coded cases.
+	for i, test := range []struct {
+		a    [][]float64
```
vladimir-ch, line 485: `[]float64` also here

mat/svd_test.go
```diff
@@ -212,3 +212,451 @@ func extractSVD(svd *SVD) (s []float64, u, v *Dense) {
...
+		ok := svd.Factorize(a, SVDFull)
+		if !ok {
+			t.Errorf("unexpected factorization failure for test %d", i)
+			continue
+		}
+
+		var x VecDense
+		rank := svd.Rank(test.cond)
+		if rank == 0 {
+			continue
```
vladimir-ch, line 601: `t.Log`?

mat/svd_test.go
```diff
@@ -212,3 +212,451 @@ func extractSVD(svd *SVD) (s []float64, u, v *Dense) {
...
+		svd.SolveVecTo(&x, b, rank)
+		if !EqualApprox(&x, ans, 1e-12) {
+			t.Errorf("Solve answer mismatch. Want %v, got %v", ans, x)
+		}
+	}
+
+	// Random Cases.
+	for i, test := range []struct {
+		m, n int
+		cond float64
```
vladimir-ch, line 612: `rcond`

mat/svd_test.go
```diff
@@ -212,3 +212,451 @@ func extractSVD(svd *SVD) (s []float64, u, v *Dense) {
...
+		ok := svd.Factorize(a, SVDFull)
+		if !ok {
+			t.Errorf("unexpected factorization failure for test %d", i)
+			continue
+		}
+
+		var x VecDense
+		rank := svd.Rank(test.cond)
+		if rank == 0 {
+			continue
```
vladimir-ch, line 648: `t.Log`?

mat/svd_test.go
```diff
@@ -212,3 +212,451 @@ func extractSVD(svd *SVD) (s []float64, u, v *Dense) {
...
+				{0, 0, 0},
+				{0, 0, 0},
+			},
+			b: []float64{3, 2},
+		},
+	} {
+		a := NewDense(flatten(test.a))
+		b := NewVecDense(len(test.b), test.b)
+
+		var ans *VecDense
```
vladimir-ch, line 586: `want`

mat/svd_test.go
```diff
@@ -212,3 +212,451 @@ func extractSVD(svd *SVD) (s []float64, u, v *Dense) {
...
+		if test.ans != nil {
```
vladimir-ch, line 587: How does this work? What is being tested if `ans` is nil?

kortschak: Nothing. The SVD factorisation has failed.

mat/svd.go
kortschak: Done

mat/svd.go
kortschak: Done

mat/svd_test.go
kortschak: Done

mat/svd_test.go
kortschak: Done

mat/svd_test.go
kortschak: Done

mat/svd_test.go
kortschak: Done

mat/svd_test.go
kortschak: Done

mat/svd_test.go
kortschak: Done

mat/svd_test.go
kortschak: Done

mat/svd_test.go
kortschak: See https://github.com/gonum/gonum/pull/1316#discussion_r438056289 below.

mat/svd.go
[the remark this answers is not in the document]
vladimir-ch: I need to think about a proper wording. I checked the documentation for `Dense.Solve` and other Solve routines and it's not completely satisfactory because it gives the impression that different problems are being solved. The problem is the difference between 1) solving one problem "minimize |AX-B|_2 where X and B are matrices" and 2) solving k LLS problems at once "minimize |Ax_i-b_i|_2 where x_i and b_i are the columns of nxk matrices X and B". We solve 2) but the docs suggest 1).

vladimir-ch:
And arm64 is failing.

mat/svd_example_test.go
```diff
@@ -0,0 +1,100 @@
...
+
+	// Perform an SVD retaining all singular vectors.
+	var svd mat.SVD
+	ok := svd.Factorize(a, mat.SVDFull)
+	if !ok {
+		log.Fatal("failed to factorize A")
+	}
+
+	// Determine the rank of the A matrix with a near zero condition threshold.
+	var rcond = 1e-15
```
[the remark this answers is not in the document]
kortschak: Done

mat/svd_example_test.go
```diff
@@ -0,0 +1,100 @@
...
```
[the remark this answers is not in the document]
kortschak: Done

mat/svd_test.go
```diff
@@ -212,3 +212,452 @@ func extractSVD(svd *SVD) (s []float64, u, v *Dense) {
...
+				3,
+				2,
+			}, bc: 1,
+		},
+	} {
+		a := NewDense(test.m, test.n, test.a)
+		b := NewDense(test.m, test.bc, test.b)
+
+		var want *Dense
+		if test.want != nil {
```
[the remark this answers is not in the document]
kortschak: I don't understand how that will work; if rank is zero, `SolveTo` will panic.

mat/svd_example_test.go
```diff
@@ -0,0 +1,100 @@
...
+
+import (
+	"fmt"
+	"log"
+
+	"gonum.org/v1/gonum/mat"
+)
+
+func ExampleSVD_SolveTo() {
+	// The system defined described by A is rank deficient.
```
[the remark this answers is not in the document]
kortschak: Done

mat/svd_example_test.go
```diff
@@ -0,0 +1,100 @@
...
+	// Output:
+	// singular values = [2.685 1.526 1.835e-16]
+	// rank = 2
+	// x = ⎡  1.2120643135523468    1.5074674510939299⎤
+	//     ⎢  0.4154007382647741    -0.624498607705372⎥
+	//     ⎣-0.18318444225528013    2.2213341936891244⎦
+}
+
+func ExampleSVD_SolveVecTo() {
+	// The system defined described by A is rank deficient.
```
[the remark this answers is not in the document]
kortschak: Done

kortschak:
Build fixed on arm64 (I need to remember to build these tricky things on my arm64 builder locally before I send them).

kortschak:
*ping*

mat/svd.go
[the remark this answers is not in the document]
vladimir-ch: I think we can postpone this. Once this PR and #1368 are merged, I'll submit changes for the documentation here.

mat/svd_test.go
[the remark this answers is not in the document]
vladimir-ch: As I wrote, this is just nitpicking, let's leave it. I may submit a draft PR later, with code it's easier to explain.

PR #1318, opened 2020-05-05
mat: address two small TODOs

jonreiter:
use CopyVec per a TODO, references long-resolved issue [#167](https://github.com/gonum/gonum/issues/167)
in Cholesky.SymRankOne don't modify receiver if singular


kortschak:
Thanks

Issue #740, opened 2018-12-09
matrix/mat64: add capacity to define alternative formatted styles


kortschak:
Ideally this would be simple like time format definitions. So the current unicode style would be:

```
`[0  1]
⎡2  3⎤
⎢4  5⎥
⎣6  7⎦`
```

a MATLAB input style would be:

```
`[0 1]
[2 3; 4 5; 6 7]`
```

etc.

We need to sort out some heuristics for how this works and be able to document them simply. For example: cell width alignment is not calculated with there is no newline between 3 and 4, and 5 and 6.

btracey:
This comment is maybe outside the scope of the issue, but how would we feel about adding a specific format string for matrices? It would be nice to do something like: fmt.Printf("%0.4m\n", a) where "m" is a tag we define.

PR #1319, opened 2020-05-05
mat: provide mechanism to output Python and MATLAB syntax

kortschak:
~WIP: Needs tests.~

Updates #740.

### DIFF
--- a/mat/format.go
+++ b/mat/format.go
@@ -7,6 +7,7 @@ package mat
 import (
 	"fmt"
 	"strconv"
+	"strings"
 )
 
 // Formatted returns a fmt.Formatter for the matrix m using the given options.
@@ -27,6 +28,8 @@ type formatter struct {
 	margin  int
 	dot     byte
 	squeeze bool
+
+	format func(m Matrix, prefix string, margin int, dot byte, squeeze bool, fs fmt.State, c rune)
 }
 
 // FormatOption is a functional option for matrix formatting.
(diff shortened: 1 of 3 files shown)


kortschak:
Please take a look.

mat/format.go
```diff
@@ -27,6 +28,8 @@ type formatter struct {
 	margin  int
 	dot     byte
 	squeeze bool
+
+	format func(m Matrix, prefix string, margin int, dot byte, squeeze bool, fs fmt.State, c rune)
```
sbinet, line 32: why don't we initialize that field with `format` in the `Formatted` function above?
this would somewhat simplify the logic of the `formatter.Format(...)` method down below.

mat/format.go
```diff
@@ -51,18 +54,35 @@ func DotByte(b byte) FormatOption {
...
 // Format satisfies the fmt.Formatter interface.
 func (f formatter) Format(fs fmt.State, c rune) {
-	if c == 'v' && fs.Flag('#') {
+	if c == 'v' && fs.Flag('#') && f.format == nil {
 		fmt.Fprintf(fs, "%#v", f.matrix)
 		return
 	}
-	format(f.matrix, f.prefix, f.margin, f.dot, f.squeeze, fs, c)
+	if f.format == nil {
+		f.format = format
```
sbinet, line 83: perhaps rename `func format(...)` into `func formatGo(...)` (or `formatGonum`) so everything is symetric?

mat/format.go
```diff
@@ -193,6 +213,252 @@ func format(m Matrix, prefix string, margin int, dot byte, squeeze bool, fs fmt.
...
+		widths = new(uniformWidth)
+	}
+	switch c {
+	case 'v', 'e', 'E', 'f', 'F', 'g', 'G':
+		if c == 'v' {
+			buf, maxWidth = maxCellWidth(m, 'g', printed, prec, widths)
+		} else {
+			buf, maxWidth = maxCellWidth(m, c, printed, prec, widths)
+		}
+	default:
```
sbinet, line 385: it doesn't seem this branch is either exercized in the tests below.
do we want to give it a try?

mat/format.go
```diff
@@ -193,6 +213,252 @@ func format(m Matrix, prefix string, margin int, dot byte, squeeze bool, fs fmt.
...
```
sbinet, line 272: it doesn't seem this branch is either exercized in the tests below.
do we want to give it a try?

mat/format.go
```diff
@@ -193,6 +213,252 @@ func format(m Matrix, prefix string, margin int, dot byte, squeeze bool, fs fmt.
...
+//
+// formatMATLAB will not provide Go syntax output.
+func formatMATLAB(m Matrix, prefix string, _ int, _ byte, squeeze bool, fs fmt.State, c rune) {
+	rows, cols := m.Dims()
+
+	prec, pOk := fs.Precision()
+	width, _ := fs.Width()
+	if !fs.Flag('#') {
+		format := fmtString(fs, c, prec, width)
+		fs.Write([]byte{'['})
```
sbinet, line 230: just a general comment, but perhaps all these little `Write([]byte)` could be accumulated into a local `[]byte` (or `bufio.Writer`) and flushed at every row in the `#` case.

mat/format.go
kortschak: The behaviour of default formatting and the foreign formatting is different with `%#v`. This depends on the `nil` state of the `format` field.

mat/format.go
kortschak: There's not an underlying symmetry, so I don't think this is worth doing.

mat/format.go
kortschak: I'll have a look at doing this.

mat/format.go
kortschak: Done

mat/format.go
kortschak: Done

mat/format.go
kortschak: We don't know what the underlying `io.Writer` is, so I don't think this is worthwhile. It may already be buffering, and if it is not, that doesn't matter either.

kortschak:
PTAL

Issue #1320, opened 2020-05-11
Please add wilcoxon Signed rank test


RamakrishnaChilaka:
I can create a pull request to add wilcoxon signed rank test.

kortschak:
This sounds reasonable. We want to add Mann-Whitney U test which we fork from https://godoc.org/golang.org/x/perf/internal/stats.

PR #1321, opened 2020-05-12
stat: Add stat/test package with Wilcoxon signed-rank test

RamakrishnaChilaka:
Please take a look.
https://github.com/gonum/gonum/issues/1320

@kortschak , Please note that I am interested to add the internal stats package. Please review the code. Thank you.

### DIFF
--- a/stat/test/wilcoxon_hyp.go
+++ b/stat/test/wilcoxon_hyp.go
@@ -51,9 +51,9 @@ func rank(out, in []float64) int {
 		return pairStructArr[i].Value < pairStructArr[j].Value
 	})
 
-	mpOut := make(map[float64]int, len(out))   // tracks the rank
-	mpCount := make(map[float64]int, len(out)) // tracks the count
-	tmpRank := make(map[float64]struct{})      // tracks the tied ranks
+	mpOut := make(map[float64]int, len(out))   // tracks the rank.
+	mpCount := make(map[float64]int, len(out)) // tracks the count.
+	tmpRank := make(map[float64]struct{})      // tracks the tied ranks.
 	rank := 1
 	for i := 0; i < lengthOfIn; i++ {
 		val := pairStructArr[i].Value
@@ -135,7 +135,7 @@ func calculateExactPValue(Wmax float64, NZ int) float64 {
 		rankSum := 0
 		// Generate all possible rank sums.
 		for j := 0; j < NZ; j++ {
-			// (i >> j) & 1 extract i's j-th bit from the right
+			// (i >> j) & 1 extract i's j-th bit from the right.
 			if ((i >> j) & 1) == 1 {
 				rankSum += j + 1
 			}
@@ -151,32 +151,32 @@ func calculateExactPValue(Wmax float64, NZ int) float64 {
 }
 
 func calculateAsymptoticPValue(Wmin float64, NZ, tieAdj int) float64 {
-	// n should be number of non-zero absolute difference pairs
+	// n should be number of non-zero absolute difference pairs.
 	ES := float64(NZ*(NZ+1)) / 4.0
 	VarS := (ES * (float64(2*NZ + 1)) / 6.0) - (float64(tieAdj) / 48.0)
 
-	// - 0.5 is a continuity correction
+	// - 0.5 is a continuity correction.
 	z := (Wmin - ES - 0.5) / math.Sqrt(VarS)
 
 	standardNormal := distuv.UnitNormal
 	return 2 * standardNormal.CDF(z)
 }
 
-// WilcoxonSignedRankTest implements Wilcoxon signed rank test (https://en.wikipedia.org/wiki/Wilcoxon_signed-rank_test)
+// WilcoxonSignedRankTest implements Wilcoxon signed rank test (https://en.wikipedia.org/wiki/Wilcoxon_signed-rank_test, https://data.library.virginia.edu/the-wilcoxon-rank-sum-test/).
 // The Wilcoxon signed-rank test tests the null hypothesis that two paired samples come from the same distribution.
 // In particular, it tests whether the distribution of the differences x - y is symmetric about zero. It is a non-parametric version of the paired T-test.
-// x and y are the first and second set of measurements respectively. Exact P value computation is expensive, it is only available for the sample sizes with size less than 30.
-// WilcoxonSignedRankTest panics when exactPValue is true and len(x) is greater than 30. The two sided p-value for the test.
-func WilcoxonSignedRankTest(x, y []float64, exactPValue bool) float64 {
+// x and y are the first and second set of measurements respectively and the elements shouldn't be same otherwise this function panics. Exact P value computation is expensive, it is only available for sample sizes with size less than 30.
+// WilcoxonSignedRankTest panics when exactPValue is true and len(x) is greater than 30 otherwise it returns two sided p-value.
+func WilcoxonSignedRankTest(x, y []float64, exactPValue bool) (float64, float64) {
 	ensureDataConformance(x, y)
 	Wmax, N, tieAdj := wilCoxonSignedRankTest(x, y)
 	if exactPValue && N > 30 {
 		panic(BadNumberTooLarge)
 	}
 	if exactPValue {
-		return calculateExactPValue(Wmax, N)
+		return calculateExactPValue(Wmax, N), Wmax
 	} else {
 		Wmin := (float64(N*(N+1)) / 2.0) - Wmax
-		return calculateAsymptoticPValue(Wmin, N, tieAdj)
+		return calculateAsymptoticPValue(Wmin, N, tieAdj), Wmin
 	}
 }
--- a/stat/test/wilcoxon_hyp.go
+++ b/stat/test/wilcoxon_hyp.go
@@ -12,15 +12,11 @@ import (
 	"gonum.org/v1/gonum/stat/distuv"
 )
 
+const badLengthMismatch = "test: slice length mismatch"
+
 func ensureDataConformance(x []float64, y []float64) {
-	if len(x) == 0 {
-		panic("x array length is zero")
-	}
-	if len(y) == 0 {
-		panic("y array length is zero")
-	}
-	if len(x) != len(y) {
-		panic("dimension mismatch in ensure data conformance")
+	if len(x) == 0 || len(y) == 0 || len(x) != len(y) {
+		panic(badLengthMismatch)
 	}
 }
 
--- a/stat/test/wilcoxon_hyp.go
+++ b/stat/test/wilcoxon_hyp.go
@@ -1,9 +1,15 @@
-package wilcoxon
+// Copyright ©2019 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package test
 
 import (
-	"gonum.org/v1/gonum/stat/distuv"
 	"math"
 	"sort"
+
+	"gonum.org/v1/gonum/floats"
+	"gonum.org/v1/gonum/stat/distuv"
 )
 
 func ensureDataConformance(x []float64, y []float64) {
@@ -18,23 +24,21 @@ func ensureDataConformance(x []float64, y []float64) {
 	}
 }
 
-func CalculateDifferences(x []float64, y []float64, z []float64) {
+func calculateDifferences(dst []float64, x []float64, y []float64) {
 	ensureDataConformance(x, y)
-	ensureDataConformance(x, z)
+	ensureDataConformance(x, dst)
+	floats.SubTo(dst, x, y)
+}
+
+func calculateAbsDifference(dst []float64, x []float64, y []float64) {
+	ensureDataConformance(x, y)
+	ensureDataConformance(x, dst)
 	for i := 0; i < len(x); i++ {
-		z[i] = x[i] - y[i]
+		dst[i] = math.Abs(y[i] - x[i])
 	}
 }
 
-func CalculateAbsDifference(x []float64, y []float64, z []float64) {
-	ensureDataConformance(x, y)
-	ensureDataConformance(x, z)
-	for i := 0; i < len(x); i++ {
-		z[i] = math.Abs(y[i] - x[i])
-	}
-}
-
-func Rank(in []float64, out []float64) ([]float64, int) {
+func rank(in []float64, out []float64) ([]float64, int) {
 	ensureDataConformance(in, out)
 
 	var tieAdjust = 0
@@ -94,17 +98,17 @@ func Rank(in []float64, out []float64) ([]float64, int) {
 	return out, tieAdjust
 }
 
-func WilCoxonSignedRankTest(x []float64, y []float64) (float64, int, int) {
+func wilCoxonSignedRankTest(x []float64, y []float64) (float64, int, int) {
 	ensureDataConformance(x, y)
 
 	var z = make([]float64, len(x))
 	var absZ = make([]float64, len(x))
-	CalculateDifferences(x, y, z)
-	CalculateAbsDifference(x, y, absZ)
+	calculateDifferences(z, x, y)
+	calculateAbsDifference(absZ, x, y)
 
 	var ranks = make([]float64, len(x))
 	var tieAdj = 0
-	ranks, tieAdj = Rank(absZ, ranks)
+	ranks, tieAdj = rank(absZ, ranks)
 	var tmpLenOfX = len(x)
 
 	var WPlus = 0.0
@@ -120,7 +124,7 @@ func WilCoxonSignedRankTest(x []float64, y []float64) (float64, int, int) {
 	return math.Max(WMinus, WPlus), tmpLenOfX, tieAdj
 }
 
-func CalculateExactPValue(Wmax float64, N int) float64 {
+func calculateExactPValue(Wmax float64, N int) float64 {
 	var m = 1 << N
 
 	largerRankSums := 0
@@ -149,7 +153,7 @@ func CalculateExactPValue(Wmax float64, N int) float64 {
 	return 2 * (float64(largerRankSums) / float64(m))
 }
 
-func CalculateAsymptoticPValue(Wmin float64, NZ int, tieAdj int) float64 {
+func calculateAsymptoticPValue(Wmin float64, NZ int, tieAdj int) float64 {
 
 	// n should be number of non zeros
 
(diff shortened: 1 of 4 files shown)


stat/wilcoxontest/wilcoxon_hyp.go
```diff
@@ -0,0 +1,182 @@
+package wilcoxon
```
kortschak, line 1: Please add a copyright header. You can find these on other Gonum source files.

stat/wilcoxontest/wilcoxon_hyp.go
```diff
@@ -0,0 +1,182 @@
...
+
+import (
+	"gonum.org/v1/gonum/stat/distuv"
+	"math"
+	"sort"
```
kortschak, line 6: Please format code with goimports.

stat/wilcoxontest/wilcoxon_hyp.go
```diff
@@ -0,0 +1,182 @@
...
+	}
+}
+
+func CalculateDifferences(x []float64, y []float64, z []float64) {
+	ensureDataConformance(x, y)
+	ensureDataConformance(x, z)
+	for i := 0; i < len(x); i++ {
+		z[i] = x[i] - y[i]
+	}
+}
```
kortschak, line 27: This is `floats.SubTo`.

stat/wilcoxontest/wilcoxon_hyp.go
```diff
@@ -0,0 +1,182 @@
...
+
+func CalculateAbsDifference(x []float64, y []float64, z []float64) {
+	ensureDataConformance(x, y)
+	ensureDataConformance(x, z)
+	for i := 0; i < len(x); i++ {
+		z[i] = math.Abs(y[i] - x[i])
+	}
+}
```
kortschak, line 35: This is `f64.LinfDist`.

stat/wilcoxontest/wilcoxon_hyp.go
```diff
@@ -0,0 +1,182 @@
...
+
+func Rank(in []float64, out []float64) ([]float64, int) {
```
kortschak, line 37: Please add documentation to exported functions.

stat/wilcoxontest/wilcoxon_hyp_test.go
```diff
@@ -0,0 +1,71 @@
+package wilcoxon
```
kortschak, line 1: Copyright header.

stat/wilcoxontest/wilcoxon_hyp.go
```diff
@@ -0,0 +1,182 @@
...
+	}
+	if len(y) == 0 {
+		panic("y array length is zero")
+	}
+	if len(x) != len(y) {
+		panic("dimension mismatch in ensure data conformance")
+	}
+}
+
+func CalculateDifferences(x []float64, y []float64, z []float64) {
```
sbinet, line 21: I'd probably rename `z` into `dst` and move it as 1st argument.
(and document that exported function)

is it really needed to be exported?

stat/wilcoxontest/wilcoxon_hyp.go
RamakrishnaChilaka: Thanks for review @kortschak , I don't think f64.LinfDist is correct here, LinfDist is calculating the maximum abs difference. Do you want me to call that function with the single element length array (len(x)) times ?. Thanks @kortschak

stat/wilcoxontest/wilcoxon_hyp.go
RamakrishnaChilaka: thanks @sbinet , for reviewing the code, I don't think this function needs to be exported and a bunch of functions also don't need to be exported, I am making them private in the upcoming PR, Will make the suggested changes wrt to naming and positions of the arguments. Thanks @sbinet

RamakrishnaChilaka:
@sbinet , @kortschak , Thanks for reviewing the code again, I have made the changes requested above (except for one) and pushed the code. Please review the code whenever you guys find free time. Thanks @sbinet @kortschak .

stat/test/wilcoxon_hyp.go
```diff
@@ -0,0 +1,190 @@
...
+
+func calculateAbsDifference(dst []float64, x []float64, y []float64) {
+	ensureDataConformance(x, y)
+	ensureDataConformance(x, dst)
+	for i := 0; i < len(x); i++ {
+		dst[i] = math.Abs(y[i] - x[i])
+	}
+}
+
+func rank(in []float64, out []float64) ([]float64, int) {
```
sbinet, line 37:
```suggestion
func rank(in, out []float64) ([]float64, int) {
```

also, we should probably consider moving `out` to the 1st arg (to harmonize with other signatures)

stat/test/wilcoxon_hyp.go
```diff
@@ -0,0 +1,190 @@
...
+		Value    float64
+	}
+	var lengthOfIn = len(in)
+	var pairStructArr = make([]pairStruct, lengthOfIn)
+	for i := 0; i < lengthOfIn; i++ {
+		pairStructArr[i] = pairStruct{
+			Position: i,
+			Value:    in[i],
+		}
+	}
```
sbinet, line 52:
```suggestion
	pairs := make([]pairStruct, len(in))
	for i, v := range in {
		pairs[i] = pairStruct{
			Position: i,
			Value:    v,
		}
	}
```

stat/test/wilcoxon_hyp.go
```diff
@@ -0,0 +1,190 @@
...
+	for val := range tmpRank {
+		tieAdjust += (mpCount[val] * mpCount[val] * mpCount[val]) - (mpCount[val])
+
+	}
+	return out, tieAdjust
+}
+
+func wilCoxonSignedRankTest(x []float64, y []float64) (float64, int, int) {
```
sbinet, line 97:
```suggestion
func wilCoxonSignedRankTest(x, y []float64) (float64, int, int) {
```

stat/test/wilcoxon_hyp.go
```diff
@@ -0,0 +1,190 @@
...
+			}
+		}
+
+		if float64(rankSum) >= Wmax {
+			largerRankSums++
+		}
+	}
+
+	/*
+	 * largerRankSums / m gives the one-sided p-value, so it's multiplied
```
sbinet, line 146: please use `//` for comments.

stat/test/wilcoxon_hyp.go
```diff
@@ -0,0 +1,190 @@
...
+	// - 0.5 is a continuity correction
+	z := (Wmin - ES - 0.5) / math.Sqrt(VarS)
+
+	standardNormal := distuv.UnitNormal
+
+	return 2 * standardNormal.CDF(z)
+}
+
+
+// Perform a Wilcoxon signed rank test of the null hypothesis that the distribution of `x`
```
sbinet, line 171: please use the recommanded style for documentation:
- https://github.com/gonum/gonum/blob/master/CONTRIBUTING.md
- https://golang.org/doc/effective_go.html#commentary

stat/test/wilcoxon_hyp_test.go
```diff
@@ -0,0 +1,50 @@
+// Copyright ©2019 The Gonum Authors. All rights reserved.
```
sbinet, line 1:
```suggestion
// Copyright ©2020 The Gonum Authors. All rights reserved.
```

stat/test/wilcoxon_hyp.go
```diff
@@ -0,0 +1,190 @@
+// Copyright ©2019 The Gonum Authors. All rights reserved.
```
sbinet, line 1:
```suggestion
// Copyright ©2020 The Gonum Authors. All rights reserved.
```

stat/test/wilcoxon_hyp.go
```diff
@@ -0,0 +1,190 @@
...
+	"math"
+	"sort"
+
+	"gonum.org/v1/gonum/floats"
+	"gonum.org/v1/gonum/stat/distuv"
+)
+
+func ensureDataConformance(x []float64, y []float64) {
+	if len(x) == 0 {
+		panic("x array length is zero")
```
sbinet: we don't call them arrays in Go (as this has a different meaning).
we call them slices.

also, `ensureDataConformance` isn't always called with `x` and `y` arguments so the panic message could be misleading.
(I think we also try to prefix panic messages with the package name: `"test: bla-bla"`)

stat/wilcoxontest/wilcoxon_hyp.go
kortschak: You're correct.

stat/test/wilcoxon_hyp.go
kortschak: In general, we do these inline. They are more informative that way.

stat/test/wilcoxon_hyp.go
```diff
@@ -0,0 +1,190 @@
...
+	}
+	if len(y) == 0 {
+		panic("y array length is zero")
+	}
+	if len(x) != len(y) {
+		panic("dimension mismatch in ensure data conformance")
+	}
+}
+
+func calculateDifferences(dst []float64, x []float64, y []float64) {
```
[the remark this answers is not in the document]
kortschak: Please just delete this function and call `floats.SubTo`.

RamakrishnaChilaka:
@kortschak @sbinet , thanks for a detailed review, it helped me understand the code style of the gonum package. I have made very significant changes, please take a pull and review it when ever you guys are free. Thanks.

RamakrishnaChilaka:
@kortschak @sbinet , I cannot figure out why travis-ci is failing for arm64 platform, If you can either help me or point in me in the correct direction. As the master build is also failing. https://api.travis-ci.org/gonum/gonum.svg?branch=master

romanwerpachowski (changes requested):

stat/test/wilcoxon_hyp_test.go
```diff
@@ -0,0 +1,50 @@
...
+				y:           []float64{3.5, 5.7, 2.9, 2.4, 9.9, 3.3, 16.7, 6.0, 3.8, 4.0, 9.1, 20.9},
+				exactPValue: false,
+			},
+			want: 0.010757133119789652,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if got := WilcoxonSignedRankTest(tt.args.x, tt.args.y, tt.args.exactPValue); !strings.EqualFold(fmt.Sprintf("%0.5f", got), fmt.Sprintf("%0.5f", tt.want)) {
+				t.Errorf("WilcoxonSignedRankTest() = %v, want %v", got, tt.want)
```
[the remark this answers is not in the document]
RamakrishnaChilaka: Not sure how to add `panic` test case. Is there any standard gonum test case i can refer to.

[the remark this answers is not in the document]
romanwerpachowski: See here, for example: https://github.com/gonum/gonum/blob/master/stat/distuv/bernoulli_test.go#L130

`panics` is a short helper function which we reimplement in every package separately:

```
func panics(fun func()) (b bool) {
	defer func() {
		err := recover()
		if err != nil {
			b = true
		}
	}()
	fun()
	return
}
```

stat/test/wilcoxon_hyp.go
```diff
@@ -0,0 +1,182 @@
...
+
+	standardNormal := distuv.UnitNormal
+	return 2 * standardNormal.CDF(z)
+}
+
+// WilcoxonSignedRankTest implements Wilcoxon signed rank test (https://en.wikipedia.org/wiki/Wilcoxon_signed-rank_test)
+// The Wilcoxon signed-rank test tests the null hypothesis that two paired samples come from the same distribution.
+// In particular, it tests whether the distribution of the differences x - y is symmetric about zero. It is a non-parametric version of the paired T-test.
+// x and y are the first and second set of measurements respectively. Exact P value computation is expensive, it is only available for the sample sizes with size less than 30.
+// WilcoxonSignedRankTest panics when exactPValue is true and len(x) is greater than 30. The two sided p-value for the test.
```
romanwerpachowski, line 169: Please document also other cases when it panics (`x` having different size than `y`, all pairs being equal, etc).

stat/test/wilcoxon_hyp.go
```diff
@@ -0,0 +1,182 @@
...
+func WilcoxonSignedRankTest(x, y []float64, exactPValue bool) float64 {
+	ensureDataConformance(x, y)
+	Wmax, N, tieAdj := wilCoxonSignedRankTest(x, y)
+	if exactPValue && N > 30 {
+		panic(BadNumberTooLarge)
+	}
+	if exactPValue {
```
romanwerpachowski, line 176: Would you consider also returning the test statistic? It is something the users will often also want to report when analysing data.

romanwerpachowski (changes requested):
Thanks! We're almost there :)

stat/test/wilcoxon_hyp.go
```diff
@@ -0,0 +1,182 @@
...
+			largerRankSums++
+		}
+	}
+
+	// largerRankSums / m gives the one-sided p-value, so it's multiplied
+	// with 2 to get the two-sided p-value.
+	return 2 * (float64(largerRankSums) / float64(m))
+}
+
+func calculateAsymptoticPValue(Wmin float64, NZ, tieAdj int) float64 {
```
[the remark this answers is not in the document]
RamakrishnaChilaka: I choose 30 because it seemed to give good balance between runtime of the function (because it is a heavy computation to calculate exact p value everytime). This is my reference, https://web.archive.org/web/20070322182510/http://www.fon.hum.uva.nl/Service/Statistics/Signed_Rank_Test.html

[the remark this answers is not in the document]
romanwerpachowski: Thanks. Would you mind adding this link to the documentation?

stat/test/wilcoxon_hyp.go
```diff
@@ -0,0 +1,182 @@
...
+	VarS := (ES * (float64(2*NZ + 1)) / 6.0) - (float64(tieAdj) / 48.0)
+
+	// - 0.5 is a continuity correction.
+	z := (Wmin - ES - 0.5) / math.Sqrt(VarS)
+
+	standardNormal := distuv.UnitNormal
+	return 2 * standardNormal.CDF(z)
+}
+
+// WilcoxonSignedRankTest implements Wilcoxon signed rank test (https://en.wikipedia.org/wiki/Wilcoxon_signed-rank_test, https://data.library.virginia.edu/the-wilcoxon-rank-sum-test/).
```
romanwerpachowski, line 165: Nitpicking: it would make reviewing the PR in GitHub easier if you limited the line length to, say, 80-100 characters.

romanwerpachowski, line 165: I'm confused: this is not the Web Archive link you have posted in the comment earlier?

stat/test/wilcoxon_hyp.go
```diff
@@ -0,0 +1,182 @@
...
+// The Wilcoxon signed-rank test tests the null hypothesis that two paired samples come from the same distribution.
+// In particular, it tests whether the distribution of the differences x - y is symmetric about zero. It is a non-parametric version of the paired T-test.
```
romanwerpachowski, line 167: How about:
```
// In particular, it tests whether the distribution of the differences x - y is symmetric about zero. It is a non-parametric version of the paired T-test.
// x and y are the first and second set of paired measurements, respectively. Exact P value computation is expensive, it is only available for sample sizes less than 30.
// WilcoxonSignedRankTest returns two sided p-value. It panics when exactPValue is true and len(x) is greater than 30, `len(x) != len(y)`, `len(x) == 0` or when `x[i] == y[i]` for all `i`.
```
?

stat/test/wilcoxon_hyp.go
```diff
@@ -151,32 +151,32 @@ func calculateExactPValue(Wmax float64, NZ int) float64 {
...
 	if exactPValue && N > 30 {
 		panic(BadNumberTooLarge)
 	}
 	if exactPValue {
-		return calculateExactPValue(Wmax, N)
+		return calculateExactPValue(Wmax, N), Wmax
 	} else {
 		Wmin := (float64(N*(N+1)) / 2.0) - Wmax
-		return calculateAsymptoticPValue(Wmin, N, tieAdj)
+		return calculateAsymptoticPValue(Wmin, N, tieAdj), Wmin
```
romanwerpachowski, line 180: Shouldn't we return the same test statistic in both cases? The difference is only in how we calculate the p-value.

stat/test/wilcoxon_hyp_test.go
```diff
@@ -6,45 +6,39 @@ package test
...
-		t.Run(tt.name, func(t *testing.T) {
-			if got := WilcoxonSignedRankTest(tt.x, tt.y, tt.exactPValue); !strings.EqualFold(fmt.Sprintf("%0.5f", got), fmt.Sprintf("%0.5f", tt.want)) {
-				t.Errorf("WilcoxonSignedRankTest() = %v, want %v", got, tt.want)
+	for index, tt := range tests {
+		t.Run(strconv.Itoa(index), func(t *testing.T) {
+			gotPValue, gotTestStatistic  := WilcoxonSignedRankTest(tt.x, tt.y, tt.exactPValue)
+			if !strings.EqualFold(fmt.Sprintf("%0.5f", gotPValue), fmt.Sprintf("%0.5f", tt.pValue)) || !strings.EqualFold(fmt.Sprintf("%0.5f", gotTestStatistic), fmt.Sprintf("%0.5f", tt.testStatistic)){
+				t.Errorf("WilcoxonSignedRankTest() = %v, %v want %v, %v", gotPValue, gotTestStatistic, tt.pValue, tt.testStatistic)
 			}
 		})
```
romanwerpachowski, line 43: To be consistent with the rest of the library tests, and avoid using string formatting routines for numerical tests, please test the returned values using the functions from our `floats/scalar` package. See an example here: https://github.com/gonum/gonum/blob/master/stat/distmv/studentst_test.go#L81

stat/test/wilcoxon_hyp_test.go
```diff
@@ -6,45 +6,39 @@ package test
 
 import (
 	"fmt"
+	"strconv"
 	"strings"
 	"testing"
 )
 
 func TestWilcoxonSignedRankTest(t *testing.T) {
```
romanwerpachowski, line 14: We always (well, almost always) add `t.Parallel()` at the beginning of a unit test.

stat/test/wilcoxon_hyp_test.go
```diff
@@ -6,45 +6,39 @@ package test
...
+			pValue:        0.010757133119789652,
+			testStatistic: 7,
 		},
 	}
-	for _, tt := range tests {
-		t.Run(tt.name, func(t *testing.T) {
-			if got := WilcoxonSignedRankTest(tt.x, tt.y, tt.exactPValue); !strings.EqualFold(fmt.Sprintf("%0.5f", got), fmt.Sprintf("%0.5f", tt.want)) {
-				t.Errorf("WilcoxonSignedRankTest() = %v, want %v", got, tt.want)
+	for index, tt := range tests {
+		t.Run(strconv.Itoa(index), func(t *testing.T) {
```
romanwerpachowski, line 38: Why not run the tested function directly? It's good to keep the test code simple.

romanwerpachowski (changes requested):

romanwerpachowski:
I edited the PR name to bring it line with Gonum convention.

Issue #1322, opened 2020-05-16
amos: investigate failing fortran comparison tests


kortschak:
There are [tests in amos](https://github.com/gonum/gonum/blob/519d2b5f9afc4fd66cd9952bf5a0b0ae4e6c86e4/mathext/internal/amos/amos_fortran_test.go) that compare the results of the [Gotran code](https://github.com/gonum/gonum/blob/519d2b5f9afc4fd66cd9952bf5a0b0ae4e6c86e4/mathext/internal/amos/origcode_test.go) (Fortran code written in Go) with the Fortran code in [amoslib](https://github.com/gonum/gonum/tree/519d2b5f9afc4fd66cd9952bf5a0b0ae4e6c86e4/mathext/internal/amos). Currently some of these tests fail. The failing are only tested when using the `fortran` tag and when the `-failing` flag is passed to `go test`.

It would be good to understand why these tests fail, though the public facing `mathext.Airy` function passes its [tests](https://github.com/gonum/gonum/blob/519d2b5f9afc4fd66cd9952bf5a0b0ae4e6c86e4/mathext/airy_test.go) when it makes use of the amos package functions.

See #1275.

PR #1323, opened 2020-05-16
amos: add bug note to failing Fortran/Go comparison tests

kortschak:
Please take a look.

Updates #1322.


kortschak:
#1324 will fix the failures on Go master. I am investigating the failure on arm64 Go1.13.x.

PR #1326, opened 2020-05-17
travis: reset tree before the generate check

### LINKED ISSUE #1325: travis: travis_install_go_dependencies causes a go.mod write during build on arm64 for go1.13.x
This is recent (I did not notice it in versions of before go1.13.11), irritating (only one arch) and likely transient (it does not happen for go1.14.x). I am unable to replicate the issue on my local arm64 builder.

See https://travis-ci.org/github/gonum/gonum/jobs/687950066 for the issue behaviour.

kortschak:
Please take a look.

Updates #1325.

### DIFF
--- a/.travis/script.d/check-generate.sh
+++ b/.travis/script.d/check-generate.sh
@@ -1,5 +1,9 @@
 #!/bin/bash
 
+# Reset the tree to the current commit to handle
+# any writes during the build.
+git reset --hard
+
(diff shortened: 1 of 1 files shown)


kortschak:
I would like to reorder the check, but this breaks the build in ways that it shouldn't (unfixing this fix). I will revisit this when Go1.13 is no longer supported.

PR #1328, opened 2020-05-19
ci: make golangci-linting a github action

kortschak:
When golangci.com went away we lost annotation within the PR, depending on golangci-lint to find errors in the travis build log. This is an experiment to use GitHub actions to do essentially the same thing.

Currently, it makes annotations on the code changes rather than sending comments to the PR conversation (which I would prefer), but that is a feature request on the action. If we don't like this was can revert it, but I'd like to try it out for a bit; it is running the action on this PR, though nothing should be found since there are no Go changes here.

Please take a look.

### DIFF
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -1,0 +1,20 @@
+name: golangci-lint
+on:
+  push:
+    tags:
+      - v*
+    branches:
+      - master
+  pull_request:
+jobs:
+  golangci:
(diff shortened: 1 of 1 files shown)


kortschak:
Waiting on an issue for the action to figure out how to make this only lint new changes.

jirfag:
@kortschak we've supported showing only new issues in the github action

kortschak:
This is ready now. PTAL

PR #1330, opened 2020-05-22
ci: don't lint on merge into master

kortschak:
This otherwise causes a failure because of the logic in new issues. We don't need the comments and the travis-based linting already catches things that this would.

Please take a look.


kortschak:
Eventually, I agree. There are linting and vetting errors that should not need annotations. When those are fixed, I'm happy to remove that. As it is, we have very few linting failures.

PR #1331, opened 2020-05-22
Interp

romanwerpachowski:
interp: This change modifies Gonum by adding a new interp/ package with interpolation algorithms.

I implemented three basic 1D interpolators so far:
- constant
- piecewise linear
- piecewise constant

Submitting a PR now to start the review process and adjust the code design & style to the project requirements, before I go down some rabbit hole ;-)

Cheers,
Roman

### DIFF
--- a/interp/doc.go
+++ b/interp/doc.go
@@ -2,10 +2,5 @@
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
-// Package interp implements algorithms for interpolating sampled values.
-//
-// The routines provided by the interp package handle the following 1-dimensional interpolation algorithms:
-//  - constant
-//  - linear
-//  - piecewise constant
+// Package interp implements 1-dimensional algorithms for interpolating values.
 package interp // import "gonum.org/v1/gonum/interp"
--- a/interp/doc.go
+++ b/interp/doc.go
@@ -1,4 +1,4 @@
-// Copyright ©2015 The Gonum Authors. All rights reserved.
+// Copyright ©2020 The Gonum Authors. All rights reserved.
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
--- a/interp/doc.go
+++ b/interp/doc.go
@@ -4,8 +4,8 @@
 
 // Package interp implements algorithms for interpolating sampled values.
 //
-// Contains the following 1D interpolation algorithms:
-// 1. constant,
-// 2. linear,
-// 3. piecewise constant.
+// The routines provided by the interp package handle the following 1-dimensional interpolation algorithms:
+//  - constant
+//  - linear
+//  - piecewise constant
 package interp // import "gonum.org/v1/gonum/interp"
--- a/interp/doc.go
+++ b/interp/doc.go
@@ -1,0 +1,6 @@
+// Copyright ©2015 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Package interp implements algorithms for interpolating sampled values.
+package interp // import "gonum.org/v1/gonum/interp"
--- a/interp/doc.go
+++ b/interp/doc.go
@@ -1,7 +1,7 @@
 // Copyright ©2015 The Gonum Authors. All rights reserved.
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
-//
+
 // Package interp implements algorithms for interpolating sampled values.
 //
 // Contains the following algorithms:
--- a/interp/doc.go
+++ b/interp/doc.go
@@ -1,6 +1,10 @@
 // Copyright ©2015 The Gonum Authors. All rights reserved.
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
-
+//
 // Package interp implements algorithms for interpolating sampled values.
+//
+// Contains the following algorithms:
+// 1. constant 1D interpolator
+// 2. linear 1D interpolator
 package interp // import "gonum.org/v1/gonum/interp"
--- a/interp/doc.go
+++ b/interp/doc.go
@@ -3,4 +3,7 @@
 // license that can be found in the LICENSE file.
 
 // Package interp implements 1-dimensional algorithms for interpolating values.
+// Outside of the interpolation interval determined by the interpolated data,
+// the returned value is undefined (but we do our best to return something
+// reasonable).
 package interp // import "gonum.org/v1/gonum/interp"
--- a/interp/doc.go
+++ b/interp/doc.go
@@ -4,7 +4,8 @@
 
 // Package interp implements algorithms for interpolating sampled values.
 //
-// Contains the following algorithms:
-// 1. constant 1D interpolator
-// 2. linear 1D interpolator
+// Contains the following 1D interpolation algorithms:
+// 1. constant,
+// 2. linear,
+// 3. piecewise constant.
 package interp // import "gonum.org/v1/gonum/interp"
--- a/interp/interp.go
+++ b/interp/interp.go
@@ -12,10 +12,10 @@ import (
 	"gonum.org/v1/gonum/spatial/r1"
 )
 
-// Interpolator1D represents a 1-dimensional interpolator.
+// Interpolator represents a 1-dimensional interpolator.
 // It interpolates a function over a defined closed interval range
 // (which can be infinite on both sides).
-type Interpolator1D interface {
+type Interpolator interface {
 	// Interval returns the valid closed interval for interpolation.
 	Interval() r1.Interval
 
@@ -24,25 +24,25 @@ type Interpolator1D interface {
 	ValueAt(float64) float64
 }
 
-// Constant1D is a constant 1D interpolator,
-// It is defined over the range [-infinity, infinity].
-type Constant1D struct {
+// Constant is an interpolator that predicts the same constant value
+// over the range [-infinity, infinity].
+type Constant struct {
 	// Constant Y value.
 	Value float64
 }
 
 // Interval returns the valid closed interval for interpolation.
-func (ci Constant1D) Interval() r1.Interval {
+func (ci Constant) Interval() r1.Interval {
 	return r1.Interval{Min: math.Inf(-1), Max: math.Inf(1)}
 }
 
 // ValueAt returns the interpolation value at x.
-func (ci Constant1D) ValueAt(x float64) float64 {
+func (ci Constant) ValueAt(x float64) float64 {
 	return ci.Value
 }
 
-// Linear1D is a piecewise linear 1-dimensional interpolator.
-type Linear1D struct {
+// PiecewiseLinear is a piecewise linear 1-dimensional interpolator.
+type PiecewiseLinear struct {
 	// Interpolated X values.
 	xs []float64
 
@@ -53,11 +53,11 @@ type Linear1D struct {
 	slopes []float64
 }
 
-// NewLinear1D creates a new linear 1-dimensional interpolator.
+// NewPiecewiseLinear creates a new linear 1-dimensional interpolator.
 // xs and ys should contain the X and Y values of interpolated nodes, respectively.
-// NewLinear1D panics if len(xs) < 2, elements of xs are not strictly increasing or
+// NewPiecewiseLinear panics if len(xs) < 2, elements of xs are not strictly increasing or
 // len(xs) != len(ys).
-func NewLinear1D(xs []float64, ys []float64) *Linear1D {
+func NewPiecewiseLinear(xs []float64, ys []float64) *PiecewiseLinear {
 	n := len(xs)
 	if len(ys) != n {
 		panic("interp: xs and ys have different lengths")
@@ -74,16 +74,16 @@ func NewLinear1D(xs []float64, ys []float64) *Linear1D {
 		}
 		slopes[i] = (ys[i+1] - ys[i]) / dx
 	}
-	return &Linear1D{xs, ys, slopes}
+	return &PiecewiseLinear{xs, ys, slopes}
 }
 
 // Interval returns the valid closed interval for interpolation.
-func (li Linear1D) Interval() r1.Interval {
+func (li PiecewiseLinear) Interval() r1.Interval {
 	return r1.Interval{Min: li.xs[0], Max: li.xs[len(li.xs)-1]}
 }
 
 // ValueAt returns the interpolation value at x.
-func (li Linear1D) ValueAt(x float64) float64 {
+func (li PiecewiseLinear) ValueAt(x float64) float64 {
 	i, xI := findSegment(li.xs, x)
 	if x == xI {
 		return li.ys[i]
@@ -92,8 +92,8 @@ func (li Linear1D) ValueAt(x float64) float64 {
 	return li.ys[i] + li.slopes[i]*(x-xI)
 }
 
-// PiecewiseConstant1D is a piecewise constant 1-dimensional interpolator.
-type PiecewiseConstant1D struct {
+// PiecewiseConstant is a piecewise constant 1-dimensional interpolator.
+type PiecewiseConstant struct {
 	// Interpolated X values.
 	xs []float64
 
@@ -104,13 +104,13 @@ type PiecewiseConstant1D struct {
 	leftContinuous bool
 }
 
-// NewPiecewiseConstant1D creates a new piecewise constant 1-dimensional interpolator.
+// NewPiecewiseConstant creates a new piecewise constant 1-dimensional interpolator.
 // xs and ys should contain the X and Y values of interpolated nodes, respectively.
 // If leftContinuous == true, then y(xs[i]) == y(xs[i] - eps) for small eps > 0. Otherwise,
 // y(xs[i]) == y(xs[i] + eps).
-// NewPiecewiseConstant1D panics if len(xs) < 2, elements of xs are not strictly increasing or
+// NewPiecewiseConstant panics if len(xs) < 2, elements of xs are not strictly increasing or
 // len(xs) != len(ys).
-func NewPiecewiseConstant1D(xs []float64, ys []float64, leftContinuous bool) *PiecewiseConstant1D {
+func NewPiecewiseConstant(xs []float64, ys []float64, leftContinuous bool) *PiecewiseConstant {
 	n := len(xs)
 	if len(ys) != n {
 		panic("interp: xs and ys have different lengths")
@@ -123,16 +123,16 @@ func NewPiecewiseConstant1D(xs []float64, ys []float64, leftContinuous bool) *Pi
 			panic("interp: xs values not strictly increasing")
 		}
 	}
-	return &PiecewiseConstant1D{xs, ys, leftContinuous}
+	return &PiecewiseConstant{xs, ys, leftContinuous}
 }
 
 // Interval returns the valid closed interval for interpolation.
-func (pci PiecewiseConstant1D) Interval() r1.Interval {
+func (pci PiecewiseConstant) Interval() r1.Interval {
 	return r1.Interval{Min: pci.xs[0], Max: pci.xs[len(pci.xs)-1]}
 }
 
 // ValueAt returns the interpolation value at x.
-func (pci PiecewiseConstant1D) ValueAt(x float64) float64 {
+func (pci PiecewiseConstant) ValueAt(x float64) float64 {
 	i, xI := findSegment(pci.xs, x)
 	if x == xI {
 		return pci.ys[i]
--- a/interp/interp.go
+++ b/interp/interp.go
@@ -76,6 +76,20 @@ type LinearInterpolator1D struct {
 	slopes []float64
 }
 
+// NewLinearInterpolator1D creates a new linear 1D interpolator.
+// xs and ys should contain the X and Y values of interpolated nodes, respectively.
+// Panics if len(xs) < 2, elements of xs are not strictly increasing or
+// len(xs) != len(ys).
+func NewLinearInterpolator1D(xs []float64, ys []float64) *LinearInterpolator1D {
+	validateXsAndYs(xs, ys)
+	m := len(xs) - 1
+	slopes := make([]float64, m)
+	for i := 0; i < m; i++ {
+		slopes[i] = (ys[i+1] - ys[i]) / (xs[i+1] - xs[i])
+	}
+	return &LinearInterpolator1D{xs, ys, slopes}
+}
+
 // Interval implements Interpolator1D.Interval.
 func (li LinearInterpolator1D) Interval() r1.Interval {
 	return r1.Interval{Min: li.xs[0], Max: li.xs[len(li.xs)-1]}
@@ -108,20 +122,6 @@ func validateXsAndYs(xs []float64, ys []float64) {
 	}
 }
 
-// NewLinearInterpolator1D creates a new linear 1D interpolator.
-// xs and ys should contain the X and Y values of interpolated nodes, respectively.
-// Panics if len(xs) < 2, elements of xs are not strictly increasing or
-// len(xs) != len(ys).
-func NewLinearInterpolator1D(xs []float64, ys []float64) *LinearInterpolator1D {
-	validateXsAndYs(xs, ys)
-	m := len(xs) - 1
-	slopes := make([]float64, m)
-	for i := 0; i < m; i++ {
-		slopes[i] = (ys[i+1] - ys[i]) / (xs[i+1] - xs[i])
-	}
-	return &LinearInterpolator1D{xs, ys, slopes}
-}
-
 // PiecewiseConstInterpolator1D is a piecewise constant 1D interpolator.
 // It is defined over [xs[0], xs[len(xs)-1]].
 // If leftContinuous == true, then y(xs[i]) == y(xs[i] - eps) for small
@@ -135,6 +135,15 @@ type PiecewiseConstInterpolator1D struct {
 
 	// Whether the interpolated function is left- or right-continuous.
 	leftContinuous bool
+}
+
+// NewPiecewiseConstInterpolator1D creates a new piecewise constant 1D interpolator.
+// xs and ys should contain the X and Y values of interpolated nodes, respectively.
+// Panics if len(xs) < 2, elements of xs are not strictly increasing or
+// len(xs) != len(ys).
+func NewPiecewiseConstInterpolator1D(xs []float64, ys []float64, leftContinuous bool) *PiecewiseConstInterpolator1D {
+	validateXsAndYs(xs, ys)
+	return &PiecewiseConstInterpolator1D{xs, ys, leftContinuous}
 }
 
 // Interval implements Interpolator1D.Interval.
@@ -153,13 +162,4 @@ func (pci PiecewiseConstInterpolator1D) Eval(x float64) float64 {
 		return pci.ys[i+1]
 	}
 	return pci.ys[i]
-}
-
-// NewPiecewiseConstInterpolator1D creates a new piecewise constant 1D interpolator.
-// xs and ys should contain the X and Y values of interpolated nodes, respectively.
-// Panics if len(xs) < 2, elements of xs are not strictly increasing or
-// len(xs) != len(ys).
-func NewPiecewiseConstInterpolator1D(xs []float64, ys []float64, leftContinuous bool) *PiecewiseConstInterpolator1D {
-	validateXsAndYs(xs, ys)
-	return &PiecewiseConstInterpolator1D{xs, ys, leftContinuous}
 }
--- a/interp/interp.go
+++ b/interp/interp.go
@@ -68,7 +68,7 @@ func find_segment(xs []float64, x float64) (int, float64) {
 	// Find minimum i s.t. xs[i] >= x, or len(xs) if not found.
 	n := len(xs)
 	i := sort.Search(n, func(i int) bool { return xs[i] > x })
-	if i < len(xs) {
+	if i < n {
 		if i == 0 {
 			// x < begin()
 			panic(fmt.Sprintf("interp: eval() argument %g outside range", x))
--- a/interp/interp.go
+++ b/interp/interp.go
@@ -1,3 +1,7 @@
+// Copyright ©2020 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
 package interp
 
 import (
--- a/interp/interp.go
+++ b/interp/interp.go
@@ -12,7 +12,7 @@ import (
 // interpolation and extrapolation.
 type Predictor interface {
 	// Predict returns the predicted value at x.
-	Predict(float64) float64
+	Predict(float64 x) float64
 }
 
 // Fitter fits a predictor to data.
--- a/interp/interp.go
+++ b/interp/interp.go
@@ -146,8 +146,9 @@ func (pc PiecewiseConstant) Predict(x float64) float64 {
 	return pc.ys[i+1]
 }
 
-// findSegment returns 0 <= i < len(xs) such that xs[i] <= x < xs[i + 1], or -1
-// if no such i is found. It assumes that len(xs) >= 2 without checking.
+// findSegment returns 0 <= i < len(xs) such that xs[i] <= x < xs[i + 1], where xs[len(xs)]
+// is assumed to be +Inf. If no such i is found, it returns -1. It assumes that len(xs) >= 2
+// without checking.
 func findSegment(xs []float64, x float64) int {
 	return sort.Search(len(xs), func(i int) bool { return xs[i] > x }) - 1
 }
--- a/interp/interp.go
+++ b/interp/interp.go
@@ -45,7 +45,7 @@ func NewConstInterpolator1D(value float64) *ConstInterpolator1D {
 	return &ConstInterpolator1D{value}
 }
 
-// LinearInterpolator1D is a linear 1D interpolator. Defined over the range [xs[0], xs[len(xs) - 1]].
+// LinearInterpolator1D is a linear 1D interpolator, defined over the range [xs[0], xs[len(xs) - 1]].
 type LinearInterpolator1D struct {
 	xs     []float64
 	ys     []float64
@@ -60,9 +60,8 @@ func (i1d LinearInterpolator1D) end() float64 {
 	return i1d.xs[len(i1d.xs)-1]
 }
 
-// Returns a tuple of: (i such that xs[i] <= x < xs[i + 1], xs[i]).
-// Assumes len(xs) >= 2.
-// Panics if such i is not found.
+// findSegment returns a tuple of: (i such that xs[i] <= x < xs[i + 1], xs[i]),
+// or panics if such i is not found. Assumes that len(xs) >= 2 without checking.
 func findSegment(xs []float64, x float64) (int, float64) {
 	// Find minimum i s.t. xs[i] >= x, or len(xs) if not found.
 	n := len(xs)
@@ -92,6 +91,7 @@ func (i1d LinearInterpolator1D) eval(x float64) float64 {
 }
 
 // NewLinearInterpolator1D creates a new linear 1D interpolator.
+// xs and ys should contain the X and Y values of interpolated nodes, respectively.
 // Panics if len(xs) < 2, elements of xs are not strictly increasing or
 // len(xs) != len(ys).
 func NewLinearInterpolator1D(xs []float64, ys []float64) *LinearInterpolator1D {
--- a/interp/interp.go
+++ b/interp/interp.go
@@ -6,11 +6,10 @@ import (
 	"sort"
 )
 
-// 1D interpolator interface.
+// Interpolator1D represents a 1D interpolator interface.
 //
 // Interpolates a sequence of (x, y) pairs on a range [begin(), end()].
-// Both begin() and end() can be infinite, but begin() must be lower
-// than end().
+// Both begin() and end() can be infinite, but begin() must be lower than end().
 type Interpolator1D interface {
 	// Returns the lowest allowed argument for evaluate().
 	begin() float64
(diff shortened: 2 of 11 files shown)


kortschak:
@btracey Do you want to take a look at this?

interp/interp.go
```diff
@@ -0,0 +1,176 @@
...
+type Interpolator1D interface {
+	// Returns the lowest allowed argument for evaluate().
+	begin() float64
+
+	// Returns the highest allowed argument for evaluate().
+	end() float64
+
+	// Evaluates interpolated sequence at x in [begin(), end()].
+	// Panics if the argument is outside this range.
+	eval(float64) float64
```
sbinet, line 25: hum... why aren't any of these methods exported?
shouldn't `eval` be at least exported?

interp/interp.go
```diff
@@ -0,0 +1,176 @@
...
+	"math"
+	"sort"
+)
+
+// Interpolator1D represents a 1D interpolator.
+// It interpolates a y(x) function based on some data over a range [begin(), end()].
+// Both begin() and end() can be infinite, but begin() must be lower than end().
+type Interpolator1D interface {
+	// Returns the lowest allowed argument for evaluate().
+	begin() float64
```
sbinet, line 18: perhaps lump `begin` and `end` together into, say, `Interval`, that would return a `r1.Interval` from http://godoc.org/gonum.org/v1/gonum/spatial/r1#Interval ?

interp/interp.go
```diff
@@ -0,0 +1,176 @@
...
+}
+
+// ConstInterpolator1D is a constant 1D interpolator,
+// It is defined over the range [-infinity, infinity].
+type ConstInterpolator1D struct {
+	// Constant Y value.
+	value float64
+}
+
+func (i1d ConstInterpolator1D) begin() float64 {
```
sbinet, line 35:
```suggestion
func (ci ConstInterpolator1D) begin() float64 {
```
(and below as well)

interp/interp.go
```diff
@@ -0,0 +1,176 @@
...
+	// Evaluates interpolated sequence at x in [begin(), end()].
+	// Panics if the argument is outside this range.
+	eval(float64) float64
+}
+
+// ConstInterpolator1D is a constant 1D interpolator,
+// It is defined over the range [-infinity, infinity].
+type ConstInterpolator1D struct {
+	// Constant Y value.
+	value float64
```
sbinet, line 32: if we were to export that field, we could get rid of the `NewConstInterpolator1D` function.

interp/interp.go
```diff
@@ -0,0 +1,176 @@
...
+			panic("interp: xs values not strictly increasing")
+		}
+	}
+}
+
+// NewLinearInterpolator1D creates a new linear 1D interpolator.
+// xs and ys should contain the X and Y values of interpolated nodes, respectively.
+// Panics if len(xs) < 2, elements of xs are not strictly increasing or
+// len(xs) != len(ys).
+func NewLinearInterpolator1D(xs []float64, ys []float64) *LinearInterpolator1D {
```
sbinet, line 124:
```suggestion
func NewLinearInterpolator1D(xs, ys []float64) *LinearInterpolator1D {
```
also, I'd move the function closer to the type definition.
(ie: just after, really)

interp/interp.go
```diff
@@ -0,0 +1,176 @@
...
+	xs []float64
+
+	// Interpolated Y data values, same len as ys.
+	ys []float64
+
+	// Whether the interpolated function is left- or right-continuous.
+	leftContinuous bool
+}
+
+func (i1d PiecewiseConstInterpolator1D) begin() float64 {
```
sbinet, line 149:
```suggestion
func (pci PiecewiseConstInterpolator1D) begin() float64 {
```
(and below)

interp/interp_test.go
```diff
@@ -0,0 +1,219 @@
...
+		panicked = r != nil
+		message = fmt.Sprint(r)
+	}()
+	fn()
+	return
+}
+
+func TestNewConstInterpolator1D(t *testing.T) {
+	t.Parallel()
+	const value float64 = 42.0
```
sbinet, line 25:
```suggestion
	const value = 42.0
```

interp/interp_test.go
```diff
@@ -0,0 +1,219 @@
...
+	i1d := NewConstInterpolator1D(value)
```
sbinet, line 26:
```suggestion
	ci := NewConstInterpolator1D(value)
```

interp/interp_test.go
```diff
@@ -0,0 +1,219 @@
...
+	if i1d.begin() != math.Inf(-1) {
```
sbinet, line 27:
```suggestion
	if !math.IsInf(ci.begin(), -1) {
```

interp/interp_test.go
```diff
@@ -0,0 +1,219 @@
...
+		t.Errorf("unexpected begin() value: got: %g want: %g", i1d.begin(), math.Inf(-1))
+	}
+	if i1d.end() != math.Inf(1) {
```
sbinet, line 30:
```suggestion
	if !math.Inf(ci.end(), +1) {
```

interp/interp_test.go
```diff
@@ -0,0 +1,219 @@
...
+		t.Errorf("unexpected end() value: got: %g want: %g", i1d.end(), math.Inf(1))
+	}
+}
+
+func TestConstInterpolator1DEval(t *testing.T) {
+	t.Parallel()
+	const value float64 = 42.0
```
sbinet, line 37:
```suggestion
	const value = 42.0
```

interp/interp_test.go
```diff
@@ -0,0 +1,219 @@
...
+	i1d := NewConstInterpolator1D(value)
+	xs := [...]float64{math.Inf(-1), -11, 0.4, 1e9, math.Inf(1)}
```
sbinet, line 39:
```suggestion
	xs := []float64{math.Inf(-1), -11, 0.4, 1e9, math.Inf(1)}
```
(why create an array? instead of a (more common) slice?)

interp/interp_test.go
```diff
@@ -0,0 +1,219 @@
...
+			}
+		}
+	}
+}
+
+func TestLinearInterpolator1DEval(t *testing.T) {
+	t.Parallel()
+	xs := []float64{0, 1, 2}
+	ys := []float64{-0.5, 1.5, 1}
+	i1d := NewLinearInterpolator1D(xs, ys)
```
sbinet, line 144:
```suggestion
	lin := NewLinearInterpolator1D(xs, ys)
```

btracey:
Yea, I can try and take a look tomorrow as well

sbinet:
doing a bit of research on the different strategies for interpolation (_i.e.:_ [scipy.interpolate](https://docs.scipy.org/doc/scipy/reference/interpolate.html)), I would perhaps suggest to split between univariate and multivariate, like we did for the statistical distributions (`stat/distuv` and `stat/distmv`).
or go a bit finer and do `interp1d`, `interp2d` and then `interpmv`?

also: another high-level question.
should the main interface interpolate one value at a time (like is done now), or a whole slice?
or give up and ask for 2 methods?
- `Eval(x float64) float64` and 
- `EvalN(dst, x []float64) []float64`?

interp/interp.go
romanwerpachowski: Right, that's my lack of experience with Go showing :) I thought it's enough to capitalise the type.

interp/interp.go
romanwerpachowski: Will do.

romanwerpachowski:
> doing a bit of research on the different strategies for interpolation (_i.e.:_ [scipy.interpolate](https://docs.scipy.org/doc/scipy/reference/interpolate.html)), I would perhaps suggest to split between univariate and multivariate, like we did for the statistical distributions (`stat/distuv` and `stat/distmv`).
> or go a bit finer and do `interp1d`, `interp2d` and then `interpmv`?

I think 1D (and maybe 2D) is a special case - there is much more algorithms for 1D interpolation, e.g. I don't know about any monotone convex interpolation for multi-dimensional data.

Can this split be made later if needed? I understand the user will be importing from the package.

> 
> also: another high-level question.
> should the main interface interpolate one value at a time (like is done now), or a whole slice?
> or give up and ask for 2 methods?
> 
> * `Eval(x float64) float64` and
> * `EvalN(dst, x []float64) []float64`?

In my experience interpolating a single value is very common, the user should not pay the price of creating a 1-element array to do that.

sbinet:
> Better naming of "this" arguments.

they are called "receivers" in Go :)

> Can this split be made later if needed?

sure. no worries. I was just thinking out loud.

> In my experience interpolating a single value is very common, the user should not pay the price of creating a 1-element array to do that.

I am not disputing that.
but in Go, (at least with the Go compiler from golang.org), function calls are relatively expensive.
(of course, depending on the underlying algorithm that does the interpolation, that cost may be dwarfed by the actual interpolation, so YMMV.)

romanwerpachowski:
> > Better naming of "this" arguments.
> 
> they are called "receivers" in Go :)

OK :)

> > In my experience interpolating a single value is very common, the user should not pay the price of creating a 1-element array to do that.
> 
> I am not disputing that.
> but in Go, (at least with the Go compiler from golang.org), function calls are relatively expensive.
> (of course, depending on the underlying algorithm that does the interpolation, that cost may be dwarfed by the actual interpolation, so YMMV.)

OK, I can add another function which loops over an array, but this will either hide the calls to Eval() inside (so no reduction of function calls) or duplicate the interpolation code. What's the attitude to code reuse vs performance in Gonum?

interp/interp_test.go
romanwerpachowski: I thought it makes more sense for literal values, but I'll change that.

interp/doc.go
```diff
@@ -0,0 +1,11 @@
+// Copyright ©2020 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Package interp implements algorithms for interpolating sampled values.
+//
+// Contains the following 1D interpolation algorithms:
+// 1. constant,
+// 2. linear,
+// 3. piecewise constant.
```
kortschak, line 10:
```
// The routines provided by the interp package handle the following 1-dimensional interpolation algorithms:
//  - constant
//  - linear
//  - piecewise constant
```

interp/interp.go
```diff
@@ -0,0 +1,165 @@
...
+	"sort"
+
+	"gonum.org/v1/gonum/spatial/r1"
+)
+
+// Interpolator1D represents a 1D interpolator.
+// It interpolates a y(x) function based on some data over a range [begin(), end()].
+// Both begin() and end() can be infinite, but begin() must be lower than end().
+type Interpolator1D interface {
+
```
kortschak, line 19: Please remove this blank line.

interp/interp.go
```diff
@@ -0,0 +1,165 @@
...
+	"fmt"
+	"math"
+	"sort"
+
+	"gonum.org/v1/gonum/spatial/r1"
+)
+
+// Interpolator1D represents a 1D interpolator.
+// It interpolates a y(x) function based on some data over a range [begin(), end()].
+// Both begin() and end() can be infinite, but begin() must be lower than end().
```
kortschak, line 17:
```
// Interpolator1D represents a 1-dimensional interpolator.
// It interpolates a function over a defined closed interval range.
```

interp/interp.go
```diff
@@ -0,0 +1,165 @@
...
+type Interpolator1D interface {
+
+	// Returns the closed interval over which one can interpolate.
```
kortschak, line 20:
```
// Interval returns the valid closed interval for interpolation.
```

interp/interp.go
```diff
@@ -0,0 +1,165 @@
...
+	Interval() r1.Interval
+
+	// Evaluates interpolated sequence at x in [begin(), end()].
```
kortschak, line 23:
```
// Eval returns the interpolation value at x.
// It panics if x is not in the interpolation interval.
Eval(x float64) float64
```

Bikeshed: `ValueAt`, `Value`? instead of `Eval`.

interp/interp.go
```diff
@@ -0,0 +1,165 @@
...
+	// Panics if the argument is outside this range.
+	Eval(float64) float64
+}
+
+// ConstInterpolator1D is a constant 1D interpolator,
```
kortschak, line 28:
```
// ConstInterpolator1D is a constant 1-dimensional interpolator,
```

Bikeshed: `Constant1D`.

interp/interp.go
```diff
@@ -0,0 +1,165 @@
...
+// It is defined over the range [-infinity, infinity].
+type ConstInterpolator1D struct {
+	// Constant Y value.
+	Value float64
+}
+
+// Interval implements Interpolator1D.Interval.
```
kortschak, line 35:
```
// Interval returns the valid closed interval for interpolation.
```

interp/interp.go
```diff
@@ -0,0 +1,165 @@
...
+func (ci ConstInterpolator1D) Interval() r1.Interval {
+	return r1.Interval{Min: math.Inf(-1), Max: math.Inf(1)}
+}
+
+// Eval implements Interpolator1D.Eval.
```
kortschak, line 40:
```
// Eval returns the interpolation value at x.
```

mod naming.

interp/interp.go
```diff
@@ -0,0 +1,165 @@
...
+		} else {
+			return i - 1, xs[i-1]
+		}
+	} else {
+		if xs[n-1] == x {
+			return n - 1, x
+		}
+		panic(fmt.Sprintf("interp: x value %g above upper bound %g", x, xs[n-1]))
+	}
+}
```
kortschak, line 64: Please move helpers below callers.

interp/interp.go
```diff
@@ -0,0 +1,165 @@
...
+
+// LinearInterpolator1D is a piecewise linear 1D interpolator, defined
+// over the range [xs[0], xs[len(xs) - 1]].
```
kortschak, line 67:
```
// Linear1D is a piecewise linear 1-dimensional interpolator.
```

The inclusion of `[xs[0], xs[len(xs) - 1]]` does not help in the docs here since the users will not see that.

interp/interp.go
```diff
@@ -0,0 +1,165 @@
...
+	ys []float64
+
+	// Slopes of Y between neighbouring X values. len(slopes) + 1 == len(xs) == len(ys).
+	slopes []float64
+}
+
+// NewLinearInterpolator1D creates a new linear 1D interpolator.
+// xs and ys should contain the X and Y values of interpolated nodes, respectively.
+// Panics if len(xs) < 2, elements of xs are not strictly increasing or
+// len(xs) != len(ys).
```
kortschak, line 82:
```
// NewLinear1D creates a new linear 1-dimensional interpolator.
// xs and ys should contain the X and Y values of interpolated nodes, respectively.
// NewLinear1D panics if len(xs) < 2, elements of xs are not strictly increasing or
// len(xs) != len(ys).
```

interp/interp.go
```diff
@@ -0,0 +1,165 @@
...
+func NewLinearInterpolator1D(xs []float64, ys []float64) *LinearInterpolator1D {
+	validateXsAndYs(xs, ys)
```
kortschak, line 84: Prefer in place.

```
if len(xs) != len(ys) {
    panic("interp: xs and ys have different lengths")
}
if len(xs) < 2 {
   panic("interp: too few points for interpolation")
}
if !sort.Float64sAreSorted(xs) {
    panic("interp: xs values not strictly increasing")
}
```

interp/interp.go
```diff
@@ -0,0 +1,165 @@
...
+		if xs[i] <= xs[i-1] {
+			panic("interp: xs values not strictly increasing")
+		}
+	}
+}
+
+// PiecewiseConstInterpolator1D is a piecewise constant 1D interpolator.
+// It is defined over [xs[0], xs[len(xs)-1]].
+// If leftContinuous == true, then y(xs[i]) == y(xs[i] - eps) for small
+// eps > 0. Otherwise, y(xs[i]) == y(xs[i] + eps).
```
kortschak, line 128:
```
// PiecewiseConstant1D is a piecewise constant 1-dimensional interpolator.
```

interp/interp.go
```diff
@@ -0,0 +1,165 @@
...
+	ys []float64
+
+	// Whether the interpolated function is left- or right-continuous.
+	leftContinuous bool
+}
+
+// NewPiecewiseConstInterpolator1D creates a new piecewise constant 1D interpolator.
+// xs and ys should contain the X and Y values of interpolated nodes, respectively.
+// Panics if len(xs) < 2, elements of xs are not strictly increasing or
+// len(xs) != len(ys).
```
kortschak, line 143:
```
// NewPiecewiseConstant1D creates a new piecewise constant 1-dimensional interpolator.
// xs and ys should contain the X and Y values of interpolated nodes, respectively.
// If leftContinuous == true, then y(xs[i]) == y(xs[i] - eps) for small eps > 0. Otherwise,
// y(xs[i]) == y(xs[i] + eps).
// NewPiecewiseConstant1D panics if len(xs) < 2, elements of xs are not strictly increasing or
// len(xs) != len(ys).
```

interp/interp.go
```diff
@@ -0,0 +1,165 @@
...
+func NewPiecewiseConstInterpolator1D(xs []float64, ys []float64, leftContinuous bool) *PiecewiseConstInterpolator1D {
+	validateXsAndYs(xs, ys)
```
kortschak, line 145: Prefer in place.

```
if len(xs) != len(ys) {
    panic("interp: xs and ys have different lengths")
}
if len(xs) < 2 {
   panic("interp: too few points for interpolation")
}
if !sort.Float64sAreSorted(xs) {
    panic("interp: xs values not strictly increasing")
}
```

interp/interp.go
romanwerpachowski: I'll go for `ValueAt`.

interp/interp.go
romanwerpachowski: I think it could be useful for the user if I exported `leftContinuous`?

romanwerpachowski:
Thanks for the review!

interp/interp.go
kortschak: What do you think the use cases would be for this being exported? Would it be something that needs to be changed after constructions or would that lead to potential confusion. If in doubt err on the side on not exporting since that can be undone later.

romanwerpachowski: For example, if you're integrating a stochastic differential equation for a jump process via Monte Carlo, you will require the paths to be right-continuous. But that's not a crucial requirement. We can leave it as it is now.

interp/interp.go
```diff
@@ -0,0 +1,166 @@
...
+	xs []float64
+
+	// Interpolated Y data values, same len as ys.
+	ys []float64
+
+	// Slopes of Y between neighbouring X values. len(slopes) + 1 == len(xs) == len(ys).
+	slopes []float64
+}
+
+// NewLinear1D creates a new linear 1-dimensional interpolator.
```
btracey, line 56: I don't think I like these New functions. I was expecting so see something like 

```
type Fitter interface {
   Fit(x, y []float64) error
}

type FittableInterpolator interface {
   Fitter
   Interpolator
}
```
In the simple case, this looks something like:
```
f := interp.PiecewiseLinear{}
err := f.Fit(xs, ys)
f.Predict(10)
```

But the power is that it enables easy building of routines that need a way of interpolating, but don't care about how.


You can imagine a function like 

```
func PredictAtPoints(locs []float64, f interp.FittableInterpolator, xdata, ydata []float64) ([]float64, error) {
    if err := f.Fit(xdata, ydata); err != nil{
         return nil, error
    }
    outs := make([]float64, len(locs))
    for i, v := range locs {
         outs[i] = f.Predict(v)
    }
    return outs, nil
}
```

And you would call it with something like

```
outs := make([]float64, 101)
floats.Span(outs, 0, 1)
scheme  := interp.Tikhonov{C: 10}
ys, err := mypkg.PredictAtPoints(outs, scheme, xdata, ydata)
```

You can then imagine building a loop to compare different predictors, etc.

interp/interp.go
```diff
@@ -0,0 +1,166 @@
...
+	xs []float64
+
+	// Interpolated Y data values, same len as ys.
+	ys []float64
+
+	// Whether the interpolated function is left- or right-continuous.
+	leftContinuous bool
+}
+
+// NewPiecewiseConstant1D creates a new piecewise constant 1-dimensional interpolator.
```
btracey, line 107: For this, in the above scheme, it would look something like

```
type PiecewiseConstant struct {
   LeftContinuous bool
   
   xs []float64
   ys []float64
}
```

And now a person uses it by

```
f := PiecewiseConstant{LeftContinuous: false}
err := f.Fit(xs, ys)
```

interp/interp.go
```diff
@@ -0,0 +1,166 @@
...
+	"math"
+	"sort"
+
+	"gonum.org/v1/gonum/spatial/r1"
+)
+
+// Interpolator1D represents a 1-dimensional interpolator.
+// It interpolates a function over a defined closed interval range
+// (which can be infinite on both sides).
+type Interpolator1D interface {
```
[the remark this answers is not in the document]
btracey: Er, not nice code

interp/interp.go
kortschak: The question is whether it's needed to change after it's been constructed.

interp/interp.go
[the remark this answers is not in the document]
kortschak: Agree

interp/doc.go
```diff
@@ -0,0 +1,11 @@
+// Copyright ©2020 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Package interp implements algorithms for interpolating sampled values.
+//
+// The routines provided by the interp package handle the following 1-dimensional interpolation algorithms:
```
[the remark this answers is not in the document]
kortschak: Agree

interp/interp.go
```diff
@@ -0,0 +1,166 @@
...
+	// Interval returns the valid closed interval for interpolation.
+	Interval() r1.Interval
```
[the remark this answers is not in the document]
kortschak: Agree

interp/interp.go
```diff
@@ -0,0 +1,166 @@
...
+	"sort"
+
+	"gonum.org/v1/gonum/spatial/r1"
+)
+
+// Interpolator1D represents a 1-dimensional interpolator.
+// It interpolates a function over a defined closed interval range
+// (which can be infinite on both sides).
+type Interpolator1D interface {
+	// Interval returns the valid closed interval for interpolation.
```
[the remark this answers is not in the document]
kortschak: If it's going away this ceases to be an issue.

interp/interp.go
```diff
@@ -0,0 +1,166 @@
...
+	Interval() r1.Interval
+
+	// ValueAt returns the interpolation value at x.
```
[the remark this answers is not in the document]
kortschak: Nice.

interp/interp.go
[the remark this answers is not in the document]
romanwerpachowski: It makes a difference for piecewise-continuous interpolation. Also, `r1.Interval` has fields `Min` and `Max`, which to me suggests an interval closed on both sides. I would rather keep "closed", it seems to me the most common use case.

interp/doc.go
```diff
@@ -0,0 +1,11 @@
+// Copyright ©2020 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Package interp implements algorithms for interpolating sampled values.
```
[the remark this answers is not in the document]
romanwerpachowski: OK.

interp/doc.go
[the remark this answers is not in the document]
romanwerpachowski: Done.

interp/interp.go
```diff
@@ -0,0 +1,166 @@
...
+	// It panics if x is not in the interpolation interval.
+	ValueAt(float64) float64
+}
+
+// Constant1D is a constant 1D interpolator,
```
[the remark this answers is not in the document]
romanwerpachowski:
```
Constant1D is an interpolator that predicts the same constant value
over the range [-infinity, infinity].
```

?

interp/interp.go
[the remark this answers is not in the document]
kortschak: We actually carefully avoided stating whether the `r1.Interval` types talks about closed or open end points.

interp/interp.go
```diff
@@ -0,0 +1,166 @@
...
+func (ci Constant1D) Interval() r1.Interval {
+	return r1.Interval{Min: math.Inf(-1), Max: math.Inf(1)}
+}
+
+// ValueAt returns the interpolation value at x.
+func (ci Constant1D) ValueAt(x float64) float64 {
+	return ci.Value
+}
+
+// Linear1D is a piecewise linear 1-dimensional interpolator.
```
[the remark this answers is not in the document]
romanwerpachowski: Done.

interp/interp.go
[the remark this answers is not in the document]
romanwerpachowski: I'm not sure. Interpolation and extrapolation are two different concepts, and you often use a different algorithm for one and for the other. For example, one often uses flat extrapolation even if using a piecewise linear / cubic interpolation between the known values (to avoid the divergence of extrapolated values). If you're interpolating interest rates and the last known one is a 50year swap rate, you usually just extend it to +infinity, even if you used a completely different scheme between the known rates. On the other hand, if using piecewise linear interpolation to model a growth of something, you could extrapolate the last known slope. Baking in a fixed extrapolation scheme in every interpolator will remove the flexibility a typical user will need.

I would rather build predictors from interpolators and extrapolators then enforce that everything has to be a predictor defined over the entire real axis.

Also, extrapolation in two or more dimensions becomes more problematic.

interp/interp.go
romanwerpachowski: That's a good idea, thanks.

interp/interp.go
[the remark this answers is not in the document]
romanwerpachowski: OK.

interp/interp.go
romanwerpachowski: No, just to provide the information. But I guess I'm overthinking it.

interp/interp.go
romanwerpachowski: One question: looking at `mat` package for example, I had the impression that when a numerical method fails, `gonum` prefers to panic than to return an error. But in your example `Fit` returns an error. What is the preferred way in `gonum`?

interp/interp.go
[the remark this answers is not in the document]
romanwerpachowski: An open end point cannot be a maximum, since a maximum of a set is its largest element. An open interval doesn't have a largest element.

Am I being too picky about maths terminology here? ;-)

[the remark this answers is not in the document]
kortschak: We don't say whether the `r1.Interval{a, b}` is [a, b], (a, b), [a, b) or (a, b]. This is left up to the user.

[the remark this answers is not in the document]
romanwerpachowski: OK, I guess it's just me for whom the names `Min` and `Max` suggest a closed interval.

interp/interp.go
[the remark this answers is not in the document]
romanwerpachowski: OK, I changed my mind - we can provide the necessary flexibility without `Interval()`. For more advanced usage, we could provide a method for concatenating two or more `Predictor` interfaces and switching between them based on the value of the argument.

interp/interp.go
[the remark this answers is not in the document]
romanwerpachowski: OK.

romanwerpachowski:
Pushed changes in line with btracey's request.

interp/interp.go
```diff
@@ -0,0 +1,158 @@
...
+
+import (
+	"sort"
+)
+
+// Predictor predicts the value of a function. It handles both
+// interpolation and extrapolation.
+type Predictor interface {
+	// Predict returns the predicted value at x.
+	Predict(float64) float64
```
vladimir-ch, line 15: Perhaps directly extend this to take another parameter specifying the order of the derivative returned?

interp/interp.go
```diff
@@ -0,0 +1,158 @@
...
+
+package interp
+
+import (
+	"sort"
+)
+
+// Predictor predicts the value of a function. It handles both
+// interpolation and extrapolation.
+type Predictor interface {
```
vladimir-ch, line 13: I'm not convinced by the name because to me "predict" has a temporal meaning but I don't have a better suggestion other than the obvious Interpolator. On the other hand, the interface doesn't transmit anything related to interpolation apart from the name and being in the `interp` package. After the interpolator is created it behaves just as a regular function with the restriction that values outside of the interpolation region are undefined or defined arbitrarily. Perhaps this could be reflected in the interface by adding a function that provides the region of validity?

interp/interp.go
```diff
@@ -0,0 +1,158 @@
...
+	return pl.ys[i] + pl.slopes[i]*(x-xI)
+}
+
+// PiecewiseConstant is a piecewise constant 1-dimensional interpolator.
+// It extrapolates flat forwards (backwards) the last (first) known Y value.
+type PiecewiseConstant struct {
+	// Whether the interpolated function is left- or right-continuous.
+	// If LeftContinuous == true, then y(xs[i]) == y(xs[i] - eps) for small eps > 0. Otherwise,
+	// y(xs[i]) == y(xs[i] + eps).
+	LeftContinuous bool
```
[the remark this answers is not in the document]
romanwerpachowski: If you want to model stochastic processes, their trajectories can be either left- or right-continuous. We can provide a default value for `LeftContinuous`, which should probably be `false`.

interp/interp.go
```diff
@@ -0,0 +1,158 @@
...
+}
+
+// FittablePredictor is a Predictor which can fit itself to data.
+type FittablePredictor interface {
+	Fitter
+	Predictor
+}
+
+// Constant predicts a constant value.
+type Constant struct {
```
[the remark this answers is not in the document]
romanwerpachowski: I can replace it with a function which creates a `PiecewiseConstant` interpolator between -infinity and +infinity.

interp/interp.go
```diff
@@ -0,0 +1,158 @@
...
+	Value float64
+}
+
+// Predict returns the predicted value at x.
+func (c Constant) Predict(x float64) float64 {
+	return c.Value
+}
+
+// PiecewiseLinear is a piecewise linear 1-dimensional interpolator.
+// It extrapolates flat forwards (backwards) the last (first) known Y value.
```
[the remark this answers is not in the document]
romanwerpachowski: Fine by me.

interp/interp.go
romanwerpachowski: I wouldn't do it right now.

interp/interp.go
romanwerpachowski: See here: https://stats.stackexchange.com/questions/205352/what-is-the-exact-difference-between-prediction-and-extrapolation. I think `Predict` is a good name for a function which does both interpolation and extrapolation.

There was an `Interval` function in the previous version of the PR, but it was removed following previous reviews.

interp/interp.go
```diff
@@ -0,0 +1,158 @@
...
+package interp
+
+import (
+	"sort"
+)
+
+// Predictor predicts the value of a function. It handles both
+// interpolation and extrapolation.
+type Predictor interface {
+	// Predict returns the predicted value at x.
```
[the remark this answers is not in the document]
kortschak: Prefer to add the `x` to the signature for the sake of documentation readability.

interp/interp.go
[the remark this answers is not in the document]
kortschak: Make it `type Constant float64`. If it is going to exist, this is the simplest and clearest way it can.

interp/interp.go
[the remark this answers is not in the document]
kortschak: I'm still not clear why it needs to be alterable. If this is defined as an unexported field set at construction, the details don't need to be propagated to other types in a visible way.

interp/interp.go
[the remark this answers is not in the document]
romanwerpachowski: Can it still have a method `Predict` then? just returning itself?

interp/interp.go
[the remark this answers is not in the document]
romanwerpachowski: Initially it was not alterable, but then one of the reviewers requested to export it... I can hide it back.

interp/interp.go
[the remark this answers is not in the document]
kortschak:
```
type Constant float64

func (c Constant) Predict(float64) float64 { return float64(c) }
```

interp/interp.go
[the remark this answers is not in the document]
kortschak: Yes, Brendan did. If there is a reason for it to be exported, then fine, but it hasn't been given yet.

interp/interp.go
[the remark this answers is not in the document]
romanwerpachowski: OK.

interp/interp.go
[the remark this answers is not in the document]
romanwerpachowski: Thanks!

interp/interp.go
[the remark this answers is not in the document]
romanwerpachowski: I'll let you two come to a consensus and do what you ask ;-)

interp/interp.go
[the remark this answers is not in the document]
vladimir-ch: But does it need to exist at all?

interp/interp.go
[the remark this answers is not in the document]
vladimir-ch: In floating-point arithmetic, does it matter if the function is left or right continuous? If not, it could be omitted even from construction.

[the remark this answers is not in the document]
romanwerpachowski: It does. There are many numbers which can be represented exactly in floating point. For example, 0.

interp/interp.go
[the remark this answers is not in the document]
romanwerpachowski: I can replace it with a function which creates a PiecewiseConstant interpolator between -infinity and +infinity.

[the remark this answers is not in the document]
romanwerpachowski: It is a convenience for what in my experience is not an uncommon use case. For example, when modelling survival curves, you often provide an interpolator for the hazard rate. Often you want a constant hazard rate from 0 to +infinity.

But if it's a problem, I can remove it. Is the preference in `Gonum` for smaller interfaces?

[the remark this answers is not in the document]
kortschak: If there is a utility for constant interpolation parameters in user functions, then this should probably exist for the documentational value; seeing `interpConsumer(dst, interp.Constant(c), ...)` at a call site is more informative than using an `interp.PiecewiseConstant`, although the user could define the type trivially I guess.

As an aside, there is probably at some point value in adding
```
type Function func(float64) float64

func (fn Function) Predict(x float64) float64 { return fn(x) }
```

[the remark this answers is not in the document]
romanwerpachowski: Done.

romanwerpachowski:
PTAL

romanwerpachowski:
Hey guys,

Did you arrive at a consensus w/r to the API?

kortschak:
Waiting on @btracey.

romanwerpachowski:
Sorry, clicked on "close" by mistake :)

romanwerpachowski:
*Ping*

interp/interp.go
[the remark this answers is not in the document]
btracey: The reason why I suggested in be exported was to do as we do in the stat/sample package, something like

```
f := PiecewiseConstant{LeftContinuous: True}.Fit(x, y)
```

That said, I think I'm with Vladimir. I understand Roman's point that it can matter, but I think having it be left continuous is far more common, especially for novice users. You want it to be the typical case that if I give an {x, y} point, the prediction at x is y. 

It seems to me that the best thing to do for this package is to leave it out. For a specialized package like stochastic processes, it is easy to add an independent `RightContiguous` type that fits the interface. If we find it is sufficiently common to need, we could possibly add it here. It seems this is more complicated than a standard user would need though.

interp/interp.go
```diff
@@ -0,0 +1,161 @@
...
+		return pc.ys[i+1]
+	}
+	return pc.ys[i]
+}
+
+// findSegment returns 0 <= i < len(xs) such that xs[i] <= x < xs[i + 1], or -1
+// if no such interval containing x is found. It assumes that len(xs) >= 2
+// without checking.
+func findSegment(xs []float64, x float64) int {
+	return sort.Search(len(xs), func(i int) bool { return xs[i] > x }) - 1
```
btracey, line 160: Why -1 rather than just using `>=`?

romanwerpachowski: Because we're looking for i in [0, len(xs)) such that xs[i] <= x < xs[i + 1], or -1 if there is no such interval. Removing -1 and replacing > with >= would break the function.

romanwerpachowski:
Removed LeftContinuous option (it's left-continuous by default). PTAL.

interp/interp.go
vladimir-ch: The code could be modified so that the not-found value is `len(xs)` in `sort`'s style instead of -1 and then @btracey suggestion should work (I didn't check in detail). Maybe it could simplify some logic here. Either way, the current function comment is not accurate, it implies that the function returns -1 also to the right of `xs` which is not the case.

romanwerpachowski: Good point. I'll improve the comment.

romanwerpachowski:
Clarified the comment for findSegment.

interp/interp.go
```diff
@@ -0,0 +1,154 @@
...
+type Predictor interface {
+	// Predict returns the predicted value at x.
+	Predict(x float64) float64
+}
+
+// Fitter fits a predictor to data.
+type Fitter interface {
+	// Fit fits a predictor to (X, Y) value pairs provided as two slices.
+	// It panics if len(xs) < 2, elements of xs are not strictly increasing or len(xs) != len(ys).
+	Fit(xs, ys []float64)
```
btracey, line 22: Sorry, last thing. This should return an error. We definitely want this interface to be able to support least squares regression, where the prediction could error.

romanwerpachowski: Good point.

romanwerpachowski:
PTAL.

romanwerpachowski:
*ping*

romanwerpachowski:
Thanks!

kortschak:
For future:
```
interp: add new package for 1D interpolation

Includes algorithms for constant, piecewise linear and piecewise constant value interpolation.
```

PR #1334, opened 2020-05-27
floats: Neumaier's compensated summation algorithm

romanwerpachowski:
This change modifies Gonum by adding Neumaier's compensated summation algorithm (https://en.wikipedia.org/wiki/Kahan_summation_algorithm) to `floats` package. 

`NeumaierSum` produces a more accurate sum of a large slice of `float64` values without having to sort them by magnitude.

Benchmark comparison between `floats.Sum` and the new method:

```
BenchmarkNeumaierSumSmall-8    	100000000	        11.7 ns/op	       0 B/op	       0 allocs/op
BenchmarkNeumaierSumMedium-8   	 1000000	      1041 ns/op	       0 B/op	       0 allocs/op
BenchmarkNeumaierSumLarge-8    	   10000	    102234 ns/op	       0 B/op	       0 allocs/op
BenchmarkNeumaierSumHuge-8     	     100	  10711364 ns/op	       0 B/op	       0 allocs/op
BenchmarkSumSmall-8            	465654849	         2.60 ns/op	       0 B/op	       0 allocs/op
BenchmarkSumMedium-8           	11793008	       102 ns/op	       0 B/op	       0 allocs/op
BenchmarkSumLarge-8            	   75109	     15060 ns/op	       0 B/op	       0 allocs/op
BenchmarkSumHuge-8             	     370	   3175291 ns/op	       0 B/op	       0 allocs/op
```

`benchstat` output for `NeumaierSum` (20 reps):
```
name                 time/op
NeumaierSumSmall-8   11.6ns ± 2%
NeumaierSumMedium-8  1.02µs ± 2%
NeumaierSumLarge-8    101µs ± 3%
NeumaierSumHuge-8    10.6ms ± 4%
```

`benchstat` output for `Sum` (20 reps):
```
name         time/op
SumSmall-8   2.52ns ± 1%
SumMedium-8  98.2ns ± 1%
SumLarge-8   14.8µs ± 6%
SumHuge-8    3.17ms ± 1%
```

Please take a look.

### DIFF
--- a/floats/floats.go
+++ b/floats/floats.go
@@ -923,3 +923,18 @@ func Within(s []float64, v float64) int {
 	}
 	return -1
 }
+
+// KahanSum returns the sum of the elements of the slice, calculated in a more
+// accurate but also more expensive way. It uses Kahan's compensated summation
+// algorithm: https://en.wikipedia.org/wiki/Kahan_summation_algorithm.
+func KahanSum(s []float64) float64 {
+	sum := 0.
+	c := 0.
+	for _, x := range s {
+		y := x - c
+		t := sum + y
+		c = (t - sum) - y
+		sum = t
+	}
+	return sum
+}
--- a/floats/floats.go
+++ b/floats/floats.go
@@ -930,14 +930,15 @@ func SumCompensated(s []float64) float64 {
 	// SumCompensated uses an improved version of Kahan's compensated
 	// summation algorithm proposed by Neumaier.
 	// See https://en.wikipedia.org/wiki/Kahan_summation_algorithm for details.
-	sum := 0.
-	c := 0.
+	var sum, c float64
 	for _, x := range s {
-		t := sum + x
+		// This type conversion is here to prevent a sufficiently
+		// "clever" compiler from optimising away these operations.
+		t := float64(sum + x)
 		if math.Abs(sum) >= math.Abs(x) {
-			c += ((sum - t) + x)
+			c += (sum - t) + x
 		} else {
-			c += ((x - t) + sum)
+			c += (x - t) + sum
 		}
 		sum = t
 	}
--- a/floats/floats.go
+++ b/floats/floats.go
@@ -924,21 +924,6 @@ func Within(s []float64, v float64) int {
 	return -1
 }
 
-// KahanSum returns the sum of the elements of the slice, calculated in a more
-// accurate but also more expensive way than Sum. It uses Kahan's compensated
-// summation algorithm: https://en.wikipedia.org/wiki/Kahan_summation_algorithm.
-func KahanSum(s []float64) float64 {
-	sum := 0.
-	c := 0.
-	for _, x := range s {
-		y := x - c
-		t := sum + y
-		c = (t - sum) - y
-		sum = t
-	}
-	return sum
-}
-
 // PairwiseSum returns the sum of the elements of the slice, calculated in a more
 // accurate but also more expensive way than Sum. It uses pairwise summation,
 // which divides the summed slice in two equal parts, sums them independently (with
@@ -951,4 +936,23 @@ func PairwiseSum(s []float64) float64 {
 		return PairwiseSum(s[0:m]) + PairwiseSum(s[m:n])
 	}
 	return Sum(s)
+}
+
+// NeumaierSum returns the sum of the elements of the slice, calculated in a more
+// accurate but also more expensive way than Sum. It uses an improved version of
+// Kahan's compensated summation algorithm proposed by Neumaier: see
+// https://en.wikipedia.org/wiki/Kahan_summation_algorithm for details.
+func NeumaierSum(s []float64) float64 {
+	sum := 0.
+	c := 0.
+	for _, x := range s {
+		t := sum + x
+		if math.Abs(sum) >= math.Abs(x) {
+			c += ((sum - t) + x)
+		} else {
+			c += ((x - t) + sum)
+		}
+		sum = t
+	}
+	return sum + c
 }
--- a/floats/floats.go
+++ b/floats/floats.go
@@ -932,8 +932,8 @@ func SumCompensated(s []float64) float64 {
 	// See https://en.wikipedia.org/wiki/Kahan_summation_algorithm for details.
 	var sum, c float64
 	for _, x := range s {
-		// This type conversion is here to prevent a sufficiently
-		// "clever" compiler from optimising away these operations.
+		// This type conversion is here to prevent a sufficiently smart compiler
+		// from optimising away these operations.
 		t := float64(sum + x)
 		if math.Abs(sum) >= math.Abs(x) {
 			c += (sum - t) + x
--- a/floats/floats.go
+++ b/floats/floats.go
@@ -924,11 +924,12 @@ func Within(s []float64, v float64) int {
 	return -1
 }
 
-// SumCompensated returns the sum of the elements of the slice, calculated in a more
-// accurate but also more expensive way than Sum. It uses an improved version of
-// Kahan's compensated summation algorithm proposed by Neumaier: see
-// https://en.wikipedia.org/wiki/Kahan_summation_algorithm for details.
+// SumCompensated returns the sum of the elements of the slice calculated with greater
+// accuracy than Sum at the expense of additional computation.
 func SumCompensated(s []float64) float64 {
+	// SumCompensated uses an improved version of Kahan's compensated
+	// summation algorithm proposed by Neumaier.
+	// See https://en.wikipedia.org/wiki/Kahan_summation_algorithm for details.
 	sum := 0.
 	c := 0.
 	for _, x := range s {
--- a/floats/floats.go
+++ b/floats/floats.go
@@ -924,20 +924,6 @@ func Within(s []float64, v float64) int {
 	return -1
 }
 
-// PairwiseSum returns the sum of the elements of the slice, calculated in a more
-// accurate but also more expensive way than Sum. It uses pairwise summation,
-// which divides the summed slice in two equal parts, sums them independently (with
-// recursion) and adds the sums. For slices shorter than 8 elements it simply calls
-// Sum. See https://en.wikipedia.org/wiki/Pairwise_summation for details.
-func PairwiseSum(s []float64) float64 {
-	n := len(s)
-	if n >= 8 {
-		m := n / 2
-		return PairwiseSum(s[0:m]) + PairwiseSum(s[m:n])
-	}
-	return Sum(s)
-}
-
 // NeumaierSum returns the sum of the elements of the slice, calculated in a more
 // accurate but also more expensive way than Sum. It uses an improved version of
 // Kahan's compensated summation algorithm proposed by Neumaier: see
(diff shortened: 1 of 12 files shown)


sbinet:
could you add the result of the benchmark (for `Sum` and `KahanSum`) to the commit message?
so we'd get a sense of the added cost and a point of reference for later.

(would be great to pass it through `benchcmp` or at least `benchstat`)

romanwerpachowski:
Pasted the benchmarks. AFAIK `benchcmp` and `benchstat` compare results of the same benchmark before and after a change, while I added a new function + a new benchmark?

kortschak:
`benchstat` with no comparison will give a statistical summary of runs when (use `-count=<number>`).

The WP article mentions further enhancements. Are these worth looking at?

romanwerpachowski:
Great, I'll run it this evening.

I will have a look at the alternatives.

floats/floats.go
```diff
@@ -923,3 +923,18 @@ func Within(s []float64, v float64) int {
...
+// KahanSum returns the sum of the elements of the slice, calculated in a more
+// accurate but also more expensive way. It uses Kahan's compensated summation
+// algorithm: https://en.wikipedia.org/wiki/Kahan_summation_algorithm.
+func KahanSum(s []float64) float64 {
+	sum := 0.
+	c := 0.
+	for _, x := range s {
+		y := x - c
+		t := sum + y
+		c = (t - sum) - y
```
vladimir-ch, line 936: @kortschak , is it guaranteed that the compiler will not optimize these calculations away? Does it need to be sparkled with `float64()`?

romanwerpachowski: The benchmark timings suggest that it doesn't. If the calculations are optimised away when compiling tests, the test for KahanSum will fail. Do you think that's a sufficient guarantee?

kortschak: @josharian AFAIU `float64` wrapping wouldn't do anything since there's no multiply here. Is that correct?

josharian: I believe that that is correct, yes. Never hurts to look at the actual generated machine code, though. `GOSSAFUNC=KahanSum go build .` and then look at ssa.html

romanwerpachowski:
> Great, I'll run it this evening.
> 
> I will have a look at the alternatives.

Interesting. Neumaier summation is both faster and more accurate than Kahan.

romanwerpachowski:
Since Neumaier's algorithm is both faster (less FP calculations inside the loop) and more accurate (passes tests which Kahan summation didn't), I got rid of KahanSum and replaced it with NeumaierSum.

PTAL

floats/floats.go
romanwerpachowski: For `NeumaierSum` (a better version of Kahan):

```
genssa
# C:\Users\Roman\go\src\gonum\floats\floats.go
00000 (945) TEXT "".NeumaierSum(SB), ABIInternal
00001 (945) PCDATA $0, $-2
00002 (945) PCDATA $1, $-2
00003 (945) FUNCDATA $0, gclocals·1a65e721a2ccc325b382662e7ffee780(SB)
00004 (945) FUNCDATA $1, gclocals·69c1753bd5f81501d95132d08af04464(SB)
00005 (945) FUNCDATA $2, gclocals·568470801006e5c0dc3947ea998fe279(SB)
v107
00006 (948) PCDATA $0, $0
v107
00007 (948) PCDATA $1, $0
v107
00008 (948) MOVQ "".s+8(SP), AX
v67
00009 (948) PCDATA $0, $1
v67
00010 (948) PCDATA $1, $1
v67
00011 (948) MOVQ "".s(SP), CX
v54
00012 (948) XORL DX, DX
v52
00013 (948) MOVSD $(0.0), X0
v40
00014 (948) MOVSD $(0.0), X1
b1
00015 (948) JMP 18
v98
00016 (+948) INCQ DX
v38
00017 (957) MOVUPS X2, X1
v89
00018 (+948) CMPQ DX, AX
b2
00019 (948) JGE 45
v19
00020 (948) MOVSD (CX)(DX*8), X2
v65
00021 (+949) MOVUPS X2, X3
v21
00022 (949) ADDSD X1, X2
v22
00023 (?) NOP
v51
00024 (?) NOP
v41
00025 (?) NOP
v27
00026 (?) NOP
# $GOROOT\src\math\unsafe.go
v32
00027 (+23) MOVQ X1, BX
# $GOROOT\src\math\abs.go
v36
00028 (+13) BTRQ $63, BX
# $GOROOT\src\math\unsafe.go
v46
00029 (+29) MOVQ BX, X4
v68
00030 (?) NOP
v55
00031 (?) NOP
v60
00032 (+23) MOVQ X3, BX
# $GOROOT\src\math\abs.go
v63
00033 (+13) BTRQ $63, BX
# $GOROOT\src\math\unsafe.go
v73
00034 (+29) MOVQ BX, X5
# C:\Users\Roman\go\src\gonum\floats\floats.go
v45
00035 (+950) UCOMISD X5, X4
b11
00036 (950) JCS 41
v84
00037 (+951) SUBSD X2, X1
v86
00038 (951) ADDSD X1, X3
v87
00039 (951) ADDSD X3, X0
b13
00040 (951) JMP 16
v91
00041 (+953) SUBSD X2, X3
v93
00042 (953) ADDSD X1, X3
v94
00043 (953) ADDSD X3, X0
b14
00044 (953) JMP 16
v101
00045 (+957) PCDATA $0, $0
v101
00046 (+957) ADDSD X0, X1
v104
00047 (957) MOVSD X1, "".~r1+24(SP)
b5
00048 (957) RET
00049 (?) END
```

romanwerpachowski:
Sorry, one more change: I decided to remove `PairwiseSum` because it's only faster than Neumaier for very small inputs, but less accurate.

floats/floats.go
```diff
@@ -923,3 +923,22 @@ func Within(s []float64, v float64) int {
 	}
 	return -1
 }
+
+// NeumaierSum returns the sum of the elements of the slice, calculated in a more
+// accurate but also more expensive way than Sum. It uses an improved version of
+// Kahan's compensated summation algorithm proposed by Neumaier: see
+// https://en.wikipedia.org/wiki/Kahan_summation_algorithm for details.
+func NeumaierSum(s []float64) float64 {
```
kortschak, line 931: I'd suggest considering calling this `SumCompensated` so that we are not constrained by the algorithm if we decide at a later stage to change it. This would the become
```
// SumCompensated returns the sum of the elements of the slice calculated with greater
// accuracy than Sum at the expense of additional computation.
func SumCompensated(s []float64) float64 {
	// SumCompensated uses an improved version of Kahan's compensated
	// summation algorithm proposed by Neumaier.
	// See https://en.wikipedia.org/wiki/Kahan_summation_algorithm for details.
```

floats/floats_test.go
```diff
@@ -1762,6 +1762,53 @@ func TestWithin(t *testing.T) {
...
+	s2[k] = -10
+
+	for i, test := range []struct {
+		s    []float64
+		want float64
+	}{
+		{
+			// Fails if we use simple Sum.
+			s:    s1,
+			want: 0.,
```
kortschak, line 1785: s/0./0/

floats/floats_test.go
```diff
@@ -1762,6 +1762,53 @@ func TestWithin(t *testing.T) {
...
+		},
+		{
+			// Fails if we use simple Sum.
+			s:    s2,
+			want: 0.,
```
kortschak, line 1790: s/0./0/

floats/floats_test.go
```diff
@@ -1762,6 +1762,53 @@ func TestWithin(t *testing.T) {
...
+			s:    []float64{1, 2, 3, 4, 5, 6, 7, 8, 9, 10},
+			want: 55,
+		},
+		{
+			s:    []float64{1.2e20, 0.1, -2.4e20, -0.1, 1.2e20, 0.2, 0.2},
+			want: 0.4,
+		},
+		{
+			s:    []float64{1, 1e100, 1, -1e100},
+			want: 2.,
```
kortschak, line 1802: s/2./2/

romanwerpachowski:
PTAL.

floats/floats.go
kortschak: I'd prefer this wording and arrangement please.

romanwerpachowski: Sorry, missed that. Will do.

romanwerpachowski:
Travis build shows as "Pending" here, but it's actually [done](https://travis-ci.org/github/gonum/gonum/builds/692477316).

floats/floats.go
```diff
@@ -923,3 +923,23 @@ func Within(s []float64, v float64) int {
...
 }
+
+// SumCompensated returns the sum of the elements of the slice calculated with greater
+// accuracy than Sum at the expense of additional computation.
+func SumCompensated(s []float64) float64 {
+	// SumCompensated uses an improved version of Kahan's compensated
+	// summation algorithm proposed by Neumaier.
+	// See https://en.wikipedia.org/wiki/Kahan_summation_algorithm for details.
+	sum := 0.
+	c := 0.
```
kortschak, line 934: `var sum, c float64`

The use of a single `.` is a small mark to denote type, so better to be explicit.

floats/floats.go
```diff
@@ -923,3 +923,23 @@ func Within(s []float64, v float64) int {
...
+	for _, x := range s {
+		t := sum + x
+		if math.Abs(sum) >= math.Abs(x) {
+			c += ((sum - t) + x)
```
kortschak, line 938: The outer parens here and below can be lost, no?

floats/floats.go
```diff
@@ -923,3 +923,23 @@ func Within(s []float64, v float64) int {
...
+// SumCompensated returns the sum of the elements of the slice calculated with greater
+// accuracy than Sum at the expense of additional computation.
+func SumCompensated(s []float64) float64 {
+	// SumCompensated uses an improved version of Kahan's compensated
+	// summation algorithm proposed by Neumaier.
+	// See https://en.wikipedia.org/wiki/Kahan_summation_algorithm for details.
+	sum := 0.
+	c := 0.
+	for _, x := range s {
+		t := sum + x
```
kortschak, line 936: I think this may want to have a caveat comment for future us for if the compiler gets smart enough to optimise these away as @vladimir-ch wrote [here](https://github.com/gonum/gonum/pull/1334#discussion_r431744475). It doesn't now, and as you said the tests will fail if it does, but if we have a comment to watch out, it might be easier to find if it does eventuate.

The spec has [this to say](https://golang.org/ref/spec#Arithmetic_operators):
>An implementation may combine multiple floating-point operations into a single fused operation, possibly across statements, and produce a result that differs from the value obtained by executing and rounding the instructions individually. An explicit floating-point type conversion rounds to the precision of the target type, preventing fusion that would discard that rounding.

Maybe even just adding `t := float64(sum + x)` with that comment to head it off paranoically?

romanwerpachowski: Done with a small tweak ;-)

romanwerpachowski:
PTAL.

kortschak:
My wording was wrong. The phrase should be ["sufficiently smart compiler"](http://wiki.c2.com/?SufficientlySmartCompiler). It shouldn't be quoted as it refers to a well understood concept.

romanwerpachowski:
I quoted it because I think it's hyperbole to attribute intelligence to any compiler - if it was really smart it would grok that we're doing compensated summation and do not optimise the algorithm away.

I would rather use "sufficiently aggressive compiler", but can use your wording too.

PR #1335, opened 2020-05-27
float: Improve test coverage.

romanwerpachowski:
This change modifies Gonum to increase test coverage of package `float`. I did not reach 100% coverage because some statements seem to be unreachable. In one case (line 580 in `floats.go`) I am not sure if the value returned is correct - why not return `n-1`?

Please take a look.

### DIFF
--- a/floats/floats.go
+++ b/floats/floats.go
@@ -577,7 +577,7 @@ func NearestIdxForSpan(n int, l, u float64, v float64) int {
 			if math.Copysign(1, v) == math.Copysign(1, l) {
 				return 0
 			}
-			return n - 1
+			return n/2 + 1
 		}
 		if math.Copysign(1, v) == math.Copysign(1, l) {
 			return 0
--- a/floats/floats.go
+++ b/floats/floats.go
@@ -577,7 +577,7 @@ func NearestIdxForSpan(n int, l, u float64, v float64) int {
 			if math.Copysign(1, v) == math.Copysign(1, l) {
 				return 0
 			}
-			return n/2 + 1
+			return n - 1
 		}
 		if math.Copysign(1, v) == math.Copysign(1, l) {
 			return 0
--- a/floats/floats_test.go
+++ b/floats/floats_test.go
@@ -1127,7 +1127,7 @@ func TestNearestIdxForSpan(t *testing.T) {
 			lower:  math.Inf(-1),
 			upper:  math.Inf(1),
 			value:  math.Inf(1),
-			idx:    4,
+			idx:    3,
 		},
 		{
 			length: 4,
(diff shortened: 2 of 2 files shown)


kortschak:
>In one case (line 580 in floats.go) I am not sure if the value returned is correct - why not return `n-1`?

Do you mean [here](https://github.com/gonum/gonum/blob/043b26cc1c488e3c0021b93d56a66d9d03ec836a/floats/floats.go#L580)?

If you get to that line `l` and `u` are infinite but not the same, the query is finite, and the number of elements is odd, so we pick the middle element.

romanwerpachowski:
Doesn't [this line (574)](https://github.com/gonum/gonum/blob/043b26cc1c488e3c0021b93d56a66d9d03ec836a/floats/floats.go#L574) guarantee that at [this point (line 580)](https://github.com/gonum/gonum/blob/043b26cc1c488e3c0021b93d56a66d9d03ec836a/floats/floats.go#L580) `v` is infinite?

```
if !math.IsInf(v, 0) {
	return n / 2
}
if math.Copysign(1, v) == math.Copysign(1, l) {
	return 0
}
	return n/2 + 1
```

kortschak:
>Doesn't this line (574) guarantee that at this point (line 580) v is infinite?

Sorry, yes, you're absolutley right, good catch. The issue that that code should reflect is #413.

Please fix L580 too.

romanwerpachowski:
Great. Will do.

kortschak:
This looks like it will be ready soon; can you send another PR with the commit message "A+C: add Roman Werpachowski" adding yourself to CONTRIBUTORS and you or your employer (as applies to your situation) to AUTHORS.

romanwerpachowski:
Done in https://github.com/gonum/gonum/pull/1336

romanwerpachowski:
What about [this line](https://github.com/gonum/gonum/blob/043b26cc1c488e3c0021b93d56a66d9d03ec836a/floats/floats.go#L585)? Doesn't it have the same problem? It is covered by a [test case](https://github.com/gonum/gonum/blob/master/floats/floats_test.go#L1117), but I'm not sure if the test is correct.

kortschak:
OK, I've recovered the thinking that went into that.

If you look at the playground [link here](https://play.golang.org/p/oIT75L_A6HI) you can see the behaviour of `Span` which is the model that `NearestIdxForSpan` has to match. If the number of elements is odd, the middle value is 0, otherwise there is a crossover between the left and right of the slice. This is why there is a special case for odd-lengthed slices; finite queries can match the middle element. Now we need to explain the remaining cases; this is that queries match the lowest indexed element of the span, so if we have and 4-long [-Inf, Inf] span, [-Inf -Inf Inf Inf], -2 would match element 0, and 2 would match element 2 (as would Inf). This is where the `n/2 + 1` and `n/2` expressions come from.

So this means the 580 line fix needs to be reverted.

romanwerpachowski:
Done. PTAL.

PR #1340, opened 2020-05-31
stat/distuv: Improve test coverage for `Categorical` distribution + add documentation to `interfaces.go`

romanwerpachowski:
This change modifies the `stat/distuv` package in Gonum to:
- improve test coverage for `Categorical` distribution
- add benchmarks for `Rand` function on `Categorical`
- add documentation comments to `interfaces.go`

Please take a look.

### DIFF
--- a/stat/distuv/interfaces.go
+++ b/stat/distuv/interfaces.go
@@ -4,19 +4,23 @@
 
 package distuv
 
+// LogProber calculates the log-probability for a given value of a random variable.
 type LogProber interface {
 	LogProb(float64) float64
 }
 
+// Rander samples a random variable.
 type Rander interface {
 	Rand() float64
 }
 
+// RandLogProber is a Rander and a LogProber.
 type RandLogProber interface {
 	Rander
 	LogProber
 }
 
+// Quantiler calculates the quantile of a distribution for given probability.
 type Quantiler interface {
 	Quantile(p float64) float64
 }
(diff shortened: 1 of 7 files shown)


kortschak:
Please start work off origin/master so that we don't have so many commits in the PR.

stat/distuv/interfaces.go
```diff
@@ -4,19 +4,23 @@
 
 package distuv
```
kortschak, line 5: The changes in this file really should be a separate PR.

romanwerpachowski:
> Please start work off origin/master so that we don't have so many commits in the PR.

Can you recommend a workflow? I am not terribly familiar with pull requests.

kortschak:
When I'm working on repos that I'm not an owner of, I keep two remotes, one is the origin and the other is my fork. Before I do some work, I make sure I pull from the origin/master and then branch from there. This way, when I push to my kortschak/feature-branch remote branch, I only have the changes that I am actually making in that branch.

After the PR has been merged, I delete my feature branch, and the remote branch and pull the origin/master again, which will now contain the added feature.

With repos that I own, it's a bit easier, since I only need to have one remote, but the process is essentially still the same.

romanwerpachowski:
Thanks.

romanwerpachowski:
I think I messed up this branch, I'll make two new separate PRs.

PR #1341, opened 2020-06-01
stat/distuv: Add documentation to `interfaces.go`

romanwerpachowski:
Please take a look.

### DIFF
--- a/stat/distuv/interfaces.go
+++ b/stat/distuv/interfaces.go
@@ -4,19 +4,23 @@
 
 package distuv
 
+// LogProber calculates the log-probability for a given value of a random variable.
 type LogProber interface {
 	LogProb(float64) float64
 }
 
+// Rander samples a random variable.
 type Rander interface {
 	Rand() float64
 }
 
+// RandLogProber is a Rander and a LogProber.
 type RandLogProber interface {
 	Rander
 	LogProber
 }
 
+// Quantiler calculates the quantile of a distribution for given probability.
 type Quantiler interface {
 	Quantile(p float64) float64
 }
--- a/stat/distuv/interfaces.go
+++ b/stat/distuv/interfaces.go
@@ -6,7 +6,10 @@ package distuv
 
 // LogProber wraps the LogProb method.
 type LogProber interface {
-	LogProb(float64) float64
+	// LogProb returns the natural logarithm of the
+	// value of the probability density or probability
+	// mass function at x.
+	LogProb(x float64) float64
 }
 
 // Rander wraps the Rand method.
@@ -23,6 +26,7 @@ type RandLogProber interface {
 
 // Quantiler wraps the Quantile method.
 type Quantiler interface {
-	// Quantile returns the minimum value of x from amongst all those values whose CDF value exceeds or equals p.
+	// Quantile returns the minimum value of x from amongst
+	// all those values whose CDF value exceeds or equals p.
 	Quantile(p float64) float64
 }
--- a/stat/distuv/interfaces.go
+++ b/stat/distuv/interfaces.go
@@ -4,23 +4,25 @@
 
 package distuv
 
-// LogProber calculates the log-probability for a given value of a random variable.
+// LogProber wraps the LogProb method.
 type LogProber interface {
 	LogProb(float64) float64
 }
 
-// Rander samples a random variable.
+// Rander wraps the Rand method.
 type Rander interface {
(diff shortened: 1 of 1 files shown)


stat/distuv/interfaces.go
```diff
@@ -4,19 +4,23 @@
 
 package distuv
 
+// LogProber calculates the log-probability for a given value of a random variable.
```
kortschak, line 7:
```
// LogProber wraps the LogProb method.
```

stat/distuv/interfaces.go
```diff
@@ -4,19 +4,23 @@
...
 type LogProber interface {
 	LogProb(float64) float64
 }
 
+// Rander samples a random variable.
```
kortschak, line 12:
```
// Rander wraps the Rand method.
```

stat/distuv/interfaces.go
```diff
@@ -4,19 +4,23 @@
...
 type Rander interface {
 	Rand() float64
 }
 
+// RandLogProber is a Rander and a LogProber.
```
kortschak, line 17:
```
// RandLogProber is the interface that groups the Rander and LogProber methods.
```

stat/distuv/interfaces.go
```diff
@@ -4,19 +4,23 @@
...
 type RandLogProber interface {
 	Rander
 	LogProber
 }
 
+// Quantiler calculates the quantile of a distribution for given probability.
```
kortschak, line 23:
```
// Quantiler wraps the Quantile method.
```

stat/distuv/interfaces.go
```diff
@@ -4,19 +4,23 @@
 
 package distuv
 
+// LogProber calculates the log-probability for a given value of a random variable.
 type LogProber interface {
 	LogProb(float64) float64
```
kortschak, line 9:
```
// LogProb returns the natural logarithm of the
// value of the probability density function at x.
LogProb(x float64) float64
```

stat/distuv/interfaces.go
```diff
@@ -4,19 +4,23 @@
...
 }
 
+// Rander samples a random variable.
 type Rander interface {
 	Rand() float64
```
kortschak, line 14:
```
// Rand returns a random sample drawn from the distribution.
Rand() float64
```

stat/distuv/interfaces.go
```diff
@@ -4,19 +4,23 @@
...
 
+// RandLogProber is a Rander and a LogProber.
 type RandLogProber interface {
 	Rander
 	LogProber
 }
 
+// Quantiler calculates the quantile of a distribution for given probability.
 type Quantiler interface {
 	Quantile(p float64) float64
```
kortschak, line 25:
```
// Quantile returns inverse of the cumulative distribution
// function for the probability p.
```

romanwerpachowski: I'll use the new wording which we agreed for Bernoulli.

stat/distuv/interfaces.go
romanwerpachowski: What about discrete distributions, which do not have the PDF but a probability mass function?

romanwerpachowski:
PTAL. I left some questions.

stat/distuv/interfaces.go
kortschak: Can you wrap the text more tightly; godoc will format text in non-code chunk doc comments, but does not do this for comments in struct or interface definitions, so we need to be more careful there. Something on the order of 50 characters is pretty good.

stat/distuv/interfaces.go
kortschak:
```
// LogProb returns the natural logarithm of the
// value of the probability density or probability
// mass function at x.
LogProb(x float64) float64
```
?

Presumably equivalent text in `Prober`.

romanwerpachowski: Done. Where's `Prober`?

stat/distuv/interfaces.go
romanwerpachowski: Done.

romanwerpachowski:
PTAL.

stat/distuv/interfaces.go
kortschak: Doesn't exist yet I suppose. We don't need it if it's not there.

PR #1351, opened 2020-06-08
stat/distuv: use math.Expm1 when calculating Exponential.CDF to improve accuracy for small x

romanwerpachowski:
Please take a look.

This changes the way the CDF is calculated for exponential distribution from

```
1 - math.Exp(-e.Rate * x)
```

to

```
- math.Expm1(-e.Rate * x)
```

giving higher accuracy for very small x.

### DIFF
--- a/stat/distuv/exponential.go
+++ b/stat/distuv/exponential.go
@@ -24,7 +24,7 @@ func (e Exponential) CDF(x float64) float64 {
 	if x < 0 {
 		return 0
 	}
-	return 1 - math.Exp(-e.Rate*x)
+	return -math.Expm1(-e.Rate * x)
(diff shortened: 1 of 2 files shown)


romanwerpachowski:
Changed the PR description.

kortschak:
For the record, PR titles and commits follow [the Go convention](https://golang.org/doc/contribute.html#commit_messages). So, given that this is a single commit PR, here both would be "stat/distuv: improve accuracy of Exponential.CDF for small values" or something similar (the code change says what happens, the commit message tells the intention - you could append " by using math.Expm1" if you really want, but I don't think it's necessary since it's in the diff).

PR #1353, opened 2020-06-10
stat/distuv: provide complete test coverage for F distribution

romanwerpachowski:
Also added a function `checkMode` to `distribution_test.go` which compares the analytic and estimated mode.

Please take a look.

### DIFF
--- a/stat/distuv/distribution_test.go
+++ b/stat/distuv/distribution_test.go
@@ -69,6 +69,10 @@ type cumulantProber interface {
 	probLogprober
 }
 
+type moder interface {
+	Mode() float64
+}
+
 func checkMean(t *testing.T, i int, x []float64, m meaner, tol float64) {
 	mean := stat.Mean(x, nil)
 	if !floats.EqualWithinAbsOrRel(mean, m.Mean(), tol, tol) {
@@ -213,6 +217,18 @@ func checkProbQuantContinuous(t *testing.T, i int, xs []float64, c cumulantProbe
 			t.Errorf("Empirical CDF doesn't match quantile. Case %v.", i)
 		}
 		xp = x
+	}
+}
+
+func checkMode(t *testing.T, i int, xs []float64, m moder, tol float64) {
+	rXs := make([]float64, len(xs))
+	for i := 0; i < len(xs); i++ {
+		rXs[i] = floats.RoundEven(xs[i]*tol, 0) / tol
+		expected, _ := stat.Mode(rXs, nil)
+		actual := m.Mode()
+		if !floats.EqualWithinAbs(expected, actual, tol) {
+			t.Errorf("Mode mismatch case %d: want %g, got %v", i, expected, actual)
+		}
 	}
 }
 
--- a/stat/distuv/distribution_test.go
+++ b/stat/distuv/distribution_test.go
@@ -69,10 +69,6 @@ type cumulantProber interface {
 	probLogprober
 }
 
-type moder interface {
-	Mode() float64
-}
-
 func checkMean(t *testing.T, i int, x []float64, m meaner, tol float64) {
 	mean := stat.Mean(x, nil)
 	if !floats.EqualWithinAbsOrRel(mean, m.Mean(), tol, tol) {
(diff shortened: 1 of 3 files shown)


kortschak:
Kicked it.

stat/distuv/f.go
```diff
@@ -109,9 +109,6 @@ func (f F) Skewness() float64 {
 //
 // StdDev returns NaN if the D2 parameter is less than or equal to 4.
 func (f F) StdDev() float64 {
-	if f.D2 <= 4 {
-		return math.NaN()
-	}
```
kortschak, line 114: I think this is likely there to avoid to the SQRTSD call. Removing these lines actually significantly increases the complexity of the generated code, so can you revert this.

stat/distuv/distribution_test.go
```diff
@@ -216,6 +220,18 @@ func checkProbQuantContinuous(t *testing.T, i int, xs []float64, c cumulantProbe
 	}
 }
 
+func checkMode(t *testing.T, i int, xs []float64, m moder, dx float64, tol float64) {
+	rXs := make([]float64, len(xs))
+	for j, x := range xs {
+		rXs[j] = floats.RoundEven(x/dx, 0) * dx
+	}
+	expected, _ := stat.Mode(rXs, nil)
```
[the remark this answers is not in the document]
romanwerpachowski: Python habits :D

stat/distuv/f.go
romanwerpachowski: But this this only matters if D2 <= 4. Do you think that in actual usage Variance will be called often for cases where D2 <= 4?

kortschak: My point was that the generated code is more complex and longer for no real gain.

stat/distuv/distribution_test.go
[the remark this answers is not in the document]
kortschak: One beat words are best.

stat/distuv/f.go
romanwerpachowski: Done.

stat/distuv/distribution_test.go
```diff
@@ -216,6 +216,22 @@ func checkProbQuantContinuous(t *testing.T, i int, xs []float64, c cumulantProbe
...
 }
 
+type moder interface {
+	Mode() float64
+}
+
+func checkMode(t *testing.T, i int, xs []float64, m moder, dx float64, tol float64) {
+	rXs := make([]float64, len(xs))
+	for j, x := range xs {
+		rXs[j] = floats.RoundEven(x/dx, 0) * dx
```
kortschak, line 226: Please use [`math.RountToEven`](https://golang.org/pkg/math/#RoundToEven). The implementation there is better than ours and ours should be fixed (and even then warned against).

stat/distuv/f_test.go
```diff
@@ -90,4 +91,62 @@ func testF(t *testing.T, f F, i int) {
 	checkExKurtosis(t, i, x, f, 1e-1)
 	checkSkewness(t, i, x, f, 5e-2)
 	checkQuantileCDFSurvival(t, i, x, f, 5e-3)
+	checkMode(t, i, x, f, 2e-2, 3e-2)
```
kortschak, line 94: What are the magic numbers based on?

stat/distuv/distribution_test.go
romanwerpachowski: Maybe the Gonum one should be removed then? Done.

stat/distuv/f_test.go
romanwerpachowski: Testing `Mode` empirically requires discretising the values (binning). The first argument is the size of the bin. F distribution's support extends over a range of the order of 1-10, so 0.02 seems like a decent size. The second argument is the tolerance. It's larger than the bin size due to finite sample size.

stat/distuv/distribution_test.go
kortschak: Yeah, maybe. They're different and the philosophy/practicality around them is sort of vexed. The one in math doesn't do precisions other than to integer. There are (entirely valid arguments) that this is because it's not possible to correctly do anything else. But on the other hand people want rounding to arbitrary digits. We have been avoiding looking under that carpet for a while now.

stat/distuv/f_test.go
kortschak: Thanks

stat/distuv/distribution_test.go
romanwerpachowski: Would there be interest in an arbitrary precision module for Gonum?

kortschak: There is math/big for that. There are also decimal packages for arbitrary decimal precision (finance) outside the standard library.

kortschak:
Thanks

PR #1355, opened 2020-06-12
stat/distuv: Add missing tests for Mode and fix bugs in some implementations

romanwerpachowski:
1. Adds tests for Mode function where missing.
2. Fixes a bug in lognormal Mode.
3. Fixes a bug in Weibull Mode.

Please take a look.

### DIFF
--- a/stat/distuv/gamma_test.go
+++ b/stat/distuv/gamma_test.go
@@ -5,6 +5,7 @@
 package distuv
 
 import (
+	"math"
 	"sort"
 	"testing"
 
@@ -64,4 +65,11 @@ func testGamma(t *testing.T, f Gamma, i int) {
 	checkExKurtosis(t, i, x, f, 2e-1)
 	checkProbContinuous(t, i, x, f, 1e-3)
 	checkQuantileCDFSurvival(t, i, x, f, 5e-2)
+	if f.Alpha >= 1 {
+		checkMode(t, i, x, f, 2e-1, 1)
+	} else {
+		if !math.IsNaN(f.Mode()) {
+			t.Errorf("Expected NaN mode for alpha < 1, got %v", f.Mode())
+		}
+	}
 }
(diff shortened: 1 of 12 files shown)


stat/distuv/gamma_test.go
```diff
@@ -64,4 +65,11 @@ func testGamma(t *testing.T, f Gamma, i int) {
 	checkExKurtosis(t, i, x, f, 2e-1)
 	checkProbContinuous(t, i, x, f, 1e-3)
 	checkQuantileCDFSurvival(t, i, x, f, 5e-2)
+	if f.Alpha >= 1 {
+		checkMode(t, i, x, f, 2e-1, 1)
+	} else {
+		if !math.IsNaN(f.Mode()) {
+			t.Errorf("Expected NaN mode for alpha < 1, got %v", f.Mode())
+		}
+	}
```
kortschak, line 74: Nit. Bad things happen first if possible. (This also makes the condition obviously the same as the message).

```
	if f.Alpha < 1 {
		if !math.IsNaN(f.Mode()) {
			t.Errorf("Expected NaN mode for alpha < 1, got %v", f.Mode())
		}
	} else {
		checkMode(t, i, x, f, 2e-1, 1)
	}
```

stat/distuv/weibull_test.go
```diff
@@ -247,6 +248,13 @@ func testWeibull(t *testing.T, dist Weibull, i int) {
 	checkSkewness(t, i, x, dist, tol)
 	checkMedian(t, i, x, dist, tol)
 	checkQuantileCDFSurvival(t, i, x, dist, tol)
-	checkProbContinuous(t, i, x, dist, 1e-10)
+	var probTol float64
```
kortschak, line 251: It's worth having a comment to explain why this is necessary.

stat/distuv/gamma_test.go
romanwerpachowski: Done.

stat/distuv/weibull_test.go
romanwerpachowski: Done.

romanwerpachowski:
Thanks, PTAL.

kortschak:
Thanks

PR #1357, opened 2020-06-15
stat/distuv: Provide full test coverage for Gamma distribution and small fixes

romanwerpachowski:
- Full test coverage
- Removed redundant calculations in the small alpha version of Rand()
- Lowered the threshold for small alpha to 0.2 in accordance with https://www4.stat.ncsu.edu/~rmartin/Codes/rgamss.R
- Added the option to pass integration bounds to PDF normalisation check
- Other small fixes

Please take a look.

### DIFF
--- a/stat/distuv/distribution_test.go
+++ b/stat/distuv/distribution_test.go
@@ -161,9 +161,10 @@ func checkProbContinuous(t *testing.T, i int, x []float64, p probLogprober, tol 
 	checkProbContinuousWithBounds(t, i, x, math.Inf(-1), math.Inf(1), p, tol)
 }
 
-func checkProbContinuousWithBounds(t *testing.T, i int, x []float64, lb float64, ub float64, p probLogprober, tol float64) {
-	// Check that the PDF is consistent (integrates to 1). Takes lower and upper bounds for X.
-	q := quad.Fixed(p.Prob, lb, ub, 1000000, nil, 0)
+// checkProbContinuousWithBounds checks that the PDF is consistent
+// with LogPDF and integrates to 1 from the lower to upper bound.
+func checkProbContinuousWithBounds(t *testing.T, i int, x []float64, lower float64, upper float64, p probLogprober, tol float64) {
+	q := quad.Fixed(p.Prob, lower, upper, 1000000, nil, 0)
 	if math.Abs(q-1) > tol {
 		t.Errorf("Probability distribution doesn't integrate to 1. Case %v: Got %v", i, q)
 	}
--- a/stat/distuv/distribution_test.go
+++ b/stat/distuv/distribution_test.go
@@ -158,8 +158,12 @@ func checkQuantileCDFSurvival(t *testing.T, i int, xs []float64, c cumulanter, t
 }
 
 func checkProbContinuous(t *testing.T, i int, x []float64, p probLogprober, tol float64) {
-	// Check that the PDF is consistent (integrates to 1).
-	q := quad.Fixed(p.Prob, math.Inf(-1), math.Inf(1), 1000000, nil, 0)
+	checkProbContinuousWithBounds(t, i, x, math.Inf(-1), math.Inf(1), p, tol)
+}
+
+func checkProbContinuousWithBounds(t *testing.T, i int, x []float64, lb float64, ub float64, p probLogprober, tol float64) {
+	// Check that the PDF is consistent (integrates to 1). Takes lower and upper bounds for X.
+	q := quad.Fixed(p.Prob, lb, ub, 1000000, nil, 0)
 	if math.Abs(q-1) > tol {
 		t.Errorf("Probability distribution doesn't integrate to 1. Case %v: Got %v", i, q)
 	}
@@ -261,7 +265,7 @@ func testRandLogProbContinuous(t *testing.T, i int, min float64, x []float64, f 
 			return math.Exp(f.LogProb(x))
 		}
 		// Integrate the PDF to find the CDF
-		estCDF := quad.Fixed(prob, min, pt, 1000, nil, 0)
+		estCDF := quad.Fixed(prob, min, pt, 10000, nil, 0)
 		if !floats.EqualWithinAbsOrRel(cdf, estCDF, tol, tol) {
 			t.Errorf("Mismatch between integral of PDF and empirical CDF. Case %v. Want %v, got %v", i, cdf, estCDF)
 		}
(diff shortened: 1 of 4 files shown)


stat/distuv/distribution_test.go
```diff
@@ -158,8 +158,12 @@ func checkQuantileCDFSurvival(t *testing.T, i int, xs []float64, c cumulanter, t
 }
 
 func checkProbContinuous(t *testing.T, i int, x []float64, p probLogprober, tol float64) {
-	// Check that the PDF is consistent (integrates to 1).
-	q := quad.Fixed(p.Prob, math.Inf(-1), math.Inf(1), 1000000, nil, 0)
+	checkProbContinuousWithBounds(t, i, x, math.Inf(-1), math.Inf(1), p, tol)
+}
+
+func checkProbContinuousWithBounds(t *testing.T, i int, x []float64, lb float64, ub float64, p probLogprober, tol float64) {
+	// Check that the PDF is consistent (integrates to 1). Takes lower and upper bounds for X.
```
kortschak, line 165:
```
// checkProbContinuousWithBounds check that the PDF is consistent,
// it integrates to 1 from the lower to upper bound.
func checkProbContinuousWithBounds(t *testing.T, i int, x []float64, lower, upper float64, p probLogprober, tol float64) {
```

and associated param changes in the body.

If there is no `checkProbContinuous`, remove the `WithBounds` suffix. This is testing code, so the API doesn't need to be friendly and efficient, just obviously conprehensible.

stat/distuv/distribution_test.go
```diff
@@ -261,7 +265,7 @@ func testRandLogProbContinuous(t *testing.T, i int, min float64, x []float64, f
 			return math.Exp(f.LogProb(x))
 		}
 		// Integrate the PDF to find the CDF
-		estCDF := quad.Fixed(prob, min, pt, 1000, nil, 0)
+		estCDF := quad.Fixed(prob, min, pt, 10000, nil, 0)
```
kortschak, line 268: Why was this necessary?

stat/distuv/gamma.go
```diff
@@ -93,6 +93,10 @@ func (g Gamma) Quantile(p float64) float64 {
 //
 // Rand panics if either alpha or beta is <= 0.
 func (g Gamma) Rand() float64 {
+	const (
+		// 0.2 threshold is from https://www4.stat.ncsu.edu/~rmartin/Codes/rgamss.R
```
kortschak, line 97: The convention is fully formed sentences (also single line decl).

```
// The 0.2 threshold is from https://www4.stat.ncsu.edu/~rmartin/Codes/rgamss.R
// described in detail in https://arxiv.org/abs/1302.1884.
const smallAlphaThresh = 0.2
```

stat/distuv/gamma.go
```diff
@@ -124,9 +128,8 @@ func (g Gamma) Rand() float64 {
 
 		// Algorithm adjusted to work in log space as much as possible.
 		lambda := 1/a - 1
-		lw := math.Log(a) - 1 - math.Log(1-a)
-		lr := -math.Log(1 + math.Exp(lw))
-		lc, _ := math.Lgamma(a + 1)
+		w := 1 / lambda / math.E
+		lr := -math.Log1p(w)
```
kortschak, line 132: Please revert this. The original code intentionally works in log space (as described in the comment above).

stat/distuv/gamma.go
```diff
@@ -135,18 +138,19 @@ func (g Gamma) Rand() float64 {
...
-				lEta = lc - z
+				lEta = -z
 			} else {
-				lEta = lc + lw + math.Log(lambda) + lambda*z
+				lEta = -1 + lambda*z
 			}
 			if lh-lEta > -exprnd() {
-				return math.Exp(-z/a) / b
+				return eza / b
 			}
```
kortschak, line 151: Here too.

stat/distuv/gamma_test.go
```diff
@@ -31,7 +31,9 @@ func TestGamma(t *testing.T) {
 	}
 	src := rand.New(rand.NewSource(1))
 	for i, g := range []Gamma{
+		{Alpha: 0.1, Beta: 0.8, Src: src},
```
kortschak, line 35: Remove this blank line?

stat/distuv/distribution_test.go
romanwerpachowski: There is :) I could add the bounds to the original function, but it'd cause a cascade of changes everywhere.

stat/distuv/distribution_test.go
romanwerpachowski: PDFs with singularities (like Gamma for alpha < 1).

stat/distuv/gamma.go
romanwerpachowski: But it still does!

stat/distuv/distribution_test.go
kortschak: OK, just the signature/doc change then.

stat/distuv/distribution_test.go
kortschak: Thanks.

stat/distuv/gamma.go
romanwerpachowski: `lc` is unnecessary (we only ever need the difference `lh - lEta`).

kortschak: Sorry, missed this. This is nice.

stat/distuv/gamma.go
kortschak: The question that is in my mind is whether `-math.Log1p(1 / lambda / math.E)` is better or worse than `-math.Log1p(math.Exp(-math.Log(lambda) - 1))`. I think you're probably right, that it's better. (They look to be about 1 ULP different in the end for reasonable small alpha.)

Given this, keep yours. Sorry.

romanwerpachowski:
PTAL.

kortschak:
You can go ahead and merge (please squash with the following commit message):

```
stat/distuv: provide full test coverage for Gamma distribution and small fixes

* Removed redundant calculations in the small alpha version of Rand.
* Lowered the threshold for small alpha to 0.2 in accordance with
  https://www4.stat.ncsu.edu/~rmartin/Codes/rgamss.R.
* Added the option to pass integration bounds to PDF normalisation check.
* Other small fixes.
```

romanwerpachowski:
Thank you!

kortschak:
For future, the single header line goes in the squash subject (no PR number link) and the body goes in the body. We try to match the convention used in the Go project.

PR #1358, opened 2020-06-16
stat/distuv: run tests in parallel

romanwerpachowski:
Please take a look.


romanwerpachowski:
> Please squash to make the commit message `stat/distuv: run tests in parallel` (remove the `(#1358)` that github will helpfully add).
> 
> For future, if you make the commit message what you would have as the PR title, and there's a single commit, the PR gets that title by default. Commit messages should be in that form, and then there's no need to squash unless the PR has multiple commits that need to be squashed either during review or otherwise.

Thanks, will do that in the future.

PR #1359, opened 2020-06-16
stat/distuv: Add integration bounds to PDF normalisation checks and tighten tolerances where possible

romanwerpachowski:
Please take a look.


romanwerpachowski:
Thanks.

PR #1360, opened 2020-06-18
stat/distuv: provide complete Gumbel test coverage and fix its documentation

romanwerpachowski:
Please take a look.

### DIFF
--- a/stat/distuv/gumbel.go
+++ b/stat/distuv/gumbel.go
@@ -57,7 +57,7 @@ func (g GumbelRight) Mean() float64 {
 	return g.Mu + g.Beta*eulerMascheroni
 }
 
-// Median returns the median of the normal distribution.
+// Median returns the median of the Gumbel distribution.
 func (g GumbelRight) Median() float64 {
 	return g.Mu - g.Beta*math.Log(math.Ln2)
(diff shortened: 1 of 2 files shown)


romanwerpachowski:
Github iOS app merged it without letting me change the commit message :(

kortschak:
The Clippit disease it widespread; I would love to ban easy to use but incorrect access to interacting with the tree, but it's not possible. Does the app, not allow squash merging? If it doesn't, and a commit message change has been requested, please don't use it, it breaks the commit convention and this cannot be fixed.

To avoid the problem, also please try to make sure the commit message (and not necessarily the PR title, though this will follow if it's a single commit PR) follows the convention. The convention is detailed [here](https://golang.org/doc/contribute.html#commit_messages). Also see [here](https://kortschak.io/courses/git_workshop/communication/#commit-messages) for discussion and links about why this is important.

romanwerpachowski:
Sure, thanks for taking the time to explain.

The iOS app gives the option to squash the commits and merge, but this option doesn't allow to edit the commit message (unlike the web app). Maybe there is another option which allows editing the message, but I will just stick to the web app.

PR #1363, opened 2020-06-22
stat/distuv: provide complete InverseGamma test coverage

romanwerpachowski:
Please take a look.

### DIFF
--- a/stat/distuv/inversegamma.go
+++ b/stat/distuv/inversegamma.go
@@ -35,9 +35,6 @@ func (g InverseGamma) CDF(x float64) float64 {
 	if x < 0 {
 		return 0
 	}
-	// TODO(btracey): Replace this with a direct call to the upper regularized
-	// gamma function if mathext gets it.
-	//return 1 - mathext.GammaInc(g.Alpha, g.Beta/x)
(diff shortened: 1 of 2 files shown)


stat/distuv/inversegamma.go
```diff
@@ -35,9 +35,6 @@ func (g InverseGamma) CDF(x float64) float64 {
 	if x < 0 {
 		return 0
 	}
-	// TODO(btracey): Replace this with a direct call to the upper regularized
-	// gamma function if mathext gets it.
-	//return 1 - mathext.GammaInc(g.Alpha, g.Beta/x)
```
kortschak, line 40: @btracey This was as it was when the file was originally added. Which do you want, `1 - mathext.GammaIncReg(g.Alpha, g.Beta/x)` or `mathext.GammaIncRegComp(g.Alpha, g.Beta/x)` since both are available.

This should be a separate PR though.

romanwerpachowski: I'll revert it and let's fix it in a separate PR.

romanwerpachowski:
PTAL.

PR #1366, opened 2020-06-24
stat/distuv: add Score method to Triangle and Uniform

romanwerpachowski:
Also tests and fix Laplace Score.

Please take a look.

### DIFF
--- a/stat/distuv/laplace.go
+++ b/stat/distuv/laplace.go
@@ -183,7 +183,7 @@ func (l Laplace) Rand() float64 {
 // For more information, see https://en.wikipedia.org/wiki/Score_%28statistics%29.
 //
 // Special cases:
-//  Score(0) = [0, -0.5/l.Scale]
+//  Score(Mu) = [NaN, -1/l.Scale]
 func (l Laplace) Score(deriv []float64, x float64) []float64 {
 	if deriv == nil {
 		deriv = make([]float64, l.NumParameters())
@@ -196,8 +196,6 @@ func (l Laplace) Score(deriv []float64, x float64) []float64 {
 		deriv[0] = 1 / l.Scale
 	} else if diff < 0 {
 		deriv[0] = -1 / l.Scale
-	} else if diff == 0 {
-		deriv[0] = 0
 	} else {
 		// must be NaN
 		deriv[0] = math.NaN()
--- a/stat/distuv/laplace.go
+++ b/stat/distuv/laplace.go
@@ -104,17 +104,6 @@ func (l Laplace) LogProb(x float64) float64 {
 	return -math.Ln2 - math.Log(l.Scale) - math.Abs(x-l.Mu)/l.Scale
 }
 
-// MarshalParameters implements the ParameterMarshaler interface
-func (l Laplace) MarshalParameters(p []Parameter) {
-	if len(p) != l.NumParameters() {
-		panic(badLength)
-	}
-	p[0].Name = "Mu"
-	p[0].Value = l.Mu
-	p[1].Name = "Scale"
-	p[1].Value = l.Scale
-}
-
 // Mean returns the mean of the probability distribution.
 func (l Laplace) Mean() float64 {
 	return l.Mu
@@ -133,6 +122,21 @@ func (l Laplace) Mode() float64 {
 // NumParameters returns the number of parameters in the distribution.
 func (l Laplace) NumParameters() int {
 	return 2
+}
+
+// parameters returns the parameters of the distribution.
+func (l Laplace) parameters(p []Parameter) []Parameter {
+	nParam := l.NumParameters()
+	if p == nil {
+		p = make([]Parameter, nParam)
+	} else if len(p) != nParam {
+		panic(badLength)
+	}
+	p[0].Name = "Mu"
+	p[0].Value = l.Mu
+	p[1].Name = "Scale"
+	p[1].Value = l.Scale
+	return p
 }
 
 // Quantile returns the inverse of the cumulative probability distribution.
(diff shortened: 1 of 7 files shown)


kortschak:
For the sake of diff reduction, making the name changes in place in this PR would be best. The order of the functions to match alphabetical ordering can happen later.

The failures are due to the now-unexported methods not being tested. Those would need to be added.

romanwerpachowski:
PTAL.

stat/distuv/triangle.go
```diff
@@ -79,6 +79,7 @@ func (t Triangle) Median() float64 {
 	if t.c >= (t.a+t.b)/2 {
 		return t.a + math.Sqrt((t.b-t.a)*(t.c-t.a)/2)
 	}
+
```
kortschak, line 82: Please revert this blank line insertion.

In general the convention for returning if-else constructs is `if cond { return res }; return altres`, which in other languages would be written as `if cond { return res } else { return altres }` (there are historical reasons for this relating to return value compiler checks that are not relevant in modern Go code, but the convention remains). This means that the alternative branch is part of the stanza, just as is would be if there were an explicit `else`.

stat/distuv/triangle.go
```diff
@@ -134,6 +135,66 @@ func (t Triangle) Rand() float64 {
...
+// of the log-likelihood at x with respect to the parameters
+//  (∂/∂θ) log(p(x;θ))
+// If deriv is non-nil, len(deriv) must equal the number of parameters otherwise
+// Score will panic, and the derivative is stored in-place into deriv. If deriv
+// is nil a new slice will be allocated and returned.
+//
+// The order is [∂LogProb / ∂Mu, ∂LogProb / ∂Sigma].
+//
+// For more information, see https://en.wikipedia.org/wiki/Score_%28statistics%29.
+func (t Triangle) Score(deriv []float64, x float64) []float64 {
```
kortschak, line 149: This is a new addition; it and its tests should be another PR (possibly the fix to `Laplace` too).

stat/distuv/uniform.go
```diff
@@ -116,6 +120,40 @@ func (u Uniform) Rand() float64 {
...
+// of the log-likelihood at x with respect to the parameters
+//  (∂/∂θ) log(p(x;θ))
+// If deriv is non-nil, len(deriv) must equal the number of parameters otherwise
+// Score will panic, and the derivative is stored in-place into deriv. If deriv
+// is nil a new slice will be allocated and returned.
+//
+// The order is [∂LogProb / ∂Mu, ∂LogProb / ∂Sigma].
+//
+// For more information, see https://en.wikipedia.org/wiki/Score_%28statistics%29.
+func (u Uniform) Score(deriv []float64, x float64) []float64 {
```
kortschak, line 134: Same here.

stat/distuv/laplace.go
```diff
@@ -203,7 +203,7 @@ func (l Laplace) Score(deriv []float64, x float64) []float64 {
 		deriv[0] = math.NaN()
 	}
 
-	deriv[1] = math.Abs(diff)/(l.Scale*l.Scale) - 0.5/(l.Scale)
+	deriv[1] = math.Abs(diff)/(l.Scale*l.Scale) - 1/l.Scale
```
kortschak, line 206: I would do this in the PR that adds `Score` to the other distributions.

stat/distuv/triangle.go
romanwerpachowski: If I don't add the Score methods, I can't test parameters and setParameters using the existing testDerivativeParam function.

stat/distuv/laplace.go
romanwerpachowski: If I don't add this fix, the function which tests parameters and setParameters fails.

stat/distuv/triangle.go
kortschak: Mmm, OK. I think that changes the focus of the PR then, which is fine. It would become

```
stat/distuv: add Score method to Triangle and Uniform

Also tests and fix Laplace Score.
```

romanwerpachowski: Done.

stat/distuv/triangle.go
romanwerpachowski: Done.

romanwerpachowski:
PTAL.

stat/distuv/laplace.go
```diff
@@ -199,7 +203,7 @@ func (l Laplace) Score(deriv []float64, x float64) []float64 {
...
```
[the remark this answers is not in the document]
romanwerpachowski: Good catch, and you actually helped me to find another bug ;-)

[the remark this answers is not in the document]
romanwerpachowski: PTAL.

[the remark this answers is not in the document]
kortschak: Yeah, I was thinking that might be the case.

stat/distuv/laplace.go
```diff
@@ -192,14 +196,12 @@ func (l Laplace) Score(deriv []float64, x float64) []float64 {
 		deriv[0] = 1 / l.Scale
 	} else if diff < 0 {
 		deriv[0] = -1 / l.Scale
-	} else if diff == 0 {
-		deriv[0] = 0
 	} else {
 		// must be NaN
 		deriv[0] = math.NaN()
 	}
```
kortschak, line 202: Is this described as the convention somewhere?

romanwerpachowski: I didn't find anything like that. If the derivative doesn't exist, we should return NaN IMHO.

PR #1367, opened 2020-06-29
stat/distuv: fix Weibull LogProb and test function absEq

romanwerpachowski:
Modifies absEq to catch cases when one or both arguments is NaN.


kortschak:
Weird failure on arm64. Re-run and cleared.

romanwerpachowski:
Thanks.

PR #1369, opened 2020-06-30
stat/distuv: add ScoreInput implementations for Uniform and Triangle

romanwerpachowski:
Also fixes a bug in Laplace ScoreInput, and tests existing ScoreInput implementations.

Please take a look.

### DIFF
--- a/stat/distuv/triangle.go
+++ b/stat/distuv/triangle.go
@@ -194,6 +194,22 @@ func (t Triangle) Score(deriv []float64, x float64) []float64 {
 	return deriv
 }
 
+// ScoreInput returns the score function with respect to the input of the
+// distribution at the input location specified by x. The score function is the
+// derivative of the log-likelihood
+//  (d/dx) log(p(x)) .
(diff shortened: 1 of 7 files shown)


stat/distuv/triangle.go
```diff
@@ -194,6 +194,22 @@ func (t Triangle) Score(deriv []float64, x float64) []float64 {
 	return deriv
 }
 
+// ScoreInput returns the score function with respect to the input of the
+// distribution at the input location specified by x. The score function is the
+// derivative of the log-likelihood
+//  (d/dx) log(p(x)) .
+// Special cases:
+//  ScoreInput(t.c) = NaN
```
kortschak, line 202: `.c` never shows up in the godoc, so this should probably just be `c`, possibly with a reminder that c is the mode.

Is it worth noting it is not defined outside of (a, b)?

romanwerpachowski: It is. Done. PTAL.

PR #1374, opened 2020-07-01
stat/distuv: provide complete test coverage for Poisson

romanwerpachowski:
Also uses mathext.GammaIncReg to implement Poisson Survival.

Please take a look.

### DIFF
--- a/stat/distuv/poisson.go
+++ b/stat/distuv/poisson.go
@@ -129,7 +129,10 @@ func (p Poisson) StdDev() float64 {
 
 // Survival returns the survival function (complementary CDF) at x.
 func (p Poisson) Survival(x float64) float64 {
-	return 1 - p.CDF(x)
+	if x < 0 {
+		return 1
+	}
+	return mathext.GammaIncReg(math.Floor(x+1), p.Lambda)
 }
 
 // Variance returns the variance of the probability distribution.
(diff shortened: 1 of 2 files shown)


stat/distuv/poisson.go
```diff
@@ -129,7 +129,10 @@ func (p Poisson) StdDev() float64 {
 
 // Survival returns the survival function (complementary CDF) at x.
 func (p Poisson) Survival(x float64) float64 {
-	return 1 - p.CDF(x)
```
vladimir-ch, line 132: I think the original is better. It is obviously correct and consistent with the documentation.

romanwerpachowski: But less accurate for x close to xm.

romanwerpachowski:
PTAL.

romanwerpachowski:
*ping*

romanwerpachowski:
@vladimir-ch Is that OK now?

stat/distuv/poisson_test.go
```diff
@@ -125,7 +126,36 @@ func testPoisson(t *testing.T, p Poisson, i int) {
...
+	}
+	logProb := p.LogProb(-0.0001)
+	if !math.IsInf(logProb, -1) {
+		t.Errorf("Mismatch in LogProb for x < 0: got %v, want -Inf", logProb)
+	}
+	logProb = p.LogProb(1.5)
+	if !math.IsInf(logProb, -1) {
+		t.Errorf("Mismatch in LogProb for non-integer x: got %v, want -Inf", logProb)
+	}
+	for _, xx := range x {
```
kortschak, line 154: Not worth changing, but we tend to avoid the double letter convention for name collisions (except in the blas and lapack code I think). This would probably be `v`.

romanwerpachowski: Thanks, noted.

PR #1382, opened 2020-07-03
graph/path: properly report negative cycles in Shortest.To

kortschak:
This actually does fix the issue with reporting negative cycles (within reason - Weight will still refuse to be definitive). The previous change adding the negative cycles did nothing since the test rejected them early.

Please take a look.

### DIFF
--- a/graph/path/bellman_ford_moore_test.go
+++ b/graph/path/bellman_ford_moore_test.go
@@ -26,9 +26,7 @@ func TestBellmanFordFrom(t *testing.T) {
 			if ok {
 				t.Errorf("%q: expected negative cycle", test.Name)
 			}
-			continue
-		}
-		if !ok {
+		} else if !ok {
 			t.Fatalf("%q: unexpected negative cycle", test.Name)
 		}
 
@@ -41,7 +39,7 @@ func TestBellmanFordFrom(t *testing.T) {
 			t.Errorf("%q: unexpected weight from To: got:%f want:%f",
 				test.Name, weight, test.Weight)
 		}
-		if weight := pt.WeightTo(test.Query.To().ID()); weight != test.Weight {
+		if weight := pt.WeightTo(test.Query.To().ID()); !math.IsInf(test.Weight, -1) && weight != test.Weight {
 			t.Errorf("%q: unexpected weight from Weight: got:%f want:%f",
 				test.Name, weight, test.Weight)
 		}
--- a/graph/path/internal/testgraphs/shortest.go
+++ b/graph/path/internal/testgraphs/shortest.go
@@ -594,7 +594,13 @@ var ShortestPathTests = []struct {
 		HasNegativeWeight: true,
 		HasNegativeCycle:  true,
 
-		Query: simple.Edge{F: simple.Node(0), T: simple.Node(1)},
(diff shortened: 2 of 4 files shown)


graph/path/shortest.go
```diff
@@ -103,6 +109,14 @@ func (p *Shortest) add(u graph.Node) int {
 func (p Shortest) set(to int, weight float64, mid int) {
 	p.dist[to] = weight
 	p.next[to] = mid
+	if p.cycCosts != nil {
```
kortschak, line 112: I think this can be optimised by requiring `p.cycCosts != nil && weight < 0`, since only negative weights will ever lead to a negative cycle, so we should only bother storing things when they have reached negativity.

kortschak: Actually, it can just be `weight < 0`, since if we have the possibility of negative weights (we are using Bellman-Ford), we have already made the map.

graph/path/internal/testgraphs/shortest.go
```diff
@@ -644,7 +658,13 @@ var ShortestPathTests = []struct {
 		HasNegativeWeight: true,
 		HasNegativeCycle:  true,
 
-		Query: simple.Edge{F: simple.Node(0), T: simple.Node(3)},
+		Query:  simple.Edge{F: simple.Node(0), T: simple.Node(3)},
+		Weight: math.Inf(-1),
```
romanwerpachowski, line 662: If this is a "special" value, perhaps it should be a constant.

graph/path/bellman_ford_moore_test.go
```diff
@@ -41,7 +39,7 @@ func TestBellmanFordFrom(t *testing.T) {
 			t.Errorf("%q: unexpected weight from To: got:%f want:%f",
 				test.Name, weight, test.Weight)
 		}
-		if weight := pt.WeightTo(test.Query.To().ID()); weight != test.Weight {
+		if weight := pt.WeightTo(test.Query.To().ID()); !math.IsInf(test.Weight, -1) && weight != test.Weight {
```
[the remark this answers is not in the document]
kortschak: Negative infinite path weights fall out of finding shortest (lowest cost) paths in a graph with cycles with single traversal costs that are negative; repeated traversal of the cycle reduces the total cost of the path by the single traversal cost each time. Since there is no constraint to limit the number of traversals you end up with -Inf.

graph/path/internal/testgraphs/shortest.go
kortschak: It's not special and since these are test cases, things should be explicit.

graph/path/bellman_ford_moore_test.go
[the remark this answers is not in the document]
romanwerpachowski: Is there a comment somewhere saying this? Unless you think the reader should know this.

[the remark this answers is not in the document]
kortschak: The reader should know this, though there are examples that demonstrates how to find them: https://godoc.org/gonum.org/v1/gonum/graph/path#example-BellmanFordFrom--Negativecycles and https://godoc.org/gonum.org/v1/gonum/graph/path#example-FloydWarshall--Negativecycles.

Issue #1378, opened 2020-07-02
stat: consider adding an example for KolmogorovSmirnov


sbinet:
this exported function has no example for now.

(we should also probaly move all the examples from `stat_example_test.go` to an `example_stat_test.go` file under a `package stat_test` clause)

PR #1384, opened 2020-07-03
stat: add an example for the KolmogorovSmirnov test

sbinet:
Fixes gonum/gonum#1378.

Please take a look.


stat/stat_example_test.go
```diff
@@ -6,12 +6,77 @@ package stat_test
...
+		-1.52, -0.06, -1.02, 1.06, 0.60, 1.15, 1.92, -0.06, -0.19, 0.67,
+		0.29, 0.58, 0.02, 2.18, -0.04, -0.13, -0.79, -1.28, -1.41, -0.23,
+		0.65, -0.26, -0.17, -1.53, -1.69, -1.60, 0.09, -1.11, 0.30, 0.71,
+		-0.88, -0.03, 0.56, -3.68, 2.40, 0.62, 0.52, -1.25, 0.85, -0.09,
+		-0.23, -1.16, 0.22, -1.68, 0.50, -0.35, -0.35, -0.33, -0.24, 0.25,
+	}
+
+	sort.Float64s(xobs)
+
+	var (
```
btracey, line 39: I think this example is confusing. I would instead do somethng like:
1) Compare 100 samples of a standard normal with 1000 (should match)
2) Compare 100 samples of standard normal with Students-T (don't match)
3) Compare 100 samples with importance-sampling corrected Students-T (should match)

sbinet: In what way do you find it confusing?
I'm obviously biaised but I believe it shows the typical question an experimentalist would want to answer (is my data close to some PDF?)

That doesn't preclude, of course, to add the examples you're talking about.

romanwerpachowski: I think the examples proposed by @btracey should be turned into unit tests. They verify that the implementation is correct.

stat/stat_example_test.go
```diff
@@ -6,12 +6,77 @@ package stat_test
...
+		exp[i] = norm.CDF(v)
+		diff[i] = math.Abs(xedf[i] - exp[i])
+	}
+
+	dist := stat.KolmogorovSmirnov(xobs, xedf, xobs, exp)
+	fmt.Printf("stat:     %d\n", n)
+	fmt.Printf("diff:     %2.3f\n", floats.Max(diff))
+	fmt.Printf("dist-KS:  %2.3f\n", dist)
+
+	// critical value @95% confidence level
```
romanwerpachowski, line 59: `// Critical value at the 95% confidence level.`

stat/stat_example_test.go
```diff
@@ -6,12 +6,77 @@ package stat_test
...
+	var h0 string
+	switch {
+	case dist < crit:
+		h0 = "do not reject"
+	default:
+		h0 = "reject"
+	}
+	fmt.Printf("H0:       %s\n", h0)
+
+	// Output:
```
romanwerpachowski, line 72: `// Expected output:`

stat/stat_example_test.go
sbinet: done.

stat/stat_example_test.go
sbinet: the `// Output:\n` comment is the magic comment to tell the Go tool this example has (indeed) an expected output and should be run as part of the `go test` suite.

romanwerpachowski: I see :)

stat/stat_example_test.go
btracey: The reason why I think it's confusing is that this function implements a two-sided KS test, while the example is to implement a one-sided KS test. Wouldn't it be better to add a one-sided function and then modify this example?

stat/stat_example_test.go
```diff
@@ -6,12 +6,77 @@ package stat_test
...
+	crit := 1.36 / math.Sqrt(float64(n))
+	fmt.Printf("dist(3σ): %2.3f\n", crit)
+
+	var h0 string
+	switch {
+	case dist < crit:
+		h0 = "do not reject"
+	default:
+		h0 = "reject"
+	}
```
kortschak, line 69: It seems odd to use a switch here.

sbinet:
PTAL.

sbinet:
wrt @btracey 's comment: should we first introduce/split KS test into 2 versions?

sbinet:
ie:
```go
func KolmogorovSmirnov(x, xw, y, yw []float64) float64 { ... }

func KolmogorovSmirnov1Sided(x, cdf []float64) float64 { ... }
```
?
(not particularly attached to the naming I chose, feel free to refine/improve)

kortschak:
Perhaps you should wait for @btracey.

romanwerpachowski:
> ie:
> 
> ```go
> func KolmogorovSmirnov(x, xw, y, yw []float64) float64 { ... }
> 
> func KolmogorovSmirnov1Sided(x, cdf []float64) float64 { ... }
> ```
> 
> ?
> (not particularly attached to the naming I chose, feel free to refine/improve)

I would split to avoid user confusion.

vladimir-ch:
@sbinet What do you want to do with this PR?

PR #1389, opened 2020-07-08
graph/path: add doc comments for unexported path construction methods

kortschak:
Please take a look.

### DIFF
--- a/graph/path/shortest.go
+++ b/graph/path/shortest.go
@@ -55,11 +55,13 @@ type Shortest struct {
 	// returned the Shortest value.
 	hasNegativeCycle bool
 
-	// cycCosts holds costs between
-	// pairs of nodes to report negative
-	// cycles. cycCosts is only used
-	// if it is non-nil.
-	cycCosts map[[2]int]float64
+	// negCosts holds negative costs
+	// between pairs of nodes to report
+	// negative cycles.
+	// negCosts must be initialised by
+	// routines that can handle negative
+	// edge weights.
+	negCosts map[[2]int]float64
 }
 
 // newShortestFrom returns a shortest path tree for paths from u
@@ -115,9 +117,9 @@ func (p Shortest) set(to int, weight float64, mid int) {
 	p.dist[to] = weight
 	p.next[to] = mid
 	if weight < 0 {
-		c, ok := p.cycCosts[[2]int{mid, to}]
+		c, ok := p.negCosts[[2]int{mid, to}]
 		if !ok {
-			p.cycCosts[[2]int{mid, to}] = weight
+			p.negCosts[[2]int{mid, to}] = weight
 		} else if weight < c {
 			// The only ways that we can have a new weight that is
 			// lower than the previous weight is if either the edge
@@ -125,7 +127,7 @@ func (p Shortest) set(to int, weight float64, mid int) {
 			// the edge is reachable from a negative cycle.
 			// Either way the reported path is returned with a
 			// negative infinite path weight.
-			p.cycCosts[[2]int{mid, to}] = math.Inf(-1)
+			p.negCosts[[2]int{mid, to}] = math.Inf(-1)
 		}
 	}
 }
@@ -160,7 +162,7 @@ func (p Shortest) To(vid int64) (path []graph.Node, weight float64) {
 		seen.Add(from)
 		for to != from {
 			next := p.next[to]
-			if math.IsInf(p.cycCosts[[2]int{next, to}], -1) {
+			if math.IsInf(p.negCosts[[2]int{next, to}], -1) {
 				weight = math.Inf(-1)
 			}
 			if seen.Has(to) {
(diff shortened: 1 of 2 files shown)


graph/path/shortest.go
```diff
@@ -106,6 +109,8 @@ func (p *Shortest) add(u graph.Node) int {
 	return idx
 }
 
+// set sets the weight of the path from the node in p.nodes indexed by to to the node
+// indexed by to.
```
romanwerpachowski, line 113: "to" or "mid"?

graph/path/shortest.go
```diff
@@ -238,6 +249,9 @@ var (
 	defacedBits = math.Float64bits(defaced)
 )
 
+// newAllShortest returns a all-pairs shortest path forrest for paths with the
```
romanwerpachowski, line 252: a -> an

romanwerpachowski, line 252: "forrest"?

graph/path/shortest.go
kortschak: Fixed

graph/path/shortest.go
kortschak: Fixed

kortschak: Full of trrees.

romanwerpachowski: That's a thing in graph theory?

kortschak: The typo is a Dan thing. I have corrected it.

romanwerpachowski: Sorry, missed the joke :D

graph/path/shortest.go
```diff
@@ -114,6 +119,12 @@ func (p Shortest) set(to int, weight float64, mid int) {
 		if !ok {
 			p.cycCosts[[2]int{mid, to}] = weight
 		} else if weight < c {
+			// The only ways that we can have a new weight that is
```
vladimir-ch, line 122: way

graph/path/shortest.go
```diff
@@ -260,15 +274,21 @@ func newAllShortest(nodes []graph.Node, forward bool) AllShortest {
 	}
 }
 
+// at returns a slice of node indexed into p.nodes for nodes that are mid points
```
vladimir-ch, line 277: This sentence is quite hard to parse so not sure but "a slice of nodes" (plural)?

graph/path/shortest.go
```diff
@@ -62,6 +62,9 @@ type Shortest struct {
 	cycCosts map[[2]int]float64
 }
 
+// newShortestFrom returns a shortest path tree for paths from u
+// intialised with the given nodes. The nodes held by the returned
```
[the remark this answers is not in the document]
kortschak: Fixed

graph/path/shortest.go
kortschak: No, there are two ways.

graph/path/shortest.go
kortschak: A common typo for me is -ed (_p.p._)-> -es (_pl._). I noticed what I though was this here and miscorrected the -es to -ed. I've fixed this and hopefully also clarified it (and below). It now means what I want it to mean, but still is difficult to parse.

kortschak:
PTAL

graph/path/shortest.go
romanwerpachowski: I would write "the only two ways" then.

graph/path/shortest.go
```diff
@@ -55,13 +55,18 @@ type Shortest struct {
 	// returned the Shortest value.
 	hasNegativeCycle bool
 
-	// cycCosts holds costs between
-	// pairs of nodes to report negative
-	// cycles. cycCosts is only used
-	// if it is non-nil.
-	cycCosts map[[2]int]float64
+	// negCosts holds negative costs
```
romanwerpachowski, line 58: Why are these lines so short?

graph/path/shortest.go
```diff
@@ -264,19 +282,30 @@ func (p *ShortestAlts) add(u graph.Node) int {
...
 	p.next[to] = []int{mid}
 	if weight < 0 {
-		c, ok := p.cycCosts[[2]int{mid, to}]
+		c, ok := p.negCosts[[2]int{mid, to}]
 		if !ok {
-			p.cycCosts[[2]int{mid, to}] = weight
+			p.negCosts[[2]int{mid, to}] = weight
 		} else if weight < c {
-			p.cycCosts[[2]int{mid, to}] = math.Inf(-1)
+			// The only ways that we can have a new weight that is
```
romanwerpachowski, line 295: "two ways"

graph/path/shortest.go
```diff
@@ -638,6 +676,10 @@ func (p AllShortest) AllBetween(uid, vid int64) (paths [][]graph.Node, weight fl
 	return paths, weight
 }
 
+// allBetween recursively constructs a slice of paths extending from the node
+// indexed into p.nodes by from to the node indexed by to. The len(seen) must
```
romanwerpachowski, line 680: I'm not sure if "The" is needed here.

graph/path/shortest.go
kortschak: This is for a couple of loosely related reasons; the first is that it is like this for the same reason that newsprint is in narrow column and the second is that formatting is wysiswyg in source code (that is godoc has no power here). In code, comments that explain hairy sections should make life easier; this is one way to achieve that.

graph/path/shortest.go
kortschak: "only" is used here as an emphatic. I think this is the correct wording.

graph/path/shortest.go
kortschak: Hmmm. I read this as "The length of seen must ...". However, this isn't user-facing, so I'm not particularly fussed. I'll change it.

graph/path/shortest.go
romanwerpachowski: For me it makes the text harder, not easier to read. But I won't insist.

kortschak: Maybe this is a generational thing. I grew up reading newspapers.

PR #1390, opened 2020-07-08
mat: update docs for SVD.SolveTo and SVD.SolveVecTo

vladimir-ch:
This updates the documentation as promised in #1316.

Please take a look.

### DIFF
--- a/mat/svd.go
+++ b/mat/svd.go
@@ -272,13 +272,24 @@ func (svd *SVD) VTo(dst *Dense) {
 	dst.Copy(tmp.T())
 }
 
-// SolveTo calculates the least-squares solution to a linear matrix equation,
-// AX = B, using a rank-truncated SVD. The result is placed in dst.
-// It returns the residuals calculated from the complete SVD. For this value to be
-// valid the factorization must have been performed with at least SVDFullU.
-// The decomposition must have been factorized computing both the U and V singular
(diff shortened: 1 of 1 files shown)


mat/svd.go
```diff
@@ -272,13 +272,23 @@ func (svd *SVD) VTo(dst *Dense) {
...
-// AX = B, using a rank-truncated SVD. The result is placed in dst.
-// It returns the residuals calculated from the complete SVD. For this value to be
-// valid the factorization must have been performed with at least SVDFullU.
-// The decomposition must have been factorized computing both the U and V singular
-// vectors, and dst must be either empty or have a row dimension equal to rank.
-// The supplied rank must be a positive end-slice index into svd.Values.
+// SolveTo calculates the minimum-norm solution to a linear least squares problem
+//  minimize over n-element vectors x: |b - A*x|_2 and |x|_2
+// where b is a given m-element vector, using the SVD of m×n matrix A stored in
+// the receiver. A may be rank-deficient, that is, the given effective rank can be
```
vladimir-ch, line 278: Should the text suggest explicitly using SVD.Rank?

kortschak: I think it probably should.

vladimir-ch: Done.

vladimir-ch:
PTAL

Issue #1395, opened 2020-07-12
all: extract examples into a 'package xxx_test' file


sbinet:
there are a couple of examples which are written as part of the package they test.
_i.e.:_ they are in a file with a `package xxx` clause.

they should be in a file with a `package xxx_test` clause.

here's the list:
```
$> for i in `git grep 'func Example' -- . | cut -d: -f 1 | sort | uniq`; do git grep 'package ' -- $i | egrep -v '_test$'; done
cmplxs/examples_test.go:5:package cmplxs
floats/examples_test.go:5:package floats
graph/community/bisect_test.go:5:package community
stat/sampleuv/example_burnin_test.go:5:package sampleuv
stat/sampleuv/example_rate_test.go:5:package sampleuv
```

PR #1396, opened 2020-07-12
all: put examples into a xxx_test package clause

sbinet:
Fixes gonum/gonum#1395.

Please take a look.


stat/sampleuv/example_burnin_test.go
```diff
@@ -29,9 +32,9 @@ func ExampleMetropolisHastings_burnin() {
 	// typical problems, if Sigma is too small, it takes a lot of samples
 	// to move around the distribution. If Sigma is too large, it can be hard
 	// to find acceptable samples.
-	proposal := ProposalDist{Sigma: 0.2}
+	proposal := sampleuv.ProposalDist{Sigma: 0.2}
```
kortschak, line 35: This is actually defined in this file. It's also used in the rate example which presumably also needs to be made a sampleuv_test package member.

sbinet:
PTAL

Issue #1393, opened 2020-07-12
ci: remove golangci-lint from .travis


sbinet:
`golangci-lint` is currently run:
- as a dedicated github action
- as part of a shell script, `check-lint`:
  https://github.com/gonum/gonum/blob/27220e4f5c22c596d48f0e07e48c301b57b7e0df/.travis/script.d/check-lint.sh

`check-lint.sh` is -AFAICT- only run on Linux (so amd64 and arm, currently).

perhaps we could remove the `check-lint.sh` script?
or only run it on linux+arm?

kortschak:
`check-lint.sh` doesn't run on arm64 because the git version on Travis' arm64 builders doesn't support the necessary commands.

Does the action correctly only lint on differences? It didn't a while back, but I don't know if that has been fixed.

sbinet:
ah. right.
I'll probably start a PR to remove all the lint, then. (holidays! you can't spend all your time in the garden or outdoor.)

kortschak:
golangci/golangci-lint-action#16, so I guess the script can be removed.

kortschak:
I would also consider making it a required check.

sbinet:
> I would also consider making it a required check.

I am not sure to know where that is specified.

kortschak:
I can do that later.

kortschak:
The linting check is now required for changed lines.

PR #1397, opened 2020-07-12
ci: remove golangci-lint from travis checks

sbinet:
golangci-lint is also run as part of a GitHub-Actions.

Fixes gonum/gonum#1393.

Please take a look.


sbinet:
done

PR #1399, opened 2020-07-13
interp: add Akima splines and generic piecewise cubic interpolators

romanwerpachowski:
For now all code is in interp.go, but the file starts getting rather large. Should I split it?

Please take a look.

### DIFF
--- a/interp/interp.go
+++ b/interp/interp.go
@@ -12,7 +12,7 @@ import (
 )
 
 const (
-	differentLengths        = "interp: xs and ys have different lengths"
+	differentLengths        = "interp: input slices have different lengths"
 	tooFewPoints            = "interp: too few points for interpolation"
 	xsNotStrictlyIncreasing = "interp: xs values not strictly increasing"
 )
@@ -86,8 +86,10 @@ func (pl *PiecewiseLinear) Fit(xs, ys []float64) error {
 		}
 		pl.slopes[i] = (ys[i+1] - ys[i]) / dx
 	}
-	pl.xs = xs
-	pl.ys = ys
+	pl.xs = make([]float64, n)
+	pl.ys = make([]float64, n)
+	copy(pl.xs, xs)
+	copy(pl.ys, ys)
 	return nil
 }
 
@@ -137,8 +139,10 @@ func (pc *PiecewiseConstant) Fit(xs, ys []float64) error {
 			return errors.New(xsNotStrictlyIncreasing)
 		}
 	}
-	pc.xs = xs
-	pc.ys = ys
+	pc.xs = make([]float64, n)
+	pc.ys = make([]float64, n)
+	copy(pc.xs, xs)
+	copy(pc.ys, ys)
 	return nil
 }
 
@@ -171,30 +175,70 @@ type PiecewiseCubic struct {
 	// for xs[i] <= x < xs[i + 1] is defined as
 	//   sum_{k = 0}^3 coeffs.At(i, k) * (x - xs[i])^k
 	// To guarantee left-continuity, coeffs.At(i, 0) == ys[i].
-	coeffs mat.Dense
+	coeffs *mat.Dense
 
 	// Last interpolated Y value, corresponding to xs[len(xs) - 1].
 	lastY float64
 }
 
 // Predict returns the interpolation value at x.
-func (pc *PiecewiseCubic) Predict(x float64) float64 {
+func (pc PiecewiseCubic) Predict(x float64) float64 {
 	i := findSegment(pc.xs, x)
 	if i < 0 {
 		return pc.coeffs.At(0, 0)
 	}
+	m := len(pc.xs) - 1
 	// i < len(pc.xs)
 	if x == pc.xs[i] {
-		return pc.coeffs.At(0, 0)
+		if i < m {
+			return pc.coeffs.At(i, 0)
+		}
+		return pc.lastY
 	}
-	n := len(pc.xs)
-	if i == n-1 {
+	if i == m {
 		// x > pc.xs[i]
 		return pc.lastY
 	}
 	dx := x - pc.xs[i]
 	a := pc.coeffs.RawRowView(i)
 	return ((a[3]*dx+a[2])*dx+a[1])*dx + a[0]
+}
+
+// FitWithDerivatives fits a piecewise cubic predictor to (X, Y, dY/dX) value
+// triples provided as three slices.
+// It returns an error if len(xs) < 2, elements of xs are not strictly
+// increasing, len(xs) != len(ys) or len(xs) != len(dydxs).
+func (pc *PiecewiseCubic) FitWithDerivatives(xs, ys, dydxs []float64) error {
+	n := len(xs)
+	if len(ys) != n {
+		return errors.New(differentLengths)
+	}
+	if len(dydxs) != n {
+		return errors.New(differentLengths)
+	}
+	if n < 2 {
+		return errors.New(tooFewPoints)
+	}
+	m := n - 1
+	pc.coeffs = mat.NewDense(m, 4, nil)
+	for i := 0; i < m; i++ {
+		dx := xs[i+1] - xs[i]
+		if dx <= 0 {
+			return errors.New(xsNotStrictlyIncreasing)
+		}
+		dy := ys[i+1] - ys[i]
+		// a_0
+		pc.coeffs.Set(i, 0, ys[i])
+		// a_1
+		pc.coeffs.Set(i, 1, dydxs[i])
+		// Solve a linear equation system for a_2 and a_3.
+		pc.coeffs.Set(i, 2, (3*dy-(2*dydxs[i]+dydxs[i+1])*dx)/dx/dx)
+		pc.coeffs.Set(i, 3, (-2*dy+(dydxs[i]+dydxs[i+1])*dx)/dx/dx/dx)
+	}
+	pc.xs = make([]float64, n)
+	copy(pc.xs, xs)
+	pc.lastY = ys[m]
+	return nil
 }
 
 // findSegment returns 0 <= i < len(xs) such that xs[i] <= x < xs[i + 1], where xs[len(xs)]
--- a/interp/interp.go
+++ b/interp/interp.go
@@ -281,12 +281,7 @@ func (as *AkimaSpline) Fit(xs, ys []float64) error {
 	for i := 0; i < n; i++ {
 		wLeft := math.Abs(slopes[i+2] - slopes[i+3])
 		wRight := math.Abs(slopes[i+1] - slopes[i])
-		w := wLeft + wRight
-		if w > 0 {
-			dydxs[i] = (wLeft*slopes[i+1] + wRight*slopes[i+2]) / w
-		} else {
-			dydxs[i] = (slopes[i+1] + slopes[i+2]) / 2
-		}
+		dydxs[i] = weightedAverage(slopes[i+1], slopes[i+2], wLeft, wRight)
 	}
 	as.cubic.FitWithDerivatives(xs, ys, dydxs)
 	return nil
@@ -297,4 +292,14 @@ func (as *AkimaSpline) Fit(xs, ys []float64) error {
 // without checking.
 func findSegment(xs []float64, x float64) int {
 	return sort.Search(len(xs), func(i int) bool { return xs[i] > x }) - 1
+}
+
+// weightedAverage returns (v1 * w1 + v2 * w2) / (w1 + w2) for w1, w2 >= 0 (not checked).
+// If w1 == w2 == 0, it returns a simple average of v1 and v2.
+func weightedAverage(v1, v2, w1, w2 float64) float64 {
+	w := w1 + w2
+	if w > 0 {
+		return (v1*w1 + v2*w2) / w
+	}
+	return 0.5*v1 + 0.5*v2
 }
--- a/interp/interp.go
+++ b/interp/interp.go
@@ -6,6 +6,7 @@ package interp
 
 import (
 	"errors"
+	"math"
 	"sort"
 
 	"gonum.org/v1/gonum/mat"
@@ -239,6 +240,59 @@ func (pc *PiecewiseCubic) FitWithDerivatives(xs, ys, dydxs []float64) error {
 	copy(pc.xs, xs)
 	pc.lastY = ys[m]
 	return nil
+}
+
+// AkimaSplines is a left-continuous, piecewise cubic
+// 1-dimensional interpolator which can be fitted to
+// (X, Y) value pairs without providing derivatives.
+// See https://www.iue.tuwien.ac.at/phd/rottinger/node60.html for more details.
+type AkimaSplines struct {
+	PiecewiseCubic
+}
+
+// Fit fits a predictor to (X, Y) value pairs provided as two slices.
+// It returns an error if len(xs) < 2, elements of xs are not strictly
+// increasing or len(xs) != len(ys).
+// If len(xs) == 2, we set both derivatives dY/dX to the slope
+// (ys[1] - ys[0]) / (xs[1] - xs[0]).
+func (as *AkimaSplines) Fit(xs, ys []float64) error {
+	n := len(xs)
+	if len(ys) != n {
+		return errors.New(differentLengths)
+	}
+	dydxs := make([]float64, n)
+
+	if n == 2 {
+		slope := (ys[1] - ys[0]) / (xs[1] - xs[0])
+		dydxs[0] = slope
+		dydxs[1] = slope
+		return as.FitWithDerivatives(xs, ys, dydxs)
+	}
+	
+	m := n - 1
+	slopes := make([]float64, m + 4)
+	for i := 0; i < m; i++ {
+		dx := xs[i+1] - xs[i]
+		if dx <= 0 {
+			return errors.New(xsNotStrictlyIncreasing)
+		}
+		slopes[i + 2] = (ys[i+1] - ys[i]) / dx
+	}
+	slopes[1] = 2 * slopes[2] - slopes[3]
+	slopes[0] = 3 * slopes[2] - 2 * slopes[3]
+	slopes[m + 2] = 2 * slopes[n] - slopes[m]
+	slopes[m + 3] = 3 * slopes[n] - 2 * slopes[m]
+	for i := 0; i < n; i++ {
+		wLeft := math.Abs(slopes[i+2] - slopes[i+3])
+		wRight := math.Abs(slopes[i+1] - slopes[i])
+		w := wLeft + wRight
+		if w > 0 {
+			dydxs[i] = (wLeft * slopes[i+1] + wRight * slopes[i+2]) / w
+		} else {
+			dydxs[i] = (slopes[i+1] + slopes[i+2]) / 2
+		}
+	}
+	return as.FitWithDerivatives(xs, ys, dydxs)
 }
 
 // findSegment returns 0 <= i < len(xs) such that xs[i] <= x < xs[i + 1], where xs[len(xs)]
--- a/interp/interp.go
+++ b/interp/interp.go
@@ -28,8 +28,7 @@ type Predictor interface {
 type Fitter interface {
 	// Fit fits a predictor to (X, Y) value pairs provided as two slices.
 	// It panics if len(xs) < 2, elements of xs are not strictly increasing
-	// or len(xs) != len(ys). It returns an error if there is an unpredictable
-	// failure to fit.
+	// or len(xs) != len(ys). Returns an error if fitting fails.
 	Fit(xs, ys []float64) error
 }
 
@@ -69,8 +68,7 @@ type PiecewiseLinear struct {
 
 // Fit fits a predictor to (X, Y) value pairs provided as two slices.
 // It panics if len(xs) < 2, elements of xs are not strictly increasing
-// or len(xs) != len(ys). It returns an error if there is an unpredictable
-// failure to fit.
+// or len(xs) != len(ys). Always returns nil.
 func (pl *PiecewiseLinear) Fit(xs, ys []float64) error {
 	n := len(xs)
 	if len(ys) != n {
@@ -124,8 +122,7 @@ type PiecewiseConstant struct {
 
 // Fit fits a predictor to (X, Y) value pairs provided as two slices.
 // It panics if len(xs) < 2, elements of xs are not strictly increasing
-// or len(xs) != len(ys). It returns an error if there is an unpredictable
-// failure to fit.
+// or len(xs) != len(ys). Always returns nil.
 func (pc *PiecewiseConstant) Fit(xs, ys []float64) error {
 	n := len(xs)
 	if len(ys) != n {
@@ -173,14 +170,14 @@ type PiecewiseCubic struct {
 	// for xs[i] <= x < xs[i + 1] is defined as
 	//   sum_{k = 0}^3 coeffs.At(i, k) * (x - xs[i])^k
 	// To guarantee left-continuity, coeffs.At(i, 0) == ys[i].
-	coeffs *mat.Dense
+	coeffs mat.Dense
 
 	// Last interpolated Y value, corresponding to xs[len(xs) - 1].
 	lastY float64
 }
 
 // Predict returns the interpolation value at x.
-func (pc PiecewiseCubic) Predict(x float64) float64 {
+func (pc *PiecewiseCubic) Predict(x float64) float64 {
 	i := findSegment(pc.xs, x)
 	if i < 0 {
 		return pc.coeffs.At(0, 0)
@@ -200,11 +197,11 @@ func (pc PiecewiseCubic) Predict(x float64) float64 {
 	return ((a[3]*dx+a[2])*dx+a[1])*dx + a[0]
 }
 
-// fitWithDerivatives fits a piecewise cubic predictor to (X, Y, dY/dX) value
+// FitWithDerivatives fits a piecewise cubic predictor to (X, Y, dY/dX) value
 // triples provided as three slices.
 // It panics if len(xs) < 2, elements of xs are not strictly increasing,
 // len(xs) != len(ys) or len(xs) != len(dydxs).
-func (pc *PiecewiseCubic) fitWithDerivatives(xs, ys, dydxs []float64) {
+func (pc *PiecewiseCubic) FitWithDerivatives(xs, ys, dydxs []float64) {
 	n := len(xs)
 	if len(ys) != n {
 		panic(differentLengths)
@@ -216,7 +213,8 @@ func (pc *PiecewiseCubic) fitWithDerivatives(xs, ys, dydxs []float64) {
 		panic(tooFewPoints)
 	}
 	m := n - 1
-	pc.coeffs = mat.NewDense(m, 4, nil)
+	pc.coeffs.Reset()
+	pc.coeffs.ReuseAs(m, 4)
 	for i := 0; i < m; i++ {
 		dx := xs[i+1] - xs[i]
 		if dx <= 0 {
@@ -251,10 +249,7 @@ func (as *AkimaSpline) Predict(x float64) float64 {
 
 // Fit fits a predictor to (X, Y) value pairs provided as two slices.
 // It panics if len(xs) < 2, elements of xs are not strictly increasing
-// or len(xs) != len(ys). It returns an error if there is an unpredictable
-// failure to fit.
-// If len(xs) == 2, we set both derivatives dY/dX to the slope
-// (ys[1] - ys[0]) / (xs[1] - xs[0]).
+// or len(xs) != len(ys). Always returns nil.
 func (as *AkimaSpline) Fit(xs, ys []float64) error {
 	n := len(xs)
 	if len(ys) != n {
@@ -266,23 +261,23 @@ func (as *AkimaSpline) Fit(xs, ys []float64) error {
 		slope := (ys[1] - ys[0]) / (xs[1] - xs[0])
 		dydxs[0] = slope
 		dydxs[1] = slope
-		as.cubic.fitWithDerivatives(xs, ys, dydxs)
+		as.cubic.FitWithDerivatives(xs, ys, dydxs)
 		return nil
 	}
 
-	m := n - 1
-	slopes := make([]float64, m+4)
-	for i := 0; i < m; i++ {
-		dx := xs[i+1] - xs[i]
+	m := n + 3
+	slopes := make([]float64, m)
+	slopes[0] = 3*slopes[2] - 2*slopes[3]
+	slopes[1] = 2*slopes[2] - slopes[3]
+	for i := 2; i < m-2; i++ {
+		dx := xs[i-1] - xs[i-2]
 		if dx <= 0 {
 			panic(xsNotStrictlyIncreasing)
 		}
-		slopes[i+2] = (ys[i+1] - ys[i]) / dx
+		slopes[i] = (ys[i-1] - ys[i-2]) / dx
 	}
-	slopes[1] = 2*slopes[2] - slopes[3]
-	slopes[0] = 3*slopes[2] - 2*slopes[3]
-	slopes[m+2] = 2*slopes[n] - slopes[m]
-	slopes[m+3] = 3*slopes[n] - 2*slopes[m]
+	slopes[m-2] = 2*slopes[m-3] - slopes[m-4]
+	slopes[m-1] = 3*slopes[m-3] - 2*slopes[m-4]
 	for i := 0; i < n; i++ {
 		wLeft := math.Abs(slopes[i+2] - slopes[i+3])
 		wRight := math.Abs(slopes[i+1] - slopes[i])
@@ -293,7 +288,7 @@ func (as *AkimaSpline) Fit(xs, ys []float64) error {
 			dydxs[i] = (slopes[i+1] + slopes[i+2]) / 2
 		}
 	}
-	as.cubic.fitWithDerivatives(xs, ys, dydxs)
+	as.cubic.FitWithDerivatives(xs, ys, dydxs)
 	return nil
 }
 
--- a/interp/interp.go
+++ b/interp/interp.go
@@ -7,6 +7,8 @@ package interp
 import (
 	"errors"
 	"sort"
+
+	"gonum.org/v1/gonum/mat"
 )
 
 const (
@@ -95,17 +97,17 @@ func (pl PiecewiseLinear) Predict(x float64) float64 {
 	if i < 0 {
 		return pl.ys[0]
 	}
-	// i < len(pci.xs)
+	// i < len(pl.xs)
 	xI := pl.xs[i]
 	if x == xI {
 		return pl.ys[i]
 	}
 	n := len(pl.xs)
 	if i == n-1 {
-		// x > li.xs[i]
+		// x > pl.xs[i]
 		return pl.ys[n-1]
 	}
-	// i < len(i1d.xs) - 1
+	// i < len(pl.xs) - 1
 	return pl.ys[i] + pl.slopes[i]*(x-xI)
 }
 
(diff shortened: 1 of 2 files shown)


interp/interp.go
```diff
@@ -146,18 +153,148 @@ func (pc PiecewiseConstant) Predict(x float64) float64 {
...
+}
+
+// Predict returns the interpolation value at x.
+func (pc PiecewiseCubic) Predict(x float64) float64 {
+	i := findSegment(pc.xs, x)
+	if i < 0 {
+		return pc.coeffs.At(0, 0)
+	}
+	m := len(pc.xs) - 1
+	// i < len(pc.xs)
```
kortschak, line 192: Necessary?

interp/interp.go
```diff
@@ -146,18 +153,148 @@ func (pc PiecewiseConstant) Predict(x float64) float64 {
...
+	if x == pc.xs[i] {
+		if i < m {
+			return pc.coeffs.At(i, 0)
+		}
+		return pc.lastY
+	}
+	if i == m {
+		// x > pc.xs[i]
```
kortschak, line 200: Necessary?

interp/interp.go
```diff
@@ -146,18 +153,148 @@ func (pc PiecewiseConstant) Predict(x float64) float64 {
...
+type AkimaSplines struct {
+	PiecewiseCubic
+}
+
+// Fit fits a predictor to (X, Y) value pairs provided as two slices.
+// It returns an error if len(xs) < 2, elements of xs are not strictly
+// increasing or len(xs) != len(ys).
+// If len(xs) == 2, we set both derivatives dY/dX to the slope
+// (ys[1] - ys[0]) / (xs[1] - xs[0]).
+func (as *AkimaSplines) Fit(xs, ys []float64) error {
```
kortschak, line 258: Here also. The length concordance can be known prior to the call.

interp/interp.go
```diff
@@ -146,18 +153,148 @@ func (pc PiecewiseConstant) Predict(x float64) float64 {
...
+	}
+	dx := x - pc.xs[i]
+	a := pc.coeffs.RawRowView(i)
+	return ((a[3]*dx+a[2])*dx+a[1])*dx + a[0]
+}
+
+// FitWithDerivatives fits a piecewise cubic predictor to (X, Y, dY/dX) value
+// triples provided as three slices.
+// It returns an error if len(xs) < 2, elements of xs are not strictly
+// increasing, len(xs) != len(ys) or len(xs) != len(dydxs).
```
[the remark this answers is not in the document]
romanwerpachowski: Returning errors was requested by @btracey when reviewing the previous PR for interp/.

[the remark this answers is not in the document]
kortschak: I'll have a look at the conversation.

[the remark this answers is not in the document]
btracey: The motivation was that if you wanted something like Least Squares, doing the matrix inversion can cause an error. It seemed like supporting Least Squares is a necessary condition for fitting. Maybe this isn't needed in 1-d though.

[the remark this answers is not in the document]
kortschak: Thanks. Yes, if there is an errorable case that can occur during the calculation but can't be determined prior to that, those would return an error. These cases can be known prior, so they should panic. Whether there are errorable inputs it something that you two are more likely to know.

interp/interp.go
```diff
@@ -84,8 +87,10 @@ func (pl *PiecewiseLinear) Fit(xs, ys []float64) error {
 		}
 		pl.slopes[i] = (ys[i+1] - ys[i]) / dx
 	}
-	pl.xs = xs
-	pl.ys = ys
+	pl.xs = make([]float64, n)
+	pl.ys = make([]float64, n)
+	copy(pl.xs, xs)
+	copy(pl.ys, ys)
```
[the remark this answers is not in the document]
romanwerpachowski: It can lead to hard to detect bugs if xs or ys is modified after falling Fit.

interp/interp.go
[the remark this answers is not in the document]
romanwerpachowski: In this case I will revert to panicking on preventable errors, but leave the option to return an error from Fit. Does it work for everyone?

interp/interp.go
romanwerpachowski: Nothing is :)

interp/interp.go
romanwerpachowski: See above.

interp/interp.go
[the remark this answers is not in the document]
kortschak: That works for me. It would be nice if there were some examples of places where it's likely to be needed so we can make a clearer decision. At the moment, we're not in a release, so it's easy to change. After Augustish this will have been released and so it will be harder (though not impossible by any means) to remove.

interp/interp.go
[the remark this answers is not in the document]
kortschak: Can you elaborate? I don't think we've ever considered this a reason to allocate, but maybe there is a first time.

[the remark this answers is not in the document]
romanwerpachowski: I mean sth like this:

```
xs = make([]float64, 10)
ys = make([]float64, 10)
// set values
fitter1.Fit(xs, ys)
ys[0] = -10
fitter2.Fit(xs, ys)
// fitter1 is now broken because we made a shallow copy
```

interp/interp.go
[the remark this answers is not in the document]
romanwerpachowski: OK.

interp/interp.go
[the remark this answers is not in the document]
kortschak: Yeah, understood. Either we document that a copy is made, or we document that the receiver takes ownership of the slices that it is given and they should not be altered after they have been passed to it. Both of these approaches are used.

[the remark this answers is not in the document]
kortschak: I'm inclined towards the no-copy, but it's worth waiting for other opinions.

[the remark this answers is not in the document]
romanwerpachowski: If the dominant style in `gonum` is to avoid defensive copying, then I'll revert this.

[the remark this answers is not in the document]
kortschak: Yeah, it varies. I don't use this kind of thing often enough to have a strongly defensible opinion, so I would wait for one of the others.

interp/interp.go
```diff
@@ -146,18 +150,142 @@ func (pc PiecewiseConstant) Predict(x float64) float64 {
...
+
+// Fit fits a predictor to (X, Y) value pairs provided as two slices.
+// It returns an error if len(xs) < 2, elements of xs are not strictly
+// increasing or len(xs) != len(ys).
+// If len(xs) == 2, we set both derivatives dY/dX to the slope
+// (ys[1] - ys[0]) / (xs[1] - xs[0]).
+func (as *AkimaSplines) Fit(xs, ys []float64) error {
+	n := len(xs)
+	if len(ys) != n {
+		return errors.New(differentLengths)
```
vladimir-ch, line 252: I think the conclusion was to panic in this case.

interp/interp.go
```diff
@@ -146,18 +150,142 @@ func (pc PiecewiseConstant) Predict(x float64) float64 {
...
+}
+
+// FitWithDerivatives fits a piecewise cubic predictor to (X, Y, dY/dX) value
+// triples provided as three slices.
+// It returns an error if len(xs) < 2, elements of xs are not strictly
+// increasing, len(xs) != len(ys) or len(xs) != len(dydxs).
+func (pc *PiecewiseCubic) FitWithDerivatives(xs, ys, dydxs []float64) error {
+	n := len(xs)
+	if len(ys) != n {
+		return errors.New(differentLengths)
```
vladimir-ch, line 207: I think the conclusion was to panic in these cases.

interp/interp.go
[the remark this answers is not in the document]
vladimir-ch: I'm also slightly inclined towards the no-copy. If the user modifies the slices, they're probably not interested in the interpolation or should not use it as a backup storage. It could even hide their bugs because with a copy the interpolator would keep returning good looking results, with the no-copy something noticeably bad would happen.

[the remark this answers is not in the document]
btracey: I think we should copy. IMO it's generally good practice to do defensive copying when passing into types that have state (or retrieving from them). It's so easy to cause bugs you don't expect. When you call a constructor, it "feels like" the type now just exists and is immutable unless you call one of the methods. IMO things should act like that unless there's a really good reason not to.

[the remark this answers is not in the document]
vladimir-ch: Ok, fine for me.

interp/interp.go
```diff
@@ -146,18 +150,142 @@ func (pc PiecewiseConstant) Predict(x float64) float64 {
...
+	copy(pc.xs, xs)
+	pc.lastY = ys[m]
+	return nil
+}
+
+// AkimaSplines is a left-continuous, piecewise cubic 1-dimensional interpolator
+// which can be fitted to (X, Y) value pairs without providing derivatives.
+// See https://www.iue.tuwien.ac.at/phd/rottinger/node60.html for more details.
+type AkimaSplines struct {
+	PiecewiseCubic
```
vladimir-ch, line 241: The `AkimaSpline` type should not expose the whole `PiecewiseCubic` interface. The essence of this spline is how it chooses the derivatives so you don't want them to be freely customizable via `FitWithDerivatives`.

interp/interp.go
vladimir-ch: Oh, this has already be resolved above, I missed the thread.

interp/interp.go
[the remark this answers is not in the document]
romanwerpachowski: I'll leave it as it is then.

interp/interp.go
```diff
@@ -146,18 +150,142 @@ func (pc PiecewiseConstant) Predict(x float64) float64 {
...
 	}
 	n := len(pc.xs)
 	if i == n-1 {
-		// x > pci.xs[i]
 		return pc.ys[n-1]
 	}
 	return pc.ys[i+1]
 }
 
+// PiecewiseCubic is a left-continuous, piecewise cubic 1-dimensional interpolator.
```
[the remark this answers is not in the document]
romanwerpachowski: True.

interp/interp.go
```diff
@@ -146,18 +150,142 @@ func (pc PiecewiseConstant) Predict(x float64) float64 {
...
+		pc.coeffs.Set(i, 2, (3*dy-(2*dydxs[i]+dydxs[i+1])*dx)/dx/dx)
+		pc.coeffs.Set(i, 3, (-2*dy+(dydxs[i]+dydxs[i+1])*dx)/dx/dx/dx)
+	}
+	pc.xs = make([]float64, n)
+	copy(pc.xs, xs)
+	pc.lastY = ys[m]
+	return nil
+}
+
+// AkimaSplines is a left-continuous, piecewise cubic 1-dimensional interpolator
```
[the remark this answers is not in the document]
romanwerpachowski: Good point.

interp/interp.go
romanwerpachowski: OK. I won't export `fitWithDerivatives` for now.

interp/interp.go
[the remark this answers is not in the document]
romanwerpachowski: This distinction between panicking on predictable error conditions and returning errors on unpredictable ones reminds me the distinction between unchecked and checked exceptions in Java, BTW.

interp/interp_test.go
```diff
@@ -191,3 +194,228 @@ func TestPiecewiseConstantPredict(t *testing.T) {
...
+
+func rightPoly(x float64) float64 {
+	return x*x*x - x + 1
+}
+
+func rightPolyDerivative(x float64) float64 {
+	return 3*x*x - 1
+}
+
+func testFittedPolys(t *testing.T, pc *PiecewiseCubic) {
```
vladimir-ch, line 292: The testing for Cubic should be structured differently. For example, this function works only if `pc` was fitted to particular points using particular functions (`rightPoly` and `leftPoly`). This hard-wired connection is surprising, also because the name `testFittedPolys` looks quite general. Also, the package-level functions `rightPoly`, `leftPoly` are not very informative.

I would have a table of test cases, something like you have for Akima:
```
type testCaseCubic struct {
    x []float64
    y []float64
    dy []float64

    want func(float64) float64
}

for _, test := range []testCaseCubic {
    {
        x: []float64{0, 0.9},
        y: []float64{0, 0},
        dy: []float64{0, 0},
        want: func(float64) float64 {return 0},
    },
    {
        x: []float64{0, 0.9, 2.3},
        y: []float64{1, 1, 1},
        dy: []float64{0, 0, 0},
        want: func(float64) float64 {return 1},
    },
    ...
}
```
You can have helper functions to create test cases:
```
func newTestCaseCubic(n int, f, df func(float64) float64, rnd *rand.Rand) testCaseCubic {
    x := make([]float64, n)
    // fill x, for example by jittering uniformly spaced points
    y := make([]float64, n)
    dy := make([]float64, n)
    for i, xi := range x {
        y[i] = f(xi)
        dy[i] = df(xi)
    }
    return testCaseCubic {
        x: x,
        y: y,
        dy: dy,
        want: f,
    }
}
```
In this way you can have a variety of interesting test cases, from some special "corner cases" like when the cubic is actually a constant or linear function to a general piecewise cubic. The input data and the expected output are grouped together, so the test code is easier to document and follow.

interp/interp_test.go
```diff
@@ -191,3 +194,228 @@ func TestPiecewiseConstantPredict(t *testing.T) {
...
+		{2, lastY},
+	} {
+		got := pc.Predict(test.x)
+		if math.Abs(got-test.want) > 1e-14 {
+			t.Errorf("Mismatch in test case %d for x = %g: got %v, want %g", i, test.x, got, test.want)
+		}
+	}
+}
+
+func TestAkimaSplinesSingleFunction(t *testing.T) {
```
vladimir-ch, line 321: Why the `SingleFunction` suffix? If it refers to how the test is done, then it should go away.

interp/interp_test.go
```diff
@@ -191,3 +194,228 @@ func TestPiecewiseConstantPredict(t *testing.T) {
...
+			got := as.Predict(x)
+			want := test.f(math.Min(x1, math.Max(x0, x)))
+			if math.Abs(got-want) > tol {
+				t.Errorf("Mismatch in interpolated value at x == %g for test case %d: got %v, want %g", x, i, got, want)
+			}
+		}
+	}
+}
+
+func TestAkimaSplinesNoWiggles(t *testing.T) {
```
vladimir-ch, line 366: I think this can be part of `TestAkimaSpline`.

romanwerpachowski:
> Is there any plan what will/should go into this package?

Nothing concrete ATM, but I want to include some algorithms used in financial mathematics.

interp/interp_test.go
romanwerpachowski: OK.

interp/interp_test.go
romanwerpachowski: OK.

interp/interp_test.go
romanwerpachowski: I cleaned it up.

romanwerpachowski:
PTAL.

interp/interp_test.go
```diff
@@ -191,3 +190,341 @@ func TestPiecewiseConstantPredict(t *testing.T) {
...
+			if j < n-1 {
+				got = pc.coeffs.At(j, 1)
+				if math.Abs(got-want) > valueTol {
+					t.Errorf("Mismatch in 1-st order interpolation coefficient in %d-th node for test case %d: got %v, want %g", j, i, got, want)
+				}
+			}
+			if j > 0 {
+				dx := test.xs[j] - test.xs[j-1]
+				got = (3*pc.coeffs.At(j-1, 3)*dx+2*pc.coeffs.At(j-1, 2))*dx + pc.coeffs.At(j-1, 1)
+				if math.Abs(got-want) > valueTol {
```
kortschak, line 293: You will need to increase the tolerance here; arm64 gives a slightly different result.

```
--- FAIL: TestPiecewiseCubic (0.00s)
    interp_test.go:294: Interpolation coefficients in 3-th node produce mismatch in interpolated derivative value at 2.99 for test case 5: got 19.88568249156475, want 19.88568249156473
```

Evicting the FMAs doesn't fix this.

interp/interp_test.go
```diff
@@ -68,21 +70,18 @@ func BenchmarkFindSegment(b *testing.B) {
...
+		{[]float64{0.3}, []float64{0}},
+		{[]float64{0.3, 0.3}, []float64{0, 0}},
+		{[]float64{0.3, -0.3}, []float64{0, 0}},
 	}
 	for _, params := range errorParamSets {
-		err := fp.Fit(params.xs, params.ys)
-		expectedMessage := fmt.Sprintf("interp: %s", params.expectedMessage)
-		if err == nil || err.Error() != expectedMessage {
-			t.Errorf("expected error for xs: %v and ys: %v with message: %s", params.xs, params.ys, expectedMessage)
+		if !panics(func() { fp.Fit(params.xs, params.ys) }) {
```
kortschak, line 83: `if !panics(func() { _ = fp.Fit(params.xs, params.ys) }) {`

interp/interp_test.go
```diff
@@ -191,3 +190,341 @@ func TestPiecewiseConstantPredict(t *testing.T) {
...
+			xs: []float64{0},
+			ys: []float64{0},
+		},
+		{
+			xs: []float64{0, 1, 1},
+			ys: []float64{10, 20, 10},
+		},
+	} {
+		var as AkimaSpline
+		if !panics(func() { as.Fit(test.xs, test.ys) }) {
```
kortschak, line 507: `if !panics(func() { _ = as.Fit(test.xs, test.ys) }) {`

interp/interp_test.go
romanwerpachowski: Oops. Thanks for spotting.

interp/interp_test.go
romanwerpachowski: It seems to work without the underscore, so why make the code longer?

kortschak: To keep the linter quiet.

kortschak: It leaves annotations in the files changed tab.

romanwerpachowski:
PTAL

interp/interp.go
```diff
@@ -146,18 +152,146 @@ func (pc PiecewiseConstant) Predict(x float64) float64 {
...
+	dx := x - pc.xs[i]
+	a := pc.coeffs.RawRowView(i)
+	return ((a[3]*dx+a[2])*dx+a[1])*dx + a[0]
+}
+
+// fitWithDerivatives fits a piecewise cubic predictor to (X, Y, dY/dX) value
+// triples provided as three slices.
+// It panicks if len(xs) < 2, elements of xs are not strictly increasing,
+// len(xs) != len(ys) or len(xs) != len(dydxs).
+func (pc *PiecewiseCubic) fitWithDerivatives(xs, ys, dydxs []float64) {
```
vladimir-ch, line 207: Unexporting this function makes this type unusable because there's no way to fit it. I would export this function and embed `PiecewiseCubic` as an unexported field in `Akima`.

interp/interp.go
```diff
@@ -146,18 +152,146 @@ func (pc PiecewiseConstant) Predict(x float64) float64 {
...
+// AkimaSpline is a piecewise cubic 1-dimensional interpolator with
+// continuous value and first derivative, which can be fitted to (X, Y)
+// value pairs without providing derivatives.
+// See https://www.iue.tuwien.ac.at/phd/rottinger/node60.html for more details.
+type AkimaSpline struct {
+	PiecewiseCubic
+}
+
+// Fit fits a predictor to (X, Y) value pairs provided as two slices.
+// It panicks if len(xs) < 2, elements of xs are not strictly increasing
```
vladimir-ch, line 248: I think this is spelled as "panics". Also at several other places in this PR.

interp/interp_test.go
```diff
@@ -191,3 +190,341 @@ func TestPiecewiseConstantPredict(t *testing.T) {
...
+			x := test.xs[j]
+			got := pc.Predict(x)
+			want := test.f(x)
+			if math.Abs(got-want) > valueTol {
+				t.Errorf("Mismatch in interpolated value at x == %g for test case %d: got %v, want %g", x, i, got, want)
+			}
+			if j < n-1 {
+				got = pc.coeffs.At(j, 0)
+				if math.Abs(got-want) > valueTol {
+					t.Errorf("Mismatch in 0-th order interpolation coefficient in %d-th node for test case %d: got %v, want %g", j, i, got, want)
```
vladimir-ch, line 257: I don't think it's necessary to check values of the internal coefficients. Removing it would simplify the test a lot.

interp/interp_test.go
```diff
@@ -191,3 +190,341 @@ func TestPiecewiseConstantPredict(t *testing.T) {
...
+		},
+		{
+			xs: []float64{-1.001, 0.2, 10},
+			f:  func(x float64) float64 { return -1 },
+			df: func(x float64) float64 { return 0 },
+		},
+		{
+			xs: []float64{-1.1, 0.2, 0.99, 2.5, 2.99},
+			f:  math.Cos,
+			df: math.Sin,
```
vladimir-ch, line 229: `cos'(x) = -sin(x)` and why does the test passes?

interp/interp_test.go
```diff
@@ -191,3 +190,341 @@ func TestPiecewiseConstantPredict(t *testing.T) {
...
+		},
+		{
+			xs: []float64{-1.1, 0.2, 0.99, 2.5, 2.99},
+			f:  math.Exp,
+			df: math.Exp,
+		},
+		{
+			xs: []float64{-1.1, 0.2, 0.99, 2.5, 2.99},
+			f:  func(x float64) float64 { return math.Sin(x * x) },
+			df: func(x float64) float64 { return -2 * x * math.Cos(x*x) },
```
vladimir-ch, line 239: `sin'(x) = cos(x)`

interp/interp_test.go
```diff
@@ -191,3 +190,341 @@ func TestPiecewiseConstantPredict(t *testing.T) {
...
+			df: func(x float64) float64 { return 1.5 },
+		},
+		{
+			xs: []float64{-1.001, 0.2, 10},
+			f:  func(x float64) float64 { return -1 },
+			df: func(x float64) float64 { return 0 },
+		},
+		{
+			xs: []float64{-1.1, 0.2, 0.99, 2.5, 2.99},
+			f:  math.Cos,
```
vladimir-ch, line 228: If you use either only at most 3. degree polynomials or if you add a flag to indicate that a test case is such, you could check the fit (which should be very good) also at points that are not knots.

interp/interp_test.go
```diff
@@ -191,3 +190,341 @@ func TestPiecewiseConstantPredict(t *testing.T) {
...
+				got = (3*pc.coeffs.At(j-1, 3)*dx+2*pc.coeffs.At(j-1, 2))*dx + pc.coeffs.At(j-1, 1)
+				if math.Abs(got-want) > valueTol {
+					t.Errorf("Interpolation coefficients in %d-th node produce mismatch in interpolated derivative value at %g for test case %d: got %v, want %g", j-1, x, i, got, want)
+				}
+			}
+		}
+	}
+}
+
+func TestPiecewiseCubicExactFit(t *testing.T) {
```
vladimir-ch, line 301: What aspect tests this function that's not already tested by `TestPiecewiseCubic`?

interp/interp_test.go
romanwerpachowski: Hmm. This is very strange. I'll check.

interp/interp_test.go
romanwerpachowski: Oops. I hope they won't take back my diploma for this :D

interp/interp_test.go
romanwerpachowski: Good point.

interp/interp_test.go
romanwerpachowski: If I implement your suggestion from previous comment, nothing.

interp/interp.go
romanwerpachowski: Yeah, I used the Mason & Dixon spelling. I'll fix that.

interp/interp.go
romanwerpachowski: Right, that's the pattern I wanted all the time.

interp/interp_test.go
romanwerpachowski: I disagree. I think it gives additional security that the implementation is correct.

interp/interp_test.go
romanwerpachowski: I'll leave the part of this function which tests internals (coefficients).

interp/interp_test.go
romanwerpachowski: The test passes because the interpolation algorithm doesn't care what function I got the (x, y, dy/dx) points from - it just interpolates those points.

interp/interp_test.go
romanwerpachowski: (It was still a valid test, but not educational.)

romanwerpachowski:
PTAL 2.

interp/interp.go
```diff
@@ -146,18 +152,151 @@ func (pc PiecewiseConstant) Predict(x float64) float64 {
...
+		panic(differentLengths)
+	}
+	if len(dydxs) != n {
+		panic(differentLengths)
+	}
+	if n < 2 {
+		panic(tooFewPoints)
+	}
+	m := n - 1
+	pc.coeffs = mat.NewDense(m, 4, nil)
```
vladimir-ch, line 219: I think that `coeffs` should be a value and this should be `pc.coeffs.ReuseAs(m, 4)`.

interp/interp.go
```diff
@@ -146,18 +152,151 @@ func (pc PiecewiseConstant) Predict(x float64) float64 {
...
+// or len(xs) != len(ys). It returns an error if there is an unpredictable
+// failure to fit.
+// If len(xs) == 2, we set both derivatives dY/dX to the slope
+// (ys[1] - ys[0]) / (xs[1] - xs[0]).
+func (as *AkimaSpline) Fit(xs, ys []float64) error {
+	n := len(xs)
+	if len(ys) != n {
+		panic(differentLengths)
+	}
+	dydxs := make([]float64, n)
```
vladimir-ch, line 263: @kortschak , is it worth making this and `slopes` fields of AkimaSpline and reuse them?

interp/interp.go
```diff
@@ -146,18 +152,151 @@ func (pc PiecewiseConstant) Predict(x float64) float64 {
...
+		dy := ys[i+1] - ys[i]
+		// a_0
+		pc.coeffs.Set(i, 0, ys[i])
+		// a_1
+		pc.coeffs.Set(i, 1, dydxs[i])
+		// Solve a linear equation system for a_2 and a_3.
+		pc.coeffs.Set(i, 2, (3*dy-(2*dydxs[i]+dydxs[i+1])*dx)/dx/dx)
+		pc.coeffs.Set(i, 3, (-2*dy+(dydxs[i]+dydxs[i+1])*dx)/dx/dx/dx)
+	}
+	pc.xs = make([]float64, n)
```
vladimir-ch, line 234: @kortschak , also reuse here?

interp/interp.go
kortschak: If this is likely to be used in tight loops then yes that would be best. There's not really a down side.

interp/interp.go
kortschak: As above.

interp/interp_test.go
romanwerpachowski: Hmm, OK.

interp/interp_test.go
```diff
@@ -191,3 +190,336 @@ func TestPiecewiseConstantPredict(t *testing.T) {
...
+			want := test.f(x)
+			if math.Abs(got-want) > valueTol {
+				t.Errorf("Mismatch in interpolated value at x == %g for test case %d: got %v, want %g", x, i, got, want)
+			}
+			if j < n-1 {
+				got = pc.coeffs.At(j, 0)
+				if math.Abs(got-want) > valueTol {
+					t.Errorf("Mismatch in 0-th order interpolation coefficient in %d-th node for test case %d: got %v, want %g", j, i, got, want)
+				}
+				if test.exact {
```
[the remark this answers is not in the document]
romanwerpachowski: OK, point taken.

interp/interp.go
```diff
@@ -146,18 +152,151 @@ func (pc PiecewiseConstant) Predict(x float64) float64 {
...
+}
+
+// Predict returns the interpolation value at x.
+func (as *AkimaSpline) Predict(x float64) float64 {
+	return as.cubic.Predict(x)
+}
+
+// Fit fits a predictor to (X, Y) value pairs provided as two slices.
+// It panics if len(xs) < 2, elements of xs are not strictly increasing
+// or len(xs) != len(ys). It returns an error if there is an unpredictable
```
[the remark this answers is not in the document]
romanwerpachowski: I copied the comment from the interface. Is it standard practice to change the comments to fit implementation details?

interp/interp_test.go
```diff
@@ -191,3 +190,336 @@ func TestPiecewiseConstantPredict(t *testing.T) {
...
+		for j := 0; j < m; j++ {
+			x0 := test.xs[j]
+			x1 := test.xs[j+1]
+			yMin := math.Min(ys[j], ys[j+1])
+			yMax := math.Max(ys[j], ys[j+1])
+			dx := (x1 - x0) / nPts
+			for k := 0; k <= nPts; k++ {
+				x := x0 + float64(k)*dx
+				y := as.Predict(x)
+				if y < yMin-wiggleTol || y > yMax+wiggleTol {
```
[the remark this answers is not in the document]
romanwerpachowski: [Kind of](https://en.wikipedia.org/wiki/Akima_spline): 

> Akima spline avoids unphysical wiggles in regions where the second derivative in the underlying curve is rapidly changing.

I tried to test it here. I think it's a useful test.

interp/interp.go
romanwerpachowski: Thanks for the suggestion! I am not used to Gonum Matrix API.

interp/interp.go
```diff
@@ -146,18 +152,151 @@ func (pc PiecewiseConstant) Predict(x float64) float64 {
...
+		slope := (ys[1] - ys[0]) / (xs[1] - xs[0])
+		dydxs[0] = slope
+		dydxs[1] = slope
+		as.cubic.fitWithDerivatives(xs, ys, dydxs)
+		return nil
+	}
+
+	m := n - 1
+	slopes := make([]float64, m+4)
+	for i := 0; i < m; i++ {
```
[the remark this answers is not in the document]
romanwerpachowski: Is calculating `m - 3` inside the loop going to be a problem?

interp/interp.go
vladimir-ch: I don't think fitting would be done in tight loops, so this can be postponed, this review is starting to be already quite long.

interp/interp.go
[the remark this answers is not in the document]
vladimir-ch: I'd say that the documentation for a concrete type can be tailored to the type and doesn't have to be a copy of the interface that the type satisfies. Maybe @kortschak has a suggestion here.

Either way, the word "unpredictable" seems out of place. At least I would say something like "Fit returns an error if the fitting fails.".

interp/interp_test.go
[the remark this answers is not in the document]
vladimir-ch: I'm not saying it's not a useful test, I just don't think that "avoids unphysical wiggles" means "monotonicity of input implies monotonicity of the interpolant".

Can you try the following input to see if this test passes with it?
```
xs: []float64{7.99, 8.09, 8.19, 8.7, 9.2, 10, 12, 15, 20},
ys: []float64{0, 2.76429e-5, 4.37498e-2, 0.169183, 0.469428, 0.943740, 0.998636, 0.999919, 0.999994},
```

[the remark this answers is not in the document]
romanwerpachowski: I will. The test doesn't check whether the interpolant is monotonic, just that it doesn't shoot up or down too much.

interp/interp.go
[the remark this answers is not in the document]
romanwerpachowski: OK.

interp/interp.go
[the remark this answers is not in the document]
vladimir-ch: I wouldn't be surprised if the Go compiler optimized this and evaluated it only once. Even if it didn't, I don't think that one additional integer operation per iteration would be any issue, especially for creating the interpolator.

interp/interp_test.go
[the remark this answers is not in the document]
vladimir-ch: "too much" is relative and the current tolerance of 0.1 very loose and arbitrary. I haven't seen anything about Akima splines that would guarantee this property theoretically and so justify this test. If you had a spline that does provide theoretical guarantees, then this test with a tight tolerance should definitely stay.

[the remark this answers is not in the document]
romanwerpachowski: It's no less arbitrary than many of the tolerances in the codebase...

interp/interp.go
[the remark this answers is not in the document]
kortschak: This implementation always returns nil, so the documentation should say that.

interp/interp.go
kortschak: Can you file an issue to look into this?

vladimir-ch: Done #1409

interp/interp_test.go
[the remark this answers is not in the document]
vladimir-ch: Akima splines don't have the property that you're testing for, therefore this part of the test should be removed. To give another example input where this is demonstrated even better:
```
xs: []float64{1, 2, 4, 6.5, 8, 10, 10.5, 11, 13, 14},
ys: []float64{0, 0, 0, 0, 0.1, 1, 4.5, 8, 10, 15},
```
In `[8,10]` the test expects the interpolant to have values in `[0,1.1]. In reality, the spline goes down to roughly -1.3.

Not that it is relevant to this review but I'm not aware of any place in Gonum where we don't try to have the test tolerances as tight as possible. If we had to tweak them, it's either because the numerical solution is difficult for floating point arithmetic (e.g. in optimize, mat) or that we want the tests to finish in a reasonable time (in stat).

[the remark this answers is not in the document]
romanwerpachowski: Fair enough. I'll delete it.

interp/interp_test.go
```diff
@@ -191,3 +190,336 @@ func TestPiecewiseConstantPredict(t *testing.T) {
...
+		wiggleTol = 1e-1
+	)
+	for i, test := range []struct {
+		xs  []float64
+		f   func(float64) float64
+		tol float64
+	}{
+		{
+			xs:  []float64{-1, 0, 1},
+			f:   func(x float64) float64 { return x * x },
```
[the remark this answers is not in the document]
vladimir-ch: This is actually more complicated than that. Akima splines are piecewise cubic but they are accurate only for 2. degree polynomials and only if the points are uniformly spaced. So the testing strategy here is again unclear. There's one test case with a quadratic but only 3 points which means the whole curve is determined by the boundary conditions, there is no true interior point. Then two test cases with a linear polynomial, with 2 and 3 points. And then some test cases with 5 points but non-polynomials where no good match can be expected which shows by suddenly using loose tolerances, up to 0.5.

I think the simplest and most reliable test here would be to take a trustworthy implementation and use it to generate  `want` at each `samples` in:
```
	for i, test := range []struct {
		x  []float64
		y []float64

		samples []float64
		want []float64
	}
```
I've already pasted here two candidate input data, others can be found for example in "A Method of Univariate Interpolation That Has the Accuracy of a Third-Degree Polynomial".

interp/interp.go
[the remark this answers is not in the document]
romanwerpachowski: OK.

interp/interp.go
```diff
@@ -146,18 +152,151 @@ func (pc PiecewiseConstant) Predict(x float64) float64 {
...
+// Predict returns the interpolation value at x.
+func (as *AkimaSpline) Predict(x float64) float64 {
+	return as.cubic.Predict(x)
+}
+
+// Fit fits a predictor to (X, Y) value pairs provided as two slices.
+// It panics if len(xs) < 2, elements of xs are not strictly increasing
+// or len(xs) != len(ys). It returns an error if there is an unpredictable
+// failure to fit.
+// If len(xs) == 2, we set both derivatives dY/dX to the slope
```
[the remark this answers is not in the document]
romanwerpachowski: Can you clarify?

interp/interp.go
vladimir-ch: Also note that the matrix must be empty before `ReuseAs`, therefore you must call `Reset` first.

interp/interp.go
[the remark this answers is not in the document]
vladimir-ch: It's just a nit but what I meant is that this sentence describes an uninteresting, useless cornercase in a disproportionate level of detail. The user can't expect an interpolation to do more than that with two points. With the quadratic comment I thought that maybe with 3 points the curve becomes quadratic but with the artificial boundary values I'm not sure about that and won't put time into checking it.

interp/interp.go
romanwerpachowski: Oops.

interp/interp_test.go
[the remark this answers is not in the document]
romanwerpachowski: I don't like relying on third-party code in tests when not necessary. I understand using Mathematica to test an implementation of a Bessel function, for example, but here it's not necessary IMHO.

I like the idea of using Akima's original test examples, though. Thanks for the pointer.

romanwerpachowski:
I changed the testing strategy for AkimaSpline. In this case it's IMHO better to test individual sub-steps of the algorithm (which can be tricky and prone to off-by-1 errors, etc) than to test the whole thing. As a composition of `akimaSlopes`, `akimaWeights` and `akimaWeightedAverage`, the `Fit` function is pretty straightforward.

WDYT?

romanwerpachowski:
*ping*

interp/interp_test.go
```diff
@@ -191,3 +190,465 @@ func TestPiecewiseConstantPredict(t *testing.T) {
...
+		if !panics(func() { pc.FitWithDerivatives(test.xs, test.ys, test.dydxs) }) {
+			t.Errorf("expected panic for xs: %v, ys: %v and dydxs: %v", test.xs, test.ys, test.dydxs)
+		}
+	}
+}
+
+func TestAkimaSpline(t *testing.T) {
+	t.Parallel()
+	const (
+		nPts = 40
```
vladimir-ch, line 374: Unused.

interp/interp_test.go
```diff
@@ -191,3 +190,465 @@ func TestPiecewiseConstantPredict(t *testing.T) {
...
+		h    = 1e-8
```
vladimir-ch, line 375: Unused.

interp/interp_test.go
```diff
@@ -191,3 +190,465 @@ func TestPiecewiseConstantPredict(t *testing.T) {
...
+		if !panics(func() { _ = as.Fit(test.xs, test.ys) }) {
+			t.Errorf("expected panic for xs: %v and ys: %v", test.xs, test.ys)
+		}
+	}
+}
+
+func TestAkimaWeightedAverage(t *testing.T) {
+	t.Parallel()
+	for i, test := range []struct {
+		v1, v2, w1, w2, want float64
```
vladimir-ch, line 482: Please add a comment saying where the `want` results come from here and in the other Akima tests.

romanwerpachowski:
Thanks! PTAL.

PR #1401, opened 2020-07-14
stat/distuv: add complete test coverage for Uniform distribution

romanwerpachowski:
Please take a look.

### DIFF
--- a/floats/floats.go
+++ b/floats/floats.go
@@ -250,6 +250,14 @@ func EqualFunc(s1, s2 []float64, f func(float64, float64) bool) bool {
 	return true
 }
 
+// EqualOrBothNaN returns true if a and b are exactly equal or both are NaN.
+func EqualOrBothNaN(a, b float64) bool {
+	if math.IsNaN(a) {
+		return math.IsNaN(b)
+	}
+	return a == b
+}
+
 // EqualWithinAbs returns true if a and b have an absolute
 // difference of less than tol.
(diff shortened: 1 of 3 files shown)


floats/floats.go
```diff
@@ -250,6 +250,14 @@ func EqualFunc(s1, s2 []float64, f func(float64, float64) bool) bool {
 	return true
 }
 
+// EqualOrBothNaN returns true if a and b are exactly equal or both are NaN.
+func EqualOrBothNaN(a, b float64) bool {
```
kortschak, line 254: We spell this `Same`, though that label is taken; we need to discuss this because the best fix will break a lot of code. For the moment, use a `same` that's local to the tests.

The body would be `return a == b || (math.IsNaN(a) && math.IsNaN(b)`.

romanwerpachowski: So not in floats package?

kortschak: Yeah, you can leave it just below the function that needs it. We do this in other places as well (suggesting the discussion is needed). Unfortunately, `floats.Same(a, b []float64) bool` already exists (with the same semantics, but for slices).

romanwerpachowski: Heh. How about `floats.SameScalar`?

kortschak: I don't think the discussion should happen here. Please file an issue.

romanwerpachowski:
PTAL.

PR #1402, opened 2020-07-14
stat/distuv: fix CDF and score functions for Weibull distribution

romanwerpachowski:
Also completes test coverage for Weibull.

### DIFF
--- a/stat/distuv/weibull.go
+++ b/stat/distuv/weibull.go
@@ -6,7 +6,6 @@ package distuv
 
 import (
 	"math"
-	"math/cmplx"
 
 	"golang.org/x/exp/rand"
 )
@@ -28,7 +27,7 @@ func (w Weibull) CDF(x float64) float64 {
 	if x < 0 {
 		return 0
 	}
-	return 1 - cmplx.Abs(cmplx.Exp(w.LogCDF(x)))
+	return -math.Expm1(-math.Pow(x/w.Lambda, w.K))
 }
 
 // Entropy returns the entropy of the distribution.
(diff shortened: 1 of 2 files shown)


stat/distuv/weibull.go
```diff
@@ -46,14 +45,6 @@ func (w Weibull) gammaIPow(i, pow float64) float64 {
...
 }
 
-// LogCDF computes the value of the log of the cumulative density function at x.
-func (w Weibull) LogCDF(x float64) complex128 {
-	if x < 0 {
-		return 0
-	}
-	return cmplx.Log(-1) + complex(-math.Pow(x/w.Lambda, w.K), 0)
-}
-
```
kortschak, line 56: Is there a way to retain the `LogCDF` without wandering into complex territory?

romanwerpachowski: There is, but does anyone need it? AFAIK it's the only distribution in `stat/distuv` which has this function.

kortschak: No, it's not.

```
$ go doc -all gonum.org/v1/gonum/stat/distuv | grep func.*LogProb
func (b Bernoulli) LogProb(x float64) float64
func (b Beta) LogProb(x float64) float64
func (b Binomial) LogProb(x float64) float64
func (c Categorical) LogProb(x float64) float64
func (c ChiSquared) LogProb(x float64) float64
func (e Exponential) LogProb(x float64) float64
func (f F) LogProb(x float64) float64
func (g Gamma) LogProb(x float64) float64
func (g GumbelRight) LogProb(x float64) float64
func (g InverseGamma) LogProb(x float64) float64
func (l Laplace) LogProb(x float64) float64
func (l LogNormal) LogProb(x float64) float64
func (n Normal) LogProb(x float64) float64
func (p Pareto) LogProb(x float64) float64
func (p Poisson) LogProb(x float64) float64
func (s StudentsT) LogProb(x float64) float64
func (t Triangle) LogProb(x float64) float64
func (u Uniform) LogProb(x float64) float64
func (w Weibull) LogProb(x float64) float64
```

I think there's only one that doesn't have it, `AlphaStable`.

romanwerpachowski: You searched for `LogProb`, but I deleted `LogCDF`.

kortschak: Eeep. Yes. Apologies. Delete away.

PR #1407, opened 2020-07-16
floats: Rename Same to AllSame. Add Same for scalars.

### LINKED ISSUE #1405: floats: add SameScalars to compare two float64 values with NaN == NaN.
**What**

`floats/floats.go` has a function

```
// Same returns true if the input slices have the same length and all elements
// have the same value with NaN treated as the same.
func Same(s, t []float64) bool {
...}
````

I would like to add there a function `SameScalars` which does the same for scalar inputs:

```
// SameScalars returns true if the inputs have the same value with NaN treated as the same.
func SameScalars(a, b float64) bool {
}
```

**Rationale**

Such a function will be useful for unit testing functions which are expected to return NaNs for some inputs.

**Updated proposal**

Rename `Same` to `AllSame` (breaks code) and call the new function `Same`.

romanwerpachowski:
Fixes #1405. Please take a look.

### DIFF
--- a/blas/testblas/common_test.go
+++ b/blas/testblas/common_test.go
@@ -155,7 +155,7 @@ func TestFlattenBanded(t *testing.T) {
 	} {
 		condensed := flattenBanded(test.dense, test.ku, test.kl)
 		correct := flatten(test.condensed)
-		if !floats.Same(condensed, correct) {
+		if !floats.AllSame(condensed, correct) {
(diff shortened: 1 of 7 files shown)


kortschak:
See https://github.com/gonum/gonum/issues/1405#issuecomment-659701641

romanwerpachowski:
Fixed by https://github.com/gonum/gonum/pull/1418.

PR #1414, opened 2020-07-30
Add format for complex matrix

acheron2302:
This change add CFormatted function for mat package. It is like Formatted function but for complex matrix.
Note: Haven't support for python format and matlab format


kortschak:
It seems to me that the best way to do this, given that we can't shoe horn complex matrices into the same infrastructure that we have for reals, would be to just copy the whole file and adjust it for complex values. This would leave only minor edits for the documentation and would ensure that the behaviour is consistent. It's not entirely clear to me why you stripped the documentation and other comments.

acheron2302:
> It seems to me that the best way to do this, given that we can't shoe horn complex matrices into the same infrastructure that we have for reals, would be to just copy the whole file and adjust it for complex values. This would leave only minor edits for the documentation and would ensure that the behaviour is consistent. It's not entirely clear to me why you stripped the documentation and other comments.

wait i didn't modify anything excpet adding cformat.go and cformat_test.go

kortschak:
Please just start from the complete copy and delete/modify as needed.

kortschak:
Although it is worth spending some time to see if we can get away with less code duplication. That will not be trivial.

vladimir-ch:

PR #1415, opened 2020-07-30
interp: add a DerivativePredictor interface and implement it in AkimaSpline and PiecewiseCubic

romanwerpachowski:
Please take a look.


romanwerpachowski:
*ping*

PR #1417, opened 2020-08-03
stat/distmat: add UniformPermutation

jonreiter:
Simple wrapper to generate random permutation matrices build around math/rand/Shuffle.

### DIFF
--- a/stat/distmat/permutation.go
+++ b/stat/distmat/permutation.go
@@ -1,0 +1,53 @@
+// Copyright ©2020 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package distmat
+
+import (
+	"math/rand"
+
+	"gonum.org/v1/gonum/mat"
+)
+
+// UniformPermutation is a distribution over permutation matrices
+type UniformPermutation struct {
+	src rand.Source
+}
+
+// NewUniformPermutation constructs a new permutation matrix
+// generator using the given random source.
+func NewUniformPermutation(src rand.Source) *UniformPermutation {
+	up := UniformPermutation{src: src}
+	return &up
+}
+
+// Matrix draws a random permutuation matrix of dimension n.
+func (up *UniformPermutation) Matrix(n int) *mat.Dense {
+	m := mat.NewDense(n, n, nil)
+	up.MatrixTo(m)
+	return m
+}
+
+// MatrixTo sets the given matrix to be a random permutation matrix
+// panics if m is not square
+// Note this does not zero the matrix -- before calling this ensure it
+// is full of 0s.
+func (up *UniformPermutation) MatrixTo(m *mat.Dense) {
+	r, c := m.Dims()
+	if r != c {
+		panic(mat.ErrShape)
+	}
+	if r == 0 {
+		return
+	}
+	iList := make([]int, r)
+	for i := range iList {
+		iList[i] = i
+	}
+	rnd := rand.New(up.src)
+	rnd.Shuffle(r, func(i, j int) { iList[i], iList[j] = iList[j], iList[i] })
+	for i, j := range iList {
+		m.Set(i, j, 1.0)
+	}
+}
--- a/stat/distmat/permutation.go
+++ b/stat/distmat/permutation.go
@@ -5,7 +5,7 @@
 package distmat
 
 import (
-	"math/rand"
+	"golang.org/x/exp/rand"
 
 	"gonum.org/v1/gonum/mat"
 )
--- a/stat/distmat/permutation.go
+++ b/stat/distmat/permutation.go
@@ -10,30 +10,25 @@ import (
 	"gonum.org/v1/gonum/mat"
 )
 
-// UniformPermutation is a distribution over permutation matrices
+// UniformPermutation is a uniform distribution over the n!
+// permutation matrices of size n x n for a given n.
 type UniformPermutation struct {
-	src rand.Source
+	rnd     *rand.Rand
+	indices []int
 }
 
 // NewUniformPermutation constructs a new permutation matrix
 // generator using the given random source.
 func NewUniformPermutation(src rand.Source) *UniformPermutation {
-	up := UniformPermutation{src: src}
-	return &up
+	return &UniformPermutation{rnd: rand.New(src)}
 }
 
-// Matrix draws a random permutuation matrix of dimension n.
-func (up *UniformPermutation) Matrix(n int) *mat.Dense {
-	m := mat.NewDense(n, n, nil)
-	up.MatrixTo(m)
-	return m
-}
-
-// MatrixTo sets the given matrix to be a random permutation matrix
-// panics if m is not square
-// Note this does not zero the matrix -- before calling this ensure it
-// is full of 0s.
-func (up *UniformPermutation) MatrixTo(m *mat.Dense) {
+// PermTo sets the given matrix to be a random permutation matrix.
+// It does not zero the destination's elements, so it is the responsibility
+// of the caller to ensure it is correctly conditioned prior to the call.
+//
+// PermTo panics if m is not square.
+func (p *UniformPermutation) PermTo(m *mat.Dense) {
 	r, c := m.Dims()
 	if r != c {
 		panic(mat.ErrShape)
@@ -41,13 +36,14 @@ func (up *UniformPermutation) MatrixTo(m *mat.Dense) {
 	if r == 0 {
 		return
 	}
-	iList := make([]int, r)
-	for i := range iList {
-		iList[i] = i
+	if len(p.indices) != r {
+		p.indices = make([]int, r)
 	}
-	rnd := rand.New(up.src)
-	rnd.Shuffle(r, func(i, j int) { iList[i], iList[j] = iList[j], iList[i] })
-	for i, j := range iList {
-		m.Set(i, j, 1.0)
+	for k := range p.indices {
+		p.indices[k] = k
+	}
+	p.rnd.Shuffle(r, func(i, j int) { p.indices[i], p.indices[j] = p.indices[j], p.indices[i] })
+	for i, j := range p.indices {
+		m.Set(i, j, 1)
 	}
 }
--- a/stat/distmat/permutation.go
+++ b/stat/distmat/permutation.go
@@ -11,7 +11,7 @@ import (
 )
 
 // UniformPermutation is a uniform distribution over the n!
-// permutation matrices of size n x n for a given n.
+// permutation matrices of size n×n for a given n.
 type UniformPermutation struct {
 	rnd     *rand.Rand
 	indices []int
@@ -27,9 +27,9 @@ func NewUniformPermutation(src rand.Source) *UniformPermutation {
 // It does not zero the destination's elements, so it is the responsibility
 // of the caller to ensure it is correctly conditioned prior to the call.
 //
-// PermTo panics if m is not square.
-func (p *UniformPermutation) PermTo(m *mat.Dense) {
-	r, c := m.Dims()
+// PermTo panics if dst is not square.
+func (p *UniformPermutation) PermTo(dst *mat.Dense) {
+	r, c := dst.Dims()
 	if r != c {
 		panic(mat.ErrShape)
 	}
@@ -38,12 +38,12 @@ func (p *UniformPermutation) PermTo(m *mat.Dense) {
 	}
 	if len(p.indices) != r {
 		p.indices = make([]int, r)
-	}
-	for k := range p.indices {
-		p.indices[k] = k
+		for k := range p.indices {
+			p.indices[k] = k
+		}
 	}
 	p.rnd.Shuffle(r, func(i, j int) { p.indices[i], p.indices[j] = p.indices[j], p.indices[i] })
 	for i, j := range p.indices {
-		m.Set(i, j, 1)
+		dst.Set(i, j, 1)
 	}
 }
--- a/stat/distmat/permutation_test.go
+++ b/stat/distmat/permutation_test.go
@@ -6,11 +6,11 @@ package distmat
 
 import (
 	"math/cmplx"
-	"math/rand"
 	"testing"
 
-	"gonum.org/v1/gonum/floats"
+	"golang.org/x/exp/rand"
 
+	"gonum.org/v1/gonum/floats"
 	"gonum.org/v1/gonum/mat"
 )
 
--- a/stat/distmat/permutation_test.go
+++ b/stat/distmat/permutation_test.go
@@ -16,15 +16,7 @@ func TestUniformPermutation(t *testing.T) {
 	up := NewUniformPermutation(rand.NewSource(1))
 	for _, n := range []int{10, 32, 64, 100} {
 		m := mat.NewDense(n, n, nil)
-		if m == nil {
-			t.Error("Matrix failed")
-		}
 		up.PermTo(m)
-		// Ensure Dims() unchanged.
-		r, c := m.Dims()
-		if r != n || c != n {
-			t.Errorf("got back matrix of wrong size. expected %d x %d, got %d x %d", n, n, r, c)
-		}
 		// Test that each row and column satisfies the permutation matrix
 		// invariant that all rows and columns have a single unit element
 		// and the remaining elements are zero.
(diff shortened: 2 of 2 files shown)


stat/distmat/permutation.go
```diff
@@ -0,0 +1,53 @@
...
+
+// NewUniformPermutation constructs a new permutation matrix
+// generator using the given random source.
+func NewUniformPermutation(src rand.Source) *UniformPermutation {
+	up := UniformPermutation{src: src}
+	return &up
+}
+
+// Matrix draws a random permutuation matrix of dimension n.
+func (up *UniformPermutation) Matrix(n int) *mat.Dense {
```
romanwerpachowski, line 26: We have another matrix distribution in this package, `Wishart`, in which the function sampling a random matrix is called `RandSymTo`. Neither this name or `Matrix` seems very fitting to me. We should decide on a consistent API now that we have more than one distribution in the package.

@kortschak ?

stat/distmat/permutation.go
```diff
@@ -0,0 +1,53 @@
...
+	m := mat.NewDense(n, n, nil)
+	up.MatrixTo(m)
+	return m
+}
+
+// MatrixTo sets the given matrix to be a random permutation matrix
+// panics if m is not square
+// Note this does not zero the matrix -- before calling this ensure it
+// is full of 0s.
+func (up *UniformPermutation) MatrixTo(m *mat.Dense) {
```
romanwerpachowski, line 36: We have another matrix distribution in this package, Wishart, in which the function sampling a random matrix is called RandSymTo. Neither this name or `MatrixTo` seems very fitting to me. We should decide on a consistent API now that we have more than one distribution in the package.

@kortschak ?

stat/distmat/permutation.go
romanwerpachowski, line 26: I think this function is unnecessary.

stat/distmat/permutation.go
```diff
@@ -0,0 +1,53 @@
...
+// Matrix draws a random permutuation matrix of dimension n.
+func (up *UniformPermutation) Matrix(n int) *mat.Dense {
+	m := mat.NewDense(n, n, nil)
+	up.MatrixTo(m)
+	return m
+}
+
+// MatrixTo sets the given matrix to be a random permutation matrix
+// panics if m is not square
+// Note this does not zero the matrix -- before calling this ensure it
```
romanwerpachowski, line 34: Better something like that:

```
// MatrixTo sets the given matrix to be a random permutation matrix.
// Note that it does not zero the matrix. Before calling it, ensure m
// is full of 0s.
//
// MatrixTo panics if m is not square.
```

stat/distmat/permutation.go
```diff
@@ -0,0 +1,53 @@
...
+	if r == 0 {
+		return
+	}
+	iList := make([]int, r)
+	for i := range iList {
+		iList[i] = i
+	}
+	rnd := rand.New(up.src)
+	rnd.Shuffle(r, func(i, j int) { iList[i], iList[j] = iList[j], iList[i] })
+	for i, j := range iList {
```
romanwerpachowski, line 50: `i` is first used as the permuted index (line 45) and then as the row index (line 50). Can we have consistent naming?

stat/distmat/permutation_test.go
```diff
@@ -0,0 +1,63 @@
...
+	"golang.org/x/exp/rand"
+
+	"gonum.org/v1/gonum/floats"
+	"gonum.org/v1/gonum/mat"
+)
+
+func TestUniformPermutation(t *testing.T) {
+	up := NewUniformPermutation(rand.NewSource(1))
+	for _, n := range []int{10, 32, 64, 100} {
+		m := up.Matrix(n)
```
romanwerpachowski, line 20: I'd test `MatrixTo` directly.

stat/distmat/permutation_test.go
```diff
@@ -0,0 +1,63 @@
...
+	n, _ := m.Dims()
+	var e mat.Eigen
+	e.Factorize(m, mat.EigenLeft)
+	values := make([]complex128, n)
+	e.Values(values)
+	for _, v := range values {
+		a := cmplx.Abs(v)
+		if !floats.EqualWithinAbs(a, 1.0, 1e-12) {
+			return false
+		}
```
romanwerpachowski, line 44: This doesn't test that v^n = 1 for some integer n.

stat/distmat/permutation_test.go
```diff
@@ -0,0 +1,63 @@
...
+	}
+	return true
+}
+
+func confirmEachColumnOneNZ(m mat.Matrix) bool {
```
romanwerpachowski, line 49: This function could also test that m is square and that each column has a 1 in a different row and 0s everywhere else. Then the eigenvalue-based test (which is IMHO quite expensive) would not be necessary.

stat/distmat/permutation.go
kortschak: Agreed. I was thinking the same thing last night. In other places we use `{noun,verb}To(dst *T)` for things like this. Figuring out what the correct noun or verb is is not necessarily easy. I don't think `Matrix` is the right noun.

stat/distmat/permutation.go
kortschak: Agreed.

stat/distmat/permutation_test.go
kortschak: The because necessary when `Matrix` goes away.

stat/distmat/permutation.go
kortschak:
```
// MatrixTo sets the given matrix to be a random permutation matrix.
// It does not zero the destination's elements, so it is the responsibility
// of the caller to ensure it is correctly conditioned prior to the call.
//
// MatrixTo panics if m is not square.
```

stat/distmat/permutation.go
```diff
@@ -0,0 +1,53 @@
...
+// is full of 0s.
+func (up *UniformPermutation) MatrixTo(m *mat.Dense) {
+	r, c := m.Dims()
+	if r != c {
+		panic(mat.ErrShape)
+	}
+	if r == 0 {
+		return
+	}
+	iList := make([]int, r)
```
kortschak, line 44: This can be a field of the receiver and be reused.

stat/distmat/permutation.go
```diff
@@ -0,0 +1,53 @@
...
+
+package distmat
+
+import (
+	"golang.org/x/exp/rand"
+
+	"gonum.org/v1/gonum/mat"
+)
+
+// UniformPermutation is a distribution over permutation matrices
```
kortschak, line 13:
```
// UniformPermutation is a distribution over uniform permutation matrices.
```

stat/distmat/permutation.go
```diff
@@ -0,0 +1,53 @@
...
+		panic(mat.ErrShape)
+	}
+	if r == 0 {
+		return
+	}
+	iList := make([]int, r)
+	for i := range iList {
+		iList[i] = i
+	}
+	rnd := rand.New(up.src)
```
kortschak, line 48: This too; rather than keeping `p.src`, make the field `p.rnd` and have the constructor return `&UniformPermutation{rnd: rand.New(src)}`.

stat/distmat/permutation.go
jonreiter: I'm being very pedantic here as I needed to google that term.  Seems "uniform permutation matrix" is used, very occasionally in a handful of references, to mean "random permutation matrix uniformly chosen among all the N! choices." (https://arxiv.org/pdf/1111.4662.pdf)

So that change is fine as far as it goes.  But I'm unlikely to be the only person checking to see if there are other special properties (there are not).  A style question I guess on the vocabulary.

kortschak: If there is convention, use that. It may be worth providing a link if there is work on this that supports it.

stat/distmat/permutation.go
jonreiter: RandDenseTo would be consistent with wishart.  uniformPerm.RandDenseTo is a bit verbose though.

DenseTo() would go with SymTo() , CholTo() etc.  a distribution is surely spitting out random matrices so i don't think that is terribly unclear.

kortschak: FWIW, `uniformPerm.RandDenseTo(dst)` is not likely to exist in code anyway, you would be more likely to see `p.RandDenseTo(dst)`.

If you take a look at the factorisation types, they have things like `LTo`, `QTo`, etc. This reflects the semantics of the method rather than the type (this is also true for `SymTo` and `CholTo`, though the similarities between the semantics and the syntax confuse this). This is the basis of calling it `PermTo`.

stat/distmat/permutation_test.go
jonreiter: now directly checks for 1 1, n-1 0s and nothing else in each row and column.

stat/distmat/permutation.go
jonreiter: fair enough, i guess that is somewhat the price for not having overloading.  adjusted to PermTo.

stat/distmat/permutation.go
jonreiter: changed to: a uniform distribution over the n! permutation matrices of size n x n for a given n.  i can find no widely-used  naming and expect that statement is clear to the audience using gonum.

stat/distmat/permutation.go
```diff
@@ -0,0 +1,49 @@
...
+func NewUniformPermutation(src rand.Source) *UniformPermutation {
+	return &UniformPermutation{rnd: rand.New(src)}
+}
+
+// PermTo sets the given matrix to be a random permutation matrix.
+// It does not zero the destination's elements, so it is the responsibility
+// of the caller to ensure it is correctly conditioned prior to the call.
+//
+// PermTo panics if m is not square.
+func (p *UniformPermutation) PermTo(m *mat.Dense) {
```
kortschak, line 31: s/m/dst/g

stat/distmat/permutation.go
```diff
@@ -0,0 +1,49 @@
...
+// generator using the given random source.
+func NewUniformPermutation(src rand.Source) *UniformPermutation {
+	return &UniformPermutation{rnd: rand.New(src)}
+}
+
+// PermTo sets the given matrix to be a random permutation matrix.
+// It does not zero the destination's elements, so it is the responsibility
+// of the caller to ensure it is correctly conditioned prior to the call.
+//
+// PermTo panics if m is not square.
```
kortschak, line 30:
```
// PermTo sets the elements of a random permutation matrix into dst.
// It does not zero the destination's elements, so it is the responsibility
// of the caller to ensure it is correctly conditioned prior to the call.
//
// PermTo panics if dst is not square.
```

I'm not convinced about the first sentence's language yet, but I can't think of better wording yet.

stat/distmat/permutation.go
```diff
@@ -0,0 +1,49 @@
...
+	}
+	if r == 0 {
+		return
+	}
+	if len(p.indices) != r {
+		p.indices = make([]int, r)
+	}
+	for k := range p.indices {
+		p.indices[k] = k
+	}
```
kortschak, line 44: Strictly speaking, this could be in the conditional block above, but it likely makes very little performance impact.

stat/distmat/permutation_test.go
```diff
@@ -0,0 +1,62 @@
...
+		up.PermTo(m)
+		if !confirmEachRowAndColumnOneNZ(m) {
+			t.Error("didnt get back a permutation matrix")
+		}
+	}
+
+}
+
+func confirmEachRowAndColumnOneNZ(m mat.Matrix) bool {
+	r, c := m.Dims()
```
romanwerpachowski, line 32: You may want to test that `r == c` here, to make this function independent from how the test calling it is written. Up to you.

stat/distmat/permutation_test.go
```diff
@@ -0,0 +1,62 @@
...
+			} else {
+				return false
+			}
+			if floats.EqualWithinAbs(m.At(j, i), 0, 1e-12) {
+				colNZ++
+			} else if floats.EqualWithinAbs(m.At(j, i), 1, 1e-12) {
+				colN1++
+			} else {
+				return false
+			}
```
sbinet, line 52: perhaps use a `switch` instead of these if/else-if/else ?

stat/distmat/permutation_test.go
```diff
@@ -0,0 +1,62 @@
...
+		if m == nil {
+			t.Error("Matrix failed")
+		}
+		up.PermTo(m)
+		if !confirmEachRowAndColumnOneNZ(m) {
+			t.Error("didnt get back a permutation matrix")
+		}
+	}
+
+}
```
kortschak, line 29:
```
func TestUniformPermutation(t *testing.T) {
	up := NewUniformPermutation(rand.NewSource(1))
	for _, n := range []int{10, 32, 64, 100} {
		m := mat.NewDense(n, n, nil)
		up.PermTo(m)
		for i := 0; i < n; i++ {
			checkPerm(t, "row", i, mat.Row(nil, i, m))
			checkPerm(t, "col", i, mat.Col(nil, i, m))
		}
	}
}

func checkPerm(t *testing.T, dir string, n int, v []float64) {
	var sum float64
	for i, x := range v {
		switch x {
		case 0, 1:
			sum += x
		default:
			t.Errorf("unexpected value in %s %d position %d: %v", dir, n, i, v)
		}
	}
	if sum != 1 {
		t.Errorf("%s %d is not a valid vector: %v", dir, n, v)
	}
}
```

stat/distmat/permutation.go
romanwerpachowski: LGTM, thanks.

stat/distmat/permutation_test.go
romanwerpachowski: Thanks!

stat/distmat/permutation_test.go
romanwerpachowski: This doesn't check that the 1 appears in a different position in every row / column.

kortschak: > this doesn't check that the 1 appears in a different position in every row / column

I don't think that's the case; this tests that every row and every column sums to one which is not possible if there is more than a single unit element in any row or column if only unit and zero elements are allowed which is also checked. Can you show me a counter example that this doesn't find?

romanwerpachowski: I mean that the test is correct, but not terribly explicit.

kortschak: With the change of name of the helper I think it meets the usual Gonum test standards.

stat/distmat/permutation_test.go
```diff
@@ -0,0 +1,58 @@
...
+		}
+		up.PermTo(m)
+		r, c := m.Dims()
+		if r != n || c != n {
+			t.Error("got back matrix of wrong size")
+		}
+		confirmPermMatrix(t, m)
+	}
+
+}
```
kortschak, line 31: Please use the test code in https://github.com/gonum/gonum/pull/1417#discussion_r466353985

stat/distmat/permutation_test.go
romanwerpachowski: I guess you're right, the comment inside the loop is very clear. LGTM.

kortschak: Could also check that `var ppT mat.Dense; ppT.Mul(up, up.T())` gives you a `ppT` that is equal to I, but at that point we are testing mathematics rather than the implementation.

stat/distmat/permutation_test.go
```diff
@@ -21,38 +20,33 @@ func TestUniformPermutation(t *testing.T) {
 			t.Error("Matrix failed")
 		}
 		up.PermTo(m)
+		// Ensure Dims() unchanged.
 		r, c := m.Dims()
 		if r != n || c != n {
-			t.Error("got back matrix of wrong size")
+			t.Errorf("got back matrix of wrong size. expected %d x %d, got %d x %d", n, n, r, c)
+		}
```
kortschak: This test is really not necessary; we can see that `PermTo` does not alter the dimensions of `dst`. Similarly the check for a nil `m` is not needed; `NewDense` cannot fail without a panic.

Please delete L19-27; the extra lines don't add safety but they do distract from the substance of the test.

kortschak:
Thanks. I'll merge this when the tests have finished.

PR #1418, opened 2020-08-04
cmplxs,floats: factor out non-vector functions

kortschak:
Please take a look.


kortschak:
When I write the release notes, I go through the changes and comment in the blurb about breaking changes.

cmplxs/cscalar/cscalar.go
```diff
@@ -0,0 +1,94 @@
...
+
+import (
+	"math"
+	"math/cmplx"
+
+	"gonum.org/v1/gonum/floats/scalar"
+)
+
+// EqualWithinAbs returns true if a and b have an absolute difference
+// of less than tol.
```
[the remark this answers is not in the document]
kortschak: Done ("not greater than").

floats/scalar/scalar.go
```diff
@@ -0,0 +1,191 @@
...
+
+package scalar
+
+import (
+	"math"
+	"strconv"
+)
+
+// EqualWithinAbs returns true if a and b have an absolute
+// difference of less than tol.
```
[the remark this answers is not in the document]
kortschak: Done

cmplxs/cscalar/cscalar.go
```diff
@@ -0,0 +1,94 @@
...
+	delta := cmplx.Abs(a - b)
+	if delta <= minNormalFloat64 {
+		return delta <= tol*minNormalFloat64
+	}
+	// We depend on the division in this relationship to identify
+	// infinities.
+	return delta/math.Max(cmplx.Abs(a), cmplx.Abs(b)) <= tol
+}
+
+// EqualWithinAbsOrRel returns true if parts of a and b are equal to within
```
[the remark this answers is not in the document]
kortschak: Done

floats/scalar/scalar.go
```diff
@@ -0,0 +1,191 @@
...
+		return delta <= tol*minNormalFloat64
+	}
+	// We depend on the division in this relationship to identify
+	// infinities (we rely on the NaN to fail the test) otherwise
+	// we compare Infs of the same sign and evaluate Infs as equal
+	// independent of sign.
+	return delta/math.Max(math.Abs(a), math.Abs(b)) <= tol
+}
+
+// EqualWithinAbsOrRel returns true if a and b are equal to within
```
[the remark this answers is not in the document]
kortschak: Done

kortschak:
I have removed the "derecated.go" file in cmplxs since this package was new since v0.7, so we don't have any need to be gentle here.

PTAL

cmplxs/cscalar/cscalar.go
```diff
@@ -0,0 +1,91 @@
...
+// the absolute or relative tolerances.
+func EqualWithinAbsOrRel(a, b complex128, absTol, relTol float64) bool {
+	return EqualWithinAbs(a, b, absTol) || EqualWithinRel(a, b, relTol)
+}
+
+// ParseWithNA converts the string s to a complex128 in v.
+// If s equals missing, w is returned as 0, otherwise 1.
```
romanwerpachowski, line 46: This is quite cryptic. Could the doc explain what is the purpose of `w`?

cmplxs/cscalar/cscalar.go
```diff
@@ -0,0 +1,91 @@
...
+func RoundEven(x complex128, prec int) complex128 {
+	if x == 0 {
+		// Make sure zero is returned
+		// without the negative bit set.
+		return 0
+	}
+	return complex(scalar.RoundEven(real(x), prec), scalar.RoundEven(imag(x), prec))
+}
+
+// Same returns true if the inputs have the same value with NaN treated as the same.
```
romanwerpachowski, line 88: I think "with two NaNs" would be clearer.

floats/scalar/scalar.go
```diff
@@ -0,0 +1,188 @@
...
+// EqualWithinAbs returns true if a and b have an absolute difference
+// not greater than tol.
+func EqualWithinAbs(a, b, tol float64) bool {
+	return a == b || math.Abs(a-b) <= tol
+}
+
+const minNormalFloat64 = 2.2250738585072014e-308
+
+// EqualWithinRel returns true if the difference between a and b
+// is not greater than tol times the greater value.
```
romanwerpachowski, line 21: "greater absolute value"

floats/scalar/scalar.go
```diff
@@ -0,0 +1,188 @@
...
+func NaNPayload(f float64) (payload uint64, ok bool) {
+	b := math.Float64bits(f)
+	if b&nanBits != nanBits {
+		return 0, false
+	}
+	return b &^ nanMask, true
+}
+
+// ParseWithNA converts the string s to a float64 in v.
+// If s equals missing, w is returned as 0, otherwise 1.
```
romanwerpachowski, line 89: This is quite cryptic. Can we have a comment explaining what the purpose of `w` is?

floats/scalar/scalar.go
```diff
@@ -0,0 +1,188 @@
...
+	return x / pow
+}
+
+func isHalfway(x float64) bool {
+	_, frac := math.Modf(x)
+	frac = math.Abs(frac)
+	return frac == 0.5 || (math.Nextafter(frac, math.Inf(-1)) < 0.5 && math.Nextafter(frac, math.Inf(1)) > 0.5)
+}
+
+// Same returns true if the inputs have the same value with NaN treated as the same.
```
romanwerpachowski, line 185: "with two NaNs"

cmplxs/cscalar/cscalar.go
kortschak: It indicates a weight, but I don't think it's cryptic, it says what it does.

floats/scalar/scalar.go
kortschak: As above.

cmplxs/cscalar/cscalar.go
kortschak: `// Same returns true if the inputs have the same value, allowing NaN equality.`?

floats/scalar/scalar.go
kortschak: I'll copy the comment from the cscalar function.

floats/scalar/scalar.go
kortschak: As above?

cmplxs/cscalar/cscalar.go
romanwerpachowski: Can it say that `w` is a weight? Knowing that, everything makes sense to me now.

cmplxs/cscalar/cscalar.go
romanwerpachowski: LGTM.

floats/scalar/scalar.go
romanwerpachowski: LGTM.

floats/scalar/scalar.go
romanwerpachowski: Can it say that w is a weight? Knowing that, everything makes sense to me now.

floats/scalar/scalar.go
romanwerpachowski: SGTM.

cmplxs/cscalar/cscalar.go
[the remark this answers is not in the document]
vladimir-ch: I don't want to nitpick too much but what does the "parts of" refer to? The comparison internally relies on `cmplx.Abs(a-b)` which considers the complex numbers as a whole, not as real and imaginary parts separately. Isn't this a left-over from a previous implementation?

cmplxs/cscalar/cscalar.go
kortschak: Please read it in the context of a godoc rendering (the code it old, so you can see it [here](https://godoc.org/gonum.org/v1/gonum/floats#ParseWithNA)). Does it still need to be expanded?

Part of the reason I'm pushing back is that there's a philosophical (in the correct sense) issue with saying that `w` is a weight; it may be used as one, and that is the intention, but it's not a weight, it is a conditional value that is `float64(0)` when the input matches the missing value text and `float64(1)` otherwise. It *can* be used as a weight, and this is demonstrated in the example immediately below the godoc signature and description.

cmplxs/cscalar/cscalar.go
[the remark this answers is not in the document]
kortschak: >Isn't this a left-over from a previous implementation?

I think that that is exactly why it's wrong. Fixed (also in scalar, where it's completely wrong).

kortschak:
PTAL

cmplxs/cscalar/cscalar.go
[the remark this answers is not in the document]
vladimir-ch: Thanks.

cmplxs/cscalar/cscalar.go
romanwerpachowski: But when I'm reading the code (say, when refactoring this file), I don't see the godoc rendering. IMHO locality of information is useful for code readability.

Why not just add what you wrote above: "w is intended to be used as a weight for the parsed value"?

floats/scalar/scalar.go
kortschak: Expanded reluctantly.

cmplxs/cscalar/cscalar.go
kortschak: Expanded reluctantly.

romanwerpachowski: Thanks.

floats/scalar/scalar.go
romanwerpachowski: Thanks.

PR #1421, opened 2020-08-12
cmplxs: add Real, Imag and Complex functions

### LINKED ISSUE #1420: cmplxs: add Real, Imag and Complex functions
I can see a benefit to adding a pair of functions to obtain the real and imaginary components of elements of a `[]complex128` and a cognate to compose a pair of `[]float64` into a `[]complex128`.

This would be

```
func Imag(dst []float64, src []complex128) []float64

func Real(dst []float64, src []complex128) []float64

func Complex(dst []complex128, real, imag []float64) []complex128
```

kortschak:
Please take a look.

Closes #1420.

### DIFF
--- a/cmplxs/cmplxs.go
+++ b/cmplxs/cmplxs.go
@@ -85,6 +85,22 @@ func Count(f func(complex128) bool, s []complex128) int {
 	return n
 }
 
+// Complex fills each of the elements of dst with the complex number
+// constructed from the corresponding elements of real and imag.
+// A panic will occur if the lengths of arguments do not match.
+func Complex(dst []complex128, real, imag []float64) []complex128 {
+	if len(dst) != len(real) || len(dst) != len(imag) {
+		panic("cmplxs: length of destination does not match length of the source")
+	}
+	if len(dst) == 0 {
+		return dst
+	}
+	for i, r := range real {
+		dst[i] = complex(r, imag[i])
+	}
+	return dst
+}
+
 // CumProd finds the cumulative product of elements of s and store it in
 // place into dst so that
 //  dst[i] = s[i] * s[i-1] * s[i-2] * ... * s[0]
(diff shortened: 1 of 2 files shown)


cmplxs/cmplxs.go
```diff
@@ -85,6 +85,22 @@ func Count(f func(complex128) bool, s []complex128) int {
 	return n
 }
 
+// Complex fills each of the elements of dst with the complex number
+// constructed from the corresponding elements of real and imag.
+// A panic will occur if the lengths of arguments do not match.
+func Complex(dst []complex128, real, imag []float64) []complex128 {
+	if len(dst) != len(real) || len(dst) != len(imag) {
+		panic("cmplxs: length of destination does not match length of the source")
```
sbinet, line 93: lift as an unexposed error?
(it's also used in `Imag` and `Real`, with same wording)

cmplxs/cmplxs.go
```diff
@@ -85,6 +85,22 @@ func Count(f func(complex128) bool, s []complex128) int {
 	return n
 }
 
+// Complex fills each of the elements of dst with the complex number
+// constructed from the corresponding elements of real and imag.
+// A panic will occur if the lengths of arguments do not match.
```
sbinet, line 90: it seems to me we usually don't use much passive form for that kind of documentation.
_i.e.:_ I think we usually word it as:
```
// Complex panics if ...
```

kortschak: This wording is lifted directly from another function.

cmplxs/cmplxs.go
kortschak: I think that probably this and the other comment should be addressed in a sweep, this panic string exists in a variety of other functions as well.

kortschak:
I have filed #1422 to address the issues raised more generally.

PR #1423, opened 2020-08-12
cmplxs,floats: harmonise docs and panic strings

### LINKED ISSUE #1422: cmplxs,floats: clean up docs and panic strings
See https://github.com/gonum/gonum/pull/1421#discussion_r469354738 and https://github.com/gonum/gonum/pull/1421#discussion_r469358082.

kortschak:
Please take a look.

Fixes #1422.

### DIFF
--- a/cmplxs/cmplxs.go
+++ b/cmplxs/cmplxs.go
@@ -13,11 +13,18 @@ import (
 	"gonum.org/v1/gonum/internal/asm/c128"
 )
 
+const (
+	zeroLength   = "cmplxs: zero length slice"
+	shortSpan    = "cmplxs: slice length less than 2"
+	badLength    = "cmplxs: slice lengths do not match"
+	badDstLength = "cmplxs: destination slice length does not match input"
+)
+
 // Abs calculates the absolute values of the elements of s, and stores them in dst.
-// It panics if the lengths of dst and s do not match.
+// It panics if the argument lengths do not match.
 func Abs(dst []float64, s []complex128) {
 	if len(dst) != len(s) {
-		panic("cmplxs: length of the slices do not match")
+		panic(badDstLength)
 	}
 	for i, v := range s {
 		dst[i] = cmplx.Abs(v)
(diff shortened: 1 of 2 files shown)


cmplxs/cmplxs.go
```diff
@@ -395,23 +413,26 @@ func MinAbsIdx(s []complex128) int {
...
 
 // Mul performs element-wise multiplication between dst
-// and s and stores the result in dst. Panics if the
-// lengths of s and t are not equal.
+// and s and stores the result in dst.
+// It panics if the argument lengths do not match.
 func Mul(dst, s []complex128) {
 	if len(dst) != len(s) {
-		panic("cmplxs: slice lengths do not match")
+		panic(badLength)
```
vladimir-ch, line 420:
```suggestion
		panic(badDstLength)
```

cmplxs/cmplxs.go
```diff
@@ -52,22 +60,25 @@ func AddConst(c complex128, dst []complex128) {
 }
 
 // AddScaled performs dst = dst + alpha * s.
-// It panics if the lengths of dst and s are not equal.
+// It panics if the slice argument lengths do not match.
 func AddScaled(dst []complex128, alpha complex128, s []complex128) {
 	if len(dst) != len(s) {
-		panic("cmplxs: length of destination and source to not match")
+		panic(badLength)
```
vladimir-ch, line 66:
```suggestion
		panic(badDstLength)
```

cmplxs/cmplxs.go
```diff
@@ -165,31 +179,34 @@ func Distance(s, t []complex128, L float64) float64 {
...
 
 // Div performs element-wise division dst / s
-// and stores the result in dst. It panics if the
-// lengths of s and t are not equal.
+// and stores the result in dst.
+// It panics if the argument lengths do not match.
 func Div(dst, s []complex128) {
 	if len(dst) != len(s) {
-		panic("cmplxs: slice lengths do not match")
+		panic(badLength)
```
vladimir-ch, line 186:
```suggestion
		panic(badDstLength)
```

cmplxs/cmplxs.go
```diff
@@ -605,23 +626,24 @@ func Span(dst []complex128, l, u complex128) []complex128 {
...
 }
 
-// Sub subtracts, element-wise, the elements of s from dst. Panics if
-// the lengths of dst and s do not match.
+// Sub subtracts, element-wise, the elements of s from dst.
+// It panics if the argument lengths do not match.
 func Sub(dst, s []complex128) {
 	if len(dst) != len(s) {
-		panic("cmplxs: length of the slices do not match")
+		panic(badLength)
```
vladimir-ch, line 633:
```suggestion
		panic(badDstLength)
```

cmplxs/cmplxs.go
kortschak: I agonised about these; in the case of `Mul` and other work-in-place functions, I think that `badLength` is actually a better message. It looks surprising because of the name of the source/destination parameter, but I think that it would be more confusing if you saw the message "destination slice length does not match input" for a call `floats.Mul(a, b)` when you think of both `a` and `b` as inputs.

vladimir-ch: Ok, seems you gave it more thought that I did :-)

PR #1425, opened 2020-08-12
ci: bump travis, appveyor and module versions

kortschak:
Please take a look.

### DIFF
--- a/version.go
+++ b/version.go
@@ -2,8 +2,6 @@
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
-// +build go1.12
-
 package gonum
 
 import (
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,8 +5,8 @@ language: go
 # Do not move these lines; they are referred to by README.md.
 # Versions of go that are explicitly supported by Gonum, plus Go tip.
 go:
+ - 1.15.x
  - 1.14.x
- - 1.13.x
  - master
 
 os:
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,27 +1,24 @@
 build: off
 
+image: Visual Studio 2019
+
+stack: go 1.14
+
 clone_folder: c:\gopath\src\gonum.org\v1\gonum
 
 cache:
   - '%LocalAppData%\go-build'
 
+environment:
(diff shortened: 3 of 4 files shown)


kortschak:
@sbinet Can we drop appveyor? They are always late and the documentation is pretty poor.

sbinet:
Sure.
As long as we keep a Windows build+test, the ci platform doesn't matter much.
(We could try the GitHub one if/when the Travis one is getting overcrowded)

kortschak:
It's working now though, so I'll file an issue to deal with that.

kortschak:
*ping*

version.go
```diff
@@ -2,8 +2,6 @@
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
-// +build go1.12
```
sbinet, line 5: Is that really necessary?

kortschak: We don't support Go 1.12, so deleting it is fine. We needed it when we added this file (4d9eed1802090095b5558e6483bb79999fa51a7a) because we also supported Go 1.11 and 1.10, which didn't have the function that this code depends on.

sbinet: Leaving it isn't bugging us either.
But ok.

kortschak: Code that isn't doing anything does hurt.

kortschak: I should expand on this for this particular case.

We have two options, to leave the build tag in, or take it out. In the happy case where everything works fine because people have maintained their build environment this is little cost (some because when we read the code, we need to think about why it is that we have a build tag for a version that is not supported by the project). In the unhappy case where someone is using a pre-go1.12 build environment the outcomes are quite different:

1. with build tag we see an error like 
    ```
    ./main.go:10:14: undefined: gonum.Version
    ```
   which is an indirection of the issue.

2. without the build tag, we see an error like
    ```
    # gonum.org/v1/gonum
    gonum.org/v1/gonum/version.go:26:11: undefined: debug.ReadBuildInfo
    ```
    which clearly points to the issue.

It's also worth noting that there are a swag of other things where we have used the most recent possible version and removed the build tag that protects this. The tracking of recent versions is sometimes mechanical (`range` expression simplification for example) and sometimes for readability (binary exponent float representation).

PR #1432, opened 2020-08-22
stat/distmat: UnitVector, uniform distribution over surface of n-sphere

jonreiter:
Adds a UnitVector distribution.  This is a uniform distribution over the surface of a sphere.

### DIFF
--- a/stat/distmat/general.go
+++ b/stat/distmat/general.go
@@ -5,3 +5,4 @@
 package distmat
 
 const badDim = "distmat: dimension mismatch"
+const zeroDim = "distmat: zero dimension"
--- a/stat/distmat/general.go
+++ b/stat/distmat/general.go
@@ -4,5 +4,7 @@
 
 package distmat
 
-const badDim = "distmat: dimension mismatch"
-const zeroDim = "distmat: zero dimension"
+const (
+	badDim  = "distmat: dimension mismatch"
+	zeroDim = "distmat: zero dimension"
+)
--- a/stat/distmat/unit_vector.go
+++ b/stat/distmat/unit_vector.go
@@ -1,0 +1,39 @@
+package distmat
+
+import (
+	"golang.org/x/exp/rand"
+
+	"gonum.org/v1/gonum/mat"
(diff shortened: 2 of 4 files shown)


stat/distmat/unit_vector.go
```diff
@@ -0,0 +1,50 @@
...
+package distmat
+
+import (
+	"golang.org/x/exp/rand"
+
+	"gonum.org/v1/gonum/mat"
+	"gonum.org/v1/gonum/stat/distuv"
+)
+
+const unitVecMinLength = 1e-5
```
romanwerpachowski, line 14: This is a curious name: a unit name has length 1 by definition. Maybe `sampledVecMinLength`?

romanwerpachowski (changes requested):

stat/distmat/unit_vector.go
kortschak: A comment explaining would be helpful, or moving it to immediately near where it's used and calling it `minLength` so that the context explains it.

stat/distmat/unit_vector_test.go
```diff
@@ -0,0 +1,66 @@
...
+	tot := 0.0
+	for i := 0; i < int(n); i++ {
+		tot += rnd.Float64()
+	}
+	tot = tot / n
+	if !scalar.EqualWithinAbs(tot, 0.5, 1e-4) {
+		t.Errorf("floating problem: %v", tot)
+	}
+
+}
```
kortschak, line 66: What is this for?

romanwerpachowski (approved):

stat/distmat/general.go
```diff
@@ -5,3 +5,4 @@
 package distmat
 
 const badDim = "distmat: dimension mismatch"
+const zeroDim = "distmat: zero dimension"
```
kortschak, line 8: Last thing.
```
const (
    badDim  = "distmat: dimension mismatch"
    zeroDim = "distmat: zero dimension"
)
```

kortschak (changes requested):

kortschak (approved):
Thank

PR #1435, opened 2020-08-25
interp: Add Fritsch-Butland piecewise cubic interpolation method

romanwerpachowski:
Also adds example of fit output for piecewise constant, piecewise linear, Akima and Fritsch-Butland methods.

Please take a look.


interp/interp_example_test.go
```diff
@@ -0,0 +1,91 @@
...
+func ExamplePredictor() {
+	xs := []float64{0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11}
+	ys := []float64{0, 0.001, 0.002, 0.1, 1, 2, 2.5, -10, -10.01, 2.49, 2.53, 2.55}
+
+	var pc interp.PiecewiseConstant
+	var pl interp.PiecewiseLinear
+	var as interp.AkimaSpline
+	var fb interp.FritschButland
+
+	_ = pc.Fit(xs, ys)
```
sbinet, line 23: ignoring errors in an example might be an incentive for "bad" code spreading in the wild.

please consider either adding the `3x4` lines of error code checking, or create a slice `[]interp.Fitter{&pc,&pl,&as,&fb}` and apply the fit in a for-loop.

interp/interp_example_test.go
```diff
@@ -0,0 +1,91 @@
...
+		x := xs[0] + float64(i)*dx
+		fmt.Printf("%.2f", x)
+		for _, predictor := range predictors {
+			y := predictor.Predict(x)
+			fmt.Printf("\t%.2f", y)
+		}
+		fmt.Printf("\n")
+	}
+	// Output:
+	// x		PC		PL		AS		FB
```
sbinet, line 45: it's not really visually super appealing.
couldn't we find a better formatting strategy?
use [text/tabwriter](https://godoc.org/text/tabwriter) perhaps?

interp/interp_example_test.go
```diff
@@ -0,0 +1,91 @@
...
+	dx := 0.25
+	nPts := int(math.Round(float64(n-1)/dx)) + 1
+
+	fmt.Printf("x\t\tPC\t\tPL\t\tAS\t\tFB\n")
+	for i := 0; i < nPts; i++ {
+		x := xs[0] + float64(i)*dx
+		fmt.Printf("%.2f", x)
+		for _, predictor := range predictors {
+			y := predictor.Predict(x)
+			fmt.Printf("\t%.2f", y)
```
sbinet, line 40:
```suggestion
			fmt.Printf("\t\t%.2f", y)
```
to better align column labels and column data?

interp/interp_example_test.go
romanwerpachowski: On my Windows PC the output was correct using the current code. I'll check out tabwriter.

interp/interp_example_test.go
romanwerpachowski: See above.

interp/interp_example_test.go
```diff
@@ -0,0 +1,91 @@
...
+	"fmt"
+	"math"
+
+	"gonum.org/v1/gonum/interp"
+)
+
+func ExamplePredictor() {
+	xs := []float64{0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11}
+	ys := []float64{0, 0.001, 0.002, 0.1, 1, 2, 2.5, -10, -10.01, 2.49, 2.53, 2.55}
+
```
kortschak, line 17: Can we have some commentary in this example telling the user what they are supposed to be seeing or getting from it?

interp/interp_example_test.go
kortschak: It's not safe to assume that everyone has the same tabstops set, please use the tabwriter. It won't make much difference to the code here since it take tab separated lines as input.

romanwerpachowski:
All done. PTAL.

sbinet (approved):

kortschak (approved):

PR #1437, opened 2020-09-06
dsp/window: add Tukey

brandondube:
This PR adds the tukey window and changes the Gaussian test infrastructure to handle any parametric window of one parameter.

I did not modify the complex interface.  These changes are from before my suggestion on gonum-dev to change the API for complex transformations.  After this is merged, I will make a PR for that.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -12,7 +12,6 @@ Alexander Egurnov <alexander.egurnov@gmail.com>
 Andrei Blinnikov <goofinator@mail.ru>
 Bill Gray <wgray@gogray.com>
 Bill Noon <noon.bill@gmail.com>
-Brandon Dube <brandon.dube@jpl.nasa.gov> <brandon@retrorefractions.com>
 Brendan Tracey <tracey.brendan@gmail.com>
 Brent Pedersen <bpederse@gmail.com>
 Chad Kunde <kunde21@gmail.com>
--- a/AUTHORS
+++ b/AUTHORS
@@ -12,6 +12,7 @@ Alexander Egurnov <alexander.egurnov@gmail.com>
 Andrei Blinnikov <goofinator@mail.ru>
 Bill Gray <wgray@gogray.com>
 Bill Noon <noon.bill@gmail.com>
+Brandon Dube <brandon.dube@jpl.nasa.gov> <brandon@retrorefractions.com>
 Brendan Tracey <tracey.brendan@gmail.com>
 Brent Pedersen <bpederse@gmail.com>
 Chad Kunde <kunde21@gmail.com>
--- a/dsp/window/window.go
+++ b/dsp/window/window.go
@@ -330,7 +330,7 @@ func (g Gaussian) Transform(seq []float64) []float64 {
 	return seq
 }
 
-// B Dube, 2020-09-06: the values for specrtal leakage come from
+// the values for specrtal leakage come from
 // A.D. Poularikas, "The Handbook of Formulas and Tables for Signal Processing"
 // table 7.1
 
@@ -338,49 +338,43 @@ func (g Gaussian) Transform(seq []float64) []float64 {
 // See https://en.wikipedia.org/wiki/Window_function#Tukey_window
 // and https://prod-ng.sandia.gov/techlib-noauth/access-control.cgi/2017/174042.pdf page 88
 //
-// The Tukey window is an adjustible window.  It can be thought of as something
+// The Tukey window is an adjustible window. It can be thought of as something
 // between a rectangular and a Hann window, with a flat center and tapered edges.
 //
-// The properties of the window depend on the value of α (alpha).  α controls
-// the fraction of the window which contains a cosine taper.  α = 0.5 gives a
+// The properties of the window depend on the value of α (alpha). α controls
+// the fraction of the window which contains a cosine taper. α = 0.5 gives a
 // window whos central 50% is a flat top and outer quartiles are tapered.
 // 0 < α < 1; if α is outside the bounds, it is treated as 0 or 1.
 //
-// Spectral leakage parameters for alpha=0.5 are
+// Spectral leakage parameters for Alpha=0.5 are
 // ΔF_0 = 1.22, ΔF_0.5 = 1.15, K = 1.3, ɣ_max = -15, β = -2.5.
 type Tukey struct {
 	Alpha float64
 }
 
 // Transform applies the Tukey transformation to seq in place, using the value
-// of the receiver as the sigma parameter, and returning the result
-func (tuk Tukey) Transform(seq []float64) []float64 {
-	if tuk.Alpha <= 0 {
+// of the receiver as the Alpha parameter, and returning the result
+func (t Tukey) Transform(seq []float64) []float64 {
+	if t.Alpha <= 0 {
 		return Rectangular(seq)
-	} else if tuk.Alpha >= 1 {
+	} else if t.Alpha >= 1 {
 		return Hann(seq)
 	}
 
-	var (
-		cos = math.Cos
-		pi  = math.Pi
-		a   = tuk.Alpha
-	)
-
-	sep := 0.5 * a * float64(len(seq)-1)
+	sep := 0.5 * t.Alpha * float64(len(seq)-1)
 	sepI := int(math.Floor(sep)) + 1
-	sep2 := float64((len(seq) - 1)) * (1 - .5*a)
+	sep2 := float64((len(seq) - 1)) * (1 - .5*t.Alpha)
 	sep2I := int(math.Ceil(sep2))
 	for i := range seq {
 		var w float64
 		if i < sepI {
 			ii := float64(i)
-			w = 0.5 * (1 + cos(pi*(ii/sep-1)))
+			w = 0.5 * (1 + math.Cos(math.Pi*(ii/sep-1)))
 		} else if i < sep2I {
 			w = 1
 		} else {
 			ii := float64(i)
-			w = 0.5 * (1 + cos(pi*(ii/sep-2/a+1)))
+			w = 0.5 * (1 + math.Cos(math.Pi*(ii/sep-2/t.Alpha+1)))
 		}
 		seq[i] *= w
 	}
--- a/dsp/window/window.go
+++ b/dsp/window/window.go
@@ -330,6 +330,63 @@ func (g Gaussian) Transform(seq []float64) []float64 {
 	return seq
 }
 
+// B Dube, 2020-09-06: the values for specrtal leakage come from
+// A.D. Poularikas, "The Handbook of Formulas and Tables for Signal Processing"
+// table 7.1
+
+// Tukey can modify a sequence by the Tukey window and return the result.
+// See https://en.wikipedia.org/wiki/Window_function#Tukey_window
+// and https://prod-ng.sandia.gov/techlib-noauth/access-control.cgi/2017/174042.pdf page 88
+//
+// The Tukey window is an adjustible window.  It can be thought of as something
+// between a rectangular and a Hann window, with a flat center and tapered edges.
+//
+// The properties of the window depend on the value of α (alpha).  α controls
+// the fraction of the window which contains a cosine taper.  α = 0.5 gives a
+// window whos central 50% is a flat top and outer quartiles are tapered.
+// 0 < α < 1; if α is outside the bounds, it is treated as 0 or 1.
+//
+// Spectral leakage parameters for alpha=0.5 are
+// ΔF_0 = 1.22, ΔF_0.5 = 1.15, K = 1.3, ɣ_max = -15, β = -2.5.
+type Tukey struct {
+	Alpha float64
+}
+
+// Transform applies the Tukey transformation to seq in place, using the value
+// of the receiver as the sigma parameter, and returning the result
+func (tuk Tukey) Transform(seq []float64) []float64 {
+	if tuk.Alpha <= 0 {
+		return Rectangular(seq)
+	} else if tuk.Alpha >= 1 {
+		return Hann(seq)
+	}
+
+	var (
+		cos = math.Cos
+		pi  = math.Pi
+		a   = tuk.Alpha
+	)
+
+	sep := 0.5 * a * float64(len(seq)-1)
+	sepI := int(math.Floor(sep)) + 1
+	sep2 := float64((len(seq) - 1)) * (1 - .5*a)
+	sep2I := int(math.Ceil(sep2))
+	for i := range seq {
+		var w float64
+		if i < sepI {
+			ii := float64(i)
+			w = 0.5 * (1 + cos(pi*(ii/sep-1)))
+		} else if i < sep2I {
+			w = 1
+		} else {
+			ii := float64(i)
+			w = 0.5 * (1 + cos(pi*(ii/sep-2/a+1)))
+		}
+		seq[i] *= w
+	}
+	return seq
+}
+
 // Values is an arbitrary real window function.
 type Values []float64
 
--- a/dsp/window/window.go
+++ b/dsp/window/window.go
@@ -330,10 +330,6 @@ func (g Gaussian) Transform(seq []float64) []float64 {
 	return seq
 }
 
-// The values for spectral leakage come from
-// A.D. Poularikas, "The Handbook of Formulas and Tables for Signal Processing"
-// table 7.1
-
 // Tukey can modify a sequence using the Tukey window and return the result.
 // See https://en.wikipedia.org/wiki/Window_function#Tukey_window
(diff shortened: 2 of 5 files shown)


dsp/window/window.go
```diff
@@ -330,6 +330,63 @@ func (g Gaussian) Transform(seq []float64) []float64 {
...
+			ii := float64(i)
+			w = 0.5 * (1 + cos(pi*(ii/sep-1)))
+		} else if i < sep2I {
+			w = 1
+		} else {
+			ii := float64(i)
+			w = 0.5 * (1 + cos(pi*(ii/sep-2/a+1)))
+		}
+		seq[i] *= w
+	}
```
kortschak, line 386: This seems more convoluted than it needs to be. The WP article and the sandia paper have simpler work to do this.

```

	alphaL := t.Alpha * float64(len(seq))
	width := 0.5 * alphaL
	for i := range seq[:(len(seq)+1)/2] {
		if i >= int(width) {
			// Do nothing within the rectangular window.
			continue
		}
		w := 0.5 * (1 - math.Cos(2*math.Pi*float64(i)/alphaL))
		seq[i] *= w
		seq[len(seq)-1-i] *= w
	}
```

brandondube: The code you pasted does not give a correct answer:
```
--- FAIL: TestMonoParametricWindows (0.00s)
    --- FAIL: TestMonoParametricWindows/Tukey_(param=0.5) (0.00s)
        window_test.go:227: unexpected result for window function "Tukey":
            got:[]float64{0, 0.0954915, 0.345492, 0.654508, 0.904508, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 0.904508, 0.654508, 0.345492, 0.0954915, 0}
            want:[]float64{0, 0.10543, 0.377257, 0.700847, 0.939737, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 0.939737, 0.700847, 0.377257, 0.105429, 0}
        window_test.go:236: unexpected result for lookup window function "Tukey":
            got:[]float64{0, 0.0954915, 0.345492, 0.654508, 0.904508, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 0.904508, 0.654508, 0.345492, 0.0954915, 0}
            want:[]float64{0, 0.10543, 0.377257, 0.700847, 0.939737, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 0.939737, 0.700847, 0.377257, 0.105429, 0}
FAIL
FAIL	gonum.org/v1/gonum/dsp/window	0.323s
FAIL
```

I do not agree that it is simpler, either.  There are fewer operations but the code is more difficult (for me) to understand.

brandondube:
I made most of the changes your requested, excepting the one that gives incorrect output.  PTAL, thanks!

dsp/window/window.go
kortschak: The difference is an off by one in the definition of `alphaL` If you change that to `alphaL := t.Alpha * float64(len(seq)+1)` then you get the same result. I was working from the definition of L that is in the WP article. It's unclear to me why you are subtracting 1. However, that change giving this
```
        alphaL := t.Alpha * float64(len(seq)-1)
        width := 0.5*alphaL + 1
        for i := range seq[:(len(seq)+1)/2] {
                if i >= int(width) {
                        // Do nothing within the rectangular window.
                        continue
                }
                w := 0.5 * (1 - math.Cos(2*math.Pi*float64(i)/alphaL))
                seq[i] *= w
                seq[len(seq)-1-i] *= w
        }
```
now matches your expected output.

The reason I argue this is simpler is because it (in order of importance IMO) 1. reflects the symmetry of the window function (it works only on the lobes and does them symmetrically - it reflects the definition in the sandia paper), 2. does less work and 3. does not perform operations that already result from type conversions.

dsp/window/window.go
```diff
@@ -330,6 +330,63 @@ func (g Gaussian) Transform(seq []float64) []float64 {
 	return seq
 }
 
+// B Dube, 2020-09-06: the values for specrtal leakage come from
```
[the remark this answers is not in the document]
kortschak: Can you fix the grammar and spelling. Actually, have a table at the bottom of the comment as exists for Gaussian.

```
// Spectral leakage parameters are summarized in the table:
//         | v=0.25 |  α=0.5 | α=0.75 |
//  -------|--------------------------|
//  ΔF_0   |   1.10 |   1.22 |   1.36 |
//  ΔF_0.5 |   1.01 |   1.15 |   1.31 |
//  K      |    efr |   1.3  |   efr  |
//  ɣ_max  | -14    | -15    | -19    |
//  β      |    efr |  -2.5  |   efr  |
// Values for spectral leakage come from A.D. Poularikas, "The Handbook of
// Formulas and Tables for Signal Processing" table 7.1.
```

I have left values not in table 7.1 or your comment as exercises for the reader.

dsp/window/window_test.go
```diff
@@ -113,31 +122,53 @@ var windowTests = []struct {
 	},
 }
 
-var gausWindowTests = []struct {
+var monoParamWindowTests = []struct {
 	name  string
-	sigma float64
+	param float64
+	win   parametricWindow
```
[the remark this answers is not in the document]
kortschak: `windowType`, please. We prefer tests to be more explicit.

dsp/window/window.go
kortschak: On more thinking, it simplifies to
```
	alphaL := t.Alpha * float64(len(seq)-1)
	width := int(0.5*alphaL) + 1
	for i := range seq[:width] {
		w := 0.5 * (1 - math.Cos(2*math.Pi*float64(i)/alphaL))
		seq[i] *= w
		seq[len(seq)-1-i] *= w
	}
```

Please use this. It improves on the previous as it only iterates for the lobes, so for lower alpha, there is a saving.

dsp/window/window.go
```diff
@@ -330,6 +330,57 @@ func (g Gaussian) Transform(seq []float64) []float64 {
 	return seq
 }
 
+// the values for specrtal leakage come from
```
romanwerpachowski, line 333: Typo.

romanwerpachowski, line 333: "The values"

dsp/window/window.go
```diff
@@ -330,6 +330,57 @@ func (g Gaussian) Transform(seq []float64) []float64 {
...
+// A.D. Poularikas, "The Handbook of Formulas and Tables for Signal Processing"
+// table 7.1
+
+// Tukey can modify a sequence by the Tukey window and return the result.
```
romanwerpachowski, line 337: "using the Tukey window"?

dsp/window/window.go
```diff
@@ -330,6 +330,57 @@ func (g Gaussian) Transform(seq []float64) []float64 {
...
+// See https://en.wikipedia.org/wiki/Window_function#Tukey_window
+// and https://prod-ng.sandia.gov/techlib-noauth/access-control.cgi/2017/174042.pdf page 88
+//
+// The Tukey window is an adjustible window. It can be thought of as something
+// between a rectangular and a Hann window, with a flat center and tapered edges.
+//
+// The properties of the window depend on the value of α (alpha). α controls
+// the fraction of the window which contains a cosine taper. α = 0.5 gives a
+// window whos central 50% is a flat top and outer quartiles are tapered.
```
romanwerpachowski, line 346: "whose"

dsp/window/window.go
```diff
@@ -330,6 +330,57 @@ func (g Gaussian) Transform(seq []float64) []float64 {
...
+		return Rectangular(seq)
+	} else if t.Alpha >= 1 {
+		return Hann(seq)
+	}
+
+	sep := 0.5 * t.Alpha * float64(len(seq)-1)
+	sepI := int(math.Floor(sep)) + 1
+	sep2 := float64((len(seq) - 1)) * (1 - .5*t.Alpha)
+	sep2I := int(math.Ceil(sep2))
+	for i := range seq {
```
romanwerpachowski, line 368: Some opportunities for optimisation here, to avoid `if` statements inside the `for` loop.

kortschak: Yes, see mine [here](https://github.com/gonum/gonum/pull/1437#discussion_r484200077).

dsp/window/window.go
brandondube: I like your new version more.  My objection to the previous one is the density of `:(len(seq)+1)/2` and treatment of the flat region as a special continue instead of unity weight.

brandondube:
I made the requested changes, I also touched the Gaussian doc for the same "by" vs "using" topic.

dsp/window/window.go
```diff
@@ -330,6 +330,54 @@ func (g Gaussian) Transform(seq []float64) []float64 {
 	return seq
 }
 
+// The values for specrtal leakage come from
```
romanwerpachowski, line 333: Typo.

dsp/window/window.go
```diff
@@ -330,6 +330,54 @@ func (g Gaussian) Transform(seq []float64) []float64 {
...
+// Transform applies the Tukey transformation to seq in place, using the value
+// of the receiver as the Alpha parameter, and returning the result
+func (t Tukey) Transform(seq []float64) []float64 {
+	if t.Alpha <= 0 {
+		return Rectangular(seq)
+	} else if t.Alpha >= 1 {
+		return Hann(seq)
+	}
+
+	alphaL := t.Alpha * float64(len(seq)-1)
```
romanwerpachowski, line 371: This could be cached in a local variable.

dsp/window/window.go
```diff
@@ -330,6 +330,54 @@ func (g Gaussian) Transform(seq []float64) []float64 {
...
+	width := int(0.5*alphaL) + 1
+	for i := range seq[:width] {
```
romanwerpachowski, line 373: Nitpicking: wouldn't calculating the required loop length beforehand be faster?

romanwerpachowski (changes requested):

dsp/window/window.go
kortschak: Do nothing is unity when the operation as multiply. That version is logically identical to the most recent one, though not as efficient.

dsp/window/window.go
kortschak: This is; `alphaL` is local.

dsp/window/window.go
kortschak: That's what this does.

dsp/window/window.go
kortschak: Yes, and please move the text below the table so users can see it as I have [here](https://github.com/gonum/gonum/pull/1437#discussion_r484182529). They may not be able to get the document, but they should see where it comes from.

dsp/window/window_test.go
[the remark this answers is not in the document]
brandondube: Are you objecting to `winType` instead of `windowType` ?  I find that needlessly pedantic since we are in the context of `window_test.`

I would prefer to not expend any further effort on the tests as they exist at this moment.  The relevant infrastructure will be remade for the change to the complex interface anyway.

dsp/window/window.go
brandondube: Which is also functionally equivalent to the original version I committed.  I am well aware that multiplying by unity doesn't do anything except (perhaps) accumulate a floating point error.

Never doing anything (i.e., no code considers those points) or explicitly using weight of 1 is clearer, without any indirection, vs a continue statement which is indirect.

dsp/window/window.go
brandondube: I had placed it outside the docstring intentionally but it's not a hill I want to die on.  Updated.

dsp/window/window_test.go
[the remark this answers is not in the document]
kortschak: OK. Yes, that what was I was asking for. If you are going to go through this an make additional changes, you can do it then. Otherwise I'll fix it.

FTR though, calling this pedantry is unnecessary; "pedantry" and "care" are two words for the same thing through the lens of different interests. My interest is the long term success of the project and ease of maintenance since that is where I spend much of my time. Conventionally in tests we are more explicit with naming, which is why my initial comment suggested the use of the longer label.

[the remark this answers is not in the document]
brandondube: I am going to rewrite and simplify the entire window test suite when I change the complex interface.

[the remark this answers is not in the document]
kortschak: No worries.

dsp/window/window_test.go
```diff
@@ -168,20 +195,21 @@ func TestWindows(t *testing.T) {
...
 				src[i] = 1
 			}
-
-			gaussian := Gaussian{test.sigma}
-
-			dst := gaussian.Transform(src)
+			type transformer interface {
+				Transform([]float64) []float64
+			}
+			trans := []transformer{Gaussian{test.param}, Tukey{test.param}}[test.winType]
```
kortschak, line 211: In future, please leave this over multiple lines to avoid diff noise when other mono-parametric windows are added..

dsp/window/window_test.go
```diff
@@ -190,7 +218,7 @@ func TestGausWindows(t *testing.T) {
 				src[i] = 1
 			}
 
-			dst = NewValues(gaussian.Transform, len(src)).Transform(src)
+			dst = NewValues(trans.Transform, len(src)).Transform(src)
 			if !floats.EqualApprox(dst, test.want, tol) {
 				t.Errorf("unexpected result for lookup window function %q:\ngot:%#.6v\nwant:%#.6v", test.name, dst, test.want)
```
kortschak, line 223: This should have a verb for the window type. Given that, maybe the look up above should be `map[string]transformer` and the test struct's window type field should be `string`.

dsp/window/window_test.go
```diff
@@ -230,29 +258,31 @@ func TestGausWindowComplex(t *testing.T) {
...
 
-	for _, test := range gausWindowTests {
-		t.Run(fmt.Sprintf("%sComplex (sigma=%.1f)", test.name, test.sigma), func(t *testing.T) {
-			src := make([]complex128, 20)
-			for i := range src {
-				src[i] = complex(1, 1)
-			}
+	for _, test := range monoParamWindowTests {
+		if test.winType == gaussWin {
+			t.Run(fmt.Sprintf("%sComplex (sigma=%.1f)", test.name, test.param), func(t *testing.T) {
```
kortschak, line 263: The test name handed to `t.Run` should reflect that this is a Gaussian test.

dsp/window/window_test.go
```diff
@@ -230,29 +258,31 @@ func TestGausWindowComplex(t *testing.T) {
...
 
-			gaussian := GaussianComplex{test.sigma}
+				gaussian := GaussianComplex{test.param}
 
-			dst := gaussian.Transform(src)
-			if !equalApprox(dst, test.want, tol) {
-				t.Errorf("unexpected result for window function %q:\ngot:%#.6v\nwant:%#.6v", test.name, dst, test.want)
-			}
+				dst := gaussian.Transform(src)
+				if !equalApprox(dst, test.want, tol) {
```
kortschak, line 272: `equalApprox` is no longer needed to defined locally since we have `cmplxs.EqualApprox` now.

kortschak (dismissed):
On the basis that the test code is being revised in a future PR I'm approving this.

Thanks for doing this.

Please file another PR adding you to CONTRIBUTORS and you or your employer as relevant to AUTHORS with the commit message "A+C: add Brandon Dube".

AUTHORS
```diff
@@ -12,6 +12,7 @@ Alexander Egurnov <alexander.egurnov@gmail.com>
 Andrei Blinnikov <goofinator@mail.ru>
 Bill Gray <wgray@gogray.com>
 Bill Noon <noon.bill@gmail.com>
+Brandon Dube <brandon.dube@jpl.nasa.gov> <brandon@retrorefractions.com>
```
kortschak, line 15: I'm going to squash this PR, so we will lose the commit message. Please revert this and [make a new PR making this change](https://github.com/gonum/gonum/pull/1437#pullrequestreview-483727703).

brandondube: ahead of you in parallel -- sorry, I'm too accustomed to single branch or only two branch development.

kortschak: No worries. Yes, GitHub's PR mechanism is not the best.

kortschak (approved):

dsp/window/window.go
romanwerpachowski: I see, thanks.

dsp/window/window.go
romanwerpachowski: I meant `len(seq)`.

kortschak: Ah. No, it's not worth it; `len(v)` is as fast as accessing a local.

romanwerpachowski: I see... Thanks.

romanwerpachowski (approved):

PR #1440, opened 2020-09-10
dsp/window: refactor for GaussianComplex, revised test suite

brandondube:
As discussed on gonum-dev, this PR removes the GaussianComplex type, moving its Transform method
to Gaussian.TransformComplex.  It also implements TransformComplex for Tukey.

The test suite was re-written for improved clarity by disentangling the testing of real and complex windows.  Unfortunately, the same was not true of the parametric windows.  The asymmetry bothers me.

The tests rely on subtle Go behavior, which I added a comment for in the test file.  `var a complex128 = 1` has value `1+0i`.  It isn't insensible for the language to have that behavior, it just surprised me that untyped constants are valid as complex values.

I removed the test bodies that make a `Values` out of each window.  I do not see the utility in doing it twice -- the test already checks weights against a unity input.  It feels like the previous way mixed testing of `Values` and other windows which, to me, is not the way to go.

The new complex tests rely somewhat on `(x + yi)(u + vi) = (xu-yv) + (xv + yu)i`.  It has `y=v=0` which zeros three of the terms.  It is modest busywork to change that, the process for which means updating the truth tables.

The test file defines five interfaces or types, which is the essential value to this PR:
```go

type transformFunc func([]float64) []float64
type transformFuncComplex func([]complex128) []complex128

type transformer interface {
	Transform([]float64) []float64
}

type transformerComplex interface {
	TransformComplex([]complex128) []complex128
}

type dualTransformer interface {
	transformer
	transformerComplex
}
```

The previous `Transform([]complex128) []complex128` interface on `GaussianComplex` overlapped with the `Transform` method on `Gaussian`.  This made the `dualTransformer` interface in the test file impossible.  The composability is not so important as the improved clarity and homogeneity with the non parametric windows, to me.

### DIFF
--- a/dsp/window/window_test.go
+++ b/dsp/window/window_test.go
@@ -12,12 +12,8 @@ import (
 	"gonum.org/v1/gonum/floats/scalar"
 )
 
-// parametricWindow is used in a switch when testing monoparametric windows
-// to change which struct is created
-type parametricWindow int
-
 const (
-	gaussWin parametricWindow = iota
+	gaussWin = iota
 	tukeyWin
 )
 
@@ -123,47 +119,47 @@ var windowTests = []struct {
 }
 
 var monoParamWindowTests = []struct {
-	name  string
-	param float64
-	win   parametricWindow
-	want  []float64
+	name    string
+	param   float64
+	winType int
+	want    []float64
 }{
 	{
-		name: "Gaussian", param: 0.3, win: gaussWin,
+		name: "Gaussian", param: 0.3, winType: gaussWin,
 		want: []float64{
 			0.006645, 0.018063, 0.043936, 0.095634, 0.186270, 0.324652, 0.506336, 0.706648, 0.882497, 0.986207,
 			0.986207, 0.882497, 0.706648, 0.506336, 0.324652, 0.186270, 0.095634, 0.043936, 0.018063, 0.006645},
 	},
 	{
-		name: "Gaussian", param: 0.5, win: gaussWin,
+		name: "Gaussian", param: 0.5, winType: gaussWin,
 		want: []float64{
 			0.164474, 0.235746, 0.324652, 0.429557, 0.546074, 0.666977, 0.782705, 0.882497, 0.955997, 0.995012,
 			0.995012, 0.955997, 0.882497, 0.782705, 0.666977, 0.546074, 0.429557, 0.324652, 0.235746, 0.164474,
 		},
 	},
 	{
-		name: "Gaussian", param: 1.2, win: gaussWin,
+		name: "Gaussian", param: 1.2, winType: gaussWin,
 		want: []float64{
 			0.730981, 0.778125, 0.822578, 0.863552, 0.900293, 0.932102, 0.958357, 0.978532, 0.992218, 0.999132,
 			0.999132, 0.992218, 0.978532, 0.958357, 0.932102, 0.900293, 0.863552, 0.822578, 0.778125, 0.730981,
 		},
 	},
 	{
-		name: "Tukey", param: 1, win: tukeyWin,
+		name: "Tukey", param: 1, winType: tukeyWin,
 		want: []float64{ // copied from Hann
 			0.006155, 0.054496, 0.146447, 0.273005, 0.421783, 0.578217, 0.726995, 0.853553, 0.945503, 0.993844,
 			0.993844, 0.945503, 0.853553, 0.726995, 0.578217, 0.421783, 0.273005, 0.146447, 0.054496, 0.006155,
 		},
 	},
 	{
-		name: "Tukey", param: 0, win: tukeyWin,
+		name: "Tukey", param: 0, winType: tukeyWin,
 		want: []float64{ // copied from rectangular
 			1, 1, 1, 1, 1, 1, 1, 1, 1, 1,
 			1, 1, 1, 1, 1, 1, 1, 1, 1, 1,
 		},
 	},
 	{
-		name: "Tukey", param: 0.5, win: tukeyWin,
+		name: "Tukey", param: 0.5, winType: tukeyWin,
 		want: []float64{
 			0.000000, 0.105430, 0.377257, 0.700847, 0.939737, 1.000000, 1.000000, 1.000000, 1.000000, 1.000000,
 			1.000000, 1.000000, 1.000000, 1.000000, 1.000000, 0.939737, 0.700847, 0.377257, 0.105429, 0.000000,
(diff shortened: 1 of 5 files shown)


brandondube:
It looks like my choice of branching commit lead to a big overlap with the other PR and a mess in the commits :( 

Only 53373c14559a4d268ae33ecde7844f9c943da359 is relevant

kortschak:
Yeah, branch management is important. I can look at that commit, but this won't merge as it is due to the conflicts. Can you please clean that up?.

dsp/window/window_test.go
```diff
@@ -8,170 +8,317 @@ import (
...
+// these types describe zero, mono, and multi parameter windowing
+// functions for both real and complex data
+//
+// when looking at the truth tables, it is helpful to know that
+// Go treats []complex128{1} as [1+0i].  More specifically, a variable which
+// is of type complex128 to which you assign an untyped constant has a zero
+// imaginary part.
+
+type transformFunc func([]float64) []float64
+type transformFuncComplex func([]complex128) []complex128
```
kortschak, line 30: Please delete these and use the untyped function declarations in the structs.

dsp/window/window_test.go
```diff
@@ -8,170 +8,317 @@ import (
...
-	fn      func([]float64) []float64
-	fnCmplx func([]complex128) []complex128
-	want    []float64
+// these types describe zero, mono, and multi parameter windowing
+// functions for both real and complex data
+//
+// when looking at the truth tables, it is helpful to know that
+// Go treats []complex128{1} as [1+0i].  More specifically, a variable which
+// is of type complex128 to which you assign an untyped constant has a zero
+// imaginary part.
```
kortschak, line 27: I don't think this is necessary. The interface definitions say the first part and the second part is described in the language spec.

dsp/window/window_test.go
```diff
@@ -8,170 +8,317 @@ import (
...
+}
+
+type transformerComplex interface {
+	TransformComplex([]complex128) []complex128
+}
+
+type dualTransformer interface {
+	transformer
+	transformerComplex
+}
```
kortschak, line 43: The first two types are only used to define the last one and it's only used once, so I would delete these all and define the dual case where it's used.

dsp/window/window_test.go
```diff
@@ -182,121 +329,81 @@ func TestWindows(t *testing.T) {
...
-				}
-
-				for i := range src {
-					src[i] = complex(1, 1)
-				}
+		t.Run(fmt.Sprintf("%s (param=%.1f)", test.name, test.param), func(t *testing.T) {
+			trans := []dualTransformer{
+				Gaussian{test.param},
+				Tukey{test.param},
+			}[test.windowType]
```
kortschak, line 388:
```
			type transformer interface {
				Transform([]float64) []float64
				TransformComplex([]complex128) []complex128
			}
			trans := []transformer{
				Gaussian{test.param},
				Tukey{test.param},
			}[test.windowType]
```

brandondube:
> Yeah, branch management is important. I can look at that commit, but this won't merge as it is due to the conflicts. Can you please clean that up?.

How?  I chose this branch point specifically because if the previous PR and this fanned out, there would be a conflict.  The extra commits here -- I think -- will evaporate once gonum/gonum:master contains them.

dsp/window/window_test.go
brandondube: I agree that the locality of define-at-place-of-use is better.  It's been said that for the gonum projects, often the tests serve as examples.  It is also pragmatic Go to define minimal interfaces.  I think it is better to show them as separate, at the top of the file before the reader's eyes glaze over scrolling past the truth tables.  There is no requirement that a type implement one or the other.  Especially, Values implements one and ValuesComplex the other.

dsp/window/window_test.go
brandondube: Why?

dsp/window/window_test.go
brandondube: The second part is not part of [the spec](https://golang.org/ref/spec#Complex_numbers), at least not explicitly.  I also think that the most useful comments are the ones that explain something subtle about code, and this is a subtle behavior that you all but certainly will not _unambiguously_ be able to predict without seeing it before.

kortschak:
The previous PR *has* been merged. There *are* conflcts with that merged PR. I could resolve those with the GitHub UI, but I won't because it is a terrible experience and you better understand the changes where the conflicts exist.

To clean up the conflicts, either rebase this branch onto master and resolve the conflicts that arise or make a temporary duplicate of this branch reset this branch hard to master and then cherrypick the relevant commit onto that.

dsp/window/window_test.go
kortschak: Locality. Name indirection doesn't help here, it just adds words.

dsp/window/window_test.go
kortschak: The [spec explicitly shows that a `1` can be converted to a `comple128`](https://golang.org/ref/spec#Conversions); that's what is happening here.

dsp/window/window_test.go
kortschak: The tests serve maybe three purposes, testing the code, showing use of the code and demonstrating idiomatic Gonum usage. Making a single-use interface distant to its use is not idiomatic use.

brandondube:
Tone does not carry well in purely textual communication; is my read of hostility in your words as you intend?

kortschak:
No, not intended at all, sorry if it came across that way; the emphatics were to counterpoint the claims that the merging of the previous PR would resolve the merge conflict. The "terrible experience" is attempting to resolve conflicts with the GitHub UI and I don't think in general that resolving conflicts should be done by third parties if avoidable given that it depends on the intention of the merging author.

I can understand the reason for you having chosen to do what you've done with the branches, but I think the squash merge brought that undone.

kortschak:
I've had a look through the relevant commit and I'm not convince that the testing changes are an improvement overall. We lose testing of `NewValues*` and, significantly, that using a `Values*` gives identical results to using the window functions directly (this is a design goal of the package since using a precomputed window is a significant performance benefit in tight loops), we also now have two sources of truth for the window functions (one for real and one for complex windows) that are readily interchangeable (I know I said [this](https://github.com/gonum/gonum/pull/1437#discussion_r484588480) - I was wrong).

My preferred option would be just to do the work on the `Gaussian` and `Tukey` types placing them in a new file, "window_parametric.go" and then clean up the changes that were deferred in #1437. I would also roll the parametric tests into the others using received-curried method functions, e.g.
```
	{
		name: "Gaussian_1.2", fn: Gaussian{1.2}.Transform, fnCmplx: Gaussian{1.2}.TransformComplex,
		want: []float64{
			0.730981, 0.778125, 0.822578, 0.863552, 0.900293, 0.932102, 0.958357, 0.978532, 0.992218, 0.999132,
			0.999132, 0.992218, 0.978532, 0.958357, 0.932102, 0.900293, 0.863552, 0.822578, 0.778125, 0.730981,
		},
	},
```

In spending some time thinking about the package again, there is potential room for improving the complex window functions' performance by replacing direct complex multiplies with explicit scalar multiplication, e.g.
```
func SineComplex(seq []complex128) []complex128 {
	k := math.Pi / float64(len(seq))
	for i, v := range seq {
		f := math.Sin(k*(float64(i)+0.5))
		seq[i] = complex(f*real(v), f*imag(v))
	}
	return seq
}
```
This should be a separate PR from the work above and should have benchmarking to support it (though it has correctness benefits in rare cases; when the sequence element is Inf which probably justifies it without benchmarking).

brandondube:
withdrawn

PR #1441, opened 2020-09-11
mat: add BandCholesky type

vladimir-ch:
This is an initial draft for a BandCholesky type. At the moment I've added only the minimal functionality to solve banded systems and to satisfy the Matrix interface. I don't plan to support updates to the matrix as `Cholesky` does. WIP, so there are no tests and no documentation comments.

### DIFF
--- a/mat/cholesky.go
+++ b/mat/cholesky.go
@@ -20,6 +20,11 @@ const (
 var (
 	_ Matrix    = (*Cholesky)(nil)
 	_ Symmetric = (*Cholesky)(nil)
+
+	_ Matrix    = (*BandCholesky)(nil)
+	_ Symmetric = (*BandCholesky)(nil)
+	_ Banded    = (*BandCholesky)(nil)
+	_ SymBanded = (*BandCholesky)(nil)
 )
 
 // Cholesky is a symmetric positive definite matrix represented by its
@@ -100,7 +105,7 @@ func (c *Cholesky) At(i, j int) float64 {
 	return val
 }
 
-// T returns the the receiver, the transpose of a symmetric matrix.
+// T returns the receiver, the transpose of a symmetric matrix.
 func (c *Cholesky) T() Matrix {
 	return c
 }
@@ -264,7 +269,7 @@ func (a *Cholesky) SolveCholTo(dst *Dense, b *Cholesky) error {
 	return nil
 }
 
-// SolveVecTo finds the vector X that solves A * x = b where A is represented
+// SolveVecTo finds the vector x that solves A * x = b where A is represented
 // by the Cholesky decomposition. The result is stored in-place into
 // dst.
 func (c *Cholesky) SolveVecTo(dst *VecDense, b Vector) error {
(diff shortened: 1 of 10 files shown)


mat/triband.go
```diff
@@ -260,6 +260,94 @@ func (t *TriBandDense) Reset() {
 	t.mat.Data = t.mat.Data[:0]
 }
 
+// ReuseAsTriBand changes the receiver to be of size n×n, bandwidth k+1 and of
+// the given kind, re-using the backing data slice if it has sufficient capacity
+// and allocating a new slice otherwise. The backing data is zero on return.
+//
+// The receiver must be empty, n must be positive and k must be non-negative and
+// less than n, otherwise ReuseAsTriBand will panic. To empty the receiver for
+// re-use, Reset should be used.
```
kortschak, line 269: There should probably be a warning that changing between U and L will cause a panic if the receiver is not empty.

kortschak (commented):
I've looked through all but the actual `BandCholesky` addition.

mat/cholesky.go
```diff
@@ -697,3 +701,132 @@ func (c *Cholesky) SymRankOne(orig *Cholesky, alpha float64, x Vector) (ok bool)
...
+		Stride: c.chol.RawTriBand().Stride,
+	}
+	_, ok = lapack64.Pbtrf(cSym)
+	if !ok {
+		c.Reset()
+		return false
+	}
+	work := getFloats(3*n, false)
+	iwork := getInts(n, false)
+	aNorm := glapack.Implementation{}.Dlansb(CondNorm, cSym.Uplo, cSym.N, cSym.K, cSym.Data, cSym.Stride, work)
```
kortschak, line 737: Would it be worth adding `Lansb` (and `Lantb`) to lapack64, but not having them use the `lapack64.lapack64` value? Then `Lansb` would look like
```
// Lansb computes the specified norm of an n×n symmetric band matrix. If
// norm == lapack.MaxColumnSum or norm == lapack.MaxRowSum, work must have length
// at least n and this function will panic otherwise.
// There are no restrictions on work for the other matrix norms.
//
// Dlansb is not part of the lapack.Float64 interface and so calls to Lansb are always
// executed by the Gonum implementation.
func Lansb(norm lapack.MatrixNorm, a blas64.SymmetricBand, work []float64) float64 {
	return gonum.Implementation{}.Dlansb(norm, a.Uplo, a.N, a.K, a.Data, max(1, a.Stride), work)
}
```
This way we're not faking it, but we do have the nicer API.

mat/triband.go
vladimir-ch: The receiver must be empty and then changing between U/L won't cause a panic.

mat/cholesky.go
vladimir-ch: Good idea, done.

mat/triband.go
kortschak: Yes. Sorry.

vladimir-ch:
I've added the missing essential methods, documentation and tests, and removed the "Draft" status. This can be reviewed.

mat/triband.go
```diff
@@ -369,3 +457,50 @@ func (t *TriBandDense) Trace() float64 {
...
+			} else {
+				// dst is upper triangular, s is in lower triangle.
+				// Commented out due to lack of test coverage.
+				// for i := 0; i < n; i++ {
+				// 	ilen := min(k+1, n-i)
+				// 	for j := 0; j < ilen; j++ {
+				// 		t.Data[i*t.Stride+j] = s.Data[(i+j)*s.Stride+k-j]
+				// 	}
+				// }
+				panic("not implemented")
```
vladimir-ch, line 489: These panics certainly don't look nice but is it really worth the trouble of implementing them and covering with tests? `dst` will most likely always be lower triangular, that eliminates two panics. `s` comes from the outside but how likely is that something that looks like a `SymBandDense` will be stored in the lower triangle?

mat/cholesky.go
```diff
@@ -696,3 +702,214 @@ func (c *Cholesky) SymRankOne(orig *Cholesky, alpha float64, x Vector) (ok bool)
...
+	iwork := getInts(n, false)
+	aNorm := lapack64.Lansb(CondNorm, cSym, work)
+	ch.cond = 1 / lapack64.Pbcon(cSym, aNorm, work, iwork)
+	putInts(iwork)
+	putFloats(work)
+	return true
+}
+
+// SolveTo finds the matrix X that solves A * X = B where A is represented by
+// the Cholesky decomposition. The result is stored in-place into dst.
```
kortschak, line 759: We don't do this elsewhere, though we probably should wherever it's returned; we should say that the error type returned is a `Condition` error when non-nil. This should be a separate PR dealing with all places where this happens.

mat/cholesky_test.go
```diff
@@ -707,3 +708,219 @@ func BenchmarkCholeskyInverseTo(b *testing.B) {
...
+	rnd := rand.New(rand.NewSource(1))
+	for _, n := range []int{1, 2, 3, 5, 10} {
+		for _, k := range []int{0, 1, n / 2, n - 1} {
+			k := min(k, n-1)
+
+			a := NewSymBandDense(n, k, nil)
+			for i := 0; i < n; i++ {
+				a.SetSymBand(i, i, rnd.Float64()+float64(n))
+				for j := i + 1; j < min(i+k+1, n); j++ {
+					a.SetSymBand(i, j, rnd.Float64())
```
kortschak, line 728: Do we want to test with internal/testrand?

mat/triband.go
kortschak: Do you want to add a TODO that can be done when someone hits this? (and below)

mat/cholesky.go
vladimir-ch: I've created #1470  for this.

mat/cholesky_test.go
vladimir-ch: I've thought about it but in my view testrand is most useful for testing numerical properties/robustness which I think should be done in the lapack package. Testing in mat should focus on the added value/functionality of mat which is the handling of variety of matrix types. Do you agree?

I've never used testrand but it looks like it could be used more in testlapack to ease (after some extending) the generation of random test matrices.

mat/triband.go
vladimir-ch: Do you mean adding a TODO comment, extending the panic string or creating a ticket?

kortschak: I was thinking just to add a TODO comment.

mat/cholesky_test.go
kortschak: Yes, I agree.

mat/triband.go
vladimir-ch: Done.

vladimir-ch:
PTAL

kortschak (approved):

vladimir-ch:
Is it ok if I ignore the linting failure
```
func `(*TriBandDense).reuseAsNonZeroed` is unused (unused)
```
or should I remove the method?

kortschak:
It's fine to leave it if you think we'll be using it soonish.

vladimir-ch:
I'm not sure if soonish but I'll take a look at what would be the most useful method calling it that we could add.

PR #1443, opened 2020-09-12
dsp/window: fix window sizing

### LINKED ISSUE #1442: dsp/window: window widths of Tukey v other functions do not match
It's been gently pointed out to me that #1242 was the wrong thing to do; the way we decided to make the windows in #1240 was to only keep the non-zero weights based on the fact that the region outside the window would be zeroed.

This is not the behaviour of the new Tukey implementation (which can be seen by removing the Hann call optimisation). If the windows' behaviours are known to only contain the non-zero weights then the user can zero the preceding and following values. However, this is less ergnomic than it could be, and the behaviour differs from the Rectangular window where this should not be done, leaving it not possibly to have drop-in replacement between these different classes of window.

On this basis, I'm going to revert #1242. The alternative is to bring Tukey into the same scheme.

kortschak:
Please take a look.

~~This takes ideas suggested by @brandondube in #1440 and extends them.~~

Fixes #1442.


dsp/window/window.go
```diff
@@ -92,14 +92,14 @@ func Triangular(seq []float64) []float64 {
...
+//  w[k] = 0.5*(1 - cos(2*π*k/(N-1))),
 // for k=0,1,...,N-1 where N is the length of the window.
 //
 // Spectral leakage parameters: ΔF_0 = 4, ΔF_0.5 = 1.5, K = 2, ɣ_max = -31.5, β = -6.
 func Hann(seq []float64) []float64 {
-	k := 2 * math.Pi / float64(len(seq))
+	k := 2 * math.Pi / float64(len(seq)-1)
 	for i := range seq {
-		seq[i] *= 0.5 * (1 - math.Cos(k*(float64(i)+0.5)))
+		seq[i] *= 0.5 * (1 - math.Cos(k*float64(i)))
```
romanwerpachowski, line 102: I'm slightly worried that, numerically, `(2 * math.Pi / float64(len(seq) - 1)) * float64(i) != 2 * math.Pi` for `i == len(seq) - 1`, and we lose the symmetry at the edges. A slightly more expensive but maybe more accurate way of computing this would be `2 * math.Pi * (float64(i) / float64(len(seq) - 1))`. WTDYT?

kortschak: Do you have an estimate of the impact or examples where it presents itself? Ideally, this commit would be a pure reversion (it's not because other code has changed, but it's fairly close), so I'd be happy to look at it, but probably after this has settled and also since Vladimir wants to look into it more before the review goes much further (the non-reversion parts of this PR are worth looking at though).

romanwerpachowski: I did a quick test
```
package main

import (
	"fmt"
	"math"
)

func main() {
	n := 75
	k := 2 * math.Pi / float64(n - 1)
	a := math.Cos(k * 0)
	b := math.Cos(k * float64(n - 1))
	fmt.Printf("a - b == %g", a - b)
}
```
and it doesn't seem to be a problem. Sorry for jumping the gun.

kortschak:
I'm going to reduce this PR to purely the the reversion of the window offsets and deal with the other parts later.

vladimir-ch (dismissed):

kortschak:
I need to resolve conflicts due to removal of Tukey. I missed those.

vladimir-ch (approved):

PR #1445, opened 2020-09-12
graph/encoding/{digraph6,graph6}: allow packages to build on 32-bit arches

### LINKED ISSUE #1444: gonum fails to build when ints are 32 bits
Running tests with GOARCH=386 results in:

```
# gonum.org/v1/gonum/graph/encoding/digraph6 [gonum.org/v1/gonum/graph/encoding/digraph6.test]
./digraph6.go:65:9: constant 68719476736 overflows int
FAIL	gonum.org/v1/gonum/graph/encoding/digraph6 [build failed]
```

And same problem in graph6.

Tested with 0.8.1 and HEAD (1a536f9a771e1c098630abc9a8bf53b7e5e1994f) with go1.15 on linux

Yes

This simple patch fixes it but doesn't seem very elegant

```
diff --git a/graph/encoding/digraph6/digraph6.go b/graph/encoding/digraph6/digraph6.go
index 5b53136f..f172e41e 100644
--- a/graph/encoding/digraph6/digraph6.go
+++ b/graph/encoding/digraph6/digraph6.go
@@ -62,7 +62,7 @@ func Encode(g graph.Graph) Graph {
 		buf.WriteByte(byte(n) + 63)
 	case n < 258048:
 		buf.Write([]byte{126, byte(n>>12) + 63, byte(n>>6) + 63, byte(n) + 63})
-	case n < 68719476736:
+	case int64(n) < 68719476736:
 		buf.Write([]byte{126, 126, byte(n>>30) + 63, byte(n>>24) + 63, byte(n>>18) + 63, byte(n>>12) + 63, byte(n>>6) + 63, byte(n) + 63})
 	default:
 		panic("digraph6: too large")
diff --git a/graph/encoding/graph6/graph6.go b/graph/encoding/graph6/graph6.go
index 90c84117..18153a83 100644
--- a/graph/encoding/graph6/graph6.go
+++ b/graph/encoding/graph6/graph6.go
@@ -63,7 +63,7 @@ func Encode(g graph.Graph) Graph {
 		buf.WriteByte(byte(n) + 63)
 	case n < 258048:
 		buf.Write([]byte{126, byte(n>>12) + 63, byte(n>>6) + 63, byte(n) + 63})
-	case n < 68719476736:
+	case int64(n) < 68719476736:
 		buf.Write([]byte{126, 126, byte(n>>30) + 63, byte(n>>24) + 63, byte(n>>18) + 63, byte(n>>12) + 63, byte(n>>6) + 63, byte(n) + 63})
 	default:
 		panic("graph6: too large")
```

kortschak:
Please take a look.

Fixes #1444.

/cc @pterjan

### DIFF
--- a/graph/encoding/digraph6/digraph6.go
+++ b/graph/encoding/digraph6/digraph6.go
@@ -57,7 +57,12 @@ func Encode(g graph.Graph) Graph {
 
 	var buf strings.Builder
 	buf.WriteByte('&')
-	switch {
+	// digraph6 specifies graphs of order up to 68719476735
+	// which overflows int on 32 bit architectures. We know
+	// that on those machines n will not be this large, since
+	// it came from a length, but explicitly convert to 64
+	// bits to allow the package to build on those architectures.
+	switch n := int64(n); {
 	case n < 63:
 		buf.WriteByte(byte(n) + 63)
 	case n < 258048:
--- a/graph/encoding/digraph6/digraph6.go
+++ b/graph/encoding/digraph6/digraph6.go
@@ -57,11 +57,11 @@ func Encode(g graph.Graph) Graph {
 
 	var buf strings.Builder
 	buf.WriteByte('&')
-	// digraph6 specifies graphs of order up to 68719476735
-	// which overflows int on 32 bit architectures. We know
-	// that on those machines n will not be this large, since
-	// it came from a length, but explicitly convert to 64
-	// bits to allow the package to build on those architectures.
+	// digraph6 specifies graphs of order up to 2^36-1 which
+	// overflows int on 32-bit architectures. We know that on
+	// those machines n will not be this large, since it came
+	// from a length, but explicitly convert to 64 bits to
+	// allow the package to build on those architectures.
 	switch n := int64(n); {
 	case n < 63:
 		buf.WriteByte(byte(n) + 63)
--- a/graph/encoding/graph6/graph6.go
+++ b/graph/encoding/graph6/graph6.go
@@ -58,7 +58,12 @@ func Encode(g graph.Graph) Graph {
 	}
 
 	var buf strings.Builder
-	switch {
+	// graph6 specifies graphs of order up to 68719476735
+	// which overflows int on 32 bit architectures. We know
+	// that on those machines n will not be this large, since
+	// it came from a length, but explicitly convert to 64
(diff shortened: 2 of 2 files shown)


graph/encoding/graph6/graph6.go
```diff
@@ -58,7 +58,12 @@ func Encode(g graph.Graph) Graph {
 	}
 
 	var buf strings.Builder
-	switch {
+	// graph6 specifies graphs of order up to 68719476735
```
romanwerpachowski, line 61: Can this number be expressed using some power of 2, to make it easier to understand?

graph/encoding/digraph6/digraph6.go
```diff
@@ -57,7 +57,12 @@ func Encode(g graph.Graph) Graph {
 
 	var buf strings.Builder
 	buf.WriteByte('&')
-	switch {
+	// digraph6 specifies graphs of order up to 68719476735
```
romanwerpachowski, line 60: Can this number be expressed using some power of 2, to make it easier to understand?

graph/encoding/graph6/graph6.go
kortschak: Normally I would say yes, but this comes from an encoding specification, where this is how it's written.

http://users.cecs.anu.edu.au/~bdm/data/formats.html

romanwerpachowski: I found it here: https://users.cecs.anu.edu.au/~bdm/data/formats.txt

> Let n be an integer in the range 0-68719476735 (2^36-1)

Why not add "(2^36-1)" after the long number?

kortschak: Oh, nice! Thanks, I missed that.

kortschak: Just for the comment or would you like the constants changed too?

romanwerpachowski: I personally think it's sufficient to change the comment.

kortschak: Done. PTAL

graph/encoding/digraph6/digraph6.go
kortschak: Done

romanwerpachowski (approved):

PR #1446, opened 2020-09-12
dsp/window: revert tukey

brandondube:
This removes materials which are or may reasonably be believed to be the intellectual property of @brandondube.


romanwerpachowski:
Hi Brandon,

If possible, could you add some context to this PR? If not here than on gonum-dev.

It's OK if you don't want to do this, in which case I'll approve the PR as it is.

romanwerpachowski (approved):

romanwerpachowski:
Approved. Note that the reverted changes include also reviewer's suggestions.

PR #1449, opened 2020-09-13
mat: don't clobber optimisation for TriDense matrix input

### LINKED ISSUE #1448: mat: missing return for VecDense.MulVec in TriDenseCase
The [VecDense.MulVec](https://github.com/gonum/gonum/blob/f883758f082bd1f138ab0155a5c2677edd09b320/mat/vector.go#L556) function has the [TriDense case](https://github.com/gonum/gonum/blob/f883758f082bd1f138ab0155a5c2677edd09b320/mat/vector.go#L631):
```go
switch aU := aU.(type) {
...
case *TriDense:
	v.CopyVec(b)
	aU.checkOverlap(v.asGeneral())
	ta := blas.NoTrans
	if trans {
		ta = blas.Trans
	}
	blas64.Trmv(ta, aU.mat, v.mat)
...
}
```
that lacks a return value, meaning that the code then proceeds on to the inefficient backup code at the bottom of the function. This causes slowdown in cases where MulVec is called with a TriDense matrix (for example, distmv.StudentsT.Rand).

kortschak:
Please take a look.

Fixes #1448.

### DIFF
--- a/mat/vector.go
+++ b/mat/vector.go
@@ -629,13 +629,16 @@ func (v *VecDense) MulVec(a Matrix, b Vector) {
 			return
 		}
 	case *TriDense:
-		v.CopyVec(b)
-		aU.checkOverlap(v.asGeneral())
-		ta := blas.NoTrans
-		if trans {
-			ta = blas.Trans
+		if fast {
+			v.CopyVec(b)
+			aU.checkOverlap(v.asGeneral())
+			ta := blas.NoTrans
+			if trans {
+				ta = blas.Trans
+			}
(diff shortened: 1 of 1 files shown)


romanwerpachowski (approved):
LGTM. Did you see an improvement in benchmarks?

kortschak:
I didn't look. I will now.

romanwerpachowski:
> ~I didn't look. I will now.~ There are no benchmarks that cover this AFAICS.

Would be nice to have some :)

PR #1450, opened 2020-09-14
graph/layout: allow Isomap tests to pass on GOARCH=386

kortschak:
Please take a look.

Updates #1444.


vladimir-ch:
Those strange glyphs in `graph/layout/testdata/wp_page_isomap_386_golden.png` are expected?

kortschak:
The isomap layout doesn't prevent nodes from overlapping, and nodes that have the same edge relationships as other nodes will be given the same coordinates. So those strange glyphs are just two node IDs written on top of each other. Isomap is not a good layout for integral edge weights for this reason.

vladimir-ch:
Thanks.

vladimir-ch (approved):

PR #1455, opened 2020-09-15
dsp/window: add Tukey, clean up tests, simplify API and fix scaling infinities by zero

kortschak:
Please take a look.


dsp/window/window_complex.go
```diff
@@ -79,8 +81,9 @@ func LanczosComplex(seq []complex128) []complex128 {
 // Spectral leakage parameters: ΔF_0 = 4, ΔF_0.5 = 1.33, K = 2, ɣ_max = -26.5, β = -6.
 func TriangularComplex(seq []complex128) []complex128 {
 	a := float64(len(seq)-1) / 2
-	for i := range seq {
-		seq[i] *= complex(1-math.Abs(float64(i)/a-1), 0)
+	for i, v := range seq {
+		w := 1 - math.Abs(float64(i)/a-1)
```
romanwerpachowski, line 85: The window is symmetric, isn't there some optimisation opportunity? we could avoid calling math.Abs.

dsp/window/window_parametric.go
```diff
@@ -0,0 +1,165 @@
...
+	if len(v) != len(seq) {
+		panic("window: length mismatch")
+	}
+	for i, w := range v {
+		seq[i] *= w
+	}
+	return seq
+}
+
+// Transform applies the weights in the receiver to seq in place, returning the
```
vladimir-ch, line 150: TransformComplex

dsp/window/window_complex.go
kortschak: Yeah. They're all symmetric. Is it OK if we get to that later (we already have a perfect optimisation by using the `Values` type)?

romanwerpachowski: Sure.

dsp/window/window_parametric.go
kortschak: Done

dsp/window/window_parametric.go
```diff
@@ -147,31 +147,19 @@ func (v Values) Transform(seq []float64) []float64 {
...
-	for i := range v {
-		v[i] = 1
-	}
-	return window(v)
-}
-
-// Transform applies the weights in the receiver to seq in place, returning the
-// result. If v is nil, Transform is a no-op, otherwise the length of v must
+// TransformComplex applies the weights in the receiver to seq in place, returning
+// the result. If v is nil, Transform is a no-op, otherwise the length of v must
```
romanwerpachowski, line 151: I don't understand this change, TBH. weighting floats with complex numbers is not a replacement for weighting complex numbers with complex numbers?

dsp/window/window_parametric.go
```diff
@@ -0,0 +1,165 @@
...
+		panic("window: length mismatch")
+	}
+	for i, w := range v {
+		seq[i] *= w
+	}
+	return seq
+}
+
+// TransformComplex applies the weights in the receiver to seq in place, returning
+// the result. If v is nil, Transform is a no-op, otherwise the length of v must
```
vladimir-ch, line 151: TransformComplex also here. Sorry I didn't see it before.

dsp/window/window_parametric.go
kortschak: You're absolutely right; the weighting (I'm sort of reluctant to call it weighting) complex numbers with complex numbers should not have been here. The point of the package is to provide scalar weighting of inputs for Fourier transforms. The approach that was used before was a lazy way to apply scalar weightings (my blame).

dsp/window/window_parametric.go
kortschak: ... or I could use some initiative.

dsp/window/window_parametric.go
romanwerpachowski: I get it now. Thanks.

romanwerpachowski (approved):

vladimir-ch (approved):

PR #1456, opened 2020-09-15
stat/distuv: Improve Gamma.Rand performance in alpha >1 case

### LINKED ISSUE #1453: stat/distuv: Improve performance of Gamma.Rand in Alpha > 1 case
In the current Gamma.Rand code, when Alpha > 1, it uses [the following loop](https://github.com/gonum/gonum/blob/1839007961357eb896846276604907b28f214187/stat/distuv/gamma.go#L222):

```go
u := -exprnd()
x := normrnd()
v := 1 + x*c
v = v * v * v
if u < 0.5*x*x+d*(1-v+math.Log(v)) {
	return d * v / b
}
```

Replacing the above with code based on [numpy's logic](https://github.com/numpy/numpy/blob/9ee262b5bf89e8c866a507e4d62b78532361adc2/numpy/random/src/legacy/legacy-distributions.c#L67):
```go
x := normrnd()
v := 1 + x*c
if v <= 0.0 {
	continue
}
v = v * v * v
u := unifrnd()
if u < 1.0-0.0331*(x*x)*(x*x) {
	return d * v / b
}
if math.Log(u) < 0.5*x*x+d*(1-v+math.Log(v)) {
	return d * v / b
}
```

Runs much faster - as the new initial check (`if u < ...`) catches most (specifically, ~92% of) cases and avoids an expensive call to `math.Log`. For a simple benchmark:
```go
package main

import (
	"testing"

	"gonum.org/v1/gonum/stat/distuv"
)

func BenchmarkGamma(b *testing.B) {
	dist := distuv.Gamma{Alpha: 2.0, Beta: 1.0}
	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		dist.Rand()
	}
}
```

`go test -bench=. -benchtime=10s -cpu=1`
Old code: `BenchmarkGamma  207709526               57.0 ns/op`
New code: `BenchmarkGamma  345146806               34.6 ns/op`
Gives a 40% performance boost (windows/amd64). This should also improve performance for some other distributions (Beta/Dirichlet, StudentsT, etc).

A few existing statistical tests (specifically, InverseGamma - ExKurtosis, Beta - Entroy, and StudentsT - Variance) fail with the new code until you sufficiently increase the `n` sample sizes, indicating it's just sampling error.

argusdusty:
Please take a look.

Fixes #1453.

benchstat (from the new GammaRand benchmark):
```
name              old time/op  new time/op  delta
GammaRand/case_0  90.3ns ± 5%  91.1ns ± 3%     ~     (p=0.152 n=10+8)
GammaRand/case_1   215ns ± 4%   213ns ± 3%     ~     (p=0.238 n=10+10)
GammaRand/case_2  25.5ns ± 3%  25.7ns ± 3%     ~     (p=0.534 n=9+10)
GammaRand/case_3  72.4ns ± 4%  48.5ns ± 5%  -32.97%  (p=0.000 n=9+10)
GammaRand/case_4  72.0ns ± 6%  47.1ns ± 6%  -34.50%  (p=0.000 n=10+10)
```

### DIFF
--- a/stat/distuv/gamma.go
+++ b/stat/distuv/gamma.go
@@ -219,11 +219,17 @@ func (g Gamma) Rand() float64 {
 		d := a - 1.0/3
 		c := 1 / (3 * math.Sqrt(d))
 		for {
-			u := -exprnd()
 			x := normrnd()
 			v := 1 + x*c
+			if v <= 0.0 {
+				continue
+			}
 			v = v * v * v
-			if u < 0.5*x*x+d*(1-v+math.Log(v)) {
+			u := unifrnd()
+			if u < 1.0-0.0331*(x*x)*(x*x) {
+				return d * v / b
+			}
+			if math.Log(u) < 0.5*x*x+d*(1-v+math.Log(v)) {
 				return d * v / b
 			}
 		}
--- a/stat/distuv/beta_test.go
+++ b/stat/distuv/beta_test.go
@@ -49,7 +49,7 @@ func TestBetaRand(t *testing.T) {
 func testBeta(t *testing.T, b Beta, i int) {
 	const (
 		tol  = 1e-2
-		n    = 5e4
+		n    = 1e5
 		bins = 10
 	)
 	x := make([]float64, n)
--- a/stat/distuv/gamma_test.go
+++ b/stat/distuv/gamma_test.go
@@ -5,6 +5,7 @@
 package distuv
 
 import (
+	"fmt"
 	"math"
 	"sort"
 	"testing"
@@ -114,5 +115,22 @@ func TestGammaPanics(t *testing.T) {
 	g = Gamma{0, 1, nil}
 	if !panics(func() { g.Rand() }) {
 		t.Errorf("Expected Rand panic for Alpha <= 0")
+	}
+}
+
+func BenchmarkGammaRand(b *testing.B) {
+	src := rand.New(rand.NewSource(1))
+	for i, g := range []Gamma{
+		{Alpha: 0.1, Beta: 0.8, Src: src},
+		{Alpha: 0.5, Beta: 0.8, Src: src},
+		{Alpha: 1, Beta: 1, Src: src},
+		{Alpha: 1.6, Beta: 0.4, Src: src},
+		{Alpha: 30, Beta: 1.7, Src: src},
+	} {
+		b.Run(fmt.Sprintf("case %d", i), func(b *testing.B) {
+			for i := 0; i < b.N; i++ {
+				g.Rand()
+			}
+		})
 	}
 }
--- a/stat/distuv/inversegamma_test.go
+++ b/stat/distuv/inversegamma_test.go
@@ -42,7 +42,7 @@ func TestInverseGamma(t *testing.T) {
 func testInverseGamma(t *testing.T, f InverseGamma, i int) {
 	const (
 		tol  = 1e-2
-		n    = 1e6
+		n    = 2e6
 		bins = 50
 	)
 	x := make([]float64, n)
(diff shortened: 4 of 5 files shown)


stat/distuv/gamma.go
```diff
@@ -219,11 +219,17 @@ func (g Gamma) Rand() float64 {
 		d := a - 1.0/3
```
romanwerpachowski, line 219: This above comment is out of date now. Please add a reference for the modification you added.

romanwerpachowski (changes requested):

romanwerpachowski:
Thanks. Can you also add the benchmark from the issue to the PR?

stat/distuv/gamma.go
argusdusty: The comment is still correct. The cited paper also describes the new approach.

romanwerpachowski: Oh, I see. Thanks.

romanwerpachowski (dismissed):

romanwerpachowski:
Thanks for the PR, and welcome!

sbinet:
thanks as well for the nice improvement.

we are a bit nit-picky (sorry about that), but could you remove the "A+C" authorship modification from this PR and send another one with only the authorship bits in?
this way, we can have a nice commit log, say:
```
stat/distuv: improve Gamma.Rand performances
A+C: add Mark Canning
```

we do like nice commit logs :) (like the ones from the Go project itself. at least that's the goal)

thanks again.

sbinet:
(btw, I see you have a nice fft package, faster than the one from scientificgo.org. did you benchmark it against gonum's?
well, we could discuss this on gonum-dev instead of hijacking that PR)

argusdusty:
I've removed the new contributor stuff. I'll send out another pull request containing it after this one. 

Regarding my FFT library, I did benchmark it against gonum/dsp (that would be the "BenchmarkGoDSPFFT" benchmarks), it's significantly faster, but as gonum/dsp is a multithreaded library and mine is single-threaded, it's not a drop-in replacement.

romanwerpachowski (dismissed):

sbinet (changes requested):
LGTM.

(just marking as needing more changes so we don't collect it before the PR w/ authorship is collected)

kortschak:
@argusdusty Our FFT implementation isn't multi-threaded. The current implementation is in an internal package so that we can swap it out when a better one becomes available, keeping the API provided in dsp/fourier the same. The requirements are (and this can be done piecemeal); real sequence FFT, complex sequence FFT, DCT, DST and QW FFT. The current implementation is sequence-length agnostic (no 2^n length requirement) and that behaviour should be kept.

argusdusty:
Apparently I got go-dsp mixed up with gonum/dsp. I added gonum's fft to my benchmarks and my code still runs a few times faster, but only supports 2^n length inputs, so still not a drop-in replacement.

kortschak:
Continue off-topic: I just went back and had another look and NumPy uses the same algorithm as we do, but in C, being a translation of FFTPACK's code. Maybe it's just a matter of looking at where we are losing performance.

argusdusty:
numpy's fft is actually rather slow (though some of that is probably from type conversion/python overhead). That's part of why I made my fft library. fftw is the best, but it's not easy to use from Go/Python without C wrappers, and the code is very complex and hard to translate. Going by http://www.fftw.org/speed/Pentium4-3.06GHz/ fftw runs ~3-4x faster than fftpack, which sounds about right.

sbinet:
needs https://github.com/gonum/gonum/pull/1460

sbinet (approved):

sbinet:
thanks a lot, and welcome :)

PR #1462, opened 2020-09-17
stat/distuv: improve Gamma.Rand performances for 0.2 <= alpha < 1

argusdusty:
Please take a look.

The paper cited in the comment for the `a > 1` (now `a >= smallAlphaThresh`) case mentions this approach as a small note on the final page, so there is no need to update the comments.

I also looked into [numpy's](https://github.com/numpy/numpy/blob/9ee262b5bf89e8c866a507e4d62b78532361adc2/numpy/random/src/legacy/legacy-distributions.c#L47) logic for this case, but a) couldn't determine a source for it, except for what appears to be [a more complex variant](https://link.springer.com/article/10.1007/BF02293108) (Algorithm GS), and b) it's *slightly* slower than this method, at least for `a >= 0.2`.

benchstat:
```
name              old time/op  new time/op  delta
GammaRand/case_0  89.4ns ± 5%  90.2ns ± 3%     ~     (p=0.084 n=20+20)
GammaRand/case_1   212ns ± 3%   101ns ± 5%  -52.28%  (p=0.000 n=20+20)
GammaRand/case_2  25.4ns ± 3%  25.3ns ± 4%     ~     (p=0.369 n=20+20)
GammaRand/case_3  48.6ns ± 6%  48.8ns ± 6%     ~     (p=0.702 n=20+20)
GammaRand/case_4  47.1ns ± 6%  47.6ns ± 3%     ~     (p=0.147 n=20+20)
```

### DIFF
--- a/stat/distuv/gamma.go
+++ b/stat/distuv/gamma.go
@@ -150,73 +150,17 @@ func (g Gamma) Rand() float64 {
 				return eza / b
 			}
 		}
-	case a >= smallAlphaThresh && a < 1:
-		// Generate using:
-		//  Kundu, Debasis, and Rameshwar D. Gupta. "A convenient way of generating
-		//  gamma random variables using generalized exponential distribution."
-		//  Computational Statistics & Data Analysis 51.6 (2007): 2796-2802.
-
-		// TODO(btracey): Change to using Algorithm 3 if we can find the bug in
-		// the implementation below.
-
-		// Algorithm 2.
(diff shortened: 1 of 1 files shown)


romanwerpachowski (approved):
Thanks!

sbinet (approved):

sbinet:
thanks as well.
keep them coming :)

PR #1465, opened 2020-09-21
stat/distuv: improve Poisson.Rand performance for Lambda >= 10

argusdusty:
Please take a look.

benchstat:
```
name                old time/op  new time/op  delta
PoissonRand/case_0   297ns ± 5%    76ns ±10%  -74.56%  (p=0.000 n=10+10)
PoissonRand/case_1   284ns ± 6%   107ns ± 7%  -62.21%  (p=0.000 n=10+10)
PoissonRand/case_2   284ns ± 6%   118ns ± 4%  -58.34%  (p=0.000 n=10+10)
PoissonRand/case_3   236ns ± 3%   239ns ± 6%     ~     (p=0.381 n=10+10)
PoissonRand/case_4   105ns ± 6%   107ns ± 8%     ~     (p=0.236 n=10+10)
PoissonRand/case_5  73.8ns ± 4%  75.1ns ± 4%     ~     (p=0.148 n=10+10)
PoissonRand/case_6  60.3ns ± 5%  60.9ns ± 5%     ~     (p=0.307 n=10+10)
```

### DIFF
--- a/stat/distuv/poisson_test.go
+++ b/stat/distuv/poisson_test.go
@@ -5,6 +5,7 @@
 package distuv
 
 import (
+	"fmt"
 	"math"
 	"sort"
 	"testing"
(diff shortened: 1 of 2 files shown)


romanwerpachowski (approved):
Thanks.

stat/distuv/poisson_test.go
```diff
@@ -120,7 +121,7 @@ func TestPoisson(t *testing.T) {
 func testPoisson(t *testing.T, p Poisson, i int) {
 	const (
 		tol = 1e-2
-		n   = 1e6
+		n   = 2e6
```
sbinet, line 124: out of interest, why do we increase the sample size?

sbinet (approved):

stat/distuv/poisson_test.go
argusdusty: The skewness test fails with a small error on the 1e6 sample. Higher sample sizes pass the test, so I assume it's just sampling error. We can also set the random seed to 2 instead of 1 if test slowdown is a concern.

sbinet: nah, that's fine. I was just being curious.

PR #1467, opened 2020-09-21
interp: Move out cubic spline code to separate files.

romanwerpachowski:
Please take a look.


sbinet (approved):
SGTM (whoever commits/squashes will have to remember to fix the commit message though :))

PR #1468, opened 2020-09-21
dsp/fourier: add radix-2 and radix-4 in-place FFTs

kortschak:
The comments in #1456 about FFT packages was what spurred me to look at cleaning up dsp/internal/fftpack, but I also spent some time thinking about my comment https://github.com/gonum/gonum/pull/1456#issuecomment-693677437 regarding the desire to have a non-radix constrained implementation. I still think this is true since it makes life easier for the general user, but it doesn't preclude us from providing radix-2 and radix-4 FFTs.

These are faster than the current implementation (even with the improvements made in #1463), and they don't allocate, but they do behave the same way as the `Coefficients` and `Sequence` methods on the FFT types we have which the exception that they are less numerically precise (this comes from the approach used to generate the twiddle factors by successive powers rather than an allocated look-up table or `cmplx.Exp` calls ~~(this isn't noted in the docs, but I can add a note to that effect)~~).

Please take a look.

### DIFF
--- a/dsp/fourier/radix24.go
+++ b/dsp/fourier/radix24.go
@@ -1,0 +1,343 @@
+// Copyright ©2020 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package fourier
+
+import (
+	"math"
+	"math/bits"
+	"math/cmplx"
+)
+
+// CoefficientsRadix2 computes the Fourier coefficients of the input
+// sequence, converting the time series in seq into the frequency spectrum,
+// in place and returning it. This transform is unnormalized; a call to
+// CoefficientsRadix2 followed by a call of SequenceRadix2 will multiply the
+// input sequence by the length of the sequence.
+//
+// CoefficientsRadix2 does not allocate, requiring that FFT twiddle factors
+// be calculated lazily. For performance reasons, this is done by successive
+// multiplication, so numerical accuracies can accumulate for large inputs.
+// If accuracy is needed, the FFT or CmplxFFT types should be used.
+//
+// If the length of seq is not an integer power of 2, CoefficientsRadix2 will
+// panic.
+func CoefficientsRadix2(seq []complex128) []complex128 {
+	x := seq
+	switch len(x) {
+	default:
+		if bits.OnesCount(uint(len(x))) != 1 {
+			panic("fourier: radix-2 fft called with non-power 2 length")
+		}
+
+	case 0, 1:
+		return x
+
+	case 2:
+		x[0], x[1] =
+			x[0]+x[1],
+			x[0]-x[1]
+		return x
+
+	case 4:
+		t := x[1] + x[3]
+		u := x[2]
+		v := negi(x[1] - x[3])
+		x[0], x[1], x[2], x[3] =
+			x[0]+u+t,
+			x[0]-u+v,
+			x[0]+u-t,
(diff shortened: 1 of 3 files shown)


kortschak:
@argusdusty Are you interested in reviewing this?

argusdusty:
Afraid I won't have any spare time to commit to that for the foreseeable future.

kortschak:
No worries.

dsp/fourier/radix24.go
```diff
@@ -0,0 +1,343 @@
...
+// series, in place and returning it. This transform is unnormalized; a
+// call to CoefficientsRadix2 followed by a call of SequenceRadix2 will
+// multiply the input sequence by the length of the sequence.
+//
+// SequenceRadix2 does not allocate, requiring that FFT twiddle factors
+// be calculated lazily. For performance reasons, this is done by successive
+// multiplication, so numerical accuracies can accumulate for large inputs.
+// If accuracy is needed, the FFT or CmplxFFT types should be used.
+//
+// If the length of seq is not an integer power of 2, SequenceRadix2
```
vladimir-ch, line 120: s/seq/coeff

dsp/fourier/radix24.go
```diff
@@ -0,0 +1,343 @@
...
+	b := bits.Len(uint(len(x)))
+	if len(x) == 1<<(b-1) {
+		return x
+	}
+	p := make([]complex128, 1<<b)
+	copy(p, x)
+	return p
+}
+
+// TrimRadix2 return the largest slice of x that is has an integer
```
vladimir-ch, line 148: returns

dsp/fourier/radix24.go
```diff
@@ -0,0 +1,343 @@
...
+// series, in place and returning it. This transform is unnormalized; a
+// call to CoefficientsRadix4 followed by a call of SequenceRadix4 will
+// multiply the input sequence by the length of the sequence.
+//
+// SequenceRadix4 does not allocate, requiring that FFT twiddle factors
+// be calculated lazily. For performance reasons, this is done by successive
+// multiplication, so numerical accuracies can accumulate for large inputs.
+// If accuracy is needed, the FFT or CmplxFFT types should be used.
+//
+// If the length of seq is not an integer power of 4, SequenceRadix4
```
vladimir-ch, line 270: s/seq/coeff

dsp/fourier/radix24.go
```diff
@@ -0,0 +1,343 @@
...
+	b := bits.Len(uint(len(x)))
+	if len(x) == 1<<(b-1) && b&0x1 == 1 {
+		return x
+	}
+	p := make([]complex128, 1<<((b+1)&^0x1))
+	copy(p, x)
+	return p
+}
+
+// TrimRadix4 return the largest slice of x that is has an integer
```
vladimir-ch, line 298: returns

dsp/fourier/radix24.go
kortschak: Done

dsp/fourier/radix24.go
kortschak: Done

dsp/fourier/radix24.go
kortschak: Done

dsp/fourier/radix24.go
kortschak: Done

vladimir-ch (approved):

Issue #1469, opened 2020-09-24
About Interpolate package


lkeix:
### Background
When I handle time series data as Approximated function by collecting time series data using Go, Gonum provide plotting the time series data, we can see the time series data outline, but Gonum doesn't provide returning function's coefficient based on the time series data.
we have to use other language (e.g. Python) to calculate function's coefficient based on time series data. 

### Proposal
I want to calculate function's coefficient not using other language. So, I want to add interpolate package for Gonum.

### Potential impact of proposal
* Extent of change
  * add directory interpolate.
  * add cubic.go on interpolate dir.
  * add lagrange.go on interpolate dir.

* implementation detail
  * Cubic Interpolate.
  * Lagrange interpolation.

* who
  * me.

kortschak:
Have you looked at the interp package? https://pkg.go.dev/gonum.org/v1/gonum/interp

If there are things that you need that are not there, suggesting addtions to that package would be the best way forward.

lkeix:
I overlooked. Sorry.
I didn't know AkimaSpline. AkimaSpline seems better than Cubic Spline on not happening over shoot.

Thank you for your response.

romanwerpachowski:
Hi Kei,

I'm the main author of the interp/ package. I was planning on adding natural cubic splines and some other methods to it. If you want to contribute to this package, I'll be happy to help.

Best,
Roman

lkeix:
Hi Roman!

Thank you for replay.
I want to contribute this package.
If you have plan to implement other methods now, I'll help you!

romanwerpachowski:
Great :) Do you want to take a stab at natural cubic splines? We don't have any C2 interpolator in the package now.

lkeix:
OK!
I'll try to implement other spline methods (e.g. cubic spline, cubic B spline, Bezier spline). I'll mention you again, when I implemented it!

romanwerpachowski:
Great :) Let me know if you need any help.

Thanks and best regards,
Roman

romanwerpachowski:
Natural cubic spline was added in #1657.

PR #1475, opened 2020-09-29
Add Natural Cubic interpolator

lkeix:
I implemented Natural Cubic Interpolator.
liked to issue #1469 .
This change modified Gonum to interp/cubic.go, interp_example_test.go.

Please confirm changed points.


interp/cubic.go
```diff
@@ -297,3 +297,68 @@ func fritschButlandEdgeDerivative(xs, ys, slopes []float64, leftEdge bool) float
...
+		} else if coeffIdx == n-2 {
+			coeffIdx--
+		}
+		coeffIdx++
+	}
+	a := cs.coeff.At(4*coeffIdx+3, 0)
+	b := cs.coeff.At(4*coeffIdx+2, 0)
+	c := cs.coeff.At(4*coeffIdx+1, 0)
+	d := cs.coeff.At(4*coeffIdx, 0)
+	return a*math.Pow(x, 3) + b*math.Pow(x, 2) + c*x + d
```
romanwerpachowski, line 363: The cubic splines are usually implemented as polynomials of `x - xs[i]`, where i := argmax_i xs[i] s.t. xs[i] <= x. See [here](https://en.wikipedia.org/wiki/Spline_(mathematics)#Algorithm_for_computing_natural_cubic_splines) for example. Using `x` instead is probably going to be less accurate.

interp/cubic.go
```diff
@@ -297,3 +297,68 @@ func fritschButlandEdgeDerivative(xs, ys, slopes []float64, leftEdge bool) float
...
+			coeffs.Set(i+(n-1)*3-1, 4*(i+1)+3, float64(-3)*xs[i+1])
+			coeffs.Set(i+(n-1)*3-1, 4*(i+1)+2, float64(-1))
+		}
+		sy.Set(i, 0, xy[i])
+		sy.Set(i+n-1, 0, xy[i+1])
+	}
+	coeffs.Set(4*(n-1)-2, 3, float64(3)*xs[0])
+	coeffs.Set(4*(n-1)-2, 2, float64(1))
+	coeffs.Set(4*(n-1)-1, 4*(n-1)-1, float64(3)*xs[n-1])
+	coeffs.Set(4*(n-1)-1, 4*(n-1)-2, float64(1))
```
romanwerpachowski, line 337: The cubic splines are usually implemented as polynomials of `x - xs[i]`, where `i := argmax_i xs[i] s.t. xs[i] <= x`. See [here ](https://en.wikipedia.org/wiki/Spline_(mathematics)#Algorithm_for_computing_natural_cubic_splines)for example. Using x instead is probably going to be less accurate.

romanwerpachowski (changes requested):
Thanks and welcome! I left some comments which hopefully should improve the PR.

What is missing from it are unit tests for the new algorithm. We should cover:

- are provided (X, Y) nodes fitted exactly?
- is the curve C2?
- are we matching the assumed boundary conditions?

Please don't be discouraged by the number of comments. Some of them are meant to make your code compatible with the conventions in the package, to make it easier to use by others :)

interp/cubic.go
```diff
@@ -297,3 +297,68 @@ func fritschButlandEdgeDerivative(xs, ys, slopes []float64, leftEdge bool) float
...
+	coeffsinv.Inverse(coeffs)
```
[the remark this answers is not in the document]
kortschak: This should really be a solve, rather than a multiply with an inverse, with an appropriate change of the type's field to being a `mat.VecDense` (concrete, not pointer), this can become `return cs.coeff.SolveVec(coeffs, &cs.coeff)`, with the opening preamble becoming `cs.coeff.ReuseAsVec((n - 1) * 4)` rather than the allocation that's there (and using it for both the temporary `sy` that's here and the final result).

Also note that there are `float64` conversions of untyped constants that are unnecessary in a few places above.

interp/cubic.go
```diff
@@ -297,3 +297,68 @@ func fritschButlandEdgeDerivative(xs, ys, slopes []float64, leftEdge bool) float
...
+	xs []float64
+
+	coeff *mat.Dense
+}
+
+// Fit fits a predictor to (X, Y) value pairs provided as two slices.
+func (cs *CubicSpline) Fit(xs, xy []float64) error {
+	n := len(xs)
+	cs.coeff = mat.NewDense((n-1)*4, 1, nil)
+	coeffs := mat.NewDense((n-1)*4, (n-1)*4, nil)
```
kortschak, line 312: I'm wondering if there is not a more efficient representation of this; `coeffs` is very sparse, with obvious structure in the data. @romanwerpachowski is there a banded representation that works for this? Outside my area, but brief reading says that this can be represented as a tridiagonal which would take us from O(n^2) space to O(n) space.

(Note that we don't currently implement GB solvers).

romanwerpachowski: Yes, it's a tridiagonal banded matrix. See e.g. http://www.inase.org/library/2014/athens/bypaper/MMCTSE/MMCTSE-01.pdf

kortschak: Yeah, that's what I thought, though the matrix that's created here is not tridiagonal, so I was unsure.

romanwerpachowski: I suspect that's because it's for the coefficients of a different polynomial (see one of my review comments).

lkeix:
Hi roman!
Thank you for code review!
I'll reconstruct Natural Cubic Spline method based on your pointed out again.

vladimir-ch:

PR #1476, opened 2020-09-29
lapack/gonum: add Dtbtrs

vladimir-ch:
Please take a look.

### DIFF
--- a/lapack/testlapack/general.go
+++ b/lapack/testlapack/general.go
@@ -1464,9 +1464,10 @@ func dlansy(norm lapack.MatrixNorm, uplo blas.Uplo, n int, a []float64, lda int)
 			}
 		}
 		return max
-	default:
-		// lapack.Frobenius:
+	case lapack.Frobenius:
 		panic("not implemented")
+	default:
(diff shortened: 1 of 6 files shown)


lapack/testlapack/general.go
```diff
@@ -1754,15 +1748,147 @@ func dlantr(norm lapack.MatrixNorm, uplo blas.Uplo, diag blas.Diag, m, n int, a
...
 					}
 				}
-				return maxsum
 			}
 		}
-	default:
-		// lapack.Frobenius
+	case lapack.Frobenius:
 		panic("not implemented")
 	}
```
kortschak, line 1759: Add `default` back with a `panic("invalid norm")` (and below)?

vladimir-ch: I've dropped the commit that modified `dlantr` because `dlansy` would need the same modification, it is not related to the purpose of the PR and doesn't add much value anyway. I've replaced it with a commit that adds the default cases as you suggested.

kortschak (approved):

PR #1478, opened 2020-10-04
lapack/gonum: don't loop unnecessarily in Dlanv2

vladimir-ch:
Please take a look.


kortschak (approved):
Yes, sorry for not noticing that.

vladimir-ch:
Nah, this is fully on me. And I had to fix it again (－‸ლ). PTAL

kortschak (approved):
I can't believe that I still get caught by that. It's a thing that I think Go should not have brought from C.

PR #1482, opened 2020-10-07
graph/flow: skip rather than erroring when there is no benchmark data

kortschak:
Please take a look.

### DIFF
--- a/graph/flow/control_flow_bench_test.go
+++ b/graph/flow/control_flow_bench_test.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"math"
+	"os"
 	"path/filepath"
 	"strings"
 	"testing"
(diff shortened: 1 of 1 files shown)


graph/flow/control_flow_bench_test.go
```diff
@@ -31,6 +32,9 @@ func BenchmarkDominators(b *testing.B) {
 
 	fis, err := ioutil.ReadDir(testdata)
 	if err != nil {
+		if os.IsNotExist(err) {
```
sbinet, line 35: just thinking out loud, but isn't now the recommended way to use `errors.Is` or `errors.As`?

sbinet (approved):

graph/flow/control_flow_bench_test.go
kortschak: That seems heavy for this.

PR #1484, opened 2020-10-10
mat: add TriBandDense.SolveTo and SolveVecTo

### LINKED ISSUE #1471: mat: linear equation solver for triangular band matrices missing

I want to solve a linear equation system described by a triangular banded matrix. This is for an implementation of an interpolation algorithm.

### What did you try?

I looked at `mat/solve.go`, but it only handles general dense matrices.

### How does Gonum not allow you to achieve your goal?

Current `mat.Solve()` is too slow for the intended purpose (fitting a spline interpolator, which can be done many times when calculating sensitivities to inputs, for example).

go version go1.14.3 windows/amd64

Gonum is current master.

### Is this feature absent from the current master?

Yes.

### Are you able to help contribute the feature?

Yes.

vladimir-ch:
Please take a look.

I also cleaned up the `basicVector` type a bit so that the tests for `SolveTo` and `SolveVecTo` are basically identical.

Closes #1471 

cc @romanwerpachowski

### DIFF
--- a/mat/matrix_test.go
+++ b/mat/matrix_test.go
@@ -372,48 +372,6 @@ func TestDet(t *testing.T) {
 	testOneInputFunc(t, "DetVsChol", f, denseComparison, sameAnswerFloatApproxTol(1e-10), isAnyType, isWide)
 }
 
-type basicVector struct {
-	m []float64
-}
-
-func (v *basicVector) AtVec(i int) float64 {
-	if i < 0 || i >= v.Len() {
-		panic(ErrRowAccess)
-	}
-	return v.m[i]
-}
-
-func (v *basicVector) At(r, c int) float64 {
-	if c != 0 {
-		panic(ErrColAccess)
-	}
-	return v.AtVec(r)
-}
-
-func (v *basicVector) Dims() (r, c int) {
-	return v.Len(), 1
-}
-
-func (v *basicVector) T() Matrix {
-	return Transpose{v}
-}
-
-func (v *basicVector) Len() int {
-	return len(v.m)
-}
-
-type rawVector struct {
-	basicVector
-}
-
-func (v *rawVector) RawVector() blas64.Vector {
-	return blas64.Vector{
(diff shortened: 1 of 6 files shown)


mat/matrix_test.go
```diff
@@ -782,9 +740,9 @@ func TestMulVecToer(t *testing.T) {
 						}
 						x = dst
 					case 2:
-						x = &rawVector{basicVector{random(n)}}
+						x = &rawVector{(*basicVector)(NewVecDense(n, random(n)))}
```
romanwerpachowski, line 743: Why not using `asBasicVector` here?

romanwerpachowski (commented):
Thanks!

mat/matrix_test.go
vladimir-ch: Because `asBasicVector` returns a `Vector` but `rawVector` wraps `*basicVector` so I would need to extract the concrete type anyway. Maybe the `asBasicSomething` functions should return concrete types instead of interfaces. Regarding `rawVector`, I'd like to revisit this test and look if I can remove this type and simplify the test.

romanwerpachowski (approved):

vladimir-ch:
I'd like to wait also for @kortschak if he has time to take a look.

mat/mul_test.go
```diff
@@ -218,6 +218,24 @@ func (m *basicMatrix) T() Matrix {
 	return Transpose{m}
 }
 
+type basicVector VecDense
```
kortschak, line 221: Maybe these should all go in a separate file in future? (not just `basicVector`)

kortschak (approved):

kortschak:
Was already reviewing when that call-out happened.

mat/mul_test.go
vladimir-ch: Yes, that's exactly what I was thinking too.

vladimir-ch:
Thanks!

PR #1487, opened 2020-10-12
mat: consolidate basic types for testing in single file

vladimir-ch:
Please take a look.


mat/basictypes_test.go
```diff
@@ -0,0 +1,114 @@
...
+import "gonum.org/v1/gonum/blas/blas64"
+
+func asBasicMatrix(d *Dense) *basicMatrix              { return (*basicMatrix)(d) }
+func asBasicVector(d *VecDense) *basicVector           { return (*basicVector)(d) }
+func asBasicSymmetric(s *SymDense) *basicSymmetric     { return (*basicSymmetric)(s) }
+func asBasicTriangular(t *TriDense) *basicTriangular   { return (*basicTriangular)(t) }
+func asBasicBanded(b *BandDense) *basicBanded          { return (*basicBanded)(b) }
+func asBasicSymBanded(s *SymBandDense) *basicSymBanded { return (*basicSymBanded)(s) }
+func asBasicTriBanded(t *TriBandDense) *basicTriBanded { return (*basicTriBanded)(t) }
+func asBasicDiagonal(d *DiagDense) *basicDiagonal      { return (*basicDiagonal)(d) }
```
kortschak, line 16: Are these all in a block at the top so we know what we have?

kortschak (approved):

mat/basictypes_test.go
vladimir-ch: Possibly. I was split having a block here or have them with the type. I can change it to your preference.

kortschak: Either way is fine.

vladimir-ch: Then I'll leave it here, less work.

Issue #1488, opened 2020-10-13
graph/community: modularisation goes into infinite loop when receiving small deltaQ in Louvain


asdftamir:
### What are you trying to do?
Hey all, part of my code uses the Louvain algorithm implemented in gonum.
I've encountered an infinite loop that might happen.

### What did you do?

My code creates a large graph and runs the Louvain algorithm multiple times with different resolutions.
Some of the runs never end, while stuck in an infinite loop in function `localMovingHeuristic` in `louvain_undirected.go`.
The problem occurs when the `deltaQ` function returns a **really** small result (for example 0.000000000000000000002730334092404377).
As a result, the `if` on line 489 is false and the node keeps moving between two communities.
I couldn't reproduce a small example yet, but still working on it.

### What did you expect to happen?
In other implementations of the louvain algorithm I saw the modularity score was compared to some small number instead of 0.

### What actually happened?
The loop never ended.

### What version of Go and Gonum are you using?

go version go1.14.4 windows/amd64 
gonum.org/v1/gonum v0.8.1 h1:wGtP3yGpc5mCLOLeTeBdjeui9oZSz5De0eOjMLC/QuQ=

### Does this issue reproduce with the current master?
yes

kortschak:
> In other implementations of the louvain algorithm I saw the modularity score was compared to some small number instead of 0.

Can you point to these so we can see what other people consider to be a reasonable tolerance?

asdftamir:
https://github.com/taynaud/python-louvain/blob/master/community/community_louvain.py - See the __MIN variable

https://github.com/riyadparvez/louvain-method/blob/66a6ff75fcf0eaa7f83f10e8ec41f91459aeb949/community.cpp#L358 - min_modularity is a parameter passed from user input with the description : "given pass stops when the modularity is increased by less than epsilon"

kortschak:
Thanks.

kortschak:
@asdftamir Are you happy to have this closed?

PR #1489, opened 2020-10-13
graph/community: add threshold for modularity gain

asdftamir:
Added modularityEpsilon constant and using it in all "localMovingHeuristic" implementations.
Addresses issue https://github.com/gonum/gonum/issues/1488

### DIFF
--- a/graph/community/louvain_common.go
+++ b/graph/community/louvain_common.go
@@ -357,9 +357,8 @@ const (
 	negativeWeight = "community: unexpected negative edge weight"
 	positiveWeight = "community: unexpected positive edge weight"
 
-	// Threshold for modularity change.
-	// If a modularity gain is lower than this value, we assume there is no gain.
-	modularityEpsilon = 0.0000001
+	// deltaQtol is the tolerance for progression of the local moving heuristic's improvement of Q.
+	deltaQtol = 1e-15
 )
 
 // positiveWeightFuncFor returns a constructed weight function for the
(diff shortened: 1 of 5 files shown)


graph/community/louvain_common.go
```diff
@@ -356,6 +356,10 @@ func (s *slice) TakeMin(p *int) bool {
 const (
 	negativeWeight = "community: unexpected negative edge weight"
 	positiveWeight = "community: unexpected positive edge weight"
+
+	// Threshold for modularity change.
+	// If a modularity gain is lower than this value, we assume there is no gain.
+	modularityEpsilon = 0.0000001
```
kortschak, line 362: Please rename this to `deltaQtol`, use e notation and change the comment to 

```
// deltaQtol is the tolerance for progression of the local moving heuristic's improvement of Q.
```

The value seems high compared to your identified problem. Does a value of 1e-14 or thereabouts fix the issue for you?

kortschak (changes requested):
Thanks for send this contribution, there are couple of minor changes that I would like you to make.

graph/community/louvain_common.go
asdftamir: Done, and yes that fixes the issue

asdftamir:
Thanks for the quick reply! pushed the changes :)

kortschak (approved):
I think we need to get a reproducer (possibly even just a computationally generated one if the input data are difficult/private in order the understand the actual tolerances that we have (we may need to adjust `deltaQtol` to be less tight if someone else comes along with a hanging graph.

However, I'm happy with this until we find a bad case.

Can you please send another PR with a single commit with the commit message "A+C: add \<your name\>" adding yourself to CONTRIBUTORS and you or your employer as appropriate to AUTHORS. After that we can merge this.

asdftamir:
opened https://github.com/gonum/gonum/pull/1490 
Ill try to think of a way to create a minimal reproducer (the original graph had 6K nodes and weighted edges)

kortschak:
No worries. I don't think there is a rush here.

PR #1491, opened 2020-10-20
lapack: add Dlangt

vladimir-ch:
A first step towards adding a solver for tridiagonal matrices.

Please take a look.

cc @romanwerpachowski

### DIFF
--- a/lapack/gonum/dlangt.go
+++ b/lapack/gonum/dlangt.go
@@ -1,0 +1,115 @@
+// Copyright ©2020 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package gonum
+
+import (
+	"math"
+
+	"gonum.org/v1/gonum/lapack"
+)
+
+// Dlangt returns the value of the given norm of an n×n tridiagonal matrix
+// represented by the three diagonals.
+//
+// d must have length at least n and dl and du must have length at least n-1.
+func (impl Implementation) Dlangt(norm lapack.MatrixNorm, n int, dl, d, du []float64) float64 {
+	switch {
+	case norm != lapack.MaxAbs && norm != lapack.MaxRowSum && norm != lapack.MaxColumnSum && norm != lapack.Frobenius:
+		panic(badNorm)
+	case n < 0:
+		panic(nLT0)
+	}
+
+	if n == 0 {
+		return 0
+	}
+
+	switch {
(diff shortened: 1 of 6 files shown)


lapack/lapack64/lapack64.go
```diff
@@ -19,6 +19,14 @@ func Use(l lapack.Float64) {
 	lapack64 = l
 }
 
+// Tridiagonal represents a tridiagonal matrix using its three diagonals.
+type Tridiagonal struct {
```
vladimir-ch, line 23: I added the type here because it exists only in Lapack. An alternative would be in `blas64` where all the other types are.

lapack/gonum/dlangt.go
```diff
@@ -0,0 +1,111 @@
...
+
+package gonum
+
+import (
+	"math"
+
+	"gonum.org/v1/gonum/lapack"
+)
+
+func (impl Implementation) Dlangt(norm lapack.MatrixNorm, n int, dl, d, du []float64) float64 {
```
kortschak, line 13: docs?

lapack/lapack64/lapack64.go
kortschak: Keeping it here is reasonable.

vladimir-ch: Ok, thanks.

lapack/gonum/dlangt.go
vladimir-ch: I can't believe I forgot them but I've added them now. PTAL

vladimir-ch:
@kortschak Could you please take a look at the failure for Arm64?
```
=== CONT  TestExponentialFitPrior
    general_test.go:127: nil and uniform weighted prior doesn't match for incremental update for (0, 4). Uniform weighted is &{38.08903260230703 <nil>}, nil is &{38.08903260230704 <nil>}
--- FAIL: TestExponentialFitPrior (0.08s)
```

kortschak:
I can't replicate that, but that test uses global rand state which should be fixed. I'll send that soon.

lapack/gonum/dlangt.go
kortschak: Did you push that? I can't find them.

vladimir-ch: I did not, I stashed them with other changes I'm working on. But now they're here.

kortschak (approved):

PR #1494, opened 2020-10-21
lapack: add Dlagtm

vladimir-ch:
This is just a draft for now because:

* `Dlagtm` takes a `trans` parameter but with the tridiagonal matrices the caller can easily handle tranposes by swapping the arguments for `dl` and `du`. This is actually what the solve/factorization calls like `Dgtsv` et al. do, they don't take a `trans` parameter. Since Lapacke doesn't wrap this call anyway, we have more freedom of departing from the Lapack interface for tridiagonal matrices (which feels like it didn't receive much attention anyway) and remove the parameter here.
* Lapack's `DLAGTM` assumes that `alpha` and `beta` are in `{-1,0,1}` but I don't see a reason to not support any values. This would be another deviation from the reference Lapack. I think we should do it but would like to check first.

Please take a look.

### DIFF
--- a/lapack/gonum/dlagtm.go
+++ b/lapack/gonum/dlagtm.go
@@ -1,0 +1,109 @@
+// Copyright ©2020 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package gonum
+
+import "gonum.org/v1/gonum/blas"
+
+// Dlagtm performs one of the matrix-matrix operations
+//  C = alpha * A * B + beta * C   if trans == blas.NoTrans
+//  C = alpha * Aᵀ * B + beta * C  if trans == blas.Trans or blas.ConjTrans
+// where A is an m×m tridiagonal matrix represented by its diagonals dl, d, du,
+// B and C are m×n dense matrices, and alpha and beta are scalars.
+func (impl Implementation) Dlagtm(trans blas.Transpose, m, n int, alpha float64, dl, d, du []float64, b []float64, ldb int, beta float64, c []float64, ldc int) {
+	switch {
+	case trans != blas.NoTrans && trans != blas.Trans && trans != blas.ConjTrans:
+		panic(badTrans)
+	case m < 0:
+		panic(mLT0)
+	case n < 0:
+		panic(nLT0)
+	case ldb < max(1, n):
+		panic(badLdB)
+	case ldc < max(1, n):
+		panic(badLdC)
+	}
+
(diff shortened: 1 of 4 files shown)


lapack/gonum/dlagtm.go
```diff
@@ -0,0 +1,133 @@
...
+package gonum
+
+import "gonum.org/v1/gonum/blas"
+
+// Dlagtm performs one of the matrix-matrix operations
+//  C = alpha * A * B + beta * C   if trans == blas.NoTrans
+//  C = alpha * Aᵀ * B + beta * C  if trans == blas.Trans or blas.ConjTrans
+// where A is an m×m tridiagonal matrix represented by its diagonals dl, d, du,
+// B and C are m×n dense matrices, and alpha and beta are scalars.
+func (impl Implementation) Dlagtm(trans blas.Transpose, m, n int, alpha float64, dl, d, du []float64, b []float64, ldb int, beta float64, c []float64, ldc int) {
```
kortschak, line 14: Remove the redundant `[]float64` after `du`?

lapack/gonum/dlagtm.go
```diff
@@ -0,0 +1,133 @@
...
+		} else {
+			for j := 0; j < n; j++ {
+				c[j] += alpha * d[0] * b[j]
+			}
+		}
+		return
+	}
+
+	if alpha == 1 {
+		if trans == blas.NoTrans {
```
kortschak, line 81: Rather than duplicating the code, perhaps just do `du, dl = dl, du` if `trans != blas.NoTrans`?

kortschak:
I think that allowing beta to be any value seems reasonable — in fact I'll go further and say that the LAPACK behaviour seems unreasonable; it promotes sloppy code.

For the transpose, there's a middle way that take the transpose flag and swaps `du` and `dl` as necessary.

lapack/gonum/dlagtm.go
vladimir-ch: Good idea, done.

lapack/gonum/dlagtm.go
vladimir-ch: It's only a stylistic detail but I'd rather leave it because I like how it separates dl, d, du, that belong together, from b.

vladimir-ch:
I've removed the Draft tag.

lapack/gonum/dlagtm.go
kortschak: I thought that's might be why it's there.

kortschak (approved):

PR #1495, opened 2020-10-23
lapack: add Dgtsv

vladimir-ch:
Please take a look.


lapack/gonum/dgtsv.go
```diff
@@ -0,0 +1,100 @@
...
+
+// Dgtsv solves the equation
+//  A * X = B
+// where A is an n×n tridiagonal matrix. It uses Gaussian elimination with
+// partial pivoting. The equation Aᵀ * X = B may be solved by swapping the
+// arguments for du and dl.
+//
+// On entry, dl, d and du contain the sub-diagonal, the diagonal and the
+// super-diagonal, respectively, of A. On return, the first n-2 elements of dl
+// will be overwritten by the n-2 elements of the second super diagonal of the
```
vladimir-ch, line 17: I'm split on documenting this. I think it's useless without the matrix L and it's practically impossible to test. Saying it's overwritten on output would be much easier without any loss.

vladimir-ch: Besides, if nrhs is 0, correctly we should still compute the factorization which is not what the code currently does.

lapack/lapack64/lapack64.go
```diff
@@ -420,6 +420,21 @@ func Ggsvd3(jobU, jobV, jobQ lapack.GSVDJob, a, b blas64.General, alpha, beta []
...
+// where A is an n×n tridiagonal matrix. It uses Gaussian elimination with
+// partial pivoting.
+//
+// On entry, a contains the matrix A, on return it will be overwritten.
+//
+// On entry, b contains the n×nrhs right-hand side matrix B. On return, if ok is
+// true, it will be overwritten by the solution matrix X.
+//
+// Gtsv returns whether the solution X has been successfuly computed.
+func Gtsv(a Tridiagonal, b blas64.General) (ok bool) {
```
vladimir-ch, line 434: Having a `trans` parameter here would actually be useful. Are you fine with adding it, @kortschak ?

lapack/lapack64/lapack64.go
```diff
@@ -420,6 +420,21 @@ func Ggsvd3(jobU, jobV, jobQ lapack.GSVDJob, a, b blas64.General, alpha, beta []
 	return lapack64.Dggsvd3(jobU, jobV, jobQ, a.Rows, a.Cols, b.Rows, a.Data, max(1, a.Stride), b.Data, max(1, b.Stride), alpha, beta, u.Data, max(1, u.Stride), v.Data, max(1, v.Stride), q.Data, max(1, q.Stride), work, lwork, iwork)
 }
 
+// Gtsv solves the equation
+//  A * X = B
+// where A is an n×n tridiagonal matrix. It uses Gaussian elimination with
+// partial pivoting.
+//
+// On entry, a contains the matrix A, on return it will be overwritten.
```
vladimir-ch, line 428: I gave up on documenting the contents of A upon return.

lapack/gonum/dgtsv.go
```diff
@@ -0,0 +1,100 @@
...
+	case n < 0:
+		panic(nLT0)
+	case nrhs < 0:
+		panic(nrhsLT0)
+	case ldb < max(1, nrhs):
+		panic(badLdB)
+	}
+
+	if n == 0 || nrhs == 0 {
+		return true
```
kortschak, line 36: Add the `nhrs == 0` case of the back solve?

lapack/lapack64/lapack64.go
```diff
@@ -420,6 +420,21 @@ func Ggsvd3(jobU, jobV, jobQ lapack.GSVDJob, a, b blas64.General, alpha, beta []
...
+//
+// On entry, b contains the n×nrhs right-hand side matrix B. On return, if ok is
+// true, it will be overwritten by the solution matrix X.
```
kortschak, line 431: Is it worth noting that `b`'s contents are destroyed either way?

lapack/testlapack/locallapack.go
```diff
@@ -0,0 +1,162 @@
...
+package testlapack
+
+import (
+	"math"
+
+	"gonum.org/v1/gonum/blas"
+	"gonum.org/v1/gonum/lapack"
+)
+
+func dlagtm(trans blas.Transpose, m, n int, alpha float64, dl, d, du []float64, b []float64, ldb int, beta float64, c []float64, ldc int) {
```
kortschak, line 15: Note that this is copied from `gonum.Implementation`'s method to avoid cyclic dependencies.

Maybe we should do the testing for lapack/gonum in gonum_test? (#future)

lapack/testlapack/locallapack.go
```diff
@@ -0,0 +1,162 @@
...
+				c[i*ldc+j] += alpha * (dl[i-1]*b[(i-1)*ldb+j] + d[i]*b[i*ldb+j] + du[i]*b[(i+1)*ldb+j])
+			}
+		}
+		for j := 0; j < n; j++ {
+			c[(m-1)*ldc+j] += alpha * (dl[m-2]*b[(m-2)*ldb+j] + d[m-1]*b[(m-1)*ldb+j])
+		}
+	}
+}
+
+func dlangt(norm lapack.MatrixNorm, n int, dl, d, du []float64) float64 {
```
kortschak, line 86: Same here.

lapack/testlapack/locallapack.go
```diff
@@ -160,3 +161,516 @@ func dlangt(norm lapack.MatrixNorm, n int, dl, d, du []float64) float64 {
...
+		}
+	case lapack.Frobenius:
+		panic("not implemented")
+	default:
+		panic("invalid norm")
+	}
+	return value
+}
+
+func dlantr(norm lapack.MatrixNorm, uplo blas.Uplo, diag blas.Diag, m, n int, a []float64, lda int, work []float64) float64 {
```
kortschak, line 354: Reason comment; it wasn't on the previous copy, but the others have one.

lapack/gonum/dgtsv.go
vladimir-ch: > ... it's practically impossible to test

I realized that of course I could test by comparing to the result of Dgetrf but it still complicates the interface and the test for no clear benefit. If the LU factorization of A is needed, then Dgttrf can be used which I'll add soon.

> What can be done with the elements of dl with this information?

Upon a successful return, d, du and dl contain the matrix U from the LU factorization of A. I don't see any use for U without L, I think this is more of a side-effect of using the slices for storing the factorization somewhere. And if the factorization is needed, then Dgttrf will be available. For me Dgtsv is just a convenience function for one-time solution of a linear system without getting the factorization explicitly. I plan to use Dgtsv in mat.TridiagDense.SolveTo and Dgttrf in mat.TridiagLU.Factorize.

In total, I've change the documentation to say that dl, d and du will be overwritten without specifying how.

lapack/lapack64/lapack64.go
vladimir-ch: In an isolated type `.T() Tridiagonal` would be best but as you write it's different to all other blas64 types, so I've added a trans parameter here. I think we have some freedom in deciding lapack64 package API.

lapack/lapack64/lapack64.go
vladimir-ch: Thanks.

lapack/gonum/dgtsv.go
vladimir-ch: I thought about that but `b` is already accessed during the factorization. And we check the length of b too, so the order would be: access b during the factorization (even though it's not actually accessed when `nrhs == 0`, quick return if `nrhs == 0`, and only then check length of b which I think is a weird order. And all this just so that the caller has U available (in a documented way).

lapack/lapack64/lapack64.go
vladimir-ch: Done.

lapack/testlapack/locallapack.go
vladimir-ch: Would there actually be cyclic dependencies? I'm not sure now. I started making copies of `Dlan*` functions because some of them are not in lapacke. If not, couldn't we call into `gonum.Implementation` directly without putting them into the impl parameter's interface? If we keep these copies, maybe move them into `testlapack.internal`?

lapack/testlapack/locallapack.go
vladimir-ch: Done.

lapack/testlapack/locallapack.go
vladimir-ch: Done.

vladimir-ch:
PTAL

lapack/testlapack/locallapack.go
vladimir-ch: Of course there are currently cyclic dependencies. Now I fully understand your comment. Yes, #future.

kortschak (approved):

PR #1496, opened 2020-10-24
README: specify which compiler we test on and support

kortschak:
Please take a look.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ go get -u gonum.org/v1/gonum/...
 
 ## Supported Go versions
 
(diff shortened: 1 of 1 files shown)


sbinet (approved):
LGTM.

I guess we could revisit this when/if the ci infrastructures we have access to provide easy access to other compilers.

kortschak:
Yes, it's the availability of compilers, but also availability of time to work on additional moving targets.

sbinet:
Yes, access and womanpower/time.
The usual conundrum.

PR #1497, opened 2020-10-27
mat: document that Condition errors are returned by Cholesky methods

### LINKED ISSUE #1470: mat: document places where Condition error is returned
We don't do this elsewhere, though we probably should wherever it's returned; we should say that the error type returned is a `Condition` error when non-nil. This should be a separate PR dealing with all places where this happens.

_Originally posted by @kortschak in https://github.com/gonum/gonum/pull/1441#discussion_r494624335_

kortschak:
Please take a look.

Closes #1470.

### DIFF
--- a/mat/cholesky.go
+++ b/mat/cholesky.go
@@ -225,6 +225,8 @@ func (c *Cholesky) LogDet() float64 {
 
 // SolveTo finds the matrix X that solves A * X = B where A is represented
 // by the Cholesky decomposition. The result is stored in-place into dst.
+// If the Cholesky decomposition is singular or near-singular a Condition error
+// is returned. See the documentation for Condition for more information.
 func (c *Cholesky) SolveTo(dst *Dense, b Matrix) error {
 	if !c.valid() {
 		panic(badCholesky)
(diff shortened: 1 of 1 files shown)


mat/cholesky.go
```diff
@@ -225,6 +225,8 @@ func (c *Cholesky) LogDet() float64 {
 
 // SolveTo finds the matrix X that solves A * X = B where A is represented
 // by the Cholesky decomposition. The result is stored in-place into dst.
+// If the Cholesky docomposition is singular or near-singular a Condition error
```
vladimir-ch, line 228: decomposition (with e) everywhere.

mat/cholesky.go
```diff
@@ -778,6 +786,7 @@ func (ch *BandCholesky) SolveTo(dst *Dense, b Matrix) error {
 
 // SolveVecTo finds the vector x that solves A * x = b where A is represented by
 // the Cholesky decomposition. The result is stored in-place into dst.
+// If a non-nil error is returned it will be a Condition error.
```
vladimir-ch, line 789: Why a different wording here?

kortschak: Missed it.

mat/cholesky.go
kortschak: My eyes are getting bad.

kortschak:
Sorry about that. PTAL

vladimir-ch (approved):

PR #1500, opened 2020-11-04
graph/layout: update for changes in gonum/plot@v0.8.1

kortschak:
I'm not entirely sure why only one label is slightly off, and in only some plots. It suggests to me that perhaps gonum/plot#294 is not completely fixed.

Please take a look.


sbinet (approved):
it seems only labels `0`, `4` and `7` are affected. (arf, no, `2` and `8` as well)

it'd be interesting to check whether that's inherent to these labels or their position in the graph.
_eg:_ replace `0` with, say, `1` or `x` or `g` in `sheet_isomap_golden.png` and see what gives?

hum... it seems `tube_isomap_golden.png` has 2 displaced labels (`2` and `8`)

anyways. LGTM.

kortschak:
Yeah, I might have more of a play later and see what is going on.

PR #1503, opened 2020-11-19
Prevent possibility of panic in self-referential graphs

samsondav:
Decoding can panic in self-referential graphs. This panic is not caught. Since graphs may be untrusted input, this is dangerous.

This PR replaces the panic with an error.

### DIFF
--- a/graph/encoding/dot/decode.go
+++ b/graph/encoding/dot/decode.go
@@ -84,7 +84,7 @@ func copyGraph(dst encoding.Builder, src *ast.Graph) (err error) {
 		case error:
 			err = e
 		default:
-			panic(e)
+			err = fmt.Errorf(fmt.Sprintf("error decoding graph %s", e))
 		}
 	}()
 	gen := &simpleGraph{
--- a/graph/encoding/dot/decode_test.go
+++ b/graph/encoding/dot/decode_test.go
@@ -393,6 +393,19 @@ func TestMultigraphDecoding(t *testing.T) {
 	}
 }
 
+func TestSelfReference(t *testing.T) {
+	var dst encoding.Builder = simple.NewDirectedGraph()
+
+	err := Unmarshal([]byte(selfReferentialGraph), dst)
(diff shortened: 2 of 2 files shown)


kortschak:
Thanks for sending this change.

I don't think this is the right fix for this, but I'm also not entirely sure what the right fix is. I think you're right that we should do something about this, but exactly what is not clear; there are a few options. It would be good to file an issue with context to discuss prior to sending a fix.

Briefly, the possible fixes are:

1. Document that given that adding a self edge to simple graphs [is allowed to panic](https://pkg.go.dev/gonum.org/v1/gonum/graph/#EdgeAdder) that either a wrapper for the destination that checks for self edges be used by the user or the user wraps the `Unmarshal` call in a recover.
2. Require that self-edge setting panics are of type `error`.
3. Wrap the loop [here](https://github.com/gonum/gonum/blob/cb7e73b504732e1bcf42dc7a02654bcf7b3f33d2/graph/encoding/dot/decode.go#L278-L286) in a recover that either passes an `error` up or wraps any other type in an `error` and panics that `error`.

I think that 3. is probably the right thing to do; it doesn't place any additional requirements on the user and reduces the possibility of converting other panics into an `error` (the helpers in that loop don't hold code that has a reasonably likelihood of causing a panic).

samsondav:
@kortschak It is not clear that this code can panic. Given that this is a parsing library quite likely to be dealing with untrusted input, is panic-ing really ever the right thing to do?

kortschak:
The parser itself is fuzzed extensively (the corpus and fuzz test are [here](https://github.com/gonum/gonum/tree/master/graph/formats/dot/fuzz)).

There are a variety of panics that we can't know about before hand that should be allowed to panic, runtime problems for example. The only things that are allowed in the graph API that can cause panics are re-addition of nodes which is already handled by the decoder and setting self-loops, which is fixed by #1504 by lowering them to an `error`. By placing the handling of the self-loop panic in the edge additional handler I minimise the possibility of unnecessarily catching extraneous panics.

PR #1505, opened 2020-11-21
dsp/fourier,stat/samplemv: fix 32 bit architecture bugs

kortschak:
The Halton fix is pretty hacky, but I've spent a fair bit of time looking through the Halton code and the R original and I'm not completely happy with my understanding of the relationship between the two (input from @btracey would be helpful here). I also think that the behaviour of the R code depends on R's sloppy behaviour in implicitly converting between `double` and `integer` types and allowing the use of `double` values to index into arrays.

Please take a look.

Updates #1444

### DIFF
--- a/dsp/fourier/radix24_test.go
+++ b/dsp/fourier/radix24_test.go
@@ -10,6 +10,7 @@ import (
 	"math/bits"
 	"strconv"
 	"testing"
+	"unsafe"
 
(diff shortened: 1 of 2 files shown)


kortschak:
*ping*

sbinet (approved):

vladimir-ch (approved):

PR #1506, opened 2020-11-21
graph/path: fix build for 386

kortschak:
Please take a look.

Updates #1444

### DIFF
--- a/graph/path/bellman_ford_moore.go
+++ b/graph/path/bellman_ford_moore.go
@@ -39,7 +39,7 @@ func BellmanFordFrom(u graph.Node, g graph.Graph) (path Shortest, ok bool) {
 
 	// The maximum number of edges in a graph is |V| * (|V|-1) which is also
 	// the worst case complexity.
-	// If the queue-loop has more iterations than the amount of maximum edges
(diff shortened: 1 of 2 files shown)


graph/path/bellman_ford_moore.go
```diff
@@ -41,8 +41,8 @@ func BellmanFordFrom(u graph.Node, g graph.Graph) (path Shortest, ok bool) {
 	// the worst case complexity.
 	// If the queue-loop has more iterations than the amount of maximum edges
 	// it indicates that we have a negative cycle.
-	maxEdges := len(nodes) * (len(nodes) - 1)
-	var loops int
+	maxEdges := int64(len(nodes)) * int64(len(nodes)-1)
```
kortschak, line 44: The alternative for the negative cycle detection in Bellman-Ford is to use division at the condition below and keep it in `int`. I chose this approach because it has zero performance impact on amd64 while a division per search loop would be non-zero.

kortschak:
*ping*

sbinet (approved):

vladimir-ch (approved):

PR #1512, opened 2020-11-26
graph/path: allow incremental Bellman-Ford search

kortschak:
This completes something that I had intended to do a couple of years ago (#486), but thought wasn't efficiently feasible. It is.

This makes time limited Bellman-Ford searches a possibility (this is used for example in routing decisions where there is a limited amount of time allocated to finding the best route).

Please take a look.

\<thinks\>Maybe A* is possible\</thinks\>

### DIFF
--- a/graph/path/bellman_ford_moore.go
+++ b/graph/path/bellman_ford_moore.go
@@ -7,17 +7,32 @@ package path
 import (
 	"gonum.org/v1/gonum/graph"
 	"gonum.org/v1/gonum/graph/internal/linear"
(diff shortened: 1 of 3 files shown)


kortschak:
I recommend hiding white space for viewing the diffs in the test code.

kortschak:
*ping*

vladimir-ch:
Apologies for the silence, I'll try to take a look soon.

sbinet (approved):
nothing jumped at me.

but we should probably wait for Vladimir's review.

vladimir-ch (approved):

Issue #1513, opened 2020-11-26
spatial/{r2,r3}: provide a way to rotate a vector


sbinet:
### What are you trying to do?

it should be possible to rotate a, say, `r3.Vec` by some angle around a given axis (x,y,z) or around a user provided `r3.Vec`.

a possible API (for `r3`, `r2` is expected to be rather similar):
```go
package r3

func (p Vec) Rotate(angle float64, q Vec) Vec { ... }
```

we could also imagine providing a way to rotate a vector via rotation matrix:
```go
package r3

func Rotate(m *mat.Dense, p Vec) Vec { ... }
```

API to be discussed.

kortschak:
Rotating around a 3 vector as shown in the first case is best done with a quaternion (lightly tested):
```
func (p Vec) Rotate(alpha float64, axis Vec) Vec {
	alpha /= 2
	q := raise(axis)
	q = quat.Scale(math.Sin(alpha)/quat.Abs(q), q)
	q.Real += math.Cos(alpha)
	if len := quat.Abs(q); len != 1 {
		q = quat.Scale(1/len, q)
	}
	pp := quat.Mul(quat.Mul(q, raise(p)), quat.Conj(q))
	return Vec{X: pp.Imag, Y: pp.Jmag, Z: pp.Kmag}
}

func raise(p Vec) quat.Number {
	return quat.Number{Imag: p.X, Jmag: p.Y, Kmag: p.Z}
}
```

R2 is trivial.

The API presented matches what we have (I recall an issue to discuss this, but I can't find it now).

sbinet:
one could also expose a new type `Rotation`:
```go
type Rotation struct {
    q quat.Number
}

func NewRotation(alpha float64, axis Vec) Rotation {
 // lift from Vec.Rotate
}

func (rot Rotation) Rotate(p Vec) Vec {
    pp := quat.Mul(quat.Mul(rot.q, raise(p)), quat.Conj(rot.q))
    return Vec{X: pp.Imag, Y: pp.Jmag, Z: pp.Kmag}
}
```

to be able to repeatedly apply a 3d rotation, w/o incurring the cost of re-computing the same sin/cos values.

kortschak:
That's probably a reasonably middle ground between a pure function and being able to do all the things that you can with quaternions in their pure form.

PR #1516, opened 2020-11-27
spatial/r3: implement Vec.Rotate

sbinet:
Updates gonum/gonum#1513.

Please take a look.


sbinet:
I took the executive decision of handling the special case of axis==(0,0,0) as the identity.
we could of course just panic. (I am on the fence on this)

spatial/r3/vector.go
```diff
@@ -49,6 +53,29 @@ func (p Vec) Cross(q Vec) Vec {
 	}
 }
 
+// Rotate returns a new vector, rotated by alpha around the provided axis.
+func (p Vec) Rotate(alpha float64, axis Vec) Vec {
+	if alpha == 0 || (axis == Vec{}) {
```
kortschak, line 58: This just explodes into NaN if axis is zero. I think that's OK. Maybe leaving the zero test for alpha though.

sbinet: ok.

sbinet:
PTAL (added some possible TODOs for later. I don't have the need for them right now, but I suspect they could be useful for some)

spatial/r3/rotation.go
```diff
@@ -0,0 +1,43 @@
...
+// TODO:
+//  - create rotations from Euler angles (NewRotationFromEuler?)
+//  - create rotations from rotation matrices (NewRotationFromMatrix?)
+//  - return the equivalent Euler angles from a Rotation
+//  - return the equivalent rotation matrix from a Rotation
+
+// Rotation describes a rotation in space.
+type Rotation struct {
+	q quat.Number
+}
```
kortschak, line 22: If this is `type Rotation quat.Number` then the system is more flexible and allows the user to do scaling and so on as well.

spatial/r3/rotation.go
```diff
@@ -0,0 +1,43 @@
...
+
+// NewRotation creates a rotation by alpha, around axis.
+func NewRotation(alpha float64, axis Vec) Rotation {
+	var (
```
kortschak, line 26: Do you want to have the `alpha == 0` case return the identity Q independent of `axis`? It would be surprising to have the two different rotations approaches give different answers for this case.

spatial/r3/rotation.go
```diff
@@ -0,0 +1,43 @@
...
+	q.Real += cos
+	if len := quat.Abs(q); len != 1 {
+		q = quat.Scale(1/len, q)
+	}
+
+	return Rotation{q: q}
+}
+
+// Rotate returns the rotated vector according to the definition of rot.
+func (rot Rotation) Rotate(p Vec) Vec {
```
kortschak, line 40: s/rot/r/g

spatial/r3/rotation.go
```diff
@@ -0,0 +1,43 @@
+// Copyright ©2020 The Gonum Authors. All rights reserved.
```
kortschak, line 1: I think this belongs in vector.go.

spatial/r3/rotation.go
```diff
@@ -0,0 +1,43 @@
...
+// Rotation describes a rotation in space.
+type Rotation struct {
+	q quat.Number
+}
+
+// NewRotation creates a rotation by alpha, around axis.
+func NewRotation(alpha float64, axis Vec) Rotation {
+	var (
+		q        = raise(axis)
+		sin, cos = math.Sincos(0.5 * alpha)
```
kortschak, line 28: Very minor aesthetic thing, but I prefer the gopher operator for these. Up to you.

kortschak:
Given that we're tying r3 to quat now, I think it probably does justify the change in API to be more consistent between the two. Future PR?

spatial/r3/rotation.go
sbinet: you mean:
```go
q := raise(axis)
sin, cos := math.Sincos(0.5*alpha)
```

spatial/r3/rotation.go
sbinet: you mean move everything that is inside `rotation.go` over to `vector.go`?

spatial/r3/rotation.go
kortschak: Yep

spatial/r3/rotation.go
sbinet: SGTM.

spatial/r3/rotation.go
sbinet: SGTM.

spatial/r3/rotation.go
kortschak: Yep

sbinet:
yeah, let's leave the API revamp for another PR :)

sbinet:
PTAL

sbinet:
ping?
(I have the same kind of PR in queue for `r2`, then I'll tackle the API revamp)

kortschak:
I haven't forgotten, just been working on rdf and dot (among otherthings). Hopefully later today.

spatial/r3/vector.go
```diff
@@ -49,6 +53,15 @@ func (p Vec) Cross(q Vec) Vec {
...
 }
 
+// Rotate returns a new vector, rotated by alpha around the provided axis.
+func (p Vec) Rotate(alpha float64, axis Vec) Vec {
+	return NewRotation(alpha, axis).Rotate(p)
+}
+
+func raise(p Vec) quat.Number {
+	return quat.Number{Imag: p.X, Jmag: p.Y, Kmag: p.Z}
+}
```
kortschak, line 63: This should come after its use.

spatial/r3/vector.go
```diff
@@ -79,3 +92,46 @@ func Cos(p, q Vec) float64 {
...
+	}
+	qq := quat.Number(r)
+	pp := quat.Mul(quat.Mul(qq, raise(p)), quat.Conj(qq))
+	return Vec{X: pp.Imag, Y: pp.Jmag, Z: pp.Kmag}
+}
+
+func (r Rotation) isIdentity() bool {
+	return quat.Number(r) == quat.Number{
+		Real: 1, Imag: 1, Jmag: 1, Kmag: 1,
+	}
```
kortschak, line 136: Quaternion identity is `quat.Number{Real: 1}`. (Note that all rotation-only quaternions must have an absolute value of 1).

`return r == Rotation{Real: 1}`

spatial/r3/vector.go
```diff
@@ -79,3 +92,46 @@ func Cos(p, q Vec) float64 {
 type Box struct {
 	Min, Max Vec
 }
+
+// TODO:
+//  - create rotations from Euler angles (NewRotationFromEuler?)
+//  - create rotations from rotation matrices (NewRotationFromMatrix?)
+//  - return the equivalent Euler angles from a Rotation
+//  - return the equivalent rotation matrix from a Rotation
```
kortschak, line 100: I am ambivalent about including Euler angles from quaternions. I would be OK with adding an example for how to do it, but it's really a bad approach to handling spatial rotation and I don't really want to encourage it.

spatial/r3/vector.go
```diff
@@ -79,3 +92,46 @@ func Cos(p, q Vec) float64 {
...
+
+// Rotation describes a rotation in space.
+type Rotation quat.Number
+
+// NewRotation creates a rotation by alpha, around axis.
+func NewRotation(alpha float64, axis Vec) Rotation {
+	if alpha == 0 {
+		return Rotation(quat.Number{
+			Real: 1, Imag: 1, Jmag: 1, Kmag: 1,
+		})
```
kortschak, line 110: `return Rotation{Real: 1}`

spatial/r3/vector_test.go
```diff
@@ -236,6 +236,34 @@ func TestCos(t *testing.T) {
...
+		{Vec{1, 0, 0}, Vec{0, 0, 0}, math.Pi / 2, Vec{math.NaN(), math.NaN(), math.NaN()}},
+		{Vec{1, 0, 0}, Vec{0, 1, 0}, math.Pi / 2, Vec{0, 0, -1}},
+		{Vec{1, 0, 0}, Vec{0, 1, 0}, math.Pi, Vec{-1, 0, 0}},
+		{Vec{2, 0, 0}, Vec{0, 1, 0}, math.Pi, Vec{-2, 0, 0}},
+		{Vec{1, 2, 3}, Vec{1, 1, 1}, 2. / 3. * math.Pi, Vec{3, 1, 2}},
+	} {
+		t.Run("", func(t *testing.T) {
+			got := test.v.Rotate(test.alpha, test.axis)
+			if !vecApproxEqual(got, test.want, tol) {
+				t.Fatalf(
```
kortschak, line 258: This should be an `Errorf`.

sbinet: It's in a subtest so it's ok to use Fatalf, isn't it?

kortschak: Does it need to me a subtest?

sbinet: consistency (with all the other `r3` and `r2` tests)?
I find it easier to run/test a particular test case with sub tests.

but if you prefer "flat" tests, I won't insist.

spatial/r3/vector.go
sbinet: I'll reword so as to not sound as a "hard" TODO but merely as a set of possible suggestions.

sbinet:
PTAL

spatial/r3/vector_test.go
kortschak: I don't really see the point of them as subtests; they're not named and there are not groups of tests there. Deleting the `t.Run("", func(t *testing.T) {` and matching `})` leaves the syntax almost identical, but simpler.

sbinet: ok. do you want me to go in `r3` and "flatten" other `r3.Vec` tests as well? in this PR?

kortschak: I think in another PR.

sbinet: ok. done w/ PS6

kortschak: thanks

kortschak (approved):

PR #1518, opened 2020-11-28
graph/encoding/dot: fix line visit marking

### LINKED ISSUE #1517: graph/encoding/dot: MarshalMulti incorrectly marks lines as visited
The logic for marking visited lines does not take into account directionality of the graph and incorrectly assumes that line IDs are unique within a multigraph. While our implementations are implemented this way, the documentation merely states that "A Line returns an ID that must distinguish Lines sharing Node end points."

I have a fix, but I need to write a test for it.

Since v0.8.2 is not yet tagged, I'd like to get this fix into that release.

kortschak:
Please take a look.

Fixes #1517.

### DIFF
--- a/graph/encoding/dot/encode.go
+++ b/graph/encoding/dot/encode.go
@@ -210,16 +210,17 @@ func (p *simpleGraphPrinter) print(g graph.Graph, name string, needsIndent, isSu
 		sort.Sort(ordered.ByID(to))
 		for _, t := range to {
 			tid := t.ID()
+			f := edge{inGraph: name, from: nid, to: tid}
 			if isDirected {
-				if p.visited[edge{inGraph: name, from: nid, to: tid}] {
+				if p.visited[f] {
 					continue
 				}
-				p.visited[edge{inGraph: name, from: nid, to: tid}] = true
+				p.visited[f] = true
 			} else {
-				if p.visited[edge{inGraph: name, from: nid, to: tid}] {
+				if p.visited[f] {
 					continue
 				}
-				p.visited[edge{inGraph: name, from: nid, to: tid}] = true
+				p.visited[f] = true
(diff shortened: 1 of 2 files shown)


kortschak:
*ping*

graph/encoding/dot/encode.go
```diff
@@ -510,10 +512,18 @@ func (p *multiGraphPrinter) print(g graph.Multigraph, name string, needsIndent,
 
 			for _, l := range lines {
 				lid := l.ID()
-				if p.visited[line{inGraph: name, id: lid}] {
-					continue
+				if isDirected {
+					if p.visited[line{inGraph: name, from: nid, to: tid, id: lid}] {
+						continue
+					}
+					p.visited[line{inGraph: name, from: nid, to: tid, id: lid}] = true
```
sbinet, line 519: perhaps construct that key once (before the `if` at line 516) and use in both places?

graph/encoding/dot/encode.go
```diff
@@ -510,10 +512,18 @@ func (p *multiGraphPrinter) print(g graph.Multigraph, name string, needsIndent,
...
+				} else {
+					if p.visited[line{inGraph: name, from: nid, to: tid, id: lid}] {
```
sbinet, line 521: ditto.

sbinet (dismissed):
LGTM

(but I am no expert)

graph/encoding/dot/encode.go
kortschak: Done

graph/encoding/dot/encode.go
kortschak: Done

kortschak:
PTAL

sbinet (approved):

PR #1519, opened 2020-12-01
graph/formats/rdf: new package for RDF N-Quad parsing

kortschak:
This includes a commit that is part of #1518 but which will be rebased away on merge. That fix is necessary for the example in this PR (where it was found).

Please take a look.

@chewxy @owulveryck Please try this in your setting to see if it fits what you'd be using it for. Review comments are welcome and will help Gonum reviewers get this reviewed quicker. (Also note that this is readily vendored being self-contained in the package).


kortschak:
Linting fails due to the generated code. I tried specifying nolint, but the linter gets confused with the `//line` directives.

kortschak:
Ubuntu dist is updated to bionic to match ragel version.

graph/formats/rdf/rdf.go
```diff
@@ -0,0 +1,386 @@
...
+	if ok {
+		return id
+	}
+	id = int64(len(dec.ids))
+	dec.ids[s] = id
+	return id
+}
+
+// Terms returns the mapping between terms and graph node IDs constructed
+// during decoding the RDF statement stream.
```
kortschak, line 267: This and the comment on `Unmarshal` should note that the IDs are keyed on the unescaped quoted node text as it appears in the original statement stream (reflected in the contents of the `Term.Value` field and no folding is performed.

graph/formats/rdf/rdf.go
```diff
@@ -0,0 +1,386 @@
...
+func (s *Statement) ID() int64 { return s.Predicate.UID }
+
+// ReversedEdge returns a graph.Edge with the subject and object of the
+// statement swapped. It is implemented to satisfy the graph.Edge
+// interface.
+func (s *Statement) ReversedEdge() graph.Edge {
+	r := *s
+	r.Subject, r.Object = r.Object, r.Subject
+	return &r
+}
```
kortschak, line 173: I'm having doubts about the behaviour of this and `ReversedLine`. ISTM given the absence of intrinsic meaning to the reversal of an RDF statement (without domain-specific knowledge) that this should return the receiver.

If the edge is not returned with the ends swapped, the documentation for `graph.Edge` should be altered so that the `ReversedEdge` method has:
```
    // ReversedEdge returns the edge reversal of the receiver
    // if a reversal is valid for the data type.
    // When a reversal is valid an edge of the same type as
    // the receiver with nodes of the receiver swapped should
    // be returned, otherwise the receiver should be returned
    // unaltered.
```

with the equivalent change for `graph.Line`.

kortschak:
I am looking into the licensing of the test data. There are license links on the pages, but they both 404. I've asked the W3C webmaster about this and they have forwarded that on to the content owner.

graph/formats/rdf/rdf.go
kortschak: Having had some time to think about this, I think this is the correct thing to do. It has virtually no impact on tests within the graph packages (changing simple and multi edges and lines to return only the unreversed edge/line - to simulate a user using a graph that follows now-allowed semantics - causes failures in simple as expected since they test for this specific behaviour (not in multi - I should fix that), and in `graph.Copy` which makes sense given what the test is testing).

I've added an example that demonstrated how to make use of this feature.

kortschak:
@dennwc Are you available to provide an external review of this purely for correctness?

I have added semantic checks to IRI parser compared to the original cayley parser, so this parser now passes the RDF Working Group test suite in its entirety. I'd ba happy to contribute those changes to the cayley project if they are wanted.

graph/formats/rdf/rdf.go
```diff
@@ -0,0 +1,373 @@
...
+
+func (dec *Decoder) idFor(s string) int64 {
+	id, ok := dec.ids[s]
+	if ok {
+		return id
+	}
+	id = int64(len(dec.ids))
+	dec.ids[s] = id
+	return id
+}
```
kortschak, line 257: IDs should be assigned from 1 rather than 0 so that graph containers can distinguish already ID-assigned terms from terms that have been manually constructed without an ID.

chewxy:
I have gone thru this. It looks good.

kortschak:
Thanks for doing that @chewxy.

kortschak:
I'd like to get this and its partner (#1524) into the next release if that's possible. Is someone available to review this?

graph/formats/rdf/rdf.go
```diff
@@ -0,0 +1,384 @@
...
+
+// Terms returns the mapping between terms and graph node IDs constructed
+// during decoding the RDF statement stream.
+func (dec *Decoder) Terms() map[string]int64 {
+	return dec.ids
+}
+
+// store is a string internment implementation.
+type store map[string]string
```
kortschak, line 271: This is not the best approach to string interning. I am waiting to see what happens in the json package and the approach that ends up being used there and will probably base the approach on that.

vladimir-ch (approved):

graph/formats/rdf/rdf.go
```diff
@@ -0,0 +1,384 @@
...
+	if dec.ids == nil {
+		dec.ids = make(map[string]int64)
+	}
+}
+
+// Unmarshal returns the next statement from the input stream.
+func (dec *Decoder) Unmarshal() (*Statement, error) {
+	for dec.scanner.Scan() {
+		data := bytes.TrimSpace(dec.scanner.Bytes())
+		if len(data) != 0 && data[0] != '#' {
```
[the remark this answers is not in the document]
kortschak: Worth doing. Thanks

kortschak:
PTAL

vladimir-ch (approved):

PR #1522, opened 2020-12-02
graph/path: simplify and clean up union-find code and Kruskal

kortschak:
Please take a look.

### DIFF
--- a/graph/path/disjoint.go
+++ b/graph/path/disjoint.go
@@ -4,84 +4,53 @@
 
 package path
 
-// A disjoint set is a collection of non-overlapping sets. That is, for any two sets in the
-// disjoint set, their intersection is the empty set.
-//
-// A disjoint set has three principle operations: Make Set, Find, and Union.
-//
-// Make set creates a new set for an element (presuming it does not already exist in any set in
-// the disjoint set), Find finds the set containing that element (if any), and Union merges two
-// sets in the disjoint set. In general, algorithms operating on disjoint sets are "union-find"
-// algorithms, where two sets are found with Find, and then joined with Union.
-//
-// A concrete example of a union-find algorithm can be found as discrete.Kruskal -- which unions
-// two sets when an edge is created between two vertices, and refuses to make an edge between two
-// vertices if they're part of the same set.
-type disjointSet struct {
-	master map[int64]*disjointSetNode
-}
+// djSet implement a disjoint set finder using the union-find algorithm.
+type djSet map[int64]*dsNode
 
-type disjointSetNode struct {
-	parent *disjointSetNode
-	rank   int
-}
-
-func newDisjointSet() *disjointSet {
-	return &disjointSet{master: make(map[int64]*disjointSetNode)}
-}
-
-// If the element isn't already somewhere in there, adds it to the master set and its own tiny set.
-func (ds *disjointSet) makeSet(e int64) {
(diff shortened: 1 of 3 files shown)


graph/path/disjoint.go
```diff
@@ -4,84 +4,53 @@
...
-// sets in the disjoint set. In general, algorithms operating on disjoint sets are "union-find"
-// algorithms, where two sets are found with Find, and then joined with Union.
-//
-// A concrete example of a union-find algorithm can be found as discrete.Kruskal -- which unions
-// two sets when an edge is created between two vertices, and refuses to make an edge between two
-// vertices if they're part of the same set.
-type disjointSet struct {
-	master map[int64]*disjointSetNode
-}
+// djSet implement a disjoint set finder using the union-find algorithm.
```
vladimir-ch, line 7: nit: implements

graph/path/disjoint.go
```diff
@@ -4,84 +4,53 @@
...
+type djSet map[int64]*dsNode
 
-type disjointSetNode struct {
-	parent *disjointSetNode
-	rank   int
-}
-
-func newDisjointSet() *disjointSet {
-	return &disjointSet{master: make(map[int64]*disjointSetNode)}
+// add adds e to the collections of sets held by the disjoint set.
```
vladimir-ch, line 10: nit: collection (?)

kortschak:
Supposedly compression and/or recursion in union-find improve performance, but benchmarking that I've done shows worse performance than the naive implementation.

graph/path/disjoint.go
```diff
@@ -4,84 +4,53 @@
...
+		return
+	}
+	if ra.rank < rb.rank {
+		ra.parent = rb
 		return
 	}
-	dsNode := &disjointSetNode{rank: 0}
-	dsNode.parent = dsNode
-	ds.master[e] = dsNode
+	rb.parent = ra
```
vladimir-ch, line 30: Can you swap ra and rb here so that this part is congruent with the if block above? Then in both cases ra.parent is set and rb.rank is incremented and therefore the greater of the two.

graph/path/disjoint.go
kortschak: fixed

graph/path/disjoint.go
kortschak: fixed

graph/path/disjoint.go
kortschak: I'm not sure what you mean. If I change it that way
```
func (djSet) union(a, b *dsNode) {
	ra := find(a)
	rb := find(b)
	if ra == rb {
		return
	}
	if ra.rank < rb.rank {
		ra.parent = rb
		return
	}
	ra.parent = rb
	if rb.rank == ra.rank {
		ra.rank++
	}
}
```
the algorithm is not correct. This changes passes in the test here, but they are small; on data sets I have here it fails to terminate. It does work with the recursive approach be cause that updates the parent pointer, but it defeats the chain length optimisation and is approximately half as fast.

Have I missed what you mean?

I will change the order of the equality test for aesthetics.

vladimir-ch: I don't know what I was thinking, you're of course right that it is not correct.

vladimir-ch (approved):

PR #1526, opened 2020-12-13
github: add CI action

kortschak:
Please take a look.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,83 @@
+name: CI
+
+on:
+  pull_request:
+    branches: [ master ]
+
+jobs:
+
+  build:
+    name: Build
+    strategy:
+      matrix:
+        go-version: [1.15.x, 1.14.x]
+        platform: [ubuntu-latest, macos-latest]
+        env:
+          - TAGS=""
+          - TAGS="-tags bounds"
+          - TAGS="-tags noasm"
+          - TAGS="-tags safe"
+          - FORCE_GOARCH=386
+        exclude:
+          - platform: macos-latest
+            env: TAGS="-tags bounds"
+          - platform: macos-latest
+            env: TAGS="-tags noasm"
+          - platform: macos-latest
+            env: TAGS="-tags safe"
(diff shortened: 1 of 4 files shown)


kortschak:
I have disabled the requirement for travis build success.

kortschak:
I have tried to make Coveralls work with this, but it's pretty resistant. I think maybe we should drop it and remove the badge.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,83 @@
...
+        restore-keys: |
+          ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
+
+    - name: Checkout code
+      uses: actions/checkout@v2
+      with:
+          path: ${{ env.GOPATH }}/src/gonum.org/v1/gonum
+
+    - name: Check copyrights+imports+formatting+generate
+      if: matrix.platform == 'ubuntu-latest' && matrix.tags == ''
```
sbinet, line 64: I guess this relies on the assumption that the test variants are executed in the same order than the `matrix/env` list.
I guess that's a rather solid assumption.

sbinet (approved):
LGTM.

nice use of the matrix/exclude feature.

I think it's OK to either remove the coveralls badge or leave it as part of the travis pipeline.

sbinet:
OTOH, using only one `matrix/env` (`TAGS: ""`) and handling the `ubuntu` build+test case specially (ie: manually building+testing the matrix with all variants of `TAGS`) would perhaps speedup the pipeline (and save resources).

.github/workflows/ci.yml
kortschak: It would be a very surprising behaviour if they didn't, it maps to an array. We'll know if this doesn't work at any point and we can fix it them.

kortschak:
The single variant with all tags was how we used to have it, it took too long.

kortschak:
I'll leave the Coveralls badge for the moment and see what happens with travis. If we ditch travis, we can get rid of Coveralls (maybe lose it anyway).

PR #1527, opened 2020-12-13
codecov: make coverage report relative to root of the repo

kortschak:
The reports can't link to the source code because that have `src/gonum.org/v1/gonum/` prefix. This should fix that.

Please take a look.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,0 +1,2 @@
+fixes:
+  - "src/gonum.org/v1/gonum/::"


sbinet (dismissed):

kortschak:
No, this didn't work. I'll play around locally and see what's going on. I still reports them with the src/gon... prefix.

kortschak:
OK. This is working now. The same change is needed in gonum/plot.

sbinet (approved):

PR #1531, opened 2020-12-22
graph/set/uid: make uid package publicly available

kortschak:
Please take a look.

This change makes it easier for users to copy/paste and alter/specialise the graphs in simple and multi.


graph/set/uid/uid.go
```diff
@@ -27,10 +31,10 @@ func (s *Set) NewID() int64 {
 	for id := range s.free {
 		return id
 	}
-	if s.maxID != Max {
+	if s.maxID != math.MaxInt64 {
```
vladimir-ch, line 34: Why not keep `Max` here?

kortschak: It seems redundant; IDs are `int64` and there is already a label for the maximum `int64` value.

vladimir-ch (approved):

PR #1533, opened 2020-12-28
ci: add action for uploading coverage report on PR merge

kortschak:
We have been missing base reports since the migration from travis since GitHub Actions (per default) don't run on merge. This and the rebase/squash merge policy that we use means that the reports that have been uploaded to Codecov are never relevant to future reports since they are always off the master branch. This does a run of coverage only when a PR is closed and the closure was a merge. It only does this for the tags and only for ubuntu, but the infrastructure for extension is left in in case we want to extend to macOS later.

No change will be visible in the pre-merge stage of this PR (unfortunately).

Please take a look.

### DIFF
--- a/.github/workflows/codecov.yml
+++ b/.github/workflows/codecov.yml
@@ -1,0 +1,63 @@
+name: Codecov
+
+on:
+  pull_request:
+    branches: [ master ]
+    types:
+      - closed
+
+jobs:
+
+  build:
+    name: Build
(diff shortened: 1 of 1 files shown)


sbinet (approved):

kortschak:
Here goes nothing...

kortschak:
Didn't quite work, but we are closer I think.

PR #1534, opened 2020-12-29
ci: lie to codecov action about the origin of the report

kortschak:
This is another leap into the darkness.

Please take a look.


sbinet (approved):

kortschak:
This did not work.

PR #1535, opened 2021-01-02
mat/cdense: Add Caps() and Slice()

notmgsk:


kortschak:
Thanks for sending this.

We will need to have tests for the new code. This is just a matter of copying `TestDenseGrow` in "dense_test.go" to its cognate in "cdense_test.go". This will require that `CDense.Grow` is added. There is also room for adding `TestCDenseOverlaps` in "shadow_test.go", but I'm happy to do that later if you don't want to do it.

mat/cdense_test.go
```diff
@@ -47,3 +47,59 @@ func TestCDenseNewAtSet(t *testing.T) {
 		}
 	}
 }
+
+func TestCDenseGrow(t *testing.T) {
+	t.Parallel()
+	m := &Dense{}
```
kortschak, line 53: This will need to be `m := &CDense{}`.

mat/cdense_test.go
```diff
@@ -47,3 +47,59 @@ func TestCDenseNewAtSet(t *testing.T) {
...
+	m = m.Grow(10, 10).(*Dense)
```
kortschak, line 54: So here and below mean that `.Grow` needs to be implemented.

kortschak (approved):
LGTM, thanks and welcome.

The next step is for you to send a PR adding yourself(/employer if relevant) to AUTHORS and CONTRIBUTORS with the commit message "A+C: add Mark Skilbeck". When that's done, I'll merge both of the PRs.

kortschak:
I just want to ask anyone else whether they are concerned about having `Grow` and  `Slice` not be prefixed with `C`. There is close to zero chance of interface collision, and there already would be with `At` and `Set`, so I think that concern ship has sailed, but it's worth asking now.

notmgsk:
@kortschak should I fix the commit messages to use the prefix "mat:" rather than "mat/cdense:"?

kortschak:
You can if you want. If you do, it would be a single commit, "mat: add Caps, Grow and Slice methods to CDense", but I can do it as a squash in merge. It's probably easier to leave for that. rather than re-run all the tests.

notmgsk:
> You can if you want. If you do, it would be a single commit, "mat: add Caps, Grow and Slice methods to CDense", but I can do it as a squash in merge. It's probably easier to leave for that rather than re-run all the tests.

Agreed, I'll leave it for you :)

vladimir-ch (approved):
> I just want to ask anyone else whether they are concerned about having Grow and Slice not be prefixed with C.

I'm not concerned. LGTM.

Welcome and thanks, @notmgsk

PR #1537, opened 2021-01-03
Cdense arithmetic

notmgsk:
Very much a work-in-progress. Making it visible here for early commentary so that I don't go too far down a wrong path.

Note: this diff includes the changes in #1535 but once those are in master I'll remove them to clean the diff.


kortschak:
The pool code provides temporary workspaces for intermediate calculations and to allow update operations on receivers that would not be possible otherwise, e.g. `m.Mul(m, b)`. The pool for each type is a set `sync.Pool` that return exponentially sized (wrt backing slice cap) slices/matrices. When a workspace is needed for type `T`, `getWorkspaceT` is called which gets a `T` value, slices it to the right size and returns it and a closure that puts it back into the pool with `putWorkspaceT` when called. All this is done to reduce garbage collection pressure. It can be got wrong with bad consequences.

There are parts of the work here that will depend on the addition of the non-conjugate transpose infrastructure, so it's worth waiting to see about that proposal (I have the draft ready, just waiting for use case examples). Some of the untranspose/unconjugate logic is going to be interesting and may need some hashing out.

mat/cdense_arithmetic.go
```diff
@@ -0,0 +1,376 @@
...
+			}
+			for ja, jb, jm := 0, 0, 0; ja < ar*amat.Stride; ja, jb, jm = ja+amat.Stride, jb+bmat.Stride, jm+m.mat.Stride {
+				for i, v := range amat.Data[ja : ja+ac] {
+					m.mat.Data[i+jm] = v + bmat.Data[i+jb]
+				}
+			}
+			return
+		}
+	}
+}
```
kortschak, line 46: This would have something like the following as the non-fast path:

```
	m.checkOverlapMatrix(aU)
	m.checkOverlapMatrix(bU)
	var restore func()
	if aTrans != aConj && m == aU {
		m, restore = m.isolatedWorkspace(aU)
		defer restore()
	} else if bTrans != bConj && m == bU {
		m, restore = m.isolatedWorkspace(bU)
		defer restore()
	}

	for r := 0; r < ar; r++ {
		for c := 0; c < ac; c++ {
			m.set(r, c, a.At(r, c)+b.At(r, c))
		}
	}
```

(with `aTrans`, `aConj`, `bTrans` and `bConj` coming from the relevant `untransposeCmplx` call).

The other element-wise operations will have the analogous addition.

What we're doing here is making sure that if the destination overlaps with the src and there is a transposition that we make a temporary work place to put the result in when then gets copied back into the destination. Otherwise the partial work will clobber the source.

mat/pool.go
```diff
@@ -63,6 +63,9 @@ var (
 
 	// poolCmplx is the CDense equivalent of pool.
 	poolCmplx [63]sync.Pool
+
+	// TODO(notmgsk): wtf
+	poolZ [63]sync.Pool
```
kortschak, line 68: `poolComplexes` I think.

mat/pool.go
```diff
@@ -63,6 +63,9 @@ var (
 
 	// poolCmplx is the CDense equivalent of pool.
 	poolCmplx [63]sync.Pool
+
+	// TODO(notmgsk): wtf
```
kortschak, line 67: The magic here is that there are a collection of exponentially sized sources of temporary work slices. This means that when we need a temporary slice, we ask for one and we know that it will be not more that 2*n-1 the length of what we need. We do this to avoid allocation of temporary slices.

kortschak (commented):
Hopefully helpful comments.

mat/cdense_arithmetic.go
notmgsk: @kortschak I'm not sure I understand this pattern. It looks like the fast path includes type assertions (i.e. `arm, ok := a.(*CDense); ok { /* CDense methods are now available */ }`. In what case would you fall through into the slow path? The input is, by definition, always going to be a `CMatrix` which, if I understand correctly, will always satisfy the assertion `a.(*CDense)`?

kortschak: It's true that `*CDense` is only only `CMatrix` implementation that we *currently* provide, but there will be others. There may be other implementations in the wild (unlikely, but possible). The design of the package is that we accept the interface, not the implementation.

mat/cdense_arithmetic.go
```diff
@@ -0,0 +1,414 @@
...
+	m.checkOverlapMatrix(aU)
+	m.checkOverlapMatrix(bU)
+	var restore func()
+	if aTrans && m == aU {
+		m, restore = m.isolatedWorkspace(aU)
+		defer restore()
+	} else if bTrans && m == bU {
+		m, restore = m.isolatedWorkspace(bU)
+		defer restore()
+	}
```
kortschak, line 53: These are not correct; if `a` is `m.T().H()` or `m.H().T()` then is is safe to write from `a` directly back to `m`. Equally, if `a` is `m.H()`, it's not. The approach here doesn't handle these cases, but the code in the comment below does; it's safe when the elements are transposed an even number of times (both H and T or neither).

PR #1538, opened 2021-01-04
ci: use new codecov action API to fix coverage report issues

kortschak:
Hopefully this is going to work. Again, we won't know until it's merged.

The [new API](https://github.com/codecov/codecov-action#arguments) was added at least in part to address the issue we have been having, so thanks to Codecov for that.

Please take a look.


sbinet (approved):

kortschak:
And.... it still doesn't work.

PR #1541, opened 2021-01-05
mat: add support for tridiagonal matrices

vladimir-ch:
After a short hiatus I'm trying to pick up my effort to support tridiagonal matrices in `mat`. At the moment this is just a draft because there are not tests yet but please take a look and leave comments.

### DIFF
--- a/mat/index_bound_checks.go
+++ b/mat/index_bound_checks.go
@@ -346,3 +346,52 @@ func (d *DiagDense) setDiag(i int, v float64) {
 	}
 	d.mat.Data[i*d.mat.Inc] = v
 }
+
+// At returns the element at row i, column j.
+func (a *Tridiag) At(i, j int) float64 {
+	return a.at(i, j)
+}
+
+func (a *Tridiag) at(i, j int) float64 {
+	if uint(i) >= uint(d.mat.N) {
+		panic(ErrRowAccess)
+	}
+	if uint(j) >= uint(d.mat.N) {
+		panic(ErrColAccess)
+	}
+	switch i - j {
+	case -1:
+		return a.mat.DU[i]
+	case 0:
+		return a.mat.D[i]
+	case 1:
+		return a.mat.DL[j]
+	default:
+		return 0
+	}
+}
+
+// SetBand sets the element at row i, column j to the value v.
+// It panics if the location is outside the appropriate region of the matrix.
+func (a *Tridiag) SetBand(i, j int, v float64) {
+	a.set(i, j, v)
+}
+
+func (a *Tridiag) set(i, j int, v float64) {
+	if uint(i) >= uint(a.mat.N) {
+		panic(ErrRowAccess)
+	}
+	if uint(j) >= uint(a.mat.N) {
+		panic(ErrColAccess)
+	}
+	switch i - j {
+	case -1:
+		a.mat.DU[i] = v
+	case 0:
+		a.mat.D[i] = v
+	case 1:
+		a.mat.DL[j] = v
+	default:
+		panic(ErrBandSet)
+	}
+}
--- a/mat/tridiag.go
+++ b/mat/tridiag.go
@@ -1,0 +1,397 @@
+// Copyright ©2020 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package mat
+
+import (
+	"math"
+
+	"gonum.org/v1/gonum/blas"
+	"gonum.org/v1/gonum/blas/blas64"
+	"gonum.org/v1/gonum/internal/asm/f64"
+	"gonum.org/v1/gonum/lapack/lapack64"
+)
+
+var (
+	tridiagDense *Tridiag
+	_            Matrix           = tridiagDense
+	_            allMatrix        = tridiagDense
+	_            denseMatrix      = tridiagDense
+	_            Banded           = tridiagDense
+	_            MutableBanded    = tridiagDense
+	_            RawTridiagonaler = tridiagDense
+)
+
+// A RawTridiagonaler can return a lapack64.Tridiagonal representation of the
+// receiver. Changes to the elements of DL, D, DU in lapack64.Tridiagonal will
+// be reflected in the original matrix, changes to the N field will not.
+type RawTridiagonaler interface {
+	RawTridiagonal() lapack64.Tridiagonal
+}
+
+// Tridiag represents a tridiagonal matrix by its three diagonals.
+type Tridiag struct {
+	mat lapack64.Tridiagonal
+}
+
+// NewTridiag creates a new n×n tridiagonal matrix with the first sub-diagonal
+// in dl, the main diagonal in d and the first super-diagonal in du. If all of
+// dl, d, and du are nil, new backing slices will be allocated for them. If dl
+// and du have length n-1 and d has length n, they will be used as backing
+// slices, and changes to the elements of the returned Tridiag will be reflected
+// in dl, d, du. If neither of these is true, NewTridiag will panic.
+func NewTridiag(n int, dl, d, du []float64) *Tridiag {
+	if n <= 0 {
+		if n == 0 {
+			panic(ErrZeroLength)
+		}
+		panic(ErrNegativeDimension)
+	}
+	if dl != nil || d != nil || du != nil {
+		if len(dl) != n-1 || len(d) != n || len(du) != n-1 {
+			panic(ErrShape)
+		}
+	} else {
+		d = make([]float64, n)
+		if n > 1 {
+			dl = make([]float64, n-1)
+			du = make([]float64, n-1)
+		}
+	}
+	return &Tridiag{
+		mat: lapack64.Tridiagonal{
+			N:  n,
+			DL: dl,
+			D:  d,
+			DU: du,
+		},
+	}
+}
+
+// Dims returns the number of rows and columns in the matrix.
+func (a *Tridiag) Dims() (r, c int) {
+	return a.mat.N, a.mat.N
+}
+
+// Bandwidth returns 1, 1 - the upper and lower bandwidths of the matrix.
+func (a *Tridiag) Bandwidth() (kl, ku int) {
+	return 1, 1
+}
+
+// T performs an implicit transpose by returning the receiver inside a Transpose.
+func (a *Tridiag) T() Matrix {
+	// An alternative would be to return the receiver with DL,DU swapped; the
+	// untranspose function would then always return false. With Transpose the
+	// diagonal swapping will be done in tridiagonal routines in lapack like
+	// lapack64.Gtsv or gonum.Dlagtm based on the trans parameter.
+	return Transpose{a}
+}
+
+// TBand performs an implicit transpose by returning the receiver inside a
+// TransposeBand.
+func (a *Tridiag) TBand() Banded {
+	// An alternative would be to return the receiver with DL,DU swapped; see
+	// explanation in T above.
+	return TransposeBand{a}
+}
+
+// RawTridiagonal returns the underlying lapack64.Tridiagonal used by the
+// receiver. Changes to elements in the receiver following the call will be
+// reflected in the returned matrix.
+func (a *Tridiag) RawTridiagonal() lapack64.Tridiagonal {
+	return a.mat
+}
+
+// SetRawTridiagonal sets the underlying lapack64.Tridiagonal used by the
+// receiver. Changes to elements in the receiver following the call will be
+// reflected in the input.
+func (a *Tridiag) SetRawTridiagonal(mat lapack64.Tridiagonal) {
+	a.mat = mat
+}
+
+// IsEmpty returns whether the receiver is empty. Empty matrices can be the
+// receiver for size-restricted operations. The receiver can be zeroed using
+// Reset.
+func (a *Tridiag) IsEmpty() bool {
+	return a.mat.N == 0
+}
+
+// Reset empties the matrix so that it can be reused as the receiver of a
+// dimensionally restricted operation.
+//
+// Reset should not be used when the matrix shares backing data. See the Reseter
+// interface for more information.
+func (a *Tridiag) Reset() {
+	a.mat.N = 0
+	a.mat.DL = a.mat.DL[:0]
+	a.mat.D = a.mat.D[:0]
+	a.mat.DU = a.mat.DU[:0]
+}
+
+// Clone makes a copy of the input Tridiag into the receiver, overwriting the
+// previous value of the receiver. Clone does not place any restrictions on
+// receiver shape.
+func (a *Tridiag) Clone(from *Tridiag) {
+	n := from.mat.N
+	switch n {
+	case 0:
+		panic(ErrZeroLength)
+	case 1:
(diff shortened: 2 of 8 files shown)


mat/index_bound_checks.go
```diff
@@ -346,3 +346,52 @@ func (d *DiagDense) setDiag(i int, v float64) {
...
+	switch i - j {
+	case -1:
+		return a.mat.DU[i]
+	case 0:
+		return a.mat.D[i]
+	case 1:
+		return a.mat.DL[j]
+	default:
+		return 0
+	}
```
kortschak, line 371: This is elegant.

mat/tridiag.go
```diff
@@ -0,0 +1,371 @@
...
+	switch {
+	case n <= 0:
+		if n == 0 {
+			panic(ErrZeroLength)
+		}
+		panic(ErrNegativeDimension)
+	case n == 1:
+		if dl != nil || du != nil || (d != nil && len(d) != 1) {
+			panic(ErrShape)
+		}
```
kortschak, line 54: Can we allow `len(dl) == 0` and `len(du) == 0`?

mat/tridiag.go
```diff
@@ -0,0 +1,371 @@
...
+}
+
+// Tridiag represents a tridiagonal matrix by its three diagonals.
+type Tridiag struct {
+	mat lapack64.Tridiagonal
+}
+
+// NewTridiag creates a new n×n tridiagonal matrix with the first sub-diagonal
+// in dl, the main diagonal in d and the first super-diagonal in du. If dl, d,
+// and du are nil, new backing slices will be allocated for them. If dl and du
```
kortschak, line 39: This is ambiguous; it could be read as "if any of the inputs are nil, those nil inputs will be created at the appropriate size", but the behaviour is that if any of them are not nil, they must be already correctly sized. I think the behaviour is correct, but the documentation could be more clear. Adding "all of", "If all of dl, d and du..." would do this I think.

mat/tridiag.go
```diff
@@ -0,0 +1,371 @@
...
+}
+
+// Bandwidth returns 1, 1 - the upper and lower bandwidths of the matrix.
+func (a *Tridiag) Bandwidth() (kl, ku int) {
+	return 1, 1
+}
+
+// T performs an implicit transpose by returning the receiver inside a Transpose.
+func (a *Tridiag) T() Matrix {
+	// TODO(vladimir-ch): should we return the receiver with DL,DU swapped?
```
kortschak, line 91: I guess this depends on how backing routines handle it. From memory it's just a swapping of diags rather than an trans param, no?

mat/index_bound_checks.go
vladimir-ch: Thanks. Btw, I didn't know it's possible to comment on a range of lines, nice.

mat/tridiag.go
vladimir-ch: I think we can, I'll make the change.

mat/tridiag.go
vladimir-ch: I'm glad you're bringing up the correctness of the behavior. I thought about being completely flexible and allowing nil for any of dl, d, du but then didn't see much benefit in it and went with the current stricter version.

I'll add "all of", thanks.

mat/tridiag.go
vladimir-ch: `Dlagtm` (and `Lagtm`) take the trans parameter, `Dgtsv` does not but we've added it to `Gtsv` for consistency and so that the caller doesn't have to fiddle with the fields of `lapack64.Triangular`. But I think that doesn't affect the choice here. With swapping `untransposeExtract` will just always return false, so in the end there's not much difference. Maybe we should keep using the Transpose type for consistency unless we want to take advantage of the easy transposing that Tridiag representation offers.

mat/tridiag.go
vladimir-ch: Done.

mat/tridiag.go
vladimir-ch: Done.

mat/index_bound_checks.go
```diff
@@ -346,3 +346,52 @@ func (d *DiagDense) setDiag(i int, v float64) {
...
+	return a.at(i, j)
+}
+
+func (a *Tridiag) at(i, j int) float64 {
+	if uint(i) >= uint(d.mat.N) {
+		panic(ErrRowAccess)
+	}
+	if uint(j) >= uint(d.mat.N) {
+		panic(ErrColAccess)
+	}
```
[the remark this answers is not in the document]
kortschak: `int` is the standard integer. This test is just a way of testing for overflow and underflow in a single test.

vladimir-ch:
I've removed the Draft tag, this is ready for review. Sorry it took me so long.

mat/tridiag.go
```diff
@@ -0,0 +1,401 @@
...
+	_            Matrix           = tridiagDense
+	_            allMatrix        = tridiagDense
+	_            denseMatrix      = tridiagDense
+	_            Banded           = tridiagDense
+	_            MutableBanded    = tridiagDense
+	_            RawTridiagonaler = tridiagDense
+)
+
+// A RawTridiagonaler can return a lapack64.Tridiagonal representation of the
+// receiver. Changes to the DL, D, DU slices in lapack64.Tridiagonal will be
```
kortschak, line 26:
```
// receiver. Changes to the elements of DL, D, DU in lapack64.Tridiagonal will be
```

mat/tridiag.go
```diff
@@ -0,0 +1,401 @@
...
+}
+
+// Bandwidth returns 1, 1 - the upper and lower bandwidths of the matrix.
+func (a *Tridiag) Bandwidth() (kl, ku int) {
+	return 1, 1
+}
+
+// T performs an implicit transpose by returning the receiver inside a Transpose.
+func (a *Tridiag) T() Matrix {
+	// An alternative would be to return the receiver with DL,DU swapped.
```
kortschak, line 83: It's probably worth having a brief comment explaining why this approach was chosen so when we come back to it the thinking is present.

mat/tridiag.go
```diff
@@ -0,0 +1,401 @@
...
+// Reset empties the matrix so that it can be reused as the receiver of a
+// dimensionally restricted operation.
+//
+// Reset should not be used when the matrix shares backing data. See the Reseter
+// interface for more information.
+func (a *Tridiag) Reset() {
+	a.mat.N = 0
+	a.mat.DL = a.mat.DL[:0:0]
+	a.mat.D = a.mat.D[:0:0]
+	a.mat.DU = a.mat.DU[:0:0]
```
kortschak, line 124: Do we want to set cap to zero? This is equivalent to setting the fields to nil as far as the resizing goes, but with the cost of keeping the allocation. If future allocations are being avoided, shouldn't it be 
```
	a.mat.DL = a.mat.DL[:0]
	a.mat.D = a.mat.D[:0]
	a.mat.DU = a.mat.DU[:0]
```

mat/tridiag.go
```diff
@@ -0,0 +1,401 @@
...
+}
+
+// Clone makes a copy of the input Tridiag into the receiver, overwriting the
+// previous value of the receiver. Clone does not place any restrictions on
+// receiver shape.
+func (a *Tridiag) Clone(from *Tridiag) {
+	n := from.mat.N
+	switch n {
+	case 0:
```
kortschak, line 133: Does this ever happen?

mat/tridiag.go
```diff
@@ -0,0 +1,401 @@
...
+			DU: use(a.mat.DU, 0),
+		}
+	case 1:
+		a.mat = lapack64.Tridiagonal{
+			N:  1,
+			DL: use(a.mat.DL, 0),
+			D:  use(a.mat.D, 1),
+			DU: use(a.mat.DU, 0),
+		}
+		copy(a.mat.D, from.mat.D)
```
kortschak, line 147: `a.mat.D[0] = from.mat.D[0]`

mat/tridiag.go
```diff
@@ -0,0 +1,401 @@
...
+	zero(a.mat.DU)
+}
+
+// Trace returns the trace of the matrix.
+func (a *Tridiag) Trace() float64 {
+	var sum float64
+	for _, aii := range a.mat.D {
+		sum += aii
+	}
+	return sum
```
kortschak, line 185: `return f64.Sum(a.mat.D)`?

mat/tridiag.go
```diff
@@ -0,0 +1,401 @@
...
+		}
+	}
+}
+
+// DoRowNonZero calls the function fn for each of the non-zero elements of row i
+// of A. The function fn takes a row/column index and the element value of A at
+// (i,j).
+func (a *Tridiag) DoRowNonZero(i int, fn func(i, j int, v float64)) {
+	n := a.mat.N
+	if i < 0 || n <= i {
```
kortschak, line 308: `uint(i) >= uint(n)`?

mat/tridiag.go
```diff
@@ -0,0 +1,401 @@
...
+		}
+	}
+}
+
+// DoColNonZero calls the function fn for each of the non-zero elements of
+// column j of A. The function fn takes a row/column index and the element value
+// of A at (i, j).
+func (a *Tridiag) DoColNonZero(j int, fn func(i, j int, v float64)) {
+	n := a.mat.N
+	if j < 0 || n <= j {
```
kortschak, line 358: `uint(j) >= uint(n)`?

mat/tridiag_test.go
```diff
@@ -0,0 +1,529 @@
+// Copyright ©2021 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package mat
+
+import (
+	"fmt"
+	"math"
+	"math/rand"
```
kortschak, line 10: "golang.org/x/exp/rand"

mat/tridiag.go
vladimir-ch: I copied it from `BandDense` and `SymBandDense` although I did find it strange. We should change it in those two places as well.

mat/tridiag.go
vladimir-ch: It can, after a `Reset`. I'd prefer not to care about zero-sized matrices but I think with Reset it's necessary, isn't it?

mat/tridiag_test.go
vladimir-ch: Oh. Thanks, done.

mat/tridiag.go
vladimir-ch: Done.

mat/tridiag.go
vladimir-ch: Done.

mat/tridiag.go
vladimir-ch: Yes, done.

mat/tridiag.go
vladimir-ch: Done.

mat/tridiag.go
vladimir-ch: I've extended the comment a bit, PTAL.

mat/tridiag.go
vladimir-ch: Done.

mat/tridiag.go
```diff
@@ -0,0 +1,404 @@
...
+
+// Trace returns the trace of the matrix.
+func (a *Tridiag) Trace() float64 {
+	return f64.Sum(a.mat.D)
+}
+
+// Norm returns the specified norm of the receiver. Valid norms are:
+//  1 - The maximum absolute column sum
+//  2 - The Frobenius norm, the square root of the sum of the squares of the elements
+//  Inf - The maximum absolute row sum
```
romanwerpachowski, line 194: Why not the maximum absolute element, like in https://en.wikipedia.org/wiki/Matrix_norm#Max_norm?

mat/tridiag.go
```diff
@@ -0,0 +1,404 @@
...
+	t := blas.NoTrans
+	if trans {
+		t = blas.Trans
+	}
+	xMat, _ := untransposeExtract(x)
+	if xVec, ok := xMat.(*VecDense); ok && dst != xVec {
+		dst.checkOverlap(xVec.mat)
+		lapack64.Lagtm(t, 1, a.mat, xVec.asGeneral(), 0, dst.asGeneral())
+	} else {
+		xCopy := getWorkspaceVec(n, false)
```
romanwerpachowski, line 217: Is there a way to avoid allocating memory here?

mat/tridiag.go
kortschak: Is it reasonable for a user to clone from an empty matrix? I'm not sure that it is. What would it mean? I seems to me like a long way of saying `a.Reset()`.

mat/tridiag.go
kortschak: Yes, we should.

mat/tridiag.go
kortschak: What you have now is good. Maybe on the second one, just refer to the first for the explanation.
```
	// An alternative would be to return the receiver with DL,DU swapped; see
	// explanation in T above.
```

mat/tridiag.go
kortschak: No, not if we want to provide a safe way for users to update their parameters as the receiver. In most cases this won't happen and there will just be the overlap check in the `ok && dst != xVec` condition. If the copy is an issue for performance, it's easy to have a pair of alternating destinations.

mat/tridiag.go
vladimir-ch: I agree, so should I rather panic with `ErrZeroLength`?

kortschak: Yes, I think that's right.

mat/tridiag.go
vladimir-ch: I think it's a limitation of the Norm API that specifies the norm by means of a float64, thus limiting the norms to p-norms and excluding element-wise norms like the max norm. Since the max norm is the only one missing and supported by Lapack, we could provide it as `MaxAbs` instead of changing Norm?

mat/tridiag.go
vladimir-ch: Done.

mat/tridiag.go
romanwerpachowski: Makes sense. Doesn't have to be done in this PR, of course.

mat/tridiag.go
```diff
@@ -0,0 +1,399 @@
...
+}
+
+// TBand performs an implicit transpose by returning the receiver inside a
+// TransposeBand.
+func (a *Tridiag) TBand() Banded {
+	// An alternative would be to return the receiver with DL,DU swapped; the
+	// untranspose function would then always return false. With Transpose the
+	// diagonal swapping will be done in tridiagonal routines in lapack like
+	// lapack64.Gtsv or gonum.Dlagtm based on the trans parameter.
+	return TransposeBand{a}
```
kortschak, line 98: Trim this comment?

vladimir-ch: Done.

mat/tridiag.go
vladimir-ch: Thanks, done.

kortschak (approved):

PR #1542, opened 2021-01-06
mat: add non-conjugate transpose for complex matrices

### LINKED ISSUE #1539: proposal: mat: provide non-conjugate transpose for CMatrix
### Background

I have had a query about non-conjugate transpose for complex matrices [on the gopher slack data-science channel](https://gophers.slack.com/archives/C0UKS8V9S/p1609624024198600). While my initial suggestion of implementing a local transpose type works (below), it prevents that transpose information being available to a BLAS or LAPACK back-end, both of which accept (variously) T and H (by various names) as trans parameters.

```
type Transpose struct {
	CMatrix mat.CMatrix
}

func (t Transpose) At(i, j int) complex128 {
	return t.CMatrix.At(j, i)
}

func (t Transpose) Dims() (r, c int) {
	c, r = t.CMatrix.Dims()
	return r, c
}

func (t Transpose) H() mat.CMatrix {
	return mat.Conjugate{t}
}
```

The inclusion of T in addition to H was briefly considered in gonum/matrix#283. Note the [comment](https://github.com/gonum/matrix/issues/283#issuecomment-173796281) which points to use of T/H confusion in the [LAPACK errata](https://www.netlib.org/lapack/Errata/errata-lapack-3.3.0.html#_strong_span_class_green_bug0024_span_strong_the_notation_a_is_confusing_for_complex_matrices_does_a_stand_for_a_transpose_of_a_hermitian_transpose). It's worth noting that while the documentation [linked in the comment](https://www.netlib.org/lapack/complex/) always refers to conjugate transpose, that it is in the context of describing `X**H`; the BLAS and LAPACK code *does* differentiate between T and H.

### Proposal

I propose adding a `T() CMatrix` method to `CMatrix`, the backing `CDense` implementation, and `CTranspose` and support types (`CUntransposer`?).

Adding the mechanisms for implicit transpose allows the transpose state to be given to backing routines in an efficient manner in the same way as is currently done for the real matrix types (when those API additions are made in mat).

The addition of a new method to `CMatrix` is a breaking API change. However the number of users of complex matrices is likely to be low at the moment since very little API is provided (just enough to support `Eigen`). So if it should be added at some time it should be earlier than later.

There is some subtlety in how to deal with the case of `m.H().T()`/`m.T().H()` as inputs to back-end routines; this doesn't need to be dealt with immediately though probably just conjugating the elements of `m` into a workspace and working on that with `NoTrans` would be the best. I think to support this we should add a `Conj(CMatrix)` method that performs an element-wise conjugation on the argument, placing the result in the receiver.

It's worth noting that NumPy uses the matrix conjugate to [refer to element-wise conjugation](https://numpy.org/doc/stable/reference/generated/numpy.matrix.conjugate.html), not the conjugate transpose (which it uses [H as a method for](https://numpy.org/doc/stable/reference/generated/numpy.matrix.H.html) matching our approach). It may be worth considering renaming the `Conjugate` type (`ConjTranspose`?) and the associated interfaces. Again, there are few users of `CMatrix` now meaning that if the change should be made, it should be made earlier than later.

The implementation cost is minimal and I'm happy to do it.

/cc @notmgsk @btracey @vladimir-ch

kortschak:
Please take a look.

Closes #1539.

### DIFF
--- a/mat/cdense.go
+++ b/mat/cdense.go
@@ -44,10 +44,10 @@ func (m *CDense) T() CMatrix {
 	return CTranspose{m}
 }
 
-// ConjElem calculates the element-wise conjugate of a and stores the result
-// in the receiver.
-// ConjElem will panic if m and a do not have the same dimension unless m is empty.
-func (m *CDense) ConjElem(a CMatrix) {
+// Conj calculates the element-wise conjugate of a and stores the result in the
+// receiver.
+// Conj will panic if m and a do not have the same dimension unless m is empty.
+func (m *CDense) Conj(a CMatrix) {
 	ar, ac := a.Dims()
 	aU, aTrans, aConj := untransposeExtractCmplx(a)
 	m.reuseAsNonZeroed(ar, ac)
@@ -82,7 +82,7 @@ func (m *CDense) ConjElem(a CMatrix) {
 	}
 }
 
-// Slice returns a new Matrix that shares backing data with the receiver.
+// Slice returns a new CMatrix that shares backing data with the receiver.
 // The returned matrix starts at {i,j} of the receiver and extends k-i rows
 // and l-j columns. The final row in the resulting matrix is k-1 and the
 // final column is l-1.
--- a/mat/cdense.go
+++ b/mat/cdense.go
@@ -4,7 +4,11 @@
 
 package mat
 
-import "gonum.org/v1/gonum/blas/cblas128"
+import (
+	"math/cmplx"
+
+	"gonum.org/v1/gonum/blas/cblas128"
+)
 
 var (
 	cDense *CDense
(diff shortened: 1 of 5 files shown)


kortschak:
On this basis of https://mathworld.wolfram.com/ConjugateMatrix.html, I'm wondering whether we should rename `Conjugate` to `ConjTranspose`.

mat/cdense.go
```diff
@@ -29,9 +33,53 @@ func (m *CDense) Dims() (r, c int) {
...
+// T performs an implicit transpose by returning the receiver inside a
+// CTranspose.
+func (m *CDense) T() CMatrix {
+	return CTranspose{m}
+}
+
+// ConjElem calculates the element-wise conjugate of a and stores the result
+// in the receiver.
+// ConjElem will panic if m and a do not have the same dimension unless m is empty.
+func (m *CDense) ConjElem(a CMatrix) {
```
vladimir-ch, line 50: Even though I proposed it, I'm wondering if just `Conj` would be enough or better now that Conjugate is renamed to ConjTranspose. The naming and effect would then be consistent with `cmplx.Conj`.

mat/cmatrix.go
```diff
@@ -35,53 +41,109 @@ type RawCMatrixer interface {
...
+
+// At returns the value of the element at row i and column j of the conjugate
+// transposed matrix, that is, row j and column i of the CMatrix field.
+func (t CTranspose) At(i, j int) complex128 {
+	return t.CMatrix.At(j, i)
+}
+
+// Dims returns the dimensions of the transposed matrix. The number of rows returned
+// is the number of columns in the CMatrix field, and the number of columns is
+// the number of rows in the Matrix field.
```
vladimir-ch, line 101: CMatrix

mat/cmatrix.go
```diff
@@ -35,53 +41,109 @@ type RawCMatrixer interface {
...
+func (t CTranspose) Dims() (r, c int) {
+	c, r = t.CMatrix.Dims()
+	return r, c
+}
+
+// H performs an implicit transpose by  returning the receiver inside a
```
vladimir-ch, line 107: nit: double space

mat/pool.go
```diff
@@ -236,3 +246,28 @@ func getInts(l int, clear bool) []int {
 func putInts(w []int) {
 	poolInts[bits(uint64(cap(w)))].Put(&w)
 }
+
+// getWorkspaceCmplx returns a *CDense of size r×c and a data slice
+// with a cap that is less than 2*r*c. If clear is true, the
+// data slice visible through the Matrix interface is zeroed.
```
vladimir-ch, line 252: CMatrix

mat/cdense.go
kortschak: Thanks, this was what I was thinking too.

mat/cmatrix.go
kortschak: Fixed

mat/cmatrix.go
kortschak: Fixed (and below)

mat/pool.go
kortschak: Fixed

vladimir-ch (approved):

kortschak:
Do you have any views about the renaming of `Conjugate` for another PR?

vladimir-ch:
I would leave it in this PR, at most split into a separate commit if it's not too onerous.

kortschak:
I'll send another.

kortschak:
I must have been very tired when I wrote that. It is already done.

PR #1547, opened 2021-01-08
dsp/window: add Values.TransformTo window functions and example

kortschak:
Please take a look.


dsp/window/window_example_test.go
```diff
@@ -118,3 +118,42 @@ func ExampleValues() {
 	//
 	// dst: [0.000000 0.164595 0.324699 0.475947 0.614213 0.735724 0.837166 0.915773 0.969400 0.996584 0.996584 0.969400 0.915773 0.837166 0.735724 0.614213 0.475947 0.324699 0.164595 0.000000]
 }
+
+func ExampleValues_TransformTo_gabor() {
```
sbinet, line 122:
```suggestion
func ExampleValues_transformTo_gabor() {
```
(I think)

sbinet (approved):
LGTM

dsp/window/window_example_test.go
kortschak: No, `TransformTo` is a method of `Values`. The example title renders as "Values.TransformTo (Gabor)".

PR #1550, opened 2021-01-21
Code cleanup

notsobad:
It seems `compensation` is always 0 here.

### DIFF
--- a/stat/stat.go
+++ b/stat/stat.go
@@ -1277,7 +1277,7 @@ func MeanVariance(x, weights []float64) (mean, variance float64) {
 			ss += d * d
 			compensation += d
 		}
-		variance = (ss - compensation*compensation/float64(len(x))) / float64(len(x)-1)
+		variance = ss / float64(len(x)-1)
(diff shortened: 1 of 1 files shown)


kortschak (changes requested):
This is not correct. Please read the paper that is referred to in the comment at the preamble to the function. The compensation may be zero for the values you observe, but will not be for all cases. 

> Chan and Lewis (1978) first derived the condition number, 𝜿, of a sample {x_i} (with respect to computing the variance). This condition number measures the sensitivity of S for the given data set. If relative errors of size 𝜸 are introduced into the x_i, then the relative change in S is bounded by 𝜿𝜸.

These two plots show the response of the naive approach (what this PR changes the behaviour to) and the corrected two-pass algorithm (what was here originally). The x axis shows 𝜿 and the y axis is the relative error in S.

Screenshot from 2021-01-21 19-12-24

Screenshot from 2021-01-21 19-12-51

notsobad:
Thank you, I will read that paper

Issue #1553, opened 2021-01-28
spatial/{r2,r3}: harmonize API with num/quat


sbinet:
as alluded to in https://github.com/gonum/gonum/pull/1516#issuecomment-734966207, we should probably harmonize the `{r2,r3}.Vec` API with the one from `num/quat`.

kortschak:
SGTM

PR #1554, opened 2021-01-28
spatial/r3: harmonize Vec API with num/quat.Number

sbinet:
Migration to the new API can be achieved with this rsc.io/rf script:

```
rf ex {
	import "gonum.org/v1/gonum/spatial/r3";
	var p,q r3.Vec;
	var f float64;

	p.Add(q) -> r3.Add(p, q);
	p.Sub(q) -> r3.Sub(p, q);
	p.Scale(f) -> r3.Scale(f, p);
	p.Dot(q) -> r3.Dot(p, q);
	p.Cross(q) -> r3.Cross(p, q);
        p.Rotate(f, q) -> r3.Rotate(p, f, q);
}
```

and, similarly, for `r2`:
```
    rf ex {
            import "gonum.org/v1/gonum/spatial/r2";
            var p,q r2.Vec;
            var f float64;
    
            p.Add(q) -> r2.Add(p, q);
            p.Sub(q) -> r2.Sub(p, q);
            p.Scale(f) -> r2.Scale(f, p);
            p.Dot(q) -> r2.Dot(p, q);
            p.Cross(q) -> r2.Cross(p, q);
            p.Rotate(f, q) -> r2.Rotate(p, f, q);
    }
```

Fixes gonum/gonum#1553.

Please take a look.

### DIFF
--- a/spatial/r2/deprecated.go
+++ b/spatial/r2/deprecated.go
@@ -1,0 +1,49 @@
+// Copyright ©2021 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package r2
+
+// TODO(sbinet): remove this file for Gonum-v0.10.0.
+
+// Add returns the vector sum of p and q.
+//
+// DEPRECATED: use r2.Add.
+func (p Vec) Add(q Vec) Vec {
+	return Add(p, q)
+}
+
+// Sub returns the vector sum of p and -q.
+//
+// DEPRECATED: use r2.Sub.
+func (p Vec) Sub(q Vec) Vec {
+	return Sub(p, q)
+}
+
+// Scale returns the vector p scaled by f.
+//
+// DEPRECATED: use r2.Scale.
+func (p Vec) Scale(f float64) Vec {
+	return Scale(f, p)
+}
+
+// Dot returns the dot product p·q.
+//
+// DEPRECATED: use r2.Dot.
+func (p Vec) Dot(q Vec) float64 {
+	return Dot(p, q)
+}
+
+// Cross returns the cross product p×q.
+//
+// DEPRECATED: use r2.Cross.
+func (p Vec) Cross(q Vec) float64 {
+	return Cross(p, q)
+}
+
+// Rotate returns a new vector, rotated by alpha around the provided point, q.
+//
+// DEPRECATED: use r2.Rotate.
+func (p Vec) Rotate(alpha float64, q Vec) Vec {
+	return Rotate(p, alpha, q)
+}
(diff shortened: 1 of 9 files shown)


spatial/r2/vector.go
```diff
@@ -12,34 +12,72 @@ type Vec struct {
...
+}
+
+// Cross returns the cross product p×q.
+func Cross(p, q Vec) float64 {
+	return p.X*q.Y - p.Y*q.X
+}
+
+// Add returns the vector sum of p and q.
+//
+// DEPRECATED: use r2.Add.
```
kortschak, line 50: Since we are using modules and we are in v0.\*, we don't need to deprecate, although we have done this once before. If we deprecate, can you add a todo to remove these at the v0.10.0 release. For the floats breaking change I moved the old API into a deprecated.go file with a TODO note.

spatial/r3/vector.go
```diff
@@ -16,43 +16,81 @@ type Vec struct {
...
-func (p Vec) Cross(q Vec) Vec {
+func Cross(p, q Vec) Vec {
 	return Vec{
 		p.Y*q.Z - p.Z*q.Y,
 		p.Z*q.X - p.X*q.Z,
 		p.X*q.Y - p.Y*q.X,
 	}
 }
 
+// Add returns the vector sum of p and q.
```
kortschak, line 59: Same comment here.

spatial/r2/vector.go
sbinet: I'll move these into a `deprecate.go` file.

keeping these around for a release allows mechanization of the migration (via the `rf` scripts I put in the commit messages)

kortschak (dismissed):

sbinet:
done.

note that I've left `func (p Vec) Rotate(angle float64, axis Vec) Vec { ... }` alone.

should I also migrate it to `func Rotate(p Vec, angle float64, axis Vec) Vec { ...}` ?

spatial/r2/deprecated.go
```diff
@@ -0,0 +1,40 @@
+// Copyright ©2021 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package r2
```
kortschak, line 5:
```
// TODO: Delete this file for v0.10.0.
```

and in the other deprecated.go as well (you can put me in the TODO parens if you want).

kortschak (commented):
> should I also migrate it to, say, `func Rotate(p Vec, angle float64, axis Vec) Vec { ...}`

Yes, I think that would be good.

kortschak (approved):

PR #1558, opened 2021-02-05
stat: don't adjust unobserved bins in ROC

### LINKED ISSUE #1557: stat: ROC does not correctly handle bins beyond highest observation
package main

import (
	"fmt"
	"math"
	"gonum.org/v1/gonum/stat"
)

func main() {
	scores := []float64{0.1, 0.35, 0.4, 0.8}
	cutoffs := []float64{math.Inf(-1), 0.1, 0.35, 0.4, 0.8, 0.9, 1}
	tpr, fpr, thresh := stat.ROC(cutoffs, scores, classes, nil)
	fmt.Printf("thresh: %v\n", thresh)
}

true  positive rate: [0 0 0 0 0.5 0.5 1]
false positive rate: [0 0 0 0.5 0.5 1 1]
thresh: [+Inf 1 0.9 0.8 0.4 0.35 0.1]

### What actually happened?
true  positive rate: [1 1 0 0 0.5 0.5 1]
false positive rate: [1 1 0 0.5 0.5 1 1]
thresh: [+Inf 1 0.9 0.8 0.4 0.35 0.1]

go version
go version go1.15.6 windows/amd64

rg gonum.org/v1/gonum go.sum
18:gonum.org/v1/gonum v0.0.0-20180816165407-929014505bf4/go.mod h1:Y+Yx5eoAFn32cQvJDxZx5Dpnq+c3wtXuadVZAcxbbBo=

require gonum.org/v1/gonum v0.8.2

kortschak:
Please take a look.

Fixes #1557.

/cc @zhongsigang

### DIFF
--- a/stat/roc.go
+++ b/stat/roc.go
@@ -106,7 +106,7 @@ func ROC(cutoffs, y []float64, classes []bool, weights []float64) (tpr, fpr, thr
 
 	invNeg := 1 / nNeg
 	invPos := 1 / nPos
-	for i := range tpr {
+	for i := range tpr[:bin+1] {
 		// Prevent fused float operations by
 		// making explicit float64 conversions.
 		tpr[i] = 1 - float64(tpr[i]*invPos)
--- a/stat/roc_test.go
+++ b/stat/roc_test.go
@@ -210,6 +210,14 @@ func TestROC(t *testing.T) {
 			wantFPR:    nil,
 			wantThresh: nil,
 		},
+		{ // 23
+			y:          []float64{0.1, 0.35, 0.4, 0.8},
+			c:          []bool{true, false, true, false},
+			cutoffs:    []float64{-1, 0.1, 0.35, 0.4, 0.8, 0.9, 1},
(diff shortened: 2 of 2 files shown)


stat/roc.go
```diff
@@ -106,7 +106,7 @@ func ROC(cutoffs, y []float64, classes []bool, weights []float64) (tpr, fpr, thr
 
 	invNeg := 1 / nNeg
 	invPos := 1 / nPos
-	for i := range tpr {
+	for i := range tpr[:bin+1] {
```
Armadilloa16, line 109: Unless I'm mistaken, this would handle when there is a single "extra" bin above the maximum value in y, but not multiple. I could add a test case for this? I agree it's prettier but I don't think it's equivalent

kortschak: This leaves all elements in `tpr[bin+1:]` and `fpr[bin+1:]` as zero, which AFAICS is the outcome of the existing code in #1561.

kortschak:
Closing in favour of #1561.

Issue #1559, opened 2021-02-06
stat:ROC has a wrong result when  the span in y is not consistent with that in cutoffs


zhongsigang:
### What are you trying to do?

package main

import (
	"fmt"
	"gonum.org/v1/gonum/stat"
	"math"
)

func main() {

	y := []float64{0.1, 0.35, 0.4, 0.8}
	c := []bool{true, false, true, false}
	cutoffs := []float64{math.Inf(-1), 0.1, 0.36, 0.8}

	tpr, fpr, thresh := stat.ROC(cutoffs, y, c, nil)
	fmt.Printf("true  positive rate: %v\n", tpr)
	fmt.Printf("false positive rate: %v\n", fpr)
	fmt.Printf("thresh: %v\n", thresh)
}

### What did you expect to happen?
true  positive rate: [0 0 0.5 1]
false positive rate: [0 0.5 0.5 1]
thresh: [+Inf 0.8 0.36 0.1]

### What actually happened?
true  positive rate: [0 0.5 0.5 1]
false positive rate: [0 0.5 1 1]
thresh: [+Inf 0.8 0.36 0.1]

### What version of Go and Gonum are you using?

go version
go version go1.15.6 windows/amd64

rg gonum.org/v1/gonum go.sum
18:gonum.org/v1/gonum v0.0.0-20180816165407-929014505bf4/go.mod h1:Y+Yx5eoAFn32cQvJDxZx5Dpnq+c3wtXuadVZAcxbbBo=

### Does this issue reproduce with the current master?
require gonum.org/v1/gonum v0.8.2

zhongsigang:
I suggest that we do not use the inverse of cutoffs, but use the meaning of **<=** directly.

zhongsigang:
Or cancel the parameter cutoff, that is, it always equals nil.

kortschak:
Please explain with more detail why you think this is wrong and why your suggestion is appropriate; this bug is very terse.

Armadilloa16:
I can confirm I've reproduced this bug, and that it is a bug --- output is incorrect. 

I had a look through the src and I'm not sure what's causing it. The codes been re-written a few times as well, I'd suggest it could be re-written to be more readable which might make debugging easier. I can have a go at this if that's desirable.

Armadilloa16:
But basically what I'm saying is I agree --- we should not use the inverse of cutoffs, and should calculate the output directly rather than transforming afterwards.

kortschak:
[...]

Issue #708, opened 2016-06-01
stat: ROC signature


Armadilloa16:
I came up with what I think is an improvement to the ROC function, 
in which the cutoffs are defined explicitly instead of the number of cutoffs n.
This allows for unequally-spaced cutoffs, allows for all cutoffs to be used
more naturally, and simplifies the code itself.

Is this a good idea -- should I put a PR in for it, @sbinet @kortschak ?

Also, I was thinking maybe ROC should output an error as well, instead of all
the panic statements? Or is it appropriate that it panic?

What I did is here, in case you wanted to look at the specifics of what I am 
suggesting:
https://github.com/Armadilloa16/stat/tree/roc

kortschak:
@Armadilloa16 I have salvaged the branch you had at your fork of stat and cleaned it up and rebased it onto master. I've also (I think correctly - though not convinced) altered the tests so that they pass.

Are you happy to have that sent as a PR and would you have a quick look over the changes to the test expectation that I've made to see if you agree with them.

The branch is [here](https://github.com/gonum/gonum/tree/stat/roc).

Armadilloa16:
Hey @kortschak sorry for the sluggish responses...

Yes I'm happy for you to send this as a PR, however I took a quick look at the tests: these are all cases in which the issue is the FPR when there are no true negatives, and the TPR when there are no true positives. Returning 0 in these cases doesn't necessarily make sense, although returning 1 might not either... I feel like deciding what to return in these cases should probably be a deliberate decision, maybe returning nil or NA or something like that would be more appropriate in these cases. I.e. maybe there should be a check for these cases. This could be done pretty easily at lines 103 and 106 of roc.go, the condition we are checking for is the else clause: nNeg == 0 or nPos == 0.

Armadilloa16:
Nice example of good testing bringing up special cases worth addressing. :)

kortschak:
OK. So previously it was that the final element was 1. I'm reluctant to add a NaN and we don't have an NA, so I'll revert the test and change the behaviour to match.

Armadilloa16:
Yeah I dunno NaN might be appropriate, you're essentially dividing by zero in that case. Or couldn't you just return an empty vector or something? It makes sense to not calculate FPR when there are no true negatives and not calculate TPR when there are no true positives.

kortschak:
OK, I'll place a NaN. A float division by zero gives Inf BTW.

Armadilloa16:
[...]

Issue #316, opened 2017-11-27
stat: Change parity convention in ROC


MaxHalford:
I was going through the ROC AUC examples and found that the results were not very intuitive. Indeed [the basic example](https://github.com/gonum/gonum/blob/master/stat/roc_example_test.go#L41) gives a ROC AUC of 0.75 but if I run it with sklearn I get:

```python
>>> from sklearn import metrics
>>> y = [0.1, 0.35, 0.4, 0.8]
>>> classes = [1, 0, 1, 0]
>>> metrics.roc_auc_score(classes, y)
0.25
```

Moreover, according to gonum a perfect match between two slices results in a ROC AUC of 0, whereas intuitively it should be 1. Initially I made [a pull request](https://github.com/gonum/gonum/pull/315) because I thought it was a typo in the examples but it turns out the problem is a bit more complicated.

btracey:
It's not that a perfect match gives an AUC of 0, it's that our definition of a perfect match is opposite of what I would intuitively think is a perfect match. ROC treats values below the cutoff as `true` (rather than `false`), and so effectively this means that the `y` values are  "probability of being `false`", instead of the more intuitive "probability of being `true`".  I would vote we switch the function to match sklearn.

sbinet:
fine by me.

MaxHalford:
Are we waiting on a pull request from someone?

btracey:
Sounds like it. Would you like to do so?

MaxHalford:
Sure I'm on it.

PR #1561, opened 2021-02-07
stat: ROC revert functionality when cutoffs is not nil, code readability improvements, and bug fixes

### LINKED ISSUE #1557: stat: ROC does not correctly handle bins beyond highest observation

Armadilloa16:
The main change is when cutoffs is not nil, returned thresholds are now cutoffs (sorted in decreasing order), and the fpr and tpr values are such that they are correct for these thresholds. 

This: 
 - maintains the parity convention decision made in #316 and documented in the ROC description.
 - is (I believe) closer to the original intent of #708
 - should resolve #1557 and #1559 (although changes the expected output)
 - includes extra test case from #1557 as per #1558

### DIFF
--- a/stat/roc.go
+++ b/stat/roc.go
@@ -58,9 +58,8 @@ func ROC(cutoffs, y []float64, classes []bool, weights []float64) (tpr, fpr, thr
 		} else {
 			cutoffs = cutoffs[:len(y)+1]
 		}
-		cutoffs[0] = math.Inf(-1)
 		// Choose all possible cutoffs for unique values in y.
-		bin := 1
+		bin := 0
 		cutoffs[bin] = y[0]
 		for i, u := range y[1:] {
 			if u == y[i] {
@@ -69,7 +68,8 @@ func ROC(cutoffs, y []float64, classes []bool, weights []float64) (tpr, fpr, thr
 			bin++
 			cutoffs[bin] = u
 		}
-		cutoffs = cutoffs[:bin+1]
+		cutoffs[bin+1] = math.Inf(1)
+		cutoffs = cutoffs[:bin+2]
 	} else {
 		// Don't mutate the provided cutoffs.
 		tmp := cutoffs
@@ -84,7 +84,7 @@ func ROC(cutoffs, y []float64, classes []bool, weights []float64) (tpr, fpr, thr
 	for i, u := range classes {
 		// Update the bin until it matches the next y value
 		// skipping empty bins.
-		for bin < len(cutoffs)-1 && y[i] > cutoffs[bin] {
+		for bin < len(cutoffs)-1 && y[i] >= cutoffs[bin] {
 			bin++
 			tpr[bin] = tpr[bin-1]
 			fpr[bin] = fpr[bin-1]
@@ -98,7 +98,8 @@ func ROC(cutoffs, y []float64, classes []bool, weights []float64) (tpr, fpr, thr
 		}
 		nPos += posWeight
 		nNeg += negWeight
-		if y[i] <= cutoffs[bin] {
+		// Count false negatives (in tpr) and true negatives (in fpr).
+		if y[i] < cutoffs[bin] {
 			tpr[bin] += posWeight
 			fpr[bin] += negWeight
 		}
@@ -106,7 +107,10 @@ func ROC(cutoffs, y []float64, classes []bool, weights []float64) (tpr, fpr, thr
 
 	invNeg := 1 / nNeg
 	invPos := 1 / nPos
-	for i := range tpr {
+	// Convert negative counts to TPR and FPR.
+	// Bins beyond the maximum value in y are skipped
+	// leaving these fpr and tpr elements as zero.
+	for i := range tpr[:bin+1] {
 		// Prevent fused float operations by
 		// making explicit float64 conversions.
 		tpr[i] = 1 - float64(tpr[i]*invPos)
@@ -116,10 +120,9 @@ func ROC(cutoffs, y []float64, classes []bool, weights []float64) (tpr, fpr, thr
 		tpr[i], tpr[j] = tpr[j], tpr[i]
 		fpr[i], fpr[j] = fpr[j], fpr[i]
 	}
-	for i, j := 1, len(cutoffs)-1; i < j; i, j = i+1, j-1 {
+	for i, j := 0, len(cutoffs)-1; i < j; i, j = i+1, j-1 {
 		cutoffs[i], cutoffs[j] = cutoffs[j], cutoffs[i]
 	}
-	cutoffs[0] = math.Inf(1)
 
 	return tpr, fpr, cutoffs
 }
--- a/stat/roc_test.go
+++ b/stat/roc_test.go
@@ -23,9 +23,9 @@ func TestROC(t *testing.T) {
 		wantFPR    []float64
 		wantThresh []float64
 	}{
-		// Test cases were calculated using sklearn metrics.roc_curve when
-		// cutoffs is nil. Where cutoffs is not nil, a visual inspection is
-		// used.
+		// Test cases were informed by using sklearn metrics.roc_curve when
+		// cutoffs is nil, but all test cases (including when cutoffs is not
+		// nil) where calculated manually.
 		// Some differences exist between unweighted ROCs from our function
 		// and metrics.roc_curve which appears to use integer cutoffs in that
 		// case. sklearn also appears to do some magic that trims leading zeros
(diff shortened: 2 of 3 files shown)


stat/roc.go
```diff
@@ -98,28 +96,35 @@ func ROC(cutoffs, y []float64, classes []bool, weights []float64) (tpr, fpr, thr
...
+		tpr[bin] = tpr[bin-1]
+		fpr[bin] = fpr[bin-1]
+	}
 
 	invNeg := 1 / nNeg
 	invPos := 1 / nPos
 	for i := range tpr {
 		// Prevent fused float operations by
 		// making explicit float64 conversions.
+		// Convert negative counts to TPR and FPR
```
kortschak, line 119: I think this would be better above the for loop. Also add final full stop.

stat/roc.go
```diff
@@ -98,28 +96,35 @@ func ROC(cutoffs, y []float64, classes []bool, weights []float64) (tpr, fpr, thr
 		}
 		nPos += posWeight
 		nNeg += negWeight
-		if y[i] <= cutoffs[bin] {
+		if y[i] < cutoffs[bin] {
+			// Actually counting false negatives
```
kortschak, line 102: Full stops for sentences.

stat/roc.go
```diff
@@ -98,28 +96,35 @@ func ROC(cutoffs, y []float64, classes []bool, weights []float64) (tpr, fpr, thr
...
 			tpr[bin] += posWeight
+			// Actually counting true negatives
 			fpr[bin] += negWeight
 		}
 	}
+	for bin < len(cutoffs)-1 {
+		bin++
+		tpr[bin] = tpr[bin-1]
+		fpr[bin] = fpr[bin-1]
+	}
```
kortschak, line 112:
```
lastTpr = tpr[bin]
lastFpr = fpr[bin]
for bin++; bin < len(cutoffs); bin++ {
    tpr[bin] = lastTpr
    fpr[bin] = lastFpr
}
```

stat/roc.go
```diff
@@ -55,12 +55,9 @@ func ROC(cutoffs, y []float64, classes []bool, weights []float64) (tpr, fpr, thr
 	if len(cutoffs) == 0 {
 		if cutoffs == nil || cap(cutoffs) < len(y)+1 {
 			cutoffs = make([]float64, len(y)+1)
-		} else {
-			cutoffs = cutoffs[:len(y)+1]
 		}
-		cutoffs[0] = math.Inf(-1)
 		// Choose all possible cutoffs for unique values in y.
-		bin := 1
+		var bin int
```
kortschak, line 62: `bin := 0`

stat/roc_test.go
```diff
@@ -210,6 +210,14 @@ func TestROC(t *testing.T) {
 			wantFPR:    nil,
 			wantThresh: nil,
 		},
+		{ // 23
```
kortschak, line 213: Can you add a test case for #1559.

```
		{ // 24
			y:          []float64{0.1, 0.35, 0.4, 0.8},
			c:          []bool{true, false, true, false},
			cutoffs:    []float64{-1, 0, 0.1, 0.36, 0.8},
			wantTPR:    []float64{...
			wantFPR:    []float64{...
			wantThresh: []float64{...
		},
```

stat/roc.go
Armadilloa16: Yep

stat/roc.go
Armadilloa16: Yep

stat/roc.go
Armadilloa16: Yeah that makes sense.

stat/roc.go
Armadilloa16: Ok

stat/roc_test.go
Armadilloa16: Done.

stat/roc.go
```diff
@@ -98,14 +98,22 @@ func ROC(cutoffs, y []float64, classes []bool, weights []float64) (tpr, fpr, thr
...
+	lastFpr := fpr[bin]
+	for bin++; bin < len(cutoffs); bin++ {
+		tpr[bin] = lastTpr
+		fpr[bin] = lastFpr
+	}
 
 	invNeg := 1 / nNeg
 	invPos := 1 / nPos
+	// Convert negative counts to TPR and FPR.
 	for i := range tpr {
```
kortschak, line 117: Out of curiosity (and laziness), does [my approach](https://github.com/gonum/gonum/pull/1558/files#diff-07d8944b2b9171f3dfbb983bfc016c30c93b829e6d2b2c4c91b21de24f3c8dccR109) used in #1558 work? It's simpler than this and does less work.

Armadilloa16: I could be wrong, but I think the case for which it wouldn't work is when there are more cutoffs above the maximum y value, I added a test case for this case (test case 26). I agree your solution is prettier though.

kortschak: Just tested with your new cases and yes, this simpler approach is correct.

Armadilloa16: Huh, yes I see what you mean now. This is confusing, but I believe it is actually equivalent. How about I incorporate you're simpler approach, but also add a comment to try and explain what it's doing?

kortschak: SGTM. I would say,

```
// Convert negative counts to TPR and FPR for non-zero bins.
```

Note that the reason this works is that these bins are zero in the final result, but `make` has already initialised the elements to be zero. So we just work on bins that are non-zero and use the existing zeros for the final result.

stat/roc.go
```diff
@@ -104,17 +104,13 @@ func ROC(cutoffs, y []float64, classes []bool, weights []float64) (tpr, fpr, thr
...
 		}
 	}
-	lastTpr := tpr[bin]
-	lastFpr := fpr[bin]
-	for bin++; bin < len(cutoffs); bin++ {
-		tpr[bin] = lastTpr
-		fpr[bin] = lastFpr
-	}
+	// here bins beyond the maximum value in y are skipped, and the
+	// fpr and tpr are set to zero below by ommission.
```
kortschak: I'd put this after the comment with the for loop.

```
	// Bins beyond the maximum value in y are skipped
	//  and the fpr and tpr elements are left as zero.
```

stat/roc.go
Armadilloa16: Yeah, cool. It's always an interesting tradeoff between efficiency and clarity. I suppose that's the point of comments. You happy for me to squash all these commit's together now?

kortschak: Yes, please. Note that this is not strictly a reversion. I'd say "stat: fix ROC cutoffs".

Do you want/need to add UWA to AUTHORS (separate commit)?

stat/roc.go
Armadilloa16: Done (?).

kortschak: LGTM

stat/roc.go
Armadilloa16: Sounds good. 

Hmm, no need to add UWA but maybe I should change the contact to my personal gmail, because I don't have access to the adelaide email address anymore... good point.

kortschak: OK. Make that a new PR. It would just be adding your personal email address to the lines in A+C.

stat/roc.go
```diff
@@ -98,15 +98,19 @@ func ROC(cutoffs, y []float64, classes []bool, weights []float64) (tpr, fpr, thr
...
 		nNeg += negWeight
-		if y[i] <= cutoffs[bin] {
+		// Count false negatives (in tpr) and true negatives (in fpr).
+		if y[i] < cutoffs[bin] {
 			tpr[bin] += posWeight
 			fpr[bin] += negWeight
 		}
 	}
+	// Bins beyond the maximum value in y are skipped
+	// leaving these fpr and tpr elements as zero.
```
kortschak, line 108: Sorry, this should be after L112, just before the for loop.

Armadilloa16: Oh yeah that makes sense

kortschak (approved):
Thanks

kortschak:
Do you need me to merge?

Armadilloa16:
Sure, I don't think I have merge rights. All looks good to me.

PR #959, opened 2019-04-14
stat: define ROC cutoffs instead of using number of cutoffs

kortschak:
Please take a look.

Closes #708.
Fixes #316.
Closes #960.

PR #1563, opened 2021-02-09
stat: add TOC function

kortschak:
Please take a look.

Tests are light because there is little to test.

/cc @Armadilloa16

### DIFF
--- a/stat/roc.go
+++ b/stat/roc.go
@@ -7,6 +7,8 @@ package stat
 import (
 	"math"
 	"sort"
+
+	"gonum.org/v1/gonum/floats"
 )
 
 // ROC returns paired false positive rate (FPR) and true positive rate
@@ -131,34 +133,63 @@ func ROC(cutoffs, y []float64, classes []bool, weights []float64) (tpr, fpr, thr
 // TOC returns the Total Operating Characteristic for the classes provided
 // and the minimum and maximum bounds for the TOC.
 //
-// The input classes should be sorted according to the rank variable.
-// SortWeightedLabeled can be used to sort classes together with the rank
-// variable. The returned ntp values can be interpreted as true positives
-// where values at or above the given rank are assigned class true for each
-// given rank from 1 to len(classes). The values of min and max provide the
-// minimum and maximum possible number of true values for the set of classes.
+// The input classes should be sorted with their weights according to the
+// rank variable. SortWeightedLabeled can be used to sort classes together
+// with weights by the rank variable.
+//
+// The returned ntp values can be interpreted as true positives where values
+// at or above the given rank are assigned class true for each given rank
+// from 1 to len(classes). The values of min and max provide the minimum and
+// maximum possible number of true values for the set of classes.
+//
+// If weights is nil, all weights are treated as 1. If weights is not nil
+// it must have the same length as classes, otherwise TOC will panic.
 //
 // More details about TOC curves are available at
 // https://en.wikipedia.org/wiki/Total_operating_characteristic
-func TOC(classes []bool) (min, ntp, max []float64) {
+func TOC(classes []bool, weights []float64) (min, ntp, max []float64) {
+	if weights != nil && len(classes) != len(weights) {
+		panic("stat: slice length mismatch")
+	}
 	if len(classes) == 0 {
 		return nil, nil, nil
 	}
+
 	ntp = make([]float64, len(classes)+1)
+	min = make([]float64, len(ntp))
+	max = make([]float64, len(ntp))
+	if weights == nil {
+		for i := range ntp[1:] {
+			ntp[i+1] = ntp[i]
+			if classes[i] {
+				ntp[i+1] += 1
+			}
+		}
+		totalPositive := ntp[len(ntp)-1]
+		for i := range min {
+			min[i] = math.Max(0, totalPositive-float64(len(classes)-i))
+		}
+		for i := range max {
+			max[i] = math.Min(totalPositive, float64(i))
+		}
+		return min, ntp, max
+	}
+
 	for i := range ntp[1:] {
 		ntp[i+1] = ntp[i]
 		if classes[i] {
-			ntp[i+1] += 1
+			ntp[i+1] += weights[i]
 		}
 	}
+	cumw := make([]float64, len(ntp))
+	floats.CumSum(cumw[1:], weights)
+	totw := cumw[len(cumw)-1]
 	totalPositive := ntp[len(ntp)-1]
-	min = make([]float64, len(ntp))
 	for i := range min {
-		min[i] = math.Max(0, totalPositive-float64(len(classes)-i))
+		min[i] = math.Max(0, totalPositive-(totw-cumw[i]))
 	}
-	max = make([]float64, len(ntp))
 	for i := range max {
-		max[i] = math.Min(totalPositive, float64(i))
+		max[i] = math.Min(totalPositive, cumw[i])
 	}
 	return min, ntp, max
 }
--- a/stat/roc.go
+++ b/stat/roc.go
@@ -134,8 +134,8 @@ func ROC(cutoffs, y []float64, classes []bool, weights []float64) (tpr, fpr, thr
 // and the minimum and maximum bounds for the TOC.
 //
 // The input classes should be sorted with their weights according to the
-// rank variable. SortWeightedLabeled can be used to sort classes together
-// with weights by the rank variable.
+// rank variable ascending. SortWeightedLabeled can be used to sort classes
+// together with weights by the rank variable.
 //
 // The returned ntp values can be interpreted as true positives where values
 // at or above the given rank are assigned class true for each given rank
@@ -161,7 +161,7 @@ func TOC(classes []bool, weights []float64) (min, ntp, max []float64) {
 	if weights == nil {
 		for i := range ntp[1:] {
 			ntp[i+1] = ntp[i]
-			if classes[i] {
+			if !classes[i] {
 				ntp[i+1] += 1
 			}
 		}
@@ -177,7 +177,7 @@ func TOC(classes []bool, weights []float64) (min, ntp, max []float64) {
 
 	for i := range ntp[1:] {
 		ntp[i+1] = ntp[i]
-		if classes[i] {
+		if !classes[i] {
 			ntp[i+1] += weights[i]
 		}
 	}
--- a/stat/roc.go
+++ b/stat/roc.go
@@ -127,3 +127,38 @@ func ROC(cutoffs, y []float64, classes []bool, weights []float64) (tpr, fpr, thr
 
 	return tpr, fpr, cutoffs
 }
+
+// TOC returns the Total Operating Characteristic for the classes provided
+// and the minimum and maximum bounds for the TOC.
+//
+// The input classes should be sorted according to the rank variable.
+// SortWeightedLabeled can be used to sort classes together with the rank
+// variable. The returned ntp values can be interpreted as true positives
+// where values at or above the given rank are assigned class true for each
+// given rank from 1 to len(classes). The values of min and max provide the
+// minimum and maximum possible number of true values for the set of classes.
+//
+// More details about TOC curves are available at
+// https://en.wikipedia.org/wiki/Total_operating_characteristic
+func TOC(classes []bool) (min, ntp, max []float64) {
+	if len(classes) == 0 {
+		return nil, nil, nil
+	}
+	ntp = make([]float64, len(classes)+1)
+	for i := range ntp[1:] {
+		ntp[i+1] = ntp[i]
+		if classes[i] {
+			ntp[i+1] += 1
+		}
+	}
+	totalPositive := ntp[len(ntp)-1]
+	min = make([]float64, len(ntp))
+	for i := range min {
+		min[i] = math.Max(0, totalPositive-float64(len(classes)-i))
+	}
+	max = make([]float64, len(ntp))
+	for i := range max {
+		max[i] = math.Min(totalPositive, float64(i))
+	}
+	return min, ntp, max
+}
--- a/stat/roc.go
+++ b/stat/roc.go
@@ -140,8 +140,13 @@ func ROC(cutoffs, y []float64, classes []bool, weights []float64) (tpr, fpr, thr
 //
 // The returned ntp values can be interpreted as the number of true positives
 // where values above the given rank are assigned class true for each given
-// rank from 1 to len(classes). The values of min and max provide the minimum
-// and maximum possible number of true values for the set of classes.
+// rank from 1 to len(classes).
+//  ntp_i = sum_{j = 0}^{i} [ !classes_j ] * weight_j
+// The values of min and max provide the minimum and maximum possible number
+// of false values for the set of classes. The first element of ntp, min and
+// max are always zero as this corresponds to assigning all data class false
+// and the last elements are always weighted sum of classes as this corresponds
+// to assigning every data class false.
 //
 // If weights is nil, all weights are treated as 1. When weights are not nil,
 // the calculation of min and max allows for partial assignment of single data
@@ -165,14 +170,12 @@ func TOC(classes []bool, weights []float64) (min, ntp, max []float64) {
 		for i := range ntp[1:] {
 			ntp[i+1] = ntp[i]
 			if !classes[i] {
-				ntp[i+1] += 1
+				ntp[i+1]++
 			}
 		}
 		totalPositive := ntp[len(ntp)-1]
-		for i := range min {
+		for i := range ntp {
 			min[i] = math.Max(0, totalPositive-float64(len(classes)-i))
-		}
-		for i := range max {
 			max[i] = math.Min(totalPositive, float64(i))
 		}
 		return min, ntp, max
@@ -188,10 +191,8 @@ func TOC(classes []bool, weights []float64) (min, ntp, max []float64) {
 	floats.CumSum(cumw[1:], weights)
 	totw := cumw[len(cumw)-1]
 	totalPositive := ntp[len(ntp)-1]
-	for i := range min {
+	for i := range ntp {
 		min[i] = math.Max(0, totalPositive-(totw-cumw[i]))
-	}
-	for i := range max {
 		max[i] = math.Min(totalPositive, cumw[i])
 	}
 	return min, ntp, max
--- a/stat/roc.go
+++ b/stat/roc.go
@@ -136,7 +136,7 @@ func ROC(cutoffs, y []float64, classes []bool, weights []float64) (tpr, fpr, thr
 // The input y values that correspond to classes and weights must be sorted
 // in ascending order. classes[i] is the class of value y[i] and weights[i]
 // is the weight of y[i]. SortWeightedLabeled can be used to sort classes
-// together with weights by the rank variable.
+// together with weights by the rank variable, i+1.
 //
 // The returned ntp values can be interpreted as the number of true positives
 // where values above the given rank are assigned class true for each given
--- a/stat/roc.go
+++ b/stat/roc.go
@@ -146,7 +146,7 @@ func ROC(cutoffs, y []float64, classes []bool, weights []float64) (tpr, fpr, thr
 // of false values for the set of classes. The first element of ntp, min and
 // max are always zero as this corresponds to assigning all data class false
 // and the last elements are always weighted sum of classes as this corresponds
-// to assigning every data class false.
+// to assigning every data class true.
 //
 // If weights is nil, all weights are treated as 1. When weights are not nil,
 // the calculation of min and max allows for partial assignment of single data
--- a/stat/roc.go
+++ b/stat/roc.go
@@ -141,7 +141,9 @@ func ROC(cutoffs, y []float64, classes []bool, weights []float64) (tpr, fpr, thr
 // The returned ntp values can be interpreted as the number of true positives
 // where values above the given rank are assigned class true for each given
 // rank from 1 to len(classes).
-//  ntp_i = sum_{j = 0}^{i} [ !classes_j ] * weight_j
+//  ntp_i = sum_{j = 1}^{i} [ !classes_{j-1} ] * weights_{j-1}
+// and
+//  ntp_0 = 0
 // The values of min and max provide the minimum and maximum possible number
 // of false values for the set of classes. The first element of ntp, min and
 // max are always zero as this corresponds to assigning all data class false
--- a/stat/roc.go
+++ b/stat/roc.go
@@ -141,7 +141,7 @@ func ROC(cutoffs, y []float64, classes []bool, weights []float64) (tpr, fpr, thr
 // The returned ntp values can be interpreted as the number of true positives
 // where values above the given rank are assigned class true for each given
 // rank from 1 to len(classes).
-//  ntp_i = sum_{j = 1}^{i} [ !classes_{j-1} ] * weights_{j-1}
+//  ntp_i = sum_{j = 1}^{i} [ !classes_{j-1} ] * weights_{j-1}, where [x] = 1 if x else 0.
 // and
 //  ntp_0 = 0
 // The values of min and max provide the minimum and maximum possible number
--- a/stat/roc.go
+++ b/stat/roc.go
@@ -139,9 +139,9 @@ func ROC(cutoffs, y []float64, classes []bool, weights []float64) (tpr, fpr, thr
 // together with weights by the rank variable.
 //
 // The returned ntp values can be interpreted as the number of true positives
-// where values at or above the given rank are assigned class true for each
-// given rank from 1 to len(classes). The values of min and max provide the
-// minimum and maximum possible number of true values for the set of classes.
+// where values above the given rank are assigned class true for each given
+// rank from 1 to len(classes). The values of min and max provide the minimum
+// and maximum possible number of true values for the set of classes.
 //
 // If weights is nil, all weights are treated as 1. If weights is not nil
 // it must have the same length as classes, otherwise TOC will panic.
--- a/stat/roc.go
+++ b/stat/roc.go
@@ -143,8 +143,10 @@ func ROC(cutoffs, y []float64, classes []bool, weights []float64) (tpr, fpr, thr
 // rank from 1 to len(classes). The values of min and max provide the minimum
 // and maximum possible number of true values for the set of classes.
 //
-// If weights is nil, all weights are treated as 1. If weights is not nil
-// it must have the same length as classes, otherwise TOC will panic.
+// If weights is nil, all weights are treated as 1. When weights are not nil,
+// the calculation of min and max allows for partial assignment of single data
+// points. If weights is not nil it must have the same length as classes,
+// otherwise TOC will panic.
 //
 // More details about TOC curves are available at
 // https://en.wikipedia.org/wiki/Total_operating_characteristic
--- a/stat/roc.go
+++ b/stat/roc.go
@@ -7,8 +7,6 @@ package stat
 import (
 	"math"
 	"sort"
-
-	"gonum.org/v1/gonum/floats"
 )
 
 // ROC returns paired false positive rate (FPR) and true positive rate
@@ -141,7 +139,7 @@ func ROC(cutoffs, y []float64, classes []bool, weights []float64) (tpr, fpr, thr
 // The returned ntp values can be interpreted as the number of true positives
 // where values above the given rank are assigned class true for each given
 // rank from 1 to len(classes).
-//  ntp_i = sum_{j = 1}^{i} [ !classes_{j-1} ] * weights_{j-1}, where [x] = 1 if x else 0.
+//  ntp_i = sum_{j = 1}^{len(ntp)-i-1} [ classes_j ] * weights_j, where [x] = 1 if x else 0.
 // and
 //  ntp_0 = 0
 // The values of min and max provide the minimum and maximum possible number
@@ -171,7 +169,7 @@ func TOC(classes []bool, weights []float64) (min, ntp, max []float64) {
 	if weights == nil {
 		for i := range ntp[1:] {
 			ntp[i+1] = ntp[i]
-			if !classes[i] {
+			if classes[len(classes)-i-1] {
 				ntp[i+1]++
 			}
 		}
@@ -183,14 +181,15 @@ func TOC(classes []bool, weights []float64) (min, ntp, max []float64) {
 		return min, ntp, max
 	}
 
+	cumw := max // Reuse max for cumulative weight. Update its elements last.
 	for i := range ntp[1:] {
 		ntp[i+1] = ntp[i]
-		if !classes[i] {
-			ntp[i+1] += weights[i]
+		w := weights[len(weights)-i-1]
+		cumw[i+1] = cumw[i] + w
+		if classes[len(classes)-i-1] {
+			ntp[i+1] += w
 		}
 	}
-	cumw := make([]float64, len(ntp))
-	floats.CumSum(cumw[1:], weights)
 	totw := cumw[len(cumw)-1]
 	totalPositive := ntp[len(ntp)-1]
 	for i := range ntp {
--- a/stat/roc.go
+++ b/stat/roc.go
@@ -133,14 +133,15 @@ func ROC(cutoffs, y []float64, classes []bool, weights []float64) (tpr, fpr, thr
 // TOC returns the Total Operating Characteristic for the classes provided
 // and the minimum and maximum bounds for the TOC.
 //
-// The input classes should be sorted with their weights according to the
-// rank variable ascending. SortWeightedLabeled can be used to sort classes
+// The input y values that correspond to classes and weights must be sorted
+// in ascending order. classes[i] is the class of value y[i] and weights[i]
+// is the weight of y[i]. SortWeightedLabeled can be used to sort classes
 // together with weights by the rank variable.
 //
-// The returned ntp values can be interpreted as true positives where values
-// at or above the given rank are assigned class true for each given rank
-// from 1 to len(classes). The values of min and max provide the minimum and
-// maximum possible number of true values for the set of classes.
+// The returned ntp values can be interpreted as the number of true positives
+// where values at or above the given rank are assigned class true for each
+// given rank from 1 to len(classes). The values of min and max provide the
+// minimum and maximum possible number of true values for the set of classes.
 //
 // If weights is nil, all weights are treated as 1. If weights is not nil
 // it must have the same length as classes, otherwise TOC will panic.
--- a/stat/roc_example_test.go
+++ b/stat/roc_example_test.go
@@ -108,7 +108,7 @@ func ExampleROC_equallySpacedCutoffs() {
 	// false positive rate: [0 0 0 1 1 1 1 1 1]
 }
 
-func ExampleROC_aUC() {
+func ExampleROC_aUC_unweighted() {
 	y := []float64{0.1, 0.35, 0.4, 0.8}
 	classes := []bool{true, false, true, false}
 
@@ -124,6 +124,22 @@ func ExampleROC_aUC() {
 	// true  positive rate: [0 0 0.5 0.5 1]
 	// false positive rate: [0 0.5 0.5 1 1]
 	// auc: 0.25
+}
+
+func ExampleROC_aUC_weighted() {
+	y := []float64{0.1, 0.35, 0.4, 0.8}
+	classes := []bool{true, false, true, false}
+	weights := []float64{1, 2, 2, 1}
+
+	tpr, fpr, _ := stat.ROC(nil, y, classes, weights)
+
+	// Compute Area Under Curve.
+	auc := integrate.Trapezoidal(fpr, tpr)
+	fmt.Printf("auc: %f\n", auc)
+
+	// Output:
+	// auc: 0.444444
+
 }
 
 func ExampleTOC() {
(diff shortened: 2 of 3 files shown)


Armadilloa16:
Ahh yes, this makes sense. The danger of TOC is misuse --- people tend to misunderstand that the point of the ROC is that it is the prevalence-independent quantity, and that although TOC has it's advantages for describing data, it is prevalence dependant when applied to populations. I.e. if the ratio of true to false in classes is different between your sample and your population, TOC does not compensate for that while ROC does. Of course all that aside, yes we can make a function to calculate it of course. I'll have a look at the code later today.

kortschak:
I'd happily take advice on appropriate doc comments for this.

Armadilloa16:
Yeah, I'm not sure if the docs are a relevant place for that (?) because its not really to do with what a TOC is or how the function behaves, but rather on the decision-making of if to use it or ROC.

stat/roc.go
```diff
@@ -126,3 +126,32 @@ func ROC(cutoffs, y []float64, classes []bool, weights []float64) (tpr, fpr, thr
...
 	return tpr, fpr, cutoffs
 }
+
+// TOC returns the Total Operating Characteristic for the classes provided,
+// which should be a boolean variable ordered by a rank variable. TOC also
+// returns the minimum and maximum bounds for the TOC.
+//
+// More details about TOC curves are available at
+// https://en.wikipedia.org/wiki/Total_operating_characteristic
+func TOC(classes []bool) (min, toc, max []float64) {
```
Armadilloa16, line 136: It's a technicality, but I would call the toc' output something like ntp' for number of true positives, as the "TOC" refers to the curve, y- and x- coordinates, while this slice is just the y-coordinates. I don't feel strongly about it though, `toc' is fine too.

stat/roc.go
```diff
@@ -126,3 +126,32 @@ func ROC(cutoffs, y []float64, classes []bool, weights []float64) (tpr, fpr, thr
 
 	return tpr, fpr, cutoffs
 }
+
+// TOC returns the Total Operating Characteristic for the classes provided,
+// which should be a boolean variable ordered by a rank variable. TOC also
+// returns the minimum and maximum bounds for the TOC.
+//
```
Armadilloa16, line 133: // The input classes should be sorted according to the rank variable, if this is the case, the toc values can be interpreted as true positives where values above or equal to the given rank are assigned class true for each given rank from 1 to len(classes)+1.

Armadilloa16: // SortWeightedLabeled can be used to sort classes together with the rank variable.

Armadilloa16: Also, could include weights, in which case copy comment for SortWeightedLabeled from ROC.

stat/roc.go
```diff
@@ -126,3 +126,32 @@ func ROC(cutoffs, y []float64, classes []bool, weights []float64) (tpr, fpr, thr
...
+	for i := range toc[1:] {
+		toc[i+1] = toc[i]
+		if classes[i] {
+			toc[i+1] += 1
+		}
+	}
+	last := toc[len(toc)-1]
+	min = make([]float64, len(toc))
+	for i := range min {
+		min[i] = math.Max(0, last+float64(i-len(classes)))
```
Armadilloa16, line 150: why do this as a float operation rather than an int operation? These should both be int's. That said, if we add functionality to allow weights, then this will need to be a float operation but in that case it would be a different total not `len(classes)'.

Armadilloa16: Also, here `last + i - len(classes)' would make more sense to me as `i - (len(classes) - last)' I realise this is doing slightly more work (?) but just though I'd mention it has a more obvious meaning. Maybe a comment here explaining that `last' is actually the number of positives in the data, and so `len(classes) - last' is the number of negatives in the data, while i is the number of data assigned positive.

stat/roc_example_test.go
```diff
@@ -125,3 +125,23 @@ func ExampleROC_aUC() {
...
+	min, toc, max := stat.TOC(classes)
+	fmt.Printf("minimum bound: %v\n", min)
+	fmt.Printf("TOC:           %v\n", toc)
+	fmt.Printf("maximum bound: %v\n", max)
+
+	// Output:
+	// minimum bound: [0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 1 2 3 4 5 6 7 8 9 10]
+	// TOC:           [0 0 1 1 1 2 2 3 3 3 4 5 6 7 8 9 10 10 10 10 10 10 10 10 10 10 10 10 10 10 10]
+	// maximum bound: [0 1 2 3 4 5 6 7 8 9 10 10 10 10 10 10 10 10 10 10 10 10 10 10 10 10 10 10 10 10 10]
+}
```
Armadilloa16, line 147: I would add an example of using SortWeightedLabeled or SortLabeled as in the ROC examples.

stat/roc_test.go
```diff
@@ -256,3 +256,41 @@ func TestROC(t *testing.T) {
...
+				false, true, false, false, true, false,
+				true, false, false, true, true, true,
+				true, true, true, true, false, false,
+				false, false, false, false, false, false,
+				false, false, false, false, false, false,
+			},
+			wantMin: []float64{0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10},
+			wantMax: []float64{0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 10, 10, 10, 10, 10, 10, 10, 10, 10, 10, 10, 10, 10, 10, 10, 10, 10, 10, 10, 10},
+			wantTOC: []float64{0, 0, 1, 1, 1, 2, 2, 3, 3, 3, 4, 5, 6, 7, 8, 9, 10, 10, 10, 10, 10, 10, 10, 10, 10, 10, 10, 10, 10, 10, 10},
+		},
```
Armadilloa16, line 282: I would include tests for all true, all false, len(classes) = 0, etc. similar to ROC

Armadilloa16:
So one question is adding weights (simple enough to do and people always seem to want that). The other question I have is about the lack of functionality to specify cutoffs. Similar to the original UI for ROC, which then later got modified to allow for cutoffs to be specified. So the question is essentially: is it intentional that TOC have a different UI than ROC? do they have different use-cases? otherwise I'd like to see them share a similar signature, so although you don't need to provide the ranking variable (the same could be said about ROC, we could remove the input `y`), we should probably add an input for `weights` and `cutoffs`, this would still be the behaviour when those inputs are `nil` of course. Then maybe we could revisit removing `y` as an input from ROC, or add a ranking variable here. Either way. Thoughts?

stat/roc.go
kortschak: Yeah, I was wondering about that. How would be the best way to include weights in this?

stat/roc.go
kortschak: It's floats because it will get used as floats later and we don't have (thankfully) vector coercion.

kortschak: I've reorganised, but repeated reconversion is not a thing I want to do.

Armadilloa16: that's fair, especially since we'll probably introduce weights. Without weights, all three outputs (min, toc, max) could all be []int though at least as far as the scope of this function is concerned, but that's probably not important.

stat/roc.go
Armadilloa16: I can have a go at it if you like, and then send the changes to your branch?

stat/roc_test.go
kortschak: Done

stat/roc_example_test.go
kortschak: Done

stat/roc.go
kortschak: Done

kortschak:
ISTM that `cutoffs` are less valuable here while they are an intrinsic part of ROC curves. I'm happy to have weights, it just wasn't immediately obvious to me from the paper how that would be handled. It's OK for the two to have different signatures, though having `y` in `TOC` would give the (potentially misleading) sense that we're looking after the user and making sure the data are correctly sorted. (It's not at all clear to me how removing `y` from `ROC` would work).

kortschak:
PTAL

kortschak:
Oh, really! Weighted TOC is just adding non-unit counts based on `weights`.

stat/roc.go
kortschak: I just had a go and it gives unintuitive results for the max bound. I put the commit, but I'm not sure.

stat/roc.go
```diff
@@ -139,17 +139,31 @@ func ROC(cutoffs, y []float64, classes []bool, weights []float64) (tpr, fpr, thr
...
+				ntp[i+1] += 1
+			}
+		}
+	} else {
+		for i := range ntp[1:] {
+			ntp[i+1] = ntp[i]
+			if classes[i] {
+				ntp[i+1] += weights[i]
+			}
 		}
```
Armadilloa16, line 164: Hah, I just finished doing this, I look back and I see you've done it. Take a look at how I did it and tell me what you think? 
https://github.com/Armadilloa16/gonum/commit/a6176817cab5793242f25706c21bc637951fceac

Armadilloa16: In particular, with weights the calculation of min and max need to be adjusted as well.

kortschak: Yeah, that explains why I didn't understand mine.

Armadilloa16:
Indeed --- and Weighted ROC is exactly the same, just uses non unit counts based on weights. 

Cutoffs are just as intrinsically a part of TOC as they are ROC, in exactly the same way. In ROC the `y` serves to rank the observations, and that is it's only function. If you'd like I could make a seperate PR to suggest changes to ROC to remove `y` if nothing else than just to show how it would look. TOC and ROC are so similar, I think it would be really strange to construct such different UIs for them. 

Apart from adjusting the calculation of min and max for weights, all looks good to me. I can check the tests and examples more carefully as well if you like but I don't think I can approve a review since I don't have write access.

kortschak:
OK, I've fixed the weighted case, and now properly adjust the min and max slices. Please take a look at the tests. An informal review is OK (GitHub has this capacity).

For the cutoffs, the two paths are 1. to file an issue to remove `y` from ROC or 2. to add `cutoffs` here (or in a follow-up PR).

stat/roc.go
```diff
@@ -126,3 +128,62 @@ func ROC(cutoffs, y []float64, classes []bool, weights []float64) (tpr, fpr, thr
 
 	return tpr, fpr, cutoffs
 }
+
+// TOC returns the Total Operating Characteristic for the classes provided
+// and the minimum and maximum bounds for the TOC.
+//
+// The input classes should be sorted according to the rank variable.
+// SortWeightedLabeled can be used to sort classes together with the rank
+// variable. The returned ntp values can be interpreted as true positives
```
Armadilloa16, line 137: together with weights by the rank variable

stat/roc.go
```diff
@@ -126,3 +128,62 @@ func ROC(cutoffs, y []float64, classes []bool, weights []float64) (tpr, fpr, thr
 
 	return tpr, fpr, cutoffs
 }
+
+// TOC returns the Total Operating Characteristic for the classes provided
+// and the minimum and maximum bounds for the TOC.
+//
```
Armadilloa16, line 134: // If weights is nil, all weights are treated as 1.

stat/roc.go
kortschak: Yes. Out of practice.

stat/roc.go
kortschak: Done, and a warning about the panic (this needs to be added to ROC as well).

kortschak: And.... I've just discovered through a fumble that GitHub's branch protection doesn't work. The fix for ROC is now accidentally merged without a PR.

stat/roc_example_test.go
```diff
@@ -125,3 +125,41 @@ func ExampleROC_aUC() {
...
+	min, ntp, max := stat.TOC(classes, weights)
+	fmt.Printf("minimum bound: %v\n", min)
+	fmt.Printf("TOC:           %v\n", ntp)
+	fmt.Printf("maximum bound: %v\n", max)
+
+	// Output:
+	// minimum bound: [0 0 0 1 4 6 8]
+	// TOC:           [0 0 1 1 4 6 8]
+	// maximum bound: [0 4 5 8 8 8 8]
+}
```
Armadilloa16, line 165: There is an issue here, you've sorted in increasing order of y, which makes this inconsistent with ROC and the documentation... we should clarify. If TOC is used in the fashsion shown in this example, then the interpretation would be consistent with the original version of ROC (where values LESS than or equal to a particular y cutoff are classified as "positive", this was later changed (parity reversed) to be consistent with sklearn so maybe we should be consistent to that. All you would need to do then is to specify that the classes should be ranked in _increasing_ order of the ranking variable y and if this is done then the interpretation of the output would be consistent with sklearn.

Armadilloa16:
Aside from the unsorted example (should clarify in the documentation as well), all LGTM. I'll have a go at making a version of ROC sans `y` and `cutoffs`.

Armadilloa16 (changes requested):

stat/roc_example_test.go
kortschak: Struggling with the wording for this. So pushing the logic changes.

Armadilloa16: I'm struggling with the wording too, my brain is getting a bit mushy today I might need to come back to this another day if that's alright with you.

kortschak: SGTM

Armadilloa16: Alright so I've thought about this abit more (sorry for the delay), and the issue is this: if there is some ranking variable y, and you sort the classes and weights in increasing order of y, this is in decreasing order of rank. I.e. the smallest value is y has the largest rank. So I think what we should do about it is clarify in the documentation. The issue originates from the phrase "where values at or above the given rank are assigned class true" because this is different to "where values at or above the given value of the ranking variable are assigned class true". One easy fix would be to just specify that we assign the smallest value in the ranking variable y rank "1", and the largest rank "len(y)". Then the interpretation as ">= a given rank" is the same as ">= a given value of the ranking variable". Does that make sense?

stat/roc.go
```diff
@@ -126,3 +128,67 @@ func ROC(cutoffs, y []float64, classes []bool, weights []float64) (tpr, fpr, thr
...
+		return nil, nil, nil
+	}
+
+	ntp = make([]float64, len(classes)+1)
+	min = make([]float64, len(ntp))
+	max = make([]float64, len(ntp))
+	if weights == nil {
+		for i := range ntp[1:] {
+			ntp[i+1] = ntp[i]
+			if !classes[i] {
```
Armadilloa16, line 163: When you implemented the "parity switch" you changed this `classes[i]` to `!classes[i]`, I don't think this is correct --- it should still be `classes[i]`, otherwise this is no longer calculating true positives at all. The confusion is to do with the ranking variable vs the ranks themselves as I discuss in the other thread.

stat/roc_example_test.go
Armadilloa16: Alternatively, we could leave it as is and just add the note that sorting in ascending order of the ranking variable is the same as in decending order of the rank variable. In which case that comment is fine.

Armadilloa16:
Really the question is about the intended behaviour. I think the best solution is to just clarify that increasing order of the ranking variable is decreasing order of rank, that way the original implementation (with `classes[i]` not `!classes[i]` is correct... I don't think the `!classes[i]` implementation is correct either way but I am willing to be convinced otherwise.

stat/roc.go
kortschak: This matches the changes in [#959](https://github.com/gonum/gonum/pull/959/files#diff-07d8944b2b9171f3dfbb983bfc016c30c93b829e6d2b2c4c91b21de24f3c8dccR96). If this is wrong, so is that. This was part of the reason for making the AUC examples for the TOC and extending the AUC examples to cover the weighted case; they must match.

stat/roc_example_test.go
kortschak: > One easy fix would be to just specify that we assign the smallest value in the ranking variable y rank "1", and the largest rank "len(y)".

This is the natural interpretation. I though that was known.

kortschak:
*ping* @Armadilloa16

stat/roc_example_test.go
Armadilloa16: Yeah, ok fair enough. It confused me though, but maybe that's because I'm over-thinking it.

stat/roc.go
Armadilloa16: Hmm, I need to think this through again from the beggining...

Armadilloa16:
Alright, well there are a couple of details I feel like we should revisit at some point (AUC calculation, and conventions for parity etc.) but otherwise LGTM for now.

kortschak:
I'm going to leave this until after the 0.9 branch is cut since there still seems to be enough confusion. I don't want changes going in that will likely only last a single minor version.

kortschak:
@romanwerpachowski Do you want to take a quick look at this?

stat/roc.go
```diff
@@ -127,3 +129,67 @@ func ROC(cutoffs, y []float64, classes []bool, weights []float64) (tpr, fpr, thr
...
 }
+
+// TOC returns the Total Operating Characteristic for the classes provided
+// and the minimum and maximum bounds for the TOC.
+//
+// The input classes should be sorted with their weights according to the
+// rank variable ascending. SortWeightedLabeled can be used to sort classes
+// together with weights by the rank variable.
+//
+// The returned ntp values can be interpreted as true positives where values
```
romanwerpachowski, line 140: What are "ntp values"?

romanwerpachowski, line 140: or "as numbers of true positives"?

stat/roc.go
```diff
@@ -127,3 +129,67 @@ func ROC(cutoffs, y []float64, classes []bool, weights []float64) (tpr, fpr, thr
 
 	return tpr, fpr, cutoffs
 }
+
+// TOC returns the Total Operating Characteristic for the classes provided
+// and the minimum and maximum bounds for the TOC.
+//
+// The input classes should be sorted with their weights according to the
+// rank variable ascending. SortWeightedLabeled can be used to sort classes
```
romanwerpachowski, line 137: This is confusing for me. The rank variable itself relates to the order of data in the sample and where a particular value fits in it. I think it would be *much* clearer for the user if we just said sth like "y values must be sorted in ascending order, classes[i] is the class of value y[i] and weights[i] is the weight of y[i]", even at the cost of some verbosity.

stat/roc.go
kortschak: Number of true positives.

kortschak: Yes, I'll add that.

stat/roc.go
kortschak: Done.

stat/roc.go
romanwerpachowski: The docstring _here_ says "where values at or above the given rank are assigned class true". The discussion at #316 says "ROC treats values below the cutoff as true (rather than false)", leading me to think that the docstring here is out of sync with the code.

kortschak: Yes, this reflects the doc comment for ROC, which I neglected to invert when I inverted the logic. That should be fixed as well (separate PR). Do you agree?

romanwerpachowski: I disagree. I think the docstring for the new function should be fixed now, to avoid submitting a description which we know is incorrect just for symmetry's sake. The ROC comment can be fixed in another PR.

kortschak: To explain more clearly, the ROC doc comment should be fixed in another PR. I'm still not convinced that this PR should go into v0.9, but a fix for the ROC docs should. An inversion of the TOC doc comment should go in this PR.

romanwerpachowski: I agree :)

kortschak: OK, I think that's fixed here too.

stat/roc.go
```diff
@@ -127,3 +129,68 @@ func ROC(cutoffs, y []float64, classes []bool, weights []float64) (tpr, fpr, thr
...
+
+// TOC returns the Total Operating Characteristic for the classes provided
+// and the minimum and maximum bounds for the TOC.
+//
+// The input y values that correspond to classes and weights must be sorted
+// in ascending order. classes[i] is the class of value y[i] and weights[i]
+// is the weight of y[i]. SortWeightedLabeled can be used to sort classes
+// together with weights by the rank variable.
+//
+// The returned ntp values can be interpreted as the number of true positives
```
Armadilloa16, line 141: I think I can explain why I think this is confusing by re-phrasing it to be more specific: `ntp[i] can be interpreted as the number of true positives where values with ranks greater than (len(classes) - i) are assigned class true for each i from 0 to len(classes).` Optionally, you could also add `So ntp[0] is always zero as this corresponds to assigning all data class false and ntp[len(classes)] is always len(classes) as this corresponds to assigning all data class true.`. It's the `(len(classes) - i)` part that I think is confusing, which is why I would prefer the more specific explanation.

stat/roc.go
```diff
@@ -127,3 +129,68 @@ func ROC(cutoffs, y []float64, classes []bool, weights []float64) (tpr, fpr, thr
...
+// where values above the given rank are assigned class true for each given
+// rank from 1 to len(classes). The values of min and max provide the minimum
+// and maximum possible number of true values for the set of classes.
+//
+// If weights is nil, all weights are treated as 1. If weights is not nil
+// it must have the same length as classes, otherwise TOC will panic.
```
Armadilloa16, line 147: I would add here `When weights are not nil, the calculation of min and max allows for partial assignment of single data points`. To explain in more detail: for example if the total weight of the true values is 8, and the first three values have weights [3 3 3] the third (well fourth) max value will be 8, i.e. min(8, 3 + 3 + 3), even though there is no way to actually make 8 from any combination of [3 3 3]. I agree this is the correct implementation to use, primarily because computing the actual max in the weighted case is combinatorially hard, but I think we should note that we are using this simplification.

stat/roc_example_test.go
```diff
@@ -125,3 +125,126 @@ func ExampleROC_aUC() {
...
+	stat.SortWeightedLabeled(y, classes, weights)
+
+	min, ntp, max := stat.TOC(classes, weights)
+	fmt.Printf("minimum bound: %v\n", min)
+	fmt.Printf("TOC:           %v\n", ntp)
+	fmt.Printf("maximum bound: %v\n", max)
+
+	// Output:
+	// minimum bound: [0 0 0 3 6 8 10]
+	// TOC:           [0 4 4 10 10 10 10]
```
Armadilloa16, line 179: So I'm still completely confuddled by the parity change you made, and the correspondence (or seeming contradiction) to the docstrings, etc. My understanding was you wanted consistency with sklearn i.e. that values above or equal to the cutoff/ threshold/ value of the ranking variable are assigned class true. If that is the case, the TOC for this example should be [0 2 4 7 7 8 8].

Armadilloa16 (changes requested):
I'm happy to hand this over to you since you seem to understand what you want better than I do at this point, but I've highlighted the things that don't make sense to me in case that's still useful. Sorry for disappearing, I had surgery 12 days ago and I've been all over the place on pain meds since so excuse me please if some of my comments were nonsense.

stat/roc.go
kortschak: Done

stat/roc_example_test.go
kortschak: The issue comes down to; 1. the change asked for by #316 was to make ROC here behave the same way that sklearn's ROC behaves, and 2. that the AUC for a ROC curve will be the same as the AUC for a TOC parallelogram. If 1. was the correct thing to do then TOC has to behave this way AFAICS. Maybe it wasn't, but if that's the case, then that should be discussed and fixed.

kortschak:
I'm going to leave this for a while. If someone wants to pick it up, please feel free.

stat/roc_example_test.go
Armadilloa16: Yeah, maybe it's your point 2. that I don't understand then, because I still maintain the trapezoidal rule is an incorrect way to calculate the AUC for both ROC and a TOC because it incorrectly interpolates between points linearly, when this interpolation should actually be done in a stepwise manner (going left to right on the plot, horizontal and then vertical in the case of a ROC, horizontal and then parallel to the max bound in the case of a TOC). The trapezoidal rule calculates neither of these correctly in general, although in the case of TOC, so long as all cutoffs are calculated (as per the current implementation), it will actually happen to calculate the correct thing. In the case of ROC it will not, even in the case when all cutoffs are calculated, although in the case that a smaller number of cutoffs are used instead it might make more sense to use the trapezoidal rule approximation, and maybe that was the reasoning for doing it that way? If we don't allow that, as in this implementation of TOC, then maybe we should also go back to calculating the AUC correctly, but I can see why that should be dealt with in a seperate issue.

kortschak: Why do you think that it won't calculate correctly when the cutoffs are calculated? In that case, all (x,y) pairs are ends of straight lines along an ROC/TOC curve. If it's wrong, it's another issue though, as you say.

Armadilloa16: Oh wait, you're completely correct and I'm not. This is actually completely fine. I was thinking about a completely different implementation of ROC calculations in which only vertical (TPR) changes are tracked, so the output is condensed... we don't do that here at all, so it's all good. Sorry for the confusion.

kortschak: :tada:

So the only remaining things is the API redesign to consider.

Armadilloa16: Yep.

Armadilloa16 (approved):
Aside from the UI redesign this LGTM, and the UI redesign should probably go into another issue so this seems good for now.

kortschak:
v0.9.0 has been released, so this can be merged when it is approved.

kortschak:
*ping*

stat/roc_example_test.go
romanwerpachowski: I would still spell it out for the benefit of simpletons like yours truly :) At least that the rank variable starts from 0 not from 1, since the convention in statistics seems to be to start from 1: https://en.wikipedia.org/wiki/Ranking#Ranking_in_statistics

romanwerpachowski (changes requested):

stat/roc.go
```diff
@@ -127,3 +129,70 @@ func ROC(cutoffs, y []float64, classes []bool, weights []float64) (tpr, fpr, thr
...
 	return tpr, fpr, cutoffs
 }
+
+// TOC returns the Total Operating Characteristic for the classes provided
+// and the minimum and maximum bounds for the TOC.
+//
+// The input y values that correspond to classes and weights must be sorted
+// in ascending order. classes[i] is the class of value y[i] and weights[i]
+// is the weight of y[i]. SortWeightedLabeled can be used to sort classes
+// together with weights by the rank variable.
```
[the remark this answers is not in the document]
kortschak: Yes, rank is i+1.

stat/roc_example_test.go
kortschak: The zero'th element is not a ranked value.

stat/roc.go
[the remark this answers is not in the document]
romanwerpachowski: Agreed.

stat/roc_example_test.go
romanwerpachowski: What do you mean?

stat/roc.go
kortschak: It is working. I initially had the approach wrong.

stat/roc.go
```diff
@@ -127,3 +129,70 @@ func ROC(cutoffs, y []float64, classes []bool, weights []float64) (tpr, fpr, thr
...
+		return nil, nil, nil
+	}
+
+	ntp = make([]float64, len(classes)+1)
+	min = make([]float64, len(ntp))
+	max = make([]float64, len(ntp))
+	if weights == nil {
+		for i := range ntp[1:] {
+			ntp[i+1] = ntp[i]
+			if !classes[i] {
```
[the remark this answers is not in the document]
kortschak: Can you elaborate?

stat/roc.go
kortschak: I'll see what I can come up with.

stat/roc_example_test.go
kortschak: In the output slices, the zero'th element is not actually a ranked part, it's the origin. The ranked parts start at 1.

stat/roc.go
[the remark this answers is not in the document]
romanwerpachowski: Number of true positives is the number of variables with class == True and classified as True, right? But if all classes == False, then ntp[i] == 0 by definition. But the code will be incrementing ntp[i + 1] in every step.

What am I not getting here?

stat/roc_example_test.go
romanwerpachowski: Ah, OK.

stat/roc.go
kortschak: `ntp[i] = sum_{j = 0}^{i} !classes[j]*weight[j]`

stat/roc.go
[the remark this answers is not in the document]
kortschak: Yeah, this part does my head in, but it's a consequence of the decision to invert the parity of the ROC input. If we accept that the decision taken in #316 was correct, then this follows.

kortschak:
PTAL

stat/roc.go
romanwerpachowski: I'm going to be obtuse and ask: how many developers out there will recognise the Iverson bracket? We should at least add

```
, where [x] = 1 if x else 0.
```

stat/roc.go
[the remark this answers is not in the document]
romanwerpachowski: But that's not really an answer, is it? What if the decision taken in #316 was wrong? Or maybe the consequence doesn't follow? Actually, looking at the discussion in #316 and the changes in #959, it looks to me that the change of parity you're talking about had to do with changing the interpretation of the `y` input to the `ROC` function, not of the `classes` input.

Maybe it all checks out and I'm wasting your time, but I would really feel safer approving this PR if I could see that the code works OK without referring to other PRs and other functions, just from the standard definition of TOC. As it is, even the updated docstring is clearly inconsistent with the code, unless we decided to redefine the meaning of "true positive"...

stat/roc.go
kortschak: Valid. Done.

romanwerpachowski: Thanks.

stat/roc.go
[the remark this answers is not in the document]
kortschak: Yes, I'm extremely uncomfortable with this (the part you edited out is absolutely correct).

The published definition of TOC (unfortunately an excel spread sheet is the implementation associated with the paper) is the inverse of the class parity here. However, it's important that the TOC AUC and ROC AUC agree since they are isomorphic. Under this transformation (flipping reality) the implementation here agrees with the paper.

The changes in #959 dealt with a few things, including the `classes` input. From memory much of that was driven by the stated desire to match the SciPy behaviour.

> As it is, even the updated docstring is clearly inconsistent with the code, unless we decided to redefine the meaning of "true positive"...

Yes. This makes me unhappy.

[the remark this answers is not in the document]
romanwerpachowski: So the discussion in #316 includes this statement by @btracey:

> ROC treats values below the cutoff as true (rather than false)

But this is different than what Wikipedia says: https://en.wikipedia.org/wiki/Receiver_operating_characteristic#Curves_in_ROC_space

> Given a threshold parameter T, the instance is classified as "positive" if X>T, and "negative" otherwise.

If we associate "positive" with "true" and "negative" with "false".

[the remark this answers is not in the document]
romanwerpachowski: And the documentation for sklearn says about the 2nd argument of `roc_auc_score`: https://scikit-learn.org/stable/modules/generated/sklearn.metrics.roc_auc_score.html#sklearn.metrics.roc_auc_score

> In the binary case, it corresponds to an array of shape (n_samples,). Both probability estimates and non-thresholded decision values can be provided. The probability estimates correspond to the probability of the class with the greater label, i.e. estimator.classes_[1] and thus estimator.predict_proba(X, y)[:, 1]. The decision values corresponds to the output of estimator.decision_function(X, y). See more information in the User guide;

The user guide https://scikit-learn.org/stable/modules/model_evaluation.html#roc-metrics makes it clear that they treat the values above the threshold as True.

Looking at https://github.com/gonum/gonum/blob/master/stat/roc_example_test.go#L111, the current AUC function in gonum matches sklearn, hence it treats - like sklearn and Wikipedia - the values ABOVE the threshold as True.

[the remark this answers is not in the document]
kortschak: Yeah. the issue is the behaviour of SciKit learn (where [1 is positive](https://scikit-learn.org/stable/modules/generated/sklearn.metrics.roc_curve.html)). We see it do this

```
>>> from sklearn import metrics
>>> y = [0.1, 0.35, 0.4, 0.8]
>>> classes = [1, 0, 1, 0]
>>> metrics.roc_auc_score(classes, y)
0.25
```
and the (probably) [equivalent Gonum](https://github.com/gonum/gonum/blob/eae7c2a69e552715bd866b03ccd39d0c90504af6/stat/roc_example_test.go#L111-L127) code do this

```
y := []float64{0.1, 0.35, 0.4, 0.8}

tpr, fpr, _ := stat.ROC(nil, y, classes, nil)

// Compute Area Under Curve.
auc := integrate.Trapezoidal(fpr, tpr)
fmt.Printf("auc: %v\n", auc)

// Output:
// true  positive rate: [0 0 0.5 0.5 1]
// false positive rate: [0 0.5 0.5 1 1]
// auc: 0.25
```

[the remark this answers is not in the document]
romanwerpachowski: In summary, I think the TOC function should:

1. treat values ABOVE the threshold as True
2. define `ntp[i]` as

```
ntp_i = sum_{j = i + 1}^{len(ntp) - 1} [ classes_j ] * weights_j, where [x] = 1 if x else 0.
ntp_{len(ntp) - 1} = 0
```

And then I think we will match AUC, sklearn and remain sane.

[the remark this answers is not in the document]
romanwerpachowski: > Yeah. the issue is the behaviour of SciKit learn (where [1 is positive](https://scikit-learn.org/stable/modules/generated/sklearn.metrics.roc_curve.html)). We see it do this
> 
> ```
> >>> from sklearn import metrics
> >>> y = [0.1, 0.35, 0.4, 0.8]
> >>> classes = [1, 0, 1, 0]
> >>> metrics.roc_auc_score(classes, y)
> 0.25
> ```
> 
> and the (probably) [equivalent Gonum](https://github.com/gonum/gonum/blob/eae7c2a69e552715bd866b03ccd39d0c90504af6/stat/roc_example_test.go#L111-L127) code do this
> 
> ```
> y := []float64{0.1, 0.35, 0.4, 0.8}
> classes := []bool{true, false, true, false}
> 
> tpr, fpr, _ := stat.ROC(nil, y, classes, nil)
> 
> // Compute Area Under Curve.
> auc := integrate.Trapezoidal(fpr, tpr)
> fmt.Printf("true  positive rate: %v\n", tpr)
> fmt.Printf("false positive rate: %v\n", fpr)
> fmt.Printf("auc: %v\n", auc)
> 
> // Output:
> // true  positive rate: [0 0 0.5 0.5 1]
> // false positive rate: [0 0.5 0.5 1 1]
> // auc: 0.25
> ```

We would probably gain total clarity if we also displayed the thresholds in the Gonum example.

https://scikit-learn.org/stable/modules/generated/sklearn.metrics.roc_curve.html generates the following complete output:

```
>>> import numpy as np
>>> from sklearn import metrics
>>> y = np.array([1, 1, 2, 2])
>>> scores = np.array([0.1, 0.4, 0.35, 0.8])
>>> fpr, tpr, thresholds = metrics.roc_curve(y, scores, pos_label=2)
>>> fpr
array([0. , 0. , 0.5, 0.5, 1. ])
>>> tpr
array([0. , 0.5, 0.5, 1. , 1. ])
>>> thresholds
array([1.8 , 0.8 , 0.4 , 0.35, 0.1 ])
```

Which clearly shows that the TPR goes up as the threshold goes down.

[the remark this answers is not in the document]
kortschak:
```
y := []float64{0.1, 0.4, 0.35, 0.8}
classes := []bool{false, false, true, true}
stat.SortWeightedLabeled(y, classes, nil)

tpr, fpr, thresh := stat.ROC(nil, y, classes, nil)
fmt.Printf("threshold          : %v\n", thresh)
```

```
true  positive rate: [0 0.5 0.5 1 1]
false positive rate: [0 0 0.5 0.5 1]
threshold          : [+Inf 0.8 0.4 0.35 0.1]
```

kortschak:
That was way harder than it should have been.

stat/roc.go
```diff
@@ -141,7 +139,7 @@ func ROC(cutoffs, y []float64, classes []bool, weights []float64) (tpr, fpr, thr
 // The returned ntp values can be interpreted as the number of true positives
 // where values above the given rank are assigned class true for each given
 // rank from 1 to len(classes).
```
romanwerpachowski, line 141: Is it true that `ntp[i]` corresponds to rank `len(classes) - i`? If yes, then maybe we should mention that for clarity.

romanwerpachowski (dismissed):
Thanks!

stat/roc.go
kortschak: No, it corresponds to the sum of positive weights.

kortschak: This is addressed in the paragraph after the formula though (which had a residual incorrectness — now fixed).

kortschak:
Sorry, I fixed an error in the docs, so this needs reapproval.

romanwerpachowski (approved):

kortschak:
Thanks.

So the lesson here, kids, is that published methods *should not* have their reference implementation be an Excel spread sheet and *should provide* a formal definition of the approach. For third-party implementers, if *there is no* formal definition and the reference implementation *is* a spread sheet, don't assume conventional data layout and ordering.

PR #1571, opened 2021-02-23
stat: bring ROC doc comment into line with logic inversion

kortschak:
This was missed in 50ee437.

Please take a look.


romanwerpachowski (approved):
Thanks.

PR #1574, opened 2021-03-01
floats: remove deprecated functions

kortschak:
Please take a look.


sbinet (dismissed):

kortschak:
Had to do a meaningless force push because there is no way to unstick or retrigger builds and GHA was broken when this was originally pushed.

sbinet (approved):

PR #1575, opened 2021-03-04
lapack/gonum: rewrite Dlartg and its test to use safe scaling

vladimir-ch:
This PR are two changes that could be squashed before merging but also retained for historical/educational/motivational purposes.

The first change fixes infinite loop in Dlartg if the input contains NaN or Inf. This was recently fixed in Reference LAPACK https://github.com/Reference-LAPACK/lapack/pull/419

The second change is a rewrite of Dlartg that uses safe scaling and without loops based on [this paper](https://doi.org/10.1145/3061665) and its supplemental code:
```
Edward Anderson. 2017. Algorithm 978: Safe Scaling in the Level 1 BLAS.
ACM Trans. Math. Softw. 44, 1, Article 12 (July 2017), 28 pages.
```
The same change is being reviewed in the reference here: https://github.com/Reference-LAPACK/lapack/pull/493

Please take a look.


kortschak:
This has some flow on effects in mat, stat and graph/layout. The test updates in layout need to be run for each of amd64, 386 and arm64. I'll email you a patch file.

lapack/gonum/dlartg.go
```diff
@@ -9,72 +9,50 @@ import "math"
...
-		sn = g1 / r
-	}
-	if math.Abs(f) > math.Abs(g) && cs < 0 {
-		cs *= -1
-		sn *= -1
-		r *= -1
+		sn = math.Copysign(1, g)
+		r = g1
+	case drtmin < f1 && f1 < drtmax && drtmin < g1 && g1 < drtmax:
+		d := math.Sqrt(f*f + g*g)
```
kortschak, line 40: `math.Hypot(f, g)`?

lapack/gonum/dlartg.go
```diff
@@ -9,72 +9,50 @@ import "math"
...
+		cs = f1 * p
+		sn = g * math.Copysign(p, f)
+		r = math.Copysign(d, f)
+	default:
+		maxfg := math.Max(f1, g1)
+		u := math.Min(math.Max(safmin, maxfg), safmax)
+		uu := 1 / u
+		fs := f * uu
+		gs := g * uu
+		d := math.Sqrt(fs*fs + gs*gs)
```
kortschak, line 51: `math.Hypot(fs, gs)`?

lapack/gonum/dlartg.go
vladimir-ch: It would be cleaner but `math.Hypot` does extra work to protect against things that won't happen here, so not calling it can be faster. I guess I'll have to support this with some benchmarks.

kortschak: If the overflow/underflow is not going to happen, then there's no point in the protection. Benchmarks aren't needed.

kortschak:
Found some more failures on arm64.

```
--- FAIL: TestDsterf (0.03s)
    dsterf.go:193: n=50,type=8: unexpected result; |dGot-dWant|=1.021405182655144e-14
FAIL
FAIL	gonum.org/v1/gonum/lapack/gonum	1085.949s
--- FAIL: TestTorgersonScaling (0.00s)
    mds_test.go:136: low precision (tol=1e-09) k for test 0: gotK:5 derived=want:4
    mds_test.go:144: unexpected Eigenvalues for test 0:
        got: [1.172027697614e+07 5.686036184502e+06 2.474981412369e+04 1.238300279584e+03 4.875142139613e-10 -2.471028925699e+02 -2.412961483427e+03 -8.452858567111e+04]
        want:[1.172027697614e+07 5.686036184502e+06 2.474981412369e+04 1.238300279584e+03 -5.444455825182e-10 -2.471028925712e+02 -2.412961483429e+03 -8.452858567111e+04]
FAIL
```

Running `noasm` now and will post if any additional failures are found there.

vladimir-ch:
I've pushed a commit that relaxes the Dsterf test.

kortschak:
The following fixes the Torgerson scaling failure (and is clearer):
```
diff --git a/stat/mds/mds_test.go b/stat/mds/mds_test.go
index d1be33fd..b9e87f6f 100644
--- a/stat/mds/mds_test.go
+++ b/stat/mds/mds_test.go
@@ -45,7 +45,7 @@ var torgersonScalingTests = []struct {
                }),
                wantVals: []float64{
                        1.172027697614e+07, 5.686036184502e+06, 2.474981412369e+04, 1.238300279584e+03,
-                       -5.444455825182e-10, -2.471028925712e+02, -2.412961483429e+03, -8.452858567111e+04,
+                       0, -2.471028925712e+02, -2.412961483429e+03, -8.452858567111e+04,
                },
                tol: 1e-9,
```

vladimir-ch:
Thanks, I've just been looking at TorgersonScaling.
I've now included your suggestion.

What about the log output for the `k` value? It seems that the expectation is correct because the newly positive eigenvalue is practically zero. Since there's no failure, we can leave it for later.

kortschak:
I think the real value is for the changed element is 0. R's cmdscale gives a value that's negative on the order of 1e-10 (similar to the previous expected value at the crossover). This could be fixed by providing a tolerance to Torgerson, or not returning k, or noting that ED is unstable and results near zero should be examined.

vladimir-ch:
> I think the real value is for the changed element is 0.

Agreed.

> ... fixed by providing a tolerance to Torgerson, or not returning k, or noting that ED is unstable and results near zero should be examined.

I'd say the last one.

I've rebased on top of master and regenerated images for graph/layout. Test pass.

kortschak:
I'm running the tests on arm64 now.

kortschak:
> I'd say the last one.

#1588

kortschak:
Tests fail for arm64 and 386. I'll mail a patch.

vladimir-ch:
I've applied the patch.

kortschak (approved):

Issue #1582, opened 2021-03-13
graph/simple: Feature Request: Complete, cycle, wheel, and star-making functions


mattConn:
## Feature
A useful addition to the graph/simple package could be some classic graph-making functions. The graph-making functions I have in mind would initialize and return "complete", "cycle", "wheel" and "star" graphs.

Having these functions in the package would offer a built-in convenience. These graphs are also useful as building blocks for other graphs. Lastly, for a new user of Gonum and/or student of Graph Theory, these functions could make for a good starting point.

## Implementation
Each of these functions would consist of setting edges via loop and return a concrete, mutable graph. For example, a complete graph-making function might look like:

```
func MakeUndirectedCompleteGraph(start, length int) *simple.UndirectedGraph {
	g := simple.NewUndirectedGraph()

	for i := start; i < start+length; i++ {
		for j := i + 1; j < start+length; j++ {
			g.SetEdge(g.NewEdge(simple.Node(i), simple.Node(j)))
		}
	}

	return g
}
```
This would yield a graph with nodes `[start, start+length)`, so that the programmer can make and join similar graphs without collision. The code would be similar for cycle, wheel, and star functions, each having identical type signatures.

I would like to contribute to this feature should it be deemed useful.

kortschak:
I would prefer if this were closer to the API in graph/graphs/gen and did not return a concrete type but rather took a destination graph.Builder. The example above would look like this:
```
func Complete(dst graph.Builder, n int) {
	_, isDirected := dst.(graph.Directed)
	for i := 0; i < n; i++ {
		for j := i + 1; j < n; j++ {
			u := g.NewNode()
			g.AddNode(u)
			v := g.NewNode()
			g.AddNode(v)
			g.SetEdge(g.NewEdge(v, v))
			if isDirected {
				g.SetEdge(g.NewEdge(v, v))
			}
		}
	}
}
```
The documentation needs to say that the graph being constructed will be a subgraph of the destination if there are already nodes/edges present.

This decouples the graph construction from graph/simple. The signatures for the different functions would look different, for example, star and wheel may take a parameter specifying the central node (or return the id of the central node if there is no parameter).

I would also suggest that "path(n)", "tree(d,b)" and "randomtree(n)" (in some flavour) may be useful to add.

mattConn:
> The documentation needs to say that the graph being constructed will be a subgraph of the destination if there are already nodes/edges present.

That sounds good.

[...]

PR #1584, opened 2021-03-15
graph/{multi,simple,testgraph}: add support for ID-specified node addition

kortschak:
I'm not sure about the placement of the `NodeWithIDer` interface. Given that we need to have it in testgraph, I'm thinking that maybe it *should* just go in graph.

Please take a look.

Updates #1582.

/cc @mattConn

### DIFF
--- a/graph/graph.go
+++ b/graph/graph.go
@@ -146,6 +146,18 @@ type NodeAdder interface {
 	AddNode(Node)
 }
 
+// NodeWithIDer is a graph that can return potentially new nodes with
+// a defined ID.
+type NodeWithIDer interface {
+	// NodeWithID returns a Node with the given ID if possible.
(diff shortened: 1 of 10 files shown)


mewmew (approved):
LGTM.

kortschak:
What do you think about placing interface in graph? As it is, it needs to be defined in at least one more place (the use package) since referring to testgraph feels very wrong.

mewmew:
> What do you think about placing interface in graph? As it is, it needs to be defined in at least one more place (the use package) since referring to testgraph feels very wrong.

I'm fine with placing the interface in the graph package.

kortschak:
PTAL

mewmew (approved):
LGTM.

PR #1587, opened 2021-03-16
graph/flow: improve consistency of coverage reports

kortschak:
Please take a look.

The background to this is that we regularly see graph/flow/control_flow\*.go files, and occasionally optimize/listsearch.go, in the codecov reports as having changed. This is due to algorithmic randomness that shows up in those files. In the case of graph/flow it is due to map iteration order. So the approach is just to increase the number of times the tests are run to decrease the probability of having a test ensemble that does not hit possible lines at least once.

This PR does not address the effect that we see in optimize/listsearch.go, but I'm unable to see how that comes in. I suspect it's from concurrency racing in the algorithm, but it's low enough frequency that I don't see it when I repeatedly run coverage tests. I'll look again if it continues to be a problem.

### DIFF
--- a/graph/flow/control_flow_test.go
+++ b/graph/flow/control_flow_test.go
@@ -132,53 +132,60 @@ func (n char) ID() int64      { return int64(n) }
 func (n char) String() string { return string(n) }
 
 func TestDominators(t *testing.T) {
-	for _, test := range dominatorsTests {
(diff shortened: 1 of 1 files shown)


kortschak:
*ping*

vladimir-ch (approved):

PR #1589, opened 2021-03-17
 internal/asm/f32: add assembly Sum implementation

y0ast:
Tests pass, implementation based on https://github.com/gonum/gonum/blob/master/internal/asm/f64/sum_amd64.s (including taking over the respective tests).

I know `f32` isn't really exposed, but I hope that in the future maybe it will be. At the moment I simply copy it over ("vendor") and noticed this functionality was missing.

### DIFF
--- a/internal/asm/f32/stubs_test.go
+++ b/internal/asm/f32/stubs_test.go
@@ -150,3 +150,63 @@ func TestAxpyIncTo(t *testing.T) {
 		checkValidIncGuard(t, v.dst, 0, int(v.incDst), gdLn)
 	}
 }
+
+func TestSum(t *testing.T) {
+	var srcGd float32 = -1
+	for j, v := range []struct {
+		src    []float32
+		expect float32
+	}{
+		{
+			src:    []float32{},
+			expect: 0,
+		},
+		{
+			src:    []float32{1},
+			expect: 1,
+		},
+		{
+			src:    []float32{nan},
+			expect: nan,
+		},
+		{
+			src:    []float32{1, 2, 3},
+			expect: 6,
+		},
+		{
+			src:    []float32{1, -4, 3},
+			expect: 0,
+		},
+		{
+			src:    []float32{1, 2, 3, 4},
+			expect: 10,
+		},
+		{
+			src:    []float32{1, 1, nan, 1, 1},
+			expect: nan,
+		},
+		{
+			src:    []float32{inf, 4, nan, -inf, 9},
+			expect: nan,
+		},
+		{
+			src:    []float32{1, 1, 1, 1, 9, 1, 1, 1, 2, 1, 1, 1, 1, 1, 5, 1},
+			expect: 29,
+		},
+		{
+			src:    []float32{1, 1, 1, 1, 9, 1, 1, 1, 2, 1, 1, 1, 1, 1, 5, 11, 1, 1, 1, 9, 1, 1, 1, 2, 1, 1, 1, 1, 1, 5, 1},
+			expect: 67,
+		},
+	} {
+		gdLn := 4 + j%2
+		gsrc := guardVector(v.src, srcGd, gdLn)
+		src := gsrc[gdLn : len(gsrc)-gdLn]
+		ret := Sum(src)
+		if !same(ret, v.expect) {
+			t.Errorf("Test %d Sum error Got: %v Expected: %v", j, ret, v.expect)
+		}
+		if !isValidGuard(gsrc, srcGd, gdLn) {
+			t.Errorf("Test %d Guard violated in src vector %v %v", j, gsrc[:gdLn], gsrc[len(gsrc)-gdLn:])
+		}
+	}
+}
--- a/internal/asm/f32/stubs_test.go
+++ b/internal/asm/f32/stubs_test.go
@@ -198,15 +198,17 @@ func TestSum(t *testing.T) {
 			expect: 67,
 		},
 	} {
-		gdLn := 4 + j%2
-		gsrc := guardVector(v.src, srcGd, gdLn)
-		src := gsrc[gdLn : len(gsrc)-gdLn]
-		ret := Sum(src)
-		if !same(ret, v.expect) {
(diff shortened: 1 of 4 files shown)


y0ast:
Thanks for the feedback, I've addressed it.

PTAL :)

internal/asm/f32/sum_amd64.s
```diff
@@ -0,0 +1,100 @@
...
+	JZ   no_trim     // if TAIL == 0 { goto no_trim }
+	SUBQ $16, TAIL   // TAIL -= 16
+
+sum_align: // Align on 16-byte boundary do {
+	ADDSS (X_PTR)(IDX*4), SUM // SUM += x[0]
+	INCQ  IDX                 // i++
+	DECQ  LEN                 // LEN--
+	JZ    sum_end             // if LEN == 0 { return }
+	ADDQ  $4, TAIL            // TAIL += 4
+	JNZ   sum_align           //  } while TAIL < 0
```
kortschak, line 42: Sorry. `s/  / /`.

y0ast: Haha, no problem, done!

kortschak (approved):
LGTM but I'd like to wait for a second reviewer due to asm.

internal/asm/f32/sum_amd64.s
```diff
@@ -0,0 +1,100 @@
...
+	JE   sum_end
+
+	PXOR SUM_1, SUM_1
+	PXOR SUM_2, SUM_2
+	PXOR SUM_3, SUM_3
+
+	MOVQ X_PTR, TAIL // Check memory alignment
+	ANDQ $15, TAIL   // TAIL = &x % 16
+	JZ   no_trim     // if TAIL == 0 { goto no_trim }
+	SUBQ $16, TAIL   // TAIL -= 16
```
Kunde21, line 34: A right shift here keeps the counter positive, making the alignment loop more easily readable at the same operation cost.

internal/asm/f32/sum_amd64.s
```diff
@@ -0,0 +1,100 @@
...
+	HADDPS SUM, SUM           // sum_i[:2] += sum_i[2:4]
+	HADDPS SUM, SUM           // sum_i[0] += sum_i[1]
+
+	TESTQ $2, TAIL
+	JZ    sum_tail1
+
+	ADDSS (X_PTR)(IDX*4), SUM // sum_i += x[i:i+1]
+	ADDQ  $1, IDX
+	ADDSS (X_PTR)(IDX*4), SUM // sum_i += x[i:i+1]
+	ADDQ  $1, IDX
```
Kunde21, line 90: This is an `ADDPS` operation.  You save a few operations and avoid re-indexing with `MOVSD` and `ADDPS`, then moving one of the `HADDPS` operations to `sum_tail1` block.

internal/asm/f32/stubs_test.go
```diff
@@ -150,3 +150,63 @@ func TestAxpyIncTo(t *testing.T) {
...
+		{
+			src:    []float32{1, 1, 1, 1, 9, 1, 1, 1, 2, 1, 1, 1, 1, 1, 5, 1},
+			expect: 29,
+		},
+		{
+			src:    []float32{1, 1, 1, 1, 9, 1, 1, 1, 2, 1, 1, 1, 1, 1, 5, 11, 1, 1, 1, 9, 1, 1, 1, 2, 1, 1, 1, 1, 1, 5, 1},
+			expect: 67,
+		},
+	} {
+		gdLn := 4 + j%2
```
Kunde21, line 201: Note here:  `f64` can only be aligned on or one-off of the 16-addr mark.  `f32` needs to handle more states, as there are three different mis-alignment possibilities.

internal/asm/f32/sum_amd64.s
y0ast: https://www.felixcloutier.com/x86/addps

It seems that ADDPS only does 4, 8 or 16 packed float32s or did I miss something?

Kunde21: You're not going to use the result in the rest of that register, so it's still adding 4.  
- `MOVSD` loads 2 packed f32 into the bottom half of the register
- `ADDPS` adds all 4 f32 slots
- `HADPS` combines the lower 2 f32 (the ones we care about) into slot 0
- `sum_tail1` ignores the upper 3 f32 slots

internal/asm/f32/sum_amd64.s
y0ast: I've looked into doing this, and I am not sure how to accomplish this with a right shift:

if TAIL = 4, then we should be doing 3 loops of `sum_align`.
At the moment we do TAIL -= 16, then add 4 to TAIL 3 times to reach zero.

If we shift 4 right we can obtain 2 or 1, which wouldn't help (we want to count down to zero). I've played around a bit with other flavours, but it seems that it always adds more operations.

internal/asm/f32/sum_amd64.s
y0ast: Ah, right, that makes sense. Thanks for the explanation.

The `MOVSD` would that go to a new register? Would I need to zero that register before use?

The following passes the test.
```
 sum_tail2:
      HADDPS SUM, SUM           // sum_i[:2] += sum_i[2:4]

      TESTQ $2, TAIL
      JZ    sum_tail1

      MOVSD (X_PTR)(IDX*4), X4
      ADDPS X4, SUM             // sum_i += x[i:i+2]
      ADDQ  $2, IDX

  sum_tail1:
      HADDPS SUM, SUM // sum_i[0] += sum_i[1]
      ...
```

Kunde21: I'd re-use `SUM_1` here since there's no chance of soft dependency on the upper 128-bits and the useful data was captured in ln72

y0ast:
Aside from the right shift comment, this addresses the rest. Also rebased.

Kunde21 (approved):

PR #1591, opened 2021-03-19
 internal/asm/f64: fix bug and clean up variable names

y0ast:
The code was jumping to end when tail == 0, but the correct thing is to jump when `LEN` is zero. `TAIL == 0` means we can start the summing pipeline. We also don't need to decrement the TAIL since we overwrite it at start of `no_trim`

This was not caught by the tests (they pass before and now). I am curious if anyone has an idea for a test case that would catch this.

I've also made the register names consistent.

### DIFF
--- a/internal/asm/f64/stubs_test.go
+++ b/internal/asm/f64/stubs_test.go
@@ -635,5 +635,16 @@ func TestSum(t *testing.T) {
 		if !isValidGuard(gsrc, srcGd, gdLn) {
 			t.Errorf("Test %d Guard violated in src vector %v %v", j, gsrc[:gdLn], gsrc[len(gsrc)-gdLn:])
 		}
+
+		gdLn++
+		gsrc = guardVector(v.src, srcGd, gdLn)
+		src = gsrc[gdLn : len(gsrc)-gdLn]
+		ret = Sum(src)
+		if !scalar.Same(ret, v.expect) {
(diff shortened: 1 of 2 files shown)


Kunde21:
Testing approach for this would mean forcing mis-alignment.  My approach here was in `stubs_test.go` line 628
```
		gdLn := 4 + j%2
```
Ideally, we do each test twice.  Once with `gdLn` and again with `gdLn+1`, to force that un-aligned memory.  For f64, at least, that would cover that early alignment path.

y0ast:
I've added that to the test.

However I just realized that the previous situation wasn't causing wrong output:

```
	ANDQ $15, TAIL   // TAIL = &y % 16
```
TAIL=8 if unaligned, otherwise 0

In old situation:
```
	DECQ TAIL // TAIL--
	JZ sum_end // if TAIL == 0 { return }
```
Decrementing tail by one, makes it `tail = 7`, so we never hit the spurious jump zero. We just incur a performance hit for single floats that are misaligned, and it's quite unusual to call this function on an array of length 1 (`x[0]` would be faster).

Kunde21 (approved):

kortschak:
@y0ast Please send another PR adding yourself to AUTHORS and CONTRIBUTORS with the commit message "A+C: add Joost van Amersfoort". When that's done this (and that) can be merged.

kortschak:
@y0ast Thanks for the contribution and welcome.

Issue #1457, opened 2020-09-16
dsp/window: fix spectral leakage parameter documentation


kortschak:
In the discussion around windowing end points, it was raised that the spectral leakage parameters we report were wrong for the functions that we provide. This got me interested to make sure that we are reporting the correct values.

So I wrote a small utility to leave in the dsp/window directory for use by authors of new window functions to check the spectral parameters before they're added. In doing this I checked all the parameters that we have from the original additions. They match-ish, making me feel a little uncomfortable. To address this I've extended the utility to be able to consume a spectrum profile from another source (e.g. NumPy) to confirm that the results match between implementations. NumPy only provides two windows (Blackman and Hamming), so those were checked.

The spectrum and spectral leakage parameter from Blackman match perfectly between the two implementations.

First the Gonum implementation,

![blackman_20_1000_go](https://user-images.githubusercontent.com/275221/93291067-50b6fc00-f821-11ea-8afd-eb2197a9a2bf.png)

and then the NumPy.

![blackman_20_1000_py](https://user-images.githubusercontent.com/275221/93291080-56acdd00-f821-11ea-8245-96c80ff0968e.png)

Note that the program does not calculate the beta value from the spectrum data from the python code, but the python code calculates the same value as the Go utility for the Gonum functions.

The match between Hamming implementations is close but not exact.

![hamming_20_1000_go](https://user-images.githubusercontent.com/275221/93291210-9ffd2c80-f821-11ea-9aad-8bb639a385f0.png)

![hamming_20_1000_py](https://user-images.githubusercontent.com/275221/93291227-a5f30d80-f821-11ea-89a0-dfad35203fbc.png).

The Hamming window is [based on a constant defined (variably)](https://ccrma.stanford.edu/~jos/sasp/Hamming_Window.html) as either α=25/46 or the decimal approximations α=0.54. NumPy uses the decimal approximation while we use a machine-level precision representation of the fraction.

If I replace the constant we use with the approximate value, again we match perfectly.

![hamming_20_1000_go_approx](https://user-images.githubusercontent.com/275221/93291718-fcad1700-f822-11ea-8547-24c4e51db420.png)

On the basis of this, I'm pretty confident that the utility works correctly and that at least the pair of windows that can be compared against other implementations are correct (depending on the correctness of using the higher precision constant\* - perhaps we should use the 2dp approximations to match others).

[...]

PR #1597, opened 2021-03-24
dsp/window/cmd/leakage: add utility command for parameter estimation

kortschak:
Please take a look.

Updates #1457.

### DIFF
--- a/dsp/window/cmd/leakage/leakage.go
+++ b/dsp/window/cmd/leakage/leakage.go
@@ -293,14 +293,16 @@ func funcCharacteristics(fn func([]float64) []float64, n, m int, symm bool) (c *
 	for i, c := range fft.Coefficients(nil, t) {
 		a := db(cmplx.Abs(c))
 		t[i] = a
-
-		if math.IsInf(a, -1) {
-			a = min
-		} else if a < min {
+		if !math.IsInf(a, -1) && a < min {
 			min = a
 		}
 		if i == 0 {
 			max = a
+		}
+	}
+	for i, a := range t[:m/2+1] {
+		if math.IsInf(a, -1) {
+			a = min
 		}
 		xy[i] = plotter.XY{X: float64(i) * float64(n) / float64(m), Y: a - max}
 	}
--- a/dsp/window/cmd/leakage/leakage.go
+++ b/dsp/window/cmd/leakage/leakage.go
@@ -1,0 +1,387 @@
+// Copyright ©2021 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// The leakage program provides summary characteristics and a plot
+// of spectral response for window functions or csv input. It is intended
+// to be used to verify window behaviour against foreign implementations.
+// For example, the behavior of a NumPy window can be captured using this
+// python code:
+//  import matplotlib.pyplot as plt
+//  import numpy as np
+//  from numpy.fft import fft
+//
+//  window = np.blackman(20)
+//  print("# beta = %f" % np.mean(window))
+//
+//  plt.figure()
+//
+//  A = fft(window, 1000)
+//  mag = np.abs(A)
+//  with np.errstate(divide='ignore', invalid='ignore'):
+//      mag = 20 * np.log10(mag)
+//  mag -= max(mag)
+//  freq = np.linspace(0, len(window)/2, len(A)/2)
+//
+//  for m in mag[:len(mag)/2]:
+//  	print(m)
+//
+//  plt.plot(freq, mag[:len(mag)/2])
+//  plt.title("Spectral leakage")
+//  plt.ylabel("Amplitude (dB)")
+//  plt.xlabel("DFT bin")
+//
+//  plt.show()
+// and then be exported to leakage and compared with the Gonum
+// implementation.
+package main
+
+import (
+	"bufio"
+	"flag"
+	"fmt"
+	"image/color"
+	"io"
+	"log"
+	"math"
+	"math/cmplx"
+	"os"
+	"strconv"
+	"strings"
+
+	"gonum.org/v1/gonum/dsp/fourier"
+	"gonum.org/v1/gonum/dsp/window"
+	"gonum.org/v1/gonum/floats"
+	"gonum.org/v1/gonum/stat"
+	"gonum.org/v1/plot"
+	"gonum.org/v1/plot/plotter"
+	"gonum.org/v1/plot/vg"
+)
+
+var windows = map[string]*builtin{
+	"rectangular": {
+		name: func(_ float64) string { return "Rectangular" },
+		fn:   func(_ float64) func([]float64) []float64 { return window.Rectangular },
+		ok:   func(_ float64) bool { return true },
+	},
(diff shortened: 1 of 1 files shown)


dsp/window/cmd/leakage/leakage.go
```diff
@@ -0,0 +1,384 @@
...
+		gammaLine,
+		deltaF0Line,
+		deltaFhalfLine,
+		spectrumLines,
+	)
+	p.Legend.Add(fmt.Sprintf("ΔF_0 = %.3v", c.deltaF0), deltaF0Line)
+	p.Legend.Add(fmt.Sprintf("ΔF_0.5 = %.3v", c.deltaFhalf), deltaFhalfLine)
+	p.Legend.Add(fmt.Sprintf("K = %.3v", c.k()), &blank)
+	p.Legend.Add(fmt.Sprintf("ɣ_max = %.3v", c.gammaMax), gammaLine)
+	p.Legend.Add(fmt.Sprintf("β = %.3v", c.beta), &blank)
```
kortschak, line 246: @sbinet I had difficulties getting the LaTeX equivalent of this working. when I replace these with the following and have the needed imports and setup I get a panic.
```
	p.Legend.Add(fmt.Sprintf(`${\Delta}F_0 = %.3v$`, c.deltaF0), deltaF0Line)
	p.Legend.Add(fmt.Sprintf(`${\Delta}F_{0.5} = %.3v$`, c.deltaFhalf), deltaFhalfLine)
	p.Legend.Add(fmt.Sprintf(`$K = %.3v$`, c.k()), &blank)
	p.Legend.Add(fmt.Sprintf(`${\gamma}_{max} = %.3v$`, c.gammaMax), gammaLine)
	p.Legend.Add(fmt.Sprintf(`$\beta = %.3v$`, c.beta), &blank)
```
The prefix of the panic is
```

goroutine 1 [running]:
github.com/go-latex/latex/mtex.(*visitor).Visit(0xc0001720e0, 0x7223d8, 0xc000126558, 0x0, 0x0)
	/home/daniel/pkg/mod/github.com/go-latex/latex@v0.0.0-20210118124228-b3d85cf34e07/mtex/parser.go:128 +0xec8
<snip>
```
This appears to be related to the subscript underscore.

kortschak: Also, the line spacing with the LaTeX handling looks pretty tight. This is probably related to the issues you filed wrt font description parameters.

Here is the non-LaTeX output

sine_20_1000_go_nolatex

and here is the LaTeX-using output (with removal of {} wrapping which swallows the text and the subscripts which panic).

sine_20_1000_go

kortschak:
If https://golang.org/issue/39513 is implemented, the output from this tool could be included in the docs.

dsp/window/cmd/leakage/leakage.go
```diff
@@ -0,0 +1,384 @@
...
+//  for m in mag[:len(mag)/2]:
+//  	print(m)
+//
+//  plt.plot(freq, mag[:len(mag)/2])
+//  plt.title("Spectral leakage")
+//  plt.ylabel("Amplitude (dB)")
+//  plt.xlabel("DFT bin")
+//
+//  plt.show()
+// and the be exported to leakage and compared with the Gonum
```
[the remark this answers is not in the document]
kortschak: Thanks. Fixed.

dsp/window/cmd/leakage/leakage.go
sbinet: yeah, the handling (or lack thereof) of subscripts/superscripts is a known limitation of `go-latex`.
(that's partly whay I've started `star-tex.org/x/tex`)
- https://github.com/go-latex/latex/issues/17

as for the lack of an adequate linegap for the legend, that's news to me. (but may very well be related to https://github.com/gonum/plot/issues/676)

kortschak: Yeah, I wen't looking and then saw that issue. The other thing that I noticed is the `{text}` doesn't end up rendering "text". I think it should.

dsp/window/cmd/leakage/leakage.go
```diff
@@ -0,0 +1,387 @@
...
+
+func main() {
+	name := flag.String("window", "", "specify a built-in window name")
+	param := flag.Float64("param", math.NaN(), "specify parameter for parametric windows")
+	symm := flag.Bool("symm", true, "specify whether the window is symmetric")
+	n := flag.Int("n", 20, "specify window length")
+	m := flag.Int("m", 1000, "specify sample length (must be greater than n)")
+	csv := flag.String("csv", "", "specify an input file of frequency amplitudes (dB transformed)")
+	out := flag.String("o", "", "specify output file for plot")
+	width := flag.Float64("width", 16, "specify plot width (cm)")
```
sbinet, line 148: perhaps use [font.ParseLength](https://pkg.go.dev/gonum.org/v1/plot@v0.9.0/font#ParseLength) to allow for nicer units handling? (_e.g.:_ `42cm`).

sbinet (approved):
LGTM.

perhaps add a minimal test that exercizes the `main` function? (so, probably split it into `main` and "real `main`")

dsp/window/cmd/leakage/leakage.go
kortschak: The documentation becomes unwieldy if I do that (I need to explain what units are possible and what absent units means).

sbinet: ok.

kortschak:
How confident are you with the leakage parameter estimation code? It took me a fair while to find enough information about how to calculate the values and yet they disagree with some published values (for example the values that are currently used in the documentation) and they vary with the sample and window length parameters. It would be good to have input from someone who uses DSP a lot.

It's not clear to me how I would test the program. Here also input from a DSP expert would help.

scott-cotton:
lgtm

dsp/window/cmd/leakage/leakage.go
```diff
@@ -0,0 +1,387 @@
...
+		t = append(t, v)
+	}
+
+	xy = make(plotter.XYs, len(t))
+	for i, a := range t {
+		if math.IsInf(a, -1) {
+			a = min
+		} else if a < min {
+			min = a
+		}
```
sbinet, line 342: I vaguely remember the Go style recommending to use `switch` for this case.
(feel free to ignore, though)
```suggestion
		switch {
		case math.IsInf(a, -1):
			a = min
		case if a < min:
			min = a
		}
```

sbinet (dismissed):
LGTM (with a minor stylistic comment)

it'd be great to have a test/example or a link to a gonum.org-hosted blog post explaining and showing how this can be used.
(the package-level doc comment is great and a good starting point though.)

dsp/window/cmd/leakage/leakage.go
kortschak: Looking more carefully at this, the logic is not correct; it works approximately right for most cases but can deviate from the intention with some inputs.

kortschak:
Yeah, my intention is to use the comment outputs for the documentation. I could think about a blog post.

PTAL

sbinet (approved):

PR #1601, opened 2021-04-05
blas/gonum,internal/asm/f32: harmonise gemv implementations in blas/gonum

kortschak:
This essentially kicks the can down a road a bit, but prepares the way for assembly implementations with tests now already and waiting in f32.

Please take a look.


kortschak:
Testing on arm64 passes locally.

vladimir-ch (approved):

PR #1603, opened 2021-04-09
lapack/gonum: check for overflow in Dtgsja

vladimir-ch:
Please take a look.

I don't have a failing test case for this change but it was done in the reference in https://github.com/Reference-LAPACK/lapack/pull/502 and looks like a safe change to make. The PR references a test case for CGGSVD3 which does trigger a failure but it looks like a deficiency of our test (too strict by not taking into account the norm of the input matrices).


kortschak:
it will be nice when fuzzing is built in and there is support for structured data. The first part of that is coming and the second part is proposed. With that, finding test cases will be easier for things like this.

kortschak (approved):

PR #1605, opened 2021-04-12
stat: fix TOC formula in documentation

kortschak:
This formula is simpler, has the ntp[0] included in the series and has the additional advantage of being correct.

Please take a look.

### DIFF
--- a/stat/roc.go
+++ b/stat/roc.go
@@ -139,14 +139,13 @@ func ROC(cutoffs, y []float64, classes []bool, weights []float64) (tpr, fpr, thr
 // The returned ntp values can be interpreted as the number of true positives
 // where values above the given rank are assigned class true for each given
 // rank from 1 to len(classes).
-//  ntp_i = sum_{j = 1}^{len(ntp)-i-1} [ classes_j ] * weights_j, where [x] = 1 if x else 0.
-// and
-//  ntp_0 = 0
+//  ntp_i = sum_{j ≥ len(ntp)-1 - i} [ classes_j ] * weights_j, where [x] = 1 if x else 0.
 // The values of min and max provide the minimum and maximum possible number
 // of false values for the set of classes. The first element of ntp, min and
 // max are always zero as this corresponds to assigning all data class false
 // and the last elements are always weighted sum of classes as this corresponds
-// to assigning every data class true.
+// to assigning every data class true. For len(classes) != 0, the lengths of
+// min, ntp and max are len(classes)+1.
 //
 // If weights is nil, all weights are treated as 1. When weights are not nil,
 // the calculation of min and max allows for partial assignment of single data


stat/roc.go
```diff
@@ -139,9 +139,7 @@ func ROC(cutoffs, y []float64, classes []bool, weights []float64) (tpr, fpr, thr
 // The returned ntp values can be interpreted as the number of true positives
 // where values above the given rank are assigned class true for each given
 // rank from 1 to len(classes).
-//  ntp_i = sum_{j = 1}^{len(ntp)-i-1} [ classes_j ] * weights_j, where [x] = 1 if x else 0.
-// and
-//  ntp_0 = 0
+//  ntp_i = sum_{j ≥ len(ntp)-1 - i} [ classes_j ] * weights_j, where [x] = 1 if x else 0.
```
romanwerpachowski, line 142: `>=` or `>`?

kortschak: `>=`

romanwerpachowski: But then we say "The first element of ntp, min and max are always zero"?

kortschak: The formula naturally does that.

This is the quadratic time version of the code:
```
func TOC(classes []bool, weights []float64) (min, ntp, max []float64) {
	if weights != nil && len(classes) != len(weights) {
		panic("stat: slice length mismatch")
	}
	if len(classes) == 0 {
		return nil, nil, nil
	}

	ntp = make([]float64, len(classes)+1)
	min = make([]float64, len(ntp))
	max = make([]float64, len(ntp))
	if weights == nil {
		for i := range ntp {
			var sum float64
			for j := len(ntp) - 1 - i; j < len(classes); j++ {
				if classes[j] {
					sum++
				}
			}
			ntp[i] = sum
		}
		totalPositive := ntp[len(ntp)-1]
		for i := range ntp {
			min[i] = math.Max(0, totalPositive-float64(len(classes)-i))
			max[i] = math.Min(totalPositive, float64(i))
		}
		return min, ntp, max
	}

	cumw := max // Reuse max for cumulative weight. Update its elements last.
	for i := range ntp {
		var sum, sumw float64
		for j := len(ntp) - 1 - i; j < len(classes); j++ {
			w := weights[j]
			sumw += w
			if classes[j] {
				sum += w
			}
		}
		ntp[i] = sum
		cumw[i] = sumw
	}
	totw := cumw[len(cumw)-1]
	totalPositive := ntp[len(ntp)-1]
	for i := range ntp {
		min[i] = math.Max(0, totalPositive-(totw-cumw[i]))
		max[i] = math.Min(totalPositive, cumw[i])
	}
	return min, ntp, max
}
```

romanwerpachowski: Ah, right, because `len(ntp) == len(classes) + 1`.

romanwerpachowski: The documentation doesn't actually say that, though. I think it should.

romanwerpachowski (approved):

stat/roc.go
kortschak: It's sort of implied, but I am happy to add it explicitly. Suggest:
```diff
diff --git a/stat/roc.go b/stat/roc.go
index c95d2ed7..be6c4a80 100644
--- a/stat/roc.go
+++ b/stat/roc.go
@@ -144,7 +144,8 @@ func ROC(cutoffs, y []float64, classes []bool, weights []float64) (tpr, fpr, thr
 // of false values for the set of classes. The first element of ntp, min and
 // max are always zero as this corresponds to assigning all data class false
 // and the last elements are always weighted sum of classes as this corresponds
-// to assigning every data class true.
+// to assigning every data class true. For len(classes) != 0, the lengths of
+// min, ntp and max are len(classes)+1.
 //
 // If weights is nil, all weights are treated as 1. When weights are not nil,
 // the calculation of min and max allows for partial assignment of single data
```

romanwerpachowski: Thanks.

kortschak: PTAL

romanwerpachowski (approved):

romanwerpachowski:
Thanks.

kortschak:
Thank you for the reviews.

Issue #1606, opened 2021-04-12
mat: bounds-checking index methods refer to wrong receiver label


kortschak:
Building suite with bounds tags fails.

```
$ go install -tags bounds ./...
# gonum.org/v1/gonum/mat
mat/index_bound_checks.go:356:21: undefined: d
mat/index_bound_checks.go:359:21: undefined: d
```

PR #1607, opened 2021-04-12
ci,graph/layout,mat: fix env var handling by ci.yml

kortschak:
Please take a look.

Updates #1606.


kortschak:
The changes in ci.yml are demonstrated to catch the bug via the failure of #1608.

vladimir-ch (approved):

PR #1608, opened 2021-04-12
ci: make sure build sees env vars

kortschak:
Investigating failure to set env vars noted in #1606.

DO NOT MERGE


kortschak:
Missed the FORCE_GOARCH case.

PR #1610, opened 2021-04-13
codecov: change override branch to master

kortschak:
This is to investigate the recent failure to upload coverage reports (failure starts after March 23) after merging into master as the master branch. This is required due to the model codecov uses to calculate coverage changes.

The change here was suggested by codecov support as an experiment.

It will make no difference to the outcome of the PR coverage, but may make a difference to the merged coverage report and so we won't know until this is merged.

Please take a look.


vladimir-ch (approved):

kortschak:
Nope. This did not fix the issue. Will likely revert, but will wait until I hear back from codecov support.

vladimir-ch:
Ok, SGTM.

PR #1611, opened 2021-04-14
mat: add Tridiag to list_test

vladimir-ch:
One of the tests fails due to floating point noise, so setting a draft status while we check how to best modify the test suite.

### DIFF
--- a/mat/list_test.go
+++ b/mat/list_test.go
@@ -848,6 +848,10 @@ func makeCopyOf(a Matrix) Matrix {
 		}
 		copy(d.mat.Data, diag.mat.Data)
 		return returnAs(d, t)
+	case *Tridiag:
+		var m Tridiag
+		m.CloneFromTridiag(a.(*Tridiag))
+		return returnAs(&m, t)
 	}
 }
 
--- a/mat/list_test.go
+++ b/mat/list_test.go
@@ -238,7 +238,7 @@ func legalDims(a Matrix, m, n int) bool {
 		return true
 	case *SymDense, *TriDense, *basicSymmetric, *basicTriangular,
 		*SymBandDense, *basicSymBanded, *TriBandDense, *basicTriBanded,
-		*basicDiagonal, *DiagDense:
+		*basicDiagonal, *DiagDense, *Tridiag:
 		if m < 0 || n < 0 || m != n {
 			return false
 		}
@@ -319,6 +319,13 @@ func returnAs(a, t Matrix) Matrix {
 			return mat
 		case *basicDiagonal:
 			return asBasicDiagonal(mat)
+		}
+	case *Tridiag:
+		switch t.(type) {
+		default:
+			panic("bad type")
+		case *Tridiag:
+			return mat
 		}
 	}
 }
@@ -530,6 +537,17 @@ func makeRandOf(a Matrix, m, n int, src rand.Source) Matrix {
 		}
 		for i := 0; i < n; i++ {
 			mat.SetDiag(i, rnd.Float64())
+		}
+		rMatrix = returnAs(mat, t)
+	case *Tridiag:
+		if m != n {
+			panic("bad size")
+		}
+		mat := NewTridiag(n, nil, nil, nil)
+		for i := 0; i < n; i++ {
+			for j := max(0, i-1); j <= min(i+1, n-1); j++ {
+				mat.SetBand(i, j, rnd.NormFloat64())
+			}
 		}
 		rMatrix = returnAs(mat, t)
 	}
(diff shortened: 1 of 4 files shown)


mat/list_test.go
```diff
@@ -848,6 +866,10 @@ func makeCopyOf(a Matrix) Matrix {
 		}
 		copy(d.mat.Data, diag.mat.Data)
 		return returnAs(d, t)
+	case *Tridiag:
+		var m Tridiag
+		m.Clone(a.(*Tridiag))
```
kortschak, line 871: Sorry I missed this before, but this looks wrong; on `*Dense` we have `CloneFrom` and on `*VecDense` we have `CloneFromVec`. This seems likely to be `CloneFromTridiag`?

We unfortunately also have a residual `Clone` on `*Cholesky`. Maybe we should have a discussion about how to deal with these properly.

kortschak (commented):
The failing test is in `Trace` which has two different paths because `*Tridiag` has a `Trace` method that calls the assembly routine. We know that the for loop driven calculations disagree with the assembly at low order bits from other work, so I think we can increase the tolerance of this. I think adding a `tol float64` parameter to `testOneInputFunc` (and `testTwoInputFunc`) to bring them into line with `testOneInput` and `testTwoInput` would be appropriate.

mat/list_test.go
vladimir-ch: Yes, `Cholesky.Clone` is what I followed. The other `CloneFrom` methods take an interface but `Tridiag` has no corresponding interface besides `Banded`. What about `Tridiag.CloneFromBand` which would panic if the input's bandwidth is not `1,1`?

kortschak: I'm not sure, I don't like the idea of losing type safety when we don't need to. Maybe clone methods that take a concrete type should name the concrete type, so `Cholesky.Clone` => `Cholesky.CloneFromCholesky` and `Tridiag.Clone` => `Tridiag.CloneFromTriDiag`. They seem heavy though, and there is no risk of additional interface collision with `ClonerFrom` (which they could also be renamed to because they won't ever reasonably need to implement that), but the source would be more explicit.

vladimir-ch: SGTM and done. I'll send a PR for Cholesky.Clone.

vladimir-ch:
I've realized that it's enough to use `sameAnswerFloatApproxTol` instead of `sameAnswer` as input to `testOneInputFunc`. I've removed the Draft tag.

vladimir-ch:
I'm wondering if tests like `TestTridiagTrace` or `TestTridiagNorm` could or should be removed. They're now redundant and extra test code to maintain.

mat/matrix_test.go
```diff
@@ -542,7 +542,7 @@ func TestTrace(t *testing.T) {
 	denseComparison := func(a *Dense) interface{} {
 		return Trace(a)
 	}
-	testOneInputFunc(t, "Trace", f, denseComparison, sameAnswerFloat, isAnyType, isSquare)
+	testOneInputFunc(t, "Trace", f, denseComparison, sameAnswerFloatApproxTol(1e-15), isAnyType, isSquare)
```
kortschak, line 545: I forgot we had this.

kortschak (commented):
> I'm wondering if tests like `TestTridiagTrace` or `TestTridiagNorm` could or should be removed. They're now redundant and extra test code to maintain.

Removing them sounds fine.

vladimir-ch:
I've removed TestTridiagTrace and now rely on list_test. It's a bit unclear and inconsistent in `mat` whether we want to test a functions on concrete types (Tridiag.Trace), an interface (mat.Tracer) or a package-level function (mat.Trace).

PTAL

kortschak (approved):

kortschak:
Not removing the Norm test?

kortschak:
> It's a bit unclear and inconsistent in mat whether we want to test a functions on concrete types (Tridiag.Trace), an interface (mat.Tracer) or a package-level function (mat.Trace).

Maybe we should have an issue to decide on a policy for this?

romanwerpachowski:
I would do whatever feels most natural in the context. Tests using a higher level API are more reusable, but may fail to test some particular error conditions which are implementation-specific. In my experience, in numerical code the veil of the OOP abstraction often needs to be "pierced" to achieve the best performance, or exploit the algorithm fully. Similar with testing numerical code. E.g. the tolerances may differ between different implementation of the same abstract interface.

kortschak:
> I would do whatever feels most natural in the context. Tests using a higher level API are more reusable, but may fail to test some particular error conditions which are implementation-specific.

Yes. I think this is probably pretty much what we do. The list_test.go is there to give majority coverage with little work, but we also have specific tests for particular cases.

vladimir-ch:
> Not removing the Norm test?

Not yet because otherwise there is only a test for `mat.Norm` which doesn't dispatch to `Normer` interface and so `Tridiag.Norm` wouldn't be covered. I'm planning to do that.

> Maybe we should have an issue to decide on a policy for this?

My problem with list_test and interface testing is to identify all the range loops and switches where the new type must be added (repeat until nothing panics). Instead of interface testing I'd prefer to work locally with a concrete type, possibly with the test using a private helper that would take an interface as a parameter. But the problem is not big enough to warrant an issue or a policy. I'll pay attention to it as I work in `mat` and I'll try to clean things up.

PR #1619, opened 2021-04-28
graph/graphs/gen: check for all node id collisions

### LINKED ISSUE #1614: graph/graphs/gen: reconsider signatures for deterministic graph generation routines
The routines that were added in #1590 all return errors. It is possible to know ahead of time whether their will be multiple instances of the same node in the iterators or whether central nodes are also present in leaves or cycles. For this reason, and to be consistent with other areas of the suite, I think the error return should be elevated to panics.

/cc @vladimir-ch @mattConn

kortschak:
This incurs a greater cost for well-behaved code, but prevents subtle errors that would otherwise not be caught, for example  cycle(1,2,3,4,5,3) not being a single cycle.

Please take a look.

Closes #1614.

### DIFF
--- a/graph/graphs/gen/gen.go
+++ b/graph/graphs/gen/gen.go
@@ -54,13 +54,17 @@ func (s IDSet) Len() int       { return len(s) }
 func (s IDSet) ID(i int) int64 { return s[i] }
 
 // Complete constructs a complete graph in dst using nodes with the given IDs.
-func Complete(dst NodeIDGraphBuilder, ids IDer) error {
-	if ids.Len() == 1 {
+// If any ID appears twice in ids, Complete will panic.
+func Complete(dst NodeIDGraphBuilder, ids IDer) {
+	switch ids.Len() {
+	case 0:
+		return
+	case 1:
 		u, new := dst.NodeWithID(ids.ID(0))
 		if new {
 			dst.AddNode(u)
 		}
-		return nil
+		return
 	}
 	for i := 0; i < ids.Len(); i++ {
 		uid := ids.ID(i)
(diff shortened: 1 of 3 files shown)


graph/graphs/gen/gen.go
```diff
@@ -54,106 +54,150 @@ func (s IDSet) Len() int       { return len(s) }
...
+				return fmt.Errorf("gen: node ID collision i=%d with extra j=%d: id=%d", prev, j, uid)
+			}
+			prev -= lenIDs
+			return fmt.Errorf("gen: extra node ID collision i=%d j=%d: id=%d", prev, j, uid)
+		}
+		seen[uid] = j + lenIDs
 	}
-	return Cycle(dst, cycle)
+	return nil
 }
```
mattConn, line 203: Should there be some way to let the programmer disable checking? Supposing they have already checked the input, calling `check` would be redundant. A boolean parameter could be added to the graph-building functions to achieve this, although this would modify their signatures.

graph/graphs/gen/gen.go
```diff
@@ -54,106 +54,150 @@ func (s IDSet) Len() int       { return len(s) }
...
+// nodes in path. If any ID appears twice in path, Path will panic.
+func Path(dst NodeIDGraphBuilder, path IDer) {
+	switch path.Len() {
+	case 0:
+		return
+	case 1:
 		u, new := dst.NodeWithID(path.ID(0))
 		if new {
 			dst.AddNode(u)
 		}
```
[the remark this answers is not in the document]
kortschak: That would make it impossible to compose graphs from these primitives.

graph/graphs/gen/gen.go
kortschak: I thought about that and if they are confident enough, I'd rather they copy/paste the code that does the work in the actual routines here than add a knob. These are convenience operations and the costs here are small compared to the actual graph routines being performed in most cases.

mattConn:
Looks good to me.

mattConn (approved):
I gave it another once-over and it still looks good to me.

PR #1620, opened 2021-05-06
lapack/{gonum,testlapack}: add Dlangb and its test

vladimir-ch:
Preparation for adding `mat.Normer` and using it in `mat.Norm` similarly to `mat.Trace`.

Please take a look.

### DIFF
--- a/lapack/gonum/dlangb.go
+++ b/lapack/gonum/dlangb.go
@@ -1,0 +1,98 @@
+// Copyright ©2021 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package gonum
+
+import (
+	"math"
+
+	"gonum.org/v1/gonum/internal/asm/f64"
+	"gonum.org/v1/gonum/lapack"
+)
+
+// Dlangb returns the given norm of an n×n band matrix with kl sub-diagonals and
+// ku super-diagonals.
+//
+// When norm is lapack.MaxColumnSum, the length of work must be at least n.
+func (impl Implementation) Dlangb(norm lapack.MatrixNorm, n, kl, ku int, ab []float64, ldab int, work []float64) float64 {
+	ncol := kl + 1 + ku
+	switch {
+	case norm != lapack.MaxAbs && norm != lapack.MaxRowSum && norm != lapack.MaxColumnSum && norm != lapack.Frobenius:
+		panic(badNorm)
+	case n < 0:
+		panic(nLT0)
+	case kl < 0:
+		panic(klLT0)
+	case ku < 0:
+		panic(kuLT0)
+	case ldab < ncol:
+		panic(badLdA)
+	}
+
+	// Quick return if possible.
+	if n == 0 {
+		return 0
+	}
+
+	switch {
+	case len(ab) < (n-1)*ldab+ncol:
+		panic(shortAB)
+	case len(work) < n && norm == lapack.MaxColumnSum:
+		panic(shortWork)
+	}
+
+	var value float64
+	switch norm {
+	case lapack.MaxAbs:
+		for i := 0; i < n; i++ {
+			l := max(0, kl-i)
+			u := min(n+kl-i, ncol)
+			for _, aij := range ab[i*ldab+l : i*ldab+u] {
+				aij = math.Abs(aij)
+				if aij > value || math.IsNaN(aij) {
+					value = aij
+				}
+			}
+		}
+	case lapack.MaxRowSum:
+		for i := 0; i < n; i++ {
+			l := max(0, kl-i)
+			u := min(n+kl-i, ncol)
+			sum := f64.L1Norm(ab[i*ldab+l : i*ldab+u])
+			if sum > value || math.IsNaN(sum) {
+				value = sum
+			}
+		}
+	case lapack.MaxColumnSum:
+		work = work[:n]
+		for j := range work {
+			work[j] = 0
+		}
+		for i := 0; i < n; i++ {
+			l := max(0, kl-i)
+			u := min(n+kl-i, ncol)
+			for jb, aij := range ab[i*ldab+l : i*ldab+u] {
+				j := l + jb - kl + i
+				work[j] += math.Abs(aij)
+			}
+		}
+		for _, sumj := range work {
+			if sumj > value || math.IsNaN(sumj) {
+				value = sumj
+			}
+		}
+	case lapack.Frobenius:
+		scale := 0.0
+		ssq := 1.0
+		for i := 0; i < n; i++ {
+			l := max(0, kl-i)
+			u := min(n+kl-i, ncol)
+			ilen := u - l
+			rowscale, rowssq := impl.Dlassq(ilen, ab[i*ldab+l:], 1, 0, 1)
+			scale, ssq = impl.Dcombssq(scale, ssq, rowscale, rowssq)
+		}
+		value = scale * math.Sqrt(ssq)
+	}
+	return value
+}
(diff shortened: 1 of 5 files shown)


lapack/lapack64/lapack64.go
```diff
@@ -466,6 +466,20 @@ func Lange(norm lapack.MatrixNorm, a blas64.General, work []float64) float64 {
...
 
+// Langb returns the given norm of a square, n×n band matrix with kl sub-diagonals and
+// ku super-diagonals.
+//
+// When norm is lapack.MaxColumnSum, the length of work must be at least n.
+//
+// Dlangb is not part of the lapack.Float64 interface and so calls to Langb are always
+// executed by the Gonum implementation.
+func Langb(norm lapack.MatrixNorm, a blas64.Band, work []float64) float64 {
+	if a.Rows != a.Cols {
```
vladimir-ch, line 477: This is unfortunate but there's no way around it. Lapack only computes norms of square band matrices. That means that this restriction would also propagate to `mat` and that's even more unfortunate. Since Lapacke doesn't provide this function, I think we're free to modify the signature. So that's one option.

On the other hand, Lapack only computes norms of square band matrices because it calls DLANGB only from two routines for solving square linear systems which I think is the principal utility of band matrices. So the other option would be to make a breaking change and allow only square band matrices in `mat`. Otherwise `mat` will never do anything useful with non-square band matrices anyway.

vladimir-ch: I was too quick. The solve routines (DGBTRS, DGBSV, ...) take square matrices but the LU factorization (DGBTRF) of course works for general matrices. So there's only the first option.

vladimir-ch: I've now pushed a commit that extends Dlangb to accept non-square band matrices.

lapack/gonum/dlangb.go
```diff
@@ -0,0 +1,100 @@
...
+			work[j] = 0
+		}
+		for i := 0; i < min(m, n+kl); i++ {
+			l := max(0, kl-i)
+			u := min(n+kl-i, ncol)
+			for jb, aij := range ab[i*ldab+l : i*ldab+u] {
+				j := l + jb - kl + i
+				work[j] += math.Abs(aij)
+			}
+		}
```
kortschak, line 81: Can this be done easily with `f64.L1NormInc` directly into `work`?

lapack/testlapack/dlangb.go
```diff
@@ -0,0 +1,110 @@
...
+
+	name := fmt.Sprintf("m=%v,n=%v,kl=%v,ku=%v,ldab=%v", m, n, kl, ku, ldab)
+
+	// Generate a random band matrix.
+	ab := randomSlice(m*ldab, rnd)
+	// Sometimes put a NaN into the matrix.
+	if m > 0 && n > 0 && rnd.Float64() < 0.5 {
+		i := rnd.Intn(m)
+		ab[i*ldab+kl] = math.NaN()
+	}
```
kortschak, line 54: internal/testrand would be useful here, but the `extreme` type is not exported. Is it worth exporting that so that it's not just configurable in the command line and is more generally usable? (separate PR)

lapack/lapack64/lapack64.go
vladimir-ch: For completeness: I checked `mat.LU` and noticed that we limit the factorization only to square matrices. I remember we had a discussion on this topic in the past. Then the only remaining "native" operation on general, non-square banded matrices is MulVec supported by DGBMV. Still, it seems that it's more reasonable to leave banded matrices as general in `mat` and change Dlangb.

kortschak: This seems reasonable to me.

lapack/gonum/dlangb.go
vladimir-ch: I'm sure it's doable: a formula for the top element (row and col in the band storage) in a column and a formula for the number of elements in a column ... at least `inc` would clearly be `ldab-1`. Uninviting but then work wouldn't be needed ...

lapack/testlapack/dlangb.go
vladimir-ch: Looking into how internal/testrand could be used in testlapack is on my todo list ...

vladimir-ch:
I don't understand how the codecov failure is happening.

kortschak:
> I don't understand how the codecov failure is happening.

The parameter check panics are never exercised. This can be ignored.

lapack/gonum/dlangb.go
vladimir-ch: I've pushed a commit that does it. It's not that bad, the index calculations are pretty similar but still any idea for simplification is welcome. Now `work` is not needed, which makes Dlangb an outlier among the norm routines. I don't have benchmarks for what is faster: iterating with increment one or using assembly (on amd64) but I think we can keep this.

vladimir-ch:
> The parameter check panics are never exercised.

Ok, I misunderstood the codecov report. I thought it failed because of the negative delta in optimize.

kortschak:
The change in optimize is not big enough to make a failure. It's irritating, because it's not consistent, but I haven't been able to stabilise it by repetition.

lapack/testlapack/dlangb.go
```diff
@@ -0,0 +1,109 @@
...
+
+func DlangbTest(t *testing.T, impl Dlangber) {
+	rnd := rand.New(rand.NewSource(1))
+	for _, norm := range []lapack.MatrixNorm{lapack.MaxAbs, lapack.MaxRowSum, lapack.MaxColumnSum, lapack.Frobenius} {
+		t.Run(normToString(norm), func(t *testing.T) {
+			for _, n := range []int{0, 1, 2, 3, 4, 5, 10} {
+				for _, m := range []int{0, 1, 2, 3, 4, 5, 10} {
+					for _, kl := range []int{0, 1, 2, 3, 4, 5, 10} {
+						for _, ku := range []int{0, 1, 2, 3, 4, 5, 10} {
+							for _, ldab := range []int{kl + ku + 1, kl + ku + 1 + 7} {
```
vladimir-ch, line 30: I removed the `iter` loop here because it increases the test run time disproportionately and the NaN element will be inserted in some cases anyway.

lapack/gonum/dlangb.go
kortschak: I can't see any simplifications in that. It might be interesting to see what the performance differences are though.

kortschak (dismissed):

lapack/gonum/dlangb.go
vladimir-ch: Hmm, there might be an issue with NaN handling. I don't reproduce it locally but I've seen a failure in CI after one of the pushes with `kl=ku=0`: the test expected a NaN but got a 0 (I think). Looking at `f64.L1NormInc`, I think it boils down to how MAXPD and MAXSD behave. https://www.felixcloutier.com/x86/maxsd says:

> If only one value is a NaN (SNaN or QNaN) for this instruction, the second source operand, either a NaN or a valid floating-point value, is written to the result.

That is, NaN may not be propagated as expected.

vladimir-ch: Ok, I understand. The failure is in noasm build when `ldab=1`, then `inc=0` and the for loop in the stub doesn't execute and 0 is returned. So either pass `max(1,ldab-1)` (stride should be always at least 1) or change the stubs (the asm and noasm versions should ideally behave identically).

vladimir-ch: I think it's all clear now, can you please re-approve? I'll add the benchmark later.

kortschak: Nice work. While I was looking into I found a nasty bug in extreme.

kortschak (approved):

PR #1622, opened 2021-05-07
mat: add Normer interface and use it in Norm

vladimir-ch:
Please take a look.

### DIFF
--- a/mat/matrix.go
+++ b/mat/matrix.go
@@ -11,7 +11,6 @@ import (
 	"gonum.org/v1/gonum/blas/blas64"
 	"gonum.org/v1/gonum/floats/scalar"
 	"gonum.org/v1/gonum/lapack"
-	"gonum.org/v1/gonum/lapack/lapack64"
 )
 
 // Matrix is the basic matrix interface type.
@@ -749,6 +748,11 @@ func Min(a Matrix) float64 {
 	}
 }
 
+// A Normer can compute a norm of the matrix.
+type Normer interface {
+	Norm(norm float64) float64
+}
+
 // Norm returns the specified norm of the matrix A. Valid norms are:
 //  1 - The maximum absolute column sum
 //  2 - The Frobenius norm, the square root of the sum of the squares of the elements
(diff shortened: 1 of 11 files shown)


mat/matrix.go
```diff
@@ -759,55 +763,19 @@ func Min(a Matrix) float64 {
 func Norm(a Matrix, norm float64) float64 {
 	r, c := a.Dims()
 	if r == 0 || c == 0 {
-		panic(ErrShape)
+		panic(ErrZeroLength)
```
vladimir-ch, line 766: Trace on an empty matrix returns zero, so I think we should either remove the panic in Norm or add it to Trace.

mat/matrix_test.go
```diff
@@ -501,7 +501,7 @@ func TestNormZero(t *testing.T) {
 			if !panicked {
 				t.Errorf("expected panic for Norm(&%T{}, %v)", a, norm)
 			}
-			if message != ErrShape.Error() {
+			if message != ErrZeroLength.Error() {
```
vladimir-ch, line 504: Idealy, I'd like to remove TestNormZero altogether.

mat/matrix.go
kortschak: SGTM. Which would you prefer? I'm happy either way.

mat/matrix.go
```diff
@@ -749,6 +748,11 @@ func Min(a Matrix) float64 {
 	}
 }
 
+// A Normer can compute a norm of the matrix.
+type Normer interface {
+	Norm(norm float64) float64
```
kortschak, line 753: Note that valid values of `norm` here?

mat/matrix.go
```diff
@@ -759,55 +763,19 @@ func Min(a Matrix) float64 {
...
-		}
-		return lapack64.Lantr(n, rm, work)
-	case RawSymmetricer:
-		rm := rma.RawSymmetric()
-		n := normLapack(norm, aTrans)
-		if n == lapack.MaxRowSum || n == lapack.MaxColumnSum {
-			work = getFloats(rm.N, false)
-			defer putFloats(work)
-		}
-		return lapack64.Lansy(n, rm, work)
```
kortschak, line 790: Do we want to get rid of these?

mat/matrix.go
vladimir-ch: I left it loose because I was thinking if we want to allow Normers to deviate from the valid values, for example a potential SVD.Norm(2) could return the spectral norm instead of the Frobenius norm.

kortschak: Fair enough.

mat/matrix.go
kortschak: I forgot `untransposeExtract` does all that work for us. This is fine.

kortschak (approved):

mat/matrix.go
vladimir-ch: This is almost a philosophical topic similar to empty set.

We can view an empty matrix as a zero matrix of any size; then Trace, Norm, etc would naturally return zero without panic but that zero would be different from those operations on non-empty matrices. However, does this distinction matter? This would be similar as returning zero by definition/convention.

We can also view it as an empty container and start from there. Trace is a sum and a sum of an empty sequence is by convention 0. Frobenius norm is also a sum. But 1-norm and inf-norm are _maximum_ over columns/rows and maximum over an empty set would be ... minus infinity? NaN? That seems worse than a panic. And panicking only for certain norms is too much hassle for a minor corner case.

Or we can panic always. So we're where we started :-)

I think I'm slightly leaning toward the first option - return 0?

kortschak: I'm OK with returning 0 for the trace. If we do that we should probably return the supremum for 1-norm and inf-norm since it seems like the most useful definition; if returning a value from an empty matrix is seen as OK.

vladimir-ch: It seems to me that you'd prefer a panic for empty matrices and I agree that it's much easier to consistently do it (also over time) in contrast to returning values like -inf from Norm (or mat.Max). Should we then settle on: operations on empty matrices panic unless used as a destination/receiver?

PR #1625, opened 2021-05-11
mat: add DiagDense.Norm

vladimir-ch:
Please take a look.

### DIFF
--- a/mat/diagonal.go
+++ b/mat/diagonal.go
@@ -5,6 +5,8 @@
 package mat
 
 import (
+	"math"
+
 	"gonum.org/v1/gonum/blas"
 	"gonum.org/v1/gonum/blas/blas64"
 )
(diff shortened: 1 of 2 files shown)


mat/diagonal.go
```diff
@@ -324,3 +326,24 @@ func (d *DiagDense) Trace() float64 {
 	return tr
 
 }
+
+// Norm returns the specified norm of the receiver. Valid norms are:
+//  1 or Inf - The maximum diagonal element magnitude
+//  2 - The Frobenius norm, the square root of the sum of the squares of
+//      the diagonal elements
+//
+// Norm will panic with ErrNormOrder if an illegal norm is specified.
```
kortschak, line 335: " and with ErrShape if the receiver has zero size"

vladimir-ch: Done.

kortschak (approved):

PR #1628, opened 2021-05-12
mat: remove unused BandWidther interface

### LINKED ISSUE #1626: mat: unused BandWidther interface
The `mat.BandWidther` interface is unused and no matrix type in `mat` fulfills it:
```
type BandWidther interface {
	BandWidth() (k1, k2 int)
}
```
Since there is the `Banded` interface:
```
type Banded interface {
	Matrix
	Bandwidth() (kl, ku int)
	TBand() Banded
}
```
I think that `BandWidther` should be removed.

vladimir-ch:
Please take a look.

Closes #1626

### DIFF
--- a/mat/matrix.go
+++ b/mat/matrix.go
@@ -183,12 +183,6 @@ type Grower interface {
 	Grow(r, c int) Matrix
 }
 
-// A BandWidther represents a banded matrix and can return the left and right half-bandwidths, k1 and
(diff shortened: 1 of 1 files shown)


kortschak (approved):

kortschak:
Added stability because formally, it does even though it won't affect anyone anywhere.

PR #1631, opened 2021-05-12
mat: return ErrZeroLength when size is zero and document it

vladimir-ch:
Please take a look.

### DIFF
--- a/mat/vector.go
+++ b/mat/vector.go
@@ -269,7 +269,8 @@ func (v *VecDense) CopyVec(a Vector) int {
 //  2 - The Euclidean norm, the square root of the sum of the squares of the elements
 //  Inf - The maximum element magnitude
 //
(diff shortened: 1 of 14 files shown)


mat/vector.go
```diff
@@ -269,7 +269,8 @@ func (v *VecDense) CopyVec(a Vector) int {
 //  2 - The Euclidean norm, the square root of the sum of the squares of the elements
 //  Inf - The maximum element magnitude
 //
-// Norm will panic with ErrNormOrder if an illegal norm is specified.
+// Norm will panic with ErrNormOrder if an illegal norm is specified and with
+// ErrZeroLength if the vector has zero size..
```
kortschak, line 273: Two dots.

kortschak (approved):

Issue #1623, opened 2021-05-07
blas,lapack: use safe scaling algorithms from Anderson (2017)


vladimir-ch:
From the abstract of https://doi.org/10.1145/3061665:

> The square root of a sum of squares is well known to be prone to overflow and underflow. Ad hoc scaling of intermediate results, as has been done in numerical software such as the BLAS and LAPACK, mostly avoids the problem, but it can still occur at extreme values in the range of representable numbers. More careful scaling, as has been implemented in recent versions of the standard algorithms, may come at the expense of performance or clarity. This work reimplements the vector 2-norm and the generation of Givens rotations from the Level 1 BLAS to improve their performance and design. In addition, support for negative increments is extended to the Level 1 BLAS operations on a single vector, and a comprehensive test suite for all the Level 1 BLAS is included.

It would be nice to use these algorithms in {blas,lapack}/gonum. We already added one of them in #1575 but there are others in that paper: DNRM2, DLASSQ, DROTG. The Reference started using them in:

* DROTG: https://github.com/Reference-LAPACK/lapack/pull/527
* DNRM2: https://github.com/Reference-LAPACK/lapack/pull/514
* DLASSQ: https://github.com/Reference-LAPACK/lapack/pull/494

A problem with our Dnrm2 is that it calls `f64.L2Norm` which is written in assembly and not easy to change.

PR #1634, opened 2021-05-17
blas/gonum: rewrite Drotg and extend its test

vladimir-ch:
The test is copied from `testlapack/dlartg.go` with a small adjustment for Drotg.

Please take a look.

Updates #1623


blas/gonum/level1float64.go
```diff
@@ -237,43 +237,78 @@ func (Implementation) Daxpy(n int, alpha float64, x []float64, incX int, y []flo
...
+	// ACM Trans. Math. Softw. 44, 1, Article 12 (July 2017), 28 pages.
+	// DOI: https://doi.org/10.1145/3061665
+
+	const safmin = 0x1p-1022
+	const safmax = 1 / safmin
+
+	anorm := math.Abs(a)
+	bnorm := math.Abs(b)
+	switch {
+	case bnorm == 0:
```
vladimir-ch, line 278: I put this case first to preserve what Drotg returns for `a=b=0`. I'm at the same time asking the reference in https://github.com/Reference-LAPACK/lapack/issues/557 whether they would consider restoring the previous return values, otherwise we have to adapt (the code here and the test expectation).

blas/gonum/level1float64.go
```diff
@@ -237,43 +237,78 @@ func (Implementation) Daxpy(n int, alpha float64, x []float64, incX int, y []flo
...
-	} else if c != 0 { // r == 0 case handled above
-		z = 1 / c
-	} else {
+	// Implementation based on Supplemental Material to:
+	// Edward Anderson. 2017. Algorithm 978: Safe Scaling in the Level 1 BLAS.
+	// ACM Trans. Math. Softw. 44, 1, Article 12 (July 2017), 28 pages.
+	// DOI: https://doi.org/10.1145/3061665
+
+	const safmin = 0x1p-1022
+	const safmax = 1 / safmin
```
kortschak, line 273:
```
const (
    safmin = 0x1p-1022
    safmax = 1 / safmin
)
```

blas/gonum/level1float64.go
kortschak: I think it's worth adding a comment to this effect if that issue is not resolved before this is merged.

kortschak (changes requested):
`go generate` will need to be run.

blas/gonum/level1float64.go
vladimir-ch: Yes, good idea. Done.

blas/gonum/level1float64.go
vladimir-ch: Done.

blas/gonum/single_precision.bash
```diff
@@ -35,6 +35,7 @@ cat level1float64.go \
       -e 's_^// Id_// Is_' \
       -e 's_"gonum.org/v1/gonum/internal/asm/f64"_"gonum.org/v1/gonum/internal/asm/f32"_' \
       -e 's_"math"_math "gonum.org/v1/gonum/internal/math32"_' \
+      -e 's_safmin = 0x1p-1022_safmin = 0x1p-126_' \
```
vladimir-ch, line 38: I'm not completely happy with this, it's too specific but I can't quickly think of a better solution. I thought about having a file like `constants.go` with `dsafmin, dsafmax, ssafmin, ssafmax` and then here change from the d-prefix to the s-prefix. However, having these constants even only at package level feels like making them too public, they won't be needed elsewhere in BLAS.

kortschak: We can live with it for the time being. I'm still wondering how we will deal with precision-dependent constants when we shift to a generics-based approach (assuming we do, but it's something that I'd like to do since it will reduce line count and improve testing).

blas/gonum/level1float32.go
```diff
@@ -251,45 +251,84 @@ func (Implementation) Saxpy(n int, alpha float32, x []float32, incX int, y []flo
...
+		c = 0
+		s = 1
+		r = b
 		z = 1
-	}
-	return
+	default:
+		maxab := math.Max(anorm, bnorm)
+		scl := math.Min(math.Max(safmin, maxab), safmax)
+		var sigma float32
```
kortschak, line 311: Ahhh! We need `math32.Min` and `.Max`.

vladimir-ch: Oh, right! Done.

blas/gonum/single_precision.bash
vladimir-ch: Ok, I'll leave it in. Regarding generics, I'm not following that topic much, not enough bandwidth, but sounds like something that would be nice to do.

blas/gonum/level1float64.go
vladimir-ch: The Reference has made the same change in https://github.com/Reference-LAPACK/lapack/pull/560, so I've removed the comment.

vladimir-ch:
PTAL

kortschak (approved):

PR #1640, opened 2021-05-20
stat: panic when calling CDF or Quantile with empty x

kortschak:
Please take a look.

### DIFF
--- a/stat/stat.go
+++ b/stat/stat.go
@@ -56,6 +56,7 @@ func Bhattacharyya(p, q []float64) float64 {
 //
 // The x data must be sorted in increasing order. If weights is nil then all
 // of the weights are 1. If weights is not nil, then len(x) must equal len(weights).
+// CDF will panic if the length of x is zero.
 //
 // CumulantKind behaviors:
 //  - Empirical: Returns the lowest fraction for which q is greater than or equal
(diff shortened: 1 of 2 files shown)


sbinet (approved):

romanwerpachowski:
Not sure about it. It makes no sense to calculate a quantile for an empty array. Panicking seems to be the right behaviour. Returning zero can lead to hard to detect bugs in user code.

kortschak:
R returns NULL, which we obviously can't do. This is related to the discussion @vladimir-ch and I had in https://github.com/gonum/gonum/pull/1622#discussion_r628281871 where we came to the same conclusion. The rationale for zero was that any quantile between infinities is 0, but NaN and panic are also reasonable. For consistency with #1622 and following the same rationale I'd be happy to make an explicit panic here and document that it panics, retaining the tests.

romanwerpachowski:
>  I'd be happy to make an explicit panic here and document that it panics,
retaining the tests

I'd vote for that.

kortschak:
`CDF` should also probably panic with empty input and does with the same currently uninformative message. I'll do that too.

kortschak:
PTAL

romanwerpachowski (approved):
I suggest renaming the PR now :)

kortschak:
Done (I had remembered at the beginning, but not the end — I did notice that the branch name was agnostic, which made me happy at the time, and then I moved on).

Issue #532, opened 2018-06-14
[Proposal] Want to collaborate with ODE integrator


Daniel-M:
### What are you trying to do?
Hey, I did a go package for ODE integration with some basic methods. I came aware today of this project and I'd like to propose or adapt this to gonum package
My package is at  [https://github.com/Daniel-M/odeint](https://github.com/Daniel-M/odeint) it started as an excercise but I figure it could grow as something useful.

Daniel-M:
@vladimir-ch thanks for the comments and the reading recommendation, I'm grateful for that.

* Though `complex128` or `complex64` aren't really popular numerical types to problem solving, I feel that an integrator without complex numbers integrators would be incomplete.

* I think that implementing float64 and complex128 methods would be enough to provide standard ODE integrators.

* I'm aware that the package might lack a proper coding style. It started as an excercise to learn more about Go, and I figured it could be useful someday. I lack the proper technical formation in Go (and general programming) to better implement the numerical methods in a resource efficient fashion and I really would like to collaborate in adding ODE integrators support to gonum, if you could provide guidance of some sort, I'd be glad to give it a try.

vladimir-ch:
Not updated in a long time, there is now a new initiative in https://github.com/gonum/exp/pull/54, closing.

PR #1641, opened 2021-05-21
add ivp package and subpackages for integrating initial value problems

soypat:
Added Runge Kutta 4th order multivariable integration scheme. Example added simulates a falling point-mass.

I expect some more changes may be requested. I'm all ears!

Checklist:

- API changes have been discussed: No API changes
- code is goformated correctly (goimports): Check
- packages with generated code have had code generation run: No code gen
- tests pass locally: Tests added pass!
- linked to relevant issues: None I'm aware of

### DIFF
--- a/ivp/doc.go
+++ b/ivp/doc.go
@@ -1,0 +1,7 @@
+// Copyright ©2016 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Package ivp provides interfaces and structures to solve
+// multivariable ODE initial value problems.
+package ivp // import "gonum.org/v1/gonum/ivp"
--- a/ivp/ivp.go
+++ b/ivp/ivp.go
@@ -1,0 +1,45 @@
+// Copyright ©2017 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package ivp
+
+import (
+	"gonum.org/v1/gonum/mat"
+)
+
+// IVP defines an initial value problem.
+type IVP interface {
+	// Initial values vector for state variables x and inputs u.
+	IV() (x0, u0 mat.Vector)
+	// Equations returns the coupled, non-linear algebraic differential
+	// equations for the state variables (x) and the functions for inputs (u).
+	// The input functions (ueq) are not differential equations but rather
+	// calculated directly from a given x vector and current input vector.
+	// Results are stored in y which are the length of x and u, respectively.
+	// The scalar (float64) argument is the domain over which the
+	// problem is integrated, which is usually time for most physical problems.
+	//
+	// If problem has no input functions then u supplied and ufunc returned
+	// may be nil. x equations my not be nil.
+	Equations() (xeq, ufunc func(y []float64, domain float64, x, u []float64))
+	// Dimensions of x state variables and u inputs
+	Dims() (nx, nu int)
+}
+
+// Integrator abstracts algorithm specifics. For anyone looking to
+// implement it, Set(ivp) should be called first to initialize the IVP with
+// initial values. Step will calculate the next x values and store them in y
+// u values should not be stored as they can easily be obtained if one has
+// x values. Integrator should store 1 or more (depending on algorithm used)
+// of previously calculated x values to be able to integrate.
+type Integrator interface {
+	// Set initializes an initial value problem. First argument
+	// is the initial domain integration point, is usually zero.
+	Set(float64, IVP) error
+	// Step integrates IVP and stores result in y. step is a suggested step
+	// for the algorithm to take. The algorithm may decide that it is not sufficiently
+	// small or big enough (these are adaptive algorithms) and take a different step.
+	// The resulting step is returned as the first parameter
+	Step(y []float64, step float64) (float64, error)
+}
--- a/ivp/model.go
+++ b/ivp/model.go
@@ -1,0 +1,46 @@
+// Copyright ©2017 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package ivp
+
+import "gonum.org/v1/gonum/mat"
+
+// Model implements IVP interface. X vector and equations can not be nil or zero length.
+type Model struct {
+	x0, u0   mat.Vector
+	xeq, ins func(y []float64, dom float64, x, u []float64)
+}
+
+// IV returns initial values of the IVP. First returned parameter is the
+// starting x vector and second parameter are inputs when solving non-autonomous
+// ODEs.
+func (m *Model) IV() (mat.Vector, mat.Vector) { return m.x0, m.u0 }
+
+// Equations returns differential equations relating to state vector x and input functions
+// for non-autonomous ODEs.
+//
+// Input functions may be nil (ueq).
+func (m *Model) Equations() (xeq, ueq func(y []float64, dom float64, x, u []float64)) {
+	return m.xeq, m.ins
+}
+
+// Dims returns dimension of state and input (x length and u length, respectively).
+// Dims has some dimension checking involved and can serve as a preliminary system verifier.
+//
+// Dims panics if x vector is nil.
+func (m *Model) Dims() (nx, nu int) {
+	if m.x0 == nil {
+		panic("x vector can not be nil")
+	}
+	if m.u0 == nil {
+		nu = 0
+	}
+	return m.x0.Len(), nu
+}
+
+// NewModel returns a IVP given initial conditions (x0,u0), differential equations (xeq) and
+// input functions for non-autonomous ODEs (ueq).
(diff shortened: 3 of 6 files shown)


ivp/ivp.go
```diff
@@ -0,0 +1,41 @@
+package ivp
```
kortschak, line 1: Please add appropriate copyright headers to all files.

ivp/ivp.go
```diff
@@ -0,0 +1,41 @@
...
+
+import (
+	"gonum.org/v1/gonum/mat"
+)
```
kortschak, line 5: Single line, please.

ivp/sim/ivpmodel.go
```diff
@@ -0,0 +1,32 @@
+package sim
```
kortschak, line 1: I don't think we need a proliferation of directories here. This can be merged with ivp/ivp.go.

ivp/solver/algorithm_test.go
```diff
@@ -0,0 +1,113 @@
+package solver_test
```
kortschak, line 1: Again, this can go into ivp.

ivp/solver/algorithms.go
```diff
@@ -0,0 +1,68 @@
+package solver
```
kortschak, line 1: Same.

ivp/sim/ivpmodel.go
```diff
@@ -0,0 +1,32 @@
...
+	"gonum.org/v1/gonum/mat"
+)
+
+// Model implements IVP interface. X vector can not be nil or zero length.
+type Model struct {
+	x0, u0   mat.Vector
+	xeq, ins func(y []float64, dom float64, x, u []float64)
+}
+
+func (m *Model) IV() (mat.Vector, mat.Vector) { return m.x0, m.u0 }
```
kortschak, line 13: Please put godoc comments on exported functions.

ivp/solver/algorithm_test.go
```diff
@@ -0,0 +1,113 @@
...
+			y[0], y[1] = dom*dom/2., dom
+		}, nil)
+	if err != nil {
+		t.Fatal(err)
+	}
+	Quadratic.solution = quadsol
+	Quadratic.err = func(h, i float64) float64 { return math.Pow(h*i, 4) }
+}
+func TestRK4Quadratic(t *testing.T) {
+	setup(t)
```
kortschak, line 45: Put the code in-line rather than calling out to setup.

ivp/solver/algorithm_test.go
```diff
@@ -0,0 +1,113 @@
...
+			if got > expected {
+				t.Errorf("error %e greater than threshold %e", got, expected)
+			}
+
+		}
+	}
+}
+
+func Example_fallingBall() {
+	const (
```
kortschak, line 79: Example tests are for teaching, so this needs more commentary to lead the user through how to use the package.

kortschak (changes requested):
Thank you for the contribution.

The should have a proposal issue for discussion, but I've gone through and done an initial review and what is here can be used as a focus for discussion there.

I think the most appropriate person to involve in the discussion would be @vladimir-ch, but time availability may be a limiting.

We normally wait until approval before including the changes to A+C since there will be revisions to code in the PR and we don't want non-logic changes tangled with logic changes. So would you please reset hard to the first commit of this branch? We'll ask you to send another PR for the A+C change when the time comes.

soypat:
@kortschak Thanks for the suggestions. I've applied them to the new commit after a hard reset.

ivp/ivp.go
```diff
@@ -0,0 +1,45 @@
+// Copyright ©2017 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package ivp
+
+import (
+	"gonum.org/v1/gonum/mat"
```
soypat, line 8: go imports failed me. should i reset hard again?

kortschak: No, that's fine, it can be cleaned up after discussion. Do you want to open a proposal to add this package?

ivp/doc.go
```diff
@@ -0,0 +1,7 @@
+// Copyright ©2016 The Gonum Authors. All rights reserved.
```
sbinet, line 1:
```suggestion
// Copyright ©2021 The Gonum Authors. All rights reserved.
```

ivp/ivp.go
```diff
@@ -0,0 +1,45 @@
+// Copyright ©2017 The Gonum Authors. All rights reserved.
```
sbinet, line 1:
```suggestion
// Copyright ©2021 The Gonum Authors. All rights reserved.
```

ivp/rungekutta4.go
```diff
@@ -0,0 +1,71 @@
+// Copyright ©2017 The Gonum Authors. All rights reserved.
```
sbinet, line 1:
```suggestion
// Copyright ©2021 The Gonum Authors. All rights reserved.
```

ivp/rungekutta4_test.go
```diff
@@ -0,0 +1,129 @@
+// Copyright ©2017 The Gonum Authors. All rights reserved.
```
sbinet, line 1:
```suggestion
// Copyright ©2021 The Gonum Authors. All rights reserved.
```

ivp/model.go
```diff
@@ -0,0 +1,46 @@
+// Copyright ©2017 The Gonum Authors. All rights reserved.
```
sbinet, line 1:
```suggestion
// Copyright ©2021 The Gonum Authors. All rights reserved.
```

sbinet (commented):
it's great somebody is picking up the slack on this front.

I haven't looked at the PR/proposal in detail yet, but after my 10,000ft diagonal read, I wondered: wouldn't it make sense to use the same kind of mechanism/protocol for the iteration than what we used for `optimize`? would it make sense?

kortschak:
> I wondered: wouldn't it make sense to use the same kind of mechanism/protocol for the iteration than what we used for optimize? would it make sense?

This is why I thought Vladimir would be interested. Also the relation to ODE solvers which has been on his distant radar for a long time.

vladimir-ch:
Thanks for the contribution!

@kortschak is right, an IVP solver package has been on my distant radar for a very long time, almost from the beginning of my involvement with Gonum, see https://groups.google.com/g/gonum-dev/c/gbsTBk2xY6M/m/bxZWbXJiBAAJ. Occasionally an external nudge comes showing interest in such package, for example https://github.com/gonum/gonum/issues/532 in 2018, but getting the API and implementation right is a major undertaking which doesn't fit well with my limited availability and now in the pandemic also with limited uninterrupted time to focus.

Recently I've been focusing more on the `mat` and `lapack` packages to have the basics (a more feature-complete support for solving linear systems, more matrix types like (symmetric) banded or tridiagonal matrices, maybe even sparse matrices, iterative solvers on which I have been procrastinating for several years now, ...) in place before proceeding to these more specialized packages. My reasoning was that implicit solvers need to solve linear systems at every step so let's have the base done but now I think the work on an ode package should proceed in smaller steps and start with explicit solvers only.

If we had an elementary working Go implementation of an explicit Runge-Kutta solver like [DOPRI5](http://www.unige.ch/~hairer/prog/nonstiff/dopri5.f) in http://www.unige.ch/~hairer/software.html, that would be a great start. This particular implementation is short, just 700 lines including 200 lines of documentation, and yet has stepsize control and dense output. We could use it as a basis for an API discussion and build a test suite around it.

I think this PR is not ready for merging directly into Gonum, so I'd propose to instead submit it to [the exp repo](https://github.com/gonum/exp). I also suggest to go through those old discussions, I think that a lot of the ideas are still valid.

If you want to take on this project, my availability is better recently and so I may have some time for discussions and code review.

ivp/ivp.go
```diff
@@ -0,0 +1,45 @@
...
+	Dims() (nx, nu int)
+}
+
+// Integrator abstracts algorithm specifics. For anyone looking to
+// implement it, Set(ivp) should be called first to initialize the IVP with
+// initial values. Step will calculate the next x values and store them in y
+// u values should not be stored as they can easily be obtained if one has
+// x values. Integrator should store 1 or more (depending on algorithm used)
+// of previously calculated x values to be able to integrate.
+type Integrator interface {
```
vladimir-ch, line 36: This looks as a good starting point but the question arises how the user will interact with the package. Will he instantiate a integrator and implement a loop to step forward? Then this interface is not needed (the user already has a concrete type). Or will this package implement the integration loop and take the integrator as this (or similar) interface?

soypat:
@vladimir-ch I've taken your suggestions and applied some of them here: https://github.com/gonum/exp/pull/54

I'll be adding Dormand-Prince method soon if you say so.

vladimir-ch:
> I'll be adding Dormand-Prince method soon if you say so.

I mentioned the `dopri5.f` code not because of the method itself but rather because of it being a solid, small but still fairly capable implementation of an IVP solver. Let's ignore it for now and focus on what you have.

PR #1646, opened 2021-05-27
testlapack: consolidate floating-point constants

vladimir-ch:
I prefer the more descriptive names than the `dlamch*` names and I think it's useful to have them at package level instead of scattered around in functions.

Please take a look.

### DIFF
--- a/lapack/testlapack/dgeev.go
+++ b/lapack/testlapack/dgeev.go
@@ -769,10 +769,9 @@ func residualRightEV(a, e blas64.General, wr, wi []float64) float64 {
 	}
 	bi.Dgemm(blas.NoTrans, blas.NoTrans, n, n, n, 1, a.Data, a.Stride, e.Data, e.Stride, -1, r, ldr)
 
-	unfl := dlamchS
-	ulp := dlamchE
-	anorm := math.Max(dlange(lapack.MaxColumnSum, n, n, a.Data, a.Stride), unfl)
-	enorm := math.Max(dlange(lapack.MaxColumnSum, n, n, e.Data, e.Stride), ulp)
+	const eps = dlamchE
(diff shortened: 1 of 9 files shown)


lapack/testlapack/dlatrs.go
```diff
@@ -103,8 +103,11 @@ func dlatrsResidual(uplo blas.Uplo, trans blas.Transpose, diag blas.Diag, n int,
 		}
 	}
 
-	eps := dlamchE
-	smlnum := dlamchS
+	const (
+		eps    = dlamchE
+		smlnum = safmin
```
vladimir-ch, line 108: This name collision is unfortunate.

kortschak: Is it worth calling this something else, `small`, to differentiate it?

lapack/testlapack/dgeev.go
```diff
@@ -769,9 +769,7 @@ func residualRightEV(a, e blas64.General, wr, wi []float64) float64 {
 	}
 	bi.Dgemm(blas.NoTrans, blas.NoTrans, n, n, n, 1, a.Data, a.Stride, e.Data, e.Stride, -1, r, ldr)
 
-	unfl := dlamchS
-	ulp := dlamchE
```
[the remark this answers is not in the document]
kortschak: ED is to me a precariously stacked house of cards, so I prefer that we not deviate.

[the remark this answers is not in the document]
vladimir-ch: Ha ha, that's very true. I've changed it back to dlamchE.

lapack/testlapack/dlatrs.go
vladimir-ch: Yes, that's better. I went with `tiny` and `huge` which is quite accurate and differentiates them from smlnum and bignum.

kortschak (approved):

PR #1652, opened 2021-06-05
lapack/gonum: add Dgesc2

soypat:
So the checklist is NOT complete. I intend for this to be a first review of how LAPACK routines are implemented in Go and get some feedback. I'm sure my indexing is messed up, was quite confused seeeing how fortran alternated betweeen double and single indexing.

No tests implemented.


lapack/gonum/idamax.go
```diff
@@ -0,0 +1,39 @@
...
+import (
+	"math"
+)
+
+// Idamax finds the index of the first element having maximum absolute value.
+//
+//  n is number of elements in input vector(s)
+//  dx is array, dimension ( 1 + ( N - 1 )*abs( incx ) )
+//  incx is storage spacing between elements of dx
+func (Implementation) Idamax(n int, dx []float64, incx int) (idamax int) {
```
kortschak, line 16: `Idamax` is [implemented in BLAS level 1](https://pkg.go.dev/gonum.org/v1/gonum/blas#Float64Level1), so this file is not needed.

lapack/gonum/dlabad.go
```diff
@@ -0,0 +1,22 @@
...
+
+// DLABAD takes as input the values computed by DLAMCH for underflow and
+// overflow, and returns the square root of each of these values if the
+// log of LARGE is sufficiently large.  This subroutine is intended to
+// identify machines with a large exponent range, such as the Crays, and
+// redefine the underflow and overflow limits to be the square roots of
+// the values computed by DLAMCH.  This subroutine is needed because
+// DLAMCH does not compensate for poor arithmetic in the upper half of
+// the exponent range, as is found on a Cray.
+func (impl Implementation) Dlabad(small, large float64) (float64, float64) {
```
kortschak, line 17: As mentioned by @vladimir-ch we don't use `Dlabad` since we use IEEE 754, so this file can be deleted.

lapack/gonum/dgesdc2.go
```diff
@@ -0,0 +1,50 @@
...
+
+package gonum
+
+import (
+	"math"
+
+	"gonum.org/v1/gonum/blas/blas64"
+)
+
+// Dgesdc2 solves a system of linear equations
```
kortschak, line 13: `Dgesc2`? and rename file?

lapack/gonum/dgesdc2.go
```diff
@@ -0,0 +1,50 @@
...
+//   A * X = scale* RHS
+// with a general N-by-N matrix A using the LU factorization with
+// complete pivoting computed by Dgetc2.
+func (impl Implementation) Dgesdc2(n int, a []float64, lda int, rhs []float64, ipiv, jpiv []int, scale float64) {
+	const eps = dlamchP
+	var smlnum, _ = impl.Dlabad(dlamchS/eps, 1./(dlamchS/eps))
```
kortschak, line 19: `const smlnum = dlamchS/dlamchP`

lapack/gonum/dgesdc2.go
```diff
@@ -0,0 +1,50 @@
...
+	if len(a) < (n-1)*lda+n {
+		panic(shortA)
+	}
+	//Apply permutations IPIV to RHS
```
kortschak, line 23: `// Apply permutations ipiv to RHS.` (final dot and space between comment marks to distinguish from directives). Though I'm not sure if this comment is needed (@vladimir-ch?).

Also separate this comment from the conditional panic by a blank line.

lapack/gonum/dgesdc2.go
```diff
@@ -0,0 +1,50 @@
...
+	impl.Dlaswp(1, rhs, lda, 1, n-1, ipiv, 1)
+
+	// solve for L part
```
kortschak, line 26: `// Solve for L part.`

lapack/gonum/dgesdc2.go
```diff
@@ -0,0 +1,50 @@
...
+	for i := 0; i < n; i++ {
```
kortschak, line 27: `i < n-1`?

lapack/gonum/dgesdc2.go
```diff
@@ -0,0 +1,50 @@
...
+		// of course this can be optimized
+		// I'm still struggling with this indexing.
+		for j := i + 1; j < n; j++ {
+			rhs[j] = rhs[j] - a[(i*n)+j]*rhs[i]
```
kortschak, line 31: `rhs[j] -= float64(a[j*lda + i]*rhs[i])`

Yes, this is currently optimised for column-major storage.

Apparent unnecessary conversion is there to prevent FMS on arm64.

lapack/gonum/dgesdc2.go
```diff
@@ -0,0 +1,50 @@
...
+		}
+	}
+	scale = 1.
+	//
```
kortschak, line 35:
```
<blank line>
// Solve for U part.

scale = 1.0

// Check for scaling.
```

lapack/gonum/dgesdc2.go
```diff
@@ -0,0 +1,50 @@
...
+	i := impl.Idamax(n, rhs, 1)
```
kortschak, line 36:
```
bi := blas64.Implementation()
i := bi.Iamax(blas64.Vector{N: n, Data: rhs, Inc: 1})
```

lapack/gonum/dgesdc2.go
```diff
@@ -0,0 +1,50 @@
...
+	if 2*smlnum*math.Abs(rhs[i]) > math.Abs(a[n*n+n]) {
```
kortschak, line 37: `math.Abs(a[n*lda+n])`

lapack/gonum/dgesdc2.go
```diff
@@ -0,0 +1,50 @@
...
+		temp := 0.5 / math.Abs(rhs[i])
+		blas64.Implementation().Dscal(n, temp, rhs[1:], 1) // what?
```
kortschak, line 39:
```
bi.Scal(rhs[0], blas64.Vector{N: n, Data: temp, Inc: 1})
```

This is scaling `temp` by the first element of `rhs`. The blas64 API takes a vector type here.

lapack/gonum/dgesdc2.go
```diff
@@ -0,0 +1,50 @@
...
+		scale = scale * temp
```
kortschak, line 40:
```
scale *= temp
```

lapack/gonum/dgesdc2.go
```diff
@@ -0,0 +1,50 @@
...
+	}
+	for i := n; i > 0; i-- {
```
kortschak, line 42:
```
for i := n-1; i >= 0; i-- {
```

lapack/gonum/dgesdc2.go
```diff
@@ -0,0 +1,50 @@
...
+		temp := 1 / a[(i*n)+i]
```
kortschak, line 43:
```
temp := 1 / a[i*lda + i]
```

lapack/gonum/dgesdc2.go
```diff
@@ -0,0 +1,50 @@
...
+		rhs[i] = rhs[i] * temp
```
kortschak, line 44:
```
rhs[i] *= temp
```

lapack/gonum/dgesdc2.go
```diff
@@ -0,0 +1,50 @@
...
+		for j := i + 1; j < n; j++ {
+			rhs[i] = rhs[i] - rhs[j]*a[(i*n)+j]*temp
```
kortschak, line 46:
```
rhs[i] -= float64(rhs[j]*a[i*lda + j]*temp)
```

lapack/gonum/dgesdc2.go
soypat: Right, I see it now. Thanks!

lapack/gonum/dgesdc2.go
soypat: Is this possible? I've found `Iamax` in namespace which calls Idamax and Idamax in Implementation:
```go
// one way.
bi := blas64.Implementation()
i := bi.Idamax(n,rhs,1)
// second way i managed.
i := blas64.Iamax(blas64.Vector{N: n, Data: rhs, Inc: 1})
```
Which one?

kortschak: The first one since we already know `Inc` >= 0.

lapack/gonum/dgesdc2.go
soypat: Again I'm getting the same problem where `blas64.Scal` has the signature you describe while `bi` has `Dscal` available.

Second observation: `temp` is not an array (or is it??). I'm seeing it the other way around, a part of `rhs` (a single element?) is being scaled by `temp`.  Now I'm looking at it more closely, I think it's scaling `rhs[0:]`. Please confirm.

kortschak: > Again I'm getting the same problem where `blas64.Scal` has the signature you describe while `bi` has `Dscal` available.

Sorry, `bi.Dscal`.

> `temp` is not an array (or is it??). I'm seeing it the other way around, a part of `rhs` (a single element?) is being scaled by temp. Now I'm looking at it more closely, I think it's scaling `rhs[0:]`. Please confirm.

Again, sorry. You are right. `temp` is a scalar and `rhs(1)` is being passed to be scaled.

lapack/gonum/dgesc2.go
```diff
@@ -0,0 +1,57 @@
...
+	"math"
+
+	"gonum.org/v1/gonum/blas/blas64"
+)
+
+// Dgesc2 solves a system of linear equations
+//   A * X = scale* RHS
+// with a general N-by-N matrix A using the LU factorization with
+// complete pivoting computed by Dgetc2.
+func (impl Implementation) Dgesc2(n int, a []float64, lda int, rhs []float64, ipiv, jpiv []int) (scale float64) {
```
kortschak, line 17: Please add that the result is placed in `rhs` on exit.

Also, since this is an auxiliary routine, it's worth adding
```
//
// Dgesc2 is an internal routine. It is exported for testing purposes.
```

lapack/gonum/dgesc2.go
```diff
@@ -0,0 +1,57 @@
...
+	const smlnum = dlamchS / dlamchP
+	if len(a) < (n-1)*lda+n {
+		panic(shortA)
+	}
+
+	// Apply permutations ipiv to RHS.
+	impl.Dlaswp(1, rhs, lda, 1, n-1, ipiv, 1)
+
+	// solve for L part.
```
kortschak, line 26: Uppercase 'S' in solve.

lapack/gonum/dgesc2.go
```diff
@@ -0,0 +1,57 @@
...
+package gonum
+
+import (
+	"math"
+
+	"gonum.org/v1/gonum/blas/blas64"
+)
+
+// Dgesc2 solves a system of linear equations
+//   A * X = scale* RHS
```
kortschak, line 14: Nit: clean up the equation, `//  A * X = scale * RHS` (two spaces between comment marks and text and single spaces around the operators).

kortschak (commented):
This looks good now. Thanks. Some minor nits and tests (take a look at how @vladimir-ch implements tests in the testlapack package) — here you would take random inputs with a variety of dimensions and strides, do the LU with `Dgetc2` (needs to be implemented - looks very straightforward) and pass in the LU part, checking on return that the equation in the docs is satisfied.

lapack/gonum/dgesc2.go
```diff
@@ -0,0 +1,57 @@
...
+		temp := 0.5 / math.Abs(rhs[i])
+		bi.Dscal(n, temp, rhs, 1)
+		scale *= temp
+	}
+
+	for i := n - 1; i >= 0; i-- {
+		temp := 1.0 / a[i*lda+i]
+		rhs[i] *= temp
+		for j := i + 1; j < n; j++ {
+			rhs[i] -= float64(rhs[j] * a[i*lda+j] * temp)
```
vladimir-ch, line 50: No need for an explicit conversion and please keep the parenthesis as in the reference.

lapack/gonum/dgesc2.go
```diff
@@ -0,0 +1,57 @@
...
+		}
+	}
+
+	// Apply permutations jpiv to the solution (rhs).
+	impl.Dlaswp(1, rhs, lda, 1, n-1, jpiv, -1)
```
vladimir-ch, line 55: `lda` must be `1`, 4. and 5. parameter `0` and `n-2`, and `jpiv` must be sliced to `n-1`.

vladimir-ch (changes requested):
This needs a test before merging for which `Dgetc2` must be implemented (with a test).

lapack/gonum/dgesc2.go
```diff
@@ -0,0 +1,57 @@
...
+		panic(shortA)
+	}
+
+	// Apply permutations ipiv to RHS.
+	impl.Dlaswp(1, rhs, lda, 1, n-1, ipiv, 1)
+
+	// solve for L part.
+	for i := 0; i < n-1; i++ {
+		for j := i + 1; j < n; j++ {
+			rhs[j] -= float64(a[j*lda+i] * rhs[i])
```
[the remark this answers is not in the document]
kortschak: This apparently unnecessary conversion was added at my request to prevent FMS generation on arm64, same below.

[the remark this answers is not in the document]
vladimir-ch: Ah, ok, I saw your comment in email but didn't relate it to this code when reviewing.

lapack/gonum/dgesc2.go
```diff
@@ -0,0 +1,57 @@
...
+// with a general N-by-N matrix A using the LU factorization with
+// complete pivoting computed by Dgetc2.
+func (impl Implementation) Dgesc2(n int, a []float64, lda int, rhs []float64, ipiv, jpiv []int) (scale float64) {
+	const smlnum = dlamchS / dlamchP
+	if len(a) < (n-1)*lda+n {
+		panic(shortA)
+	}
+
+	// Apply permutations ipiv to RHS.
+	impl.Dlaswp(1, rhs, lda, 1, n-1, ipiv, 1)
```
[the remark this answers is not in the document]
vladimir-ch: I thought it's strange that `k2` is `n-1` and not `n` in the Fortran reference. In fact it's because they rely on an undocumented implementation detail of DGETC2:
```
*
*     Set last pivots to N
*
      IPIV( N ) = N
      JPIV( N ) = N
```
I think it would be cleaner to pass `0, n-1, ipiv[:n], 1` here. What do you think, @kortschak ?

lapack/gonum/dgesc2.go
vladimir-ch: The same here: `0, n-1, jpiv[:n], -1`.

lapack/gonum/dgesc2.go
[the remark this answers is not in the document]
kortschak: Wouldn't that mean were doing an additional row of work that is in effect a no-op?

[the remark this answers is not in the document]
vladimir-ch: Yes but only because our `Dlaswp` implementation doesn't check whether it's doing a no-op swap (unlike the reference which does the check). I think it would be better to pass `n` pivots here and below and change `Dlaswp` to check for a no-op.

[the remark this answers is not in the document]
kortschak: Yeah, OK.

soypat:
Some lessons learned:
* Parenthesis matter. They can prevent precision loss and overflow if done correctly.
* `lda` is the "leading dimension" and used to index matrices, not `n`/`m` (actual dimensions)
* Documentation must be taken care of properly
* Add bounds/length checks not present in reference

As for the test, I need to generate not only the general matrix but also `ipiv` and `jpiv`. I'll take a look at other tests which do this i.e. `dlaqp2.go` and `dlaqps.go`

lapack/gonum/dgesc2.go
```diff
@@ -0,0 +1,71 @@
...
+
+// Dgesc2 solves a system of linear equations
+//  A * X = scale * RHS
+// with a general N-by-N matrix A using the LU factorization with
+// complete pivoting computed by Dgetc2. The result is placed in
+// rhs on exit.
+//
+// Dgesc2 is an internal routine. It is exported for testing purposes.
+func (impl Implementation) Dgesc2(n int, a []float64, lda int, rhs []float64, ipiv, jpiv []int) (scale float64) {
+	switch {
```
vladimir-ch, line 21: This part should look like this:
```
	switch {
	case n < 0:
		panic(nLT0)
	case lda < max(1, n):
		panic(badLdA)
    }

    // Quick return if possible.
    if n == 0 {
        return 0
    }

    switch {
	case len(a) < (n-1)*lda+n:
		panic(shortA)
	case len(rhs) < n:
		panic(shortRHS)
	case len(ipiv) != n:
	    panic(badLenIpiv)
	case len(jpiv) != n:
	    panic(badLenJpiv)
	}

	const smlnum = dlamchS / dlamchP
```

lapack/gonum/dgesc2.go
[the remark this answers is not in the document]
vladimir-ch: #1654 submitted

lapack/gonum/dgesc2.go
soypat: I've applied your suggestions. I'm still confused as to how much we have to test. Is it enough that it did not panic with a bunch of random matrices? Should the result be compared with another linear solver?

vladimir-ch:
> As for the test, I need to generate not only the general matrix but also ipiv and jpiv.

It will be much easier (and more correct) if you first implement Dgetc2 and use it in the test to generate input for Dgesc2.

soypat:
> It will be much easier (and more correct) if you first implement Dgetc2 and use it in the test to generate input for Dgesc2.

should I do it in this PR? should I start another?

lapack/gonum/dgesc2.go
vladimir-ch: We must test that, given the output from Dgetc2, Dgesc2 computes what it promises, that is, that:
```
A * X = scale * RHS
```
holds and that `scale` is in `[0,1]`.

I would start with:
1. Generate a random A, compute the norm |A|
2. Generate a random X, compute the norm |X|
3. Compute RHS from A*X
4. Call Dgetc2
5. Call Dgesc2
6. Check that scale is in [0,1]
7. Check that |A*X - scale*RHS| / |A| / |X| is small

If nothing special is done, the above steps will always return `scale = 1`. Then the next step would be to modify 1. and/or 2. to trigger a non-unit scaling.

vladimir-ch:
`Dgetc2` should be a separate PR.

soypat:
Dgesc2 is passing albeit without the test you mentioned. I tried programming it and ended up leaving the `t.Error` declaration commented because it was failing although `X` was being calculated correctly.

soypat:
*pinging*. Needed for https://github.com/gonum/gonum/pull/1658 and #1669.

worth mentioning https://github.com/gonum/gonum/pull/1665 is ready for merge too.

lapack/testlapack/dgesc2.go
```diff
@@ -0,0 +1,91 @@
+// Copyright ©2016 The Gonum Authors. All rights reserved.
```
vladimir-ch, line 1: 2021

lapack/gonum/dgesc2.go
```diff
@@ -0,0 +1,83 @@
...
+
+import (
+	"math"
+
+	"gonum.org/v1/gonum/blas/blas64"
+)
+
+// Dgesc2 solves a system of linear equations
+//  A * X = scale * RHS
+// with a general N-by-N matrix A using the LU factorization with
```
vladimir-ch, line 15: `n×n`

vladimir-ch (commented):
LGTM after the two comments. I'll fix the test later.

soypat:
@vladimir-ch Thanks for taking a look!

vladimir-ch (approved):

lapack/gonum/dgesc2.go
```diff
@@ -0,0 +1,83 @@
...
+		panic(badLenJpiv)
+	}
+
+	const smlnum = dlamchS / dlamchP
+	if len(a) < (n-1)*lda+n {
+		panic(shortA)
+	}
+
+	// Apply permutations ipiv to RHS.
+	impl.Dlaswp(1, rhs, 1, 0, n-1, ipiv[:n], 1)
```
soypat, line 50: @kortschak or @vladimir-ch , I went back to `Dgesc2` to see how I used `Dlaswp` and this is confusing me some. The reference material does `CALL dlaswp( 1, rhs, lda, 1, n-1, ipiv, 1 )`, which according to its documentations, does row swap for `k2-k1+1` rows, that is **n-1** rows. Are we not swapping **n** rows here? Or am I misunderstanding?

vladimir-ch: This was discussed in https://github.com/gonum/gonum/pull/1652#discussion_r646106943. The reference documents IPIV and JPIV to have length n, DGETC2 sets n elements but the only users of the PIV arrays, DGESC2 and DLATDF, refer only to the first n-1 elements. That's because n-1 swaps are enough to represent the permutation matrix from DGETC2 and from the way PIV arrays are constructed the last element must be n-1. DGESC2 uses this fact and does not call DLASWP with the last no-op swap.

I see three options:
1. Keep what we have: ipiv and jpiv have length n, Dgetc2 sets all elements, Dgesc2 and Dlatdf pass all n elements to Dlaswp, Dlaswp will not do anything for the last swap.
2. Do what the reference does: ipiv and jpiv have length n, Dgetc2 sets all elements but Dgesc2 and Dlatdf pass only n-1 elements to Dlaswp
3. ipiv and jpiv have length n-1, Dgetc2 sets all elements, Dgesc2 and Dlatdf pass all n-1 elements to Dlaswp.

I think what the reference does is the least clear of these options (set n elements but then use only n-1 elements for no documented, albeit logical, reason). Option 1 or option 3 are both fine, option 1 diverges less from the reference.

soypat: Thank you for clearing it up. I remember you and Dan talked about it but I didn't really understand what the fuss was all about. I do now, crystal clear explanation, btw!

Issue #1651, opened 2021-06-05
gonum: add CARE solver


soypat:
### What are you trying to do?
Add a Continuous-time algebraic Riccati equation (CARE) solver. 

If we are to follow in scipy's footsteps [source](https://github.com/scipy/scipy/blob/v1.6.3/scipy/linalg/_solvers.py#L325-L526) there are some LAPACK functions that are used that may be missing. Below are the LAPACK routines uses (dgges and dtgsen) and their dependencies. I've taken the liberty to checkmark the routines which have a file in [`gonum/lapack/gonum`](https://github.com/gonum/gonum/tree/master/lapack/gonum) (i presume that means they are implemented already):

- [ ] [dtgsen](https://github.com/scipy/scipy/blob/4ec4ab8d6ccc1cdb34b84fdcb66fde2cc0210dbf/scipy/linalg/_decomp_qz.py#L368)
- [ ] [dgges (numpylink)](https://github.com/scipy/scipy/blob/4ec4ab8d6ccc1cdb34b84fdcb66fde2cc0210dbf/scipy/linalg/_decomp_qz.py#L113) [lapack link](https://www.netlib.org/lapack/explore-html/d9/d8e/group__double_g_eeigen_ga8637d4b822e19d10327ddcb4235dc08e.html#ga8637d4b822e19d10327ddcb4235dc08e)
  - [ ] [daxpy](https://www.netlib.org/lapack/explore-html/de/da4/group__double__blas__level1_ga8f99d6a644d3396aa32db472e0cfc91c.html#ga8f99d6a644d3396aa32db472e0cfc91c)
  - [ ] [dggbal](https://www.netlib.org/lapack/explore-html/d9/d8e/group__double_g_eeigen_ga8637d4b822e19d10327ddcb4235dc08e.html#ga8637d4b822e19d10327ddcb4235dc08e)
  - [ ] [dswap](https://www.netlib.org/lapack/explore-html/de/da4/group__double__blas__level1_gaca2757ba2c3b2c6fc5d729b50345fac0.html#gaca2757ba2c3b2c6fc5d729b50345fac0)
  - [ ] [dlabad](https://www.netlib.org/lapack/explore-html/df/dd1/group___o_t_h_e_rauxiliary_ga57ac7409aedb2a380d5d0e625ffbeb7a.html#ga57ac7409aedb2a380d5d0e625ffbeb7a)
[...]

PR #1655, opened 2021-06-06
lapack/gonum: add Dgetc2

soypat:
In hopes of someday getting to add CARE solver for #1651, blas routine `dgetc2` is added.

I'd like some guidance on testing. I've looked at tests in the BLAS package and they're all quite different from what I've seen in the LAPACK package. What stands out to me particularily is the tables:

```go
for i, test := range []struct {
		ul   blas.Uplo
		tA   blas.Transpose
		d    blas.Diag
		n, k int
		a    [][]float64
		x    []float64
		incX int
		ans  []float64
	}{
		{
			ul: blas.Upper,
			tA: blas.NoTrans,
			d:  blas.NonUnit,
			n:  5,
			k:  1,
			a: [][]float64{
				{1, 3, 0, 0, 0},
				{0, 6, 7, 0, 0},
				{0, 0, 2, 1, 0},
				{0, 0, 0, 12, 3},
				{0, 0, 0, 0, -1},
			},
			x:    []float64{1, 2, 3, 4, 5},
			incX: 1,
			ans:  []float64{2.479166666666667, -0.493055555555556, 0.708333333333333, 1.583333333333333, -5.000000000000000},
		},
// ...
```
Is this the way to go?


vladimir-ch:
This should go into `lapack/{gonum,testlapack}`, it's not a BLAS routine.

The test should follow the `lapack/testlapack` style, for example `dpbtf2.go` is kind of related and relatively recent. `dtrtrs.go` is not so related but also quite recent and shows the style on which we've recently settled in lapack.

soypat:
Got confused since the page says "BLAS Level 2 *Algorithm*". will fix.

Thanks for the suggestions!

blas/gonum/level2float64.go
```diff
@@ -5,12 +5,115 @@
...
+			}
+		}
+		if i == 0 {
+			smin = math.Max(eps*xmax, smlnum)
+		}
+
+		// Swap rows.
+
+		if ipv != i {
+			impl.Dswap(n, a[ipv*lda:], lda, a[i*lda:], lda)
```
vladimir-ch: Gonum is row-major (rows are continuous in storage) while Fortran is column-major (columns are continuous in storage). So Lapack must use increment of `lda` to get to the next element in a row but for us the increment is 1. Below, where columns are swapped, the situation is reversed, Lapack increments by 1 to get to the next element in a column but we need to increment by `lda`. Therefore here:
```
impl.Dswap(n, a[ipv*lda:], 1, a[i*lda:], 1)
```

blas/gonum/level2float64.go
```diff
@@ -5,12 +5,115 @@
...
+		// Swap columns.
+
+		if jpv != i {
+			impl.Dswap(n, a[jpv:], lda, a[i:], lda)
+		}
+		jpiv[i] = jpv
+
+		// Check for singularity.
+
+		if math.Abs(a[i*lda+1]) < smin {
```
vladimir-ch: `i*lda+i` (the column index is `i`, not 1)

blas/gonum/level2float64.go
```diff
@@ -5,12 +5,115 @@
...
+			k = i
+			a[i*lda+1] = smin
```
vladimir-ch: Also here.

blas/gonum/level2float64.go
```diff
@@ -5,12 +5,115 @@
...
+		}
+
+		for j := i + 1; j < n; j++ {
+			a[j*lda+i] /= a[i*lda+i]
+		}
+		impl.Dger(n-i, n-i, -1.0, a[(i+1)*lda+i:], 1, a[i*lda+i+1:], lda, a[(i+1)*lda+i+1:], lda)
```
vladimir-ch:
```suggestion
		impl.Dger(n-i-1, n-i-1, -1.0, a[(i+1)*lda+i:], lda, a[i*lda+i+1:], 1, a[(i+1)*lda+i+1:], lda)
```

blas/gonum/level2float64.go
```diff
@@ -5,12 +5,115 @@
...
+	}
+
+	if math.Abs(a[n*lda+n]) < smin {
```
vladimir-ch: Here `n` is used as an index, so we must subtract 1:
```suggestion
	if math.Abs(a[(n-1)*lda+n-1]) < smin {
```

blas/gonum/level2float64.go
```diff
@@ -5,12 +5,115 @@
...
+		k = n
```
vladimir-ch:
```suggestion
		k = n-1
```

blas/gonum/level2float64.go
```diff
@@ -5,12 +5,115 @@
...
+		a[n*lda+n] = smin
```
vladimir-ch:
```suggestion
		a[(n-1)*lda+n-1] = smin
```

blas/gonum/level2float64.go
```diff
@@ -5,12 +5,115 @@
...
+	}
+
+	// Set last pivots to n.
+	ipiv[n-1], jpiv[n-1] = n, n
```
vladimir-ch:
```suggestion
	ipiv[n-1], jpiv[n-1] = n-1, n-1
```

blas/gonum/level2float64.go
```diff
@@ -5,12 +5,115 @@
...
+	return
```
vladimir-ch:
```suggestion
	return k
```

blas/gonum/level2float64.go
```diff
@@ -5,12 +5,115 @@
...
+					ipv, jpv = ip, jp
+				}
+			}
+		}
+		if i == 0 {
+			smin = math.Max(eps*xmax, smlnum)
+		}
+
+		// Swap rows.
+
```
vladimir-ch: Delete empty line.

blas/gonum/level2float64.go
```diff
@@ -5,12 +5,115 @@
...
+		if ipv != i {
+			impl.Dswap(n, a[ipv*lda:], lda, a[i*lda:], lda)
+		}
+		ipiv[i] = ipv
+
+		// Swap columns.
+
```
vladimir-ch: Delete empty line.

blas/gonum/level2float64.go
```diff
@@ -5,12 +5,115 @@
...
+		if jpv != i {
+			impl.Dswap(n, a[jpv:], lda, a[i:], lda)
+		}
+		jpiv[i] = jpv
+
+		// Check for singularity.
+
```
vladimir-ch: Delete empty line.

blas/gonum/level2float64.go
```diff
@@ -5,12 +5,115 @@
...
+	const smlnum = dlamchS / eps
+
+	// Handle n==1 case by itself.
+
+	if n == 1 {
+		ipiv[0], jpiv[0] = 1, 1
+		if math.Abs(a[0]) < smlnum {
+			a[0] = smlnum
+			return 0
+		}
```
vladimir-ch:
```suggestion
			k = 0
		}
		return k
```

blas/gonum/level2float64.go
```diff
@@ -5,12 +5,115 @@
...
+// n-by-n matrix A. The factorization has the form
+//  A = P * L * U * Q,
+// where P and Q are permutation matrices, L is lower triangular with
+// unit diagonal elements and U is upper triangular.
+//
+// Outputs are A, ipiv, jpiv and k. k is non-negative if U(k, k) is likely to produce overflow if
+// we try to solve for x in Ax = b. So U is perturbed to
+// avoid the overflow.
+func (impl Implementation) Dgetc2(n int, a []float64, lda int, ipiv, jpiv []int) (k int) {
+	// Negative k indicates U was not perturbed.
```
vladimir-ch: Move this before quick return.

blas/gonum/level2float64.go
```diff
@@ -5,12 +5,115 @@
...
+	switch {
+	case n < 0:
+		panic(nLT0)
+	case lda < max(1, n):
+		panic(badLdA)
+	}
+
+	// Quick return if possible.
+	if n == 0 {
+		return
```
vladimir-ch:
```suggestion
		return k
```

vladimir-ch (changes requested):
Let's see what Github does with these review comments done when the code was still in the blas directory.

vladimir-ch:
> Got confused since the page says "BLAS Level 2 Algorithm". will fix.

No worries. Some Lapack routines come in 2 variants, a simpler one that uses BLAS Level 2 and operates on columns/rows, and a more complex one that uses BLAS Level 3, operates on blocks and should be more performant. The comment here means that this routine _uses_ BLAS Level 2.

vladimir-ch:
> Let's see what Github does with these review comments done when the code was still in the blas directory.

Ok, it shows them as outdated but they are of course all still valid.

lapack/gonum/dgetc2.go
```diff
@@ -0,0 +1,112 @@
...
+		panic(badLenJpvt)
+	}
+
+	const eps = dlamchP
+	const smlnum = dlamchS / eps
+
+	// Handle n==1 case by itself.
+
+	if n == 1 {
+		ipiv[0], jpiv[0] = 1, 1
```
vladimir-ch, line 52:
```suggestion
		ipiv[0], jpiv[0] = 0, 0
```

blas/gonum/level2float64.go
soypat: darn, missed that one

blas/gonum/level2float64.go
soypat: Think you may have missed the last `lda` which should actually be a `1`, right?

soypat: Or no, that one **is** equal to `lda` since it is `lda` parameter that is passed into `Dger`

blas/gonum/level2float64.go
soypat: Woah, slipped right past me. Thanks Vladimir.

soypat:
I'm at a loss as to how to construct the permutation matrices from pivot indices. Cholesky factorization function does not have a similar implementation.

kortschak:
Why do you need to construct a permutation matrix?

blas/gonum/level2float64.go
vladimir-ch: `x` is i-th column of A, so Lapack has `inc=1`, we have `inc=lda`, `y` is (i+1)-th row of A, Lapack has `inc=lda`, we have `inc=1`. The last argument is the destination matrix, so `lda` is kept.

lapack/gonum/dgetc2.go
```diff
@@ -0,0 +1,111 @@
...
+					ipv, jpv = ip, jp
+				}
+			}
+		}
+		if i == 0 {
+			smin = math.Max(eps*xmax, smlnum)
+		}
+
+		// Swap rows.
+		bi := blas64.Implementation()
```
vladimir-ch, line 79: Move this out of the for loop, before `var smin float64`.

lapack/gonum/dgetc2.go
```diff
@@ -0,0 +1,111 @@
...
+	case len(ipiv) != n:
+		panic(badLenIpiv)
+	case len(jpiv) != n:
+		panic(badLenJpvt)
+	}
+
+	const eps = dlamchP
+	const smlnum = dlamchS / eps
+
+	// Handle n==1 case by itself.
```
vladimir-ch, line 49: nit: I think this comment is not so useful and I'd remove it.

lapack/gonum/dgetc2.go
```diff
@@ -0,0 +1,111 @@
...
+		}
+		bi.Dger(n-i-1, n-i-1, -1.0, a[(i+1)*lda+i:], 1, a[i*lda+i+1:], 1, a[(i+1)*lda+i+1:], lda)
+	}
+
+	if math.Abs(a[(n-1)*lda+n-1]) < smin {
+		k = n - 1
+		a[(n-1)*lda+(n-1)] = smin
+	}
+
+	// Set last pivots to n.
```
vladimir-ch, line 108: Either remove the comment or change `n` -> `n-1`.

vladimir-ch (changes requested):
`Dgetc2` looks almost ok now, I haven't looked at the test because you're still working on it.

soypat:
@kortschak I'm way out of my element here, but I'd though a good way to test `Dgetc2` was to reconstruct the matrix `A` from the results of the LU factorization `A = P * L * U * Q` where P and Q are permutation matrices, L is lower triangular with unit diagonal elements and U is upper triangular.

vladimir-ch:
> I'd thought that a good way to test Dgetc2 was to reconstruct the matrix A from the results of the LU factorization A = P * L * U * Q where P and Q are permutation matrices, L is lower triangular with unit diagonal elements and U is upper triangular.

Yes, reconstructing A is the way to test Dgetc2 but you don't need to form P and Q explicitly and multiply, you can use ipiv and jpiv to swap rows/columns to achieve the same effect.

> What my test is doing thus far is checking if the permutation indices yield unique combinations, which I'm not even sure is an expected result. It is failing for now.

Testing the content of ipiv and jpiv is easy and will not hurt. I would fill them with negative numbers before calling Dgetc2, then make their copies, sort the copies and then check that ipivCopy[i] == i for all i. Then I would use ipiv, jpiv to reconstruct A.

soypat:
> Testing the content of ipiv and jpiv is easy and will not hurt. I would fill them with negative numbers before calling Dgetc2, then make their copies, sort the copies and then check that ipivCopy[i] == i for all i. Then I would use ipiv, jpiv to reconstruct A.

I got around to implementing this and letting it fail. I've read the code and compared it with reference but nothing seems out of place. Any ideas on what I should be looking out for?

vladimir-ch:
That's my fault, I didn't pay enough attention to how the permutations are represented in ipiv and jpiv. Now I see there is not much that can be tested except that all elements of ipiv, jpiv are set, in your case that no element is -1. After that you should focus on checking that A can be reconstructed from the factorization.

soypat:
>  you can use ipiv and jpiv to swap rows/columns to achieve the same effect.
There's definitely more to reconstructing A than swapping columns and rows. I did just that in my test but noticed the lines 
```go
for j := i + 1; j < n; j++ {
			a[j*lda+i] /= a[i*lda+i]
		}
		bi.Dger(n-i-1, n-i-1, -1.0, a[(i+1)*lda+i:], lda, a[i*lda+i+1:], 1, a[(i+1)*lda+i+1:], lda)
```
modify `a` in every case. There's gotta be a better way of testing this than programming `inverseDgetc2`.

vladimir-ch:
> There's definitely more to reconstructing A than swapping columns and rows.

Yes, Dgetc2 stores the L and U triangular factors in-place into a (and P and Q into ipiv and jpiv).

soypat:
@kortschak This branch is ready to merge. `Dgetc2` passes all tests.

lapack/gonum/dgetc2.go
```diff
@@ -0,0 +1,109 @@
...
+	case len(a) < (n-1)*lda+n:
+		panic(shortA)
+	case len(ipiv) != n:
+		panic(badLenIpiv)
+	case len(jpiv) != n:
+		panic(badLenJpvt)
+	}
+
+	const eps = dlamchP
+	const smlnum = dlamchS / eps
```
kortschak, line 47:
```
const (
    eps    = dlamchP
    smlnum = dlamchS / eps
)
```

lapack/gonum/dgetc2.go
```diff
@@ -0,0 +1,109 @@
...
+// Dgetc2 computes an LU factorization with complete pivoting of the
+// n-by-n matrix A. The factorization has the form
+//  A = P * L * U * Q,
+// where P and Q are permutation matrices, L is lower triangular with
+// unit diagonal elements and U is upper triangular.
+//
+// Outputs are A, ipiv, jpiv and k. k is non-negative if U(k, k)
+// is likely to produce overflow when  e try to solve for x in Ax = b.
+// So U is perturbed to avoid the overflow.
+func (impl Implementation) Dgetc2(n int, a []float64, lda int, ipiv, jpiv []int) (k int) {
```
kortschak, line 22: Dgetc2 is auxiliary, so note that it's only exported for testing purposes.

```
//
// Dgetc2 is an internal routine. It is exported for testing purposes.
```

lapack/gonum/dgetc2.go
```diff
@@ -0,0 +1,109 @@
...
+package gonum
+
+import (
+	"math"
+
+	"gonum.org/v1/gonum/blas/blas64"
+)
+
+// Dgetc2 computes an LU factorization with complete pivoting of the
+// n-by-n matrix A. The factorization has the form
```
kortschak, line 14: s/n-b-n/n×n/

lapack/gonum/dgetc2.go
```diff
@@ -0,0 +1,109 @@
...
+//  A = P * L * U * Q,
+// where P and Q are permutation matrices, L is lower triangular with
+// unit diagonal elements and U is upper triangular.
+//
+// Outputs are A, ipiv, jpiv and k. k is non-negative if U(k, k)
+// is likely to produce overflow when  e try to solve for x in Ax = b.
+// So U is perturbed to avoid the overflow.
```
kortschak, line 21: This doesn't look like our normal style of docs. I'll let @vladimir-ch comment on this.

lapack/gonum/dgetc2.go
```diff
@@ -0,0 +1,109 @@
...
+		ipiv[0], jpiv[0] = 0, 0
+		if math.Abs(a[0]) < smlnum {
+			a[0] = smlnum
+			k = 0
+		}
+		return k
+	}
+
+	// Factorize A using complete pivoting.
+	// Set pivots less than SMIN to SMIN.
```
kortschak, line 59: lc/SMIN/g

lapack/gonum/dgetc2.go
```diff
@@ -0,0 +1,109 @@
...
+	var smin float64
+	var ipv, jpv int
+	bi := blas64.Implementation()
+	for i := 0; i < n-1; i++ {
+		xmax := 0.0
+		for ip := i; ip < n; ip++ {
+			for jp := i; jp < n; jp++ {
+				if math.Abs(a[ip*lda+jp]) >= xmax {
+					xmax = math.Abs(a[ip*lda+jp])
+					ipv, jpv = ip, jp
```
kortschak, line 69: I'm not sure that this is clearer on a single line. ¯\\\_ (ツ)\_/¯

lapack/gonum/dgetc2.go
```diff
@@ -0,0 +1,109 @@
...
+		// Check for singularity.
+		if math.Abs(a[i*lda+i]) < smin {
+			k = i
+			a[i*lda+i] = smin
+		}
+
+		for j := i + 1; j < n; j++ {
+			a[j*lda+i] /= a[i*lda+i]
+		}
+		bi.Dger(n-i-1, n-i-1, -1.0, a[(i+1)*lda+i:], lda, a[i*lda+i+1:], 1, a[(i+1)*lda+i+1:], lda)
```
kortschak, line 98: s/-1.0/-1/

lapack/gonum/dgetc2.go
```diff
@@ -0,0 +1,109 @@
...
+	}
+
+	if math.Abs(a[(n-1)*lda+n-1]) < smin {
+		k = n - 1
+		a[(n-1)*lda+(n-1)] = smin
+	}
+
+	// Set last pivots to last index.
+	ipiv[n-1], jpiv[n-1] = n-1, n-1
```
kortschak, line 107: Again, I think this probably should be two lines.

kortschak, line 107: All this get simpler if there is a `n--` at the beginning. Too clever, @vladimir-ch?

lapack/testlapack/dgetc2.go
```diff
@@ -0,0 +1,136 @@
+// Copyright ©2015 The Gonum Authors. All rights reserved.
```
kortschak, line 1: 2021

lapack/testlapack/dgetc2.go
```diff
@@ -0,0 +1,136 @@
...
+
+package testlapack
+
+import (
+	"fmt"
+	"math"
+	"testing"
+
+	"golang.org/x/exp/rand"
+	"gonum.org/v1/gonum/blas"
```
kortschak, line 13: Blank line between us and them.

lapack/testlapack/dgetc2.go
```diff
@@ -0,0 +1,136 @@
...
+
+func Dgetc2Test(t *testing.T, impl Dgetc2er) {
+	const tol = 1e-12
+	rnd := rand.New(rand.NewSource(1))
+	for _, n := range []int{0, 1, 2, 3, 4, 5, 10, 20} {
+		for _, lda := range []int{n} {
+			dgetc2Test(t, impl, rnd, n, lda, tol)
+		}
+	}
+	// specific matrix cases
```
kortschak, line 29: Capitalisation and final dot.

lapack/testlapack/dgetc2.go
```diff
@@ -0,0 +1,136 @@
...
+	// Generate a random general matrix A.
+	a := randomGeneral(n, n, lda, rnd)
+
+	// ipiv and jpiv are outputs.
+	ipiv := make([]int, n)
+	jpiv := make([]int, n)
+	for i := 0; i < n; i++ {
+		ipiv[i], jpiv[i] = -1, -1 // Set to non-indices.
+	}
+	// Copy to store output (LU decomposition)
```
kortschak, line 81: Final dot.

lapack/gonum/dgetc2.go
soypat: I've looked at `Dgelq2` to make the documentation better for last commit.

lapack/testlapack/dgetc2.go
```diff
@@ -0,0 +1,137 @@
...
+
+	// ipiv and jpiv are outputs.
+	ipiv := make([]int, n)
+	jpiv := make([]int, n)
+	for i := 0; i < n; i++ {
+		ipiv[i], jpiv[i] = -1, -1 // Set to non-indices.
+	}
+	// Copy to store output (LU decomposition).
+	lu := make([]float64, len(a.Data))
+	copy(lu, a.Data)
```
vladimir-ch, line 84:
```suggestion
	lu := cloneGeneral(a)
```

lapack/testlapack/dgetc2.go
```diff
@@ -0,0 +1,137 @@
...
+	k := impl.Dgetc2(n, lu, lda, ipiv, jpiv)
+	if k >= 0 {
+		t.Fatalf("%v: matrix was perturbed at %d", name, k)
```
vladimir-ch, line 87: I'm not sure this deserves a `Fatalf`. It's not an indication of a bug, it's about the input data. We're not hitting it but please change to `Logf`.

lapack/testlapack/dgetc2.go
```diff
@@ -0,0 +1,137 @@
...
+					L[idx] = 1.0
+				}
+			} else if i > j { // On diagonal or lower triangle.
+				L[idx] = lu[idx]
+			}
+		}
+	}
+
+	work := make([]float64, n*n)
+	bi.Dgemm(blas.NoTrans, blas.NoTrans, n, n, n, 1, L, lda, U, lda, 0, work, lda)
```
vladimir-ch, line 118: This will break once `lda != n`. Either use `ldl := n` etc or use `zeros` to allocate `L`, `U`, `work` and use their `Stride` field here.

lapack/testlapack/dgetc2.go
```diff
@@ -0,0 +1,137 @@
...
+		if ipv != i {
+			bi.Dswap(n, work[i*lda:], 1, work[ipv*lda:], 1)
+		}
+		if jpv != i {
+			bi.Dswap(n, work[i:], lda, work[jpv:], lda)
+		}
+	}
+
+	// A should be reconstructed by now.
+	for i := range lu {
```
vladimir-ch, line 132: This again assumes that `lu` (`a`) and `work` slices have the same length which will not be true when `lda != n`.

vladimir-ch (changes requested):
Also please remove `lapack/.gitignore`.

lapack/gonum/dgetc2.go
vladimir-ch: It's ok. I'll polish it after it's merged.

lapack/testlapack/dgetc2.go
```diff
@@ -0,0 +1,137 @@
...
+			dgetc2Test(t, impl, rnd, n, lda, tol)
+		}
+	}
+	// Specific matrix cases.
+	for _, tc := range []struct {
+		name       string
+		a, expect  blas64.General
+		ipiv, jpiv []int
+	}{
+		{name: "identity", a: eye(3, 3), expect: eye(3, 3), ipiv: []int{2, 2, 2}, jpiv: []int{2, 2, 2}},
```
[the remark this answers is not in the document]
soypat: Those are test cases I created by running LAPACK's `Dgetc2` and copying the output as `expect`. `"identity"` test case is supposed to be a property based test: the identity matrix is already LU.

lapack/testlapack/dgetc2.go
```diff
@@ -0,0 +1,137 @@
...
+		{
+			name:   "small",
+			a:      blas64.General{Rows: 3, Cols: 3, Stride: 3, Data: []float64{1, 2, 3, 2, 1, 6, 3, 6, 0}},
```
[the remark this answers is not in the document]
soypat: I chose an `A` matrix that had a rational LU decomposition. I'm not sure what you mean by the second sentence.

lapack/testlapack/dgetc2.go
```diff
@@ -0,0 +1,137 @@
...
+
+func Dgetc2Test(t *testing.T, impl Dgetc2er) {
+	const tol = 1e-12
+	rnd := rand.New(rand.NewSource(1))
+	for _, n := range []int{0, 1, 2, 3, 4, 5, 10, 20} {
+		for _, lda := range []int{n} {
+			dgetc2Test(t, impl, rnd, n, lda, tol)
+		}
+	}
+	// Specific matrix cases.
```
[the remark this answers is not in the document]
soypat: Only the Identity case which I thought of a special property based test. Though when you say it like that I guess there's not much added by the tests. We can remove them

lapack/testlapack/dgetc2.go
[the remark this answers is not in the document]
vladimir-ch: I mean that both 6 are equally good candidates for the first pivot and there is nothing in the docs that would say which one will be chosen. In other words, the test expectation depends on an implementation detail.

lapack/testlapack/dgetc2.go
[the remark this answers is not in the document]
vladimir-ch: Do you mean that you created the output by running your implementation of Dgetc2 or the Fortran reference? Clearly, you can't generate the expected output by the very implementation that you're testing.

My question about ipiv, jpiv again points to the ambiguity how the pivots are chosen by the algorithm when there are more candidates.

lapack/testlapack/dgetc2.go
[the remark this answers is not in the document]
vladimir-ch: Yes, I think it's ok to remove them.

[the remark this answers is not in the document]
soypat: You've made your point clear as day. Thanks for clearing the reason up!

soypat:
Changes to tests done as per your suggestion. Removed those cases generated by the reference since they may fail depending on implementation details.

vladimir-ch (approved):
LGTM, thank you for your contribution and following our guidance. I will squash the commits before merging and then submit a PR cleaning up a few nits.

I think this is your first contribution; @kortschak , what is the process?

vladimir-ch:
I found it `CONTRIBUTING.md`:

> you will need to be added to the CONTRIBUTORS and AUTHORS file; after your contribution has been accepted you will be asked to open a pull request adding yourself to them.

@soypat , can you submit a new PR where you add yourself to those two files?

kortschak:
Make the commit message "A+C: add Patricio Whittingslow".

PR #1657, opened 2021-06-08
interp: Add NaturalCubic spline interpolator

romanwerpachowski:
It's about time :)


interp/cubic.go
```diff
@@ -297,3 +297,123 @@ func fritschButlandEdgeDerivative(xs, ys, slopes []float64, leftEdge bool) float
 	}
 	return g
 }
+
+// fitWithSecondDerivatives fits a piecewise cubic predictor to (X, Y, d^2Y/dX^2) value
+// triples provided as three slices.
+// It panics if len(xs) < 2, elements of xs are not strictly increasing,
+// len(xs) != len(ys) or len(xs) != len(d2ydx2s).
```
kortschak, line 304: Can this be a bulleted list?

interp/cubic.go
```diff
@@ -297,3 +297,123 @@ func fritschButlandEdgeDerivative(xs, ys, slopes []float64, leftEdge bool) float
...
+// Note: this method does not guarantee on its own the continuity of 1st derivatives.
```
kortschak, line 305: s/Note: this/Note that this/
s/1st/first/

interp/cubic.go
```diff
@@ -297,3 +297,123 @@ func fritschButlandEdgeDerivative(xs, ys, slopes []float64, leftEdge bool) float
...
+		dy := ys[i+1] - ys[i]
+		dm := d2ydx2s[i+1] - d2ydx2s[i]
+		// a_0
+		pc.coeffs.Set(i, 0, ys[i])
+		// a_1.
+		pc.coeffs.Set(i, 1, (dy-(d2ydx2s[i]+dm/3)*dx*dx/2)/dx)
+		// a_2
+		pc.coeffs.Set(i, 2, d2ydx2s[i]/2)
+		// a_3
+		pc.coeffs.Set(i, 3, dm/6/dx)
```
kortschak, line 334:
```
		pc.coeffs.Set(i, 0, ys[i])                             // a_0
		pc.coeffs.Set(i, 1, (dy-(d2ydx2s[i]+dm/3)*dx*dx/2)/dx) // a_1
		pc.coeffs.Set(i, 2, d2ydx2s[i]/2)                      // a_2
		pc.coeffs.Set(i, 3, dm/6/dx)                           // a_3
```

interp/cubic.go
```diff
@@ -297,3 +297,123 @@ func fritschButlandEdgeDerivative(xs, ys, slopes []float64, leftEdge bool) float
...
+	}
+	pc.xs = make([]float64, n)
+	copy(pc.xs, xs)
```
kortschak, line 337:
```
	pc.xs = append(pc.xs[:0], xs...)
```

interp/cubic.go
```diff
@@ -297,3 +297,123 @@ func fritschButlandEdgeDerivative(xs, ys, slopes []float64, leftEdge bool) float
...
+	pc.lastY = ys[m]
+	lastDx := xs[m] - xs[m-1]
+	pc.lastDyDx = pc.coeffs.At(m-1, 1) + 2*pc.coeffs.At(m-1, 2)*lastDx + 3*pc.coeffs.At(m-1, 3)*lastDx*lastDx
+}
+
+// makeCubicSplineSecondDerivativeEquations generates the basic system of linear equations
+// which have to be satisfied by 2nd derivatives to make the 1st derivatives of a cubic spline
```
kortschak, line 344: s/2nd/second/g
s/1st/first/

interp/cubic.go
```diff
@@ -297,3 +297,123 @@ func fritschButlandEdgeDerivative(xs, ys, slopes []float64, leftEdge bool) float
...
+}
+
+// PredictDerivative returns the predicted derivative at x.
+func (nc *NaturalCubic) PredictDerivative(x float64) float64 {
+	return nc.cubic.PredictDerivative(x)
+}
+
+// Fit fits a predictor to (X, Y) value pairs provided as two slices.
+// It panics if len(xs) < 2, elements of xs are not strictly increasing
+// or len(xs) != len(ys). Returns an error if solving the required system
```
kortschak, line 403: s/Returns/It returns/

interp/cubic.go
```diff
@@ -297,3 +297,123 @@ func fritschButlandEdgeDerivative(xs, ys, slopes []float64, leftEdge bool) float
...
+// of linear equations fails.
+func (nc *NaturalCubic) Fit(xs, ys []float64) error {
+	A, b := makeCubicSplineSecondDerivativeEquations(xs, ys)
```
kortschak, line 406: lc/A/

interp/cubic.go
romanwerpachowski: Why? the current line is used in other Fit functions.

interp/cubic.go
romanwerpachowski: I feel strange denoting a matrix by a lower case symbol, but OK :)

interp/cubic.go
kortschak: Yeah, I saw that. I think those should change too, but that can happen in another PR.

The why is that the current approach has a guaranteed allocation for each call to `Fit`, while with the approach in the comment, there is the capacity to reuse a previously allocated slice if the `xs` slice will fit into it.

interp/cubic.go
kortschak: Yeah, it's a domain clash. It gets more comfortable as time passes; that's the nature of Stockholm Syndrome.

interp/cubic.go
romanwerpachowski: I see, thanks. Good to know. Done.

interp/cubic.go
romanwerpachowski: OK.

romanwerpachowski:
PTAL.

interp/cubic.go
kortschak: Also s/1st/first/.

romanwerpachowski: Got rid of them everywhere now, I think.

romanwerpachowski:
PTAL.

kortschak (dismissed):
LGTM, but please wait for @vladimir-ch.

romanwerpachowski:
@vladimir-ch ping :)

interp/cubic.go
```diff
@@ -297,3 +296,122 @@ func fritschButlandEdgeDerivative(xs, ys, slopes []float64, leftEdge bool) float
...
+	return nc.cubic.PredictDerivative(x)
+}
+
+// Fit fits a predictor to (X, Y) value pairs provided as two slices.
+// It panics if len(xs) < 2, elements of xs are not strictly increasing
+// or len(xs) != len(ys). It returns an error if solving the required system
+// of linear equations fails.
+func (nc *NaturalCubic) Fit(xs, ys []float64) error {
+	a, b := makeCubicSplineSecondDerivativeEquations(xs, ys)
+	// Add boundary conditions y''(left) = y''(right) = 0:
```
vladimir-ch, line 405: Could or should this be done in `makeCubicSplineSecondDerivativeEquations`?

interp/cubic.go
```diff
@@ -297,3 +296,122 @@ func fritschButlandEdgeDerivative(xs, ys, slopes []float64, leftEdge bool) float
...
+	n := len(xs)
+	b.SetVec(0, 0)
+	b.SetVec(n-1, 0)
+	a.SetBand(0, 0, 1)
```
vladimir-ch, line 409: What about the off-diagonal element?

vladimir-ch (commented):
I took a quick look, I won't have time to go through the code in detail.

interp/cubic.go
romanwerpachowski: I plan to add other boundary conditions later, so makeCubicSplineSecondDerivativeEquations should contain only the common parts. Such a split feels natural to me.

interp/cubic.go
romanwerpachowski: Already set in `makeCubicSplineSecondDerivativeEquations`.

interp/cubic_test.go
```diff
@@ -671,3 +671,134 @@ func TestFritschButlandErrors(t *testing.T) {
...
+		t.Errorf("Mismatch in b size: got %v, want %d", b.Len(), n)
+	}
+	r, c := A.Dims()
+	if r != n || c != n {
+		t.Errorf("Mismatch in A size: got %d x %d, want %d x %d", r, c, n, n)
+	}
+	expectedB := mat.NewVecDense(3, []float64{0, 3, 0})
+	var diffB mat.VecDense
+	diffB.SubVec(b, expectedB)
+	if math.Abs(diffB.Norm(math.Inf(1))) > tol {
```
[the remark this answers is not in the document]
romanwerpachowski: Good catch!

interp/cubic.go
```diff
@@ -297,3 +296,122 @@ func fritschButlandEdgeDerivative(xs, ys, slopes []float64, leftEdge bool) float
...
+				du[i] = dx / 6
+			}
+			if i < m-1 {
+				b[i+1] -= slope
+				d[i+1] += dx / 3
+				dl[i] = dx / 6
+			}
+		}
+	}
+	return mat.NewTridiag(n, dl, d, du), mat.NewVecDense(n, b)
```
vladimir-ch, line 378: There should be a ticket to use `SymTridiag` when it exists and has `SolveVecTo`. You could already use `SymBandDense` and `BandCholesky`.

romanwerpachowski: I think the matrix isn't symmetric in this case.

vladimir-ch: Ok, it's not symmetric because of the boundary conditions. Currently there is only `y''(left) = y''(right) = 0` and in this case the matrix can be made symmetric by eliminating the first and last unknown. I don't know what other BCs you plan and whether it's worth the effort.

romanwerpachowski: I plan to add those listed in https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.CubicSpline.html and custom ones (`a*y'(left) + b*y''(left) == c && d*y'(right) + e*y''(right) == f`). Started with natural just to get the overall design right without having to redo a lot of work.

interp/cubic.go
```diff
@@ -297,3 +296,122 @@ func fritschButlandEdgeDerivative(xs, ys, slopes []float64, leftEdge bool) float
...
+// triples provided as three slices.
+// It panics if any of these is true:
+// - len(xs) < 2,
+// - elements of xs are not strictly increasing,
+// - len(xs) != len(ys),
+// - len(xs) != len(d2ydx2s).
+// Note that this method does not guarantee on its own the continuity of first derivatives.
+func (pc *PiecewiseCubic) fitWithSecondDerivatives(xs, ys, d2ydx2s []float64) {
+	n := len(xs)
+	if len(ys) != n {
```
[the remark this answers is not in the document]
romanwerpachowski: Will do.

interp/cubic.go
vladimir-ch: Ok.

romanwerpachowski:
PTAL.

vladimir-ch (approved):

PR #1659, opened 2021-06-10
stat/disuv: make helpers and disambiguate case indexes

kortschak:
Please take a look.


romanwerpachowski (approved):
Why isn't "panics" also marked as t.Helper()?

kortschak:
> Why isn't "panics" also marked as t.Helper()?

Nothing ever fails there.

romanwerpachowski:
OK, thanks.

PR #1662, opened 2021-06-11
interp: add ClampedCubic interpolator

romanwerpachowski:
Please take a look.


interp/cubic_test.go
```diff
@@ -800,5 +800,37 @@ func TestNaturalCubicFit(t *testing.T) {
...
+}
 
+func TestClampedCubicFit(t *testing.T) {
+	t.Parallel()
+	xs := []float64{-1, 0, 2, 3.5}
+	ys := []float64{2, 0, 2, 1.5}
+	var cc ClampedCubic
+	err := cc.Fit(xs, ys)
+	if err != nil {
+		t.Errorf("Error when fitting NaturalCubic: %v", err)
```
vladimir-ch, line 812: ClampedCubic?

romanwerpachowski: Thanks, fixed.

romanwerpachowski:
PTAL.

soypat:
Where do the `want` values come from? Could we not make property based tests?

romanwerpachowski:
They come from `scipy.interpolate.CubicSpline`, which I treat as a trusted reference implementation until proven otherwise :) I could test properties of `ClampedCubic` directly, but it's more work and the current test is sufficient IMHO.

romanwerpachowski:
*ping*

vladimir-ch (approved):

PR #1665, opened 2021-06-13
lapack: add Dlag2 subroutine

soypat:
ARE issue #1651

### DIFF
--- a/lapack/gonum/dlag2.go
+++ b/lapack/gonum/dlag2.go
@@ -6,7 +6,7 @@ package gonum
 
 import "math"
 
-// DLAG2 computes the eigenvalues of a 2 x 2 generalized eigenvalue
+// DLAG2 computes the eigenvalues of a 2×2 generalized eigenvalue
 // problem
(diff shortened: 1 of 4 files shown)


soypat:
_ping_

vladimir-ch:
Let's first get Dorgr2 done, only little left to do there.

kortschak:
Given the limited bandwidth available for review, the optimal approach to ordering work in a complex addition like this is a strict dependency ordering, only moving on to dependent packages once dependencies have been accepted and merged; this reduces the overall cognitive load on reviewers since it ceases to be necessary to hold more than one package in mind at a time. Having too many things in the air will hamper progress rather than help it.

soypat:
> Having too many things in the air will hamper progress rather than help it.

This has started to become painfully apparent. 

> the optimal approach to ordering work in a complex addition like this is a strict dependency ordering

Sounds good. A slow but methodical and precise approach.

soypat:
closed due to branch mismanagement.

PR #1666, opened 2021-06-13
lapack: add Dlagv2 implementation.

soypat:
ARE issue #1651 

This branch is still under work. It would require Dlag2 (#1665) to be merged into master to implement tests.

I've added a private (lower cased) method called `dlagv2aux` to simplify `Dlagv2`'s implementation and make it more readable. If this is a non-permissible act it can be implemented as in the reference but do note readability will suffer greatly.

Tests pending, as mentioned above.


soypat:
Bad branch

PR #1670, opened 2021-06-14
lapack/testlapack: extend test for Dlasy2 to cover rescaling

vladimir-ch:
Please take a look.

@soypat

### DIFF
--- a/lapack/testlapack/dlasy2.go
+++ b/lapack/testlapack/dlasy2.go
@@ -6,13 +6,13 @@ package testlapack
 
 import (
 	"fmt"
-	"math"
 	"testing"
 
 	"golang.org/x/exp/rand"
 
 	"gonum.org/v1/gonum/blas"
 	"gonum.org/v1/gonum/blas/blas64"
+	"gonum.org/v1/gonum/lapack"
 )
 
 type Dlasy2er interface {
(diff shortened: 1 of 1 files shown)


lapack/testlapack/dlasy2.go
```diff
@@ -38,68 +42,83 @@ func Dlasy2Test(t *testing.T, impl Dlasy2er) {
...
+		blas64.Gemm(blas.NoTrans, blas.NoTrans, float64(isgn), x, tr, 1, diff)
 	}
-	// Compare them.
 	for i := 0; i < n1; i++ {
 		for j := 0; j < n2; j++ {
-			diff := lhsWant.Data[i*lhsWant.Stride+j] - b.Data[i*b.Stride+j]
-			if math.Abs(diff) > tol && ok {
-				t.Errorf("%v: unexpected result, diff[%v,%v]=%v", prefix, i, j, diff)
-			}
+			diff.Data[i*diff.Stride+j] -= scale * b.Data[i*b.Stride+j]
```
soypat, line 116: Should `scale * b.Data[i*b.Stride+j]` be inside a `float64` cast to enable testing on ARM?

soypat:
Apart from that `float64` for ARM cast, LGTM

lapack/testlapack/dlasy2.go
kortschak: I think this is OK here until we see that it's not. It passes on travis and locally on my builder.

kortschak (approved):

PR #1672, opened 2021-06-14
spatial/vptree: fix vantage point selection

### LINKED ISSUE #1671: vptree: construction may panic if a pair of points has zero distance

Perform k-NN on a machine learning dataset that contains some duplicate samples.

### What did you do?

Invoke `vptree.New()` with an `effort` and seeded `src` that resulted in `vptree.selectVantage()` reaching a recursion case where the remaining point choices `s` are identical.

Because [in such a recursion case] all point choices have zero distance from each other, the choices loop never finds a variance greater than 0, and so it never selects a `best, bestVar` pair. This causes the `best == nil` panic.

Minimal example:
https://play.golang.org/p/DCTJsybTtQJ

(This assumes your version of `golang.org/x/exp/rand` produces the same random sequences as mine.)

Expected VP tree creation to succeed.

### What actually happened?

`panic: vptree: could not find vantage point`

go version go1.16 darwin/amd64
Gonum 0.9.1

Yes

### Proposal

This seems to be fixed by changing [this](https://github.com/gonum/gonum/blob/85aaeceabc3dd1dbef08a30c0992482c4a68dd8f/spatial/vptree/vptree.go#L145)...
```go
if variance > bestVar {
	best, bestVar = p, variance
}
```
...to this...
```go
if bestVar == 0 || variance > bestVar {
	best, bestVar = p, variance
}
```

kortschak:
Please take a look.

Fixes #1671.

/cc @jjr90

### DIFF
--- a/spatial/vptree/vptree.go
+++ b/spatial/vptree/vptree.go
@@ -155,7 +155,7 @@ func (b *builder) selectVantage(s []Comparable, effort int) Comparable {
 
 func (b *builder) random(n int, s []Comparable) []Comparable {
 	if n >= len(s) {
-		return s
+		n = len(s)
 	}
 	b.shuf(len(s), func(i, j int) { s[i], s[j] = s[j], s[i] })
 	return s[:n]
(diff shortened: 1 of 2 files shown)


jjr90 (approved):
Thank you, looks nice.

vladimir-ch (approved):

PR #1677, opened 2021-06-19
add tait-bryan angles rotation matrix

soypat:
@sbinet Please take a look at the proposal.

 These are not Euler angles but Tait-Bryan angles which are easier to reason about since the rotations happen on three separate axis. These equations commonly describe rigid body motion, as can be seen in Hubert Hahn's comprehensive book on the matter *Rigid Body Dynamics*.

There is room for other Euler angle implementations which could use the `RotationMatrix` type.


kortschak:
It's not clear to me why this is needed (the need it what would form the basis for a proposal rather than sending some code). Also note that while it doesn't have Euler angles it still suffers from gimbal lock which was the reason to avoid Euler angles.

soypat:
Closing due to different approach

Issue #1678, opened 2021-06-19
gonum/spatial/r3: Add Euler/Tait-Bryan angle rotations


soypat:
### Background
Analysis of rigid body mechanics in 3D space are usually dealt with in Euler angles in introductory literature to simplify the algebra. Such is the case in _Rigid Body Dynamics_ by Prof. Hubert Hahn whose book deals with the XYZ permutation of Tait-Bryan angles. 

### Proposal
Add an implementation of the XYZ Tait-Bryan rotation.

### Potential impact of proposal
It is known this representation has the problem of gimbal lock. Maybe users could have the option to convert the Tait-Bryan rotation into a quaternion rotation to avoid this problem.

kortschak:
A similar proposal has been made in the past https://github.com/gonum/gonum/issues/1552 that suffers from the same problems which I wrote about [here](https://github.com/gonum/gonum/pull/1516#discussion_r532911007).

This proposal says that the algebra is easier. Can this be demonstrated? I would have thought the converse since the rotation operation with a quat is simple and operations on quats are simple, including Slerp.

If the addition of this feature for learning purposes (being able to understand the literature), then I would much rather have an example test that shows interconversion between angle-based forms and quats and which demonstrates the outcome of attempting this with a gimbal locked system (note here that converting a description of a rotation system with gimbal lock into a quat representation doesn't solve the gimbal lock).

kortschak:
A fleshed out (more detail on gimbal lock) version of this would be pretty reasonable.

```
package r3_test

import (
	"fmt"
	"math"

	"gonum.org/v1/gonum/spatial/r3"
)

// euler returns an r3.Rotation that corresponds to the Euler
// angles alpha, beta and gamma which are rotations around the x,
// y and z axes respectively.
func euler(alpha, beta, gamma float64) r3.Rotation {
	s1, c1 := math.Sincos(alpha / 2)
	s2, c2 := math.Sincos(beta / 2)
	s3, c3 := math.Sincos(gamma / 2)
	return r3.Rotation{
		Real: c2*c1*c3 - s2*s1*s3,
		Imag: s2*s1*c3 + c2*c1*s3,
		Jmag: s2*c1*c3 + c2*s1*s3,
		Kmag: c2*s1*c3 - s2*c1*s3,
	}
}

func ExampleRotation_gimbalLock() {
	orig := r3.Vec{-1, 0, 0}

[...]

PR #1681, opened 2021-06-20
spatial/r3: add example for why not to use Euler angles

kortschak:
Please take a look.

Updates #1678.

### DIFF
--- a/spatial/r3/euler_example_test.go
+++ b/spatial/r3/euler_example_test.go
@@ -1,0 +1,77 @@
+// Copyright ©2021 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package r3_test
+
+import (
+	"fmt"
+	"math"
+
+	"gonum.org/v1/gonum/num/quat"
+	"gonum.org/v1/gonum/spatial/r3"
+)
+
+// euler returns an r3.Rotation that corresponds to the Euler
+// angles alpha, beta and gamma which are rotations around the x,
+// y and z axes respectively. The order of rotations is x, y, z;
+// there are many conventions for this ordering.
+func euler(alpha, beta, gamma float64) r3.Rotation {
+	// Note that this function can be algebraically simplified
+	// to reduce floating point operations, but is left in this
+	// form for clarity.
+	var rot1, rot2, rot3 quat.Number
+	rot1.Imag, rot1.Real = math.Sincos(alpha / 2) // x-axis rotation
+	rot2.Jmag, rot2.Real = math.Sincos(beta / 2)  // y-axis rotation
+	rot3.Kmag, rot3.Real = math.Sincos(gamma / 2) // z-axis rotation
+
+	return r3.Rotation(quat.Mul(rot3, quat.Mul(rot2, rot1))) // order of rotations
+}
+
+func ExampleRotation_eulerAngles() {
+	// It is possible to interconvert between the quaternion representation
+	// of a rotation and Euler angles, but this leads to problems.
+	//
+	// The first of these is that there are a variety of conventions for
+	// application of the rotations.
+	//
+	// The more serious consequence of using Euler angles is that it is
+	// possible to put the rotation system into a singularity which results
+	// in loss of degrees of freedom and so causes gimbal lock. This happens
+	// when the second axis to be rotated around is rotated to 𝝿/2.
+	//
+	// See https://en.wikipedia.org/wiki/Euler_angles for more details.
+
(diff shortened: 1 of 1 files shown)


spatial/r3/euler_example_test.go
```diff
@@ -0,0 +1,74 @@
...
+	rot1.Imag, rot1.Real = math.Sincos(alpha / 2) // x-axis rotation
+	rot2.Jmag, rot2.Real = math.Sincos(beta / 2)  // y-axis rotation
+	rot3.Kmag, rot3.Real = math.Sincos(gamma / 2) // z-axis rotation
+
+	return r3.Rotation(quat.Mul(rot3, quat.Mul(rot2, rot1))) // order of rotations
+}
+
+func ExampleRotation_eulerAngles() {
+	// It is possible to interconvert between the quaternion representation
+	// of a rotation and Euler angles, but this leads to a variety of problem.
```
vladimir-ch, line 33: of problems.

spatial/r3/euler_example_test.go
```diff
@@ -0,0 +1,74 @@
...
+	// The first of these is that there are a variety of conventions for
```
vladimir-ch, line 34: _is_ or _are_ (a variety of conventions)?

spatial/r3/euler_example_test.go
kortschak: Fixed

spatial/r3/euler_example_test.go
kortschak: This is correct (at least for my dialect). I did a quick check and it seems prevalent. I think it would be singular if the object were uncountable, though maybe in American English it's singular.

soypat:
It could be clarified this is an **extrinsic** euler angle implementations which are is not the implementation I had proposed with my recent PR. Maybe we could clarify the difference between the two and have one implementation of each. I've never really used extrinsic rotations FTR.

kortschak:
They are equivalent, with the difference being only really visible when you use a matrix for the rotations.

Here is the change in the code that gives you the Tait-Bryan rotation:

```
func taitBryan(alpha, beta, gamma float64) r3.Rotation {
	var rot1, rot2, rot3 quat.Number
	rot1.Imag, rot1.Real = math.Sincos(alpha / 2) // x-axis rotation
	rot2.Jmag, rot2.Real = math.Sincos(beta / 2)  // y-axis rotation
	rot3.Kmag, rot3.Real = math.Sincos(gamma / 2) // z-axis rotation
	return r3.Rotation(quat.Inv(quat.Mul(rot1, quat.Mul(rot2, rot3)))) // order of rotations
}
```

soypat:
I don't think you understand what I mean by extrinsic. In all fairness, I'm taking the nomenclature from [wikipedia](https://en.wikipedia.org/wiki/Euler_angles) since pretty much every book I've read on the matter has conflicting names for conventions used to represent orientation configuration.

### Euler vs. Tait-Bryan
According to wikipedia, The difference between euler angles and Tait-bryan angles is the chosen axis over which the rotations are performed.

| Euler | Tait-Bryan |
|--------|----------------|
| zxz   | xyz |
| xyx  | yzx  |
| yzy  | zxy  |
| zyz | xzy  |
| xzx | zyx |
| yxy | yxz |

The table should be enough to notice the pattern, euler angles rely on a permutation of two axis, tait-bryan on three.

## Extrinsic vs. Intrinsic
From wikipedia
> The three elemental rotations may be extrinsic (rotations about the axes xyz of the original coordinate system, which is assumed to remain motionless), or intrinsic (rotations about the axes of the rotating coordinate system XYZ, solidary with the moving body, which changes its orientation after each elemental rotation).

Mathematically, intrinsic rotations can be represented as a product of three rotation matrices with on vector. For the case of Tait-Bryan XYZ rotation, a vector in an "inertial" or fixed frame of reference `N` can be rotated to a body frame `B` with the following operations: 
> `v^{B} = A_{z} * A_{y} * A_{x} * v^{N}`

Extrinsic rotations may be performed with both Tait-Bryan and Euler angles. Same goes for Intrinsic rotations.

kortschak:
I think this demonstrated really nicely why we shouldn't touch this any more than I have with this PR.

kortschak:
To clarify, "since pretty much every book I've read on the matter has conflicting names for conventions used to represent orientation configuration" says that there is too much complexity for us to cover. The example I have here shows how to generate any given rotation approach since they are all equivalent and representable by a quaternion rotation.

I think that there is probably merit in adding a mat→quat and quat→mat pair of functions so that rotations of multiple points are easier via matrix multiplication, but that would be a separate PR.

vladimir-ch (approved):

PR #1682, opened 2021-06-21
spatial/r3: add method to obtain a rotation matrix

kortschak:
I have not added the matrix→quat function as this requires more thought in terms of what we should accept and is arguably less valuable; the quat→mat operation means that large sets of points can be operated on with a single matrix multiply while it's not entirely clear what you get from being able to convert a mat to a quat, and the code is more complex.

Please take a look.

### DIFF
--- a/spatial/r3/vector.go
+++ b/spatial/r3/vector.go
@@ -7,6 +7,7 @@ package r3
 import (
 	"math"
 
+	"gonum.org/v1/gonum/mat"
 	"gonum.org/v1/gonum/num/quat"
 )
 
@@ -96,7 +97,6 @@ type Box struct {
 //  - create rotations from Euler angles (NewRotationFromEuler?)
 //  - create rotations from rotation matrices (NewRotationFromMatrix?)
 //  - return the equivalent Euler angles from a Rotation
-//  - return the equivalent rotation matrix from a Rotation
 //
 // Euler angles have issues (see [1] for a discussion).
 // We should think carefully before adding them in.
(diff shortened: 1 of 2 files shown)


sbinet (approved):

spatial/r3/vector.go
```diff
@@ -139,3 +139,25 @@ func (r Rotation) isIdentity() bool {
 func raise(p Vec) quat.Number {
 	return quat.Number{Imag: p.X, Jmag: p.Y, Kmag: p.Z}
 }
+
+// Matrix returns a 3×3 rotation matrix corresponding to the receiver. It
+// may be used to perform rotations on a 3-vector or to apply the rotation
+// to a 3×n matrix of column vectors. If the receiver is not a unit
+// quaternion, the returned matrix will not be a pure rotation.
+func (r Rotation) Matrix() mat.Matrix {
+	re, im, jm, km := r.Real, r.Imag, r.Jmag, r.Kmag
```
sbinet, line 148: Any reason why returning a 'mat.Matrix' instead of just returning the concrete type?

kortschak: The main reason was that we can change out the implementation if we need; `mat.Dense` is probably more heavy-weight than needed here (a `*[9]float64` that satisfies `mat.Matrix` and `mat.RawMatrixer` would suit for example).

kortschak: Like so
```
func (r Rotation) Matrix() mat.Matrix {
	re, im, jm, km := r.Real, r.Imag, r.Jmag, r.Kmag
	im2 := im * im
	jm2 := jm * jm
	km2 := km * km
	rim := re * im
	rjm := re * jm
	rkm := re * km
	ijm := im * jm
	jkm := jm * km
	kim := km * im
	return &matrix{
		1 - 2*(jm2+km2), 2 * (ijm - rkm), 2 * (kim + rjm),
		2 * (ijm + rkm), 1 - 2*(im2+km2), 2 * (jkm - rim),
		2 * (kim - rjm), 2 * (jkm + rim), 1 - 2*(im2+jm2),
	}
}

type matrix [9]float64

func (m *matrix) At(i, j int) float64 {
	if uint(i) >= 3 {
		panic(mat.ErrRowAccess)
	}
	if uint(j) >= 3 {
		panic(mat.ErrColAccess)
	}
	return m[i*3+j]
}
func (m *matrix) Dims() (r, c int) { return 3, 3 }
func (m *matrix) T() mat.Matrix    { return mat.Transpose{Matrix: m} }
func (m *matrix) RawMatrix() blas64.General {
	return blas64.General{Rows: 3, Cols: 3, Data: m[:], Stride: 3}
}
```

Do you think this is worth doing now, or leave?

sbinet: this could be worth doing. (perhaps in a separate PR, though)

PR #1684, opened 2021-06-21
gonum/blas: Sparse Blas level 1 routines draft

soypat:
Please take a look.

The purpose of this PR is to start looking at cold-hard code of what a sparse implementation would look like. There is a significant downside to how these functions work: there's no way to know if the length of y is too short. These routines assume that `index[i]` will not overstep `y`'s length. These checks could be added in a higher level function which takes a struct `blas.General`-like argument (i.e: `blas.Sparse`?)


soypat:
Closing and reopening without the commits which have nothing related to sparse proposition. I tried squashing them but I'm still a newbie with git and botched it.

PR #1687, opened 2021-06-22
graph/{multi,simple}: improve undirected graph deduplication logic

### LINKED ISSUE #1686: graph/multi: inconsistent/unexpected results from Edges() for Directed and Undirected

Count the edges of an undirected graph.

### What did you do?

Add nodes and lines to an undirected graph, call Edges(), and then call Len() on the result to get the total number of edges in the graph.
The example below shows the results of the same process being applied to Directed and Undirected graphs.
There should be 3 nodes and 2 edges.

```go
func add(gr graph.MultigraphBuilder) {
	n0 := gr.NewNode()
	gr.AddNode(n0)
	n1 := gr.NewNode()
	gr.AddNode(n1)
	n2 := gr.NewNode()
	gr.AddNode(n2)
	gr.SetLine(gr.NewLine(n0, n1))
	gr.SetLine(gr.NewLine(n1, n2))
}

func main() {
	fmt.Println("Undirected:")
	ug := multi.NewUndirectedGraph()
	add(ug)
	fmt.Println(ug.Nodes().Len(), ug.Edges().Len())

	fmt.Println("Directed:")
	dg := multi.NewDirectedGraph()
	add(dg)
	fmt.Println(dg.Nodes().Len(), dg.Edges().Len())
}
```

Output:
```
Undirected:
3 1
Directed:
3 2
```

I expected that each line would be counted as an edge.
I also expected that the number of edges would be the same for both graphs.

### What actually happened?

The Directed graph reports the expected number, but the Undirected graph reports a different and unexpected number.

Go version:
```
go version go1.15.7 windows/amd64
```

From go.mod:
```
require gonum.org/v1/gonum v0.9.2
```

I was able to repoduce using the lastest commit hash.
```
require gonum.org/v1/gonum v0.9.1-0.20210621072618-2600994ee4df
```

kortschak:
Please take a look.

Fixes #1686.

/cc @msiner


msiner:
This change provides the output I was expecting from Edges() and appears to solve #1686. When tested on a large undirected graph (https://snap.stanford.edu/data/as-733.html), I get the correct number of edges. De-duplication of lines with identical, but reversed, to and from node IDs in the undirected case also appears to be working as expected. Thank you.

kortschak:
Thanks for confirming. Working on this highlighted some overdue optimisations, so it was a definite win. We just need to wait for a review, but you can use a replace to get this commit now.

vladimir-ch (approved):

PR #1688, opened 2021-06-22
lapack/gonum: add Dgesv

vladimir-ch:
More for completeness than utility because Dgesv doesn't have a trans parameter to solve `A^T*X = B`.

Please take a look.

### DIFF
--- a/lapack/testlapack/dgesv.go
+++ b/lapack/testlapack/dgesv.go
@@ -1,0 +1,164 @@
+// Copyright ©2021 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package testlapack
+
+import (
+	"fmt"
+	"math"
+	"testing"
+
+	"golang.org/x/exp/rand"
+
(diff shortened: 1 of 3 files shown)


lapack/testlapack/dgesv.go
```diff
@@ -0,0 +1,158 @@
...
+// license that can be found in the LICENSE file.
+
+package testlapack
+
+import (
+	"fmt"
+	"math"
+	"testing"
+
+	"golang.org/x/exp/rand"
```
kortschak, line 12: Blank between.

kortschak (commented):
A nit.

lapack/testlapack/dgesv.go
vladimir-ch: Thanks, done. I've also added a check of `ok` in the test.

kortschak (approved):

vladimir-ch:
I took a look at Dgesvx but that's a much more complex call, testing it would be a hell and there are unimplemented dependencies, so I won't go there.

PR #1695, opened 2021-07-13
spatial/curve: add package to generate Hilbert curves

firelizzard18:
This PR implements 2-, 3-, and 4-dimensional Hilbert curves, based on the code sample in the Wikipedia article - the package documentation includes a permalink to the article. Initially, I copy-pasted the functions and translated them to Go, for the 2D curve. I extrapolated that to 3 and 4 dimensions and gradually rearranged the code. These changes make it more obvious that the calculations are largely bit manipulation, and make it more obvious how 2D can be extrapolated to higher dimensions.

# Performance

I included a benchmark for the 2D Hilbert curve. On my machine (i9-10850K), `Hilbert2D.Curve` takes 20-40ns and `Hilbert2D.Space2D` takes 5-20ns, for order 1-10. The execution time of `Curve` can be reduced to nearly that of `Space2D` by eliminating the allocation, `v = dup(v)`. This copy is done because the rotation operation clobbers V - clobbering an input array is generally not expected, hence the copy. Since 2D always uses 2 values, this could be eliminated in a variety of ways, most of which would degrade readability. Go 1.17 will allow `u := *(*[2]int)(v)` - the current equivalent, `u := *(*[2]int)(unsafe.Pointer(&v[0]))`, eliminates the allocation and improves runtime to 5-30ns.

The performance of 3D and 4D are an order of magnitude worse (~100-500ns). This may be due to allocations in the `rot` functions.

# Implementation

Initially, `Hilbert4D` was `Hilbert` and I intended it to be an n-dimensional Hilbert curve, which is why the bit operations are in loops. But I couldn't figure out how to extrapolate the 'rotation' operations.

The 1st order curves are constructed with XOR. Where N is the dimension, the curve coordinate D for a spatial point V is calculated by:

```golang
var d int
for i := N - 1; i >= 0; i-- {
	v := v[i] >> n & 1
	d = d<<1 | (d^v)&1
}
```

Conversely, the spatial point V for a curve coordinate D is calculated by:

```golang
v := make([]int, N)
for i, d := 0, d; i < N; i++ {
	v[i] += (d ^ d>>1) & 1 << n
	d >>= 1
}
```

Higher order curves are constructed by iterating over `[0, K)` where K is the order, shifting D appropriately, and 'rotating' the spatial coordinates (V). What 'rotation' operation is used depends on D and N.

The rotations for 2D were taken from Wikipedia. I arrived at the 3D rotations by rendering the 2nd order curve and deducing what was necessary to make the sub curves connect. I arrived at the 4D rotations by printing out the 2nd order curve as a 4x4x4x4 matrix (see the example) and deducing what was necessary to make the sub curves connect. I do not see an obvious way to extrapolate this to n-dimensions.

If the rotation for a given D and N is not its own inverse, the order used for Curve must be reversed for Space. Swap, flip, and invert operations are their own inverse, but two such operations (on different I and J) in sequence are not. The initial operations I deduced for 3D were all single swap, flip, or invert operations, but that did not produce the same curve as other sources, so I changed the rotations until the curve matched.


spatial/curve/curve.go
```diff
@@ -0,0 +1,23 @@
+// Package curve defines space filling curves. A space filling curve is a curve
```
kortschak, line 1: Each file will need to have a copyright header with 2021 as the copyright data. You can just copy/paste from another file in the repo.

spatial/curve/curve.go
```diff
@@ -0,0 +1,23 @@
...
+// who's range contains the entirety of a finite k-dimensional space. Space
```
kortschak, line 2: s/who's/whose/

spatial/curve/curve.go
```diff
@@ -0,0 +1,23 @@
...
+// filling curves can be used to map between a linear space and a 2D, 3D, or 4D
+// space.
+package curve
+
+// Point is a point on a curve.
+type Point uint64
```
kortschak, line 8: I don't think we need a type for this.

spatial/curve/curve.go
```diff
@@ -0,0 +1,23 @@
...
+
+// SpaceFilling is a space filling curve.
+type SpaceFilling interface {
+	// Size returns the spatial size of the curve. If the X, Y, and Z
+	// coordinates of spatial points mapped by a 3 dimensional curve all lie
+	// within [0, x), [0, y), and [0, z), respectively, the size of that curve
+	// is (x, y, z).
+	Size() []int
```
kortschak, line 16: Do you anticipate that there will be non-equilateral spaces being generated (this is not possible with these curves, so if you are it would depend on alternative space-filling curves). If not, this can just return an `int`.

spatial/curve/curve.go
```diff
@@ -0,0 +1,23 @@
...
+// who's range contains the entirety of a finite k-dimensional space. Space
+// filling curves can be used to map between a linear space and a 2D, 3D, or 4D
+// space.
+package curve
+
+// Point is a point on a curve.
+type Point uint64
+
+// SpaceFilling is a space filling curve.
+type SpaceFilling interface {
```
kortschak, line 11: I'm not completely convinced that we need to define an interface publicly yet.

spatial/curve/hilbert.go
```diff
@@ -0,0 +1,254 @@
...
+//
+//     func map1stOrder(k int, v []int) (d int) {
+//         for i := k - 1; i >= 0; i-- {
+//             d = d<<1 | (d^v[i])&1
+//         }
+//         return d
+//     }
+//
+// In a 2-space with the origin at the bottom left, this results in a ⊐ shape,
+// wound counter clockwise.
```
kortschak, line 22: Please place this in the same comment that documents the package (currently in curve.go, but better in doc.go).

spatial/curve/hilbert.go
```diff
@@ -0,0 +1,254 @@
...
+package curve
+
+// NewHilbert returns a 2-, 3-, or 4-dimensional Hilbert curve. NewHilbert will
+// panic if dimension is not in the range [2, 4] or if order is not in the range
+// [1, ∞).
+//
+// The runtime of Space and Curve scales as O(k∙n) where k is the dimension and
+// n is the order. The length of the curve is 2^(k∙n).
+func NewHilbert(order, dimension int) SpaceFilling {
```
kortschak, line 31: I don't think we need a constructor yet.

spatial/curve/hilbert.go
```diff
@@ -0,0 +1,254 @@
...
+type coordInvert struct{ I, J int } // invert I and J
+type coordSwap struct{ I, J int }   // swap I and J
+type coordFlip struct{ I, J int }   // swap and invert I and J
+
+func (c coordInvert) do(n int, v []int) { v[c.I], v[c.J] = v[c.I]^(1<<n-1), v[c.J]^(1<<n-1) }
+func (c coordSwap) do(n int, v []int)   { v[c.I], v[c.J] = v[c.J], v[c.I] }
+func (c coordFlip) do(n int, v []int)   { v[c.I], v[c.J] = v[c.J]^(1<<n-1), v[c.I]^(1<<n-1) }
+
+// compose multiple coordinate modifications
+type multiCoord []interface{ do(int, []int) }
```
kortschak, line 67: Please place dependency types and functions below their users.

spatial/curve/hilbert.go
```diff
@@ -0,0 +1,254 @@
...
+}
+
+// V may get mangled by Curve, so copy it
+func dup(v []int) []int {
+	u := make([]int, len(v))
+	copy(u, v)
+	return u
+}
+
+type coordInvert struct{ I, J int } // invert I and J
```
kortschak, line 58: s/coordInvert/invert/g and lc/[ij]/g

Similar changes for the next two types.

spatial/curve/hilbert.go
```diff
@@ -0,0 +1,254 @@
...
+type coordSwap struct{ I, J int }   // swap I and J
+type coordFlip struct{ I, J int }   // swap and invert I and J
+
+func (c coordInvert) do(n int, v []int) { v[c.I], v[c.J] = v[c.I]^(1<<n-1), v[c.J]^(1<<n-1) }
+func (c coordSwap) do(n int, v []int)   { v[c.I], v[c.J] = v[c.J], v[c.I] }
+func (c coordFlip) do(n int, v []int)   { v[c.I], v[c.J] = v[c.J]^(1<<n-1), v[c.I]^(1<<n-1) }
```
kortschak, line 64: Please group the methods with their types.

spatial/curve/hilbert.go
kortschak, line 67: I'd call something simpler like `ops` and define `type op interface { do(int, []int) }`.

spatial/curve/hilbert.go
```diff
@@ -0,0 +1,254 @@
...
+	case 0:
+		coordSwap{0, 1}.do(n, v)
+	case 3:
+		coordFlip{0, 1}.do(n, v)
+	}
+}
+
+// Curve returns the curve coordinate of V.
+func (h Hilbert2D) Curve(v ...int) Point {
+	v = dup(v)
```
kortschak, line 98: I'd rather not have this allocation here, but rather a warning in the docs that the parameter is mutated. Same for the others. So `dup` can be dropped.

spatial/curve/hilbert.go
```diff
@@ -0,0 +1,254 @@
...
+		ry := (v[1] >> n) & 1
+		rd := ry<<1 | (ry ^ rx)
+		d += Point(rd) << (2 * n)
+		h.rot(h.Order, v, rd)
+	}
+	return d
+}
+
+// Space2D returns the spatial coordinates of D.
+func (h Hilbert2D) Space2D(d Point) [2]int {
```
kortschak, line 111: Why do we have the array and slice versions of each of these?

spatial/curve/curve.go
```diff
@@ -0,0 +1,23 @@
...
+	// Size returns the spatial size of the curve. If the X, Y, and Z
+	// coordinates of spatial points mapped by a 3 dimensional curve all lie
+	// within [0, x), [0, y), and [0, z), respectively, the size of that curve
+	// is (x, y, z).
+	Size() []int
+
+	// Curve returns the curve coordinate of V.
+	Curve(v ...int) Point
```
kortschak, line 19: I think `Curve([]int)` rather than `Curve(...int)`.

spatial/curve/hilbert_test.go
```diff
@@ -0,0 +1,346 @@
...
+
+import (
+	"fmt"
+	"reflect"
+	"strings"
+	"testing"
+
+	"golang.org/x/exp/rand"
+)
+
```
kortschak, line 11: Please place the `func Test...` first, followed by the `func Bench...` with the callees after their last dependent.

spatial/curve/hilbert_test.go
```diff
@@ -0,0 +1,346 @@
...
+type Dim int
```
kortschak, line 12: I don't think this type is more helpful than a function with an `int` parameter, or preferably a slice of test cases to iterate over.

spatial/curve/hilbert.go
```diff
@@ -0,0 +1,254 @@
...
+
+// Hilbert4D is a 4-dimensional Hilbert curve.
+type Hilbert4D struct{ Order int }
+
+// Size returns {2ⁿ, 2ⁿ, 2ⁿ, 2ⁿ} where n is the order.
+func (h Hilbert4D) Size() []int { return []int{1 << h.Order, 1 << h.Order, 1 << h.Order, 1 << h.Order} }
+
+func (h Hilbert4D) rot(reverse bool, n int, v []int, d int) {
+	switch d {
+	case 0x0:
```
kortschak, line 200: Can you make the cases decimal, please.

kortschak (commented):
I've done a first pass.

It would be good if you could put the algorithm that generates the n-dim rot function in a comment.

spatial/curve/curve.go
firelizzard18: Given that non-equilateral space filling curves exist, I would rather use an interface that accommodates that. Personally I don't plan to implement such a curve, but IMO it is worth it to allow for that.

spatial/curve/hilbert.go
firelizzard18: `Space(d Point) []int` allows all of the curves to share an interface, but it requires allocating a slice. `Space[N]D(d Point) [N]int` does not require an allocation, but it has a different signature for each type. So if you're using `Hilbert3D` directly, you can use the allocation-free version. But if you need to support different dimensionalities, you can use the one that returns a slice.

spatial/curve/hilbert_test.go
firelizzard18: Part of the point of `testCurve` is to ensure that the curve actually describes a consistent curve where any two neighboring points are actually neighbors. To test this, I scan the entire space/curve. IMO a recursive function is the simplest way to do this. I replaced `Dim` with a function.

firelizzard18:
> It would be good if you could put the algorithm that generates the n-dim rot function in a comment.

@kortschak There is no algorithm. For 2D, I transliterated Wikipedia's function. For 3D, I rendered the curve with OpenGL and tried things until the first half looked right, then mirrored that for the second half. Then I noticed that Mathematica's rendering shows the sub curves oriented differently, so I tried things until it looked the same.

For 4D... I assumed that
- Other than the first and last (0, 15), operations are paired
- The 'middle' operations (7, 8) are special
- The remainder are mirrored about the 'middle'
  - 13 and 14 are mirrors of 1 and 2, etc

Then I spent an hour or four staring at the printout from `ExampleHilbert4D_Curve`, thinking "how do I rotate subcurves 0x1_ and 0x2_ so that 0x10 is adjacent to 0x0F and 0x20 is adjacent to 0x1F" etc.

spatial/curve/curve.go
kortschak: Yep, fair enough.

spatial/curve/curve.go
kortschak: Sure, that sounds good.

spatial/curve/hilbert.go
kortschak: That's what I was thinking, thanks.

spatial/curve/hilbert_test.go
kortschak: Thanks.

kortschak:
> Then I spent an hour or four staring at the printout

Ahh yes. I know that algorithm. I too have used it.

kortschak:
I have things on until next week, but I will take another look then. In the mean time, it occurs to me that it's probably useful to have a method to provide the full length of the curve (maybe `Max() uint64` — if the type returned were an `int` I'd suggest `Len() int`, so an alternative would be to make the returned type for curve points be `int` which should work given that these are likely to be used for indexing anyway). I think also perhaps `Size` should be called `Dims` for naming consistency within Gonum, but I'll leave that up to you.

Thank you for doing this work.

firelizzard18:
Yeah... I wanted to make an n-dimensional `Hilbert` type, but I don't see a pattern I could write an algorithm from. Based on the fact that I came up with two different versions of the 3D curve, clearly there are multiple ways to draw a self-consistent space-filling curve. I don't know whether the alternate versions are 'proper' Hilbert curves. And I wonder if there's some alternative 4D arrangement that would make the pattern clear.

To figure out the pattern, it may be useful to look at the spatial coordinates as the curve is traversed:

```go
func ExampleHilbert3D_Space() {
	h := Hilbert3D{Order: 2}

	for d := uint64(0); d < 1<<(3*h.Order); d++ {
		v := h.Space3D(d)
		for _, v := range v {
			fmt.Printf("%x", v)
		}
		fmt.Println()
	}
}
```

But I haven't gotten anywhere with that.

firelizzard18:
`Dims()` works for me.

The extent of the curve coordinates is [0, 2<sup>N*K</sup>), where N is the dimension and K is the order. That means an 8th order 4D curve or an 11th order 3D curve will overflow `int` on a 32-bit machine. That's definitely more than I need, but I'm not sure its unreasonable for someone to want to use that, which is why I used `uint64`, though I agree that `int` would be nicer.

kortschak:
If they are using it for indexing then they'll have pain anyway. A good way to avoid the problem is to just document the limitations.

spatial/curve/hilbert_test.go
```diff
@@ -0,0 +1,372 @@
...
+	dim := len(c.Dims())
+	actual := make([]int, len(expected))
+	for i := range expected {
+		v := vec(i, order, dim)
+		actual[i] = c.Curve(dup(v))
+	}
+	if !reflect.DeepEqual(expected, actual) {
+		t.Logf("expected %v", expected)
+		t.Logf("got      %v", expected)
+		t.FailNow()
```
kortschak, line 354: This can be spelled with `t.Fatalf`.

spatial/curve/hilbert_test.go
```diff
@@ -0,0 +1,372 @@
...
+			return false
+		}
+		n++
+	}
+
+	return n == 1
+}
+
+func testCurveCase(t *testing.T, c curve, order int, expected []int) {
+	dim := len(c.Dims())
```
kortschak, line 345: Add a call to `t.Helper()`.

kortschak (commented):
I'm wondering whether we should panic for cases where the order can overflow?

spatial/curve/hilbert_test.go
firelizzard18: I did that so the two error messages would be aligned. Would you prefer `t.Fatalf("expected %v\ngot      %v", ...)`? I want them on separate lines, since they're arrays.

Also, the second one definitely should be using `actual`...

spatial/curve/doc.go
```diff
@@ -0,0 +1,44 @@
...
+// In a 2-space with the origin at the bottom left, this results in a ⊐ shape,
+// wound counter clockwise.
+//
+// The runtime of Space and Curve scales as O(n∙k) where n is the dimension and
+// k is the order. The length of the curve is 2^(n∙k).
+//
+// Limitations
+//
+// An n-dimensional, k-order Hilbert curve will not be fully usable if 2ⁿᵏ does
+// not fit in a signed integer. Len will overflow if n∙k ≥ BitSize-1 (where
```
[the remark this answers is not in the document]
firelizzard18: I added a parenthetical note that "if 2ⁿᵏ overflows int" is dependent on architecture. I replaced BitSize with `sizeof(int)` - I find that more readable than the proper expression, `unsafe.Sizeof(int(0))`. Do you think developers not familiar with C will have trouble with that?

firelizzard18:
> I'm wondering whether we should panic for cases where the order can overflow?

My inclination is to say, "These functions are well-defined within a range, with undefined behavior outside of that range." Anyone using this should take care to understand how the spatial and linear resolution of the curve scales with the dimension and order. And if you understand that, you should look at `int` and think, "What happens if the order∙dimension is sufficiently large?"

Granted, I recently spent a while debugging an issue that boiled down to me assuming that `int` is always 4 bytes. I don't have a strong feeling. If you'd prefer it to panic, I'll make the change.

spatial/curve/hilbert_test.go
kortschak: The way that I approach this is the put a newline before the got and wanted lines; the testing logger automatically indents appropriately.

Like so, `t.Fatalf("unexpected result:\ngot: %v\nwant:%v", got, want)`.

spatial/curve/doc.go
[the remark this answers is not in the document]
kortschak: Yeah, true. Maybe, [`bits.UintSize`](https://pkg.go.dev/math/bits#pkg-constants).

firelizzard18:
@kortschak I've resolved everything except the question about panicking

kortschak:
Thanks.

@vladimir-ch Do you have time to take a quick look at the docs here and add a view on the panic issue?

vladimir-ch:
Yes, I will take a look.

spatial/curve/hilbert.go
```diff
@@ -0,0 +1,236 @@
...
+		rx := (v[0] >> n) & 1
+		ry := (v[1] >> n) & 1
+		rd := ry<<1 | (ry ^ rx)
+		d += rd << (2 * n)
+		h.rot(h.Order, v, rd)
+	}
+	return d
+}
+
+// Space2D returns the spatial coordinates of d.
```
vladimir-ch, line 44: What is `d`?

spatial/curve/hilbert_test.go
```diff
@@ -0,0 +1,381 @@
...
+		return
+	}
+
+	for i := 0; i < size[dim-1]; i++ {
+		v[dim-1] = i
+		dimRange(dim-1, size, v, fn)
+	}
+}
+
+// v may get mangled by Curve, so copy it.
```
vladimir-ch, line 324: This documents how/where/why the function is used, not what it does. But probably it doesn't need a comment at all.

spatial/curve/hilbert_test.go
```diff
@@ -0,0 +1,381 @@
...
+// exactly one. v and u must be the same length.
+//
+// In other words, given d(i) = abs(v[i] - u[i]), adjacent returns false if
+// d(i) > 1 for any i or if d(i) == 1 for more than a single i, or if d(i)
+// == 0 for all i.
+func adjacent(v, u []int) bool {
+	n := 0
+	for i := range v {
+		x := v[i] - u[i]
+		if x == 0 {
```
vladimir-ch, line 341: I would write it as a `switch`:
```
diff := v[i] - u[i]
switch diff {
case 0:
case 1, -1:
    n++
default:
    return false
}
```

spatial/curve/hilbert_test.go
```diff
@@ -0,0 +1,381 @@
...
+
+	for i, d := range want {
+		v := vec(i, order, dim)
+		if !reflect.DeepEqual(v, c.Space(d)) {
+			t.Fatalf("[%d] expected %v for d = %d", i, v, d)
+		}
+	}
+}
+
+func vec(n, ord, dim int) []int {
```
vladimir-ch, line 374: On the other hand, this could have a comment what it does.

vladimir-ch:
I'd be fine with merging this as it is. I can address my comments in a follow-up PR.

firelizzard18:
@vladimir-ch My day job started consuming more of my time and energy, which is why I haven't responded.

vladimir-ch:
No problem at all. I totally understand, I've been in the same situation for several years now. I was just worried that I overwhelmed you with my comments and so blocked progress.

kortschak:
Given that no one is in a rush here, I think we can remove the v0.10 milestone.

spatial/curve/doc.go
```diff
@@ -0,0 +1,44 @@
...
+// Hilbert Curves
+//
+// The Hilbert curve is a continuous, space filling curve first described by
+// David Hilbert. The implementation of Hilbert2D is based on example code
+// from the Wikipedia article
+// (https://en.wikipedia.org/w/index.php?title=Hilbert_curve&oldid=1011599190).
+// The implementation of Hilbert3D and Hilbert4D are extrapolated from
+// Hilbert2D.
+//
+// For the first-order n-dimensional Hilbert curve, a spatial point v is
```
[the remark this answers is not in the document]
firelizzard18: I haven't come up with a good way to do that. I could say it's the number of iterations, or the level of recursion, but that's about the implementation, not the curve. I could describe what different orders of curve look like when drawn, but I don't think that would be very useful. If someone asked me, I'd tell them to go look at pictures on Wikipedia.

spatial/curve/doc.go
```diff
@@ -0,0 +1,44 @@
...
+// mapped to a point on the curve d by XOR - each dimension of v is expected
+// to be 0 or 1:
+//
+//     func map1stOrder(n int, v []int) (d int) {
```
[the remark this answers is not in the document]
firelizzard18: It's a way to explain what a Hilbert curve is/how it works. "If you can visualize what an XOR operation does, this is kind of like that." I would be happy to replace this with a better explanation, but I'm having trouble coming up with one.

spatial/curve/hilbert.go
```diff
@@ -0,0 +1,236 @@
...
+	case 3:
+		flip{0, 1}.do(n, v)
+	}
+}
+
+// Curve returns the curve coordinate of v. Curve modifies v.
+//
+// Curve will overflow on a 32-bit architecture if the order is ≥ 16.
+func (h Hilbert2D) Curve(v []int) int {
+	var d int
```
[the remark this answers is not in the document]
firelizzard18: > Should there be a check that `len(v) == 2`?

My thought was more or less, if `v` is too short, it will panic anyways, and if `v` is longer than necessary *shrug* hopefully the user knows what they're doing.

> Are there any restrictions on the components of v?

The values produced by `Space` should be `0 ≤ v < 2ᵏ`. If a value outside of that range is passed to `Curve`, I'm guessing it would give the same result as calling it with each value modulo 2ᵏ. Curve takes N bits at a time, so I think it would ignore everything except the last k bits of each value.

spatial/curve/hilbert.go
```diff
@@ -0,0 +1,236 @@
...
+func (h Hilbert2D) rot(n int, v []int, d int) {
+	switch d {
+	case 0:
+		swap{0, 1}.do(n, v)
+	case 3:
+		flip{0, 1}.do(n, v)
+	}
+}
+
+// Curve returns the curve coordinate of v. Curve modifies v.
```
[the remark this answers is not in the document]
firelizzard18: If you took the curve and straightened it out, and called one end 'zero' and the other 2ⁿᵏ-1 and added a point for every integer in between, then folded the line back into a Hilbert curve, the 'curve coordinate' would be the number at any given point.

That's not a particularly good description. I haven't been able to come up with a better one.

spatial/curve/hilbert.go
firelizzard18: `d` is the input parameter, which should be a curve coordinate in the range [0, 2ⁿᵏ)

spatial/curve/hilbert_test.go
firelizzard18: First I have to remember what it does :facepalm:

spatial/curve/hilbert.go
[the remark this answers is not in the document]
firelizzard18: > If we define position 0 as the end of the curve at `(0, ...)` and position 2ⁿᵏ-1 as the other end of the curve, the 'curve coordinate' of a point on the curve is the position along the curve relative to the two ends.

WDYT?

spatial/curve/doc.go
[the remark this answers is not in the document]
firelizzard18: > Technically, a Hilbert curve is a continuous, fractal, space-filling curve of infinite length, constructed as the limit of a series of piecewise linear curves. We refer to the kth piecewise linear curve as the k-order Hilbert curve. The first-order 2D Hilbert curve looks like ⊐. The k-order n-dimensional curve can be constructed from 2ⁿ copies of the (k-1) order curve. These (finite) Hilbert curves are iterative/recursive, and the order refers to the iteration number.

WDYT?

spatial/curve/hilbert.go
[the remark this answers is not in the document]
kortschak: Since you are using position as the reference for what a coordinate it, I'd just call it the position and rename the method to "Pos" (similarlly, s/Space(?)D/Coord$1D/g and s/Space/Coord/g), with the comment 
```
// Pos returns the linear position of the spatial coordinate along the curve.
```

spatial/curve/hilbert.go
kortschak:
```
// Coord2d returns the spatial coordinate of the linear position along the curve, pos.
```

(rename the param to `pos`).

vladimir-ch:
I'm closing this PR because it has gone stale even though quite some effort went into the implementation and the review. We're still interested in contributions, so please create a new PR if you still intend to submit this change.

PR #1706, opened 2021-08-12
all: replace uses of deprecated ioutil functions

kortschak:
Please take a look.

This is ready to apply when Go1.15 is no longer supported.

### DIFF
--- a/graph/flow/control_flow_bench_test.go
+++ b/graph/flow/control_flow_bench_test.go
@@ -7,7 +7,6 @@ package flow
 import (
 	"flag"
 	"fmt"
-	"io/ioutil"
 	"math"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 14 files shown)


sbinet:
it's seems this CL is missing:
```
graph/formats/dot/internal/lexer/lexer.go:16:   "io/ioutil"
graph/formats/dot/internal/lexer/lexer.go:46:   src, err := ioutil.ReadFile(fpath)
```

kortschak:
That get's generated, so until gocc is updated, that remains.

sbinet:
ah, ok :)
perhaps file an issue for this?

sbinet (dismissed):

kortschak:
They have a long tail of supported versions, so I don't think that would be worthwhile.

sbinet (approved):

Issue #641, opened 2018-11-08
Misleading documentation about support of complex dense matrixes 


taruti:
Currently it seems gonum lacks the support for dense complex valued matrices. i.e. with entries of complex128 instead of float64.

However the documentation in ```mat/doc.go``` claims that it provides:
```
//  - The complementary types for complex matrices, CMatrix, CSymDense, etc.
```

The documentation should probably not claim that complex matrices are supported.

Fixing the documentation is trivial, but actually supporting them would be nicer.

btracey:
The long-term plan is to provide support for complex matrices. I agree that the support is sparse right now (no pun intended). We would like pull requests to help expand that support. Do you have a suggestion for making the documentation clearer while also expressing the intended scope of the package?

wiless:
Can we have support for `CMatrix` in the `mat.Formatted()` to start with ?

kortschak:
The code for that is trivial. The template would be in the following locations.

https://github.com/gonum/gonum/blob/4fb605734974cb56f4f9d2e266358ca50d5fa822/mat/format.go#L12-L30

https://github.com/gonum/gonum/blob/4fb605734974cb56f4f9d2e266358ca50d5fa822/mat/format.go#L59-L223

The relevant changes are largely obvious, and then choose `cFormatter`, `CFormatted`, `cFormat` and `cMaxCellWidth` to avoid name collisions.

brandondube:
Is there some kind of ETA or priority listing so users can get a feel for when this might happen?  Or some kind of dev docs to how the `*Dense` methods work?  I am implementing matrix triple product DFTs as a canary to a larger project but need Outer, Mul, and MulVec.  I looked at the the implementations on Dense but they are noisy and hard to understand as an unfamiliar reader.

kortschak:
> Is there some kind of ETA or priority listing so users can get a feel for when this might happen?

No there is not.

> Or some kind of dev docs to how the *Dense methods work?

What is it that you are having difficulties with? I'm happy to go through questions on golang-dev.
\

chewxy:
hi @brandondube ... try using `gorgonia.org/tensor` as a wrapper instead. Under the hood it uses Gonum's routines.

```
package main

import (
	"fmt"
	"log"

	"gorgonia.org/tensor"
)

func main() {
	a := tensor.New(tensor.WithShape(3, 2), tensor.WithBacking([]complex64{
		complex(0, 0), complex(1, 1),
[...]

PR #1707, opened 2021-08-12
mat: format complex dense matrices

wamuir:
Please take a look.

This PR adds a `CFormat` function in gonum/mat to format complex dense matrices; this function is intended to complement the current `Format` function.  Relevant issue #641 and a relevant PR #1414, although that PR appears to have stalled.  Further, PR #1414 takes the approach of wholesale duplicating `format.go`, which may not be preferred.

Summary:

Following https://github.com/gonum/gonum/pull/1414#issuecomment-666282217, an objective was to avoid code duplication while implementing a formatter for complex matrices.  To summarize the proposed approach:  value formatters are added for formatting floats and complex numbers, which are then used during matrix formatting.  Since these value formatters satisfy the `fmt.Formatter` interface, `fmt.State` can be passed, offering simplification in several areas (e.g., re-construction of a format string is no longer necessary).  As it would be inefficient to spin up a a new formatter for each element in a matrix, a value formatter holds a buffer for reuse.

To tackle widthing, a `state` type is proposed (satisfying `fmt.State`) that can be fit to a matrix and that stores information on column widths, using the existing widther type. As before, iterating through matrix elements is required.   After a call to `At`, a `state` will respond to subsequent `Width` calls with the relevant column width.  Recommendations toward improving implementation are appreciated.

Notwithstanding the need for additional internal types, this feels simpler to me on the whole and I believe this approach should make extending to new syntaxes or data types easier, if desired.  Thanks for reviewing -- will address feedback/comments. 

Bench:

I've included a few benchmarks below.  These compare the current and proposed implementation when formatting 10x10, 100x100 and 1000x1000 dense matrices.

| implementation | benchmark           |  iter |            time |          bytes |           alloc |
| :------------- | :------------------ | ----: | --------------: | -------------: | --------------: |
| current        | BenchmarkFormat10   | 36134 |     33063 ns/op |      2693 B/op |    16 allocs/op |
| current        | BenchmarkFormat100  |   358 |   3747403 ns/op |   1672025 B/op |   137 allocs/op |
| current        | BenchmarkFormat1000 |     3 | 334034817 ns/op | 163557800 B/op |  1066 allocs/op |
| proposed       | BenchmarkFormat10   | 22086 |     54554 ns/op |      2886 B/op |    19 allocs/op |
| proposed       | BenchmarkFormat100  |   170 |   7072273 ns/op |   1670752 B/op |    49 allocs/op |
| proposed       | BenchmarkFormat1000 |     2 | 697675398 ns/op | 163542896 B/op |    80 allocs/op |

### DIFF
--- a/mat/format.go
+++ b/mat/format.go
@@ -173,25 +173,25 @@ func format(m formattable, prefix string, margin int, dot byte, squeeze bool, fs
 		fmt.Fprintf(fs, "Dims(%d, %d)\n", rows, cols)
 	}
 
+	var buf = make([]byte, 0, 4)
 	for i := 0; i < rows; i++ {
 		if !first {
 			fmt.Fprint(fs, prefix)
 		}
 		first = false
-		var el string
 		switch {
 		case rows == 1:
 			fmt.Fprint(fs, "[")
-			el = "]"
+			buf = append(buf[:0], []byte("]")...)
 		case i == 0:
 			fmt.Fprint(fs, "⎡")
-			el = "⎤\n"
+			buf = append(buf[:0], []byte("⎤\n")...)
 		case i < rows-1:
 			fmt.Fprint(fs, "⎢")
-			el = "⎥\n"
+			buf = append(buf[:0], []byte("⎥\n")...)
 		default:
 			fmt.Fprint(fs, "⎣")
-			el = "⎦"
+			buf = append(buf[:0], []byte("⎦")...)
 		}
 
 		for j := 0; j < cols; j++ {
@@ -213,7 +213,7 @@ func format(m formattable, prefix string, margin int, dot byte, squeeze bool, fs
 			}
 		}
 
-		fmt.Fprint(fs, el)
+		fs.Write(buf)
 
 		if i >= printed-1 && i < rows-printed && 2*printed < rows {
 			i = rows - printed - 1
@@ -274,21 +274,21 @@ func formatMATLAB(m formattable, prefix string, _ int, _ byte, squeeze bool, fs 
 		return
 	}
 
+	var buf = make([]byte, 0, 4)
 	for i := 0; i < rows; i++ {
-		var el string
 		switch {
 		case rows == 1:
 			fmt.Fprint(fs, "[")
-			el = "]"
+			buf = append(buf[:0], []byte("]")...)
 		case i == 0:
 			fmt.Fprint(fs, "[\n"+prefix+" ")
-			el = "\n"
+			buf = append(buf[:0], []byte("\n")...)
 		case i < rows-1:
 			fmt.Fprint(fs, prefix+" ")
-			el = "\n"
+			buf = append(buf[:0], []byte("\n")...)
 		default:
 			fmt.Fprint(fs, prefix+" ")
-			el = "\n" + prefix + "]"
+			buf = append(buf[:0], []byte("\n"+prefix+"]")...)
 		}
 
 		for j := 0; j < cols; j++ {
@@ -300,7 +300,7 @@ func formatMATLAB(m formattable, prefix string, _ int, _ byte, squeeze bool, fs 
 			}
 		}
 
-		fmt.Fprint(fs, el)
+		fs.Write(buf)
 	}
 }
 
@@ -361,24 +361,24 @@ func formatPython(m formattable, prefix string, _ int, _ byte, squeeze bool, fs 
 		return
 	}
 
+	var buf = make([]byte, 0, 4)
 	for i := 0; i < rows; i++ {
 		if i != 0 {
 			fmt.Fprint(fs, prefix)
 		}
-		var el string
 		switch {
 		case rows == 1:
 			fmt.Fprint(fs, "[")
-			el = "]"
+			buf = append(buf[:0], []byte("]")...)
 		case i == 0:
 			fmt.Fprint(fs, "[[")
-			el = "],\n"
+			buf = append(buf[:0], []byte("],\n")...)
 		case i < rows-1:
 			fmt.Fprint(fs, " [")
-			el = "],\n"
+			buf = append(buf[:0], []byte("],\n")...)
 		default:
 			fmt.Fprint(fs, " [")
-			el = "]]"
+			buf = append(buf[:0], []byte("]]")...)
 		}
 
 		for j := 0; j < cols; j++ {
@@ -390,7 +390,7 @@ func formatPython(m formattable, prefix string, _ int, _ byte, squeeze bool, fs 
 			}
 		}
 
-		fmt.Fprint(fs, el)
+		fs.Write(buf)
 	}
 }
 
--- a/mat/format.go
+++ b/mat/format.go
@@ -243,20 +243,20 @@ func formatMATLAB(m formattable, prefix string, _ int, _ byte, squeeze bool, fs 
 			fmt.Fprintf(fs, "%%!%c(%T=Dims(%d, %d))", c, m.Mat(), rows, cols)
 			return
 		}
-		fs.Write([]byte{'['})
+		fmt.Fprint(fs, "[")
 		for i := 0; i < rows; i++ {
 			if i != 0 {
-				fs.Write([]byte("; "))
+				fmt.Fprint(fs, "; ")
 			}
 			for j := 0; j < cols; j++ {
 				if j != 0 {
-					fs.Write([]byte{' '})
+					fmt.Fprint(fs, " ")
 				}
 				f = m.RevalueFormatter(f, i, j)
 				f.Format(fs, c)
 			}
 		}
-		fs.Write([]byte{']'})
+		fmt.Fprint(fs, "]")
 		return
 	}
 
@@ -324,26 +324,26 @@ func formatPython(m formattable, prefix string, _ int, _ byte, squeeze bool, fs 
 			fmt.Fprintf(fs, "%%!%c(%T=Dims(%d, %d))", c, m.Mat(), rows, cols)
 			return
 		}
-		fs.Write([]byte{'['})
+		fmt.Fprint(fs, "[")
 		if rows > 1 {
-			fs.Write([]byte{'['})
+			fmt.Fprint(fs, "[")
 		}
 		for i := 0; i < rows; i++ {
 			if i != 0 {
-				fs.Write([]byte("], ["))
+				fmt.Fprint(fs, "], [")
 			}
 			for j := 0; j < cols; j++ {
 				if j != 0 {
-					fs.Write([]byte(", "))
+					fmt.Fprint(fs, ", ")
 				}
 				f = m.RevalueFormatter(f, i, j)
 				f.Format(fs, c)
 			}
 		}
 		if rows > 1 {
-			fs.Write([]byte{']'})
+			fmt.Fprint(fs, "]")
 		}
-		fs.Write([]byte{']'})
+		fmt.Fprint(fs, "]")
 		return
 	}
 
--- a/mat/format.go
+++ b/mat/format.go
@@ -32,6 +32,8 @@ type formatter struct {
 	format func(m Matrix, prefix string, margin int, dot byte, squeeze bool, fs fmt.State, c rune)
 }
 
+var _ fmt.Formatter = (*formatter)(nil)
+
 // FormatOption is a functional option for matrix formatting.
 type FormatOption func(*formatter)
 
@@ -507,10 +509,14 @@ type widther interface {
 
 type uniformWidth int
 
+var _ widther = (*uniformWidth)(nil)
+
 func (u *uniformWidth) width(_ int) int   { return int(*u) }
 func (u *uniformWidth) setWidth(_, w int) { *u = uniformWidth(w) }
 
 type columnWidth []int
+
+var _ widther = (*columnWidth)(nil)
 
 func (c columnWidth) width(i int) int   { return c[i] }
 func (c columnWidth) setWidth(i, w int) { c[i] = w }
--- a/mat/format.go
+++ b/mat/format.go
@@ -19,7 +19,7 @@ func Formatted(m Matrix, options ...FormatOption) fmt.Formatter {
 	for _, o := range options {
 		o(&f)
 	}
-	return f
+	return &f
 }
 
 type formatter struct {
--- a/mat/format.go
+++ b/mat/format.go
@@ -5,15 +5,14 @@
 package mat
 
 import (
+	"bytes"
 	"fmt"
-	"strconv"
-	"strings"
 )
 
 // Formatted returns a fmt.Formatter for the matrix m using the given options.
 func Formatted(m Matrix, options ...FormatOption) fmt.Formatter {
 	f := formatter{
-		matrix: m,
+		matrix: formattableMatrix{mat: m},
 		dot:    '.',
 	}
 	for _, o := range options {
@@ -22,14 +21,15 @@ func Formatted(m Matrix, options ...FormatOption) fmt.Formatter {
 	return &f
 }
 
+// formatter is a matrix formatter that satisfies the fmt.Formatter interface
 type formatter struct {
-	matrix  Matrix
+	matrix  formattable
 	prefix  string
 	margin  int
 	dot     byte
 	squeeze bool
 
-	format func(m Matrix, prefix string, margin int, dot byte, squeeze bool, fs fmt.State, c rune)
+	format func(m formattable, prefix string, margin int, dot byte, squeeze bool, fs fmt.State, c rune)
 }
 
 var _ fmt.Formatter = (*formatter)(nil)
@@ -78,7 +78,7 @@ func FormatPython() FormatOption {
 // Format satisfies the fmt.Formatter interface.
 func (f formatter) Format(fs fmt.State, c rune) {
 	if c == 'v' && fs.Flag('#') && f.format == nil {
-		fmt.Fprintf(fs, "%#v", f.matrix)
+		fmt.Fprintf(fs, "%#v", f.matrix.Mat())
 		return
 	}
 	if f.format == nil {
@@ -86,6 +86,52 @@ func (f formatter) Format(fs fmt.State, c rune) {
 	}
 	f.format(f.matrix, f.prefix, f.margin, f.dot, f.squeeze, fs, c)
 }
+
+// formattable holds a matrix and provides methods for formatting values.
+type formattable interface {
+	tabular
+
+	// NewValueFormatter returns a new, zero-valued fmt.Formatter
+	NewValueFormatter(options ...vFormatOption) fmt.Formatter
+
+	// RevalueFormatter sets the value of a formatter to element i, j.
+	// A use is to recycle an existing value formatter for efficiency.
+	RevalueFormatter(f fmt.Formatter, i, j int) fmt.Formatter
+
+	// ValueFormatter returns a new fmt.Formatter for element i, j
+	ValueFormatter(i, j int, options ...vFormatOption) fmt.Formatter
+
+	// Matrix returns the source matrix
+	Mat() tabular
+}
+
+// formattableMatrix is a formattable for floating point-valued matrices, those
+// that satisfy the Matrix interface.
+type formattableMatrix struct{ mat Matrix }
+
+var _ formattable = (*formattableMatrix)(nil)
+
+func (m formattableMatrix) Dims() (r, c int) { return m.mat.Dims() }
+func (m formattableMatrix) Mat() tabular     { return m.mat }
+func (m formattableMatrix) NewValueFormatter(options ...vFormatOption) fmt.Formatter {
+	return formattedFloat(0, options...)
+}
+func (m formattableMatrix) RevalueFormatter(f fmt.Formatter, i, j int) fmt.Formatter {
+	f.(*floatFormatter).value = m.mat.At(i, j)
+	return f
+}
+func (m formattableMatrix) ValueFormatter(i, j int, options ...vFormatOption) fmt.Formatter {
+	return formattedFloat(m.mat.At(i, j), options...)
+}
+
+// tabular type is a two-dimensional array having rows and columns with table
+// elements accessible by a formattable. Details regarding value type and
+// methods for element access are left to the formattable.
+type tabular interface {
+	Dims() (r, c int)
+}
+
+var _ tabular = (Matrix)(nil)
 
 // format prints a pretty representation of m to the fs io.Writer. The format character c
 // specifies the numerical representation of elements; valid values are those for float64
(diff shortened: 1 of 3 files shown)


sbinet:
just a drive-by comment:
could you run the benchmarks before/after the proposed optimization a couple of times (say 10-20) and show the result of `benchstat`?
with the limited statistics you presented in the commit message, I get this rather poor statistical power result:

```
$> benchstat ./ref.txt ./new.txt 
name        old time/op    new time/op    delta
Format10      33.1µs ± 0%    54.6µs ± 0%   ~     (p=1.000 n=1+1)
Format100     3.75ms ± 0%    7.07ms ± 0%   ~     (p=1.000 n=1+1)
Format1000     334ms ± 0%     698ms ± 0%   ~     (p=1.000 n=1+1)

name        old alloc/op   new alloc/op   delta
Format10      2.69kB ± 0%    2.89kB ± 0%   ~     (p=1.000 n=1+1)
Format100     1.67MB ± 0%    1.67MB ± 0%   ~     (p=1.000 n=1+1)
Format1000     164MB ± 0%     164MB ± 0%   ~     (p=1.000 n=1+1)

name        old allocs/op  new allocs/op  delta
Format10        16.0 ± 0%      19.0 ± 0%   ~     (p=1.000 n=1+1)
Format100        137 ± 0%        49 ± 0%   ~     (p=1.000 n=1+1)
Format1000     1.07k ± 0%     0.08k ± 0%   ~     (p=1.000 n=1+1)
```

ie:
```
$> go test -run=NONE -bench=Format -benchmem -count=20 |& tee ref.txt
## apply optimizations
$> go test -run=NONE -bench=Format -benchmem -count=20 |& tee new.txt
$> benchstat ./ref.txt ./new.txt
```

thanks.

PS: `benchstat` is available here:
- [golang.org/x/perf/benchstat](https://golang.org/x/perf/benchstat)

kortschak:
@wamuir Thank you for doing this. I've only taken a brief look, but it looks like nice work so far. I hope to be able to have a deeper look over the weekend.

mat/format_test.go
```diff
@@ -238,3 +240,16 @@ func TestFormat(t *testing.T) {
...
+	a, _ := randDense(size, 1.0, src)
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		_ = fmt.Sprintf("%v", Formatted(a))
+	}
+}
+
+func BenchmarkFormat10(b *testing.B)   { benchmarkFormat(b, 10) }
+func BenchmarkFormat100(b *testing.B)  { benchmarkFormat(b, 100) }
+func BenchmarkFormat1000(b *testing.B) { benchmarkFormat(b, 1000) }
```
kortschak, line 521: I'd be inclined to use sub-benchmarks for this. There are many places where we don't, but that is for historical reasons (the didn't exist when those benchmarks were written).

mat/format.go
```diff
@@ -19,7 +19,7 @@ func Formatted(m Matrix, options ...FormatOption) fmt.Formatter {
 	for _, o := range options {
 		o(&f)
 	}
-	return f
+	return &f
```
kortschak, line 21: This change is not necessary; all the `Format` method remains on the concrete type and the value is small.

mat/format_value.go
```diff
@@ -0,0 +1,143 @@
...
+// valueFormatter is a type that can set options for formatting values and that
+// satisfies the fmt.Formatter interface.
+type valueFormatter interface {
+	fmt.Formatter
+
+	setDotByte(b byte)
+	setSyntax(s int)
+}
+
+// vFormatOption is a functional option for value formatting.
```
kortschak, line 30: Can we expand the `v` to `value` throughout please.

mat/format_value.go
```diff
@@ -0,0 +1,143 @@
...
+		o(&f)
+	}
+	return &f
+}
+
+// floatFormatter formats 64-bit a floating point value and satifies the
+// valueFormatter interface. A floatFormatter utilizes an internal buffer while
+// generating formatting output and this buffer persists with the formatter.
+// For this reason, and because a floatFormatter might be reused or revalued,
+// the floatFormatter embeds an exclusion lock.
```
kortschak, line 71: Reuse doesn't require a lock, only if there is concurrent use, but concurrent use it not a goal, so the mutex can be dropped.

mat/format_value.go
```diff
@@ -0,0 +1,143 @@
...
+	dot    byte
+	syntax int
+
+	format func(v float64, buf []byte, dot byte, syntax int, fs fmt.State, c rune)
+}
+
+var _ valueFormatter = (*floatFormatter)(nil)
+
+// Format satisfies the fmt.Formatter interface.
+func (f *floatFormatter) Format(fs fmt.State, c rune) {
```
kortschak, line 86: These methods don't need to be on a pointer.

mat/format.go
```diff
@@ -87,6 +87,52 @@ func (f formatter) Format(fs fmt.State, c rune) {
...
+
+	// Matrix returns the source matrix
+	Mat() tabular
+}
+
+// formattableMatrix is a formattable for floating point-valued matrices, those
+// that satisfy the Matrix interface.
+type formattableMatrix struct{ mat Matrix }
+
+var _ formattable = (*formattableMatrix)(nil)
```
kortschak, line 125: This should not check against the point since the methods are on the concrete type.

mat/format.go
```diff
@@ -87,6 +87,52 @@ func (f formatter) Format(fs fmt.State, c rune) {
...
+	return f
+}
+func (m formattableMatrix) ValueFormatter(i, j int, options ...vFormatOption) fmt.Formatter {
+	return formattedFloat(m.mat.At(i, j), options...)
+}
+
+// tabular type is a two-dimensional array having rows and columns with table
+// elements accessible by a formattable. Details regarding value type and
+// methods for element access are left to the formattable.
+type tabular interface {
```
kortschak, line 162: By convention we would probably call this `dimer` (or `dimser`) rather than `tabular`.

mat/format.go
```diff
@@ -87,6 +87,52 @@ func (f formatter) Format(fs fmt.State, c rune) {
...
+	Dims() (r, c int)
+}
+
+var _ tabular = (Matrix)(nil)
```
kortschak, line 166: This is not necessary since `Matrix` is the canonical `interface{ Dim() (r, c int) }` type.

mat/format.go
```diff
@@ -427,58 +382,63 @@ func formatPython(m Matrix, prefix string, _ int, _ byte, squeeze bool, fs fmt.S
...
-		rows, cols = m.Dims()
-		max        int
-	)
+// fit fits a state to formattable m, storing a single width for the
+// formattable or, if squeeze is true, individual widths are stored for each
+// column in the formattable. After fitment, the state responds to Width calls
+// with the value in p, which may be updated for a new column by calling At.
+//
+// The fitted state is not safe for concurrent use.
+func (ws *state) fit(m formattable, printed int, squeeze bool, fs fmt.State, c rune, f fmt.Formatter) *state {
```
kortschak, line 459: s/ws/s/g

mat/format.go
```diff
@@ -427,58 +382,63 @@ func formatPython(m Matrix, prefix string, _ int, _ byte, squeeze bool, fs fmt.S
...
 
-func maxCellWidth(m Matrix, c rune, printed, prec int, w widther) ([]byte, int) {
-	var (
-		buf        = make([]byte, 0, 64)
-		rows, cols = m.Dims()
-		max        int
-	)
+// fit fits a state to formattable m, storing a single width for the
+// formattable or, if squeeze is true, individual widths are stored for each
+// column in the formattable. After fitment, the state responds to Width calls
```
kortschak, line 455: s/fitment/fitting/

mat/format.go
```diff
@@ -427,58 +382,63 @@ func formatPython(m Matrix, prefix string, _ int, _ byte, squeeze bool, fs fmt.S
...
+	flag      func(c int) bool
+
+	w widther
+	p int
+}
+
+var _ fmt.State = (*state)(nil)
+
+// newState return a new state, shadowing given fmt.State fs.
+func newState(fs fmt.State) *state {
```
kortschak, line 443: It's not clear to me why this needs to return a pointer type; the fit method returns the mutated `state`.  Either it can mutate in place and not bother returning the value (and having the remaining methods on the pointer), or preferably, returning the mutated non-pointer `state` (this would require that the `fmt.State` assignability check above changes).

mat/format.go
```diff
@@ -427,58 +382,63 @@ func formatPython(m Matrix, prefix string, _ int, _ byte, squeeze bool, fs fmt.S
...
+// fit fits a state to formattable m, storing a single width for the
+// formattable or, if squeeze is true, individual widths are stored for each
+// column in the formattable. After fitment, the state responds to Width calls
+// with the value in p, which may be updated for a new column by calling At.
+//
+// The fitted state is not safe for concurrent use.
+func (ws *state) fit(m formattable, printed int, squeeze bool, fs fmt.State, c rune, f fmt.Formatter) *state {
+	rows, cols := m.Dims()
+
+	// assign appropriate widther
```
kortschak, line 462: Initial capital and final period for these comments.

mat/format.go
```diff
@@ -173,25 +173,25 @@ func format(m formattable, prefix string, margin int, dot byte, squeeze bool, fs
 		fmt.Fprintf(fs, "Dims(%d, %d)\n", rows, cols)
 	}
 
+	var buf = make([]byte, 0, 4)
```
kortschak, line 208: `buf := make([]byte, 0, 4) // Four bytes required because of unicode decoration.`

mat/format.go
```diff
@@ -274,21 +274,21 @@ func formatMATLAB(m formattable, prefix string, _ int, _ byte, squeeze bool, fs
 		return
 	}
 
+	var buf = make([]byte, 0, 4)
```
kortschak, line 309: I don't think this presizing to 4 makes sense here since in the default case `prefix` is arbitrarily long with a minumum of two. The remainder are one byte long, so maybe just presize to 2, `buf = make([]byte, 0, 2)` and let it grow if needed in the default case when `len(prefix) != 0`.

mat/format.go
```diff
@@ -361,24 +361,24 @@ func formatPython(m formattable, prefix string, _ int, _ byte, squeeze bool, fs
...
```
kortschak, line 396: `buf = make([]byte, 0, 3)`

mat/format.go
```diff
@@ -243,20 +243,20 @@ func formatMATLAB(m formattable, prefix string, _ int, _ byte, squeeze bool, fs
 			fmt.Fprintf(fs, "%%!%c(%T=Dims(%d, %d))", c, m.Mat(), rows, cols)
 			return
 		}
-		fs.Write([]byte{'['})
+		fmt.Fprint(fs, "[")
```
kortschak, line 278: I'd prefer not to make these changes; they are more heavy weight. `fmt` depends on `reflect` and interpretation of types while write just spits bytes which the compiler has nicely prepared.

mat/format.go
```diff
@@ -21,6 +21,19 @@ func Formatted(m Matrix, options ...FormatOption) fmt.Formatter {
...
+// given options.
+func CFormatted(m CMatrix, options ...FormatOption) fmt.Formatter {
+	f := formatter{
+		matrix: formattableCMatrix{mat: m},
+		dot:    '.',
+	}
+	for _, o := range options {
+		o(&f)
+	}
+	return &f
```
kortschak, line 34: Return the non-pointer type.

mat/format_value.go
```diff
@@ -141,3 +141,113 @@ func formatFloat(v float64, buf []byte, dot byte, syntax int, fs fmt.State, c ru
...
+	dot    byte
+	syntax int
+
+	format func(v complex128, buf []byte, dot byte, syntax int, fs fmt.State, c rune)
+}
+
+var _ valueFormatter = (*complexFormatter)(nil)
+
+// Format satisfies the fmt.Formatter interface.
+func (f *complexFormatter) Format(fs fmt.State, c rune) {
```
kortschak, line 179: Here too, this doesn't need to be on a pointer for the same reasons.

mat/format_value.go
```diff
@@ -141,3 +141,113 @@ func formatFloat(v float64, buf []byte, dot byte, syntax int, fs fmt.State, c ru
...
+		o(&f)
+	}
+	return &f
+}
+
+// floatComplex formats 128-bit complex value and satifies the valueFormatter
+// interface. A complexFormatter utilizes an internal buffer while generating
+// formatting output and this buffer persists with the formatter. For this
+// reason, and because a complexFormatter might be reused or revalued,
+// the complexFormatter embeds an exclusion lock.
```
kortschak, line 164: Same situation wrt the mutex.

wamuir:
> just a drive-by comment:
> could you run the benchmarks before/after the proposed optimization a couple of times (say 10-20) and show the result of `benchstat`?
> with the limited statistics you presented in the commit message, I get this rather poor statistical power result:
> 
> ```
> $> benchstat ./ref.txt ./new.txt 
> name        old time/op    new time/op    delta
> Format10      33.1µs ± 0%    54.6µs ± 0%   ~     (p=1.000 n=1+1)
> Format100     3.75ms ± 0%    7.07ms ± 0%   ~     (p=1.000 n=1+1)
> Format1000     334ms ± 0%     698ms ± 0%   ~     (p=1.000 n=1+1)
> 
> name        old alloc/op   new alloc/op   delta
> Format10      2.69kB ± 0%    2.89kB ± 0%   ~     (p=1.000 n=1+1)
> Format100     1.67MB ± 0%    1.67MB ± 0%   ~     (p=1.000 n=1+1)
> Format1000     164MB ± 0%     164MB ± 0%   ~     (p=1.000 n=1+1)
> 
> name        old allocs/op  new allocs/op  delta
> Format10        16.0 ± 0%      19.0 ± 0%   ~     (p=1.000 n=1+1)
> Format100        137 ± 0%        49 ± 0%   ~     (p=1.000 n=1+1)
> Format1000     1.07k ± 0%     0.08k ± 0%   ~     (p=1.000 n=1+1)
> ```
> 
> ie:
> 
> ```
> $> go test -run=NONE -bench=Format -benchmem -count=20 |& tee ref.txt
> ## apply optimizations
> $> go test -run=NONE -bench=Format -benchmem -count=20 |& tee new.txt
> $> benchstat ./ref.txt ./new.txt
> ```
> 
> thanks.
> 
> PS: `benchstat` is available here:
> 
>     * [golang.org/x/perf/benchstat](https://golang.org/x/perf/benchstat)

Thanks, I wasn't aware of benchstat.  Cool.  Output is below.  I added `-benchtime=1x`, because this appears to otherwise perform tests using the pooled mean?  If true, I'd think the dispersion of ± time/op is otherwise understated. I guess that's a nit on `go test` for not having a flag to output variance (or a nit on me for not being able to find it).

| name       | old time/op | new time/op  | delta                       |
| :--------- | ----------: | -----------: | --------------------------: |
| Format10   | 44.2µs ±17% |  73.9µs ±38% |  +67.07%  (p=0.000 n=20+18) |
| Format100  | 3.28ms ± 3% |  7.55ms ± 5% | +130.21%  (p=0.000 n=20+18) |
| Format1000 |  329ms ± 1% |   764ms ± 2% | +132.17%  (p=0.000 n=20+20) |

Increase in clock time is statistically significant.  @kortschak provided review comments, which I'll work through over the weekend and then rerun the bench.  One of these comments is to use micro-benchmarking and I expect this will offer greater insight into differences.  Addressing his comments may also close the current gap (e.g., his comment regarding fmt.Fprint and reflect).

wamuir:
Most recent commit addresses review comments, except for the switch to sub-benchmarks.  I'll need to push a second commit to update the benchmarks.  I attempted some further simplification of the format functions and I think this was productive.  I also added comments along the way.  Updated benchstat output is below.

| name       | old time/op | new time/op  | delta                       |
| :--------- | ----------: | -----------: | --------------------------: |
| Format10   | 44.9µs ±10% |  62.9µs ±31% |  +40.08%  (p=0.000 n=18+20) |
| Format100  | 3.31ms ± 1% |  4.27ms ± 1% |  +29.20%  (p=0.000 n=18+20) |
| Format1000 |  326ms ± 1% |   426ms ± 1% |  +30.41%  (p=0.000 n=20+20) |

mat/format.go
wamuir: This was helpful, thanks.

wamuir:
| name                | old time/op | new time/op | delta                      |
| :------------------ | ----------: | ----------: | -------------------------: |
| Format/10/General   | 52.6µs ±17% | 68.8µs ±18% | +30.74%  (p=0.000 n=19+19) |
| Format/10/DotByte   | 49.5µs ±21% | 67.9µs ±17% | +37.19%  (p=0.000 n=19+20) |
| Format/10/Excerpt   | 50.3µs ±11% | 66.1µs ±21% | +31.40%  (p=0.000 n=19+20) |
| Format/10/Prefix    | 49.4µs ±10% | 66.4µs ±17% | +34.54%  (p=0.000 n=17+20) |
| Format/10/Squeeze   | 53.0µs ±22% | 69.7µs ±14% | +31.49%  (p=0.000 n=20+20) |
| Format/10/MATLAB    | 51.5µs ±12% | 66.4µs ±12% | +28.79%  (p=0.000 n=20+19) |
| Format/10/MATLAB#   | 43.0µs ±18% | 42.9µs ±23% |    ~     (p=0.967 n=20+20) |
| Format/10/Python    | 50.4µs ±13% | 64.0µs ± 9% | +26.99%  (p=0.000 n=20+19) |
| Format/10/Python#   | 42.4µs ±37% | 43.1µs ±30% |    ~     (p=0.757 n=20+20) |
| Format/100/General  | 3.13ms ± 7% | 4.43ms ± 9% | +41.42%  (p=0.000 n=17+20) |
| Format/100/DotByte  | 3.94ms ± 1% | 4.35ms ± 8% | +10.32%  (p=0.000 n=19+20) |
| Format/100/Excerpt  | 4.00ms ± 1% | 4.54ms ± 6% | +13.45%  (p=0.000 n=20+20) |
| Format/100/Prefix   | 4.00ms ± 1% | 5.61ms ± 1% | +40.23%  (p=0.000 n=20+16) |
| Format/100/Squeeze  | 4.42ms ± 1% | 5.35ms ±18% | +21.01%  (p=0.000 n=14+20) |
| Format/100/MATLAB   | 4.68ms ±15% | 6.46ms ±13% | +37.92%  (p=0.000 n=20+20) |
| Format/100/MATLAB#  | 2.91ms ± 3% | 2.90ms ± 2% |    ~     (p=0.247 n=20+20) |
| Format/100/Python   | 6.33ms ±16% | 6.38ms ±35% |    ~     (p=0.890 n=20+20) |
| Format/100/Python#  | 3.56ms ±13% | 1.69ms ± 1% | -52.45%  (p=0.000 n=20+19) |
| Format/1000/General |  306ms ± 0% |  405ms ± 0% | +32.08%  (p=0.000 n=19+20) |
| Format/1000/DotByte |  303ms ± 0% |  392ms ± 0% | +29.54%  (p=0.000 n=20+20) |
| Format/1000/Excerpt |  308ms ± 1% |  405ms ± 0% | +31.76%  (p=0.000 n=20+20) |
| Format/1000/Prefix  |  307ms ± 0% |  404ms ± 0% | +31.67%  (p=0.000 n=20+20) |
| Format/1000/Squeeze |  307ms ± 0% |  405ms ± 0% | +31.83%  (p=0.000 n=20+20) |
| Format/1000/MATLAB  |  307ms ± 0% |  405ms ± 0% | +31.99%  (p=0.000 n=19+20) |
| Format/1000/MATLAB# |  164ms ± 1% |  164ms ± 1% |    ~     (p=0.561 n=19+19) |
| Format/1000/Python  |  308ms ± 1% |  406ms ± 1% | +31.86%  (p=0.000 n=20+18) |
| Format/1000/Python# |  164ms ± 1% |  164ms ± 1% |    ~     (p=0.604 n=20+19) |
| [Geo mean]          |      3.74ms |      4.43ms | +18.29%                    |

wamuir:
| name                | old time/op | new time/op | delta                      |
| :------------------ | ----------: | ----------: | -------------------------: |
| Format/10/General   | 52.6µs ±17% | 68.8µs ±18% | +30.74%  (p=0.000 n=19+19) |
| Format/10/DotByte   | 49.5µs ±21% | 67.9µs ±17% | +37.19%  (p=0.000 n=19+20) |
| Format/10/Excerpt   | 50.3µs ±11% | 66.1µs ±21% | +31.40%  (p=0.000 n=19+20) |
| Format/10/Prefix    | 49.4µs ±10% | 66.4µs ±17% | +34.54%  (p=0.000 n=17+20) |
| Format/10/Squeeze   | 53.0µs ±22% | 69.7µs ±14% | +31.49%  (p=0.000 n=20+20) |
| Format/10/MATLAB    | 51.5µs ±12% | 66.4µs ±12% | +28.79%  (p=0.000 n=20+19) |
| Format/10/MATLAB#   | 43.0µs ±18% | 42.9µs ±23% |    ~     (p=0.967 n=20+20) |
| Format/10/Python    | 50.4µs ±13% | 64.0µs ± 9% | +26.99%  (p=0.000 n=20+19) |
| Format/10/Python#   | 42.4µs ±37% | 43.1µs ±30% |    ~     (p=0.757 n=20+20) |
| Format/100/General  | 3.13ms ± 7% | 4.43ms ± 9% | +41.42%  (p=0.000 n=17+20) |
| Format/100/DotByte  | 3.94ms ± 1% | 4.35ms ± 8% | +10.32%  (p=0.000 n=19+20) |
| Format/100/Excerpt  | 4.00ms ± 1% | 4.54ms ± 6% | +13.45%  (p=0.000 n=20+20) |
| Format/100/Prefix   | 4.00ms ± 1% | 5.61ms ± 1% | +40.23%  (p=0.000 n=20+16) |
| Format/100/Squeeze  | 4.42ms ± 1% | 5.35ms ±18% | +21.01%  (p=0.000 n=14+20) |
| Format/100/MATLAB   | 4.68ms ±15% | 6.46ms ±13% | +37.92%  (p=0.000 n=20+20) |
| Format/100/MATLAB#  | 2.91ms ± 3% | 2.90ms ± 2% |    ~     (p=0.247 n=20+20) |
| Format/100/Python   | 6.33ms ±16% | 6.38ms ±35% |    ~     (p=0.890 n=20+20) |
| Format/100/Python#  | 3.56ms ±13% | 1.69ms ± 1% | -52.45%  (p=0.000 n=20+19) |
| Format/1000/General |  306ms ± 0% |  405ms ± 0% | +32.08%  (p=0.000 n=19+20) |
| Format/1000/DotByte |  303ms ± 0% |  392ms ± 0% | +29.54%  (p=0.000 n=20+20) |
| Format/1000/Excerpt |  308ms ± 1% |  405ms ± 0% | +31.76%  (p=0.000 n=20+20) |
| Format/1000/Prefix  |  307ms ± 0% |  404ms ± 0% | +31.67%  (p=0.000 n=20+20) |
| Format/1000/Squeeze |  307ms ± 0% |  405ms ± 0% | +31.83%  (p=0.000 n=20+20) |
| Format/1000/MATLAB  |  307ms ± 0% |  405ms ± 0% | +31.99%  (p=0.000 n=19+20) |
| Format/1000/MATLAB# |  164ms ± 1% |  164ms ± 1% |    ~     (p=0.561 n=19+19) |
| Format/1000/Python  |  308ms ± 1% |  406ms ± 1% | +31.86%  (p=0.000 n=20+18) |
| Format/1000/Python# |  164ms ± 1% |  164ms ± 1% |    ~     (p=0.604 n=20+19) |
| [Geo mean]          |      3.74ms |      4.43ms | +18.29%                    |

vladimir-ch:
This PR has gone a bit stale. @wamuir are you still interested in submitting this change?

wamuir:
@vladimir-ch Yes. I think this needs another review. I can make revisions.

PR #1708, opened 2021-08-16
graph/topo: add 3 variants of Johnson's Cycle Finding Algorithm

wardbradt:
Please take a look.

This change modifies Gonum to add `DirectedCyclesContaining`, `DirectedCyclesOfMaxLen`, `DirectedCyclesOfMaxLenContaining` to the `topo` package (along with corresponding unit tests). 

The algorithms used in these 3 new functions are variants of Johnson's cycle finding algorithm. Each of them searches for a specific subset of cycles in a directed graph using a modified algorithm that is more efficient than post-processing the output of `DirectedCyclesIn`.

I am a first-time contributor to gonum so I am not sure if such specific functionality is desired.


graph/topo/johnson_cycles.go
```diff
@@ -70,6 +70,114 @@ func DirectedCyclesIn(g graph.Directed) [][]graph.Node {
 	return j.result
 }
 
+// DirectedCyclesContaining returns the set of elementary cycles in the graph g containing the node with ID vid.
+// This is a slight deviation from Johnson's.
```
kortschak, line 74: The final sentence is not needed.

graph/topo/johnson_cycles.go
```diff
@@ -70,6 +70,114 @@ func DirectedCyclesIn(g graph.Directed) [][]graph.Node {
...
+
+	for i, v := range j.adjacent.orig {
+		if !j.adjacent.nodes.Has(v.ID()) {
+			continue
+		}
+		if len(j.adjacent.succ[v.ID()]) > 0 {
+			j.blocked[i] = false
+			j.b[i] = make(set.Ints)
+		}
+	}
```
kortschak, line 100: Can you factor this out to a helper so it's not quadruplicated?

graph/topo/johnson_cycles.go
```diff
@@ -107,6 +215,47 @@ func (j *johnson) circuit(v int) bool {
 	return f
 }
 
+// circuitMaxLen is the CIRCUIT sub-procedure in the paper but produces circuits up to a maximum length.
+func (j *johnson) circuitMaxLen(v int, maxLen int) bool {
+	f := false
+	// TODO: can we replace the "+ 2" with "+ 3" or "+ 4"?
+	if len(j.stack)+2 > maxLen {
+		return f
+	}
```
kortschak, line 224: If this condition is added to circuit and the order of the graph is passed in, then this method addition would not be required, true?

graph/topo/johnson_cycles.go
```diff
@@ -107,6 +215,47 @@ func (j *johnson) circuit(v int) bool {
 	return f
 }
 
+// circuitMaxLen is the CIRCUIT sub-procedure in the paper but produces circuits up to a maximum length.
+func (j *johnson) circuitMaxLen(v int, maxLen int) bool {
+	f := false
+	// TODO: can we replace the "+ 2" with "+ 3" or "+ 4"?
```
kortschak, line 221: Can you clarify this? It wouldn't be enough information for me to come back to in the future.

graph/topo/johnson_cycles_test.go
```diff
@@ -115,3 +115,406 @@ func TestDirectedCyclesIn(t *testing.T) {
...
+			3: linksTo(4),
+			4: linksTo(2, 5),
+			6: linksTo(3, 5),
+			7: linksTo(0, 6),
+		},
+		want: [][][]int64{
+			0: {{0, 1, 7, 0}},
+			1: {{1, 7, 0, 1}},
+			2: {{2, 3, 4, 2},
+				{2, 6, 3, 4, 2}},
```
kortschak, line 137: I'd prefer these kinds of cases to be on a single line; they're not so long that they need two.

kortschak (commented):
Thank you for sending this.

Can you provide a provenance for these additions?

Also, I'm interested to know how you would intend using these. I have suspicions, but it would be nice to know.

graph/topo/johnson_cycles.go
wardbradt: yes, that is correct

kortschak: If that doesn't have a significant performance impact, I would prefer it done that way.

graph/topo/johnson_cycles_test.go
```diff
@@ -115,3 +115,406 @@ func TestDirectedCyclesIn(t *testing.T) {
...
+			2: linksTo(3, 6),
+			3: linksTo(4),
+			4: linksTo(2, 5),
+			6: linksTo(3, 5),
+			7: linksTo(0, 6),
+		},
+		wantMap: map[int][][]int64{
+			3: nil,
+			4: {{0, 1, 7, 0},
+				{2, 3, 4, 2}},
```
kortschak, line 264: This test is failing:
```
--- FAIL: TestDirectedCyclesOfMaxLen (0.00s)
    johnson_cycles_test.go:349: unexpected johnson result for 0:
        	got:[][]int64{[]int64{0, 1, 7, 0}}
        	want:[][]int64{[]int64{0, 1, 7, 0}, []int64{2, 3, 4, 2}}
FAIL
```

graph/topo/johnson_cycles.go
wardbradt: If you don't mind, a question I have is if and when this conditional is true https://github.com/gonum/gonum/blob/eae99adef421605fe44c4cab6e45a17e7e846560/graph/topo/johnson_cycles.go#L53. I know that it is in the original paper as `s := least vertex in V_k;` but I don't see how `j.adjacent.leastVertexIndex()` could return a number less than `j.s`.

kortschak: Yes, it can be true, both as an equality (which doesn't matter) and as `s > j.s`. This happens when the least index is not in V_k.

graph/topo/johnson_cycles.go
wardbradt: > Thank you for sending this.

Sure, thanks for working on the package. It's really helpful and well-written.

> Can you provide a provenance for these additions?

I am not 100% sure what you exactly mean by provenance. I think the answer is this pull request (I authored the modifications). However, I got the idea for `DirectedCyclesOfMaxLen` from someone else's fork of this package.

> Also, I'm interested to know how you would intend using these. I have suspicions, but it would be nice to know.

My intended use case is for a high performance application where post-processing the output of `DirectedCyclesIn` is too slow.

kortschak: >I am not 100% sure what you exactly mean by provenance. I think the answer is this pull request (I authored the modifications). However, I got the idea for `DirectedCyclesOfMaxLen` from someone else's fork of this package.

Yeah. I was just wondering if there was a paper.

>My intended use case is for a high performance application where post-processing the output of `DirectedCyclesIn` is too slow.

If it's not sensitive, I was wondering what the use is. I see crypto and finance in your repos, so I was guessing in that context, either algotrading or blockchain analysis.

graph/topo/johnson_cycles.go
wardbradt: It can be ignored. The TODO was a note to myself to think about replacing `len(j.stack)+2` with `len(j.stack)+3`. But I have realized that the latter would not work. I will delete this comment

graph/topo/johnson_cycles_test.go
wardbradt: Very strange. It did not fail on my machine. I will look into it

graph/topo/johnson_cycles.go
wardbradt: It would not. I can do that

kortschak:
I've had more of a think about the API here. If we have two function, the original and then one new one `DirectedCyclesRestrictedIn(g graph.Directed, max int, with ...int64) [][]graph.Node`then all the backing code can be shared since `DirectedCyclesIn` is essentially just the case where max is the order of the graph and with is all the nodes. It's not quite as simple as just passing to the restricted case, but it's pretty close, and it massively reduces the API surface. WDYT?

vladimir-ch:

PR #1714, opened 2021-09-02
lapack/{gonum,testlapack}: clean up Dorgr2 and its test

vladimir-ch:
PTAL

cc @soypat

### DIFF
--- a/lapack/gonum/dorgr2.go
+++ b/lapack/gonum/dorgr2.go
@@ -9,39 +9,34 @@ import (
 	"gonum.org/v1/gonum/blas/blas64"
 )
 
-// Dorgr2 generates an m×n real matrix Q with orthonormal rows,
-// which is defined as the last m rows of a product of k elementary
-// reflectors of order n
+// Dorgr2 generates an m×n real matrix Q with orthonormal rows, which is defined
+// as the last m rows of a product of k elementary reflectors of order n
 //  Q = H_0 * H_1 * ... * H_{k-1}
 // as returned by Dgerqf.
 //
-// Each entry of tau contains the scalar factor of the elementary reflector H_i
-// and the length of tau must be at least k. The length of work must be at least m.
-// a is a matrix of dimensions (n,lda). On entry the [m-k+i-1]-th row must contain (counting from zero)
-// the vector which defines the elementary reflector H_i, for i = 0,1,2,...,k-1, as
-// returned by Dgerqf in the last k rows of its array argument A.
-// On exit, the m×n matrix Q.
-//  n >= m >= k >= 0
+// On entry, the (m-k+i)-th row of A must contain the vector which defines the
+// elementary reflector H_i, for i = 0,1,...,k, as returned by Dgerqf. On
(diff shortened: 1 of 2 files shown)


lapack/gonum/dorgr2.go
```diff
@@ -54,16 +49,16 @@ func (impl Implementation) Dorgr2(m, n, k int, a []float64, lda int, tau, work [
...
-		for l := 0; l < m-k; l++ {
-			for j := 0; j < n; j++ {
-				a[l*lda+j] = 0
-			}
-			a[l*lda+n-m+l] = 1
+	// Initialise rows 0:m-k to rows of the unit matrix.
+	for l := 0; l < m-k; l++ {
+		for j := 0; j < n; j++ {
+			a[l*lda+j] = 0
 		}
```
kortschak, line 60: Is it worth helping the compiler optimise this?
```
row := a[l*lda:l*lda+n]
for j := range row {
    row[j] = 0
}
```

lapack/testlapack/dorgr2.go
```diff
@@ -35,60 +37,72 @@ func Dorgr2Test(t *testing.T, impl Dorgr2er) {
...
-		}
-	}
 	// Compute the matrix Q using Dorg2r.
-	impl.Dorgr2(m, n, k, a.Data, a.Stride, tau[m-k:m], work)
+	q := cloneGeneral(rq)
+	impl.Dorgr2(m, n, k, q.Data, q.Stride, tau[m-k:m], work)
+
 	if m == 0 {
 		return
 	}
```
kortschak, line 64: Does this happen this late for panic checks?

vladimir-ch: Yes, to check for panics but also for modifications if there are non-zero-sized const input slices.

lapack/gonum/dorgr2.go
vladimir-ch: Good idea, done.

lapack/testlapack/dorgr2.go
kortschak: > but also for modifications if there are non-zero-sized const input slices.

I'm confused about this. There are no test checks prior to this return, so if m is zero, the only thing that can be observed as a failure is a panic. What am I missing?

vladimir-ch: Ah, sorry for not being clear, I meant this comment in general. You're right that here I postpone the quick return only for a panic check. In other routines there might also be returned errors or inputs that I check for modifications between the call and the quick return (for example, Dtbtrs).

kortschak: OK. Thanks.

soypat:
Much better. LGTM. I'll definitely carry over some style decisions to future PRs. Thanks Vladimir

soypat (approved):
Much better. LGTM. I'll definitely carry over some style decisions to future PRs. Thanks Vladimir

vladimir-ch:
@kortschak Can you please approve too?

kortschak (approved):

PR #1715, opened 2021-09-02
github: remove kortschak from FUNDING.yml

kortschak:
Please take a look.


romanwerpachowski (approved):

kortschak:
The sponsorship heart remains. Probably we should remove the entire file until it's needed again.

PR #1719, opened 2021-09-13
lapack: add Dlag2 subroutine

soypat:
Please take a look. Process of solving #1651.

Dlag2 solves a 2x2 eigenvalue problem of the form `|A-xB| = 0` where `B` is upper triangular.

### DIFF
--- a/lapack/gonum/dlag2.go
+++ b/lapack/gonum/dlag2.go
@@ -1,0 +1,245 @@
+// Copyright ©2021 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package gonum
+
+import "math"
+
+// Dlag2 computes the eigenvalues of a 2×2 generalized eigenvalue
+// problem
+//  A - w*B,
+// with scaling as necessary to avoid over-/underflow.
+// The scaling factor "s" results in a modified eigenvalue equation
+//  s*A - w*B
+// where  s  is a non-negative scaling factor chosen so that  w,  w*B,
+// and  s*A  do not overflow and, if possible, do not underflow, either.
+//
+// B is an upper triangular ldb×2 matrix
+// On entry, the 2×2 upper triangular matrix B.  It is
+// assumed that the one-norm of B is less than 1/dlamchS.  The
+// diagonals should be at least sqrt(dlamchS) times the largest
+// element of B (in absolute value); if a diagonal is smaller
+// than that, then  +/- sqrt(dlamchS) will be used instead of
+// that diagonal.
+//
+// It is assumed that A's 1-norm is less than 1/safmin.  Entries less than
+// sqrt(safmin)*norm(A) are subject to being treated as zero.
+//
+// Dlag2 is an internal routine. It is exported for testing purposes.
+//
+// scale1 is used to avoid over-/underflow in the
+// eigenvalue equation which defines the first eigenvalue.  If
+// the eigenvalues are complex, then the eigenvalues are
+// ( wr1  +/-  wi*i ) / scale1  (which may lie outside the
+// exponent range of the machine), scale1=scale2, and scale1
+// will always be positive.  If the eigenvalues are real, then
+// the first (real) eigenvalue is  wr1 / scale1 , but this may
+// overflow or underflow, and in fact, scale1 may be zero or
+// less than the underflow threshold if the exact eigenvalue
+// is sufficiently large.
+//
+// scale2 is used to avoid over-/underflow in the
+// eigenvalue equation which defines the second eigenvalue.  If
+// the eigenvalues are complex, then scale2=scale1.  If the
+// eigenvalues are real, then the second (real) eigenvalue is
+// wr2 / scale2 , but this may overflow or underflow, and in
+// fact, scale2 may be zero or less than the underflow
+// threshold if the exact eigenvalue is sufficiently large.
+//
+// If the eigenvalue is real, then wr1 is scale1 times the
+// eigenvalue closest to the (1,1) element of A*B^{-1}.  If the
+// eigenvalue is complex, then wr1=wr2 is scale1 times the real
+// part of the eigenvalues.
+//
+// If the eigenvalue is real, then wr2 is scale2 times the
+// other eigenvalue.  If the eigenvalue is complex, then
+// wr1=wr2 is scale1 times the real part of the eigenvalues.
+//
+// If the eigenvalue is real, then wi is zero.  If the
+// eigenvalue is complex, then wi is scale1 times the imaginary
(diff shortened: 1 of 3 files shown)


lapack/gonum/dlag2.go
```diff
@@ -0,0 +1,245 @@
...
+		panic(shortA)
+	case len(b) < ldb+2:
+		panic(shortB)
+	}
+
+	const (
+		safmin = dlamchS
+		fuzzy1 = 1 + 1e-5
+	)
+	rtmin := math.Sqrt(safmin)
```
vladimir-ch, line 77: It's unfortunate that our `drtmin` is slightly different.

lapack/gonum/dlag2.go
```diff
@@ -0,0 +1,245 @@
...
+		wbig := shift + sum
+
+		// Compute smaller eigenvalue.
+		var wsmall, wdet float64
+		wsmall = shift + diff
+		if 0.5*math.Abs(wbig) > math.Max(math.Abs(wsmall), safmin) {
+			wdet = (a11*a22 - a12*a21) * (binv11 * binv22)
+			wsmall = wdet / wbig
+		}
+		// Choose (real) eigenvalue closest to 2,2 element of A*B**(-1) for WR1.
```
vladimir-ch, line 166:
```suggestion
		// Choose (real) eigenvalue closest to 2,2 element of A*B^{-1} for wr1.
```

lapack/testlapack/dlag2.go
```diff
@@ -0,0 +1,96 @@
...
+import (
+	"fmt"
+	"math"
+	"testing"
+
+	"golang.org/x/exp/rand"
+)
+
+type Dlag2er interface {
+	// DLAG2 computes the eigenvalues of a 2 x 2 generalized eigenvalue
```
vladimir-ch, line 16: Please remove this comment, there's no point in documenting these helper interfaces.

lapack/testlapack/dlag2.go
```diff
@@ -0,0 +1,96 @@
...
+	// than that, then  +/- sqrt(SAFMIN) will be used instead of
+	// that diagonal.
+	//
+	// Dlag2 is an internal routine. It is exported for testing purposes.
+	Dlag2(a []float64, lda int, b []float64, ldb int) (scale1, scale2, wr1, wr2, wi float64)
+}
+
+func Dlag2Test(t *testing.T, impl Dlag2er) {
+	rnd := rand.New(rand.NewSource(1))
+	for _, lda := range []int{2, 3, 4, 5, 6, 7, 30} {
```
vladimir-ch, line 39: There's no benefit from testing so many strides, 2 and say 5 is enough. On the other hand, you can add a for loop that runs for example 10-100 times to get some variability in the random test data.

lapack/testlapack/dlag2.go
```diff
@@ -0,0 +1,96 @@
...
+			dlag2Test(t, impl, rnd, lda, ldb)
+		}
+	}
+}
+
+func dlag2Test(t *testing.T, impl Dlag2er, rnd *rand.Rand, lda, ldb int) {
+	const tol = 1e-14
+	a := randomGeneral(2, 2, lda, rnd)
+	b := randomGeneral(2, 2, ldb, rnd)
+	b.Data[ldb] = 0 // Dlag2 requires b to be upper triangular
```
vladimir-ch, line 50: There is no requirement in the code that this element is set to zero, it's simply not referenced by Dlag2. The test will be in fact better if you leave it set to a random value in case our implementation does reference it by mistake.

lapack/testlapack/dlag2.go
```diff
@@ -0,0 +1,96 @@
...
+	a11 := a.Data[0]
+	a12 := a.Data[1]
+	a21 := a.Data[lda]
+	a22 := a.Data[lda+1]
+	b11 := b.Data[0]
+	b12 := b.Data[1]
+	b22 := b.Data[ldb+1]
+	name := fmt.Sprintf("a11=%.4v,a12=%.4v,a21=%.4v,a22=%.4v,b11=%.4v,b12=%.4v,b22=%.4v", a11, a12, a21, a22, b11, b12, b22)
```
vladimir-ch, line 58: I'm not sure if this is useful. If there's an error, having the elements printed to 4 decimal places will not help.

lapack/testlapack/dlag2.go
```diff
@@ -0,0 +1,96 @@
...
+	for _, lda := range []int{2, 3, 4, 5, 6, 7, 30} {
+		for _, ldb := range []int{2, 3, 4, 5, 6, 7, 30} {
+			dlag2Test(t, impl, rnd, lda, ldb)
+		}
+	}
+}
+
+func dlag2Test(t *testing.T, impl Dlag2er, rnd *rand.Rand, lda, ldb int) {
+	const tol = 1e-14
+	a := randomGeneral(2, 2, lda, rnd)
```
vladimir-ch, line 48: Generating A and B randomly is ok but I'm afraid it will not cover those parts of the code that handle under/overflow. Have you looked what the coverage of Dlag2 is which this test?

You should take a look at `TESTING/EIG/dget53.f` and how they check the eigenvalues. See also where in `TESTING/EIG` this function is being called and how the input is being generated. Maybe you can find how to generate a difficult input to Dlag2.

lapack/testlapack/dlag2.go
```diff
@@ -0,0 +1,96 @@
...
+		t.Errorf("%v: got a residual from |%0.4v*A - w1*B| > tol: %v", name, scale1, res1)
+	}
+	res2 := det2x2(scale2*a11-wr2*b11, scale2*a12-wr2*b12,
+		scale2*a21, scale2*a22-wr2*b22)
+	if math.Abs(res2) > tol {
+		t.Errorf("%v: got a residual from |s*A - w2*B| > tol: %v", name, res1)
+	}
+}
+
+// Solves complex determinant of
```
vladimir-ch, line 93:
```suggestion
// cmplxdet2x2 returns the determinant of
```
and please the body of the function on its own line.

lapack/gonum/dlag2.go
soypat: Should I change this to be drtmin?

vladimir-ch: No, it was just a comment.

lapack/testlapack/dlag2.go
soypat: I've added `dget53Test` to the test but I'm struggling reading the tests. Could we add these edge case tests to the backlog?

vladimir-ch: Yeah, no worries. I'll merge what you have here now and then take a look myself and probably submit a clean-up PR later.

vladimir-ch (approved):
Thanks.

PR #1720, opened 2021-09-18
ci: drop use of travis and coveralls

kortschak:
This is due to https://thehackernews.com/2021/09/travis-ci-flaw-exposes-secrets-of.html and the ongoing trouble that we have had with travis. We lose arm64 builds, so changes that impact arm64 (usually lapack and blas) will need to be tested locally where possible until someone else provides builders for this arch. 

Please take a look.


vladimir-ch (approved):

sbinet:
Just a drive-by comment: in the meantime, perhaps we could setup a qemu scaffolding to test arm?

kortschak:
Qemu has sometimes worked for Go builds for arm64, but at other times has not (quite catastrophically). If it works then yes, we should.

PR #1721, opened 2021-09-18
README: fix link to supported versions

kortschak:
Please take a look.

The link looks like an off-by-one, but is not; it points to the line "Versions of Go that..." and the version list.


sbinet (approved):

kortschak:
Eeep. Wrong syntax for line 15. I'll fix it later.

PR #1722, opened 2021-09-19
stat/distuv: improve Pareto.Rand performance, also adding Pareto.Rand benchmark

argusdusty:
Please take a look.

Just removes an unnecessary math.Log call.

benchstat:
```
name        old time/op  new time/op  delta
ParetoRand  38.0ns ± 8%  24.4ns ± 2%  -35.72%  (p=0.000 n=9+10)
```


romanwerpachowski (approved):
LGTM.

kortschak:
Thanks

PR #1723, opened 2021-09-26
ci: make cache works by removing comments

ylz-at:
The path is a multiline string, which includes comments in its value. This
means that the caching was not working at all, because it would include
the comment in the path. Here I'm moving the comments, which fixes the
caching so it works.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -56,29 +56,27 @@ jobs:
       with:
         go-version: ${{ matrix.go-version }}
 
-    - name: Cache-Go
-      uses: actions/cache@v1
-      
-      with:
-        # In order:
-        # * Module download cache
-        # * Build cache (Linux)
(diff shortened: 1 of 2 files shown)


kortschak:
Thank you. Would you please also send another PR adding yourself to AUTHORS and CONTRIBUTORS with the commit message "A+C: add Jinesi Yelizati".

kortschak (approved):
LGTM but waiting for A+C PR.

ylz-at:
Hi, I updated actions/cache to v2 as v1 doesn't work with multiline string.
And I moved `Checkout code` before `Cache-Go` as `Cache-Go` depends on `go.sum`.

[Have tested on my own repo](https://github.com/ylz-at/gonum/runs/3713059510?check_suite_focus=true), Cache-Go can work this time.

kortschak (approved):
Thanks

PR #1731, opened 2021-09-29
lapack: add Dlatdf

soypat:
Please take a look.

@vladimir-ch  There is something real fishy here. Like you said:
> This says that the slice rhs given to Dlaswp is too short. Based on the call above, rhs should be an (n-1)x1 matrix (that is, a column vector) with the stride ldz.
Though this is not the case! If rhs has stride ldz then this routine operates on non-acessible data. So rhs probably has stride 1 though I can't for the life of me figure out how those `Dlaswp` calls work.

### DIFF
--- a/lapack/gonum/dlatdf.go
+++ b/lapack/gonum/dlatdf.go
@@ -1,0 +1,161 @@
+package gonum
+
+import (
+	"math"
+
+	"gonum.org/v1/gonum/blas/blas64"
+	"gonum.org/v1/gonum/lapack"
+)
+
+// Contributors: Bo Kagstrom and Lars Westin,
+// Generalized Schur Methods with Condition Estimators for
+// Solving the Generalized Sylvester Equation, IEEE Transactions
+// on Automatic Control, Vol. 34, No. 7, July 1989, pp 745-751.
+
+// Dlatdf uses the LU factorization of the n×n matrix Z computed by Dgetc2 and
+// computes a contribution to the reciprocal Dif-estimate by solving
+//  Z * x = b for x
+// and choosing the rhs b such that the norm of x is as large as possible.
+// On entry rhs = b holds the contribution from earlier solved sub-systems, and on return rhs = x.
+//
+// The factorization of Z returned by Dgetc2 has the form
+//  Z = P*L*U*Q,
+// where P and Q are permutation matrices. L is lower triangular with
+// unit diagonal elements and U is upper triangular.
+//
+// On entry ipiv and jpiv are n length slices. On exit the pivot indices where row i has
+// been interchanged with ipiv[i] and column j interchanged with column jpiv[j].
+//
+// It must hold that n <= len(ipiv), n <= len(jpiv). rhs must have stride ldz and n number of accesible elements.
+// Dlatdf implicitly expects z matrix to be at most an 8×8 matrix. This is fulfilled by it's only caller, Dtgsy2.
+//
+// ijob info:
+// ijob==2: First compute an approximative null-vector e of Z using DGECON,
+// e is normalized and solve for Zx = +-e - f with the sign giving the greater value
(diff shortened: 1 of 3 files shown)


lapack/gonum/dlatdf.go
```diff
@@ -0,0 +1,161 @@
...
+//
+// Dlatdf is an internal routine. It is exported for testing purposes.
+func (impl Implementation) Dlatdf(ijob, n int, z []float64, ldz int, rhs []float64, rdsum, rdscal float64, ipiv, jpiv []int) (sum, scale float64) {
+	switch {
+	case n < 0:
+		panic(nLT0)
+	case ldz < max(1, n):
+		panic(badLdZ)
+	case len(rhs) < (n-1)*ldz:
+		// rhs is expected to have stride ldz
```
vladimir-ch, line 58: The Fortran code might make it look like that but `rhs` is just a column n-vector represented by a continuous array (increment of 1). For the Dlaswp call the Fortran code needs to represent it as a matrix and so they need a stride even though it's actually not used (there is no 2. column to which stride to). They use ldz as a placeholder, they could use n just as well. But since Gonum is row-major, our code should use 1 as the stride for rhs and xm.

lapack/gonum/dlatdf.go
```diff
@@ -0,0 +1,161 @@
...
+		panic(shortRHS)
+	case len(ipiv) < n:
+		panic(badLenIpiv)
+	case len(jpiv) < n:
+		panic(badLenJpiv)
+	case len(z) > (ldz+1)*8:
```
vladimir-ch, line 64: The correct check should be that `if n > 8 then panic`. Then the length check for `z` can be the usual `if len(z) < (n-1)*ldz + n then panic`

lapack/gonum/dlatdf.go
```diff
@@ -0,0 +1,161 @@
...
+	xp := make([]float64, n)
+	if ijob == 2 {
+		// Compute approximate nullvector xm of Z when ijob == 2.
+		xm := make([]float64, n)
+		work := make([]float64, 4*n)
+		impl.Dgecon(lapack.MaxRowSum, n, z, ldz, 1, work, make([]int, n)) // iwork is unused.
+		bi.Dcopy(n, work[n:], 1, xm, 1)
+
+		// Compute rhs.
+		impl.Dlaswp(1, xm, ldz, 0, n-2, ipiv[:n-1], -1)
```
vladimir-ch, line 80:
```suggestion
		impl.Dlaswp(1, xm, 1, 0, n-2, ipiv[:n-1], -1)
```

lapack/gonum/dlatdf.go
```diff
@@ -0,0 +1,161 @@
...
+		}
+
+		// Compute sum of squares.
+		rdscal, rdsum = impl.Dlassq(n, rhs, 1, rdscal, rdsum)
+		return rdsum, rdscal
+	}
+
+	// Apply permutations IPIV to RHS
+	// If ijob != 2 uses look-ahead strategy.
+	impl.Dlaswp(1, rhs, ldz, 0, n-2, ipiv[:n-1], 1)
```
vladimir-ch, line 99:
```suggestion
	impl.Dlaswp(1, rhs, 1, 0, n-2, ipiv[:n-1], 1)
```

soypat:
@vladimir-ch Thank you for taking the time to explain what was off in the implementation. I have applied your suggestions to the code. 

Please take a look at the first error check lines of Dlatdf, I have added a few new checks. 

I have not completed the tests as I am having a hard time replicating the conditions given in `Dtgsy2` (I suspect, not sure about this either). What about testing this routine via `Dtgsy2` (which I've already implemented)?

soypat:
Sounds good to me. Here's the PR.

Issue #1725, opened 2021-09-27
Constrained Optimization (especially lower+upper bounds)


JulianCologne:
Hello,

Coming from Python (scipy.optimize.minimize) and MATLAB (fmincon), I can't find a way to specify bounds (and other constraints), which to me are very essential parts of real world application optimization problem.

Python / Scipy:
```python
res = minimize(opti_fun, x0, bounds=mybounds, constraints=myconstraints)
```

Matlab:
```matlab
res = fmincon(opti_fun, x0, A, b, Aeq, beq, lowerbound, upperbound)
```

Looking through the documentation, I could not find that this is implemented.
Is this something that is being worked on? Or is it planned? Or is it too complicated / time consuming at the moment? Or did I just miss something here?

I would love to see this in go!

vladimir-ch:
> I would love to see this in go!

Yes, we'd love that too. I'm aware that most practical optimization problems are constrained. Unfortunately, implementing a constrained solver is a big task that requires concentrated effort and our manpower and free time are both severely limited. I don't see it being added soon but we will surely be happy to welcome any help ;-)

Maybe L-BFGS-B would be a good starting point, it's an extension of a solver that we already have, so the task is potentially not so daunting but already a big step forward.

JulianCologne:
Thanks for the response.
You are absolutely right, this is a big job. Maybe I find some time in the future and can contribute to it. Good tip with L-BFGS-B, I'll take a look at that. Thanks!

sbinet:
as I noted in https://github.com/gonum/gonum/issues/354#issuecomment-395685405, a stop-gap solution (that has been used for decades in high energy physics) is to apply a parameter change that take into account the bounds:

```
P_internal = arcsin (2 * (P_external - a) / (b - a) - 1)

P_external = a + (b-a)/2 (sin P_internal + 1)
```

- https://cds.cern.ch/record/2296388/files/minuit.pdf (page 2)

JulianCologne:
@sbinet Thanks for the suggestion.
Looks very interesting, I will try that! ;)

crhntr:
I read the [Nelder Mead implementation in nlopt and noticed how it handles bounds](https://github.com/stevengj/nlopt/blob/d676b27aecb4f43ea4766af3c0c810dd56622953/src/algs/neldermead/nldrmd.c#L52-L77). Here is a branch with what I think it might look like added to the gonum implementation.
https://github.com/crhntr/gonum/tree/add-bounds-to-neldermead

PR #1734, opened 2021-10-05
add optional bounds to NelderMead

crhntr:
API changes are backwards compatible. Zero value (nil slice of Bound) behaves exactly as it did before.

I read the Nelder Mead implementation in nlopt and noticed how it handles bounds. Here is what I think it might look like added to the gonum implementation. 

This is a possible fix for: https://github.com/gonum/gonum/issues/1725

I did not add tests or TDD because this is/was an exploration.

### DIFF
--- a/optimize/neldermead.go
+++ b/optimize/neldermead.go
@@ -5,7 +5,6 @@
 package optimize
 
 import (
-	"fmt"
 	"math"
 	"sort"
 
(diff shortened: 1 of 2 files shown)


optimize/neldermead.go
```diff
@@ -69,6 +69,7 @@ type NelderMead struct {
 	Contraction     float64 // Contraction parameter (>0, <1)
 	Shrink          float64 // Shrink parameter (>0, <1)
 	SimplexSize     float64 // size of auto-constructed initial simplex
+	Bounds          []Bound // Bounds parameter if not nil, Bounds must have length equal to dim
```
romanwerpachowski, line 72: Consider: "Bounds parameter: if not nil, must have have length equal to dim".

romanwerpachowski (commented):
I think a simple unit test would be nice, e.g. that if you minimise x^2 on the range [1, 2] the solution is close to 1.

Issue #617, opened 2018-10-06
[proposal] graph: evaluate what the Graph API may look like in Go 2


mewmew:
### Background

Given the [recent generics proposal for Go 2](https://github.com/golang/proposal/blob/master/design/go2draft-generics-overview.md), it would be an idea to evaluate what the Graph API may look like if it took *full* advantage of the proposal in its current form.

### Proposal

The intention of this issue is partly to discuss how to Graph API may evolve in the future, to allow type safe handling of user-defined node types.

But more importantly than that, the intention of this issue and evaluation is to assess whether the current generics proposal for Go 2 (using contracts) is a good fit for defining a graph API. Both in terms of capabilities, but also in terms of readability and maintainability.

If, after discussions and evaluation there does not seem to be a good fit, then I suggest we summarize the discussions and outcomes of this issue and use that as constructive feedback to further improve the proposal.

### Potential impact of proposal

For Gonum Graph users, the main impact will be to get type safety when dealing with user-defined node types and avoid having to re-implement boilerplate.

For Go 2, the impact is potentially quite substantial, as it may help iron out edge cases in the current proposal that make it unfit for implementing Graphs (and other data structures and container formats) in terms of capability, readability and maintainability.

kortschak:
This seems like a good idea, but I think it should be based on what experiences we have with the current API: where it works, where it falls down etc. Then we can look at how generics/contracts could be used to help the parts that work and fix the parts that don't.

mewmew:
Here is an example from a control flow graph library I am writing:

https://github.com/graphism/exp/blob/master/cfa/cfa.go#L529

```go
// node asserts that the given node is a control flow graph node.
func node(n graph.Node) *cfg.Node {
	if n, ok := n.(*cfg.Node); ok {
		return n
	}
	panic(fmt.Errorf("invalid node type; expected *cfg.Node, got %T", n))
}

// edge asserts that the given edge is a control flow graph edge.
func edge(e graph.Edge) *cfg.Edge {
	if e, ok := e.(*cfg.Edge); ok {
		return e
	}
	panic(fmt.Errorf("invalid edge type; expected *cfg.Edge, got %T", e))
}
```

This is a fairly common practice for me, as I need to extract values from the nodes, and e.g. the `Nodes` method returns a slice of `graph.Node`s. The control flow analysis algorithms I'm currently using depend on a lot of values stored for each node:

https://github.com/graphism/exp/blob/master/cfg/graph.go#L317

```go
[...]

PR #1735, opened 2021-10-06
graph: update API to use type parameters

mewmew:
This *draft* PR is meant to get the ball rolling on *evaluating* what the graph API may look like if using (Go 1.18) type parameters. The intention of #617 is to to make the Gonum graph API safer (and easier) to use, hopefully reducing the need for run-time type assertions.

Note, thus far, no user-facing API has been changed in any significant way.

Updates #617.

**Edit:** note, 91c920cccff3d96d33eafb212096b92c11e0bd92 adds a copy of `container/heap` from golang/go@4a37a1d. This code is not meant to be submitted to Gonum in the final PR. As Go 1.18 (or Go 1.19) is released, there should be a type parameterized standard library `heap` package. The code is included for now to make `go test` pass.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module gonum.org/v1/gonum
 
-go 1.14
+go 1.18
 
 require (
 	golang.org/x/exp v0.0.0-20191002040644-a1355ae1e2c3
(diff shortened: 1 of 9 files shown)


kortschak:
The other place where I was thinking this could happen is in graph/internal/{linear,ordered,set}. Again, these are not user facing, but would reduce code lines.

shlomi-dr:
Hi, any updates on this topic? 

Would absolutely love to see graphs parametrized with Node and Edges types etc.

vladimir-ch:

PR #1741, opened 2021-10-24
lapack/gonum: add Dtgsy2 and its test

soypat:
Seeking to solve #1651. 

At the time of making this PR the `NoTrans` case passes only for n,m pairs: (1,1), (2,2),(3,1),(1,3). I've compared with the reference line-by-line a few times already and there is no detectable difference. Any guidance would be welcome.


vladimir-ch:
@soypat Do you still need the generalized eigenvalue solver? If not, I'd prefer to close this PR, especially if you can't make the test pass with confidence.

lapack/gonum/dtgsy2.go
```diff
@@ -0,0 +1,696 @@
...
+							a[is:], lda, rhs, mb, 1, c[js:], ldc)
+						bi.Dgemm(blas.NoTrans, blas.Trans, is, nb, mb, -1,
+							d[is:], ldd, rhs, mb, 1, f[js:], ldf)
+					}
+					if j < q {
+						k = mb * nb
+						bi.Dgemm(blas.Trans, blas.NoTrans, mb, n-je-1, nb, 1,
+							rhs[k:], mb, b[js*ldb+je+1:], ldb, 1, c[is*ldc+je+1:], ldc)
+						bi.Dgemm(blas.Trans, blas.NoTrans, mb, n-je-1, nb, 1,
+							rhs[k:], mb, e[js*lde+je+1:], lde, 1, f[is*ldf+je+1:], ldf)
```
soypat, line 420: @vladimir-ch I've found the issue lies with gonum's Dgemm implementation which can't emulate the full functionality of Fortran's Dgemm for non-transposed matrices due to the row-major and column major difference. If there were a column major Dgemm in gonum I could easily fix this PR.

soypat: https://groups.google.com/g/gonum-dev/c/AJFZHgTaxZ8/m/7ahKK2SCAQAJ

vladimir-ch: This is not an issue with gonum's Dgemm. It does exactly what the reference implementation does, only the matrices are represented differently. The issue is that Dtgsy2 does not adapt to this different representation correctly.

Besides, the test fails even when m=1,2, n=1,2 which doesn't have Dgemm in its code path, so there must be other issues as well.

soypat: So oddly I've managed to have Dtgsy2 passing for cases with practically any n,m combinations using `NoTrans` but when `ijob` is zero. If ijob is not zero a code path with Dlatdf at lines ~188, ~249, ~314, ~400 is taken.

soypat:
@vladimir-ch @kortschak 
I've debugged this DTGSY2 implementation side-by side with LAPACK's version for 1-dimensional matrices and am getting identical results for ijob=2 though the tests fail in gonum. I think that may be down to what ijob != 0 implies. According to documentation in  Dtgsy2:

> Dtgsy2 also (ijob >= 1) contributes to the computation in Dtgsyl of an upper bound on the separation between to matrix pairs. Then the input (A, D), (B, E) are sub-pencils of the matrix pair in Dtgsyl. See Dtgsyl for details.

Maybe we should defer testing of ijob != 0 to Dtsyl's implementation?

soypat:
So I've tried Dtgsy**l**  with matrix sizes of 4x4 and ijob==2 and I am getting identical results with LAPACK. Dtgsyl is calling Dtgsy2 in what test consider failure conditions, so I'm thinking maybe there's an implementation detail that we are not considering in tests.

soypat:
Closing in favor of https://github.com/gonum/gonum/pull/1891

PR #1742, opened 2021-10-25
graph/formats/rdf: work around unhelpful mechanical lawyer

kortschak:
This works around [this problem](https://pkg.go.dev/gonum.org/v1/gonum@v0.9.1-0.20210927225948-aad206551871/graph/formats/rdf). After merging it will look something like [this](https://pkg.go.dev/gonum.org/v1/gonum@v0.9.1-0.20211025210127-89aa43d448e3/graph/formats/rdf).

Included in v0.10 because documentation is important.

Please take a look.


kortschak:
*ping* This should be uncontroversial and doesn't require any graph package knowledge.

mattConn (approved):
The commit message is what got my attention.

kortschak:
Thanks

PR #1752, opened 2021-11-20
integrate/quad: Fix hermite initial guess bug

notimesea:
Eventually I found and fixed bug in integrate/quad package when tried to increase number of Hermite polynomial roots to 1000 (since panic when taking negative index of array)

I have also added test, but failed to find roots precise to 1e-14 tolerance, so added precise to 1e-13 one.

Refs:
https://github.com/gonum/gonum/blob/master/integrate/quad/internal/hermpts.m


integrate/quad/hermite.go
```diff
@@ -273,7 +273,7 @@ func (h Hermite) hermiteInitialGuess(i, n int) float64 {
 	}
 
 	// Use Gatteschi guesses in the second half where x is nearer to sqrt(n+0.5)
-	i = i + 1 - m
+	i = m - i - 1
```
kortschak, line 276: Can you make this `i = m - (i + 1)` for clarity?

kortschak (commented):
Thank you for sending this fix.

It looks OK to me, but I would like to wait for someone with more knowledge in the area.

Would you please also send another PR adding yourself to AUTHORS and CONTRIBUTORS with the PR comment "A+C: add Evgeny Savinov".

integrate/quad/hermite.go
notimesea: Yes, of course

notimesea:
@kortschak Hi, any update on this? Should someone else review this PR?

vladimir-ch:
Sorry for the delay, I will take a look.

vladimir-ch (approved):
In the reference Matlab code, the index reversing corresponds to this line in `hermpts.m`:
```
x_init_airy = real(x_init(end:-1:1));
```
so yes, this change LGTM. Good catch, thanks.

PR #1755, opened 2021-12-12
spatial/r3: add Mat type

### LINKED ISSUE #1754: spatial/r3: Adding a `r3.Mat` type
I work with rotation tensors and vectors of dimension 3. I found myself writing a package for matrix and r3.Vec interoperability.

### Background
I work with rotation tensors and vectors of dimension 3. I found myself writing a package for matrix and r3.Vec interoperability.
These tensors are used for attitude representation of a launch vehicle and are the industry standard way of representing this concept. One can use euler angles or quaternions to construct/decompose these matrices. See _Modeling and simulation of aerospace vehicle dynamics_ by P. Zipfel.

### Proposal
I propose the following addition to `spatial/r3`

<details>

```go
import (
	"fmt"
	"math"

	"gonum.org/v1/gonum/mat"
	"gonum.org/v1/gonum/spatial/r3"
)

const (
	badDim = "bad matrix dimensions"
	badIdx = "bad matrix index"
)

type Mat3 [3 * 3]float64

func NewMat3(val []float64) *Mat3 {
	if len(val) != 9 {
		if val == nil {
			return &Mat3{}
		}
		panic(badDim)
	}
	M := Mat3{}
	copy(M[:], val)
	return &M
}

// IdentityMat3 returns the Identity 3x3 matrix
func IdentityMat3() *Mat3 { return &Mat3{1, 0, 0, 0, 1, 0, 0, 0, 1} }

func (m Mat3) String() string {
	for i := range m {
		if math.Abs(m[i]) < 1e-6 {
			m[i] = 0
		}
	}
	return fmt.Sprintf("[%0.6g\t%0.6g\t%0.6g]\n[%0.6g\t%0.6g\t%0.6g]\n[%0.6g\t%0.6g\t%0.6g]",
		m[0], m[1], m[2], m[3], m[4], m[5], m[6], m[7], m[8])
}

func (m Mat3) CloneDense() *mat.Dense {
	M := mat.NewDense(3, 3, m[:])
	return M
	// for i := 0; i < 3; i++ {
	// 	for j := 0; j < 3; j++ {
	// 		M.Set(i, j, m.at(i, j))
	// 	}
	// }
	// return M
}

// CopyToMat3 copies elements from src Matrix to
// dst Mat3.
func CopyToMat3(dst *Mat3, src mat.Matrix) {
	r, c := src.Dims()
	if r != 3 || c != 3 {
		panic(badDim)
	}
	for i := 0; i < 3; i++ {
		for j := 0; j < 3; j++ {
			dst.set(i, j, src.At(i, j))
		}
	}
}

func (m *Mat3) Dims() (r, c int) { return 3, 3 }

func (m *Mat3) At(i, j int) float64 {
	if uint(i) > 2 || uint(j) > 2 {
		panic(badIdx)
	}
	return m.at(i, j)
}

func (m *Mat3) at(i, j int) float64 {
	return m[i*3+j]
}

func (m *Mat3) T() mat.Matrix { return mat.Transpose{Matrix: m} }
func (m *Mat3) Set(i, j int, v float64) {
	if uint(i) > 2 || uint(j) > 2 {
		panic(badIdx)
	}
	m.set(i, j, v)
}

// set sets element at position i,j without bounds checking
func (m *Mat3) set(i, j int, v float64) {
	m[i*3+j] = v
}

// Scale multiplies the elements of a by f, placing the result in the receiver.
//
// See the Scaler interface for more information.
func (m *Mat3) Scale(f float64, a mat.Matrix) {
	r, c := a.Dims()
	if r != 3 || c != 3 {
		panic(badDim)
	}
	for i := 0; i < 3; i++ {
		for j := 0; j < 3; j++ {
			m.set(i, j, f*a.At(i, j))
		}
	}
}

// Performs matrix multiplication on v:
//  result = M * v
func (m *Mat3) MulVec(v r3.Vec) (result r3.Vec) {
	result.X = v.X*m[0] + v.Y*m[1] + v.Z*m[2]
	result.Y = v.X*m[3] + v.Y*m[4] + v.Z*m[5]
	result.Z = v.X*m[6] + v.Y*m[7] + v.Z*m[8]
	return result
}

// Performs transposed matrix multiplication on v:
//  result = Mᵀ * v
func (m *Mat3) MulVecTrans(v r3.Vec) (result r3.Vec) {
	result.X = v.X*m[0] + v.Y*m[3] + v.Z*m[6]
	result.Y = v.X*m[1] + v.Y*m[4] + v.Z*m[7]
	result.Z = v.X*m[2] + v.Y*m[5] + v.Z*m[8]
	return result
}

// Skew returns the skew symmetric matrix (right hand system) of v.
//                  [0  -z  y]
//  Skew({x,y,z}) = [z   0 -x]
//                  [-y  x  0]
func Skew(v r3.Vec) (M *Mat3) {
	return &Mat3{
		0, -v.Z, v.Y,
		v.Z, 0, -v.X,
		-v.Y, v.X, 0,
	}
}

// Mul takes the matrix product of a and b, placing the result in the receiver.
// If the number of columns in a does not equal 3, Mul will panic.
func (m *Mat3) Mul(a, b mat.Matrix) {
	// TODO: add optimized operations for Mat3*Mat3
	// A, oka := a.(*Mat3)
	// B, okb := b.(*Mat3)
	// if oka && okb {
	// 	m.mul(A, B)
	// 	return
	// }
	mat.NewDense(3, 3, m[:]).Mul(a, b)
}

// Subtracts b from a and stores result in receiver. Sub will panic if the two matrices do not have the same shape.
func (m *Mat3) Sub(a, b mat.Matrix) {
	mat.NewDense(3, 3, m[:]).Sub(a, b)
}

// Add adds a and b element-wise, placing the result in the receiver. Add will panic if the two matrices do not have the same shape.
func (m *Mat3) Add(a, b mat.Matrix) {
	mat.NewDense(3, 3, m[:]).Add(a, b)
}

func (m *Mat3) VecRow(i int) r3.Vec {
	if uint(i) > 2 {
		panic(badIdx)
	}
	return r3.Vec{X: m.at(i, 0), Y: m.at(i, 1), Z: m.at(i, 2)}
}

func (m *Mat3) VecCol(j int) r3.Vec {
	if uint(j) > 2 {
		panic(badIdx)
	}
	return r3.Vec{X: m.at(0, j), Y: m.at(1, j), Z: m.at(2, j)}
}
```

</details>

Above is a working example of a 3x3 matrix. Of course with time the routines could be optimized for the special case of 3x3 arguments to Add, Mul, etc. Let me know what you think!

`Mat` implements the `mat.Matrix` interface. Maybe it is worth having `r3.Vec` implement the `Vector` interface so that there is better gonum package interoperability.

soypat:
Please take a look. Resolves  #1754.

Sidenote: if we switched `Mat`s data field to `data[3*3]` `RawMatrix` could be implemented more easily if that were of interest.

### DIFF
--- a/spatial/r3/mat.go
+++ b/spatial/r3/mat.go
@@ -1,0 +1,250 @@
+// Copyright ©2021 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package r3
+
+import (
+	"unsafe"
+
+	"gonum.org/v1/gonum/blas/blas64"
+	"gonum.org/v1/gonum/mat"
+)
+
+const (
+	badDim = "bad matrix dimensions"
+	badIdx = "bad matrix index"
+)
+
+// Mat represents a 3×3 matrix. Useful for rotation matrices and such.
+type Mat struct {
+	data *[3][3]float64
+}
+
+var _ mat.Matrix = (*Mat)(nil)
+
+// NewMat returns a new 3×3 matrix Mat type and populates its elements
+// with values passed as argument in row-major form. If val argument
+// is nil then NewMat returns a matrix filled with zeros.
+func NewMat(val []float64) *Mat {
+	if len(val) != 9 {
+		if val == nil {
+			return &Mat{data: new([3][3]float64)}
+		}
+		panic(badDim)
+	}
+	m := Mat{}
+	m.setBackingSlice(val)
+	return &m
+}
+
+// Dims returns the number of rows and columns of this matrix.
+// This method will always return 3×3 for a Mat.
+func (m *Mat) Dims() (r, c int) { return 3, 3 }
+
+// At returns the value of a matrix element at row i, column j.
+// At expects indices in the range [0,2].
+// It will panic if i or j are out of bounds for the matrix.
+func (m *Mat) At(i, j int) float64 {
+	return m.data[i][j]
+}
+
+// Set sets the element at row i, column j to the value v.
+func (m *Mat) Set(i, j int, v float64) {
+	m.data[i][j] = v
+}
+
+// T returns the transpose of Mat. Changes in the receiver will be reflected in the returned matrix.
+func (m *Mat) T() mat.Matrix { return mat.Transpose{Matrix: m} }
+
+// RawMatrix returns the blas representation of the matrix with the backing data of this matrix.
+// Changes to the returned matrix will be reflected in the receiver.
+func (m *Mat) RawMatrix() blas64.General {
+	return blas64.General{Rows: 3, Cols: 3, Data: m.backingSlice(), Stride: 3}
+}
+
+// Eye returns the 3×3 Identity matrix
+func Eye() *Mat {
+	return &Mat{data: &[3][3]float64{
+		{1, 0, 0},
+		{0, 1, 0},
+		{0, 0, 1},
+	}}
+}
+
+// Scale multiplies the elements of a by f, placing the result in the receiver.
+//
+// See the mat.Scaler interface for more information.
+func (m *Mat) Scale(f float64, a mat.Matrix) {
+	r, c := a.Dims()
+	if r != 3 || c != 3 {
+		panic(badDim)
+	}
+	for i := 0; i < 3; i++ {
+		for j := 0; j < 3; j++ {
+			m.Set(i, j, f*a.At(i, j))
+		}
+	}
+}
+
+// Performs matrix multiplication on v:
+//  result = M * v
+func (m *Mat) MulVec(v Vec) Vec {
+	return Vec{
+		X: v.X*m.At(0, 0) + v.Y*m.At(0, 1) + v.Z*m.At(0, 2),
+		Y: v.X*m.At(1, 0) + v.Y*m.At(1, 1) + v.Z*m.At(1, 2),
+		Z: v.X*m.At(2, 0) + v.Y*m.At(2, 1) + v.Z*m.At(2, 2),
+	}
+}
+
+// Performs transposed matrix multiplication on v:
+//  result = Mᵀ * v
+func (m *Mat) MulVecTrans(v Vec) Vec {
+	return Vec{
+		X: v.X*m.At(0, 0) + v.Y*m.At(1, 0) + v.Z*m.At(2, 0),
+		Y: v.X*m.At(0, 1) + v.Y*m.At(1, 1) + v.Z*m.At(2, 1),
+		Z: v.X*m.At(0, 2) + v.Y*m.At(1, 2) + v.Z*m.At(2, 2),
+	}
+}
+
+// Skew returns the 3×3 skew symmetric matrix (right hand system) of v.
+//                  ⎡ 0 -z  y⎤
+//  Skew({x,y,z}) = ⎢ z  0 -x⎥
+//                  ⎣-y  x  0⎦
+func Skew(v Vec) (M *Mat) {
+	return &Mat{data: &[3][3]float64{
+		{0, -v.Z, v.Y},
+		{v.Z, 0, -v.X},
+		{-v.Y, v.X, 0},
+	}}
+}
+
+// Mul takes the matrix product of a and b, placing the result in the receiver.
+// If the number of columns in a does not equal 3, Mul will panic.
+func (m *Mat) Mul(a, b mat.Matrix) {
+	ra, ca := a.Dims()
+	rb, cb := b.Dims()
+	switch {
+	case ra != 3:
+		panic(badDim)
+	case cb != 3:
+		panic(badDim)
+	case ca != rb:
+		panic(badDim)
+	}
+	if ca != 3 {
+		// General matrix multiplication for the case where the inner dimension is not 3.
+		t := mat.NewDense(3, 3, m.backingSlice())
+		t.Mul(a, b)
+		return
+	}
+
+	a00 := a.At(0, 0)
+	b00 := b.At(0, 0)
+	a01 := a.At(0, 1)
+	b01 := b.At(0, 1)
+	a02 := a.At(0, 2)
+	b02 := b.At(0, 2)
+	a10 := a.At(1, 0)
+	b10 := b.At(1, 0)
+	a11 := a.At(1, 1)
+	b11 := b.At(1, 1)
+	a12 := a.At(1, 2)
+	b12 := b.At(1, 2)
+	a20 := a.At(2, 0)
+	b20 := b.At(2, 0)
+	a21 := a.At(2, 1)
+	b21 := b.At(2, 1)
+	a22 := a.At(2, 2)
+	b22 := b.At(2, 2)
+	m.data[0][0] = a00*b00 + a01*b10 + a02*b20
+	m.data[0][1] = a00*b01 + a01*b11 + a02*b21
+	m.data[0][2] = a00*b02 + a01*b12 + a02*b22
+	m.data[1][0] = a10*b00 + a11*b10 + a12*b20
+	m.data[1][1] = a10*b01 + a11*b11 + a12*b21
+	m.data[1][2] = a10*b02 + a11*b12 + a12*b22
+	m.data[2][0] = a20*b00 + a21*b10 + a22*b20
+	m.data[2][1] = a20*b01 + a21*b11 + a22*b21
+	m.data[2][2] = a20*b02 + a21*b12 + a22*b22
+}
+
+// CloneFrom makes a copy of a into the receiver m.
+// Mat expects a 3×3 input matrix.
+func (m *Mat) CloneFrom(a mat.Matrix) {
+	r, c := a.Dims()
+	if r != 3 || c != 3 {
+		panic(badDim)
+	}
+	for i := 0; i < 3; i++ {
+		for j := 0; j < 3; j++ {
+			m.Set(i, j, a.At(i, j))
+		}
+	}
+}
+
+// Sub subtracts the matrix b from a, placing the result in the receiver.
+// Sub will panic if the two matrices do not have the same shape.
+func (m *Mat) Sub(a, b mat.Matrix) {
+	if r, c := a.Dims(); r != 3 || c != 3 {
+		panic(badDim)
+	}
+	if r, c := b.Dims(); r != 3 || c != 3 {
+		panic(badDim)
+	}
+
+	m.data[0][0] = a.At(0, 0) - b.At(0, 0)
+	m.data[0][1] = a.At(0, 1) - b.At(0, 1)
+	m.data[0][2] = a.At(0, 2) - b.At(0, 2)
+	m.data[1][0] = a.At(1, 0) - b.At(1, 0)
+	m.data[1][1] = a.At(1, 1) - b.At(1, 1)
+	m.data[1][2] = a.At(1, 2) - b.At(1, 2)
+	m.data[2][0] = a.At(2, 0) - b.At(2, 0)
+	m.data[2][1] = a.At(2, 1) - b.At(2, 1)
+	m.data[2][2] = a.At(2, 2) - b.At(2, 2)
+}
+
+// Add adds a and b element-wise, placing the result in the receiver. Add will panic if the two matrices do not have the same shape.
+func (m *Mat) Add(a, b mat.Matrix) {
+	if r, c := a.Dims(); r != 3 || c != 3 {
+		panic(badDim)
+	}
+	if r, c := b.Dims(); r != 3 || c != 3 {
+		panic(badDim)
+	}
+
+	m.data[0][0] = a.At(0, 0) + b.At(0, 0)
+	m.data[0][1] = a.At(0, 1) + b.At(0, 1)
+	m.data[0][2] = a.At(0, 2) + b.At(0, 2)
+	m.data[1][0] = a.At(1, 0) + b.At(1, 0)
+	m.data[1][1] = a.At(1, 1) + b.At(1, 1)
+	m.data[1][2] = a.At(1, 2) + b.At(1, 2)
+	m.data[2][0] = a.At(2, 0) + b.At(2, 0)
+	m.data[2][1] = a.At(2, 1) + b.At(2, 1)
+	m.data[2][2] = a.At(2, 2) + b.At(2, 2)
+}
+
+// VecRow returns the elements in the ith row of the receiver.
+func (m *Mat) VecRow(i int) Vec {
+	if i > 2 {
+		panic(badIdx)
+	}
+	return Vec{X: m.At(i, 0), Y: m.At(i, 1), Z: m.At(i, 2)}
+}
+
+// VecCol returns the elements in the jth column of the receiver.
+func (m *Mat) VecCol(j int) Vec {
+	if j > 2 {
+		panic(badIdx)
+	}
+	return Vec{X: m.At(0, j), Y: m.At(1, j), Z: m.At(2, j)}
+}
+
+// setBackingSlice requires unsafe.
+func (m *Mat) setBackingSlice(vals []float64) {
+	m.data = (*[3][3]float64)(unsafe.Pointer(&vals[0]))
+}
+
+// backingSlice requires unsafe.
+func (m *Mat) backingSlice() []float64 {
+	return (*[9]float64)(unsafe.Pointer(m.data))[:]
+}
--- a/spatial/r3/mat_test.go
+++ b/spatial/r3/mat_test.go
@@ -1,0 +1,103 @@
+// Copyright ©2021 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package r3
+
+import (
+	"math"
+	"testing"
+
+	"golang.org/x/exp/rand"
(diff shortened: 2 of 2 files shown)


spatial/r3/mat.go
```diff
@@ -0,0 +1,185 @@
...
+	// A, oka := a.(*Mat3)
+	// B, okb := b.(*Mat3)
+	// if oka && okb {
+	// 	m.mul(A, B)
+	// 	return
+	// }
+	d := mat.NewDense(3, 3, nil)
+	d.Mul(a, b)
+	m.CloneFrom(d)
+}
```
kortschak, line 141: https://go.dev/play/p/iUkCC5cF_1N
```
func (m *Mat) Mul(a, b mat.Matrix) {
	if r, c := a.Dims(); r != 3 || c != 3 {
		panic(badDim)
	}
	if r, c := b.Dims(); r != 3 || c != 3 {
		panic(badDim)
	}

	a00 = a.At(0, 0); b00 = b.At(0, 0)
	a01 = a.At(0, 1); b01 = b.At(0, 1)
	a02 = a.At(0, 2); b02 = b.At(0, 2)
	a10 = a.At(1, 0); b10 = b.At(1, 0)
	a11 = a.At(1, 1); b11 = b.At(1, 1)
	a12 = a.At(1, 2); b12 = b.At(1, 2)
	a20 = a.At(2, 0); b20 = b.At(2, 0)
	a21 = a.At(2, 1); b21 = b.At(2, 1)
	a22 = a.At(2, 2); b22 = b.At(2, 2)
	m.data[0][0] = a00*b00 + a01*b10 + a02*b20
	m.data[0][1] = a00*b01 + a01*b11 + a02*b21
	m.data[0][2] = a00*b02 + a01*b12 + a02*b22
	m.data[1][0] = a10*b00 + a11*b10 + a12*b20
	m.data[1][1] = a10*b01 + a11*b11 + a12*b21
	m.data[1][2] = a10*b02 + a11*b12 + a12*b22
	m.data[2][0] = a20*b00 + a21*b10 + a22*b20
	m.data[2][1] = a20*b01 + a21*b11 + a22*b21
	m.data[2][2] = a20*b02 + a21*b12 + a22*b22
}
```
With this, the type assertion to check whether `a` and `b` are `*Mat` probably doesn't pay (it might, but the direct arithmetic and 3× reduction is indexing is paying more already).

spatial/r3/mat.go
```diff
@@ -0,0 +1,185 @@
...
+		j := iv % 3
+		m.set(i, j, a.At(i, j))
+	}
+}
+
+// Subtracts b from a and stores result in receiver. Sub will panic if the two matrices do not have the same shape.
+func (m *Mat) Sub(a, b mat.Matrix) {
+	d := mat.NewDense(3, 3, nil)
+	d.Sub(a, b)
+	m.CloneFrom(d)
```
kortschak, line 161: As for mul, these should be generated.

spatial/r3/mat.go
```diff
@@ -0,0 +1,185 @@
...
+}
+
+// Add adds a and b element-wise, placing the result in the receiver. Add will panic if the two matrices do not have the same shape.
+func (m *Mat) Add(a, b mat.Matrix) {
+	d := mat.NewDense(3, 3, nil)
+	d.Add(a, b)
+	m.CloneFrom(d)
```
kortschak, line 168: Same here.

spatial/r3/mat.go
```diff
@@ -0,0 +1,185 @@
...
+		}
+		panic(badDim)
+	}
+	M := Mat{}
+	for iv := 0; iv < 9; iv++ {
+		i := iv / 3
+		j := iv % 3
+		M.data[i][j] = val[iv]
+	}
+	return &M
```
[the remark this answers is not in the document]
soypat: Yeah I'm not fond of including unsafe in the main implementation. I know some Go based compilers break with it such as gopherjs. I'll see if I can split it up with build tags.

spatial/r3/mat.go
soypat: Copy that

spatial/r3/mat.go
soypat: Hmmmm, shouldn't we let the user multiply matrices which are not both necessarily 3x3?

kortschak: True. Make this the fast case when both a 3×3. Sorry.

spatial/r3/mat.go
```diff
@@ -0,0 +1,185 @@
...
+func (m *Mat) CloneFrom(a mat.Matrix) {
+	r, c := a.Dims()
+	if r != 3 || c != 3 {
+		panic(badDim)
+	}
+	for iv := 0; iv < 9; iv++ {
+		i := iv / 3
+		j := iv % 3
+		m.set(i, j, a.At(i, j))
+	}
```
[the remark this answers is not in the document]
soypat: good call 

```
BenchmarkCloneNested-8   	44806436	        27.63 ns/op	       0 B/op	       0 allocs/op
BenchmarkCloneMod-8      	35851317	        33.52 ns/op	       0 B/op	       0 allocs/op
```

soypat:
math.Rand is disallowed?

kortschak:
Use golang.org/x/exp/rand.

spatial/r3/mat.go
kortschak: https://go.dev/play/p/L4c-TpDOcbJ

spatial/r3/mat.go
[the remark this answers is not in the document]
kortschak: It's worth noting that gopherjs is not a target. It would be good to keep it, but if that blocks other things we are OK with losing it. The raw matrix code is only possible with unsafe if we use a 3×3 array. If it's important to have unsafe-free code, then perhaps we consider using a `[9]float64`. That decision needs benchmarking.

spatial/r3/mat.go
soypat: That is a really neat way to generate matrix operations. It would not have occurred to me to actually do this. Thanks!

spatial/r3/mat.go
```diff
@@ -0,0 +1,268 @@
...
+	m.data[1][1] = a11 - b11
+	m.data[1][2] = a12 - b12
+	m.data[2][0] = a20 - b20
+	m.data[2][1] = a21 - b21
+	m.data[2][2] = a22 - b22
+}
+
+// Add adds a and b element-wise, placing the result in the receiver. Add will panic if the two matrices do not have the same shape.
+func (m *Mat) Add(a, b mat.Matrix) {
+	a00 := a.At(0, 0)
```
kortschak, line 225: This needs a shape check. Also in `Sub`.

spatial/r3/mat.go
```diff
@@ -0,0 +1,268 @@
...
+	}
+}
+
+func (m *Mat) cloneMod(a mat.Matrix) {
+	for iv := 0; iv < 9; iv++ {
+		i := iv / 3
+		j := iv % 3
+		m.Set(i, j, a.At(i, j))
+	}
+}
```
kortschak, line 190: This is unused except in the benchmark. It can go away.

spatial/r3/mat.go
```diff
@@ -0,0 +1,268 @@
...
+	m.cloneNested(a)
+}
+
+func (m *Mat) cloneNested(a mat.Matrix) {
+	for i := 0; i < 3; i++ {
+		for j := 0; j < 3; j++ {
+			m.Set(i, j, a.At(i, j))
+		}
+	}
+}
```
kortschak, line 182: Move this up into `CloneFrom`.

spatial/r3/mat.go
```diff
@@ -0,0 +1,268 @@
...
+	result.X = v.X*m.At(0, 0) + v.Y*m.At(1, 0) + v.Z*m.At(2, 0)
+	result.Y = v.X*m.At(0, 1) + v.Y*m.At(1, 1) + v.Z*m.At(2, 1)
+	result.Z = v.X*m.At(0, 2) + v.Y*m.At(1, 2) + v.Z*m.At(2, 2)
+	return result
+}
+
+// Skew returns the 3×3 skew symmetric matrix (right hand system) of v.
+//                  [ 0 -z  y]
+//  Skew({x,y,z}) = [ z  0 -x]
+//                  [-y  x  0]
```
sbinet, line 109:
```suggestion
//                  ⎡ 0 -z  y⎤
//  Skew({x,y,z}) = ⎢ z  0 -x⎥
//                  ⎣-y  x  0⎦
```

spatial/r3/mat.go
```diff
@@ -0,0 +1,268 @@
...
+	m.data[1][0] = a10*b00 + a11*b10 + a12*b20
+	m.data[1][1] = a10*b01 + a11*b11 + a12*b21
+	m.data[1][2] = a10*b02 + a11*b12 + a12*b22
+	m.data[2][0] = a20*b00 + a21*b10 + a22*b20
+	m.data[2][1] = a20*b01 + a21*b11 + a22*b21
+	m.data[2][2] = a20*b02 + a21*b12 + a22*b22
+}
+
+// CloneFrom clones the matrix a into the receiver m. Mat expects a 3×3
+// input matrix.
```
sbinet, line 167:
```suggestion
// CloneFrom makes a copy of a into the receiver m.
// Mat expects a 3×3 input matrix.
```

spatial/r3/mat.go
```diff
@@ -0,0 +1,268 @@
...
+	r, c := a.Dims()
+	if r != 3 || c != 3 {
+		panic(badDim)
+	}
+	m.cloneNested(a)
+}
+
+func (m *Mat) cloneNested(a mat.Matrix) {
+	for i := 0; i < 3; i++ {
+		for j := 0; j < 3; j++ {
```
sbinet, line 178: perhaps reverse the loops order? (for memory locality)

spatial/r3/mat.go
```diff
@@ -0,0 +1,268 @@
...
+
+func (m *Mat) cloneMod(a mat.Matrix) {
+	for iv := 0; iv < 9; iv++ {
+		i := iv / 3
+		j := iv % 3
+		m.Set(i, j, a.At(i, j))
+	}
+}
+
+// Subtracts b from a and stores result in receiver. Sub will panic if the two matrices do not have the same shape.
```
sbinet, line 192:
```suggestion
// Sub subtracts the matrix b from a, placing the result in the receiver.
// Sub will panic if the two matrices do not have the same shape.
```

spatial/r3/mat.go
```diff
@@ -0,0 +1,268 @@
...
+	m.data[1][1] = a11 + b11
+	m.data[1][2] = a12 + b12
+	m.data[2][0] = a20 + b20
+	m.data[2][1] = a21 + b21
+	m.data[2][2] = a22 + b22
+}
+
+// VecRow returns the elements in the ith row of the receiver.
+func (m *Mat) VecRow(i int) Vec {
+	if uint(i) > 2 {
```
sbinet, line 256: isn't the `uint` too cute for its own good?
if the index happens to be negative, `At` will panic with probably a better panic message.
```suggestion
	if i > 2 {
```

spatial/r3/mat.go
```diff
@@ -0,0 +1,268 @@
...
+		panic(badIdx)
+	}
+	return Vec{X: m.At(i, 0), Y: m.At(i, 1), Z: m.At(i, 2)}
+}
+
+// VecCol returns the elements in the jth column of the receiver.
+func (m *Mat) VecCol(j int) Vec {
+	if uint(j) > 2 {
```
sbinet, line 264: ditto

sbinet (commented):
I didn't look too closely to the math.

feel free to ignore some of my nitpicking comments, though :)

spatial/r3/mat.go
```diff
@@ -0,0 +1,268 @@
...
+// NewMat returns a new 3×3 matrix Mat type and populates it's elements
+// with values passed as argument in row-major form. If val argument
+// is nil then NewMat returns a matrix filled with zeros.
+func NewMat(val []float64) *Mat {
+	if len(val) != 9 {
+		if val == nil {
+			return &Mat{}
+		}
+		panic(badDim)
+	}
```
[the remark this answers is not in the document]
kortschak: This is not right; nil and len==0 are semantically different in the mat API — nil is don't care, 0 is a length which is not valid here.

spatial/r3/mat.go
kortschak: This is iterating in the correct orientation for row major.

spatial/r3/mat.go
```diff
@@ -0,0 +1,268 @@
...
+// Scale multiplies the elements of a by f, placing the result in the receiver.
+//
+// See the mat.Scaler interface for more information.
+func (m *Mat) Scale(f float64, a mat.Matrix) {
+	r, c := a.Dims()
+	if r != 3 || c != 3 {
+		panic(badDim)
+	}
+	for i := 0; i < 3; i++ {
+		for j := 0; j < 3; j++ {
```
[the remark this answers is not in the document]
kortschak: This is correct.

spatial/r3/mat.go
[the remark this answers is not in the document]
sbinet: right.

what about:
```go
switch {
case val == nil:
    return &Mat{}
case len(val) != 9:
    panic(badDim)
}
```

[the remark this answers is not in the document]
kortschak: Why? It's not really clearer and is a minusculely greater amount of work.

spatial/r3/mat_safe.go
```diff
@@ -0,0 +1,23 @@
...
+		for j := 0; j < 3; j++ {
+			m.Set(i, j, vals[i*3+j])
+		}
+	}
+}
+
+// backingSlice requires unsafe.
+func (m *Mat) backingSlice() []float64 {
+	panic("not implemented without unsafe")
+}
```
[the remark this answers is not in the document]
soypat: It seems to not be possible to accomplish this since tests are run with -tags=safe flag. Is there a way to disable running tests in `r3` with safe?

spatial/r3/mat_unsafe.go
```diff
@@ -0,0 +1,21 @@
...
+// +build !safe
+
+package r3
+
+import "unsafe"
+
+const tagSafe = false
+
+func (m *Mat) setBackingSlice(vals []float64) {
+	copy(m.backingSlice(), vals)
```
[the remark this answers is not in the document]
soypat: I did not perform this change as requested since it is only valid from go 1.17+ and the go.mod specifies 1.14. I've instead taken the pointer from `&vals[0]`

spatial/r3/mat_safe.go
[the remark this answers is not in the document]
kortschak: Put `// +build !safe` in mat_test.go.

spatial/r3/mat_unsafe.go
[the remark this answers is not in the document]
kortschak: The go.mod needs updating, but that won't fix this issue yet. Please add a todo `// TODO(kortschak): Use array conversion with go1.16 is no longer supported.`

spatial/r3/mat.go
```diff
@@ -0,0 +1,235 @@
...
+	badDim = "bad matrix dimensions"
+	badIdx = "bad matrix index"
+)
+
+// Mat represents a 3×3 matrix. Useful for rotation matrices and such.
+type Mat struct {
+	data *[3][3]float64
+}
+
+var _ mat.Matrix = (*Mat)(nil)
```
kortschak, line 22: I'm wondering about whether we need to have a pointer on `Mat`'s methods. Formally we don't since the mutations are to the `data` field, but it makes it look like a `Mat` is immutable. I'm think I'm still leaning towards removing it because it reduces the amount of pointer walking required. It would be helpful to see the difference in the generated code between the two.

@sbinet What do you think?

spatial/r3/mat.go
```diff
@@ -0,0 +1,235 @@
...
+		}
+		panic(badDim)
+	}
+	m := Mat{}
+	m.setBackingSlice(val)
+	return &m
+}
+
+// Dims returns the number of rows and columns of this matrix.
+// This method will always return 3×3 for a r3.Mat.
```
kortschak, line 40: `r3.` prefix is not required within the package.

spatial/r3/mat.go
soypat: You do need a pointer receiver if you want to re-assign the backing slice if I'm not mistaken, as we are doing in `setBackingSlice`

kortschak: We can leave it as it is.

spatial/r3/mat_test.go
```diff
@@ -0,0 +1,106 @@
+// Copyright ©2021 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+//go:build !safe
+// +build !safe
```
kortschak, line 6: These shouldn't be here. Otherwise we are lying to people.

kortschak (approved):

spatial/r3/mat.go
sbinet: d'oh. I was probably already a bit confused by the flu...

PR #1760, opened 2022-01-05
mat: add example for polynomial regression with QR factorization

### LINKED ISSUE #1759: Would you mind writing a quick demo about "polyfit" and "polyval"? Thank you.
In matlab and numpy. There are build-in functions like "polyfit" and "polyval".
I searched some codes like https://rosettacode.org/wiki/Polynomial_regression#Library_gonum.2Fmatrix
But all of them are out of date (Use old library mat64).
I am not really good at math. I tried to understand QR matrix and wrote it myself but failed. Would you mind writing a quick demo about "polyfit" and "polyval" with QR matrix? Thank you for your help.

sbinet:
Fixes #1759.

Signed-off-by: Sebastien Binet <binet@cern.ch>

Please take a look.

### DIFF
--- a/mat/qr_example_test.go
+++ b/mat/qr_example_test.go
@@ -28,8 +28,7 @@ func ExampleQR_solveTo() {
 	var qr mat.QR
 	qr.Factorize(a)
 
-	const trans = false
-	err := qr.SolveTo(x, trans, b)
+	err := qr.SolveTo(x, false, b)
 	if err != nil {
 		log.Fatalf("could not solve QR: %+v", err)
 	}
--- a/mat/qr_example_test.go
+++ b/mat/qr_example_test.go
@@ -1,0 +1,50 @@
+// Copyright ©2022 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package mat_test
+
+import (
+	"fmt"
+	"log"
+
+	"gonum.org/v1/gonum/mat"
+)
+
+func ExampleQR_solveTo() {
+	var (
+		x = []float64{0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10}
+		y = []float64{1, 6, 17, 34, 57, 86, 121, 162, 209, 262, 321}
+
+		degree = 2
+
+		a = Vandermonde(x, degree+1)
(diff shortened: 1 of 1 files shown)


mat/qr_example_test.go
```diff
@@ -0,0 +1,50 @@
...
+
+import (
+	"fmt"
+	"log"
+
+	"gonum.org/v1/gonum/mat"
+)
+
+func ExampleQR_solveTo() {
+	var (
```
kortschak, line 15: This probably wants some lead in text.

mat/qr_example_test.go
```diff
@@ -0,0 +1,50 @@
...
+	}
+	fmt.Printf("%.3f\n", mat.Formatted(c))
+
+	// Output:
+	// ⎡1.000⎤
+	// ⎢2.000⎥
+	// ⎣3.000⎦
+}
+
+func Vandermonde(a []float64, d int) *mat.Dense {
```
kortschak, line 42: Is there a particular reason to use this rather than just demonstrating a solve with a product of two matrices? If this is needed it should be introduced in the lead in text so that readers aren't required to look up and understand the background for this in order to understand the QR solve API.

sbinet: no reason.
I was just implementing the Rosetta code example for QR factorization (that I had migrated from `github.com/gonum/matrix` to `gonum.org/v1/gonum/mat`)
https://rosettacode.org/wiki/Polynomial_regression#Library_gonum.2Fmatrix

mat/qr_example_test.go
```diff
@@ -0,0 +1,41 @@
...
+	var (
+		a = mat.NewDense(4, 2, []float64{0, 1, 1, 1, 1, 1, 2, 1})
+		b = mat.NewDense(4, 1, []float64{1, 0, 2, 1})
+		x = mat.NewDense(2, 1, nil)
+	)
+
+	var qr mat.QR
+	qr.Factorize(a)
+
+	const trans = false
```
kortschak, line 31: This will be rendered right next to the documentation, so I don't think this const is necessary.

kortschak (dismissed):

sbinet:
PTAL

kortschak (approved):
Thanks

PR #1766, opened 2022-02-02
lapack/lapack64: add Pstrf

vladimir-ch:
- lapack/gonum: extend docs for Dpstrf and Dpstf2
- lapack/lapack64: add Pstrf

Please take a look.

### DIFF
--- a/lapack/gonum/dpstf2.go
+++ b/lapack/gonum/dpstf2.go
@@ -17,10 +17,19 @@ import (
 // The factorization has the form
 //  Pᵀ * A * P = Uᵀ * U ,  if uplo = blas.Upper,
 //  Pᵀ * A * P = L  * Lᵀ,  if uplo = blas.Lower,
-// where U is an upper triangular matrix and L is lower triangular, and P is
-// stored as vector piv.
+// where U is an upper triangular matrix, L is lower triangular, and P is a
+// permutation matrix.
 //
-// Dpstf2 does not attempt to check that A is positive semidefinite.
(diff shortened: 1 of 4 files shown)


lapack/gonum/dpstf2.go
```diff
@@ -17,10 +17,19 @@ import (
...
 //  Pᵀ * A * P = Uᵀ * U ,  if uplo = blas.Upper,
 //  Pᵀ * A * P = L  * Lᵀ,  if uplo = blas.Lower,
-// where U is an upper triangular matrix and L is lower triangular, and P is
-// stored as vector piv.
+// where U is an upper triangular matrix, L is lower triangular, and P is a
+// permutation matrix.
 //
-// Dpstf2 does not attempt to check that A is positive semidefinite.
+// tol is a user-defined tolerance. The algorithm terminates if the pivot is
+// less than or equal to tol. If tol is negative, then n*eps*max(A[k,k]) will be
```
vladimir-ch, line 24: Maybe we don't need to document what value will be used and just say "a default value".

kortschak: I think that's reasonable.

kortschak (dismissed):

lapack/gonum/dpstf2.go
vladimir-ch: I changed it before seeing that you edited the comment. Now I've reverted it back, can you please re-approve?

kortschak (approved):

PR #1767, opened 2022-02-04
lapack/gonum: clarify ok return from Dpstrf and Dpstf2

vladimir-ch:
Please take a look.


vladimir-ch:
I gave some more thought to Dpstrf and read again the reference docs (and http://www.netlib.org/lapack/lawnspdf/lawn161.pdf) and realized/understood that this function doesn't have an error mode, that's why I removed the word "successfully" from the docs. This means that the caller must be sure that the matrix is positive semi-definite (or definite), otherwise Dpstrf could return incomplete rank even though the matrix has full rank but is, for example, indefinite.

kortschak (approved):

PR #1769, opened 2022-02-11
graph/formats/rdf: add Has{All,Any}{Out,In} and Repeat query methods

kortschak:
Please take a look.

### DIFF
--- a/graph/formats/rdf/query.go
+++ b/graph/formats/rdf/query.go
@@ -232,6 +232,29 @@ func (q Query) Not(p Query) Query {
 	return r
 }
 
+// Repeat repeatedly calls fn on q until the set of results is empty or
+// ok is false, and then returns the result. If the last non-empty result
+// is wanted, fn should return its input and false when the partial
+// traversal returns an empty result.
+//
+// 	result := start.Repeat(func(q rdf.Query) (rdf.Query, bool) {
+// 		r := q.Out(condition)
+// 		if r.Len() == 0 {
+// 			return q, false
+// 		}
+// 		return r, true
+// 	}).Result()
+//
+func (q Query) Repeat(fn func(Query) (q Query, ok bool)) Query {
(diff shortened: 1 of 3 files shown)


kortschak:
This needs `AnyOut` and `AnyIn` to be able to provide the power I want. Don't review yet.

kortschak:
You might want to sit down for this one.

graph/formats/rdf/query.go
```diff
@@ -153,6 +232,29 @@ func (q Query) Not(p Query) Query {
...
+// 	}).Result()
+//
+func (q Query) Repeat(fn func(Query) (q Query, ok bool)) Query {
+	for {
+		var ok bool
+		q, ok = fn(q)
+		if !ok || len(q.terms) == 0 {
+			return q
+		}
+	}
```
kortschak, line 255: The alternative implementation of `Repeat`,
```
func (q Query) Repeat(fn func(Query) (q Query, ok bool)) Query {
	for {
		r, ok := fn(q)
		if !ok || len(r.terms) == 0 {
			return q
		}
		q = r
	}
}
```
doesn't have this requirement, but I'm not sure which is better. I think being able to obtain an empty result is an important behaviour and that is slightly harder to do in the alternative formulation.

vladimir-ch: > I think being able to obtain an empty result is an important behaviour

Oh, this is very subtle. Why is it important and why the last non-empty result is more interesting/important than the previous non-empty results? Also, `fn` knows what it's returning and so when the repeat ends, so I'd be fine with keeping the current behavior.

vladimir-ch (approved):

graph/formats/rdf/query.go
vladimir-ch: Understood. Let's keep the current behaviour.

PR #1771, opened 2022-02-13
spatial/r3: add Mat Outer,Det,RotateQuat methods

### LINKED ISSUE #1757: spatial/r3: Adding more functionality to Mat type
**Vector Product** method

Adding `Mat.VecProd(v1,v2t)` or similar for multiplying two `r3.Vec` to obtain a matrix as in `Mat = Vec1 * Vec2ᵀ`. Seems to appear often in the literature I'm using.

<details><summary>VecProd()</summary>

```go
// Prod performs vector multiplication as if they were matrices
//  m = v1 * v2ᵀ
func (m *Mat) VecProd(v1, v2t r3.Vec) {
	m.Set(0, 0, v1.X*v2t.X)
	m.Set(0, 1, v1.X*v2t.Y)
	m.Set(0, 2, v1.X*v2t.Z)

	m.Set(1, 0, v1.Y*v2t.X)
	m.Set(1, 1, v1.Y*v2t.Y)
	m.Set(1, 2, v1.Y*v2t.Z)

	m.Set(2, 0, v1.Z*v2t.X)
	m.Set(2, 1, v1.Z*v2t.Y)
	m.Set(2, 2, v1.Z*v2t.Z)
}
```

</details>

**Rotation tensor from quaternion** method

Having a rotation tensor from a quaternion would be incredibly useful to have natively as a method:
`[R] = q.Real*q.Real * [E] - dot([q],[q])*[E] + 2*[q]*[q]ᵀ + 2*q.Real * skew([q])`, where `[q] = {q.Imag, q.Jmag, q.Kmag}` and `[E]=Eye()`

<details><summary>FromQuat()</summary> 

```go
// Not tested
func (m *Mat) RotationFromQuat(q quat.Number) {
	qv := r3.Vec{X: q.Imag, Y: q.Jmag, Z: q.Kmag}
	qs := Skew(qv)
	q01 := Identity()
	q01.Scale(q.Real*q.Real, q01)

	qd := Identity()
	qd.Scale(r3.Dot(qv, qv), qd)

	qs.Scale(2*q.Real, qs)

	qv = r3.Scale(2, qv)
	m.Prod(qv, qv) // m = 2*[q]*[q]ᵀ
	m.Add(m, q01)  // m += q.Real*q.Real * [E]
	m.Add(m, qd)   // m += dot([q],[q])*[E]
	m.Add(m, qs)   // m += 2*q.Real * skew([q])
}
```

</details>

I can also add the following:
* Determinant: Should be package level function? Or method? I feel it to be more of a function kind.
* Inverse: Probably a method which receives a 3x3 `mat.Matrix` and stores the result in the receiver?

### Is this feature absent from the current master?
Yes

### Are you able to help contribute the feature?
Yes

soypat:
Please take a look. Resolves #1757

Work on Inverse (could not get it to pass tests):
<details>

```go
func (m *Mat) Inverse(a mat.Matrix) error {
	if r, c := a.Dims(); r != 3 || c != 3 {
		panic("Mat.Inverse needs 3x3 matrix")
	}

	const small = 1e-12
	D := mat.Det(a)
	if math.Abs(D) < small {
		return mat.ErrSingular
	}
	// https://www.wikihow.com/Find-the-Inverse-of-a-3x3-Matrix
	for i := 0; i < 3; i++ {
		i1, i2 := (i+1)%3, (i+2)%3
		for j := 0; j < 3; j++ {
			j1, j2 := (j+1)%3, (j+2)%3
			det := a.At(i1, j1)*a.At(i2, j2) - a.At(i1, j2)*a.At(i2, j1)
			m.Set(j, i, det)
		}
	}
	m.Scale(1/D, m)
	return nil
}

func TestInverse(t *testing.T) {
	const tol = 1e-11
	for tc := 0; tc < 20; tc++ {
		m := randomMat(rnd)
		M := mat.DenseCopyOf(m)
		err := M.Inverse(M)
		if err != nil {
			t.Fatal(err)
		}
		err = m.Inverse(m)
		if err != nil {
			t.Fatal(err)
		}
		if !mat.EqualApprox(m, M, tol) {
			t.Error("expected similar inverse resultss")
		}
	}
}
```

</details>

### DIFF
--- a/spatial/r3/mat.go
+++ b/spatial/r3/mat.go
@@ -4,7 +4,10 @@
 
 package r3
 
-import "gonum.org/v1/gonum/mat"
+import (
+	"gonum.org/v1/gonum/mat"
+	"gonum.org/v1/gonum/num/quat"
+)
 
 // Mat represents a 3×3 matrix. Useful for rotation matrices and such.
 // The zero value is usable as the 3×3 zero matrix.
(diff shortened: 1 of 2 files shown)


kortschak:
I'll probably need to bump up the time for the failing colouring test.

spatial/r3/mat.go
```diff
@@ -163,3 +166,57 @@ func (m *Mat) VecCol(j int) Vec {
...
+	m.Set(1, 0, ay*y.X)
+	m.Set(1, 1, ay*y.Y)
+	m.Set(1, 2, ay*y.Z)
+
+	m.Set(2, 0, az*y.X)
+	m.Set(2, 1, az*y.Y)
+	m.Set(2, 2, az*y.Z)
+}
+
+// Sets the rotation component of this matrix to the rotation specified by q,
```
kortschak, line 190:
```
// RotationFromQuat sets the rotation component of this matrix to the rotation specified by q.
//
// See https://en.wikipedia.org/wiki/Rotation_matrix#Quaternion for details.
```

What is the behaviour is q is not a unit quaternion? Should that be documented? Should q be scaled to unit length if it is not?

spatial/r3/mat.go
```diff
@@ -163,3 +166,57 @@ func (m *Mat) VecCol(j int) Vec {
 	}
 	return Vec{X: m.At(0, j), Y: m.At(1, j), Z: m.At(2, j)}
 }
+
+// Outer calculates the outer product of the vectors x and y,
+// where x and y are treated as column vectors, and stores the result in the receiver.
+//  m = alpha * x * yᵀ
+func (m *Mat) Outer(alpha float64, x, y Vec) {
```
kortschak, line 173: This needs a test.

spatial/r3/mat.go
soypat: Definitely should be documented, slipped my mind. Should be easy enough to check in the function after calculating x2, y2, z2 and panic if not normalized or normalize then and there. I think normalizing non-unit quaternions is OK though I'm unsure of whether it's good design.

kortschak: You can just call `quat.Abs(q)` to find if it's unit. The question is whether people want to scale at the same time.

kortschak: ... and `quat.Scale(1/quat.Abs(q). q)`.

soypat: What about having a boolean argument `scale` to make it more explicit? Think it might be worthwhile since it works as explicit documentation.

kortschak: TBH, I think just documenting that no quaternion scaling takes place is fine, this favours composition over parameterisation.

spatial/r3/mat.go
```diff
@@ -163,3 +166,60 @@ func (m *Mat) VecCol(j int) Vec {
...
+	m.Set(2, 0, az*y.X)
+	m.Set(2, 1, az*y.Y)
+	m.Set(2, 2, az*y.Z)
+}
+
+// Sets the rotation component of this matrix to the rotation specified by q,
+// as outlined at https://en.wikipedia.org/wiki/Rotation_matrix#Quaternion.
+//
+// RotationFromQuat requires a normalized quaternion q. Example:
+//  q = quat.Scale(1/quat.Abs(q), q)
```
kortschak, line 194: This is only true if you want a non-scaling rotation.

soypat:
Let me know if there is anything else I can make better!

kortschak (approved):

kortschak:
Sorry. Got away from me.

PR #1773, opened 2022-02-23
spatial/r3: small cleanup

vladimir-ch:
- spatial/r3: remove netlib from dependency requirements
- spatial/r3: fix doc comments to start with function name

Please take a look.

### DIFF
--- a/spatial/r3/mat.go
+++ b/spatial/r3/mat.go
@@ -55,8 +55,7 @@ func (m *Mat) Scale(f float64, a mat.Matrix) {
 	}
 }
 
-// Performs matrix multiplication on v:
-//  result = M * v
+// MulVec returns the matrix-vector product M⋅v.
 func (m *Mat) MulVec(v Vec) Vec {
 	if m.data == nil {
 		return Vec{}
@@ -68,8 +67,7 @@ func (m *Mat) MulVec(v Vec) Vec {
 	}
 }
 
-// Performs transposed matrix multiplication on v:
-//  result = Mᵀ * v
+// MulVecTrans returns the matrix-vector product Mᵀ⋅v.
 func (m *Mat) MulVecTrans(v Vec) Vec {
 	if m.data == nil {
 		return Vec{}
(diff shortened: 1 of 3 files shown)


spatial/r3/mat.go
```diff
@@ -187,12 +185,12 @@ func (m *Mat) Outer(alpha float64, x, y Vec) {
 	m.Set(2, 2, az*y.Z)
 }
 
-// Sets the rotation component of this matrix to the rotation specified by q,
-// as outlined at https://en.wikipedia.org/wiki/Rotation_matrix#Quaternion.
+// RotationFromQuat converts the quaternion q to the corresponding orthogonal
+// matrix and stores the result in the receiver. q should be normalized but this
```
vladimir-ch, line 189: Why not normalize the quaternion inside the function?

kortschak: In the case that the user want a rotation and scale, that becomes impossible.

kortschak (approved):

spatial/r3/mat.go
vladimir-ch: But then we're in an undocumented territory and the existing documentation and naming becomes inaccurate: the matrix is not orthogonal and so not a rotation.

kortschak: Fair enough; probably we should do that then and note that it's normalised. Thanks for that.

vladimir-ch: Yeah, I think if scaling is needed it can be easily achieved with `Scale`.

vladimir-ch: I've been thinking about this a bit more and it's probably the cleanest to leave it as is. The behavior is documented for unit quaternions, non-unit behavior is not documented and we don't need to test it. We could maybe replace "should" with "must", @kortschak ?

kortschak: That's fine with me. I think that people should be using quaternions for this kind of thing where all this just works.

PR #1774, opened 2022-02-24
spatial/r3: rework test for RotationFromQuat

vladimir-ch:
Please take a look.

### DIFF
--- a/spatial/r3/mat_test.go
+++ b/spatial/r3/mat_test.go
@@ -189,30 +189,35 @@ func TestOuter(t *testing.T) {
 }
 
 func TestRotationFromQuat(t *testing.T) {
-	const tol = 1e-11
+	const tol = 1e-14
 	rnd := rand.New(rand.NewSource(1))
-	var backing [9]float64 // reuse memory.
+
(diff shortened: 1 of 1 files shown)


vladimir-ch:
I test via `Rotation.Rotate` applied on a random point instead of `Rotation.Matrix` because the latter uses the same formula as `RotationFromQuat` and so could be considered not independent enough.

Working on the test I discovered `Rotation.Matrix` which does the same thing as `Mat.RotationFromQuat`. Shouldn't only one of them exist?

kortschak:
My original intention was to not convert to a matrix. I forgot about `Rotation`. I think that the `Rotation.Matrix` code could return a `Mat` and have `RotationFromQuat` go away, maybe adding a `Mat` method that returns the type in its concrete form.

kortschak (approved):
LGTM, but I agree some of this API should go away; we now have two trivially different ways to do the same thing.

kortschak:
I'll have an API reduction ready soon.

vladimir-ch:
Good to know, I've just been looking at it too :-)

PR #1776, opened 2022-02-25
spatial/r3: remove duplicated API

kortschak:
Mainly colour and movement. The `Matrix` method is deleted as extraneous, and it never got into a release, so it's not breaking; everything else otherwise just works the same way that it did and the new tests in #1774 are hooked into the old code.

Please take a look.

### DIFF
--- a/spatial/r3/mat_safe.go
+++ b/spatial/r3/mat_safe.go
@@ -136,3 +136,25 @@ func arrayFrom(vals []float64) *array {
 	copy(a[:], vals)
 	return &a
 }
+
+// Mat returns a 3×3 rotation matrix corresponding to the receiver. It
+// may be used to perform rotations on a 3-vector or to apply the rotation
+// to a 3×n matrix of column vectors. If the receiver is not a unit
+// quaternion, the returned matrix will not be a pure rotation.
+func (r Rotation) Mat() *Mat {
+	w, i, j, k := r.Real, r.Imag, r.Jmag, r.Kmag
+	ii := 2 * i * i
+	jj := 2 * j * j
+	kk := 2 * k * k
+	wi := 2 * w * i
+	wj := 2 * w * j
+	wk := 2 * w * k
+	ij := 2 * i * j
+	jk := 2 * j * k
+	ki := 2 * k * i
+	return &Mat{&array{
+		1 - (jj + kk), ij - wk, ki + wj,
+		ij + wk, 1 - (ii + kk), jk - wi,
+		ki - wj, jk + wi, 1 - (ii + jj),
+	}}
+}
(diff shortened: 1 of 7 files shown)


spatial/r3/mat_test.go
```diff
@@ -224,9 +223,8 @@ func TestRotationFromQuat(t *testing.T) {
 
 func BenchmarkQuat(b *testing.B) {
 	rnd := rand.New(rand.NewSource(1))
-	m := NewMat(nil)
 	for i := 0; i < b.N; i++ {
 		q := quat.Number{Real: rnd.Float64(), Imag: rnd.Float64(), Jmag: rnd.Float64(), Kmag: rnd.Float64()}
-		m.RotationFromQuat(q)
+		Rotation(q).Mat()
```
vladimir-ch, line 228: Should this assign to a global variable?

spatial/r3/mat_safe.go
```diff
@@ -136,3 +136,25 @@ func arrayFrom(vals []float64) *array {
 	copy(a[:], vals)
 	return &a
 }
+
+// Mat returns a 3×3 rotation matrix corresponding to the receiver. It
+// may be used to perform rotations on a 3-vector or to apply the rotation
+// to a 3×n matrix of column vectors. If the receiver is not a unit
+// quaternion, the returned matrix will not be a pure rotation.
+func (r Rotation) Mat() *Mat {
+	re, im, jm, km := r.Real, r.Imag, r.Jmag, r.Kmag
```
vladimir-ch, line 145: It's just naming but since we're touching this I'd prefer:
```
w, i, j, k := ...
ii := 2*i*i
jj := 2*j*j
...
```

spatial/r3/mat_test.go
```diff
@@ -198,8 +198,7 @@ func TestRotationFromQuat(t *testing.T) {
 		q = quat.Scale(1/quat.Abs(q), q)
 
 		// Convert it to a rotation matrix R.
-		var r Mat
-		r.RotationFromQuat(q)
+		r := Rotation(q).Mat()
```
vladimir-ch, line 201: Rename the test to `TestRotationMat`?

sbinet (dismissed):

spatial/r3/mat_test.go
kortschak: I'll do a nil test on it.

spatial/r3/mat_safe.go
kortschak: Done.

spatial/r3/mat_test.go
kortschak: Done.

vladimir-ch (approved):

PR #1779, opened 2022-03-05
floats/scalar: make use of math.Round and math.RoundToEven

kortschak:
This reduces the code here and improves the behaviour slightly.

Please take a look.


sbinet (approved):

kortschak:
Would still like to deprecate it, but we use it in plot.

PR #1780, opened 2022-03-09
mat: add PivotedCholesky

vladimir-ch:
This is just a draft for now to discuss and collect feedback.

My doubts are:

1. Should `Factorize` return a bool? And the closely related question, should `PivotedCholesky.SolveTo/SolveVecTo` exist? If the solve routines don't exist, then Factorize doesn't need to return ok. If they exist, then they must panic if ok is false. This gives the ok return a different meaning from the other factorization types: it's not a failure of the factorization, it's just an indicator that the matrix has full rank (that is, positive definite) and can be used for linear solve. However, if one expects that the matrix is only semi-definite, then one should not expect to be able to solve linear systems with it.
2. Should `Factorize` take a tolerance parameter for `Dpstrf` with the same semantics (default value if negative)?
3. How to provide the factor? With normal Cholesky `distmv.Normal.TransformNormal` needs the factor `L` which here becomes `P*L`. However, `P*L` is not lower triangular. Returning a dense would be a waste. We could return `P` and `L` separately but then `P` should probably be a new Permuatation matrix type? We could have `PL` returning a `MulVecToer` for computing `P*L*x` which is limited but enough for `distmv` needs. Any other ideas?

### DIFF
--- a/mat/cholesky.go
+++ b/mat/cholesky.go
@@ -25,6 +25,9 @@ var (
 	_ Symmetric = (*BandCholesky)(nil)
 	_ Banded    = (*BandCholesky)(nil)
 	_ SymBanded = (*BandCholesky)(nil)
+
+	_ Matrix    = (*PivotedCholesky)(nil)
+	_ Symmetric = (*PivotedCholesky)(nil)
 )
 
 // Cholesky is a symmetric positive definite matrix represented by its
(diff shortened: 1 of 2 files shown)


kortschak:
1. If you think the solvers will be useful, then it makes sense to have them. The ok can be documented to indicate whether the solvers can be used, and the solvers can be documented to only be safe to use if the `PivotedCholesky` has previously returned an ok=true.
2. I can't see any reason not to allow that with those semantics (they match the Go-wide convention _and_ the LAPACK convention)
3. Yes, I was thinking a permutation matrix type would be the way to go (I missed the suggestion when reading it the first time — maybe it was subliminally planted).

vladimir-ch:
This is ready for review. The methods to extract `L` and `P` are still missing, so the functionality is practically equivalent to `Cholesky`. I'll have to think about the permutation matrix type.

kortschak (approved):

kortschak:
This will need to be rebased onto master to pick up the go1.21 testing.

kortschak (approved):

PR #1781, opened 2022-03-10
spatial/r3: use array conversion in safe code

kortschak:
This is the only Go version dependent change that needs to be made prior to the next release. Sadly, the unsafe conversion needs to remain in the unsafe path, otherwise we could have collapsed it back to a single file.

Please take a look.


kortschak:
I think it might be time to stop using golangci-lint soon. Just waiting for staticcheck to support linting only changes (apparently this is coming).

sbinet (approved):

PR #1790, opened 2022-03-21
lapack/gonum: add Dlapmr

vladimir-ch:
Please take a look.

### DIFF
--- a/lapack/gonum/dlapmr.go
+++ b/lapack/gonum/dlapmr.go
@@ -1,0 +1,85 @@
+// Copyright ©2022 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package gonum
+
+import "gonum.org/v1/gonum/blas/blas64"
+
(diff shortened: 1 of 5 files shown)


lapack/gonum/dlapmr.go
```diff
@@ -0,0 +1,85 @@
...
+	case n < 0:
+		panic(nLT0)
+	case ldx < max(1, n):
+		panic(badLdX)
+	}
+
+	// Quick return if possible.
+	if m == 0 || n == 0 {
+		return
+	}
```
kortschak, line 31: Move this below the parameter checks and merge with the other quick return?

kortschak (approved):
Approved if you want to merge without the suggestion. LGTM

lapack/gonum/dlapmr.go
vladimir-ch: The length check doesn't work if m==0, so this quick return has to be here. Moving the other quick return here wouldn't check that k has the right length, so I'll just leave it as is.

Issue #1791, opened 2022-05-08
proposal: spatial package: add Box, Triangle types and Elem functions


soypat:
### Background
I am developing a [replacement](https://github.com/soypat/sdf) for a [Signed Distance Function](https://en.wikipedia.org/wiki/Signed_distance_function) (SDF) library for generating 2D and 3D shapes. This module I am writing makes as heavy use of spatial types and algorithms, notably
* Vectors (Already implemented as `rn.Vec` types)
* Bounding Boxes (Implemented only as `r2.Box`)
* Triangles (Not implemented by gonum)
* [Elem functions](https://github.com/soypat/sdf/blob/main/internal/d3/vector.go#L37) such as the [Hadamard product](https://en.wikipedia.org/wiki/Hadamard_product_(matrices)). These functions operate on individual elements of vectors.

I am keen on using gonum as the defacto package for all this logic, but it requires the types be added. I am available to contribute this feature.

### Proposal
I've been working with these types for the past two weeks, so keep in mind my experience is far from thorough.

#### Adding Box types and utility methods
The Box type is already present in `r2`, this proposal would seek to add it to `r3` as well.

This proposal would also seek to add utility **methods** to the Box type. I have the library I wrote to measure how often [I've used these methods](https://github.com/soypat/sdf/blob/main/internal/d3/box.go#L11):
* `Size`.  Size returns the size of a 3d box. Very commonly used in bounding box operations.
* `Center`.  Center returns the center of a 3d box. Very commonly used in bounding box operations.
* `Vertices`.  Vertices returns a slice of 3d box corner vertices. Very commonly used in bounding box operations.
* 9 uses of `Extend`. Extend returns a box enclosing two 3d boxes.
* 8 uses of `Translate`. Translate translates a 3d box with a Vec
* 2 uses of `Enlarge` . Enlarge returns a new 3d box enlarged by a size vector
* 1 use of `func (a Box) MinMaxDist2(p r3.Vec) (min, max float64)` returns the minimum and maximum dist * dist from a point to a box. Is quite complex for such a simple objective.

I think gonum is a well suited place for these kind of functions. That said I'd like to open the floor for discussion on how best to combine the functions (or even add others not mentioned here) so that more can be done with less.

#### Add Triangle types
Adding triangle types for `r2` and `r3` packages.

Renderers require using triangles for representing 3D and 2D models. Below is the type declaration of Triangle from the [sdf package](https://github.com/soypat/sdf/blob/main/render/render.go).
```go
[...]

PR #1792, opened 2022-05-08
spatial/r3: add Triangle

soypat:
Please take a look.

Resolves part of #1791

### DIFF
--- a/spatial/r3/icosahedron_example_test.go
+++ b/spatial/r3/icosahedron_example_test.go
@@ -1,0 +1,104 @@
+// Copyright ©2022 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package r3_test
+
+import (
+	"fmt"
+
+	"gonum.org/v1/gonum/spatial/r3"
+)
+
+func ExampleTriangle_icosphere() {
+	// This example generates a 3D icosphere from
+	// a starting icosahedron by subdividing surfaces.
+	// See https://schneide.blog/2016/07/15/generating-an-icosphere-in-c/.
+	const subdivisions = 5
+	// vertices is a slice of r3.Vec
+	// triangles is a slice of [3]int indices
+	// referencing the vertices.
+	vertices, triangles := icosahedron()
+	for i := 0; i < subdivisions; i++ {
+		vertices, triangles = subdivide(vertices, triangles)
+	}
+	var faces []r3.Triangle
+	for _, t := range triangles {
+		var face r3.Triangle
+		for i := 0; i < 3; i++ {
+			face[i] = vertices[t[i]]
+		}
+		faces = append(faces, face)
+	}
+	fmt.Println(faces)
+	// The 3D rendering of the icosphere is left as an exercise to the reader.
+}
+
+// edgeIdx represents an edge of the icosahedron
+type edgeIdx [2]int
+
+func subdivide(vertices []r3.Vec, triangles [][3]int) ([]r3.Vec, [][3]int) {
+	// We generate a lookup table of all newly generated vertices so as to not
+	// duplicate new vertices. edgeIdx has lower index first.
+	lookup := make(map[edgeIdx]int)
+	var result [][3]int
+	for _, triangle := range triangles {
+		var mid [3]int
+		for edge := 0; edge < 3; edge++ {
+			lookup, mid[edge], vertices = subdivideEdge(lookup, vertices, triangle[edge], triangle[(edge+1)%3])
+		}
+		newTriangles := [][3]int{
+			{triangle[0], mid[0], mid[2]},
+			{triangle[1], mid[1], mid[0]},
+			{triangle[2], mid[2], mid[1]},
+			{mid[0], mid[1], mid[2]},
+		}
+		result = append(result, newTriangles...)
+	}
+	return vertices, result
+}
+
+// subdivideEdge takes the vertices list and indices first and second which
+// refer to the edge that will be subdivided.
+// lookup is a table of all newly generated vertices from
+// previous calls to subdivideEdge so as to not duplicate vertices.
+func subdivideEdge(lookup map[edgeIdx]int, vertices []r3.Vec, first, second int) (map[edgeIdx]int, int, []r3.Vec) {
+	key := edgeIdx{first, second}
+	if first > second {
+		// Swap to ensure edgeIdx always has lower index first.
+		key[0], key[1] = key[1], key[0]
+	}
+	vertIdx, vertExists := lookup[key]
+	if !vertExists {
+		// If edge not already subdivided add
+		// new dividing vertex to lookup table.
+		edge0 := vertices[first]
+		edge1 := vertices[second]
+		point := r3.Unit(r3.Add(edge0, edge1)) // vertex at a normalized position.
+		vertices = append(vertices, point)
+		vertIdx = len(vertices) - 1
+		lookup[key] = vertIdx
+	}
+	return lookup, vertIdx, vertices
+}
+
+// icosahedron returns an icosahedron mesh.
+func icosahedron() (vertices []r3.Vec, triangles [][3]int) {
+	const (
+		radiusSqrt = 1.0 // Example designed for unit sphere generation.
+		X          = radiusSqrt * .525731112119133606
+		Z          = radiusSqrt * .850650808352039932
+		N          = 0.0
+	)
+	return []r3.Vec{
+			{-X, N, Z}, {X, N, Z}, {-X, N, -Z}, {X, N, -Z},
+			{N, Z, X}, {N, Z, -X}, {N, -Z, X}, {N, -Z, -X},
+			{Z, X, N}, {-Z, X, N}, {Z, -X, N}, {-Z, -X, N},
+		}, [][3]int{
+			{0, 1, 4}, {0, 4, 9}, {9, 4, 5}, {4, 8, 5},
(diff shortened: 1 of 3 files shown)


spatial/r3/triangle.go
```diff
@@ -0,0 +1,72 @@
...
+
+// Triangle represents a triangle in 3D space and
+// is composed by 3 vectors corresponding to the position
+// of each of the vertices. Ordering of these vertices
+// decides the "normal" direction.
+type Triangle struct {
+	// Triangle vertices. Ordering of these vertices decides Normal vector direction.
+	// Inverting ordering of two vertices inverts the resulting direction.
+	V [3]Vec
+}
```
sbinet, line 17: just a drive-by comment:
any reason why we don't define it as:
```go
type Triangle [3]Vec
```

or, (better for the SSA pass):
```go
type Triangle struct {
    A, B, C Vec
}
```


do we expect to sub-slice `t.V`? or iterate over the vertices of a triangle?

spatial/r3/triangle.go
```diff
@@ -0,0 +1,72 @@
...
+
+// Normal returns the vector with direction
+// perpendicular to the Triangle's face. The ordering
+// of the triangle vertices decides the normal's resulting
+// direction. Normal is not guaranteed to return a unit-norm vector.
+func (t Triangle) Normal() Vec {
+	s1, s2, _ := t.sides()
```
sbinet, line 24: it seems (moderately) wasteful to ask for the computation of `Sub(t.V[0], t.V[2])` to drop it right there.
but ok.

sbinet (approved):
just a little stream of consciousness.

LGTM, perhaps add a little example?

spatial/r3/triangle.go
soypat: It's common to iterate over triangle vertices to make code more succinct. I can change it to `type Triangle [3]Vec`, which I prefer over the current implementation.

spatial/r3/triangle.go
soypat: Would the SSA not optimize the calculation? Maybe not today but it does seem like something that could done given it looks like a short enough path

soypat:
@sbinet I've added an example for which I do not have the tools to validate. I have also added the `Centroid` method here and in the `r2` implementation.

sbinet (dismissed):

sbinet:

soypat:
@kortschak  I have added the tests requested with a handful of comments to help out readers. 

I managed to fix the icosphere example. Below are 0 and 7 subdivisions of an icosahedron.
image
image

spatial/r3/triangle.go
```diff
@@ -0,0 +1,116 @@
...
+	return Cross(s1, s2)
+}
+
+// Degenerate returns true if all of triangle's vertices are
+// within tol distance of its longest side.
+func (t Triangle) Degenerate(tol float64) bool {
```
kortschak, line 27: Consider renaming this to `IsDegenerate`.

spatial/r3/triangle.go
```diff
@@ -0,0 +1,116 @@
...
+	ln := line{t[longIdx], t[(longIdx+1)%3]}
+	dist := ln.distance(t[(longIdx+2)%3])
+	return dist <= tol
+}
+
+// Area returns the surface area of the triangle.
+func (t Triangle) Area() float64 {
+	// Formula taken from Section 2, page 3.
+	// William M. Kahan (24 March 2000). "Miscalculating Area and Angles of a Needle-like Triangle"
+	// https://people.eecs.berkeley.edu/~wkahan/Triangle.pdf
```
kortschak, line 39: Note that this is Heron's formula. I would probably link to the first (as well as this reference) given the ancient history of it. https://en.wikipedia.org/wiki/Heron%27s_formula.

```
// Heron's Formula, see https://en.wikipedia.org/wiki/Heron%27s_formula.
// Also see William M. Kahan (24 March 2000). "Miscalculating Area and Angles of a Needle-like Triangle"
// for more discussion. https://people.eecs.berkeley.edu/~wkahan/Triangle.pdf.
```

spatial/r3/triangle.go
```diff
@@ -0,0 +1,116 @@
...
+		l1, l2 = l2, l1
+	}
+	if l3 < l2 {
+		l2, l3 = l3, l2
+		if l2 < l1 {
+			l1, l2 = l2, l1
+		}
+	}
+	return l1, l2, l3
+}
```
kortschak, line 95: Used only once for `Area`, so place it immediately after that.

spatial/r3/triangle.go
```diff
@@ -0,0 +1,116 @@
...
+	longIdx := 0
+	if len2[1] > longLen {
+		longLen = len2[1]
+		longIdx = 1
+	}
+	if len2[2] > longLen {
+		longIdx = 2
+	}
+	return longIdx
+}
```
kortschak, line 75: Since this is only used once, for `Degenerate`, I'd place it immediately after that.

kortschak (commented):
This is looking good.

kortschak:
Can you rebase onto the current master to get the new linter.

spatial/r3/icosahedron_example_test.go
```diff
@@ -0,0 +1,98 @@
...
+	var faces []r3.Triangle
+	for _, t := range triangles {
+		var face r3.Triangle
+		for i := 0; i < 3; i++ {
+			face[i] = vertices[t[i]]
+		}
+		faces = append(faces, face)
+	}
+	fmt.Println(faces)
+	// The 3D rendering of the icosphere is left as an exercise to the reader.
```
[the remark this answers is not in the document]
soypat: Left a small pointer-comment, good enough?

spatial/r3/icosahedron_example_test.go
```diff
@@ -0,0 +1,98 @@
...
+)
+
+// edgeIdx represents an edge of the icosahedron
+type edgeIdx [2]int
+
+// subdivideEdge takes the vertices list and indices first and second to
+// the vertices defining the edge that will be subdivided. lookup is a
+// table of all newly generated vertices from previous calls to subdivideEdge
+// so as to not duplicate vertices.
+func subdivideEdge(lookup map[edgeIdx]int, vertices []r3.Vec, first, second int) (int, []r3.Vec) {
```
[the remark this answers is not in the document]
soypat: I reversed the order of the functions but left them before the example following suit of the euler angle example. Should I add them after the example or is this OK?

spatial/r3/icosahedron_example_test.go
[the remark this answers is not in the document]
kortschak: Oh, sorry, no. OOB. This comment for the Barnes-Hut example was how the galaxy example for gio was instigated. This doesn't need to be here.

soypat:
I rebased onto current master, which is [ci: install ragel to check rdf codegen](https://github.com/gonum/gonum/commit/84fd2bc2bbc4a322613c3e2e81647e3a9682bcc8)

kortschak:
You're right, sorry. The linter change hasn't merged yet.

spatial/r3/icosahedron_example_test.go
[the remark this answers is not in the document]
kortschak: The Euler example is a special case; it was in that order because people know (or think they know) what Euler Angles are and want to use them, and it was written as a warning. This example has helpers that are not (AFAICS) something to be copy-pasted into their code.

soypat:
Refactored icosphere example:
- All triangle normals now point owtward from icosphere center
- Bug caused duplicate loose subdiving vertices has been fixed, also clearer return value
- Example has been moved to front of file.

kortschak:
It might be worth inspecting the example's mesh for defect. You can do this by exporting it as an STL and inspecting it with meshmixer. If you don't have code for STL export, you can use this https://github.com/kortschak/stl.

kortschak (dismissed):

soypat:
Do not merge. I believe I found a bug with Area function.

soypat:
Nevermind, all is good. I've added more tests to `TestTriangleArea` and a new `TestTriangleOrderedLengths` test.

soypat:
Hey there peeps! I'm gonna go ahead and share something I'm really proud of: 
image
A sphere .stl file (**left**) was taken aas input and a signed distance function generated from it (**right**) using the [`Transform` type](https://github.com/gonum/gonum/pull/1802) and a generic rewrite of `kdtree` (just for kicks). I'm looking forward to exposing this functionality with gonum's upcoming `r3.Triangle` as an argument! This would allow users of `sdf` to edit .stl files as if they were native `sdf.SDF3` objects

spatial/r3/icosahedron_example_test.go
```diff
@@ -0,0 +1,104 @@
...
+		result = append(result, newTriangles...)
+	}
+	return vertices, result
+}
+
+// subdivideEdge takes the vertices list and indices first and second which
+// refer to the edge that will be subdivided.
+// lookup is a table of all newly generated vertices from
+// previous calls to subdivideEdge so as to not duplicate vertices.
+func subdivideEdge(lookup map[edgeIdx]int, vertices []r3.Vec, first, second int) (int, []r3.Vec) {
```
sbinet, line 65: perhaps also return the lookup table, to future proof any behaviour change in the guts of the `map` implementation?

kortschak: I don't think that's necessary; maps can't change their reference value behaviour because of the compatibility promise.

kortschak (approved):

PR #1793, opened 2022-05-08
spatial/r3: add Box methods

soypat:
Please take a look.

### DIFF
--- a/spatial/r3/box.go
+++ b/spatial/r3/box.go
@@ -1,0 +1,120 @@
+// Copyright ©2022 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package r3
+
+import "math"
+
+// Box is a 3D bounding box. Well formed Boxes Min components
+// are smaller than Max components.
+type Box struct {
+	Min, Max Vec
+}
+
+// NewBox is shorthand for Box{Min:Vec{x0,y0,z0}, Max:Vec{x1,y1,z1}}.
+// The sides are swapped so that the resulting Box is well formed.
+func NewBox(x0, y0, z0, x1, y1, z1 float64) Box {
+	return Box{
+		Min: Vec{X: math.Min(x0, x1), Y: math.Min(y0, y1), Z: math.Min(z0, z1)},
+		Max: Vec{X: math.Max(x0, x1), Y: math.Max(y0, y1), Z: math.Max(z0, z1)},
+	}
+}
+
+// IsEmpty returns true if a Box's volume is zero
+// or if a Min component is greater than its Max component.
+func (a Box) Empty() bool {
+	return a.Min.X >= a.Max.X || a.Min.Y >= a.Max.Y || a.Min.Z >= a.Max.Z
+}
+
+// Size returns the size of the Box.
+func (a Box) Size() Vec {
+	return Sub(a.Max, a.Min)
+}
+
+// Center returns the center of the Box.
+func (a Box) Center() Vec {
+	return Scale(0.5, Add(a.Min, a.Max))
+}
+
+// Vertices returns a slice of the 8 vertices
+// corresponding to each of the Box's corners.
+//
+// Ordering of vertices 0-3 is CCW in the XY plane starting at box minimum.
+// Ordering of vertices 4-7 is CCW in the XY plane starting at box minimum
+// for X and Y values and maximum Z value.
+//
+// Edges for the box can be constructed with the following indices:
+//  edges := [12][2]int{
+//   {0, 1}, {1, 2}, {2, 3}, {3, 0},
+//   {4, 5}, {5, 6}, {6, 7}, {7, 4},
+//   {0, 4}, {1, 5}, {2, 6}, {3, 7},
+//  }
+func (a Box) Vertices() []Vec {
+	return []Vec{
+		0: a.Min,
+		1: {X: a.Max.X, Y: a.Min.Y, Z: a.Min.Z},
+		2: {X: a.Max.X, Y: a.Max.Y, Z: a.Min.Z},
+		3: {X: a.Min.X, Y: a.Max.Y, Z: a.Min.Z},
+		4: {X: a.Min.X, Y: a.Min.Y, Z: a.Max.Z},
+		5: {X: a.Max.X, Y: a.Min.Y, Z: a.Max.Z},
+		6: a.Max,
+		7: {X: a.Min.X, Y: a.Max.Y, Z: a.Max.Z},
+	}
+}
+
+// Union returns a box enclosing both the receiver and argument Boxes.
+func (a Box) Union(b Box) Box {
+	if a.Empty() {
+		return b
+	}
+	if b.Empty() {
+		return a
+	}
+	return Box{
+		Min: minElem(a.Min, b.Min),
+		Max: maxElem(a.Max, b.Max),
+	}
+}
+
+// Add adds v to the bounding box components.
+// It is the equivalent of translating the Box by v.
+func (a Box) Add(v Vec) Box {
+	return Box{Add(a.Min, v), Add(a.Max, v)}
+}
+
+// Scale returns a new Box scaled by a size vector around its center.
+// The scaling is done element wise which is to say the Box's X dimension
+// is scaled by scale.X. Negative elements of scale are interpreted as zero.
+func (a Box) Scale(scale Vec) Box {
+	scale = maxElem(scale, Vec{})
+	// TODO(soypat): Probably a better way to do this.
+	return centeredBox(a.Center(), mulElem(scale, a.Size()))
+}
+
+// centeredBox creates a Box with a given center and size.
+// Negative components of size will be interpreted as zero.
+func centeredBox(center, size Vec) Box {
+	size = maxElem(size, Vec{}) // set negative values to zero.
+	half := Scale(0.5, size)
+	return Box{Min: Sub(center, half), Max: Add(center, half)}
+}
+
+// Contains returns true if v is contained within the bounds of the Box.
+func (a Box) Contains(v Vec) bool {
+	if a.Empty() {
+		return v == a.Min && v == a.Max
+	}
+	return a.Min.X <= v.X && v.X <= a.Max.X &&
+		a.Min.Y <= v.Y && v.Y <= a.Max.Y &&
+		a.Min.Z <= v.Z && v.Z <= a.Max.Z
+}
+
+// Canon returns the canonical version of a. The returned Box has minimum
+// and maximum coordinates swapped if necessary so that it is well-formed.
+func (a Box) Canon() Box {
+	return Box{
+		Min: minElem(a.Min, a.Max),
+		Max: maxElem(a.Min, a.Max),
+	}
+}
--- a/spatial/r3/box_test.go
+++ b/spatial/r3/box_test.go
@@ -1,0 +1,200 @@
+// Copyright ©2022 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package r3
+
+import (
+	"testing"
+
+	"golang.org/x/exp/rand"
+)
+
+func TestBoxContains(t *testing.T) {
+	rnd := rand.New(rand.NewSource(1))
+	for i := 0; i < 200; i++ {
+		b := randomBox(rnd)
+		for j := 0; j < 10; j++ {
+			contained := b.random(rnd)
+			if !b.Contains(contained) {
+				t.Error("bounding box should contain Vec")
+			}
+		}
+		uncontained := [6]Vec{
+			Add(b.Max, Vec{1, 0, 0}),
+			Add(b.Max, Vec{0, 1, 0}),
+			Add(b.Max, Vec{0, 0, 1}),
+			Sub(b.Min, Vec{1, 0, 0}),
+			Sub(b.Min, Vec{0, 1, 0}),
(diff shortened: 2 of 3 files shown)


spatial/r3/vector.go
```diff
@@ -83,7 +83,50 @@ func Cos(p, q Vec) float64 {
...
+//  v = {a.X*b.X, a.Y*b.Y, a.Z*b.Z}
+func mulElem(a, b Vec) Vec {
+	return Vec{
+		X: a.X * b.X,
+		Y: a.Y * b.Y,
+		Z: a.Z * b.Z,
+	}
+}
+
+// mulElem erforms Hadamard product between a vector a
```
sbinet, line 123:
```suggestion
// mulElem performs Hadamard product between a vector a
```

spatial/r3/vector.go
```diff
@@ -83,7 +83,50 @@ func Cos(p, q Vec) float64 {
...
+// maxElem return a vector with the maximum components of two vectors.
+func maxElem(a, b Vec) Vec {
+	return Vec{
+		X: math.Max(a.X, b.X),
+		Y: math.Max(a.Y, b.Y),
+		Z: math.Max(a.Z, b.Z),
+	}
+}
+
+// absElem returns vector with components set to their absolute value.
```
sbinet, line 104:
```suggestion
// absElem returns a vector with components set to their absolute value.
```

spatial/r3/vector.go
```diff
@@ -83,7 +83,50 @@ func Cos(p, q Vec) float64 {
...
+func absElem(a Vec) Vec {
+	return Vec{
+		X: math.Abs(a.X),
+		Y: math.Abs(a.Y),
+		Z: math.Abs(a.Z),
+	}
+}
+
+// mulElem performs Hadamard product between two vectors.
```
sbinet, line 113: I might go with:
```suggestion
// mulElem returns the Hadamard product between the vectors a and b.
```

spatial/r3/box.go
```diff
@@ -0,0 +1,69 @@
...
+	return Sub(a.Max, a.Min)
+}
+
+// Center returns the center of the Box.
+func (a Box) Center() Vec {
+	return Add(a.Min, Scale(0.5, a.Size()))
+}
+
+// Vertices returns a slice of the 8 vertices
+// corresponding to each of the Box's corners.
```
sbinet, line 29: do we want to document or specify the order?
or document it isn't specified?

soypat: Will document that it is not specified.

soypat:
@kortschak @sbinet Could this and https://github.com/gonum/gonum/pull/1792 be merged sometime this week? I'd like to be using these types for my CERN hackathon application :grinning:

sbinet (dismissed):

sbinet:

soypat:
@kortschak It's all good- thanks for letting me know in advance!

spatial/r3/box.go
```diff
@@ -0,0 +1,79 @@
...
+		a.Min,                                // 0
+		{X: a.Max.X, Y: a.Min.Y, Z: a.Min.Z}, // 1
+		{X: a.Max.X, Y: a.Max.Y, Z: a.Min.Z}, // 2
+		{X: a.Min.X, Y: a.Max.Y, Z: a.Min.Z}, // 3
+		{X: a.Min.X, Y: a.Min.Y, Z: a.Max.Z}, // 4
+		{X: a.Max.X, Y: a.Min.Y, Z: a.Max.Z}, // 5
+		a.Max,                                // 6
+		{X: a.Min.X, Y: a.Max.Y, Z: a.Max.Z}, // 7
+	}
+}
```
kortschak, line 51: Never put in comments what can go in code.

```suggestion
	return []Vec{
		0: a.Min,
		1: {X: a.Max.X, Y: a.Min.Y, Z: a.Min.Z},
		2: {X: a.Max.X, Y: a.Max.Y, Z: a.Min.Z},
		3: {X: a.Min.X, Y: a.Max.Y, Z: a.Min.Z},
		4: {X: a.Min.X, Y: a.Min.Y, Z: a.Max.Z},
		5: {X: a.Max.X, Y: a.Min.Y, Z: a.Max.Z},
		6: a.Max,
		7: {X: a.Min.X, Y: a.Max.Y, Z: a.Max.Z},
	}
}
```

spatial/r3/box.go
```diff
@@ -0,0 +1,79 @@
...
+// Box is a 3D bounding box.
+type Box struct {
+	Min, Max Vec
+}
+
+// CenteredBox creates a Box with a given center and size.
+func CenteredBox(center, size Vec) Box {
+	half := Scale(0.5, size)
+	return Box{Min: Sub(center, half), Max: Add(center, half)}
+}
```
kortschak, line 16: How often is this needed? It seems to me that it wouldn't be often and that it's simple enough to leave to users. (Though see comment on constructor).

spatial/r3/box.go
```diff
@@ -0,0 +1,79 @@
...
+//
+// Vertex ordering between 0-3 and 4-7 outline
+// 2D boxes in the XY plane where vertices 0-3 are in Z=Box.Min.Z.
+// One can then construct the edges with indices for the
+// return value of Vertices:
+//  edges := [12][2]int{
+//   {0, 1}, {1, 2}, {2, 3}, {3, 0},
+//   {4, 5}, {5, 6}, {6, 7}, {7, 4},
+//   {0, 4}, {1, 5}, {2, 6}, {3, 7},
+//  }
```
kortschak, line 39: I think this could be clearer.

spatial/r3/box.go
```diff
@@ -0,0 +1,79 @@
+// Copyright ©2022 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package r3
+
+// Box is a 3D bounding box.
+type Box struct {
```
kortschak, line 8: Can we have a constructor like [`image.Rect`](https://pkg.go.dev/image#Rect)?

spatial/r3/box.go
```diff
@@ -0,0 +1,79 @@
...
+// the Box's X size is scaled by v.X.
+func (a Box) Scale(v Vec) Box {
+	// TODO(soypat): Probably a better way to do this.
+	return CenteredBox(a.Center(), mulElem(v, a.Size()))
+}
+
+// Contains checks if the Box contains the given vector within its bounds.
+func (a Box) Contains(v Vec) bool {
+	return a.Min.X <= v.X && a.Min.Y <= v.Y && a.Min.Z <= v.Z &&
+		v.X <= a.Max.X && v.Y <= a.Max.Y && v.Z <= a.Max.Z
```
kortschak, line 78:
```suggestion
	return a.Min.X <= v.X && v.X <= a.Max.X &&
		a.Min.Y <= v.Y && v.Y <= a.Max.Y &&
		a.Min.Z <= v.Z && v.Z <= a.Max.Z
```

kortschak (changes requested):

kortschak:
Please also rebase onto master.

spatial/r3/box.go
soypat: I'd be ok with removing this. That said this form seems to be the most common form of constructing a box from scratch from what I've been doing.
What you are suggesting below is having a `NewBox(a,b Vec)` that ensures valid box creation, right?

kortschak: > What you are suggesting below is having a `NewBox(a,b Vec)` that ensures valid box creation, right?

Yes, so that `b.Min` is actually `elemMin(a, b)` and `b.Max` is `elemMax(a, b)`.

soypat: Alright, I've added Empty method along with valid box creation algorithm for `CenteredBox`, `Union` and `Scale`.

I still feel like `CenteredBox` is the most reasonable way of initializing a Box where complexity is reasonably justfied.

I'll also comment on `NewBox`: A `NewBox` constructor which does the same thing as `Rect` is not intuitive to me at all. "NewBox creates a box that contains all components of the two vectors"? In every case where I've created a box I've had precise information on how to create `Min` and `Max` from scratch. I feel like going the `Rect` way lends itself to messy Box creation where the user hopes the resulting Box contains the desired bounds.

kortschak: The constructor defines the box by the position of its side rather than the position of two of its corners. This is the intuition. An alternative way to think about it is as the union of two zero-sized boxes at the defined min and max (which many not have been min and max).

soypat: I do not see the difference between defining a box by its side over its corner. The corners lie on the sides after all? 

Let me restate my fear: If we allow constructing valid boxes with two arbitrary vectors then a lot of readability can be lost when it is misused. When I say misuse I mean users can find clever ways of constructing boxes such that in the example`NewBox(a,b)` `a.X < b.X` while `a.Y>b.Y`, resulting in confusion over which vector defines which side/corner (at least for me). When constructing boxes with `Box{Min:a, Max:b}` the intention is clear and the code more structured.

kortschak: > I do not see the difference between defining a box by its side over its corner. The corners lie on the sides after all?

Taking r2 to reduce complsxit of example, Box{Min:Vec{X:1,Y:1}, Max:Vec{2,2}} does not work the same as Box{Min:Vec{X:1,Y:2}, Max:Vec{2,1}} (this would be an invalid box under the implementations that I was looking at yesterday). In pseudo code Box for X=1..2 Y=1..2 is guaranteed to mean what the user intends.

> If we allow constructing valid boxes with two arbitrary vectors

They are not vectors. Take a look at [`image.Rect`](https://pkg.go.dev/image#Rect); it accepts integer arguments. `NewBox` would be `NewBox(x0, y0, x1, y1 float64) Box`. These parameters merely say make a box with vertical sides at x0 and x1 and horizontal sides at y0 and y1.

soypat: OK, if we are talking about a function signature like `NewBox(x0, y0, z0, x1, y1, z1 float64)` then I can agree this is OK since now it is clear the box is being created with sides rather than points (as implied by using `Vec` type). I was initially against this signature because it had lots of arguments- however, looking through my use of `Box` it's clear this function would replace **many** `Box{Min...Max...}` initializations.
Little question on the signature: which do we prefer?
```go
NewBox(x0, y0, z0, x1, y1, z1 float64)
NewBox(x0, x1, y0, y1, z0, z1 float64)
```
I am partial to the second one since related numbers are side-by-side.

kortschak: While I prefer the second, I'm going to argue for the first on the basis of similarity with `image.Rect`.

kortschak (commented):
Comments for r2 box apply here.

spatial/r3/box.go
```diff
@@ -0,0 +1,122 @@
...
+}
+
+// Size returns the size of the Box.
+func (a Box) Size() Vec {
+	return Sub(a.Max, a.Min)
+}
+
+// Center returns the center of the Box.
+func (a Box) Center() Vec {
+	return Add(a.Min, Scale(0.5, a.Size()))
```
kortschak, line 39: If this were `Scale(0.5, Add(a.Min, a.Max))` it would always return the centre, even for non-well formed boxes, and it would be infinitesimally cheaper.

spatial/r3/box.go
```diff
@@ -0,0 +1,122 @@
+// Copyright ©2022 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package r3
+
+import (
+	"math"
+)
```
kortschak, line 9: One line.

spatial/r3/box_test.go
```diff
@@ -0,0 +1,200 @@
...
+			Add(maxElem(b1.Max, b2.Max), Vec{1, 0, 0}),
+			Add(maxElem(b1.Max, b2.Max), Vec{0, 1, 0}),
+			Add(maxElem(b1.Max, b2.Max), Vec{0, 0, 1}),
+			Sub(minElem(b1.Min, b2.Min), Vec{1, 0, 0}),
+			Sub(minElem(b1.Min, b2.Min), Vec{0, 1, 0}),
+			Sub(minElem(b1.Min, b2.Min), Vec{0, 0, 1}),
+		}
+		for _, unc := range uncontained {
+			if !b1.Contains(unc) && !b2.Contains(unc) && u.Contains(unc) {
+				t.Fatalf("union should not contain Vec")
```
kortschak, line 65: s/Fatalf/Errorf/

kortschak (commented):
Minor changes then LGTM

spatial/r3/box.go
soypat: Good catch

spatial/r3/box_test.go
```diff
@@ -0,0 +1,200 @@
...
+}
+
+func TestBoxVertices(t *testing.T) {
+	rnd := rand.New(rand.NewSource(1))
+	for i := 0; i < 300; i++ {
+		b := randomBox(rnd)
+		gots := b.Vertices()
+		wants := goldenVertices(b)
+		if len(gots) != len(wants) {
+			t.Fatal("bad length of vertices. expect 8, got ", len(gots))
```
kortschak, line 116: The addition of a space to make a `Fatal` call format correctly is a smell. I'd prefer `Fatalf` here.

kortschak (commented):
LGTM after nit (BTW it's hard to review incrementally with amended changes since the changes are not associated with commits). It's fine to squash as a final move after in principle approval to avoid GitHub's borken approach to squash merge ownership).

soypat:
@kortschak Changes applied without amending since that was probably the last of it. I'll keep that in mind for future commits on other branches.

kortschak (approved):

PR #1794, opened 2022-05-08
spatial/r2: add Box methods

soypat:
Please take a look.

### DIFF
--- a/spatial/r2/box.go
+++ b/spatial/r2/box.go
@@ -1,0 +1,111 @@
+// Copyright ©2022 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package r2
+
+import "math"
+
+// Box is a 2D bounding box. Well formed Boxes have
+// Min components smaller than Max components.
+type Box struct {
+	Min, Max Vec
+}
+
+// NewBox is shorthand for Box{Min:Vec{x0,y0}, Max:Vec{x1,y1}}.
+// The sides are swapped so that the resulting Box is well formed.
+func NewBox(x0, y0, x1, y1 float64) Box {
+	return Box{
+		Min: Vec{X: math.Min(x0, x1), Y: math.Min(y0, y1)},
+		Max: Vec{X: math.Max(x0, x1), Y: math.Max(y0, y1)},
+	}
+}
+
+// Size returns the size of the Box.
+func (a Box) Size() Vec {
+	return Sub(a.Max, a.Min)
+}
+
+// Center returns the center of the Box.
+func (a Box) Center() Vec {
+	return Scale(0.5, Add(a.Min, a.Max))
+}
+
+// IsEmpty returns true if a Box's volume is zero
+// or if a Min component is greater than its Max component.
+func (a Box) Empty() bool {
+	return a.Min.X >= a.Max.X || a.Min.Y >= a.Max.Y
+}
+
+// Vertices returns a slice of the 4 vertices
+// corresponding to each of the Box's corners.
+//
+// The order of the vertices is CCW in the XY plane starting at the box minimum.
+// If viewing box from +Z position the ordering is as follows:
+//  1. Bottom left.
+//  2. Bottom right.
+//  3. Top right.
+//  4. Top left.
+func (a Box) Vertices() []Vec {
+	return []Vec{
+		0: a.Min,
+		1: {a.Max.X, a.Min.Y},
+		2: a.Max,
+		3: {a.Min.X, a.Max.Y},
+	}
+}
+
+// Union returns a box enclosing both the receiver and argument Boxes.
+func (a Box) Union(b Box) Box {
+	if a.Empty() {
+		return b
+	}
+	if b.Empty() {
+		return a
+	}
+	return Box{
+		Min: minElem(a.Min, b.Min),
+		Max: maxElem(a.Max, b.Max),
+	}
+}
+
+// Add adds v to the bounding box components.
+// It is the equivalent of translating the Box by v.
+func (a Box) Add(v Vec) Box {
+	return Box{Add(a.Min, v), Add(a.Max, v)}
+}
+
+// Scale returns a new Box scaled by a size vector around its center.
+// The scaling is done element wise, which is to say the Box's X size is
+// scaled by v.X. Negative components of v are interpreted as zero.
+func (a Box) Scale(v Vec) Box {
+	v = maxElem(v, Vec{})
+	// TODO(soypat): Probably a better way to do this.
+	return centeredBox(a.Center(), mulElem(v, a.Size()))
+}
+
+// centeredBox creates a Box with a given center and size. Size's negative
+// components are interpreted as zero so that resulting box is well formed.
+func centeredBox(center, size Vec) Box {
+	size = maxElem(size, Vec{})
+	half := Scale(0.5, absElem(size))
+	return Box{Min: Sub(center, half), Max: Add(center, half)}
+}
+
+// Contains returns true if v is contained within the bounds of the Box.
+func (a Box) Contains(v Vec) bool {
+	if a.Empty() {
+		return v == a.Min && v == a.Max
+	}
+	return a.Min.X <= v.X && v.X <= a.Max.X &&
+		a.Min.Y <= v.Y && v.Y <= a.Max.Y
+}
+
+// Canon returns the canonical version of b. The returned Box has minimum
+// and maximum coordinates swapped if necessary so that it is well-formed.
+func (b Box) Canon() Box {
+	return Box{
+		Min: minElem(b.Min, b.Max),
+		Max: maxElem(b.Min, b.Max),
+	}
+}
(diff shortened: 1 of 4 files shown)


spatial/r2/vector.go
```diff
@@ -112,3 +107,46 @@ func (r Rotation) Rotate(p Vec) Vec {
...
+// mulElem performs Hadamard product between two vectors.
+//  v = {a.X*b.X, a.Y*b.Y, a.Z*b.Z}
+func mulElem(a, b Vec) Vec {
+	return Vec{
+		X: a.X * b.X,
+		Y: a.Y * b.Y,
+	}
+}
+
+// mulElem erforms Hadamard product between a vector a
```
sbinet, line 144:
```suggestion
// mulElem performs the Hadamard product between a vector a
```

spatial/r2/vector.go
```diff
@@ -112,3 +107,46 @@ func (r Rotation) Rotate(p Vec) Vec {
...
+
+// absElem returns vector with components set to their absolute value.
+func absElem(a Vec) Vec {
+	return Vec{
+		X: math.Abs(a.X),
+		Y: math.Abs(a.Y),
+	}
+}
+
+// mulElem performs Hadamard product between two vectors.
```
sbinet, line 135:
```suggestion
// mulElem performs the Hadamard product between two vectors.
```

soypat:
Changes applied.

sbinet (dismissed):

spatial/r2/box.go
```diff
@@ -0,0 +1,68 @@
...
+// the Box's X size is scaled by v.X.
+func (a Box) Scale(v Vec) Box {
+	// TODO(soypat): Probably a better way to do this.
+	return CenteredBox(a.Center(), mulElem(v, a.Size()))
+}
+
+// Contains checks if the Box contains the given vector within its bounds.
+func (a Box) Contains(v Vec) bool {
+	return a.Min.X <= v.X && a.Min.Y <= v.Y &&
+		v.X <= a.Max.X && v.Y <= a.Max.Y
```
kortschak, line 67:
```suggestion
	return a.Min.X <= v.X && v.X <= a.Max.X &&
		a.Min.Y <= v.Y && v.Y <= a.Max.Y
```

kortschak (changes requested):
Comments to r3 also apply here mod the indexing in the literal; I'm not entirely sure what to do there, though do we need to say which is bottom/top (if we do, full words would be better).

spatial/r2/box.go
```diff
@@ -0,0 +1,101 @@
...
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package r2
+
+import "math"
+
+// Box is a 2D bounding box. Well formed
+// Boxes have Min components smaller than
+// Max components.
```
kortschak, line 11: Would you mind going through the docs and wrapping at approximately 70-80 unless there are reasons to not.

spatial/r2/box.go
```diff
@@ -0,0 +1,101 @@
...
+
+// Center returns the center of the Box.
+func (a Box) Center() Vec {
+	return Add(a.Min, Scale(0.5, a.Size()))
+}
+
+// IsEmpty returns true if a Box's volume is zero
+// or if a Min component is greater than its Max component.
+func (a Box) Empty() bool {
+	return a.Min.X >= a.Max.X || a.Min.Y >= a.Max.Y
```
kortschak, line 38: Do we ever expect to see min > max?

spatial/r2/box.go
```diff
@@ -0,0 +1,101 @@
...
+func (a Box) Union(b Box) Box {
+	if a.Empty() {
+		return b
+	}
+	if b.Empty() {
+		return a
+	}
+	return Box{
+		Min: minElem(a.Min, b.Min),
+		Max: maxElem(a.Max, b.Max),
```
kortschak, line 65: If the `Box` is incorrectly constructed, this will fail to give the right answer. See `image.Rectangle`'s implementation. I would say that a box has an invariant that min <= max for both dimensions, but this is not currently guaranteed.

spatial/r2/box.go
```diff
@@ -0,0 +1,101 @@
...
+func (a Box) Add(v Vec) Box {
+	return Box{Add(a.Min, v), Add(a.Max, v)}
+}
+
+// Scale returns a new Box scaled by a size vector around its center.
+// The scaling is done element wise, which is to say
+// the Box's X size is scaled by v.X.
+// Negative components of v are interpreted as zero.
+func (a Box) Scale(v Vec) Box {
+	v = maxElem(v, Vec{})
```
kortschak, line 80: Why do this? Is there a reason to disallow scaling by negative values?

spatial/r2/box.go
soypat: Unless we unexport `Min`, `Max` fields users can create malformed boxes.

spatial/r2/box.go
soypat: My understanding is this does the same thing as image.Union. both implementations discard malformed boxes with `Empty` and then return the element wise minimum and maximum.

spatial/r2/box.go
soypat: Scale creates a new box at `b`'s center with dimensions multiplied by `scale`. Negative components of `scale` yield negative dimensions, which would create a box with `Max` < `Min`. Do we want to allow creating malformed boxes? Probably not. We can either
* Interpret this input as spurious or incorrect and set the dimension to zero which would create a well formed empty box.
* Interpret that the user implies they want the absolute value of the dimension and set the dimension to the result of `math.Abs(component)`.

To me the latter seems like the most user friendly way to go about it. It also feels like bad API design since gonum would not help users detect "bad" dimension input where it exists.

spatial/r2/box.go
kortschak: Maybe https://pkg.go.dev/image#Rectangle.Canon?

spatial/r2/box.go
kortschak: Yeah, OK. With the addition of a cononicalisation method this would be OK.

spatial/r2/box.go
kortschak: Yeah. OK.

soypat: Sorry I incorrectly said the first felt like bad API design when I actually meant the second. Are we still on the same page?

soypat:
Would adding `Quadtree/Octree` methods be too much?
```go
func (b Box) Octree() []Box {
	ctr := b.Center()
	min := b.Min
	max := b.Max
	return []Box{
		0: {Min: b.Min, Max: ctr},
		1: {Min: Vec{ctr.X, min.Y, min.Z}, Max: Vec{max.X, ctr.Y, ctr.Z}},
		2: {Min: Vec{min.X, ctr.Y, min.Z}, Max: Vec{ctr.X, max.Y, ctr.Z}},
		3: {Min: Vec{ctr.X, ctr.Y, min.Z}, Max: Vec{max.X, max.Y, ctr.Z}},
		4: {Min: Vec{min.X, min.Y, ctr.Z}, Max: Vec{ctr.X, ctr.Y, max.Z}},
		5: {Min: Vec{ctr.X, min.Y, ctr.Z}, Max: Vec{max.X, ctr.Y, max.Z}},
		6: {Min: Vec{min.X, ctr.Y, ctr.Z}, Max: Vec{ctr.X, max.Y, max.Z}},
		7: {Min: ctr, Max: max},
	}
}
```

kortschak:
Yes, it would. We want to be agnostic about stores. Users can always embed.

spatial/r2/box.go
```diff
@@ -0,0 +1,111 @@
...
+}
+
+// Size returns the size of the Box.
+func (a Box) Size() Vec {
+	return Sub(a.Max, a.Min)
+}
+
+// Center returns the center of the Box.
+func (a Box) Center() Vec {
+	return Add(a.Min, Scale(0.5, a.Size()))
```
kortschak, line 31: Same comment as for r3.

spatial/r2/box_test.go
```diff
@@ -0,0 +1,210 @@
...
+		}
+		uncontained := [4]Vec{
+			Add(maxElem(b1.Max, b2.Max), Vec{1, 0}),
+			Add(maxElem(b1.Max, b2.Max), Vec{0, 1}),
+			Sub(minElem(b1.Min, b2.Min), Vec{1, 0}),
+			Sub(minElem(b1.Min, b2.Min), Vec{0, 1}),
+		}
+		for _, unc := range uncontained {
+			if !b1.Contains(unc) && !b2.Contains(unc) && u.Contains(unc) {
+				t.Fatalf("union should not contain Vec")
```
kortschak, line 61: s/Fatalf/Errorf/

kortschak (commented):
LGTM after minor fixes.

spatial/r2/box_test.go
```diff
@@ -0,0 +1,210 @@
...
+}
+
+func TestBoxVertices(t *testing.T) {
+	rnd := rand.New(rand.NewSource(1))
+	for i := 0; i < 300; i++ {
+		b := randomBox(rnd)
+		gots := b.Vertices()
+		wants := goldenVertices(b)
+		if len(gots) != len(wants) {
+			t.Fatal("bad length of vertices. expect 4, got", len(gots))
```
kortschak, line 171: `Fatalf` here?

kortschak (approved):

PR #1795, opened 2022-05-08
spatial/r2: add Triangle

soypat:
Please take a look.

### DIFF
--- a/spatial/r2/triangle.go
+++ b/spatial/r2/triangle.go
@@ -1,0 +1,96 @@
+// Copyright ©2022 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package r2
+
+import "math"
+
+// Triangle represents a triangle in 2D space and
+// is composed by 3 vectors corresponding to the position
+// of each of the vertices.
+type Triangle [3]Vec
+
+// Centroid returns the intersection of the three medians of the triangle
+// as a point in space.
+func (t Triangle) Centroid() Vec {
+	return Scale(1.0/3.0, Add(Add(t[0], t[1]), t[2]))
+}
+
+// Area returns the surface area of the triangle.
+func (t Triangle) Area() float64 {
+	// Heron's Formula, see https://en.wikipedia.org/wiki/Heron%27s_formula.
+	// Also see William M. Kahan (24 March 2000). "Miscalculating Area and Angles of a Needle-like Triangle"
+	// for more discussion. https://people.eecs.berkeley.edu/~wkahan/Triangle.pdf.
+	a, b, c := t.orderedLengths()
+	A := (c + (b + a)) * (a - (c - b))
+	A *= (a + (c - b)) * (c + (b - a))
+	return math.Sqrt(A) / 4
+}
+
+// orderedLengths returns the lengths of the sides of the triangle such that
+// a ≤ b ≤ c.
+func (t Triangle) orderedLengths() (a, b, c float64) {
+	s1, s2, s3 := t.sides()
+	l1 := Norm(s1)
+	l2 := Norm(s2)
+	l3 := Norm(s3)
+	// sort-3
+	if l2 < l1 {
+		l1, l2 = l2, l1
+	}
+	if l3 < l2 {
+		l2, l3 = l3, l2
+		if l2 < l1 {
+			l1, l2 = l2, l1
+		}
+	}
+	return l1, l2, l3
+}
+
+// sides returns vectors for each of the sides of t.
+func (t Triangle) sides() (Vec, Vec, Vec) {
+	return Sub(t[1], t[0]), Sub(t[2], t[1]), Sub(t[0], t[2])
+}
+
+// IsDegenerate returns true if all of triangle's vertices are
+// within tol distance of its longest side.
+func (t Triangle) IsDegenerate(tol float64) bool {
+	sides := [3]Vec{Sub(t[1], t[0]), Sub(t[2], t[1]), Sub(t[0], t[2])}
+	len2 := [3]float64{Norm2(sides[0]), Norm2(sides[1]), Norm2(sides[2])}
+	longLen := len2[0]
+	longIdx := 0
+	if len2[1] > longLen {
+		longLen = len2[1]
+		longIdx = 1
+	}
+	if len2[2] > longLen {
+		longIdx = 2
+	}
+	// calculate vertex distance from longest side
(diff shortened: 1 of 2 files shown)


sbinet (dismissed):

spatial/r2/triangle.go
```diff
@@ -0,0 +1,64 @@
...
+	a, b, c := t.orderedLengths()
+	A := (c + (b + a)) * (a - (c - b))
+	A *= (a + (c - b)) * (c + (b - a))
+	return math.Sqrt(A) / 4
+}
+
+// Centroid returns the intersection of the three medians of the triangle
+// as a point in space.
+func (t Triangle) Centroid() Vec {
+	return Scale(1./3., Add(Add(t[0], t[1]), t[2]))
```
kortschak, line 37: s!1./3.!1.0/3.0!

spatial/r2/triangle.go
```diff
@@ -0,0 +1,64 @@
...
+	// https://stackoverflow.com/questions/33037449/given-three-side-of-a-triangle-how-can-i-define-whether-it-is-a-degenerate-tria
+	a, b, c := t.orderedLengths()
+	// The sum of the shorter sides of a valid triangle is always
+	// larger than the longest of the sides.
+	return a+b < c+tol
+}
+
+// Area returns the surface area of the triangle.
+func (t Triangle) Area() float64 {
+	// William M. Kahan (24 March 2000). "Miscalculating Area and Angles of a Needle-like Triangle"
```
kortschak, line 27: Add a link, https://people.eecs.berkeley.edu/~wkahan/Triangle.pdf, and which formula was used. I can't find a doi for the paper.

spatial/r2/triangle.go
```diff
@@ -0,0 +1,64 @@
...
+// Triangle represents a triangle in 2D space and
+// is composed by 3 vectors corresponding to the position
+// of each of the vertices.
+type Triangle [3]Vec
+
+// Degenerate returns true if triangle's vertices are collinear to within
+// a certain tolerance.
+func (t Triangle) Degenerate(tol float64) bool {
+	// TODO(soypat): Must be a better way to do this.
+	// https://stackoverflow.com/questions/33037449/given-three-side-of-a-triangle-how-can-i-define-whether-it-is-a-degenerate-tria
```
kortschak, line 18: How does this compare to finding the cross product of two vertices relative to the third, finding the area or comparing the slope of two pairs? It would be nice to have a measure where the tolerance is intuitively understandable.

spatial/r2/triangle.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+// sides returns vectors for each of the sides of the Triangle.
+func (t Triangle) sides() (Vec, Vec, Vec) {
+	return Sub(t[1], t[0]), Sub(t[2], t[1]), Sub(t[0], t[2])
+}
+
+// orderedLengths returns the lengths of the sides of the triangle such that
+// a ≤ b ≤ c.
+func (t Triangle) orderedLengths() (a, b, c float64) {
```
kortschak, line 47: This can be simpler by using swaps and sort-3.

spatial/r2/triangle.go
```diff
@@ -0,0 +1,64 @@
...
+// Centroid returns the intersection of the three medians of the triangle
+// as a point in space.
+func (t Triangle) Centroid() Vec {
+	return Scale(1./3., Add(Add(t[0], t[1]), t[2]))
+}
+
+// sides returns vectors for each of the sides of the Triangle.
+func (t Triangle) sides() (Vec, Vec, Vec) {
+	return Sub(t[1], t[0]), Sub(t[2], t[1]), Sub(t[0], t[2])
+}
```
kortschak, line 43: Callees are their callers, though in this case, I'm not sure this is needed.

spatial/r2/triangle_test.go
```diff
@@ -0,0 +1,93 @@
...
+package r2
+
+import (
+	"math"
+	"testing"
+
+	"golang.org/x/exp/rand"
+)
+
+func TestTriangleDegenerate(t *testing.T) {
```
kortschak, line 14: I think this can be done more clearly by constructing triangles with known degeneracy from vector projection from a random line.

kortschak (changes requested):

spatial/r2/triangle.go
soypat: I think checking the area lends itself to user error, as a long, needle-like triangle could have a larger area than a well-formed small triangle. I would vouch for `tol` having some spatial tie, i.e. triangle is degenerate if a cylinder of radius `tol` is able to contain all it's vertices. This would then be easy to test and intuitively understand- and I think I have some idea on how to program it. What do you think?

kortschak: > triangle is degenerate if a cylinder of radius `tol` is able to contain all of its vertices

I don't think this is right; an arbitrarily long major side with two minors that sum to |major|+eps  is degenerate to approximately eps.

A either of the longest sides or the shortest sides would give you something where that could be reasonably safely normalised with the mean of the sides' lengths (choose longest or shortest based on whether the triangle is a needle or a pancake).

soypat: > A either of the longest sides or the shortest sides would give you something where that could be reasonably safely normalised with the mean of the sides' lengths (choose longest or shortest based on whether the triangle is a needle or a pancake).

Also I'm not sure I understand

kortschak: I'm just trying to come up with a reasonable metric for degeneracy. Another alternative is to find the distance between the longest edge and its opposite vertex. This has a very clear intuitive meaning.

soypat: I like that metric. I've included it in the r3 PR and works as expected.

https://github.com/gonum/gonum/pull/1792/files#diff-8e1c51bb20ff0be4cc795a9701e444f915453f31dcfba6c4bd8ea348a1ebe71aR27

spatial/r2/triangle_test.go
kortschak: Define a pair of points, then pick a middle and project a vector of known length perpendicular to the line at the middle point for the third vertex.

soypat: Aha, that way we test for the "cap" type of degenerate triangle. Alright, I'll add tests for "cap" and "needle" degenerate triangles.

kortschak (commented):
Comments for the r3 version can be applied here.

kortschak (approved):

kortschak (approved):

PR #1798, opened 2022-05-09
mat: gofmt -w -r 'Matrix -> MatrixT' mat

soypat:
[Related Mail-list discussion](https://groups.google.com/g/gonum-dev/c/9-Fvth2ME1Q/m/AEIvUs02DwAJ?utm_medium=email&utm_source=footer). Summed up: `T()` method is not necessary to represent a matrix and can be done away with. Benefits of this are felt in libraries that provide types that implement `mat.Matrix`. Without removing the `T()` method these types will have varying gonum versions in the future and will [harm compilation speed](https://github.com/soypat/mat-import-woes). Implementing `mat.Matrix` today adds 52000 lines of code that must be resolved by the compiler.

This is a **test** to show the scale of changes if the Transpose method were removed from the mat.Matrix interface and a new mat.MatrixT interface was implemented with the same functionality as the previous mat.Matrix type.

This approach would allow gonum to gradually phase out the use of mat.MatrixT in the arguments to the functions and methods in the `mat` package with the goal of eventually replacing all `mat.MatrixT` uses with `mat.Matrix` and using a package wide `TransposeThisMatrixOverHere` or `T` function where the transpose were needed.

```go
func T(m Matrix) Matrix {
	if u, ok := m.(Untransposer); ok {
		return u.Untranspose()
	}
	if u, ok := m.(Transposer); ok {
		return u.T()
	}
	return Transpose{m}
}
```

#### Steps to reproduce

In main gonum package:
```shell
# First change all uses of Matrix interface
gofmt -w -r 'mat.Matrix -> mat.MatrixT' * 
# We now adjust mat package, renaming Matrix to MatrixT
gofmt -w -r 'Matrix -> MatrixT' mat
# Gonum should now be in a semi-consolidated state except
# for cases where Matrix was an embedded field in a struct
gofmt -w -r 'a.Matrix -> a.MatrixT' graph/spectral
# Uses are fixed except for struct declarations with the embedded type:
# i.e. return Laplacian{Matrix: l, Nodes: nodes, Index: indexOf}

# Fix other embedded struct uses.
gofmt -w -r 'a.Matrix  -> a.MatrixT' stat/mds/mds_test.go 
# Some files have mixed use of Matrix fields and Matrix methods
# such as graph/undirect_test.go, this file had to be edited manually
# along with some other 4 or 5 files. This meant adding the T suffix
# to all Matrix embedded struct access.
```

#### How do users catch up?
The following gofmt command seems to fix the usual use of the transpose method with the proposed new mat package function. For more complex uses a `go fix` alternative could be devised.
```shell
gofmt -w -r 'a.T() -> mat.T(a)' *
```


kortschak:
I'm going to restate my opposition to this.

vladimir-ch:
I'm closing this PR because it has gone stale and there was an opposition to the change.

PR #1801, opened 2022-05-10
ci: replace golangci-lint with staticcheck

kortschak:
Please take a look.


lapack/testlapack/general.go
```diff
@@ -1113,6 +1113,7 @@ func isSchurCanonicalGeneral(t blas64.General) bool {
 
 // schurBlockEigenvalues returns the two eigenvalues of the 2×2 matrix [a b; c d]
 // that must be in Schur canonical form.
+//lint:ignore U1000 This is useful for debugging.
```
kortschak, line 1116: @vladimir-ch I think this is why it was left here. Should it be removed?

sbinet (approved):

lapack/testlapack/general.go
kortschak: I'm going to merge, so if this is not right, we can fix it later.

Issue #1797, opened 2022-05-08
proposal: spatial manipulation matrices


soypat:
### Background
Most (if not all) 3D renderers use spatial transformation matrices. For example, here is the documentation of [three.js' Matrix4](https://threejs.org/docs/#api/en/math/Matrix4) type which allows for rotations and translations. Similarly, the corresponding matrix for 2D rotation+translation is a [3x3 matrix (three.js link)](https://threejs.org/docs/#api/en/math/Matrix3).

Today there is a 3x3 spatial matrix implementation, the `r3.Mat` type. This type was thought of for use as a rotation tensor, which describes a complete rotation in 3D space, without translation. This is useful for studying rigid-body dynamics of arbitrarily oriented frames in space, such as is the case with planes and drones. However, its functionality is a far cry from a `Matrix4` type, which brings rotation and translation into the equation, something which is crucial for 3D rendering.

These matrices have been implemented in `sdfx` by Jason Harris in Go: https://github.com/soypat/sdf/blob/main/matrix.go. The "important" methods this matrix should have for the correct functioning of the `sdf` package are as follows for the Matrix4 type:
* `func NewMatrix4(data []float64) Matrix4`
* `func (a Matrix4) MulPosition(b r3.Vec) r3.Vec`
* `func (a Matrix4) Inverse() Matrix4`
* `func (a Matrix4) Det() float64`
* `func (a Matrix4) MulBox(box r3.Box) r3.Box`  rotates/translates a 2d bounding box and resizes for axis-alignment.
* `func (a Matrix4) Scale(k float64) Matrix4` multiplies all values by k and returns new matrix

`Matrix3` should have corresponding methods of the same name too for `r2` types.

### Proposal
Add a similar-to Matrix3 and Matrix4 types somewhere in the `spatial` package. My guess of the package name where it belongs is at best as good as yours (`r2`, `mat2`, `mat`?) .

The code shown above uses "direct" types- they should probably be pointers to the Matrix though (?).

I am in no hurry to implement these so this proposal can serve as a floor to discuss the idea.

### Potential impact of proposal
3D renderer libraries are happy with gonum!

soypat:
I went out of my way to create a minimal quasi-working example of what the `Matrix3` type could look like: https://github.com/soypat/sdf/blob/main/internal/d2/transform.go

kortschak:
[...]

PR #1802, opened 2022-05-16
spatial/r3: add Transform

soypat:
Please take a look.

Resolves part of https://github.com/gonum/gonum/issues/1797.


spatial/r3/transform.go
```diff
@@ -0,0 +1,267 @@
...
+	return Transform{
+		d00: a[0] - 1, x01: a[1], x02: a[2], x03: a[3],
+		x10: a[4], d11: a[5] - 1, x12: a[6], x13: a[7],
+		x20: a[8], x21: a[9], d22: a[10] - 1, x23: a[11],
+		x30: a[12], x31: a[13], x32: a[14], d33: a[15] - 1,
+	}
+}
+
+// ComposeTransform creates a new transform for a given translation to
+// positon, scaling vector scale and quaternion rotation.
```
kortschak, line 61: position

spatial/r3/transform.go
```diff
@@ -0,0 +1,267 @@
...
+	//  if T == (Transform{})
+	d00, x01, x02, x03 float64
+	x10, d11, x12, x13 float64
+	x20, x21, d22, x23 float64
+	x30, x31, x32, d33 float64
+}
+
+// Transform applies the Transform to the argument vector
+// and returns the result.
+func (t Transform) Transform(v Vec) Vec {
```
kortschak, line 29: `Apply`?

spatial/r3/transform.go
```diff
@@ -0,0 +1,267 @@
...
+	t.d22 = (1-(xx+yy))*scale.Z - 1
+
+	t.x03 = position.X
+	t.x13 = position.Y
+	t.x23 = position.Z
+	return t
+}
+
+// Translate adds Vec to the positional Transform.
+func (t Transform) Translate(v Vec) Transform {
```
kortschak, line 98: `Add`?

spatial/r3/transform.go
```diff
@@ -0,0 +1,267 @@
...
+
+	t.x02 *= factor.Z
+	t.x12 *= factor.Z
+	t.d22 = (t.d22+1)*factor.Z - 1
+	t.x32 *= factor.Z
+	return t
+}
+
+// Mul multiplies the Transforms a and b and returns the result.
+// This is the equivalent of combining two transforms in one.
```
kortschak, line 135: I think this could be clearer. The text as it stands does not indicate that the operation is non-commutative.

spatial/r3/transform.go
```diff
@@ -0,0 +1,267 @@
...
+func (t Transform) Mul(b Transform) Transform {
+	if t == (Transform{}) {
```
kortschak, line 137: I think an `isZero` method would make these clearer.

spatial/r3/transform.go
```diff
@@ -0,0 +1,267 @@
...
+// If matrix is singular then Inv() returns the zero transform.
+func (t Transform) Inv() Transform {
+	if t == (Transform{}) {
+		return t
+	}
+	det := t.Det()
+	if math.Abs(det) < 1e-16 {
+		return zeroTransform
+	}
+	// Do something if singular?
```
kortschak, line 202: This troubles me. In mat we've gone out of our way to make it possible to avoid needing to calculate the inverse of matrices.

spatial/r3/transform.go
```diff
@@ -0,0 +1,267 @@
...
+	return Transform{
+		d00: t.d00, x01: t.x10, x02: t.x20, x03: t.x30,
+		x10: t.x01, d11: t.d11, x12: t.x21, x13: t.x31,
+		x20: t.x02, x21: t.x12, d22: t.d22, x23: t.x32,
+		x30: t.x03, x31: t.x13, x32: t.x23, d33: t.d33,
+	}
+}
+
+// equals tests the equality of the Transforms to within a tolerance.
+func (t Transform) equals(b Transform, tolerance float64) bool {
```
kortschak, line 239: I think this belongs in the test file. I'd also refer to use a more standard distance measure. If `Transform` implemented `mat.Matrix` this would just fall out.

spatial/r3/transform.go
```diff
@@ -0,0 +1,267 @@
...
+		math.Abs(t.x23-b.x23) < tolerance &&
+		math.Abs(t.x30-b.x30) < tolerance &&
+		math.Abs(t.x31-b.x31) < tolerance &&
+		math.Abs(t.x32-b.x32) < tolerance &&
+		math.Abs(t.d33-b.d33) < tolerance
+}
+
+// SliceCopy returns a copy of the Transform's data
+// in row major storage format. It returns 16 elements.
+func (t Transform) SliceCopy() []float64 {
```
kortschak, line 260: This also doesn't need to be exposed.

spatial/r3/transform.go
```diff
@@ -0,0 +1,267 @@
...
+
+package r3
+
+import (
+	"math"
+)
+
+// Transform represents a 3D spatial transformation.
+// The zero value of Transform is the identity transform.
+type Transform struct {
```
kortschak, line 13: What about making this structured like the `Mat` type rather than having distinct fields for each element?

spatial/r3/transform.go
soypat: Gioui uses `Transform`, I like `Apply` as well.

spatial/r3/transform.go
soypat: Gioui uses [Offset](https://pkg.go.dev/gioui.org/f32#Affine2D.Offset) and [three.js uses Translation](https://threejs.org/docs/index.html?q=matrix#api/en/math/Matrix4.makeTranslation) and [setPosition](https://threejs.org/docs/index.html?q=matrix#api/en/math/Matrix4.setPosition). I think there is a preference to not use `Add` in the industry for this operation since it lends itself to be confused with ordinary matrix additon.

spatial/r3/transform.go
soypat: Sounds good. The zero transform is the identity transform so these would be `isIdentity`

spatial/r3/transform.go
soypat: I'm not sure how I'd go about avoiding this calculation without using the `mat` package which would make the implementation noticeably slower. I'll add the benchmark source in this next push
```
goos: linux
goarch: amd64
pkg: gonum.org/v1/gonum/spatial/r3
cpu: AMD Ryzen 5 3400G with Radeon Vega Graphics    
BenchmarkTransformInverse-8      	12232880	        93.51 ns/op	       0 B/op	       0 allocs/op
BenchmarkMatTransformInverse-8   	  438561	      2564 ns/op	      96 B/op	       4 allocs/op
PASS
ok  	gonum.org/v1/gonum/spatial/r3	5.583s
```

spatial/r3/transform.go
soypat: Right now this PR makes no assumption about the underlying data type and is pretty fast, maybe owing to the SSA optimizations made possible by having each element be a field of the struct. The changes you suggest would require making substantial changes. I'd like to leave this to a future PR that 
* demonstrates that an array representation does not substantially slow down the implementation
* Maintains the zero value Transform is the identity transform

kortschak: SGTM

spatial/r3/transform.go
kortschak: And yet the docs use the "add" to describe the action it takes. I would call it add and explain in the docs with reference to translate and/or offset. The type system lets the user know that it's not ordinary matrix addiition, though note that really, it is, just with a different in-memory representation of a translation-only transform matrix.

soypat: I guess I can convince myself `Add` is fine. If I were a new user to gonum I think I'd be confused by the `Add` method and wonder how I change the Translate part of the transform but hey, it should be easy enough to get used to. Will change it

soypat: Another question, what method do we use to set the positional part of the transform to a known value?

spatial/r3/transform.go
```diff
@@ -0,0 +1,259 @@
...
+		t.d00 + 1, t.x01, t.x02, t.x03,
+		t.x10, t.d11 + 1, t.x12, t.x13,
+		t.x20, t.x21, t.d22 + 1, t.x23,
+		t.x30, t.x31, t.x32, t.d33 + 1,
+	}
+}
+
+// isIdentity returns true if receiver is the identity Transform.
+func (t Transform) isIdentity() bool {
+	return t == (Transform{})
```
kortschak, line 258: The parens are not needed now.

spatial/r3/transform.go
kortschak: `SetPosition`?

spatial/r3/transform.go
```diff
@@ -0,0 +1,259 @@
+// Copyright ©2022 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package r3
+
+import (
+	"math"
+)
```
kortschak, line 9: One line.

spatial/r3/transform.go
```diff
@@ -0,0 +1,259 @@
...
+	x30, x31, x32, d33 float64
+}
+
+// Apply applies the Transform to the argument vector
+// and returns the result.
+func (t Transform) Apply(v Vec) Vec {
+	if t.isIdentity() {
+		return v
+	}
+	// https://github.com/mrdoob/three.js/blob/dev/src/math/Vector3.js#L262
```
kortschak, line 33: This is not a permalink.

spatial/r3/transform.go
```diff
@@ -0,0 +1,259 @@
...
+
+// NewTransform returns a new Transform type and populates its elements
+// with values passed in row-major form. If val is nil then NewTransform
+// returns a Transform filled with zeros.
+func NewTransform(a []float64) Transform {
+	if a == nil {
+		return zeroTransform
+	}
+	if len(a) != 16 {
+		panic("Transform is initialized with 16 values")
```
kortschak, line 56: This is ambiguous; `panic(fmt.Sprintf("r3: NewTransform invalid slice length: %d != 16", len(a)))`.

spatial/r3/transform.go
```diff
@@ -0,0 +1,259 @@
...
+		x20: a[8], x21: a[9], d22: a[10] - 1, x23: a[11],
+		x30: a[12], x31: a[13], x32: a[14], d33: a[15] - 1,
+	}
+}
+
+// ComposeTransform creates a new transform for a given translation to
+// position, scaling vector scale and quaternion rotation.
+// The identity Transform is constructed with
+//  ComposeTransform(Vec{}, Vec{1,1,1}, Rotation{})
+func ComposeTransform(position, scale Vec, q Rotation) Transform {
```
kortschak, line 70: Suggest `position` => `translate`.

soypat: If `translate` describes that value better then shouldn't the method for setting it be `SetTranslate`?

soypat:
@kortschak I've added methods `IsZero()` and `SetTranslate`. IsZero lets users check result of `Inv()`. I'm also on the fence about the `Add(v Vec)` signature. I feel like `Translate` fits in so well with how we've named the argument to  `ComposeTransform` and the `SetTranslate` method.

spatial/r3/transform.go
kortschak: Absolutely

spatial/r3/affine.go
```diff
@@ -0,0 +1,316 @@
...
+//
+// The zero value of Warp corresponds to the identity transform, thus the diagonal
+// elements of Warp are stored subtracted 1. This means that a transformation
+// that scales all components of a vector by 3 is constructed as follows:
+//  w := Warp{XX:2, YY:2, ZZ:2}
+type Warp struct {
+	XX, XY, XZ float64
+	YX, YY, YZ float64
+	ZX, ZY, ZZ float64
+}
```
kortschak, line 300: I'm not a fan of this. It leaves all the work up to the user and is an end-run around the work that's gone in to not exporting the elements of `Affine`. I agree that the complexity of representing skew is high, but I would prefer that this be how it's represented. Probably `type Skew struct { XY, XZ, YX, YZ, ZX, ZY float64 }`.

kortschak:
> Should `SetTranslate` return a new transform instead of setting the receiver? If so, `SetTranslate` feels off for the name.

Yes, the transforms should be immutable. I agree `SetTranslate` is not right. I'll think about this on the weekend.

spatial/r3/affine.go
soypat: > It leaves all the work up to the user

I'm not sure I understand- what work do you mean? 

> end-run around the work that's gone in to not exporting the elements of Affine.

The Affine's top non-zero 3x3 elements may or may not encode Warp data. They are functionally different.

> `type Skew struct { XY, XZ, YX, YZ, ZX, ZY float64 }`

This identifier is already taken by the skew symmetric matrix constructor.

kortschak: > I'm not sure I understand- what work do you mean?

There is work that needs to be done by the user to make sure they don't shoot themselves in the foot; this is confusing:

"This means that a transformation that scales all components of a vector by 3 is constructed as follows: `w := Warp{XX:2, YY:2, ZZ:2}`"

The reason is clear when you know the code, but it will be the cause of issues.

It would be better to not have both scaling and skew together in the API. A `Mat` with the note that the diagonal is ignored would work, 

> This identifier is already taken by the skew symmetric matrix constructor.

This is unfortunate. What was the reason for introducing the skew-symmetric function? (it wasn't in the proposal that provoked the PR).

I'm starting to worry about the complexity of the API that is growing here.

soypat: > Mat with the note that the diagonal is ignored would work,

OK, this can work.

> This is unfortunate. What was the reason for introducing the skew-symmetric function? (it wasn't in the proposal that provoked the PR).

See https://github.com/gonum/gonum/issues/1754. Skew was in the proposal. In hindsight it should have been a method - this would have simplified package-level API and also prevented allocations on every Skew call. The reason I included it as a function was that the `Skew` operation to me was a basic vector commodity, much like `Cross`.

> I'm starting to worry about the complexity of the API that is growing here.

It might be a little early to set `ComposeAffine` in stone- I'm still learning a lot about Affine transforms. We can leave it for a future PR.

soypat:
I suggest making a compromise- removing API which we are uncertain about quality and leaving in the bits which seem OK. I believe the most shaky bits are
* Warp type
* Compose Affine
* Warp method
These can be left for a future PR after we have more knowledge on how Affine is used.

kortschak:
This seems like a reasonable approach.

spatial/r3/affine.go
```diff
@@ -0,0 +1,255 @@
...
+	x30, x31, x32, d33 float64
+}
+
+// Transform applies the transform to the argument vector and returns the result.
+func (a Affine) Transform(v Vec) Vec {
+	if a.isIdentity() {
+		return v
+	}
+	// See https://github.com/mrdoob/three.js/blob/019fa1ad671a1ffcf9be5828efd518fb06575c2b/src/math/Vector3.js#L263.
+	w := 1 / (a.x30*v.X + a.x31*v.Y + a.x32*v.Z + a.d33 + 1)
```
vladimir-ch, line 34: What kind of transformations does Affine represent? The 4x4 nature and the code here suggest that perspective projection is included. Is it intended? I'd say maybe it's out of scope?

Issue #1803, opened 2022-05-17
Add quaternion spherical interpolation function


soypat:
This is a common function used for computer graphics.

https://en.wikipedia.org/wiki/Slerp

A tried and tested implementation can be found at [three.js](https://github.com/mrdoob/three.js/blob/14030014c00a5a0f0e2b052d9b6773a6c49055f2/src/math/Quaternion.js#L535). I can help contribute the feature.

kortschak:
With quats, slerp is not that hard; all you need to do is take the n-th root of a transform for an n-step slerp from the start to the end of the transform. The same for the dual quat and dual complex. I don't think this is needed. You can demonstrate this with minor changes to https://pkg.go.dev/gonum.org/v1/gonum/num/quat#example-package-Rotate.

soypat:
So like `nthRoot:= quat.Number{Real:math.Pow(q.Real, 1/n), Imag: math.Pow(q.Imag, 1/n), Kmag:math.Pow(q.Kmag, 1/n),Jmag:math.Pow(q.Jmag, 1/n)}` ? I can't find anything on this algorithm with a cursory google search.

kortschak:
No, `quat.Pow(q, quat.Inv(quat.Number{Real: n}))`. I'll find some papers.

PR #1805, opened 2022-05-21
num/quat: add slerp example

soypat:
@kortschak Thanks for the papers. I think the Simo Särkkä "Notes on Quaternions" does something of the style which you speak of, taking the $t=i/n$ root.
$$ p(t) = (q_1 \cdot q_0^{-1})^t \cdot q_0$$

I used the `r3.Rotation` type to add a helpful pointer to the `r3` package.

Resolves https://github.com/gonum/gonum/issues/1803


num/quat/slerp_example_test.go
```diff
@@ -0,0 +1,51 @@
...
+	"gonum.org/v1/gonum/num/quat"
+	"gonum.org/v1/gonum/spatial/r3"
+)
+
+// slerp returns the spherical interpolation between q0 and q1
+// for t in [0,1].
+func slerp(r0, r1 r3.Rotation, t float64) r3.Rotation {
+	q0 := quat.Number(r0)
+	q1 := quat.Number(r1)
+	// Simo Särkkä "Notes on Quaternions", June 28, 2007
```
kortschak, line 20:
```
// Based on Simo Särkkä "Notes on Quaternions" Eq. 35
// https://users.aalto.fi/~ssarkka/pub/quat.pdf 
```

I'd also suggest that the initial orientation to the q0 basis be explained.

num/quat/slerp_example_test.go
```diff
@@ -0,0 +1,51 @@
...
+	q1 = quat.Mul(q1, quat.Inv(q0))
+	return r3.Rotation(quat.Mul(quat.Exp(quat.Scale(t, quat.Log(q1))), q0))
+}
+
+// Spherically interpolate between two quaternions to obtain a rotation.
+func Example_slerp() {
+	const nsamples = 10
+	noRot := r3.NewRotation(0, r3.Vec{})
```
kortschak, line 28: Given that the base case where there is no initial rotation is a lot simpler than when there is, it would be good to show that this does something. So I'd suggest having the initial rotation be pi/4 around x.

Also rename `noRot` => `rot0` and `xRot` => `rot1`.

num/quat/slerp_example_test.go
```diff
@@ -0,0 +1,51 @@
...
+	xRot := r3.NewRotation(math.Pi, r3.Vec{X: 1})
+	v := r3.Vec{X: 1, Y: 1, Z: 1}
+	for i := 0; i <= nsamples; i++ {
+		t := float64(i) / nsamples
+		interpRot := slerp(noRot, xRot, t)
+		rotated := interpRot.Rotate(v)
```
kortschak, line 34:
```
rotated := slerp(rot0, rot1, t).Rotate(v)
```
If this were being used on a collection of points it would be worth retaining, but here I think it's not needed.

num/quat/slerp_example_test.go
```diff
@@ -0,0 +1,51 @@
+// Copyright ©2022 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package quat_test
```
kortschak, line 5: I'm wondering with this doesn't belong in r3 rather than quat. I know that you are using this to link to r2, but I think that would be better done with a sentence in the package doc.

num/quat/slerp_example_test.go
```diff
@@ -0,0 +1,51 @@
...
+	q0 := quat.Number(r0)
+	q1 := quat.Number(r1)
+	// Simo Särkkä "Notes on Quaternions", June 28, 2007
+	q1 = quat.Mul(q1, quat.Inv(q0))
+	return r3.Rotation(quat.Mul(quat.Exp(quat.Scale(t, quat.Log(q1))), q0))
+}
+
+// Spherically interpolate between two quaternions to obtain a rotation.
+func Example_slerp() {
+	const nsamples = 10
```
kortschak, line 27: s/nsamples/steps/g (not actually samples)

num/quat/slerp_example_test.go
```diff
@@ -0,0 +1,51 @@
...
+	noRot := r3.NewRotation(0, r3.Vec{})
+	xRot := r3.NewRotation(math.Pi, r3.Vec{X: 1})
+	v := r3.Vec{X: 1, Y: 1, Z: 1}
+	for i := 0; i <= nsamples; i++ {
+		t := float64(i) / nsamples
```
kortschak, line 32:
```
	for i := 0.0; i <= steps; i++ {
		t := i / steps
```

num/quat/slerp_example_test.go
```diff
@@ -0,0 +1,51 @@
...
+)
+
+// slerp returns the spherical interpolation between q0 and q1
+// for t in [0,1].
+func slerp(r0, r1 r3.Rotation, t float64) r3.Rotation {
+	q0 := quat.Number(r0)
+	q1 := quat.Number(r1)
+	// Simo Särkkä "Notes on Quaternions", June 28, 2007
+	q1 = quat.Mul(q1, quat.Inv(q0))
+	return r3.Rotation(quat.Mul(quat.Exp(quat.Scale(t, quat.Log(q1))), q0))
```
kortschak, line 22: Yeah, nicer than `quat.Pow(q1, quat.Num{Real: t}`. (Maybe quat should have a `PowReal` added since this is likely more common than q1^q2.)

num/quat/slerp_example_test.go
soypat: I agree this is a better fit for `r3`. will move it over.

vladimir-ch:
@soypat Do you still want to work on this PR and address @kortschak review comments? Otherwise we can close it because it's getting a bit stale.

soypat:
@vladimir-ch @kortschak Applied suggestions! Let me know if there's anything left

num/quat/slerp_example_test.go
kortschak: There is a real power function now. https://pkg.go.dev/gonum.org/v1/gonum/num/quat#PowReal

kortschak (approved):
Thanks

PR #1806, opened 2022-05-22
num/quat: add PowReal function

kortschak:
This brings the quat package into line with the others in num, which already have `PowReal`. See https://github.com/gonum/gonum/pull/1805#discussion_r878771435 for motivation.

Please take a look.


soypat:
The API signature and implementation looks good to me.

sbinet (approved):

PR #1811, opened 2022-06-20
interp: convert panics to errors

stevenh:
Return errors instead of panicing in all cases for the interp package so its more consumable.

### DIFF
--- a/interp/cubic.go
+++ b/interp/cubic.go
@@ -74,26 +74,21 @@ func (pc *PiecewiseCubic) PredictDerivative(x float64) float64 {
 
 // FitWithDerivatives fits a piecewise cubic predictor to (X, Y, dY/dX) value
 // triples provided as three slices.
-// It panics if len(xs) < 2, elements of xs are not strictly increasing,
+// It errors if len(xs) < 2, elements of xs are not strictly increasing,
 // len(xs) != len(ys) or len(xs) != len(dydxs).
-func (pc *PiecewiseCubic) FitWithDerivatives(xs, ys, dydxs []float64) {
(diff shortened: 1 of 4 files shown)


kortschak (changes requested):
The principle used for the choice of panics v errors in Gonum is that a function will return an error is it is not possible for the programmer to know whether the call will fail but if they can know a failure can happen it will be through a panic — this is the case in these calls where the panics are for slice length mismatch. This places the use of panics within the realms of what happens on programmer error (and is less severe than the behaviour in reflect).

stevenh:
Thanks for detailing the logic, but that means callers that consume external data need to replicate the checks already done by the package to avoid a panic.

Typical idiomatic use of panic is for an internal programming issue that API caller can do nothing about, with errors for everything else including argument validation. A notable exception is functions / methods which it would be inconvenient to add an error return parameter, but in this case those already exist.

With this in mind would be be willing to reconsider?

kortschak:
> Typical idiomatic use of panic is for an internal programming issue that API caller can do nothing about, with errors for everything else including argument validation.

I don't completely agree with this. Panics are for programmer error, there is nothing that says who the programmer is. The reflect package is an excellent example of this. The line that we draw is similar to the line drawn by reflect; typish things must match to make up for domain-specific weaknesses in the Go type system (e.g. length matches here)

I doubt this will change. The context behind this decision is that Gonum is intended primarily for scientific computing and decisions are coloured by experiences in that domain. Functions that *only* return an error (or panic in our model) are the *least* likely to be changed since these are the cases that we most commonly see being ignored and so leading to silent failure. If you have time and want to understand the kind of thing that motivates this, I would suggest you watch [this video](https://www.youtube.com/watch?v=7gYIs7uYbMo).

Issue #1809, opened 2022-05-26
spatial/r3: missed opportunity on a Transformer interface?


soypat:
Is it too late to consider the following change to r3? Feel like a huge opportunity was missed when naming the Rotate method for Rotation. This would really simplify things on my end for the `sdf` package.

```go
// Could be defined on user side, no need for it to be in r3
type Transformer interface {
    Apply(Vec) Vec
}

var _ Transformer = Rotation{}
var _ Transformer = Transform{}
```

kortschak:
When the package was designed it was not intended to be something bigger than storing spatial information for statistical methods and rendering static spatial data; this explains the reason that it is the way that it is. It looks like it is growing significantly in scope with the new additions. If it is going to grow into a generalised set of spatial data packages (I'm not opposed to this, but it brings additional complexity and already seems to be happening) then how it looks should be discussed.

On the plus side of the discussion, it adds the possibility of easily supporting dualquat/dualcomplex transformations to r2 and r3. I think the name of the interface needs work.

kortschak:
On names, I'd suggest `Transform` => `Affine` and then make the interface `type Transformer interface { Transform(Vec) Vec }`.

To start the discussion, it would be helpful if you could outline where it is that you want to take the package(s) (if r3 moves, then r2 needs to as well to maintain API congruence) and what are the motivations for the extension. This can happen at gonum-dev if you want to flesh things out.

soypat:
> `Transform` => `Affine`

Yes! I had been thinking just that the other day. I'll apply this change to the `Transform` PR one of these days.

As for the end goal/future additions I do have a few comments:

I envision gonum/r3 as being a solid but basic foundation to build 3D data manipulation programs.

#### About the shape types
1. `Triangle` type is a much needed for computational geometry. It is the bread and butter of surface representation
2. `Tetrahedron` type brings in the basic `r3` volumetric representation. Yes, `Box` does volumes but is very limited and cannot represent arbitrary volumes. I'd even argue that `r3.Tetrahedron`, not `r3.Triangle`, is the `r3` complement to `r2.Triangle`.

[...]

PR #1816, opened 2022-07-21
r3.Mat: add Hessian

soypat:
Please take a look.

### DIFF
--- a/spatial/r3/mat.go
+++ b/spatial/r3/mat.go
@@ -198,8 +198,9 @@ func (m *Mat) Det() float64 {
 	return a*deta - b*detb + c*detc
 }
 
-// Hessian sets the receiver to the Hessian matrix at point p of a scalar
-// field. The field is also evaluated at points in the area surrounding p by adding
+// Hessian sets the receiver to the Hessian matrix of the scalar field at the point p,
+// approximated using finite differences with the given step sizes.
+// The field is evaluated at points in the area surrounding p by adding
 // at most 2 components of step to p. Hessian expects the field's second partial
 // derivatives are all continuous for correct results.
 func (m *Mat) Hessian(p, step Vec, field func(Vec) float64) {
--- a/spatial/r3/mat.go
+++ b/spatial/r3/mat.go
@@ -198,39 +198,40 @@ func (m *Mat) Det() float64 {
 	return a*deta - b*detb + c*detc
 }
 
-// Hessian sets the receiver to the Hessian matrix of a scalar field
-// evaluated at point p. Additional points evaluated to calculate hessian
-// are spaced at most tol away from p.
-func (m *Mat) Hessian(p, step Vec, fn func(Vec) float64) {
+// Hessian sets the receiver to the Hessian matrix at point p of a scalar
+// field. The field is also evaluated at points in the area surrounding p by adding
+// at most 2 components of step to p. Hessian expects the field's second partial
+// derivatives are all continuous for correct results.
+func (m *Mat) Hessian(p, step Vec, field func(Vec) float64) {
 	dx := Vec{X: step.X}
 	dy := Vec{Y: step.Y}
 	dz := Vec{Z: step.Z}
-	fp := fn(p)
-	fxp := fn(Add(p, dx))
-	fxm := fn(Sub(p, dx))
+	fp := field(p)
+	fxp := field(Add(p, dx))
+	fxm := field(Sub(p, dx))
 	fxx := (fxp - 2*fp + fxm) / (step.X * step.X)
 
-	fyp := fn(Add(p, dy))
-	fym := fn(Sub(p, dy))
+	fyp := field(Add(p, dy))
+	fym := field(Sub(p, dy))
 	fyy := (fyp - 2*fp + fym) / (step.Y * step.Y)
 
 	aux := Add(dx, dy)
-	fxyp := fn(Add(p, aux))
-	fxym := fn(Sub(p, aux))
+	fxyp := field(Add(p, aux))
+	fxym := field(Sub(p, aux))
 	fxy := (fxyp - fxp - fyp + 2*fp - fxm - fym + fxym) / (2 * step.X * step.Y)
 
-	fzp := fn(Add(p, dz))
-	fzm := fn(Sub(p, dz))
+	fzp := field(Add(p, dz))
+	fzm := field(Sub(p, dz))
 	fzz := (fzp - 2*fp + fzm) / (step.Z * step.Z)
 
 	aux = Add(dx, dz)
-	fxzp := fn(Add(p, aux))
-	fxzm := fn(Sub(p, aux))
+	fxzp := field(Add(p, aux))
+	fxzm := field(Sub(p, aux))
 	fxz := (fxzp - fxp - fzp + 2*fp - fxm - fzm + fxzm) / (2 * step.X * step.Z)
 
 	aux = Add(dy, dz)
-	fyzp := fn(Add(p, aux))
-	fyzm := fn(Sub(p, aux))
+	fyzp := field(Add(p, aux))
+	fyzm := field(Sub(p, aux))
 	fyz := (fyzp - fyp - fzp + 2*fp - fym - fzm + fyzm) / (2 * step.Y * step.Z)
 
 	m.Set(0, 0, fxx)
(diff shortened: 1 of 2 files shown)


spatial/r3/mat.go
```diff
@@ -197,3 +197,52 @@ func (m *Mat) Det() float64 {
...
+
+	fyp := f(Add(p, dy))
+	fym := f(Sub(p, dy))
+	fyy := (fyp - 2*fp + fym) / (tol.Y * tol.Y)
+
+	var aux Vec
+	aux = Add(dx, dy)
+	fxyp := f(Add(p, aux))
+	fxym := f(Sub(p, aux))
+	fxy := (fxyp - fxp - fyp + 2*fp - fxm - fym + fxym) / (2 * tol.X * tol.Y)
```
sbinet, line 221:
```suggestion
	fxy := (fxyp - fxp - fyp + 2*fp - fxm - fym + fxym) * 0.5 * iy2
```

sbinet (commented):
just a drive-by review.

spatial/r3/mat.go
```diff
@@ -197,3 +197,52 @@ func (m *Mat) Det() float64 {
...
+	fxp := f(Add(p, dx))
+	fxm := f(Sub(p, dx))
+	fxx := (fxp - 2*fp + fxm) / (tol.X * tol.X)
+
+	fyp := f(Add(p, dy))
+	fym := f(Sub(p, dy))
+	fyy := (fyp - 2*fp + fym) / (tol.Y * tol.Y)
+
+	var aux Vec
+	aux = Add(dx, dy)
```
kortschak, line 218:
```
	aux := Add(dx, dy)
```

spatial/r3/mat.go
```diff
@@ -197,3 +197,52 @@ func (m *Mat) Det() float64 {
 	detc := m.At(1, 0)*m.At(2, 1) - m.At(1, 1)*m.At(2, 0)
 	return a*deta - b*detb + c*detc
 }
+
+// Hessian sets the receiver to the hessian matrix of a spatial scalar field
```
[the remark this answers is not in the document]
vladimir-ch: And I'd say that "spatial" is redundant.

spatial/r3/mat.go
```diff
@@ -197,3 +197,52 @@ func (m *Mat) Det() float64 {
...
+// evaluated at point p. Additional points evaluated to calculate hessian
+// are spaced at most tol away from p.
+func (m *Mat) Hessian(p, tol Vec, f func(Vec) float64) {
```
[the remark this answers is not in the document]
vladimir-ch: I'd rename `tol` -> `step`.

spatial/r3/mat_test.go
```diff
@@ -230,3 +230,59 @@ func BenchmarkQuat(b *testing.B) {
...
+				return NewMat([]float64{
+					24 * p.X, 0, 0,
+					0, 10, 0,
+					0, 0, 36 * p.Z * p.Z,
+				})
+			},
+		},
+		{
+			f: func(p Vec) float64 {
+				// 4*x^3 + 5*y^2 + 3*z^4
```
vladimir-ch, line 261: Wrong comment.

spatial/r3/mat_test.go
```diff
@@ -230,3 +230,59 @@ func BenchmarkQuat(b *testing.B) {
...
+func TestMatHessian(t *testing.T) {
+	const (
+		tol = 1e-3
+		h   = 5e-5
+	)
+	step := Vec{X: h, Y: h, Z: h}
+	rnd := rand.New(rand.NewSource(1))
+	for _, Case := range []struct {
+		f      func(p Vec) float64
+		expect func(p Vec) *Mat
```
vladimir-ch, line 243: It's conventional to call this `want`.

spatial/r3/mat_test.go
```diff
@@ -230,3 +230,59 @@ func BenchmarkQuat(b *testing.B) {
...
+					-y4 * cx * sz, -4 * y3 * sx * sz, -y4 * sx * cz,
+					-4 * y3 * sx * sz, 12 * p.Y * p.Y * cx * sz, 4 * y3 * cx * cz,
+					-y4 * sx * cz, 4 * y3 * cx * cz, -y4 * cx * sz,
+				})
+			},
+		},
+	} {
+		for i := 0; i < 30; i++ {
+			p := randomVec(rnd)
+			m := NewMat(nil)
```
vladimir-ch, line 280: This would be typically called `got`.

spatial/r3/mat.go
soypat: This may be invalid. The cross partial derivative of x and y requires being divided by both the steps in the x and y direction, which may not be equal. From https://en.wikipedia.org/wiki/Finite_difference#Multivariate_finite_differences

spatial/r3/mat_test.go
```diff
@@ -230,3 +230,59 @@ func BenchmarkQuat(b *testing.B) {
 		}
 	}
 }
+
+func TestMatHessian(t *testing.T) {
+	const (
+		tol = 1e-3
```
vladimir-ch, line 236: In fact, with `h = 8e-4` I get the test to pass with a more stringent tolerance `tol = 5e-6`. However, that's quite tight and could cause flakiness in the test, so using `tol = 1e-5` is enough in my opinion.

soypat: @vladimir-ch good find. added!

spatial/r3/mat_test.go
```diff
@@ -230,3 +230,60 @@ func BenchmarkQuat(b *testing.B) {
...
+	const (
+		tol = 1e-5
+		h   = 8e-4
+	)
+	step := Vec{X: h, Y: h, Z: h}
+	rnd := rand.New(rand.NewSource(1))
+	for _, Case := range []struct {
+		fn   func(p Vec) float64
+		want func(p Vec) *Mat
+	}{
```
kortschak, line 244:
```suggestion
	for _, test := range matHessianTests {
```
and move the table up into package scope.

spatial/r3/mat.go
```diff
@@ -197,3 +197,51 @@ func (m *Mat) Det() float64 {
 	detc := m.At(1, 0)*m.At(2, 1) - m.At(1, 1)*m.At(2, 0)
 	return a*deta - b*detb + c*detc
 }
+
+// Hessian sets the receiver to the Hessian matrix of a scalar field
+// evaluated at point p. Additional points evaluated to calculate hessian
+// are spaced at most tol away from p.
+func (m *Mat) Hessian(p, step Vec, fn func(Vec) float64) {
```
kortschak, line 204: `fn` needs an explanation in the godoc.

spatial/r3/mat_test.go
```diff
@@ -230,3 +230,63 @@ func BenchmarkQuat(b *testing.B) {
...
+}
+
+func TestMatHessian(t *testing.T) {
+	const (
+		tol = 1e-5
+		h   = 8e-4
+	)
+	step := Vec{X: h, Y: h, Z: h}
+	rnd := rand.New(rand.NewSource(1))
+	for _, Case := range scalarFields {
```
kortschak, line 280: s/Case/test/g

spatial/r3/mat.go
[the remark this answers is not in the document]
kortschak: I'm going to reverse myself here now that the docs have been improved and I understand what is trying to be said better (this applies in the divergence PR too, but I'll add a comment there). Rather than `f` or `fn`, I think `field` documents it better, then the name parameter name doesn't need to be mentioned at all.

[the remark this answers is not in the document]
vladimir-ch: And once you're at it, I'd simplify the information about evaluating in the neighborhood of p to just a mention of central differences, as in Divergence. And I'd completely avoid mentioning the mathematical assumptions on continuity - they are already implied by the fact that we're computing the Hessian. And what is "correct" when we're computing only an approximation (and in floating point)? They are not directly related to the implementation and so are out of scope. 

I'd write something like:
```
// Hessian sets the receiver to the Hessian matrix of the scalar field at the point p,
// approximated using finite differences with the given step sizes.
```

kortschak (approved):
LGTM but please wait for @vladimir-ch

vladimir-ch (approved):
The doc comment for Hessian hasn't been modified but it can be done later.

PR #1817, opened 2022-07-26
spatial/r3: add Divergence

soypat:
Please take a look.

### DIFF
--- a/spatial/r3/vector.go
+++ b/spatial/r3/vector.go
@@ -83,6 +83,18 @@ func Cos(p, q Vec) float64 {
 	return Dot(p, q) / (Norm(p) * Norm(q))
 }
 
+// Divergence returns the divergence of the vector field at the point p,
+// approximated using finite differences with the given step sizes.
+func Divergence(p, step Vec, field func(Vec) Vec) float64 {
+	sx := Vec{X: step.X}
+	divx := (field(Add(p, sx)).X - field(Sub(p, sx)).X) / step.X
+	sy := Vec{Y: step.Y}
+	divy := (field(Add(p, sy)).Y - field(Sub(p, sy)).Y) / step.Y
+	sz := Vec{Z: step.Z}
+	divz := (field(Add(p, sz)).Z - field(Sub(p, sz)).Z) / step.Z
+	return 0.5 * (divx + divy + divz)
+}
+
 // minElem return a vector with the minimum components of two vectors.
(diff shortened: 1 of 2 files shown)


spatial/r3/vector_test.go
```diff
@@ -254,6 +256,25 @@ func TestRotate(t *testing.T) {
...
 }
 
+func TestDivergence(t *testing.T) {
+	const (
+		tol = 1e-10
+		h   = 1e-2
+	)
+	step := Vec{X: h, Y: h, Z: h}
+	rnd := rand.New(rand.NewSource(1))
+	for _, Case := range vectorFields {
```
kortschak, line 266: s/Case/test/g

spatial/r3/vector_test.go
```diff
@@ -288,3 +309,32 @@ func vecApproxEqual(a, b Vec, tol float64) bool {
 		scalar.EqualWithinAbs(a.Y, b.Y, tol) &&
 		scalar.EqualWithinAbs(a.Z, b.Z, tol)
 }
+
+var vectorFields = []struct {
```
kortschak, line 313: Please put this in front of the test function. Are you planning to use this for other tests? If not please s/vectorFields/divergenceTests/.

soypat: I'll move them. I am planning on using them for other tests that require vector fields. i.e. jacobian calculation.

spatial/r3/vector.go
```diff
@@ -83,6 +83,18 @@ func Cos(p, q Vec) float64 {
 	return Dot(p, q) / (Norm(p) * Norm(q))
 }
 
+// Divergence returns the volume density of the outward flux of a vector
```
vladimir-ch, line 86: The current comment unnecessarily focuses on the physical interpretation of the divergence. A user of the function presumably knows what the divergence is. I'd replace it with something like

```
// Divergence returns the divergence of the vector field fn at the point p,
// approximated using central differences with step sizes given in step.
```

kortschak: With reference to https://github.com/gonum/gonum/pull/1816#discussion_r930844923, s/fn // in the comment and change the parameter name to `field`.

kortschak (approved):
LGTM but please wait for @vladimir-ch

vladimir-ch (approved):

PR #1818, opened 2022-07-26
spatial/r3: Skew is now method

soypat:
I understand this has not been talked about, but it's clearly bad API design in my eyes. What do y'all think?

Changes:
* deprecate Skew package level function in favor of method to prevent heap allocations

### DIFF
--- a/spatial/r3/mat.go
+++ b/spatial/r3/mat.go
@@ -197,3 +197,20 @@ func (m *Mat) Det() float64 {
 	detc := m.At(1, 0)*m.At(2, 1) - m.At(1, 1)*m.At(2, 0)
 	return a*deta - b*detb + c*detc
 }
+
+// Skew sets the receiver to the 3×3 skew symmetric matrix
+// (right hand system) of v.
+//                  ⎡ 0 -z  y⎤
+//  Skew({x,y,z}) = ⎢ z  0 -x⎥
+//                  ⎣-y  x  0⎦
+func (m *Mat) Skew(v Vec) {
+	m.Set(0, 0, 0)
+	m.Set(0, 1, -v.Z)
+	m.Set(0, 2, v.Y)
+	m.Set(1, 0, v.Z)
+	m.Set(1, 1, 0)
+	m.Set(1, 2, -v.X)
+	m.Set(2, 0, -v.Y)
+	m.Set(2, 1, v.X)
+	m.Set(2, 2, 0)
+}
--- a/spatial/r3/mat_safe.go
+++ b/spatial/r3/mat_safe.go
@@ -60,6 +60,8 @@ func Eye() *Mat {
 //                  ⎡ 0 -z  y⎤
 //  Skew({x,y,z}) = ⎢ z  0 -x⎥
 //                  ⎣-y  x  0⎦
+//
+// DEPRECATED: use Mat.Skew()
 func Skew(v Vec) (M *Mat) {
 	return &Mat{&array{
 		0, -v.Z, v.Y,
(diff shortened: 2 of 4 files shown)


spatial/r3/mat_safe.go
```diff
@@ -60,6 +60,8 @@ func Eye() *Mat {
 //                  ⎡ 0 -z  y⎤
 //  Skew({x,y,z}) = ⎢ z  0 -x⎥
 //                  ⎣-y  x  0⎦
+//
+// DEPRECATED: use Mat.Skew()
```
vladimir-ch, line 64: @kortschak It's a matter of convenience but couldn't we have both?

spatial/r3/mat.go
```diff
@@ -197,3 +197,20 @@ func (m *Mat) Det() float64 {
 	detc := m.At(1, 0)*m.At(2, 1) - m.At(1, 1)*m.At(2, 0)
 	return a*deta - b*detb + c*detc
 }
+
+// Skew sets the receiver to the 3×3 skew symmetric matrix
+// (right hand system) of v.
+//                  ⎡ 0 -z  y⎤
+//  Skew({x,y,z}) = ⎢ z  0 -x⎥
+//                  ⎣-y  x  0⎦
+func (m *Mat) Skew(v Vec) {
```
vladimir-ch, line 206: What is the utility of this matrix? If it's to compute its product with a vector, then why not do it more efficiently with `Cross`? Are there other uses?

spatial/r3/mat_safe.go
kortschak: We've deprecated things from r2 and r3 in the past (though looking at the package I see that they are still there). There is an weird asymmetry in the deprecation (justified by implementation) in that the previous deprecation was a change from method to package function and this one is from package function to method (the implementation detail is that those are non-heap allocating returns, while this one is a heap allocating return). @sbinet?

spatial/r3/mat.go
vladimir-ch: I guess the answer is "composition".

soypat: One use I have here lying around is computing a rotation tensor from a quaternion

```go
// RotationFromQuat stores the rotation matrix from a quaternion (must be normalized).
//  [R] = q.Real*q.Real * [E] - dot([q],[q])*[E] + 2*[q]*[q]ᵀ + 2*q.Real * skew([q])
// where
//  [q] = {q.Imag, q.Jmag, q.Kmag}
// and
//  [E]=Eye()
func (m *Mat) RotationFromQuat(q quat.Number) {
	qv := r3.Vec{X: q.Imag, Y: q.Jmag, Z: q.Kmag}
	qs := Skew(qv)
	q01 := Identity()
	q01.Scale(q.Real*q.Real, q01)

	qd := Identity()
	qd.Scale(r3.Dot(qv, qv), qd)

	qs.Scale(2*q.Real, qs)

	qv = r3.Scale(2, qv)
	m.Prod(qv, qv) // m = 2*[q]*[q]ᵀ
	m.Add(m, q01)  // m += q.Real*q.Real * [E]
	m.Add(m, qd)   // m += dot([q],[q])*[E]
	m.Add(m, qs)   // m += 2*q.Real * skew([q])
}
```

The literature I follow for rigid body dynamics uses 3x3 matrices to store angular velocities, though I've still have not gotten as far as to using them in that form.

vladimir-ch (approved):

kortschak:
@soypat This will need conflicts to be resolved.

kortschak (approved):

PR #1822, opened 2022-07-30
spatial/r3: add Jacobian

soypat:
Please take a look.

### DIFF
--- a/spatial/r3/mat.go
+++ b/spatial/r3/mat.go
@@ -264,3 +264,28 @@ func (m *Mat) Hessian(p, step Vec, field func(Vec) float64) {
 	m.Set(2, 1, fyz)
 	m.Set(2, 2, fzz)
 }
+
+// Jacobian sets the receiver to the Jacobian matrix of the vector field at the point p,
+// approximated using finite differences with the given step sizes.
+// Jacobian expects the field's first order partial
+// derivatives are all continuous for correct results.
+func (m *Mat) Jacobian(p, step Vec, field func(Vec) Vec) {
+	dx := Vec{X: step.X}
+	dy := Vec{Y: step.Y}
+	dz := Vec{Z: step.Z}
+
+	dfdx := Scale(0.5/step.X, Sub(field(Add(p, dx)), field(Sub(p, dx))))
+	dfdy := Scale(0.5/step.Y, Sub(field(Add(p, dy)), field(Sub(p, dy))))
+	dfdz := Scale(0.5/step.Z, Sub(field(Add(p, dz)), field(Sub(p, dz))))
+	m.Set(0, 0, dfdx.X)
+	m.Set(0, 1, dfdy.X)
+	m.Set(0, 2, dfdz.X)
+
+	m.Set(1, 0, dfdx.Y)
+	m.Set(1, 1, dfdy.Y)
+	m.Set(1, 2, dfdz.Y)
+
+	m.Set(2, 0, dfdx.Z)
+	m.Set(2, 1, dfdy.Z)
+	m.Set(2, 2, dfdz.Z)
+}
--- a/spatial/r3/vector_test.go
+++ b/spatial/r3/vector_test.go
@@ -279,18 +279,18 @@ var vectorFields = []struct {
 	},
 	{
 		field: func(v Vec) Vec {
-			// (x*y*z*cos(y), y*z+sin(x), z*x/sin(y))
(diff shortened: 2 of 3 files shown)


spatial/r3/vector_test.go
```diff
@@ -259,17 +259,27 @@ func TestRotate(t *testing.T) {
...
+			return NewMat([]float64{
+				v.Y * v.Z, v.X * v.Z, v.X * v.Y,
+				0, v.Z, v.Y,
+				v.Z, 0, v.X,
+			})
+		},
 	},
 	{
 		field: func(v Vec) Vec {
+			// (x*y*z*cos(y), y*z+sin(x), z*x/sin(y))
```
vladimir-ch, line 282: I think it's better to use continuous functions for testing to avoid false negatives. The test points are generated randomly and it could happen that `y` is close to a multiple of π.

kortschak (approved):
LGTM for API and docs, but waiting for @vladimir-ch.

spatial/r3/mat.go
```diff
@@ -264,3 +264,28 @@ func (m *Mat) Hessian(p, step Vec, field func(Vec) float64) {
...
+// derivatives are all continuous for correct results.
+func (m *Mat) Jacobian(p, step Vec, field func(Vec) Vec) {
+	dx := Vec{X: step.X}
+	dy := Vec{Y: step.Y}
+	dz := Vec{Z: step.Z}
+
+	dfdx := Scale(0.5/step.X, Sub(field(Add(p, dx)), field(Sub(p, dx))))
+	dfdy := Scale(0.5/step.Y, Sub(field(Add(p, dy)), field(Sub(p, dy))))
+	dfdz := Scale(0.5/step.Z, Sub(field(Add(p, dz)), field(Sub(p, dz))))
+	m.Set(0, 0, dfdx.X)
```
vladimir-ch, line 280: I don't have time to check it myself but I think it would be good to make sure that this function and `fd.Jacobian` use the same convention for the Jacobian, that is, that it's _not_ the case that one returns J and the other J^T.

soypat: Got distracted these past weeks, here's a test I wrote- I can add it to the PR if we decide it is worth adding. They are not transposes of eachother.

```go

func TestFDandR3JacobianConsolidation(t *testing.T) {
	const step = 8e-4
	const tol = 1e-3
	h := Vec{step, step, step}
	for _, test := range vectorFields {
		p := randomVec(rnd)
		px := []float64{p.X, p.Y, p.Z}
		fdf := func(y, x []float64) {
			v := test.field(Vec{x[0], x[1], x[2]})
			y[0], y[1], y[2] = v.X, v.Y, v.Z
		}
		d := mat.NewDense(3, 3, nil)
		fd.Jacobian(d, fdf, px, &fd.JacobianSettings{
			Step: step,
		})
		m := NewMat(nil)
		m.Jacobian(p, h, test.field)
		if !mat.EqualApprox(d, m, tol) {
			t.Error("")
			t.Errorf("fd and r3 package Jacobian not consolidated\ngot:  %v\nwant:  %v",
				mat.Formatted(m), mat.Formatted(d))
		}
	}
}
```
Result:
```
$ go test -timeout 30s -run ^TestFDandR3JacConsolidation$ gonum.org/v1/gonum/spatial/r3

ok  	gonum.org/v1/gonum/spatial/r3
```

vladimir-ch: Let's just document explicitly how the Jacobian is oriented, the test is not necessary.

kortschak: Would this be OK?
```
// The Jacobian matrix J is the matrix of all first-order partial derivatives of f.
// If f maps an 3-dimensional vector x to an 3-dimensional vector y = f(x), J is
// a 3×3 matrix whose elements are given as
//
//	J_{i,j} = ∂f_i/∂x_j,
//
// or expanded out
//
//	    [ ∂f_1/∂x_1 ∂f_1/∂x_2 ∂f_1/∂x_3 ]
//	J = [ ∂f_2/∂x_1 ∂f_2/∂x_2 ∂f_2/∂x_3 ]
//	    [ ∂f_3/∂x_1 ∂f_3/∂x_2 ∂f_3/∂x_3 ]
```
to reflect the description used in fd?

vladimir-ch: Yes, OK! Thanks.

kortschak: I'll merge this and add that.

PR #1826, opened 2022-08-09
delete minor unreachable code caused by t.Fatal

Abirdcfly:
Signed-off-by: Abirdcfly <fp544037857@gmail.com>

Please take a look.


kortschak:
Thank you. I don't think this is the correct fix; the continue should be executed and an error raised. I'll send that.

PR #1839, opened 2022-11-24
lapack/gonum: propagate NaN from Dgecon

vladimir-ch:
Applying similar changes as the reference in https://github.com/Reference-LAPACK/lapack/pull/765

Please take a look.

### DIFF
--- a/lapack/gonum/dgecon.go
+++ b/lapack/gonum/dgecon.go
@@ -18,11 +18,11 @@ import (
 //
 // The slice a contains the result of the LU decomposition of A as computed by Dgetrf.
 //
-// anorm is the corresponding 1-norm or ∞-norm of the original matrix A.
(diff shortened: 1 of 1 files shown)


lapack/gonum/dgecon.go
```diff
@@ -51,6 +53,9 @@ func (impl Implementation) Dgecon(norm lapack.MatrixNorm, n int, a []float64, ld
 	if anorm == 0 {
 		return 0
 	}
+	if math.IsNaN(anorm) {
+		return anorm
```
kortschak, line 57: I think it's worth having a comment here explaining that while the reference implementation treats the NaN anorm as invalid, we don't since it's not helpful to do so here.

vladimir-ch: Done.

kortschak (approved):

PR #1840, opened 2022-11-24
lapack/gonum: improve accuracy of Dlartg

vladimir-ch:
This incorporates modifications introduced in LAPACK 3.11.0 in https://github.com/Reference-LAPACK/lapack/pull/631

Please take a look.


kortschak (dismissed):

kortschak:
It looks like the images in graph/layout will need to be regenerated for amd64 (and likely 386). I can do the arm64 ones in a separate PR.

kortschak:
I think the easiest is if I push those fixes here.

kortschak (approved):

PR #1841, opened 2022-11-30
docs: remove duplicated badge

haoxins:
godoc is go.dev now


kortschak:
I'd like to keep a non-go.dev docs link. For a while godoc.org worked, now it should be changed to godocs.io. I'll do this when the next release happens. Thanks

PR #1850, opened 2023-01-28
graph/path: reduce allocations in path reconstructions

### LINKED ISSUE #1848: Performance of /graph/path/shortest.go/allBetween is not great
My app does network simulations. Doing so, I call the `AllBetween` method of `AllShortest` a lot.

### What did you do?

Doing some profiling, I observed that it was one of the main reason the code was slow.

Gonum to be blazing fast :)

### What actually happened?
image
The calculation for this was taken from around 4h of simulations. As you can see, the allBetween accounts for almost 3h of execution, most of it due to the `growslice` runtime.

I seems that all the `append()` (or slicing ?) calls make this slow.
Maybe it could be rewritten using a generator pattern (channel, or func ?).

```
> go version
go version go1.19.4 darwin/amd64

> grep gonum.org/v1/gonum go.sum
gonum.org/v1/gonum v0.12.0 h1:xKuo6hzt+gMav00meVPUlXwSdoEJP46BR+wdxQEFK2o=
gonum.org/v1/gonum v0.12.0/go.mod h1:73TDxJfAAHeA8Mk9mf8NlIppyhQNo5GLTcYeqgo2lvY=
```

Code is the same, so I don't think it would change anything.

kortschak:
Requires careful review. Considering adding a naive Dijkstra implementation to the test to have as a comparator.

Please take a look.

Fixes #1848


kortschak:
*pin*

vladimir-ch:
Sorry, I forgot about this one. I'll try to take a look this week.

graph/path/bellman_ford_moore_test.go
```diff
@@ -170,6 +170,41 @@ func TestBellmanFordAllFrom(t *testing.T) {
...
 			}
 
+			paths = paths[:0]
+			pt.AllToFunc(test.Query.To().ID(), func(path []graph.Node) {
+				paths = append(paths, append([]graph.Node(nil), path...))
+			})
+			if weight != test.Weight {
+				t.Errorf("%q %s: unexpected weight from AllTo: got:%f want:%f",
+					test.Name, tg.typ, weight, test.Weight)
+			}
```
sbinet, line 180: why do we re-test this ?
didn't we test it at `L142` ?

(I am not advocating for removing it, mind you. I am just trying to see whether I am missing something)

graph/path/shortest.go
```diff
@@ -420,38 +420,67 @@ func (p ShortestAlts) AllTo(vid int64) (paths [][]graph.Node, weight float64) {
...
 	if from == to {
 		if path == nil {
-			return paths
+			return
 		}
 		ordered.Reverse(path)
-		return append(paths, path)
+		fn(path)
+		ordered.Reverse(path)
+		return
```
sbinet, line 466: isn't there a possible change of semantics here ?
before the PR, we (seemed to) reverse the order of nodes in place and then copy inside of `paths`.
but now, we (probably) copy into a user-owned `paths` the original order of nodes, and then reverse the order of nodes of `path`, but AFAICT, the order of nodes in the user-owned `paths` isn't altered. is it ?

do we test for this `from==to` case in the set of tests ?

graph/path/shortest.go
```diff
@@ -420,38 +420,67 @@ func (p ShortestAlts) AllTo(vid int64) (paths [][]graph.Node, weight float64) {
...
-	paths = p.allTo(from, to, seen, []graph.Node{p.nodes[to]}, nil)
+	p.allTo(from, to, seen, []graph.Node{p.nodes[to]}, func(path []graph.Node) {
+		paths = append(paths, append([]graph.Node(nil), path...))
+	})
 	weight = p.dist[to]
 
 	return paths, weight
 }
 
+// AllToFunc calls fn all shortest paths to v. Paths containing zero-weight
```
sbinet, line 431:
```suggestion
// AllToFunc calls fn on all shortest paths to v. Paths containing zero-weight
```

graph/path/bellman_ford_moore_test.go
kortschak: Yeah, this can go away.

graph/path/shortest.go
kortschak: We have a partial path here, so we show it to the caller in the order they are expecting it, allow them to look at it, and then put it back to how we need it to continue working with it. This is why they may not hold it after they return from their closure.

graph/path/shortest.go
kortschak: Done

sbinet (approved):

PR #1856, opened 2023-03-04
all: replace godoc.org with pkg.go.dev in badges and links

antichris:
This pull request does the following:

- replace all godoc.org README badges godocs.io ones
- replace remaining (non-badge) godoc.org links with pkg.go.dev ones
- fix a link to the `blas64` docs in the `doc.go` of `lapack/gonum`, that had "golang.org" instead of "gonum.org" in the package path
- add pkg.go.dev reference README badges to all packages that have the godocs.io ones

## Bakground

What caught my eye initially was that the main README had two seemingly identical "go reference" badges. It turned out that one of them was the legacy godoc.org reference which has been redirecting to pkg.go.dev since early 2021 (see https://github.com/golang/go/issues/43178)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -36,7 +36,7 @@ Where possible, the source of algorithms should be referenced in the comments.
 
 Here are the current repositories for the Gonum project.
 If code you want to contribute doesn't quite fit in any of them, then please start a discussion on the [mailing list](https://groups.google.com/forum/#!forum/gonum-dev).
-Code can be found at [github.com/gonum/](https://github.com/gonum/)\<repo\>, and documentation at godoc.org/github.com/gonum/\<repo\>.
+Code can be found at [github.com/gonum/](https://github.com/gonum/)\<repo\>, and documentation at gonum.org/v1/\<repo\>.
 
 * [gonum](https://github.com/gonum/gonum) — The gonum repository contains the majority of Gonum packages
 * [plot](https://github.com/gonum/plot) — A repository for plotting and visualizing data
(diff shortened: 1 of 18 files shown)


CONTRIBUTING.md
```diff
@@ -36,7 +36,7 @@ Where possible, the source of algorithms should be referenced in the comments.
 
 Here are the current repositories for the Gonum project.
 If code you want to contribute doesn't quite fit in any of them, then please start a discussion on the [mailing list](https://groups.google.com/forum/#!forum/gonum-dev).
-Code can be found at [github.com/gonum/](https://github.com/gonum/)\<repo\>, and documentation at godoc.org/github.com/gonum/\<repo\>.
+Code can be found at [github.com/gonum/](https://github.com/gonum/)\<repo\>, and documentation at pkg.go.dev/github.com/gonum/\<repo\>.
```
kortschak, line 39: This is not right (it has been so for a long time). The doc links are not relative to github.com/gonum, so this should be `gonum.org/v1/\<repo\>`

README.md
```diff
@@ -4,7 +4,6 @@
 [![Build status](https://ci.appveyor.com/api/projects/status/valslkp8sr50eepn/branch/master?svg=true)](https://ci.appveyor.com/project/Gonum/gonum/branch/master)
 [![codecov.io](https://codecov.io/gh/gonum/gonum/branch/master/graph/badge.svg)](https://codecov.io/gh/gonum/gonum)
 [![go.dev reference](https://pkg.go.dev/badge/gonum.org/v1/gonum)](https://pkg.go.dev/gonum.org/v1/gonum)
-[![GoDoc](https://godoc.org/gonum.org/v1/gonum?status.svg)](https://godoc.org/gonum.org/v1/gonum)
```
kortschak, line 7: Resurrect this as a godocs.io link.

kortschak (commented):
Thanks. I had intended to make a change from godoc.org  to godocs.io (https://github.com/gonum/gonum/pull/1841#issuecomment-1334831850). You've done more than I had thought was necessary, so I'm happy to accept this if the front page README has its godoc.org link replaced with a godocs.io link.

README.md
antichris: Shouldn't then all of the package README's that used to link to godoc.org also now link to godocs.io instead of pkg.go.dev? Or should they have both badges, just like here?

kortschak: I think that battle was lost a long time ago. By putting the link to godocs.io on the front page we give people the option to avoid pkg.go.dev if they want to. I'd say that most people do not use the old-repo level READMEs to direct their doc journey so it's less important to provide the link there. If you want to do the work, I'm happy with that, it just didn't seem as important to me. Thanks.

antichris: Done.

kortschak: Sorry for the delay.

kortschak:
Thank you for a well organised PR. Would you please add yourself to A+C with a commit message "A+C: add Chris Nobody" (or if you are comfortable adding your real name — if indeed that isn't — use that).

README.md
antichris: All's fine, things happen.

antichris:
> Would you please add yourself to A+C

I hope this is good enough. I would've preferred to remain anonymous, though, especially considering the lack of significance in my contribution. As I see it, when you pick up a candy wrapper to put it in the bin, you really don't expect your name in the newspapers for that. :sweat_smile:

kortschak:
:smile: Anonymity is acceptable. We reward all contribution with acknowledgement. Thank you.

kortschak (approved):
Thanks

PR #1862, opened 2023-04-27
graph/formats/dot/internal: regenerate with latest goccmack/gocc

kortschak:
Make use of go.mod to pin gocc version and go mod tidy.

Please take a look.


mewmew (approved):
LGTM.

PR #1863, opened 2023-05-02
graph/path: allow cost-based Yen shortest path calculation

kortschak:
This adds an additional constraint onto YenKShortestPaths to consider the paths' cost as well as the number of paths, enabling search within a constrained distance.

This is a breaking change.

Please take a look.

### DIFF
--- a/graph/path/yen_ksp.go
+++ b/graph/path/yen_ksp.go
@@ -12,9 +12,11 @@ import (
 	"gonum.org/v1/gonum/graph/iterator"
 )
 
-// YenKShortestPaths returns the k-shortest loopless paths from s to t in g.
-// YenKShortestPaths will panic if g contains a negative edge weight.
-func YenKShortestPaths(g graph.Graph, k int, s, t graph.Node) [][]graph.Node {
+// YenKShortestPaths returns the k-shortest loopless paths from s to t in g
+// with path costs no greater than cost beyond the shortest path.
+// If k is negative, only path cost will be used to limit the set of returned
+// paths. YenKShortestPaths will panic if g contains a negative edge weight.
+func YenKShortestPaths(g graph.Graph, k int, cost float64, s, t graph.Node) [][]graph.Node {
 	// See https://en.wikipedia.org/wiki/Yen's_algorithm and
 	// the paper at https://doi.org/10.1090%2Fqam%2F253822.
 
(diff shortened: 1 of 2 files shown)


graph/path/yen_ksp.go
```diff
@@ -5,16 +5,17 @@
...
 
 import (
-	"math"
 	"sort"
 
 	"gonum.org/v1/gonum/graph"
 	"gonum.org/v1/gonum/graph/iterator"
 )
 
 // YenKShortestPaths returns the k-shortest loopless paths from s to t in g.
```
vladimir-ch, line 14: No final dot, the sentence continues.

graph/path/yen_ksp.go
```diff
@@ -5,16 +5,17 @@
...
-// YenKShortestPaths will panic if g contains a negative edge weight.
-func YenKShortestPaths(g graph.Graph, k int, s, t graph.Node) [][]graph.Node {
+// with path costs no greater than cost relative to the shortest path.
```
vladimir-ch, line 15: I find the word "relative" here misleading as it evokes a ratio, a division, but there's none. I had to check the code and the tests to understand. Maybe replace it with "compared", or something like "with path cost differences no greater than cost compared to the shortest path"?

kortschak: I was thinking relative as opposed to absolute.

vladimir-ch: SGTM

graph/path/yen_ksp.go
```diff
@@ -70,7 +72,7 @@ func YenKShortestPaths(g graph.Graph, k int, s, t graph.Node) [][]graph.Node {
 			}
 
 			spath, weight := DijkstraFrom(spur, yk).To(t.ID())
-			if math.IsInf(weight, 1) {
+			if weight > cost {
```
kortschak, line 75: I think this should probably be 
```
			if weight > cost || math.IsInf(weight, 1) { {
```

kortschak:
*ping*

graph/path/yen_ksp.go
vladimir-ch: That would change the outcome only when weight=inf and cost=inf or nan. Both values of cost can be interpreted as "don't take cost into account", in which case the previous behavior should be (?) preserved, so adding `|| math.Inf(weight, 1)` would make sense.

kortschak:
PTAL

vladimir-ch (approved):

Issue #1866, opened 2023-06-13
panic in Sampleuv


khasanovbi:
### What are you trying to do?
I've got a panic when call `Weighted.Take`:

```
2023/06/13 03:57:52 http: panic serving 10.0.0.19:10934: runtime error: index out of range [5] with length 3
goroutine 5425 [running]:
net/http.(*conn).serve.func1()
        /opt/hostedtoolcache/go/1.20.4/x64/src/net/http/server.go:1854 +0xbf
panic({0xf1ffa0, 0xc00013a0d8})
        /opt/hostedtoolcache/go/1.20.4/x64/src/runtime/panic.go:890 +0x263
gonum.org/v1/gonum/stat/sampleuv.Weighted.Take({{0xc00013a0a8, 0x3, 0x3}, {0xc00013a0c0, 0x3, 0x3}, 0xc00095fc00})
        /home/runner/go/pkg/mod/gonum.org/v1/gonum@v0.13.0/stat/sampleuv/weighted.go:80 +0x345
```

### What did you do?
I wrote some fuzz tests and found some cases when this happen. Looks like this is floating numbers issue.
Some examples:
https://go.dev/play/p/MBWfsSZErYf
https://go.dev/play/p/62FWG3e8xHZ

### What version of Go and Gonum are you using?

gonum v0.13.0:
gonum.org/v1/gonum v0.13.0 h1:a0T3bh+7fhRyqeNbiC3qVHYmkiQgit3wnNan/2c0HMM=
gonum.org/v1/gonum v0.13.0/go.mod h1:/WPYRckkfWrhWefxyYTfrTtQR0KH4iyHNuzxqXAKyAU=

go: go version go1.20.5 linux/amd64

### Does this issue reproduce with the current master?

yes

PR #1867, opened 2023-06-13
Fix Weighted panic

khasanovbi:
Please take a look.

Take is divided into 2 steps index search and after that Reweight with 0.
1. I added checks for heap over bounds during tree traversal.
2. I removed last and left because in each iteration of the loop we always make a move to either the left or right subtree and this branch never reached.
3. Instead of two implementations of reweight in the `Take` and `Reweight` functions, I left only one.
4. Instead of calling reset - O(n) when you check a negative weight, I do a weight calculation as in reset, but through sift up only the desired nodes for O(log(n)). It also removed the `scalar.EqualWithinAbsOrRel` check, since we are now summing up zeros instead of accumulating a float error.

https://github.com/gonum/gonum/issues/1866

### DIFF
--- a/stat/sampleuv/weighted.go
+++ b/stat/sampleuv/weighted.go
@@ -4,9 +4,7 @@
 
 package sampleuv
 
-import (
-	"golang.org/x/exp/rand"
-)
+import "golang.org/x/exp/rand"
 
 // Weighted provides sampling without replacement from a collection of items with
 // non-uniform probability.
@@ -52,41 +50,6 @@ func NewWeighted(w []float64, src rand.Source) Weighted {
 // already taken.
 func (s Weighted) Len() int { return len(s.weights) }
 
-func (s Weighted) findIndex(r float64) int {
-	i := 0
-
-	for {
-		r -= s.weights[i]
-		if r < 0 {
-			break // Fall within item i.
-		}
-
-		ni := i<<1 + 1 // Move to left child.
-		// Left node should exist, because r is non-negative, but there are could be floating point errors,
-		// so we check index explicitly.
-		if ni >= len(s.heap) {
-			break
-		}
-
-		i = ni
-
-		d := s.heap[i]
-		if r >= d {
-			// If enough r to pass left child try to move to the right child.
-			r -= d
-			ni := i + 1
-
-			if ni >= len(s.heap) {
-				break
-			}
-
-			i = ni
-		}
-	}
-
-	return i
-}
-
 // Take returns an index from the Weighted with probability proportional
 // to the weight of the item. The weight of the item is then set to zero.
 // Take returns false if there are no items remaining.
@@ -102,7 +65,38 @@ func (s Weighted) Take() (idx int, ok bool) {
 		r = s.rnd.Float64()
 	}
 
-	i := s.findIndex(s.heap[0] * r)
+	r *= s.heap[0]
+	i := 0
+	for {
+		r -= s.weights[i]
+		if r < 0 {
+			break // Fall within item i.
+		}
+
+		li := i*2 + 1 // Move to left child.
+		// Left node should exist, because r is non-negative,
+		// but there could be floating point errors, so we
+		// check index explicitly.
+		if li >= len(s.heap) {
+			break
+		}
+
+		i = li
+
+		d := s.heap[i]
+		if r >= d {
+			// If there is enough r to pass left child try to
+			// move to the right child.
+			r -= d
+			ri := i + 1
+
+			if ri >= len(s.heap) {
+				break
+			}
+
+			i = ri
+		}
+	}
 
 	s.Reweight(i, 0)
 
@@ -113,10 +107,12 @@ func (s Weighted) Take() (idx int, ok bool) {
 func (s Weighted) Reweight(idx int, w float64) {
 	s.weights[idx] = w
 
+	// We want to keep the heap state here consistent
+	// with the result of a reset call. So we sum
+	// weights in the same order, since floating point
+	// addition is not associative.
 	for {
 		w = s.weights[idx]
-		// We want to keep heap state here like after reset call.
-		// Therefore, we sum weights in such order, because floating point sum not associative.
 
 		ri := idx*2 + 2
 		if ri < len(s.heap) {
--- a/stat/sampleuv/weighted.go
+++ b/stat/sampleuv/weighted.go
@@ -6,8 +6,6 @@ package sampleuv
 
 import (
 	"golang.org/x/exp/rand"
-
-	"gonum.org/v1/gonum/floats/scalar"
 )
 
 // Weighted provides sampling without replacement from a collection of items with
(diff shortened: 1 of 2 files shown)


kortschak:
I'll take a look on the weekend. From an initial look, I'd like more detail on the fix in the description and less code movement.

kortschak:
> can I change the golang version for staticcheck to 1.18?

Where do you want to do this?

khasanovbi:
> Where do you want to do this?

[Here](https://github.com/gonum/gonum/blob/master/.github/workflows/staticcheck.yml#L16). I want to set it to v1 to use v1.3.0

.github/workflows/staticcheck.yml
```diff
@@ -13,6 +13,6 @@ jobs:
     - uses: actions/checkout@v2
       with:
         fetch-depth: 1
-    - uses: dominikh/staticcheck-action@v1.2.0
+    - uses: dominikh/staticcheck-action@v1
```
kortschak, line 16: This should be in a separate PR.

khasanovbi: Moved this commit to https://github.com/gonum/gonum/pull/1868

kortschak: Thanks

stat/sampleuv/weighted.go
```diff
@@ -54,70 +52,89 @@ func NewWeighted(w []float64, src rand.Source) Weighted {
 // already taken.
 func (s Weighted) Len() int { return len(s.weights) }
 
+func (s Weighted) findIndex(r float64) int {
```
kortschak, line 55: Conventionally we place caller before callee, but in this case I'd prefer that this block be put back in `Take`; it's not called from more than one location and doesn't significantly complexify the calling function.

So, as
```
	r *= s.heap[0]
	i := 0
	for {
		r -= s.weights[i]
		if r < 0 {
			break // Fall within item i.
		}
...
```
after `r` decl in `Take`.

stat/sampleuv/weighted.go
```diff
@@ -54,70 +52,89 @@ func NewWeighted(w []float64, src rand.Source) Weighted {
...
+
+	for {
+		r -= s.weights[i]
+		if r < 0 {
+			break // Fall within item i.
+		}
+
+		ni := i<<1 + 1 // Move to left child.
+		// Left node should exist, because r is non-negative, but there are could be floating point errors,
+		// so we check index explicitly.
```
kortschak, line 66:
```suggestion
		// Left node should exist, because r is non-negative,
		// but there could be floating point errors, so we
		// check index explicitly.
```
(narrow comments are preferred within code)

stat/sampleuv/weighted.go
```diff
@@ -54,70 +52,89 @@ func NewWeighted(w []float64, src rand.Source) Weighted {
...
+		if ni >= len(s.heap) {
+			break
+		}
+
+		i = ni
+
+		d := s.heap[i]
+		if r >= d {
+			// If enough r to pass left child try to move to the right child.
```
kortschak, line 75:
```suggestion
			// If there is enough r to pass left child try to
			// move to the right child.
```

stat/sampleuv/weighted.go
```diff
@@ -54,70 +52,89 @@ func NewWeighted(w []float64, src rand.Source) Weighted {
...
-	idx++
-	for idx > 0 {
-		s.heap[idx-1] -= w
-		idx >>= 1
+	s.weights[idx] = w
+
+	for {
+		w = s.weights[idx]
+		// We want to keep heap state here like after reset call.
+		// Therefore, we sum weights in such order, because floating point sum not associative.
```
kortschak, line 119: Move this to above (attached to) the for loop as
```
	// We want to keep the heap state here consistent
	// with the result of a reset call. So we sum
	// weights in the same order, since floating point
	// addition is not associative.
```

stat/sampleuv/weighted_test.go
```diff
@@ -266,3 +266,48 @@ func TestWeightedNoResample(t *testing.T) {
 		}
 	}
 }
+
+func calculateNonZeroWeights(weights []float64) int {
```
kortschak, line 270: Please move this to below `FuzzWeightedTake`.

stat/sampleuv/weighted_test.go
```diff
@@ -266,3 +266,48 @@ func TestWeightedNoResample(t *testing.T) {
 		}
 	}
 }
+
```
kortschak, line 269: Please add a test for the cases in the issue
```
var issue1866Tests = []struct {
	weights []float64
	seed    uint64
}{
	{weights: []float64{72000, 166.66666666666666}, seed: 202},
	{weights: []float64{0.6666666666666666, 20412, 0}, seed: 0},
}

// See https://github.com/gonum/gonum/issues/1866
func TestIssue1866(t *testing.T) {
	for i, test := range issue1866Tests {
		w := NewWeighted(test.weights, rand.NewSource(test.seed))
		for j := 0; j < len(test.weights)+1; j++ {
			if panicked(func() { w.Take() }) {
				t.Errorf("unexpected panic for test %d iteration %d", i, j)
			}
		}
	}
}

func panicked(fn func()) (yes bool) {
	defer func() {
		if recover() != nil {
			yes = true
		}
	}()
	fn()
	return false
}
```

stat/sampleuv/weighted_test.go
```diff
@@ -266,3 +266,48 @@ func TestWeightedNoResample(t *testing.T) {
...
+	f.Fuzz(func(t *testing.T, w1 float64, w2 float64) {
+		if w1 < 0 || w2 < 0 {
+			t.Skip()
+		}
+
+		weights := []float64{w1, w2}
+
+		weighted := NewWeighted(weights, rand.NewSource(0))
+
+		expectedTakenSize := calculateNonZeroWeights(weights)
```
kortschak, line 293:
```suggestion
		got := calculateNonZeroWeights(weights)
```

stat/sampleuv/weighted_test.go
```diff
@@ -266,3 +266,48 @@ func TestWeightedNoResample(t *testing.T) {
...
+		}
+
+		if expectedTakenSize != len(takenSet) {
+			t.Errorf(
+				"unexpected taken set size: expected=%d, actual=%d, taken=%v",
+				expectedTakenSize,
+				len(takenSet),
+				takenSet,
+			)
+		}
```
kortschak, line 311: I don't think we need to keep a set, we expect n samples from an n-sized weighted, so let's use that. 

```suggestion
		want := calculateNonZeroWeights(weights)
		var got int
		for {
			_, ok := weighted.Take()
			if !ok {
				break
			}
			got++
		}

		if got != want {
			t.Errorf("unexpected taken set size: got=%d, want=%d", got, want)
		}
```

stat/sampleuv/weighted_test.go
```diff
@@ -266,3 +266,48 @@ func TestWeightedNoResample(t *testing.T) {
...
+	for _, w := range weights {
+		if w != 0 {
+			count++
+		}
+	}
+
+	return count
+}
+
+func FuzzWeighted_Take(f *testing.F) {
```
kortschak, line 281:
```suggestion
func FuzzWeightedTake(f *testing.F) {
```

kortschak (commented):
Thanks. Mainly style issues.

stat/sampleuv/weighted_test.go
khasanovbi: Fixed, but maybe it would be better to prefer subtests with t.Run?

stat/sampleuv/weighted_test.go
khasanovbi: Fixed

stat/sampleuv/weighted_test.go
khasanovbi: Fixed

stat/sampleuv/weighted_test.go
khasanovbi: Fixed

stat/sampleuv/weighted.go
khasanovbi: Fixed

stat/sampleuv/weighted.go
khasanovbi: Fixed

stat/sampleuv/weighted.go
khasanovbi: Fixed

stat/sampleuv/weighted.go
khasanovbi: Fixed

stat/sampleuv/weighted.go
```diff
@@ -6,8 +6,6 @@ package sampleuv
 
 import (
 	"golang.org/x/exp/rand"
-
-	"gonum.org/v1/gonum/floats/scalar"
 )
```
kortschak, line 9: Since this is now a single import
```
import "golang.org/x/exp/rand"
```

stat/sampleuv/weighted_test.go
kortschak: This will do; having the test table in the func body makes it more cluttered and the tests are simple enough not to justify sub-tests.

Please move this addition to above the fuzz test.

kortschak (commented):
LGTM after minor nits resolved. Thanks

kortschak (approved):
Thanks

PR #1874, opened 2023-07-02
Fixed a grammatical error in Unit docs

llamington:
Fixed a grammatical from "a run time" to "at run time"

Please take a look.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -11,6 +11,7 @@
 Alexander Egurnov <alexander.egurnov@gmail.com>
 Andrei Blinnikov <goofinator@mail.ru>
 antichris <chris@u-d13.com>
+Bailey Lissington <lissington4@gmail.com>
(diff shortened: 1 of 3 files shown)


vladimir-ch:
Thanks. Can you send another PR with the commit message "A+C: add Bailey Lissington", adding yourself to AUTHORS and CONTRIBUTORS? When that's done, I'll merge this.

kortschak:
@llamington The second commit should be in a separate PR. We will need to fix up the commit message for the first commit and we cannot do this with the GitHub UI when there is more than one commit. Can you either edit the commit history to make the commit message for the first commit be "unit: fix typo in package doc comment" or revert the second commit so that we can do the fixup in a squash.

llamington:
All sorted

kortschak (approved):
Thanks

PR #1875, opened 2023-07-03
lapack/gonum: improve exceptional shift in Dlahqr

vladimir-ch:
Replays changes done in the Reference in https://github.com/Reference-LAPACK/lapack/pull/478 and https://github.com/Reference-LAPACK/lapack/pull/487

Please take a look.

### DIFF
--- a/lapack/gonum/dlahqr.go
+++ b/lapack/gonum/dlahqr.go
@@ -150,6 +150,9 @@ func (impl Implementation) Dlahqr(wantt, wantz bool, n, ilo, ihi int, h []float6
 
 	itmax := 30 * max(10, nh) // Total number of QR iterations allowed.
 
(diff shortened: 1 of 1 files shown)


kortschak (commented):
This appears to be a partial replay with an exceptional shift at tens rather than the 8 and 6s in the reference. Is that intentional?

vladimir-ch:
Could you be more specific? I'm looking at the reference which has `KEXSH = 10`. There were two PRs in the reference and initially they had `KEXSH = 6`.

On the other hand, I noticed an unrelated mistake in `dat1`.

kortschak:
This was an oversight on my part; I missed the second PR in the reference.

kortschak (approved):

PR #1883, opened 2023-07-28
blas/testblas: clean up Drotmg test and relax tolerance

vladimir-ch:
For https://github.com/gonum/netlib/pull/92

Please take a look.

### DIFF
--- a/blas/testblas/level1double.go
+++ b/blas/testblas/level1double.go
@@ -11,6 +11,7 @@ import (
 
 	"gonum.org/v1/gonum/blas"
 	"gonum.org/v1/gonum/floats"
+	"gonum.org/v1/gonum/floats/scalar"
 )
 
 type DoubleOneVectorCase struct {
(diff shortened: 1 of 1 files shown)


blas/testblas/level1double.go
```diff
@@ -1731,461 +1732,466 @@ func DrotgTest(t *testing.T, d Drotger, skipExtreme bool) {
...
+			d1Want: 0.9999559282289687,
+			d2Want: 0.9898121986058326,
+			x1Want: 0.03662270484346241,
+			pWant: blas.DrotmParams{
+				Flag: blas.Rescaling,
+				H:    [4]float64{1.6110934624105326e-06, -0.000244140625, 0.000244140625, 1.6276041666666668e-06},
+			},
+		},
+
+		// Test cases with values taken from BLAS/TESTING/dblat1.f in the reference.
```
vladimir-ch, line 2060: It's not visible in the diff: I modified these test cases to use "exact" values from the reference.

kortschak: Why were we previously not using the exact values?

vladimir-ch: The test cases were added in https://github.com/gonum/gonum/pull/420. It looks like you used the runtime output of a printf statement, so they're practically the same values only in a different representation.

kortschak: Yes. That seems likely. I have no memory of that.

I've gone through the netlib/BLAS tests and I'm convinced that that is what I did.

kortschak (approved):

PR #1892, opened 2023-09-05
Optimize NKeeper.Keep with heap.Fix

JordanStoker:
This change modifies Gonum spatial/kdtree NKeeper by making its Keep function more computationally efficient by replacing calls to Pop and Push by changing the value at index 0 and calling Fix.

From heap.Fix description, 
```
// Fix re-establishes the heap ordering after the element at index i has changed its value.
// Changing the value of the element at index i and then calling Fix is equivalent to,
// but less expensive than, calling Remove(h, i) followed by a Push of the new value.
// The complexity is O(log n) where n = h.Len().
```
Note that Remove(h,0) is equivalent to Pop()

Please take a look.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -65,7 +65,6 @@ Jonathan J Lawlor <jonathan.lawlor@gmail.com>
 Jonathan Reiter <jonreiter@gmail.com>
 Jonathan Schroeder <jd.schroeder@gmail.com>
 Joost van Amersfoort <git@joo.st>
-Jordan Stoker <jordan_stoker@hotmail.com>
 Joseph Watson <jtwatson@linux-consulting.us>
 Josh Wilson <josh.craig.wilson@gmail.com>
 Julien Roland <juroland@gmail.com>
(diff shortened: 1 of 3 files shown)


kortschak (approved):
Thanks. Can you send another PR with a single commit adding yourself to AUTHORS and CONTRIBUTORS with the commit message "A+C: add Jordan Stoker" (assuming this is your name). When you've done that I'll merge them both.

kortschak (approved):

kortschak:
Adding the new commit to this PR complicates things. We have a conventional format for commit titles which the first commit does not conform it. I could fix that if this were a single commit PR, but cannot now. Can you send the second commit as a separate PR please.

JordanStoker:
My apologies @kortschak. I've created a new pull request with a single commit adding myself to AUTHORS and CONTRIBUTORS here https://github.com/gonum/gonum/pull/1895

kortschak:
Thanks. Can you revert the last commit here?

kortschak (approved):
Thanks

PR #1896, opened 2023-09-07
lapack/gonum: clean up Dlanhs and its test

vladimir-ch:
Please take a look.

@soypat


soypat:
Thank you @vladimir-ch

PR #1898, opened 2023-09-11
lapack/gonum: clean up Dgghrd and its test

vladimir-ch:
Please take a look.

@soypat

### DIFF
--- a/lapack/gonum/dgghrd.go
+++ b/lapack/gonum/dgghrd.go
@@ -10,57 +10,45 @@ import (
 	"gonum.org/v1/gonum/lapack"
 )
 
-// Dgghrd reduces a pair of real matrices (A,B) to generalized upper
-// Hessenberg form using orthogonal transformations, where A is a
-// general matrix and B is upper triangular.  The form of the
-// generalized eigenvalue problem is
+// Dgghrd reduces a pair of real matrices (A,B) to generalized upper Hessenberg
+// form using orthogonal transformations, where A is a general matrix and B is
+// upper triangular.
 //
-//	A*x = lambda*B*x,
+// This subroutine simultaneously reduces A to a Hessenberg matrix H
 //
-// and B is typically made upper triangular by computing its QR
-// factorization and moving the orthogonal matrix Q to the left side
-// of the equation.
-// This subroutine simultaneously reduces A to a Hessenberg matrix H:
+//	Qᵀ*A*Z = H,
 //
-//	Qᵀ*A*Z = H
+// and transforms B to another upper triangular matrix T
 //
-// and transforms B to another upper triangular matrix T:
-//
-//	Qᵀ*B*Z = T
-//
-// in order to reduce the problem to its standard form
-//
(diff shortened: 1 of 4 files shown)


lapack/gonum/dgghrd.go
```diff
@@ -10,57 +10,45 @@ import (
 	"gonum.org/v1/gonum/lapack"
 )
 
-// Dgghrd reduces a pair of real matrices (A,B) to generalized upper
-// Hessenberg form using orthogonal transformations, where A is a
-// general matrix and B is upper triangular.  The form of the
-// generalized eigenvalue problem is
```
vladimir-ch, line 16: I've removed references to the generalized eigenvalue problem from the doc comment because it's not essential for what Dgghrd does.

lapack/lapack.go
```diff
@@ -231,7 +231,7 @@ const (
 type OrthoComp byte
 
 const (
-	OrthoNone  OrthoComp = 'N' // Do not compute orthogonal matrix.
-	OrthoUnit  OrthoComp = 'I' // Argument is initialized to the unit matrix and the orthogonal matrix is returned.
-	OrthoEntry OrthoComp = 'V' // Argument Q contains orthogonal matrix Q1 on entry and the product Q1*Q is returned.
+	OrthoNone     OrthoComp = 'N' // Do not compute the orthogonal matrix.
+	OrthoExplicit OrthoComp = 'I' // The orthogonal matrix is formed explicitly and returned in the argument.
+	OrthoPostmul  OrthoComp = 'V' // The orthogonal matrix is post-multiplied into the matrix stored in the argument on entry.
```
kortschak, line 236: These names are more evocative. Is there an origin for them?

lapack/gonum/dgghrd.go
```diff
@@ -10,57 +10,45 @@ import (
...
-	case len(a) < (n-1)*lda+n:
-		panic(shortA)
-	case len(b) < (n-1)*ldb+n:
-		panic(shortB)
 	case n < 0:
 		panic(nLT0)
-	case ilo < 0:
+	case ilo < 0 || max(0, n-1) < ilo:
 		panic(badIlo)
-	case ihi < ilo-1 || ihi >= n:
```
soypat, line 63: Is `ihi >= n` not more readable since we are testing for ihi?

soypat (approved):
LGTM

lapack/gonum/dgghrd.go
kortschak: The convention we use for ranges only uses `<`. We find that the expressions are clearer this way.

lapack/lapack.go
vladimir-ch: Coming up with good names is difficult and especially for these constants. In this case they turned out quite nicely, the origin is the documentation and an effort to reduce it to a single keyword plus the nice "Ortho" prefix that @soypat chose.

kortschak (approved):

PR #1902, opened 2023-09-22
graph: fix comment of exported elements

Juneezee:
According to [Comment Sentences at github.com/golang](https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences), it is a convention to begin a comment with the name of the exported element.

### DIFF
--- a/graph/formats/gexf12/gexf.go
+++ b/graph/formats/gexf12/gexf.go
@@ -187,7 +187,7 @@ type AttValues struct {
 	AttValues []AttValue `xml:"attvalue,omitempty"`
 }
 
-// AttValues holds a single attribute value and its associated dynamics.
+// AttValue holds a single attribute value and its associated dynamics.
 type AttValue struct {
 	For       string `xml:"for,attr"`
 	Value     string `xml:"value,attr"`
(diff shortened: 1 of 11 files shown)


graph/formats/gexf12/gexf.go
```diff
@@ -197,7 +197,7 @@ type AttValue struct {
 	EndOpen   string `xml:"endopen,attr,omitempty"`
 }
 
-// EdgeShape holds the visual representation of an edge with associated
+// Edgeshape holds the visual representation of an edge with associated
```
kortschak, line 200: This one is unfortunate. I think the comment is correct, but I don't think the type name is wrong enough to justify a potential breakage.

kortschak (commented):
Thank you for doing this. Would you please add another commit with the commit message "A+C: add Eng Zer Jun" adding yourself to AUTHORS and CONTRIBUTORS. After that, LGTM.

kortschak (approved):
Thanks.

For future, we don't use signed-off-by and I don't think the first commit needs a reference; the fixes fix clearly incorrect comments.

PR #1905, opened 2023-10-03
mat: add LU.RowPivots and deprecate LU.Pivot

vladimir-ch:
Please take a look.

### DIFF
--- a/mat/lu.go
+++ b/mat/lu.go
@@ -144,16 +144,16 @@ func (lu *LU) LogDet() (det float64, sign float64) {
 	return floats.Sum(logDiag), sign
 }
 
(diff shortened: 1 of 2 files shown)


mat/lu.go
```diff
@@ -177,6 +177,11 @@ func (lu *LU) Pivot(dst []int) []int {
 	return dst
 }
 
+// Pivot is deprecated, use RowPivots instead.
```
kortschak, line 180: There is a magic incantation to get the on-line docs to hid deprecated labels.

```suggestion
// Deprecated: Use RowPivots instead.
```

vladimir-ch: Thanks, done.

kortschak (approved):

Issue #926, opened 2019-03-27
mat: LU, LQ, QR should implement matrix


btracey:
See discussion in issue #830 

- [x] LU
- [x] QR
- [x] LQ

vladimir-ch:
Done.

PR #1907, opened 2023-10-03
mat: make LU satisfy Matrix

vladimir-ch:
Updates #926

Please take a look.

### DIFF
--- a/mat/lu.go
+++ b/mat/lu.go
@@ -19,11 +19,62 @@ const (
 	badLU          = "mat: invalid LU factorization"
 )
 
-// LU is a type for creating and using the LU factorization of a matrix.
+// LU is a square n×n matrix represented by its LU factorization with partial
+// pivoting.
+//
+// The factorization has the form
+//
+//	A = P * L * U
+//
+// where P is a permutation matrix, L is lower triangular with unit diagonal
+// elements, and U is upper triangular.
+//
+// Note that this matrix representation is useful for certain operations, in
+// particular for solving linear systems of equations. It is very inefficient at
+// other operations, in particular At is slow.
 type LU struct {
 	lu    *Dense
 	swaps []int
+	piv   []int
 	cond  float64
+	ok    bool // Whether A is nonsingular
+}
+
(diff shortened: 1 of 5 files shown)


mat/lu.go
```diff
@@ -52,37 +106,51 @@ func (lu *LU) updateCond(anorm float64, norm lapack.MatrixNorm) {
...
+// result in the receiver. The LU decomposition will complete regardless of the
+// singularity of a.
 //
-// The LU factorization is computed with pivoting, and so really the decomposition
-// is a PLU decomposition where P is a permutation matrix. The individual matrix
-// factors can be extracted from the factorization using the Permutation method
-// on Dense, and the LU.LTo and LU.UTo methods.
+// The L and U matrix factors can be extracted from the factorization using the
+// LTo and UTo methods. The matrix P can be extracted as a row permutation using
+// the Pivot method and applied using Dense.PermuteRows.
```
kortschak, line 115:
```suggestion
// the RowPivots method and applied using Dense.PermuteRows.
```
?

vladimir-ch: Thanks.

vladimir-ch:
This is now ready.

mat/lu.go
```diff
@@ -19,11 +19,62 @@ const (
...
+//
+// Note that this matrix representation is useful for certain operations, in
+// particular for solving linear systems of equations. It is very inefficient at
+// other operations, in particular At is slow.
 type LU struct {
 	lu    *Dense
 	swaps []int
+	piv   []int
 	cond  float64
+	ok    bool // Whether A is singular
```
kortschak, line 40: This is a little confusing; the intention is that `!ok` → singular, but it could be read differently.

kortschak (commented):
Minor nit only.

mat/lu.go
vladimir-ch: Good point, thanks. I've also adjusted comments in the involved Lapack routines.

kortschak (approved):

PR #1909, opened 2023-10-05
mat: make QR satisfy Matrix

vladimir-ch:
Making QR satisfy Matrix is more difficult compared to the other factorization types because the matrix Q is represented only implicitly and recovering it is expensive. I opted for computing it explicitly together with the factorization at the expense of memory.

I also switched from Dormqr (prepare the identity matrix and multiply Q into it) to Dorgqr which exists for exactly this purpose.

Please take a look.

Updates #926

### DIFF
--- a/lapack/lapack64/lapack64.go
+++ b/lapack/lapack64/lapack64.go
@@ -694,6 +694,28 @@ func Ormlq(side blas.Side, trans blas.Transpose, a blas64.General, tau []float64
 	lapack64.Dormlq(side, trans, c.Rows, c.Cols, a.Rows, a.Data, max(1, a.Stride), tau, c.Data, max(1, c.Stride), work, lwork)
 }
 
+// Orgqr generates an m×n matrix Q with orthonormal columns defined by the
+// product of elementary reflectors
+//
+//	Q = H_0 * H_1 * ... * H_{k-1}
+//
+// as computed by Geqrf.
+//
+// k is determined by the length of tau.
+//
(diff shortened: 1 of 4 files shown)


lapack/lapack64/lapack64.go
```diff
@@ -725,7 +748,7 @@ func Ormlq(side blas.Side, trans blas.Transpose, a blas64.General, tau []float64
 // If lwork is -1, instead of performing Ormqr, the optimal workspace size will
 // be stored into work[0].
 func Ormqr(side blas.Side, trans blas.Transpose, a blas64.General, tau []float64, c blas64.General, work []float64, lwork int) {
-	lapack64.Dormqr(side, trans, c.Rows, c.Cols, a.Cols, a.Data, max(1, a.Stride), tau, c.Data, max(1, c.Stride), work, lwork)
+	lapack64.Dormqr(side, trans, c.Rows, c.Cols, len(tau), a.Data, max(1, a.Stride), tau, c.Data, max(1, c.Stride), work, lwork)
```
vladimir-ch, line 751: This is actually a bug fix. Since the signature of Ormqr does not specify k, it must be determined from somewhere and the length of tau is the natural (and only) option. That's also why the documentation change.

I thought about including a change in this PR to require `len(tau) == k` in the Lapack call graph of all these QR-related functions (we do it in some functions already) but the change was quite pervasive, so I'll leave it for later.

kortschak (approved):

PR #1910, opened 2023-10-06
mat: add VecDense.Permute

vladimir-ch:
I'm wondering whether it wouldn't be more `mat`-like to have signatures
```
// v = P*x or v = P^T*x
func (v *VecDense) Permute(p []int, trans bool, x Vector)

// M = P*A or M = P^T*A
func (m *Dense) PermuteRows(p []int, trans bool, a Matrix)

// M = A*P or M = A*P^T
func (m *Dense) PermuteCols(a Matrix, p []int, trans bool)
```
It complicates the implementation (special-casing on the type of A and whether it's transposed; do we care about orientation of x?) and I'm not sure if the greater flexibility is a benefit for the user. What do you think?

### DIFF
--- a/mat/vector.go
+++ b/mat/vector.go
@@ -837,3 +837,19 @@ func (v *VecDense) RowViewOf(m RawMatrixer, i int) {
 	v.mat.Data = rm.Data[i*rm.Stride : i*rm.Stride+rm.Cols]
 	v.mat.N = rm.Cols
 }
+
+// Permute rearranges the elements of the n-vector v in the receiver as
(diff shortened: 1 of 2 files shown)


kortschak:
If we use the first signature in the alternative, we have to care about vector orientation since `Vector` can be a row vector. I'm inclined not to want to have the alternative signatures; the user would already know what kind of `Matrix` they have and can easily move that into the destination in a hopefully efficient way, while we would have to have non-trivial assessments of the value to determine the best way to get it into place before doing the same work.

kortschak (approved):

vladimir-ch:
These were exactly my thoughts too.

PR #1912, opened 2023-10-07
stat/distmv: add NormalRandCov

### LINKED ISSUE #1717: distmv.NewNormal does not accept positive semidefinite matrix as Σ

I need to draw samples from a multivariate gaussian with a specific μ and Σ, Σ is positive semi-definite, which should be suitable for sampling, but is not allowed in gonum.

### What did you do?

The original draft I got is written in python, up to the normal distribution/sampling, everything works in both numpy and gonum.

Then, using gonum's distmv.NewNormal does not work, while numpy.random.multivariate_normal() can be used to draw the samples.

I can confirm Σ is the same in both.

Looking under the hood, numpy uses SVD to determine if Σ is PSD ([src](https://github.com/numpy/numpy/blob/94010e3e820998afe1cab4d5d915f34552a4476e/numpy/random/mtrand.pyx#L4120)):
```py
(u, s, v) = numpy.linalg.svd(Σ)
psd = np.allclose(np.dot(v.T * s, v), Σ)    
```
Gonum on the other hand uses Cholesky to determine if Σ is PD ([src](https://github.com/gonum/gonum/blob/6b90107d4a29dcf67ae7c214a14935d0bcfbf129/stat/distmv/normal.go#L57)), but the same SVD check works:
```go
svd := new(mat.SVD)
svd.Factorize(Σ, mat.SVDFull)

V := new(mat.Dense)
svd.VTo(V)
s := make([]float64, Σm)
svd.Values(s)

ss := make([]float64, Σm*Σm)
for i := 0; i < n; i++ {
	for j, v := range s {
		ss[i*Σm+j] = v
	}
}
S := mat.NewDense(Σm, Σm, ss)

M := new(mat.Dense)
M.MulElem(V, S)
M.Mul(M, V.T())
psd := mat.EqualApprox(Σ, M, 1e-5)
```

Sorry I cannot share the data or more of the code.

distmv.NewNormal should accept positive semidefinite Σ, as far as I understand.

### What actually happened?

distmv.NewNormal accepts only positive definite Σ.

```
$ go version
go version go1.16.6 linux/amd64
```
```
$ grep "v1/gonum" go.sum 
gonum.org/v1/gonum v0.0.0-20180816165407-929014505bf4/go.mod h1:Y+Yx5eoAFn32cQvJDxZx5Dpnq+c3wtXuadVZAcxbbBo=
gonum.org/v1/gonum v0.8.1-0.20200930085651-eea0b5cb5cc9/go.mod h1:oe/vMfY3deqTw+1EZJhuvEW2iwGF1bW9wwu7XCu0+v0=
gonum.org/v1/gonum v0.8.2/go.mod h1:oe/vMfY3deqTw+1EZJhuvEW2iwGF1bW9wwu7XCu0+v0=
gonum.org/v1/gonum v0.9.3 h1:DnoIG+QAMaF5NvxnGe/oKsgKcAc6PcUyl8q0VetfQ8s=
gonum.org/v1/gonum v0.9.3/go.mod h1:TZumC3NeyVQskjXqmyWt4S3bINhy7B4eYwW69EbyX+0=
```

It must since the code has not changed.

vladimir-ch:
Closes #1717 

Please take a look.

### DIFF
--- a/stat/distmv/general.go
+++ b/stat/distmv/general.go
@@ -6,7 +6,8 @@ package distmv
 
 const (
 	badQuantile      = "distmv: quantile not between 0 and 1"
-	badReceiver      = "distmv: input slice is not nil or the correct length"
+	badOutputLen     = "distmv: output slice is not nil or the correct length"
+	badInputLength   = "distmv: input slice length mismatch"
 	badSizeMismatch  = "distmv: size mismatch"
 	badZeroDimension = "distmv: zero dimensional input"
 	nonPosDimension  = "distmv: non-positive dimension input"
@@ -14,8 +15,8 @@ const (
 
 const logTwoPi = 1.8378770664093454835606594728112352797227949472755668
 
-// useAs gets a slice of size n. If len(x) == n, x is returned, if len(x) == 0
-// then a slice is returned of length n.
+// reuseAs returns a slice of length n. If len(x) == n, x is returned, if len(x)
+// == 0 then a slice of length n is returned, otherwise reuseAs panics.
 func reuseAs(x []float64, n int) []float64 {
 	if len(x) == n {
 		return x
@@ -26,5 +27,5 @@ func reuseAs(x []float64, n int) []float64 {
 		}
 		return make([]float64, n)
 	}
-	panic(badReceiver)
+	panic(badOutputLen)
 }
--- a/stat/distmv/normal_test.go
+++ b/stat/distmv/normal_test.go
@@ -5,6 +5,7 @@
 package distmv
 
 import (
+	"fmt"
 	"math"
 	"testing"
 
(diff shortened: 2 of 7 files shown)


stat/distmv/general.go
kortschak, line 1: `badReceiver` appears to be skewed wrt the behaviour.

stat/distmv/normal_test.go
```diff
@@ -572,3 +573,94 @@ func TestNormalScoreInput(t *testing.T) {
...
+			case "Cholesky":
+				var chol mat.Cholesky
+				ok := chol.Factorize(a)
+				if !ok {
+					t.Fatalf("%s: factorization failed, input covariance not positive definite?", name)
+				}
+				cov = &chol
+			case "PivotedCholesky":
+				var chol mat.PivotedCholesky
+				_ = chol.Factorize(a, -1)
```
kortschak, line 639: Do we need the `_ = ` prefix?

stat/distmv/general.go
vladimir-ch: I'm not sure what you exactly mean but I've changed it to `badOutputLen` and modified the string accordingly. There's a lot to clean up but I don't want to go in there.

stat/distmv/normal_test.go
vladimir-ch: I wanted to leave a hint that there's an ok return but it's not useful here. I've removed the prefix but can put it back if you want.

stat/distmv/general.go
kortschak: Sorry, I should have been clearer. The text is "output slice is *not nil* or the correct length", but there is no nilness check here. Nilness is something we use elsewhere for this (all of mat — should we be consistent?), but it's not the behaviour nor is it the documented behaviour.

vladimir-ch: Ah, yes, that's one of the cleanup candidates that I ignored on purpose. We can decide here and I'll submit another PR.

I'd prefer to be consistent with mat. We will trigger panics in case anyone relies on the undocumented behavior of non-nil zero-length slices to be acceptable for reuse but I think we can disregard that.

kortschak: I'd prefer to be consistent with mat.

vladimir-ch: I've cleaned up comments and code in distmv where the destination slice is used. Note that at a couple of places I also simplified the code to avoid allocation of temporary slices.

kortschak (approved):
We will need to remember to mark this as a breaking changing in the next release notes.

PR #1915, opened 2023-10-08
mat: make EigenSym satisfy Matrix

vladimir-ch:
Please take a look.

### DIFF
--- a/mat/eigen_test.go
+++ b/mat/eigen_test.go
@@ -131,10 +131,11 @@ func cmplxEqualTol(v1, v2 []complex128, tol float64) bool {
 	return true
 }
 
-func TestSymEigen(t *testing.T) {
+func TestEigenSym(t *testing.T) {
 	t.Parallel()
+	const tol = 1e-14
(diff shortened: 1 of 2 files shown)


mat/eigen_test.go
```diff
@@ -153,25 +154,26 @@ func TestSymEigen(t *testing.T) {
 		var es EigenSym
 		ok := es.Factorize(test.mat, true)
 		if !ok {
-			t.Errorf("bad factorization")
+			t.Errorf("bad test")
```
kortschak, line 157: These might want some identifying mark to explain which case was bad — pointless here with only one case, but below might be good. Not crucial.

kortschak (dismissed):

mat/eigen_test.go
vladimir-ch: Done.

vladimir-ch:
@kortschak Can you please approve again? I pushed changes by mistake.

kortschak (approved):
GitHub's PR review UX is horrible.

Issue #920, opened 2019-03-25
mat: Dense and other Dense types should implement SolveTo/SolveVecTo


btracey:
See discussion in #830

PR #1916, opened 2023-10-16
mat: delegate to SolveTo method in *Dense.Solve

vladimir-ch:
Please take a look.

Updates #920

### DIFF
--- a/mat/solve.go
+++ b/mat/solve.go
@@ -4,12 +4,6 @@
 
 package mat
 
-import (
-	"gonum.org/v1/gonum/blas"
-	"gonum.org/v1/gonum/blas/blas64"
-	"gonum.org/v1/gonum/lapack/lapack64"
-)
-
 // Solve solves the linear least squares problem
 //
 //	minimize over x |b - A*x|_2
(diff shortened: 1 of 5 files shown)


mat/solve.go
```diff
@@ -33,61 +27,18 @@ import (
 // If A does not have full rank, a Condition error is returned. See the
 // documentation for Condition for more information.
 func (m *Dense) Solve(a, b Matrix) error {
```
kortschak, line 29: Do we want to have a note here that if the underlying `Matrix` in `a` is a `SolveToer` that its `SolveTo` method is used? If not, why do we define `SolveToer`?

Related: We don't AFAICS have a defined notion of a backing matrix — this would be any matrix that is what you get when you have called `untransposeExtract` with it, though it can't be documented with reference to that since it is not exported. I think this would be useful to add (elsewhen) to the section "Type Switching" in the docs.

vladimir-ch: I've added a note about SolveToer, PTAL.

I'll postpone the update of "Type Switching" to later because I don't immediately see an simple way to describe it and don't want to block this PR.

kortschak (approved):

PR #1918, opened 2023-10-21
stat/distmv: add special case in NormalRandCov for mat.EigenSym

vladimir-ch:
Please take a look.

### DIFF
--- a/mat/eigen.go
+++ b/mat/eigen.go
@@ -140,6 +140,20 @@ func (e *EigenSym) Values(dst []float64) []float64 {
 	return dst
 }
 
+// RawValues returns the slice storing the eigenvalues of A in ascending order.
+//
+// If the returned slice is modified, the factorization is invalid and should
+// not be used.
+//
+// If the receiver does not contain a successful factorization, RawValues will
+// return nil.
+func (e *EigenSym) RawValues() []float64 {
+	if !e.succFact() {
+		return nil
+	}
+	return e.values
+}
+
 // VectorsTo stores the orthonormal eigenvectors of the factorized n×n matrix A
 // into the columns of dst.
 //
@@ -164,6 +178,25 @@ func (e *EigenSym) VectorsTo(dst *Dense) {
 		}
 	}
 	dst.Copy(e.vectors)
+}
+
+// RawQ returns the orthogonal matrix Q from the spectral factorization of the
+// original matrix A
+//
+//	A = Q * Λ * Qᵀ
+//
+// The columns of Q contain the eigenvectors of A.
+//
+// If the returned matrix is modified, the factorization is invalid and should
+// not be used.
+//
+// If the receiver does not contain a successful factorization or eigenvectors
+// not computed, RawU will return nil.
+func (e *EigenSym) RawQ() Matrix {
+	if !e.succFact() || !e.vectorsComputed {
+		return nil
+	}
+	return e.vectors
 }
 
 // EigenKind specifies the computation of eigenvectors during factorization.
--- a/mat/eigen.go
+++ b/mat/eigen.go
@@ -14,8 +14,12 @@ const (
 	noVectors = "mat: eigenvectors not computed"
 )
 
-// EigenSym is a type for creating and manipulating the Eigen decomposition of
-// symmetric matrices.
+// EigenSym is a type for computing all eigenvalues and, optionally,
+// eigenvectors of a symmetric matrix A.
+//
+// It is a Symmetric matrix represented by its spectral factorization. Once
+// computed, this representation is useful for extracting eigenvalues and
+// eigenvector, but At is slow.
 type EigenSym struct {
 	vectorsComputed bool
 
@@ -34,7 +38,8 @@ func (e *EigenSym) SymmetricDim() int {
 	return len(e.values)
 }
 
-// At returns the element at row i, column j.
+// At returns the element at row i, column j of the matrix A.
+//
 // At will panic if the eigenvectors have not been computed.
 func (e *EigenSym) At(i, j int) float64 {
 	if !e.vectorsComputed {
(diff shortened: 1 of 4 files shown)


stat/distmv/normal.go
```diff
@@ -365,22 +366,28 @@ func NormalRandCov(dst, mean []float64, cov mat.Symmetric, src rand.Source) []fl
...
 		dstVec := mat.NewVecDense(n, dst)
 		dstVec.MulVec(cov.RawU().T(), dstVec)
 		dstVec.Permute(cov.ColumnPivots(nil), true)
-		floats.Add(dst, mean)
-	// TODO(vladimir-ch): enable this case when EigenSym satisfies Matrix.
-	// case *mat.EigenSym:
+	case *mat.EigenSym:
+		d := cov.RawValues()
+		if d[0] < 0 {
+			panic("distmv: covariance not positive semi-definite")
```
vladimir-ch, line 376: This is not nice. And the caller doesn't have much options how to avoid it. They should not modify the slice from RawValues (as documented) even if the smallest eigenvalue is a negative almostzero. So they can only modify the covariance matrix and reattempt the factorization. We could hardcode some ad hoc mechanism with a tolerance to consider tiny negative numbers to be zero but that's probably even uglier.

mat/eigen.go
```diff
@@ -156,6 +180,25 @@ func (e *EigenSym) VectorsTo(dst *Dense) {
...
+//	A = Q * Λ * Qᵀ
+//
+// The columns of Q contain the eigenvectors of A.
+//
+// If the returned matrix is modified, the factorization is invalid and should
+// not be used.
+//
+// If the receiver does not contain a successful factorization or eigenvectors
+// not computed, RawU will return nil.
+func (e *EigenSym) RawQ() Matrix {
```
vladimir-ch, line 195: Or `RawVectors`?

kortschak: Not sure. `Cholesky` uses `RawU` for the equivalent accessor. Though that is more obvious to people who have not read the documentation and are not aware of the conventional naming of the eigenvector-containing matrix. I can see `RawVectors() Matrix` being similarly confusing to some users though. On balance, I think `RawQ` is probably the right approach.

stat/distmv/normal.go
kortschak: Can we allow a matrix type that implements the method set of `EigenSym` and use that instead of `EigenSym`? This way the user can implement an almost zero filter. If this is OK, it would be an argument for keeping `RawQ` since no-one else is using that (though I don't think that is quite protective enough).

mat/eigen.go
vladimir-ch: > ... conventional naming of the eigenvector-containing matrix.

I tried to check some literature and there doesn't seem to be a consistent notation for the factorization, unlike QR or LQ. I used Q to denote an orthogonal matrix for consistency with these two factorizations but Lapack itself uses Z: https://www.netlib.org/lapack/lug/node30.html For the caller the vectors are the focus, so it makes sense to use `RawVectors`, also to have a pair name for VectorsTo, but as you write it can be confusing to some users and the name is a bit too similar to `RawVector`.

stat/distmv/normal.go
vladimir-ch: Something like this? d617f1934dbf66c764bceb2f0c50bbc1e54739f5

kortschak: Yeah, though probably `func NewFilteredEigenSym(ed *mat.EigenSym, epsilon float64) *FilteredEigenSym {`.

vladimir-ch: Initially I had the epsilon but then decided against it. numpy use the epsilon only for checking the eigenvalues, for the computation they use `abs(val)`, so if an eigenvalue is -tol, they use tol. I find it strange, because if the concern is rounding errors and these eigenvalues should really be zero, then why use abs(val) instead of max(0, val)?

Is the name `FilteredEigenSym` good? Would `PositivePartEigenSym` be better?

kortschak: Yes, that is what I came to a few minutes after make in the suggestion.

I prefer `PositivePartEigenSym`.

vladimir-ch: > I prefer PositivePartEigenSym.

Done.

stat/distmv/normal.go
```diff
@@ -332,18 +332,81 @@ func NormalRand(dst, mean []float64, chol *mat.Cholesky, src rand.Source) []floa
...
+// that are almost zero but negative due to rounding errors.
+type PositivePartEigenSym struct {
+	ed   *mat.EigenSym
+	vals []float64
+}
+
+var _ EigenSym = (*PositivePartEigenSym)(nil)
+var _ EigenSym = (*mat.EigenSym)(nil)
+
+// NewFilteredEigenSym returns a new FilteredEigenSym, wrapping the given
```
kortschak, line 360:
```suggestion
// NewPositivePartEigenSym returns a new PositivePartEigenSym, wrapping the given
```
Though possibly with some wrapping.

stat/distmv/normal.go
```diff
@@ -332,18 +332,81 @@ func NormalRand(dst, mean []float64, chol *mat.Cholesky, src rand.Source) []floa
...
+// eigendecomposition.
+func NewPositivePartEigenSym(ed *mat.EigenSym) *PositivePartEigenSym {
+	n := ed.SymmetricDim()
+	vals := make([]float64, n)
+	for i, lamda := range ed.RawValues() {
+		if lamda >= 0 {
```
kortschak, line 366:
```suggestion
		if lamda > 0 {
```
? since `vals[i]` is already zero.

stat/distmv/normal.go
vladimir-ch: Done.

stat/distmv/normal.go
vladimir-ch: Done.

kortschak (approved):

PR #1919, opened 2023-11-02
lapack/gonum: fix accumulation in Dlassq

vladimir-ch:
See https://github.com/Reference-LAPACK/lapack/pull/909 

Please take a look.

### DIFF
--- a/lapack/gonum/dlassq.go
+++ b/lapack/gonum/dlassq.go
@@ -77,21 +77,25 @@ func (impl Implementation) Dlassq(n int, x []float64, incx int, scale float64, s
 		ax := scale * math.Sqrt(sumsq)
 		switch {
 		case ax > dtbig:
-			// We assume scale >= sqrt( TINY*EPS ) / dsbig, that is, if the
-			// scaled sum is big then its scaling factor should not be too
-			// small.
-			v := scale * dsbig
-			abig += (v * v) * sumsq
+			if scale > 1 {
+				scale *= dsbig
+				abig += scale * (scale * sumsq)
+			} else {
+				// sumsq > dtbig^2 => (dsbig * (dsbig * sumsq)) is representable.
+				abig += scale * (scale * (dsbig * (dsbig * sumsq)))
+			}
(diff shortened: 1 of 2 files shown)


lapack/testlapack/dlassq.go
```diff
@@ -21,7 +21,7 @@ type Dlassqer interface {
 func DlassqTest(t *testing.T, impl Dlassqer) {
 	values := []float64{
 		0,
-		2 * safmin,
+		0.5 * safmin,
```
vladimir-ch, line 24: I can't really justify this but it increases coverage to almost 100%.

lapack/testlapack/dlassq.go
```diff
@@ -72,15 +72,7 @@ func dlassqTest(t *testing.T, impl Dlassqer, rnd *rand.Rand, n, incx, cas int, v
 		workssq += wi * wi
 	}
 
-	// Construct a test (n+1)-vector z as
```
vladimir-ch, line 75: The diff is ugly but I only simplified the test by getting rid of slice `z` and moved the computation of `nrmWant` closer to the check.

lapack/gonum/dlassq.go
```diff
@@ -77,21 +77,25 @@ func (impl Implementation) Dlassq(n int, x []float64, incx int, scale float64, s
...
 		switch {
 		case ax > dtbig:
-			// We assume scale >= sqrt( TINY*EPS ) / dsbig, that is, if the
-			// scaled sum is big then its scaling factor should not be too
-			// small.
-			v := scale * dsbig
-			abig += (v * v) * sumsq
+			if scale > 1 {
+				scale *= dsbig
+				abig += scale * (scale * sumsq)
```
kortschak, line 82: I take it the order of operations is to prevent an overflow?

vladimir-ch: I think it's rather to prevent underflow in `scale*scale`, after scaling down (by `dsbig`) `scale` can be very small.

PR #1931, opened 2023-12-06
stat: Variance should work if the input length equals 1

haoxins:
Please take a look.

same to the NumPy, if the input length is `1`, the current function will return `NaN`

```py
>>> np.var([1])
0.0
```

### DIFF
--- a/stat/stat.go
+++ b/stat/stat.go
@@ -1319,6 +1319,11 @@ func MeanVariance(x, weights []float64) (mean, variance float64) {
 		unnormalisedVariance float64
 		sumWeights           float64
 	)
+
+	if len(x) == 1 {
+		return x[0], 0
+	}
+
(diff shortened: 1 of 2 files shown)


kortschak (changes requested):
This is not correct. The variance of a single datum is undefined. See for example from R:
```
~ $ R --quiet
> var(c(1))
[1] NA
```

haoxins:
> This is not correct. The variance of a single datum is undefined. 

Should we consider to define this?
I think it is reasonable to return the determined value (the original one).

Since the R and NumPy have different behaviors for this, should we pick the convenient one?

I got an error when I returned the variance (NaN) in the HTTP JSON response for my use case.

```
json: unsupported value: NaN
```

haoxins:
@kortschak https://github.com/gonum/gonum/pull/1932/files

Sorry, I did not notice the PopVariance, I added some tests to indicate the differences.

PR #1932, opened 2023-12-06
stat: Add tests for MeanVariance and PopMeanVariance

haoxins:
chore

Please take a look.


kortschak (commented):
Why does this need to be done? [We have tests for this](https://github.com/gonum/gonum/blob/f74f45f5f3e9cc7c1d0f0af2ffd19ccf8972a87e/stat/stat_test.go#L1787-L1837).

haoxins:
> Why does this need to be done? [We have tests for this](https://github.com/gonum/gonum/blob/f74f45f5f3e9cc7c1d0f0af2ffd19ccf8972a87e/stat/stat_test.go#L1787-L1837). Hang on, I missed that we are missing `MeanVariance`. Please make your tests confirm to the style used in the project (we use table tests for these).

done~

PR #1937, opened 2024-03-01
graph/iterator: always use safe mode

ianlancetaylor:
We can use reflect.MapIter.Reset, introduced in Go 1.18, to restart iteration without allocating a MapIter.


ianlancetaylor:
This will permit the code work even if golang/go#54766 is adopted.

But I'm not entirely sure why the unsafe code was introduced.  I don't know if this meets all the requirements.

kortschak:
The reason for the existence of the unsafe implementation is performance. When it was written there was a significant difference between perf in the unsafe approach and the reflect-based approach. This is still the case with the majority of packages showing serious performance hits when using the reflect-based implementation.

|Package|Geomean delta|
|---|---|
| gonum.org/v1/gonum/graph/coloring | +5.85% |
| gonum.org/v1/gonum/graph/community| +10.17% |
| gonum.org/v1/gonum/graph/flow | +14.67% |
| gonum.org/v1/gonum/graph/formats/rdf | -3.13% |
| gonum.org/v1/gonum/graph/path | +20.38% |
| gonum.org/v1/gonum/graph/product | -3.33% |
| gonum.org/v1/gonum/graph/topo | +14.11% |
| gonum.org/v1/gonum/graph/traverse | +34.08% |

(it's unclear to me why rdf and product show the converse pattern, and in the case of product it is more complex since the benchmarks show both directions in that package; in the packages showing regressions here the results of individual benchmarks are almost entirely in the wrong direction)

The main issue is that the iterators are used in tight loops for nearly all of the graph algorithms and so disproportionately impact on performance.

ianlancetaylor:
I assume those benchmarks are with this pull request.  Do you have any insight into why the unsafe version is faster?  I would have thought that the version in this pull request would have fewer memory allocations, as the iteration struct is a field in `Lines`, etc., rather than a pointer.  And the unsafe version just passes the struct to the reflect functions anyhow.  What is the key step that makes the unsafe version faster?

ianlancetaylor:
OK, I see the problem.  It's the call to `(*reflect.MapIter).Key().Int()`.  Because of how the reflect package represents values, the `Key` method has to copy the integer into newly allocated memory in order `Key` to return a `reflect.Value`.  Then we immediately call the `Int` method and discard the value.

If I add a `(*reflect.MapIter).KeyInt` method that just returns the integer directly, the newer code is the same or faster as the existing code.

I filed https://go.dev/issue/66125.

kortschak:
Thanks, Ian.

Issue #1691, opened 2021-07-02
spatial: R-tree


firelizzard18:
Feature request: add an [R-tree](https://en.wikipedia.org/wiki/R-tree) implementation to `spatial`.

### What are you trying to do?
Given an arbitrary polyhedron represented as an array of polygons (faces), search the array for faces that are within a given region.

### What did you try?
Linear searches.

### How does Gonum not allow you to achieve your goal?
The closest thing gonum has is `spatial/kdtree`, which stores points not regions, and cannot be queried by region.

### Are you able to help contribute the feature?
I am willing to help, but this is definitely not my area of expertise.

kortschak:
If you are willing to provide an implementation, I can review it.

I can see a couple of R-tree implementations for N-D, https://pkg.go.dev/github.com/dhconnelly/rtreego and a more recently and apparently currently maintained but only 2-D, https://pkg.go.dev/github.com/tidwall/rtree.

firelizzard18:
@kortschak I was thinking about R-trees and how to partition sets of rectangles, since inserting may result in needing to split a node, and the split node should be balanced. Wikipedia doesn't say much about that, other than mentioning that one strategy is Hilbert curves. So I went down that rabbit hole.

I ported the code on Wikipedia for a 2D Hilbert curve and extrapolated that to 3D and 4D. I was unable to extrapolate to n-D. My [r-tree](https://github.com/gonum/gonum/compare/master...firelizzard18:r-tree) branch adds `spatial/curve`, which defines `SpaceFilling`, `Hilbert2D`, `Hilbert3D`, `Hilbert4D`. I plan to use that as a partitioning scheme for an r-tree implementation, though that means it's limited to at most 4 dimensions.

firelizzard18:
@kortschak I made another branch and opened #1695 just for the Hilbert curve implementation.

FYI, I only use RNG for the coordinates I use for benchmarking. The tests iterate exhaustively over the relevant space. Which, by the way, grows exponentially as 2^(order*dimension). The upper limits on order in the test are based on what runs quickly on my machine.

firelizzard18:
I created an R-tree implementation: https://github.com/firelizzard18/gonum/compare/hilbert-curve...r-tree. There's certainly room for improvement but it works, I think. My plan is to used this in a real-world application before submitting a PR. I have one in mind (the motivation for this issue), but I'm not sure when that will happen.

[...]

PR #1939, opened 2024-03-03
spatial/curve: Hilbert curves, take 2

firelizzard18:
See #1695 - this is the same code with minor modifications as per comments on the original PR.

<details><summary>Benchmarks</summary><pre>
Running tool: /usr/local/go/bin/go test -benchmem -run=^$ -bench ^BenchmarkHilbert$ gonum.org/v1/gonum/spatial/curve -v -count=1
goos: darwin
goarch: amd64
pkg: gonum.org/v1/gonum/spatial/curve
cpu: Intel(R) Core(TM) i7-6700HQ CPU @ 2.60GHz
BenchmarkHilbert
BenchmarkHilbert/2D/Pos
BenchmarkHilbert/2D/Pos/Order_1
BenchmarkHilbert/2D/Pos/Order_1-8       100000000                11.84 ns/op            0 B/op           0 allocs/op
BenchmarkHilbert/2D/Pos/Order_2
BenchmarkHilbert/2D/Pos/Order_2-8       76892733         16.06 ns/op            0 B/op           0 allocs/op
BenchmarkHilbert/2D/Pos/Order_3
BenchmarkHilbert/2D/Pos/Order_3-8       55760536         20.37 ns/op            0 B/op           0 allocs/op
BenchmarkHilbert/2D/Pos/Order_4
BenchmarkHilbert/2D/Pos/Order_4-8       53285614                24.30 ns/op              0 B/op          0 allocs/op
BenchmarkHilbert/2D/Pos/Order_5
BenchmarkHilbert/2D/Pos/Order_5-8       38780760                26.66 ns/op              0 B/op          0 allocs/op
BenchmarkHilbert/2D/Pos/Order_6
BenchmarkHilbert/2D/Pos/Order_6-8       30895916                35.55 ns/op              0 B/op          0 allocs/op
BenchmarkHilbert/2D/Pos/Order_7
BenchmarkHilbert/2D/Pos/Order_7-8       26158227                42.13 ns/op              0 B/op          0 allocs/op
BenchmarkHilbert/2D/Pos/Order_8
BenchmarkHilbert/2D/Pos/Order_8-8       26913849                43.31 ns/op              0 B/op          0 allocs/op
BenchmarkHilbert/2D/Pos/Order_9
BenchmarkHilbert/2D/Pos/Order_9-8       26351476                43.45 ns/op              0 B/op          0 allocs/op
BenchmarkHilbert/2D/Pos/Order_10
BenchmarkHilbert/2D/Pos/Order_10-8              21631359                59.43 ns/op              0 B/op          0 allocs/op
BenchmarkHilbert/2D/Coord
BenchmarkHilbert/2D/Coord/Order_1
BenchmarkHilbert/2D/Coord/Order_1-8             176244006                7.655 ns/op           0 B/op              0 allocs/op
BenchmarkHilbert/2D/Coord/Order_2
BenchmarkHilbert/2D/Coord/Order_2-8             86029083                14.62 ns/op            0 B/op              0 allocs/op
BenchmarkHilbert/2D/Coord/Order_3
BenchmarkHilbert/2D/Coord/Order_3-8             88970055                16.94 ns/op            0 B/op              0 allocs/op
BenchmarkHilbert/2D/Coord/Order_4
BenchmarkHilbert/2D/Coord/Order_4-8             67142550                18.16 ns/op            0 B/op              0 allocs/op
BenchmarkHilbert/2D/Coord/Order_5
BenchmarkHilbert/2D/Coord/Order_5-8             54518832                20.30 ns/op            0 B/op              0 allocs/op
BenchmarkHilbert/2D/Coord/Order_6
BenchmarkHilbert/2D/Coord/Order_6-8             49024351                35.19 ns/op            0 B/op              0 allocs/op
BenchmarkHilbert/2D/Coord/Order_7
BenchmarkHilbert/2D/Coord/Order_7-8             39211636                25.72 ns/op            0 B/op              0 allocs/op
BenchmarkHilbert/2D/Coord/Order_8
BenchmarkHilbert/2D/Coord/Order_8-8             42119673                29.36 ns/op            0 B/op              0 allocs/op
BenchmarkHilbert/2D/Coord/Order_9
BenchmarkHilbert/2D/Coord/Order_9-8             34036394                39.42 ns/op            0 B/op              0 allocs/op
BenchmarkHilbert/2D/Coord/Order_10
BenchmarkHilbert/2D/Coord/Order_10-8            28563565                44.25 ns/op            0 B/op              0 allocs/op
BenchmarkHilbert/3D/Pos
BenchmarkHilbert/3D/Pos/Order_1
BenchmarkHilbert/3D/Pos/Order_1-8               45626558                23.72 ns/op            0 B/op              0 allocs/op
BenchmarkHilbert/3D/Pos/Order_2
BenchmarkHilbert/3D/Pos/Order_2-8               35097692                38.66 ns/op            0 B/op              0 allocs/op
BenchmarkHilbert/3D/Pos/Order_3
BenchmarkHilbert/3D/Pos/Order_3-8               24931842                53.10 ns/op            0 B/op              0 allocs/op
BenchmarkHilbert/3D/Pos/Order_4
BenchmarkHilbert/3D/Pos/Order_4-8               16363636                65.13 ns/op            0 B/op              0 allocs/op
BenchmarkHilbert/3D/Pos/Order_5
BenchmarkHilbert/3D/Pos/Order_5-8               17743531                73.46 ns/op            0 B/op              0 allocs/op
BenchmarkHilbert/3D/Pos/Order_6
BenchmarkHilbert/3D/Pos/Order_6-8               12450988                99.51 ns/op            0 B/op              0 allocs/op
BenchmarkHilbert/3D/Pos/Order_7
BenchmarkHilbert/3D/Pos/Order_7-8               10337702               134.7 ns/op             0 B/op              0 allocs/op
BenchmarkHilbert/3D/Pos/Order_8
BenchmarkHilbert/3D/Pos/Order_8-8                9742449               116.1 ns/op             0 B/op              0 allocs/op
BenchmarkHilbert/3D/Pos/Order_9
BenchmarkHilbert/3D/Pos/Order_9-8                9062575               145.2 ns/op             0 B/op              0 allocs/op
BenchmarkHilbert/3D/Pos/Order_10
BenchmarkHilbert/3D/Pos/Order_10-8               6391741               190.9 ns/op             0 B/op              0 allocs/op
BenchmarkHilbert/3D/Coord
BenchmarkHilbert/3D/Coord/Order_1
BenchmarkHilbert/3D/Coord/Order_1-8             79398408                20.42 ns/op            0 B/op              0 allocs/op
BenchmarkHilbert/3D/Coord/Order_2
BenchmarkHilbert/3D/Coord/Order_2-8             41471281                31.48 ns/op            0 B/op              0 allocs/op
BenchmarkHilbert/3D/Coord/Order_3
BenchmarkHilbert/3D/Coord/Order_3-8             30333662                42.67 ns/op            0 B/op              0 allocs/op
BenchmarkHilbert/3D/Coord/Order_4
BenchmarkHilbert/3D/Coord/Order_4-8             21393865                55.17 ns/op            0 B/op              0 allocs/op
BenchmarkHilbert/3D/Coord/Order_5
BenchmarkHilbert/3D/Coord/Order_5-8             17034688                73.57 ns/op            0 B/op              0 allocs/op
BenchmarkHilbert/3D/Coord/Order_6
BenchmarkHilbert/3D/Coord/Order_6-8             18116989                91.11 ns/op            0 B/op              0 allocs/op
BenchmarkHilbert/3D/Coord/Order_7
BenchmarkHilbert/3D/Coord/Order_7-8             13880876                86.80 ns/op            0 B/op              0 allocs/op
BenchmarkHilbert/3D/Coord/Order_8
BenchmarkHilbert/3D/Coord/Order_8-8             12011598               114.2 ns/op             0 B/op              0 allocs/op
BenchmarkHilbert/3D/Coord/Order_9
BenchmarkHilbert/3D/Coord/Order_9-8              9410091               111.1 ns/op             0 B/op              0 allocs/op
BenchmarkHilbert/3D/Coord/Order_10
BenchmarkHilbert/3D/Coord/Order_10-8             8406057               137.7 ns/op             0 B/op              0 allocs/op
BenchmarkHilbert/4D/Pos
BenchmarkHilbert/4D/Pos/Order_1
BenchmarkHilbert/4D/Pos/Order_1-8               44024707                26.23 ns/op            0 B/op              0 allocs/op
BenchmarkHilbert/4D/Pos/Order_2
BenchmarkHilbert/4D/Pos/Order_2-8               25374694                45.76 ns/op            0 B/op              0 allocs/op
BenchmarkHilbert/4D/Pos/Order_3
BenchmarkHilbert/4D/Pos/Order_3-8               23462036                70.91 ns/op            0 B/op              0 allocs/op
BenchmarkHilbert/4D/Pos/Order_4
BenchmarkHilbert/4D/Pos/Order_4-8               15241987                75.97 ns/op            0 B/op              0 allocs/op
BenchmarkHilbert/4D/Pos/Order_5
BenchmarkHilbert/4D/Pos/Order_5-8               13263390               100.6 ns/op             0 B/op              0 allocs/op
BenchmarkHilbert/4D/Pos/Order_6
BenchmarkHilbert/4D/Pos/Order_6-8                8060668               125.5 ns/op             0 B/op              0 allocs/op
BenchmarkHilbert/4D/Pos/Order_7
BenchmarkHilbert/4D/Pos/Order_7-8                8870961               130.7 ns/op             0 B/op              0 allocs/op
BenchmarkHilbert/4D/Pos/Order_8
BenchmarkHilbert/4D/Pos/Order_8-8                7693154               145.1 ns/op             0 B/op              0 allocs/op
BenchmarkHilbert/4D/Pos/Order_9
BenchmarkHilbert/4D/Pos/Order_9-8                6436678               187.7 ns/op             0 B/op              0 allocs/op
BenchmarkHilbert/4D/Pos/Order_10
BenchmarkHilbert/4D/Pos/Order_10-8               5614554               234.2 ns/op             0 B/op              0 allocs/op
BenchmarkHilbert/4D/Coord
BenchmarkHilbert/4D/Coord/Order_1
BenchmarkHilbert/4D/Coord/Order_1-8             70980118                16.07 ns/op            0 B/op              0 allocs/op
BenchmarkHilbert/4D/Coord/Order_2
BenchmarkHilbert/4D/Coord/Order_2-8             30415852                40.34 ns/op            0 B/op              0 allocs/op
BenchmarkHilbert/4D/Coord/Order_3
BenchmarkHilbert/4D/Coord/Order_3-8             20624990                61.09 ns/op            0 B/op              0 allocs/op
BenchmarkHilbert/4D/Coord/Order_4
BenchmarkHilbert/4D/Coord/Order_4-8             15923089                71.61 ns/op            0 B/op              0 allocs/op
BenchmarkHilbert/4D/Coord/Order_5
BenchmarkHilbert/4D/Coord/Order_5-8             13974184                88.66 ns/op            0 B/op              0 allocs/op
BenchmarkHilbert/4D/Coord/Order_6
BenchmarkHilbert/4D/Coord/Order_6-8             10822477               104.1 ns/op             0 B/op              0 allocs/op
BenchmarkHilbert/4D/Coord/Order_7
BenchmarkHilbert/4D/Coord/Order_7-8              9425492               157.4 ns/op             0 B/op              0 allocs/op
BenchmarkHilbert/4D/Coord/Order_8
BenchmarkHilbert/4D/Coord/Order_8-8              7466738               179.1 ns/op             0 B/op              0 allocs/op
BenchmarkHilbert/4D/Coord/Order_9
BenchmarkHilbert/4D/Coord/Order_9-8              6651352               174.4 ns/op             0 B/op              0 allocs/op
BenchmarkHilbert/4D/Coord/Order_10
BenchmarkHilbert/4D/Coord/Order_10-8             6441031               193.7 ns/op             0 B/op              0 allocs/op
PASS
ok      gonum.org/v1/gonum/spatial/curve        89.593s
</pre></details>

### DIFF
--- a/spatial/curve/hilbert.go
+++ b/spatial/curve/hilbert.go
@@ -11,7 +11,7 @@ import (
 
 // ErrUnderflow is returned by Hilbert curve constructors when the power is less
 // than 1.
-var ErrUnderflow = errors.New("order cannot be less than 1")
+var ErrUnderflow = errors.New("order is less than 1")
 
 // ErrOverflow is returned (wrapped) by Hilbert curve constructors when the
 // power would cause Len and Pos to overflow.
@@ -20,19 +20,22 @@ var ErrOverflow = errors.New("overflow int")
 // The size of an int. Taken from src/math/const.go.
 const intSize = 32 << (^uint(0) >> 63) // 32 or 64
 
+// Hilbert2D is a 2-dimensional Hilbert curve.
+type Hilbert2D struct{ order int }
+
 // NewHilbert2D constructs a [Hilbert2D] of the given order. NewHilbert2D
 // returns [ErrOverflow] (wrapped) if the order would cause Len and Pos to
 // overflow.
 func NewHilbert2D(order int) (Hilbert2D, error) {
-	v := Hilbert2D{Order: order}
+	v := Hilbert2D{order: order}
 
-	// The order must be greater than zero
+	// The order must be greater than zero.
 	if order < 1 {
 		return v, ErrUnderflow
 	}
 
 	// The product of the order and number of dimensions must not exceed or
-	// equal the size of an int
+	// equal the size of an int.
 	if order*2 >= intSize {
 		return v, fmt.Errorf("a 2-dimensional, %d-order Hilbert curve will %w", order, ErrOverflow)
 	}
@@ -40,59 +43,16 @@ func NewHilbert2D(order int) (Hilbert2D, error) {
 	return v, nil
 }
 
-// NewHilbert3D constructs a [Hilbert3D] of the given order. NewHilbert3D
-// returns [ErrOverflow] (wrapped) if the order would cause Len and Pos to
-// overflow.
-func NewHilbert3D(order int) (Hilbert3D, error) {
-	v := Hilbert3D{Order: order}
-
-	// The order must be greater than zero
-	if order < 1 {
-		return v, ErrUnderflow
-	}
-
-	// The product of the order and number of dimensions must not exceed or
-	// equal the size of an int
-	if order*3 >= intSize {
-		return v, fmt.Errorf("a 3-dimensional, %d-order Hilbert curve will %w", order, ErrOverflow)
-	}
-
-	return v, nil
-}
-
-// NewHilbert4D constructs a [Hilbert4D] of the given order. NewHilbert4D
-// returns [ErrOverflow] (wrapped) if the order would cause Len and Pos to
-// overflow.
-func NewHilbert4D(order int) (Hilbert4D, error) {
-	v := Hilbert4D{Order: order}
-
-	// The order must be greater than zero
-	if order < 1 {
-		return v, ErrUnderflow
-	}
-
-	// The product of the order and number of dimensions must not exceed or
-	// equal the size of an int
-	if order*4 >= intSize {
-		return v, fmt.Errorf("a 4-dimensional, %d-order Hilbert curve will %w", order, ErrOverflow)
-	}
-
-	return v, nil
-}
-
-// Hilbert2D is a 2-dimensional Hilbert curve.
-type Hilbert2D struct{ Order int }
-
 // Dims returns the spatial dimensions of the curve, which is {2ᵏ, 2ᵏ}, where k
 // is the order.
-func (h Hilbert2D) Dims() []int { return []int{1 << h.Order, 1 << h.Order} }
+func (h Hilbert2D) Dims() []int { return []int{1 << h.order, 1 << h.order} }
 
 // Len returns the length of the curve, which is 2ⁿᵏ, where n is the dimension
 // (2) and k is the order.
 //
 // Len will overflow if order is ≥ 16 on architectures where [int] is 32-bits,
 // or ≥ 32 on architectures where [int] is 64-bits.
-func (h Hilbert2D) Len() int { return 1 << (2 * h.Order) }
+func (h Hilbert2D) Len() int { return 1 << (2 * h.order) }
 
 func (h Hilbert2D) rot(n int, v []int, d int) {
 	switch d {
@@ -110,12 +70,13 @@ func (h Hilbert2D) rot(n int, v []int, d int) {
 // or ≥ 32 on architectures where [int] is 64-bits.
 func (h Hilbert2D) Pos(v []int) int {
 	var d int
-	for n := h.Order - 1; n >= 0; n-- {
+	const dims = 2
+	for n := h.order - 1; n >= 0; n-- {
 		rx := (v[0] >> n) & 1
 		ry := (v[1] >> n) & 1
 		rd := ry<<1 | (ry ^ rx)
-		d += rd << (2 * n)
-		h.rot(h.Order, v, rd)
+		d += rd << (dims * n)
+		h.rot(h.order, v, rd)
 	}
 	return d
 }
@@ -131,7 +92,7 @@ func (h Hilbert2D) Coord(dst []int, pos int) []int {
 	} else if len(dst) != 2 {
 		panic("len(dst) must equal 2")
 	}
-	for n := 0; n < h.Order; n++ {
+	for n := 0; n < h.order; n++ {
 		e := pos & 3
 		h.rot(n, dst[:], e)
 
@@ -145,18 +106,38 @@ func (h Hilbert2D) Coord(dst []int, pos int) []int {
 }
 
 // Hilbert3D is a 3-dimensional Hilbert curve.
-type Hilbert3D struct{ Order int }
+type Hilbert3D struct{ order int }
+
+// NewHilbert3D constructs a [Hilbert3D] of the given order. NewHilbert3D
+// returns [ErrOverflow] (wrapped) if the order would cause Len and Pos to
+// overflow.
+func NewHilbert3D(order int) (Hilbert3D, error) {
+	v := Hilbert3D{order: order}
+
+	// The order must be greater than zero.
+	if order < 1 {
+		return v, ErrUnderflow
+	}
+
+	// The product of the order and number of dimensions must not exceed or
+	// equal the size of an int.
+	if order*3 >= intSize {
+		return v, fmt.Errorf("a 3-dimensional, %d-order Hilbert curve will %w", order, ErrOverflow)
+	}
+
+	return v, nil
+}
 
 // Dims returns the spatial dimensions of the curve, which is {2ᵏ, 2ᵏ, 2ᵏ}, where
 // k is the order.
-func (h Hilbert3D) Dims() []int { return []int{1 << h.Order, 1 << h.Order, 1 << h.Order} }
+func (h Hilbert3D) Dims() []int { return []int{1 << h.order, 1 << h.order, 1 << h.order} }
 
 // Len returns the length of the curve, which is 2ⁿᵏ, where n is the dimension
 // (3) and k is the order.
 //
 // Len will overflow if order is ≥ 11 on architectures where [int] is 32-bits,
 // or ≥ 21 on architectures where [int] is 64-bits.
-func (h Hilbert3D) Len() int { return 1 << (3 * h.Order) }
+func (h Hilbert3D) Len() int { return 1 << (3 * h.order) }
 
 func (h Hilbert3D) rot(reverse bool, n int, v []int, d int) {
 	switch d {
@@ -180,13 +161,14 @@ func (h Hilbert3D) rot(reverse bool, n int, v []int, d int) {
 // or ≥ 21 on architectures where [int] is 64-bits.
 func (h Hilbert3D) Pos(v []int) int {
 	var d int
-	for n := h.Order - 1; n >= 0; n-- {
+	const dims = 3
+	for n := h.order - 1; n >= 0; n-- {
 		rx := (v[0] >> n) & 1
 		ry := (v[1] >> n) & 1
 		rz := (v[2] >> n) & 1
 		rd := rz<<2 | (rz^ry)<<1 | (rz ^ ry ^ rx)
-		d += rd << (3 * n)
-		h.rot(false, h.Order, v, rd)
+		d += rd << (dims * n)
+		h.rot(false, h.order, v, rd)
 	}
 	return d
 }
@@ -202,7 +184,7 @@ func (h Hilbert3D) Coord(dst []int, pos int) []int {
 	} else if len(dst) != 3 {
 		panic("len(dst) must equal 3")
 	}
-	for n := 0; n < h.Order; n++ {
+	for n := 0; n < h.order; n++ {
 		e := pos & 7
 		h.rot(true, n, dst[:], e)
 
@@ -218,18 +200,38 @@ func (h Hilbert3D) Coord(dst []int, pos int) []int {
 }
 
 // Hilbert4D is a 4-dimensional Hilbert curve.
-type Hilbert4D struct{ Order int }
+type Hilbert4D struct{ order int }
+
+// NewHilbert4D constructs a [Hilbert4D] of the given order. NewHilbert4D
+// returns [ErrOverflow] (wrapped) if the order would cause Len and Pos to
+// overflow.
+func NewHilbert4D(order int) (Hilbert4D, error) {
+	v := Hilbert4D{order: order}
+
+	// The order must be greater than zero.
+	if order < 1 {
+		return v, ErrUnderflow
+	}
+
+	// The product of the order and number of dimensions must not exceed or
+	// equal the size of an int.
+	if order*4 >= intSize {
+		return v, fmt.Errorf("a 4-dimensional, %d-order Hilbert curve will %w", order, ErrOverflow)
+	}
+
+	return v, nil
+}
 
 // Dims returns the spatial dimensions of the curve, which is {2ᵏ, 2ᵏ, 2ᵏ, 2ᵏ},
 // where k is the order.
-func (h Hilbert4D) Dims() []int { return []int{1 << h.Order, 1 << h.Order, 1 << h.Order, 1 << h.Order} }
+func (h Hilbert4D) Dims() []int { return []int{1 << h.order, 1 << h.order, 1 << h.order, 1 << h.order} }
 
 // Len returns the length of the curve, which is 2ⁿᵏ, where n is the dimension
 // (4) and k is the order.
 //
 // Len will overflow if order is ≥ 8 on architectures where [int] is 32-bits, or
 // ≥ 16 on architectures where [int] is 64-bits.
-func (h Hilbert4D) Len() int { return 1 << (4 * h.Order) }
+func (h Hilbert4D) Len() int { return 1 << (4 * h.order) }
 
 func (h Hilbert4D) rot(reverse bool, n int, v []int, d int) {
 	switch d {
@@ -261,16 +263,16 @@ func (h Hilbert4D) rot(reverse bool, n int, v []int, d int) {
 // ≥ 16 on architectures where [int] is 64-bits.
 func (h Hilbert4D) Pos(v []int) int {
 	var d int
-	N := 4
-	for n := h.Order - 1; n >= 0; n-- {
+	const dims = 4
+	for n := h.order - 1; n >= 0; n-- {
 		var e int
-		for i := N - 1; i >= 0; i-- {
+		for i := dims - 1; i >= 0; i-- {
 			v := v[i] >> n & 1
 			e = e<<1 | (e^v)&1
 		}
 
-		d += e << (N * n)
-		h.rot(false, h.Order, v, e)
+		d += e << (dims * n)
+		h.rot(false, h.order, v, e)
 	}
 	return d
 }
@@ -287,7 +289,7 @@ func (h Hilbert4D) Coord(dst []int, pos int) []int {
 		panic("len(dst) must equal 4")
 	}
 	N := 4
-	for n := 0; n < h.Order; n++ {
+	for n := 0; n < h.order; n++ {
 		e := pos & (1<<N - 1)
 		h.rot(true, n, dst[:], e)
 
--- a/spatial/curve/hilbert.go
+++ b/spatial/curve/hilbert.go
@@ -1,0 +1,332 @@
+// Copyright ©2024 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package curve
+
+import (
+	"errors"
+	"fmt"
+)
+
+// ErrUnderflow is returned by Hilbert curve constructors when the power is less
+// than 1.
+var ErrUnderflow = errors.New("order cannot be less than 1")
+
+// ErrOverflow is returned (wrapped) by Hilbert curve constructors when the
+// power would cause Len and Pos to overflow.
+var ErrOverflow = errors.New("overflow int")
+
+// The size of an int. Taken from src/math/const.go.
+const intSize = 32 << (^uint(0) >> 63) // 32 or 64
+
+// NewHilbert2D constructs a [Hilbert2D] of the given order. NewHilbert2D
+// returns [ErrOverflow] (wrapped) if the order would cause Len and Pos to
+// overflow.
+func NewHilbert2D(order int) (Hilbert2D, error) {
+	v := Hilbert2D{Order: order}
+
+	// The order must be greater than zero
+	if order < 1 {
+		return v, ErrUnderflow
+	}
+
+	// The product of the order and number of dimensions must not exceed or
+	// equal the size of an int
+	if order*2 >= intSize {
+		return v, fmt.Errorf("a 2-dimensional, %d-order Hilbert curve will %w", order, ErrOverflow)
+	}
+
+	return v, nil
+}
+
+// NewHilbert3D constructs a [Hilbert3D] of the given order. NewHilbert3D
+// returns [ErrOverflow] (wrapped) if the order would cause Len and Pos to
+// overflow.
+func NewHilbert3D(order int) (Hilbert3D, error) {
+	v := Hilbert3D{Order: order}
+
+	// The order must be greater than zero
+	if order < 1 {
+		return v, ErrUnderflow
+	}
+
+	// The product of the order and number of dimensions must not exceed or
+	// equal the size of an int
+	if order*3 >= intSize {
+		return v, fmt.Errorf("a 3-dimensional, %d-order Hilbert curve will %w", order, ErrOverflow)
+	}
+
+	return v, nil
+}
+
+// NewHilbert4D constructs a [Hilbert4D] of the given order. NewHilbert4D
+// returns [ErrOverflow] (wrapped) if the order would cause Len and Pos to
+// overflow.
+func NewHilbert4D(order int) (Hilbert4D, error) {
+	v := Hilbert4D{Order: order}
+
+	// The order must be greater than zero
+	if order < 1 {
+		return v, ErrUnderflow
+	}
+
+	// The product of the order and number of dimensions must not exceed or
+	// equal the size of an int
+	if order*4 >= intSize {
+		return v, fmt.Errorf("a 4-dimensional, %d-order Hilbert curve will %w", order, ErrOverflow)
+	}
+
+	return v, nil
+}
+
+// Hilbert2D is a 2-dimensional Hilbert curve.
+type Hilbert2D struct{ Order int }
+
+// Dims returns the spatial dimensions of the curve, which is {2ᵏ, 2ᵏ}, where k
+// is the order.
+func (h Hilbert2D) Dims() []int { return []int{1 << h.Order, 1 << h.Order} }
+
+// Len returns the length of the curve, which is 2ⁿᵏ, where n is the dimension
+// (2) and k is the order.
+//
+// Len will overflow if order is ≥ 16 on architectures where [int] is 32-bits,
+// or ≥ 32 on architectures where [int] is 64-bits.
+func (h Hilbert2D) Len() int { return 1 << (2 * h.Order) }
+
+func (h Hilbert2D) rot(n int, v []int, d int) {
+	switch d {
+	case 0:
+		swap{0, 1}.do(n, v)
+	case 3:
+		flip{0, 1}.do(n, v)
+	}
+}
+
+// Pos returns the linear position of the 3-spatial coordinate v along the
+// curve. Pos modifies v.
+//
+// Pos will overflow if order is ≥ 16 on architectures where [int] is 32-bits,
+// or ≥ 32 on architectures where [int] is 64-bits.
+func (h Hilbert2D) Pos(v []int) int {
+	var d int
+	for n := h.Order - 1; n >= 0; n-- {
+		rx := (v[0] >> n) & 1
+		ry := (v[1] >> n) & 1
+		rd := ry<<1 | (ry ^ rx)
+		d += rd << (2 * n)
+		h.rot(h.Order, v, rd)
+	}
+	return d
+}
+
+// Coord writes the spatial coordinates of pos to dst and returns it. If dst is
+// nil, Coord allocates a new slice of length 2; otherwise Coord is
+// allocation-free.
+//
+// Coord panics if dst is not nil and len(dst) is not equal to 2.
+func (h Hilbert2D) Coord(dst []int, pos int) []int {
+	if dst == nil {
+		dst = make([]int, 2)
+	} else if len(dst) != 2 {
+		panic("len(dst) must equal 2")
+	}
+	for n := 0; n < h.Order; n++ {
+		e := pos & 3
+		h.rot(n, dst[:], e)
+
+		ry := e >> 1
+		rx := (e>>0 ^ e>>1) & 1
+		dst[0] += rx << n
+		dst[1] += ry << n
+		pos >>= 2
+	}
+	return dst
+}
+
+// Hilbert3D is a 3-dimensional Hilbert curve.
+type Hilbert3D struct{ Order int }
+
+// Dims returns the spatial dimensions of the curve, which is {2ᵏ, 2ᵏ, 2ᵏ}, where
+// k is the order.
+func (h Hilbert3D) Dims() []int { return []int{1 << h.Order, 1 << h.Order, 1 << h.Order} }
+
+// Len returns the length of the curve, which is 2ⁿᵏ, where n is the dimension
+// (3) and k is the order.
+//
+// Len will overflow if order is ≥ 11 on architectures where [int] is 32-bits,
+// or ≥ 21 on architectures where [int] is 64-bits.
+func (h Hilbert3D) Len() int { return 1 << (3 * h.Order) }
+
+func (h Hilbert3D) rot(reverse bool, n int, v []int, d int) {
+	switch d {
+	case 0:
+		do2(reverse, n, v, swap{1, 2}, swap{0, 2})
+	case 1, 2:
+		do2(reverse, n, v, swap{0, 2}, swap{1, 2})
+	case 3, 4:
+		invert{0, 1}.do(n, v)
+	case 5, 6:
+		do2(reverse, n, v, flip{0, 2}, flip{1, 2})
+	case 7:
+		do2(reverse, n, v, flip{1, 2}, flip{0, 2})
+	}
+}
+
+// Pos returns the linear position of the 4-spatial coordinate v along the
+// curve. Pos modifies v.
+//
+// Pos will overflow if order is ≥ 11 on architectures where [int] is 32-bits,
+// or ≥ 21 on architectures where [int] is 64-bits.
+func (h Hilbert3D) Pos(v []int) int {
+	var d int
+	for n := h.Order - 1; n >= 0; n-- {
+		rx := (v[0] >> n) & 1
+		ry := (v[1] >> n) & 1
+		rz := (v[2] >> n) & 1
+		rd := rz<<2 | (rz^ry)<<1 | (rz ^ ry ^ rx)
+		d += rd << (3 * n)
+		h.rot(false, h.Order, v, rd)
+	}
+	return d
+}
+
+// Coord writes the spatial coordinates of pos to dst and returns it. If dst is
+// nil, Coord allocates a new slice of length 3; otherwise Coord is
+// allocation-free.
+//
+// Coord panics if dst is not nil and len(dst) is not equal to 3.
+func (h Hilbert3D) Coord(dst []int, pos int) []int {
+	if dst == nil {
+		dst = make([]int, 3)
+	} else if len(dst) != 3 {
+		panic("len(dst) must equal 3")
+	}
+	for n := 0; n < h.Order; n++ {
+		e := pos & 7
+		h.rot(true, n, dst[:], e)
+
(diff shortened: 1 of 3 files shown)


firelizzard18:
@kortschak @vladimir-ch I'm getting back to the project that I need #1691 for, so I'm revisiting this Hilbert curve code. I addressed almost all of the remaining discussions from #1695, including the `Space` and `Curve` methods to `Coord` and `Pos`. The [remaining](https://github.com/gonum/gonum/pull/1695#discussion_r711485982) [discussions](https://github.com/gonum/gonum/pull/1695#issuecomment-888678226) are about off-nominal behavior (when the caller passes out of bounds values or equivalent).

firelizzard18:
Do you think it's valuable to define a `Curve` interface in the package? For example:

```go

type Curve interface {
	Dims() []int
	Len() int
	Pos(coord []int) int
	Coord(pos int) []int
}
```

firelizzard18:
I eliminated all allocations from `Pos` and all those that were avoidable from `Coord` - excepting the return value - by replacing `type ops []op` and its method `do` with a function `do2` that uses generics to avoid allocating interface values. On my laptop I got the following, before and after:

<details><summary>benchcmp ns/op</summary><pre>
benchmark                               old ns/op     new ns/op     delta
BenchmarkHilbert/2D/Pos/Order_1-8       13.5          10.3          -24.15%
BenchmarkHilbert/2D/Pos/Order_2-8       16.2          15.2          -6.06%
BenchmarkHilbert/2D/Pos/Order_3-8       19.6          18.4          -5.92%
BenchmarkHilbert/2D/Pos/Order_4-8       24.4          22.7          -6.90%
BenchmarkHilbert/2D/Coord/Order_1-8     41.3          41.1          -0.53%
BenchmarkHilbert/2D/Coord/Order_2-8     46.2          44.7          -3.22%
BenchmarkHilbert/2D/Coord/Order_3-8     47.6          49.1          +3.00%
BenchmarkHilbert/2D/Coord/Order_4-8     56.4          51.0          -9.71%
BenchmarkHilbert/3D/Pos/Order_1-8       87.9          24.0          -72.71%
BenchmarkHilbert/3D/Pos/Order_2-8       172           42.4          -75.39%
BenchmarkHilbert/3D/Pos/Order_3-8       266           54.9          -79.33%
BenchmarkHilbert/3D/Pos/Order_4-8       277           72.8          -73.71%
BenchmarkHilbert/3D/Coord/Order_1-8     82.1          90.8          +10.64%
BenchmarkHilbert/3D/Coord/Order_2-8     233           128           -45.19%
BenchmarkHilbert/3D/Coord/Order_3-8     234           118           -49.53%
BenchmarkHilbert/3D/Coord/Order_4-8     284           142           -49.93%
BenchmarkHilbert/4D/Pos/Order_1-8       110           33.1          -70.08%
BenchmarkHilbert/4D/Pos/Order_2-8       237           59.4          -74.97%
BenchmarkHilbert/4D/Pos/Order_3-8       480           75.4          -84.30%
BenchmarkHilbert/4D/Pos/Order_4-8       553           97.1          -82.43%
BenchmarkHilbert/4D/Coord/Order_1-8     265           104           -60.82%
BenchmarkHilbert/4D/Coord/Order_2-8     343           117           -65.98%
BenchmarkHilbert/4D/Coord/Order_3-8     224           134           -40.07%
BenchmarkHilbert/4D/Coord/Order_4-8     425           158           -62.75%
</pre></details>
<details><summary>benchcmp allocs</summary><pre>
benchmark                               old allocs     new allocs     delta
BenchmarkHilbert/2D/Pos/Order_1-8       0              0              +0.00%
BenchmarkHilbert/2D/Pos/Order_2-8       0              0              +0.00%
BenchmarkHilbert/2D/Pos/Order_3-8       0              0              +0.00%
BenchmarkHilbert/2D/Pos/Order_4-8       0              0              +0.00%
BenchmarkHilbert/2D/Coord/Order_1-8     1              1              +0.00%
BenchmarkHilbert/2D/Coord/Order_2-8     1              1              +0.00%
BenchmarkHilbert/2D/Coord/Order_3-8     1              1              +0.00%
BenchmarkHilbert/2D/Coord/Order_4-8     1              1              +0.00%
BenchmarkHilbert/3D/Pos/Order_1-8       2              0              -100.00%
BenchmarkHilbert/3D/Pos/Order_2-8       4              0              -100.00%
BenchmarkHilbert/3D/Pos/Order_3-8       6              0              -100.00%
BenchmarkHilbert/3D/Pos/Order_4-8       6              0              -100.00%
BenchmarkHilbert/3D/Coord/Order_1-8     2              2              +0.00%
BenchmarkHilbert/3D/Coord/Order_2-8     6              2              -66.67%
BenchmarkHilbert/3D/Coord/Order_3-8     6              2              -66.67%
BenchmarkHilbert/3D/Coord/Order_4-8     6              2              -66.67%
BenchmarkHilbert/4D/Pos/Order_1-8       2              0              -100.00%
BenchmarkHilbert/4D/Pos/Order_2-8       4              0              -100.00%
BenchmarkHilbert/4D/Pos/Order_3-8       6              0              -100.00%
BenchmarkHilbert/4D/Pos/Order_4-8       6              0              -100.00%
BenchmarkHilbert/4D/Coord/Order_1-8     4              2              -50.00%
BenchmarkHilbert/4D/Coord/Order_2-8     6              2              -66.67%
BenchmarkHilbert/4D/Coord/Order_3-8     6              2              -66.67%
BenchmarkHilbert/4D/Coord/Order_4-8     8              2              -75.00%
</pre></details>
<details><summary>benchcmp bytes</summary><pre>
benchmark                               old bytes     new bytes     delta
BenchmarkHilbert/2D/Pos/Order_1-8       0             0             +0.00%
BenchmarkHilbert/2D/Pos/Order_2-8       0             0             +0.00%
BenchmarkHilbert/2D/Pos/Order_3-8       0             0             +0.00%
BenchmarkHilbert/2D/Pos/Order_4-8       0             0             +0.00%
BenchmarkHilbert/2D/Coord/Order_1-8     16            16            +0.00%
BenchmarkHilbert/2D/Coord/Order_2-8     16            16            +0.00%
BenchmarkHilbert/2D/Coord/Order_3-8     16            16            +0.00%
BenchmarkHilbert/2D/Coord/Order_4-8     16            16            +0.00%
BenchmarkHilbert/3D/Pos/Order_1-8       32            0             -100.00%
BenchmarkHilbert/3D/Pos/Order_2-8       64            0             -100.00%
BenchmarkHilbert/3D/Pos/Order_3-8       96            0             -100.00%
BenchmarkHilbert/3D/Pos/Order_4-8       96            0             -100.00%
BenchmarkHilbert/3D/Coord/Order_1-8     48            48            +0.00%
BenchmarkHilbert/3D/Coord/Order_2-8     112           48            -57.14%
BenchmarkHilbert/3D/Coord/Order_3-8     112           48            -57.14%
BenchmarkHilbert/3D/Coord/Order_4-8     112           48            -57.14%
BenchmarkHilbert/4D/Pos/Order_1-8       32            0             -100.00%
BenchmarkHilbert/4D/Pos/Order_2-8       64            0             -100.00%
BenchmarkHilbert/4D/Pos/Order_3-8       96            0             -100.00%
BenchmarkHilbert/4D/Pos/Order_4-8       96            0             -100.00%
BenchmarkHilbert/4D/Coord/Order_1-8     96            64            -33.33%
BenchmarkHilbert/4D/Coord/Order_2-8     128           64            -50.00%
BenchmarkHilbert/4D/Coord/Order_3-8     128           64            -50.00%
BenchmarkHilbert/4D/Coord/Order_4-8     160           64            -60.00%
</pre></details>

firelizzard18:
I replaced `CoordND(pos int) [N]int` with `WriteCoordTo(pos int, v []int)` and updated the benchmark to use that. That eliminated *all* allocations in the benchmark and improved the performance of `WriteCoordTo` (vs `CoordND`) by another ~50%. I'm doing this because my application will be making heavy use of the R-tree (once that's implemented) and the performance of lookups will meaningfully impact the user experience.

firelizzard18:
WDYT about making these generic? E.g. `type HilbertND[V constraints.Integer] struct{ Order V }`. That would allow the caller to use any integer type, including uint64 if 32-bit overflow is a concern.

spatial/curve/hilbert.go
```diff
@@ -0,0 +1,241 @@
...
+func (h Hilbert2D) rot(n int, v []int, d int) {
+	switch d {
+	case 0:
+		swap{0, 1}.do(n, v)
+	case 3:
+		flip{0, 1}.do(n, v)
+	}
+}
+
+// Pos returns the linear position of the spatial coordinate along the curve.
```
kortschak, line 29:
```suggestion
// Pos returns the linear position of the 2-spatial coordinate v along the curve.
```
?

spatial/curve/hilbert.go
```diff
@@ -0,0 +1,241 @@
...
+		rx := (v[0] >> n) & 1
+		ry := (v[1] >> n) & 1
+		rd := ry<<1 | (ry ^ rx)
+		d += rd << (2 * n)
+		h.rot(h.Order, v, rd)
+	}
+	return d
+}
+
+// WriteCoordTo writes the spatial coordinates of pos to v.
```
kortschak, line 45: Note that `len(v) >= 2` otherwise `WriteCoordTo` will panic. Consider making this strict (i.e. `len(v) == 2`) to be consistent with other length requirement check that we impose elsewhere.

spatial/curve/hilbert.go
```diff
@@ -0,0 +1,241 @@
...
+		pos >>= 2
+	}
+}
+
+// Coord returns the spatial coordinates of pos as a slice.
+func (h Hilbert2D) Coord(pos int) []int {
+	v := make([]int, 2)
+	h.WriteCoordTo(pos, v)
+	return v
+}
```
kortschak, line 64: Do we need this? It seems like a helper that suggests inefficient client code.

spatial/curve/hilbert.go
```diff
@@ -0,0 +1,241 @@
...
+type Hilbert3D struct{ Order int }
+
+// Dims returns the spatial dimensions of the curve, which is {2ᵏ, 2ᵏ, 2ᵏ}, where
+// k is the order.
+func (h Hilbert3D) Dims() []int { return []int{1 << h.Order, 1 << h.Order, 1 << h.Order} }
+
+// Len returns the length of the curve, which is 2ⁿᵏ, where n is the dimension
+// (3) and k is the order.
+//
+// Len will overflow on a 32-bit architecture if the order is ≥ 11.
```
kortschak, line 76: Warning above (or constructor) and for 64-bit.

spatial/curve/hilbert.go
```diff
@@ -0,0 +1,241 @@
...
+	case 3, 4:
+		invert{0, 1}.do(n, v)
+	case 5, 6:
+		do2(reverse, n, v, flip{0, 2}, flip{1, 2})
+	case 7:
+		do2(reverse, n, v, flip{1, 2}, flip{0, 2})
+	}
+}
+
+// Pos returns the linear position of the spatial coordinate along the curve.
```
kortschak, line 94:
```suggestion
// Pos returns the linear position of the 3-spatial coordinate v along the curve.
```

spatial/curve/hilbert.go
```diff
@@ -0,0 +1,241 @@
...
+		pos >>= 3
+	}
+}
+
+// Coord returns the spatial coordinates of pos as a slice.
+func (h Hilbert3D) Coord(pos int) []int {
+	v := make([]int, 3)
+	h.WriteCoordTo(pos, v)
+	return v[:]
+}
```
kortschak, line 132: Needed?

spatial/curve/hilbert.go
```diff
@@ -0,0 +1,241 @@
...
+		ry := (v[1] >> n) & 1
+		rz := (v[2] >> n) & 1
+		rd := rz<<2 | (rz^ry)<<1 | (rz ^ ry ^ rx)
+		d += rd << (3 * n)
+		h.rot(false, h.Order, v, rd)
+	}
+	return d
+}
+
+// WriteCoordTo writes the spatial coordinates of pos to v.
```
kortschak, line 111: Size check and note.

spatial/curve/hilbert.go
```diff
@@ -0,0 +1,241 @@
...
+type Hilbert4D struct{ Order int }
+
+// Dims returns the spatial dimensions of the curve, which is {2ᵏ, 2ᵏ, 2ᵏ, 2ᵏ},
+// where k is the order.
+func (h Hilbert4D) Dims() []int { return []int{1 << h.Order, 1 << h.Order, 1 << h.Order, 1 << h.Order} }
+
+// Len returns the length of the curve, which is 2ⁿᵏ, where n is the dimension
+// (4) and k is the order.
+//
+// Len will overflow on a 32-bit architecture if the order is ≥ 8.
```
kortschak, line 144: Same

spatial/curve/hilbert.go
```diff
@@ -0,0 +1,241 @@
...
+		do2(reverse, n, v, flip{0, 3}, flip{1, 3})
+	case 15:
+		do2(reverse, n, v, flip{1, 3}, flip{0, 3})
+	}
+}
+
+// Pos returns the linear position of the spatial coordinate along the curve.
+// Pos modifies v.
+//
+// Pos will overflow on a 32-bit architecture if the order is ≥ 8.
```
kortschak, line 173: Same as comments above.

spatial/curve/hilbert.go
```diff
@@ -0,0 +1,241 @@
...
+		pos >>= N
+	}
+}
+
+// Coord returns the spatial coordinates of pos as a slice.
+func (h Hilbert4D) Coord(pos int) []int {
+	v := make([]int, 4)
+	h.WriteCoordTo(pos, v)
+	return v
+}
```
kortschak, line 210: Needed?

spatial/curve/hilbert.go
```diff
@@ -0,0 +1,241 @@
...
+
+// swap and invert I and J
+type flip struct{ i, j int }
+
+func (c flip) do(n int, v []int) { v[c.i], v[c.j] = v[c.j]^(1<<n-1), v[c.i]^(1<<n-1) }
+
+// do2 executes the given operations, optionally in reverse.
+//
+// Generic specialization reduces allocation (because it can eliminate interface
+// value boxing) and improves performance
```
kortschak, line 232: Nice that you found this. I was not aware that this was the case.

spatial/curve/hilbert_test.go
kortschak, line 1: Do you mind changing %02X to %02x for legibility?

spatial/curve/hilbert_test.go
```diff
@@ -0,0 +1,382 @@
...
+		return Hilbert2D{Order: order}
+	case 3:
+		return Hilbert3D{Order: order}
+	case 4:
+		return Hilbert4D{Order: order}
+	}
+	panic("invalid dimension")
+}
+
+func testCurve(t *testing.T, c curve) {
```
kortschak, line 267: It would be helpful to have a comment explaining what this is testing.

spatial/curve/hilbert_test.go
```diff
@@ -0,0 +1,382 @@
...
+
+	D := 1
+	for _, v := range c.Dims() {
+		D *= v
+	}
+	for d := 0; d < D-1; d++ {
+		v, u := m[d], m[d+1]
+		if !adjacent(v, u) {
+			t.Logf("points %x and %x are not adjacent", d, d+1)
+			t.Logf("    %v -> %v", v, u)
```
kortschak, line 300: If this is done in a single log with a `\n` we reduce boilerplate in the output.

spatial/curve/hilbert_test.go
```diff
@@ -0,0 +1,382 @@
...
+		dimRange(dim-1, size, v, fn)
+	}
+}
+
+// v may get mangled by Curve, so copy it.
+func dup(v []int) []int {
+	u := make([]int, len(v))
+	copy(u, v)
+	return u
+}
```
kortschak, line 328: Replace use of this with `slices.Clone` and delete.

spatial/curve/hilbert_test.go
```diff
@@ -0,0 +1,382 @@
...
+		if x < -1 || 1 < x {
+			return false
+		}
+		n++
+	}
+
+	return n == 1
+}
+
+func testCurveCase(t *testing.T, c curve, order int, want []int) {
```
kortschak, line 352: Add a comment explaining what this is testing.

kortschak:
> Do you think it's valuable to define a `Curve` interface in the package? For example:
> 
> ```
> type Curve interface {
> 	Dims() []int
> 	Len() int
> 	Pos(coord []int) int
> 	Coord(pos int) []int
> }
> ```

I don't think this is necessary at this stage.

> WDYT about making these generic? E.g. type HilbertND[V constraints.Integer] struct{ Order V }. That would allow the caller to use any integer type, including uint64 if 32-bit overflow is a concern.

If this is a concern, why not just make it `uint64` from the outset?

firelizzard18:
> If this is a concern, why not just make it `uint64` from the outset?

I thought there was some reason we decided on `int` in #1695 but I can't find any mention of that so I'll switch it.

firelizzard18:
Possibly this comment is what I was thinking of: https://github.com/gonum/gonum/pull/1695#issuecomment-883882068

kortschak:
Thanks for following that up. How likely do you think overflowing int will be a problem?

kortschak:
Don't go ahead with a switch to uint64 until we explore this.

firelizzard18:
> How likely do you think overflowing int will be a problem. 

It seems unlikely. The vast majority of PCs/servers/etc are 64-bit, so the only scenario I can think of is embedded systems. I can imagine some scenarios where you might use a Hilbert curve on an embedded system, for example satellites (though the scenario I'm thinking of used a k-d tree) but if someone does run into that scenario, I'd suggest they create a custom implementation.

> Don't go ahead with a switch to uint64 until we explore this.

I started changing things, and then I started thinking about `uint64` vs `int64` and the more I think about it the more natural it seems to stick with `int`. It would be easiest to overflow a 4D Hilbert curve and I'd guess the majority of use cases will be using the 2D or 3D curve.

kortschak:
> if someone does run into that scenario, I'd suggest they create a custom implementation.

Agree

>  It would be easiest to overflow a 4D Hilbert curve and I'd guess the majority of use cases will be using the 2D or 3D curve.

Agree. I think any concerns here can be readily addressed by making curve construction be guarded within a constructor that returns the type and an error.

spatial/curve/hilbert.go
```diff
@@ -0,0 +1,241 @@
...
+type Hilbert2D struct{ Order int }
+
+// Dims returns the spatial dimensions of the curve, which is {2ᵏ, 2ᵏ}, where k
+// is the order.
+func (h Hilbert2D) Dims() []int { return []int{1 << h.Order, 1 << h.Order} }
+
+// Len returns the length of the curve, which is 2ⁿᵏ, where n is the dimension
+// (2) and k is the order.
+//
+// Len will overflow on a 32-bit architecture if the order is ≥ 16.
```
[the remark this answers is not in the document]
firelizzard18: I updated these comments to "on architectures where int is 32-bits" instead of "32-bit architecture" and added a comment for "where int is 64-bits".

spatial/curve/hilbert.go
firelizzard18: Done

spatial/curve/hilbert.go
firelizzard18: :shrug:. If I were writing code where the performance of `Coord` was unimportant, I'd rather use this signature. WriteCoordTo is kind of awkward.

spatial/curve/hilbert.go
firelizzard18: Done

spatial/curve/hilbert.go
firelizzard18: Done

spatial/curve/hilbert.go
firelizzard18: Done

spatial/curve/hilbert.go
firelizzard18: Done

spatial/curve/hilbert.go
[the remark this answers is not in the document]
kortschak: I don't think that makes a difference; we can still see an overflow with b4-bit ints.

spatial/curve/hilbert.go
firelizzard18: Done

spatial/curve/hilbert.go
firelizzard18: Done

spatial/curve/hilbert.go
firelizzard18: Given that we're removing `Coord`, should the other method stay `WriteCoordTo`?

kortschak: Good point. Maybe `Coord(dst []int, pos int)`? Also, on a survey, maybe we *can* have something like `Coord(dst []int, pos int) []int` which allocates the slice is `dst` is nil. I know that this goes against what I wrote above, but it's consistent both with existing code we have here and with similar functions that are in the standard library.

spatial/curve/hilbert.go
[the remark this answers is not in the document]
firelizzard18: > I do think it would be better to include an equivalent warning on the constructor

What constructor? I could add `func NewHilbert(order, dimension int) (???, error)` but that would require something like `type Curve interface`. Or I could add `func NewHilbert2D(order int) (Hilbert2D, error)` etc but the implementation would be trivial (not worth a function IMO).

spatial/curve/hilbert.go
firelizzard18: I like `Coord(dst []int, pos int) []int`. As you say that's an established pattern and it addresses both use cases.

spatial/curve/hilbert.go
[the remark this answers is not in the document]
kortschak: I was thinking func `NewHilbert2D(order int) (Hilbert2D, error)` and friends. The function is trivial, but makes the overflow checks documented in code.

spatial/curve/hilbert_test.go
```diff
@@ -342,20 +356,22 @@ func adjacent(v, u []int) bool {
...
-func testCurveCase(t *testing.T, c curve, order int, want []int) {
+// testCurveCase verifies that the curve produces the expected sequence of
+// values.
+func testCurveCase(t *testing.T, c curve, order int, expected []int) {
 	t.Helper()
 
 	dim := len(c.Dims())
-	got := make([]int, len(want))
-	for i := range want {
+	actual := make([]int, len(expected))
```
kortschak, line 365: Sad. Not worth reverting, but we lean towards the English versus the Latin.

spatial/curve/hilbert.go
firelizzard18: I knew that `var x any = valueType` results in an allocation (if `x` escapes to the heap) and I could see in profiles that `ops.do` was a significant source of allocations so I inferred that boxing the operations structs as `type op inferface` was the source of those allocations. Though (the reference implementation of) Go uses gc-shape instead of true templating (as in C++), the resulting machine code is identical to a templated version I believe, so I figured this would work. Having two separate type parameters for the two function parameters means that the end result is identical to defining a separate `do2` function hard coded for each combination we use. I *think* there's only one actual do2 implementation generated since all the operations have the same shape, but it's possible the compiler is generating a separate implementation for each combination.

spatial/curve/hilbert.go
firelizzard18: Done

spatial/curve/hilbert_test.go
firelizzard18: I tend towards got/want but using "expected" made it easier to write the function comment naturally. That way I can say "produces the expected sequence of values" and it's obvious (I hope) that expected sequence corresponds to the 'expected' parameter.

firelizzard18: I suppose "desired sequence" and 'want' might be sufficiently clear?

spatial/curve/hilbert_test.go
firelizzard18: I added a comment but I don't know that it's useful. Ultimately the answer is, "See that test case above the call? This function is verifying that the curve produces that sequence of values when you call Pos on each corresponding spatial coordinate, and vice versa." Every time I tried to go deeper than that it turned into jargon-y word salad.

spatial/curve/hilbert.go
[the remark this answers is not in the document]
firelizzard18: Done

kortschak:
This
```
2024-03-10T23:13:49.6141071Z ##[error]spatial/curve/hilbert_test.go:10:2: package slices is not in GOROOT (/home/runner/work/gonum/setup-go-faster/go/1.19.13/x64/src/slices) (compile)
```
is due to issues with how staticcheck is run by default. See #1940.

spatial/curve/hilbert.go
```diff
@@ -4,6 +4,50 @@
...
+import (
+	"errors"
+	"fmt"
+)
+
+var ErrUnderflow = errors.New("order cannot be less than 1")
+var ErrOverflow = errors.New("overflow int")
+
+// NewHilbert2D constructs a [Hilbert2D] of the given order. NewHilbert2D
+// returns [ErrOverflow] if the order would cause Len and Pos to overflow.
```
kortschak, line 16: "returns **[ErrOverflow]**" is not quite true.

spatial/curve/hilbert.go
```diff
@@ -4,6 +4,50 @@
...
+	// The order must be greater than zero
+	if ord < 1 {
+		return v, ErrUnderflow
+	}
+
+	// The product of the order and number of dimensions must not exceed or
+	// equal the size of an int
+	if ord*dim >= intSize {
+		return v, fmt.Errorf("a %d-dimensional, %d-order Hilbert curve will %w", dim, ord, ErrOverflow)
+	}
```
kortschak, line 46: ISTM that this (and the test above) is small enough to include inline in the constructors and avoid the use of this generic function.

spatial/curve/hilbert.go
firelizzard18: I added "(wrapped)"

spatial/curve/hilbert.go
firelizzard18: Done

kortschak:
The CI issues have been fixed in master. Please rebase onto master to pick those changes up.

firelizzard18:
Rebased

spatial/curve/hilbert.go
```diff
@@ -0,0 +1,332 @@
...
+package curve
+
+import (
+	"errors"
+	"fmt"
+)
+
+// ErrUnderflow is returned by Hilbert curve constructors when the power is less
+// than 1.
+var ErrUnderflow = errors.New("order cannot be less than 1")
```
kortschak, line 14:
```suggestion
var ErrUnderflow = errors.New("order is less than 1")
```

spatial/curve/hilbert.go
```diff
@@ -0,0 +1,332 @@
...
+	return v, nil
+}
+
+// NewHilbert3D constructs a [Hilbert3D] of the given order. NewHilbert3D
+// returns [ErrOverflow] (wrapped) if the order would cause Len and Pos to
+// overflow.
+func NewHilbert3D(order int) (Hilbert3D, error) {
+	v := Hilbert3D{Order: order}
+
+	// The order must be greater than zero
```
kortschak, line 49:
```suggestion
	// The order must be greater than zero.
```
(final full-stop on internal comments as well - also below)

spatial/curve/hilbert.go
```diff
@@ -0,0 +1,332 @@
...
+	// equal the size of an int
+	if order*4 >= intSize {
+		return v, fmt.Errorf("a 4-dimensional, %d-order Hilbert curve will %w", order, ErrOverflow)
+	}
+
+	return v, nil
+}
+
+// Hilbert2D is a 2-dimensional Hilbert curve.
+type Hilbert2D struct{ Order int }
```
kortschak, line 84: I think probably `Order` should be unexported.

spatial/curve/hilbert.go
```diff
@@ -0,0 +1,332 @@
...
+	var d int
+	N := 4
+	for n := h.Order - 1; n >= 0; n-- {
+		var e int
+		for i := N - 1; i >= 0; i-- {
+			v := v[i] >> n & 1
+			e = e<<1 | (e^v)&1
+		}
+
+		d += e << (N * n)
```
kortschak, line 272:
```suggestion
	const dims = 4
	for n := h.Order - 1; n >= 0; n-- {
		var e int
		for i := dims - 1; i >= 0; i-- {
			v := v[i] >> n & 1
			e = e<<1 | (e^v)&1
		}

		d += e << (dims * n)
```

spatial/curve/hilbert.go
```diff
@@ -0,0 +1,332 @@
...
+// Pos will overflow if order is ≥ 11 on architectures where [int] is 32-bits,
+// or ≥ 21 on architectures where [int] is 64-bits.
+func (h Hilbert3D) Pos(v []int) int {
+	var d int
+	for n := h.Order - 1; n >= 0; n-- {
+		rx := (v[0] >> n) & 1
+		ry := (v[1] >> n) & 1
+		rz := (v[2] >> n) & 1
+		rd := rz<<2 | (rz^ry)<<1 | (rz ^ ry ^ rx)
+		d += rd << (3 * n)
```
kortschak, line 188: Use a const for 3?

spatial/curve/hilbert.go
```diff
@@ -0,0 +1,332 @@
...
+//
+// Pos will overflow if order is ≥ 16 on architectures where [int] is 32-bits,
+// or ≥ 32 on architectures where [int] is 64-bits.
+func (h Hilbert2D) Pos(v []int) int {
+	var d int
+	for n := h.Order - 1; n >= 0; n-- {
+		rx := (v[0] >> n) & 1
+		ry := (v[1] >> n) & 1
+		rd := ry<<1 | (ry ^ rx)
+		d += rd << (2 * n)
```
kortschak, line 117: Use a const for 2?

spatial/curve/hilbert.go
```diff
@@ -0,0 +1,332 @@
...
+// power would cause Len and Pos to overflow.
+var ErrOverflow = errors.New("overflow int")
+
+// The size of an int. Taken from src/math/const.go.
+const intSize = 32 << (^uint(0) >> 63) // 32 or 64
+
+// NewHilbert2D constructs a [Hilbert2D] of the given order. NewHilbert2D
+// returns [ErrOverflow] (wrapped) if the order would cause Len and Pos to
+// overflow.
+func NewHilbert2D(order int) (Hilbert2D, error) {
```
kortschak, line 26: Can you move each of these to just below their corresponding type?

kortschak (commented):
Thanks. I think this is almost done. Thank you for bearing with us.

Can you send another PR with the commit message "A+C: add Ethan Reesor" adding yourself to AUTHORS and CONTRIBUTORS.

spatial/curve/hilbert.go
firelizzard18: :+1:

firelizzard18:
Done

kortschak (approved):

PR #1940, opened 2024-03-06
ci: update to go1.21

jbduncan:
This PR is a prerequisite for a few others I intend to raise that will better prepare `graph` and other packages for generics and other modernization, as described in https://github.com/gonum/gonum/issues/617.

Follow-on to https://github.com/gonum/gonum/pull/1934, which was my first attempt at this but was too large to properly benchmark.

PR #1949, opened 2024-03-06
floats: use more efficient "is sorted" routine

jbduncan:

Blocked by https://github.com/gonum/gonum/pull/1940.

### DIFF
--- a/floats/floats.go
+++ b/floats/floats.go
@@ -7,6 +7,7 @@ package floats
 import (
 	"errors"
 	"math"
+	"slices"
 	"sort"
 
 	"gonum.org/v1/gonum/floats/scalar"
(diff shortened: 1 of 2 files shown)


jbduncan:
Unexpectedly, there are no benchmarks that exercise `floats.Within` directly or indirectly, and the only other code change affects a test, so there's nothing to run here.

kortschak:

jbduncan:

Done.

kortschak (commented):
Any appeal to efficiency needs to be supported by benchmarks.

jbduncan:
@kortschak Thanks for challenging me on this. On re-reading the source code, I see that `sort.Float64sAreSorted` delegates to `slices.IsSorted`, so if there is any improvement, it would be so small that I don't think this PR is worth it anymore.

Closing.

PR #1954, opened 2024-03-06
unit: use more efficient sort routine

jbduncan:

Blocked by https://github.com/gonum/gonum/pull/1940.

### DIFF
--- a/unit/unit_test.go
+++ b/unit/unit_test.go
@@ -472,3 +472,44 @@ func BenchmarkMulMapLargeSame(b *testing.B) {
 		u2.Mul(u1)
 	}
 }
+
+func BenchmarkDimensionsString(b *testing.B) {
+	oneDim := Dimensions{CurrentDim: 1}
+	threeDims := Dimensions{
+		CurrentDim:           1,
+		LengthDim:            2,
+		LuminousIntensityDim: 3,
+	}
+	allDims := Dimensions{
+		CurrentDim:           1,
+		LengthDim:            2,
+		LuminousIntensityDim: 3,
+		MassDim:              4,
+		MoleDim:              5,
+		TemperatureDim:       6,
+		TimeDim:              7,
+		AngleDim:             8,
+	}
+
+	b.Run("oneDim", func(b *testing.B) {
+		for i := 0; i < b.N; i++ {
+			if oneDim.String() != "" {
+				b.FailNow()
+			}
+		}
+	})
+	b.Run("threeDims", func(b *testing.B) {
+		for i := 0; i < b.N; i++ {
+			if threeDims.String() != "" {
+				b.FailNow()
+			}
+		}
+	})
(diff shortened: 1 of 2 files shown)


kortschak:

jbduncan:

Done.

kortschak (commented):
Needs evidence.

jbduncan:
These benchmark results seem to prove that there are performance improvements:

```
goos: darwin
goarch: arm64
pkg: gonum.org/v1/gonum/unit
                             │ unit/benchmarks_dimensions_string_old.txt │ unit/benchmarks_dimensions_string_new.txt │
                             │                  sec/op                   │      sec/op        vs base                │
DimensionsString/oneDim-8                                    175.8n ± 0%         147.6n ± 0%  -16.07% (p=0.000 n=10)
DimensionsString/threeDims-8                                 488.7n ± 0%         455.1n ± 0%   -6.88% (p=0.000 n=10)
DimensionsString/allDims-8                                   1.457µ ± 0%         1.417µ ± 1%   -2.75% (p=0.000 n=10)
geomean                                                      500.3n              456.5n        -8.74%
```

jbduncan:
Hi team, this PR is ready, so PTAL.

unit/unit_test.go
```diff
@@ -472,3 +472,38 @@ func BenchmarkMulMapLargeSame(b *testing.B) {
...
+		MassDim:              4,
+		MoleDim:              5,
+		TemperatureDim:       6,
+		TimeDim:              7,
+		AngleDim:             8,
+	}
+
+	b.Run("oneDim", func(b *testing.B) {
+		for i := 0; i < b.N; i++ {
+			_ = oneDim.String()
```
kortschak, line 496: The assignment to `_` is a no-op here. It does not ensure that the `String` method call is not optimised away. So please remove it. If you are worried about the optimisation occurring, please add a cheap conditional (say `!= ""`) with a `b.FailNow()` in the fail case to elevate the value to an observable effect.

kortschak (commented):
Please separate the benchmark addition into a separate commit, prior to the logic change so that the perf impact can be confirm eternally.

unit/unit_test.go
jbduncan: Thanks for pointing this out. Actually I was already aware that this doesn't stop the method call from being optimised away, it was more of a habit to make it super clear that we're intentionally ignoring the result.

But it sounds sensible to use `b.FailNow()` anyway, so I've just force-pushed a change to apply this.

jbduncan:
> Please separate the benchmark addition into a separate commit, prior to the logic change so that the perf impact can be confirm eternally.

Alrighty! Done.

I'm ready for another review whenever you are.

kortschak (approved):
Thanks

jbduncan:
@kortschak yw! And thank you for being so patient with me, with all my PRs recently. :)

PR #1953, opened 2024-03-06
internal: generify and move ordered.BySliceValues to top-level place

jbduncan:
This PR introduces a modernization change that is not necessarily related to #617 but which was spun out from my attempts at generifying the leaves of `graph`.

Although it may not provide any value in of itself, it will be used by forthcoming PRs to remove duplicate code.

Benchmarks to follow.

Blocked by https://github.com/gonum/gonum/pull/1940.

PR #1955, opened 2024-03-06
spatial: use generic lexical sorting routine

jbduncan:

Blocked by https://github.com/gonum/gonum/pull/1940 and https://github.com/gonum/gonum/pull/1953.

### DIFF
--- a/internal/sorted/doc.go
+++ b/internal/sorted/doc.go
@@ -1,0 +1,6 @@
+// Copyright ©2024 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Package sorted provides common sorting functions.
+package sorted // import "gonum.org/v1/gonum/internal/ordered"
--- a/graph/internal/ordered/doc.go
+++ b/graph/internal/ordered/doc.go
@@ -2,5 +2,5 @@
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
-// Package ordered provides common sort ordering types.
+// Package ordered provides common sort ordering types for the graph package.
 package ordered // import "gonum.org/v1/gonum/graph/internal/ordered"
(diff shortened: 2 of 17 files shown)


jbduncan:
When checking out this PR and using IntelliJ's "Call Hierarchy" on `BySliceValues`, I see that only tests use it directly or indirectly. Furthermore, the [only additional commit](https://github.com/gonum/gonum/pull/1955/commits/9e82cf86492a8212c09d9e266d88d0536a7b204d) introduced in this PR over https://github.com/gonum/gonum/pull/1940 and https://github.com/gonum/gonum/pull/1953 affects just `TestDo` in `spatial/vptree/vptree_test.go`; #1940 only has GitHub Actions changes; and [I already checked](https://github.com/gonum/gonum/pull/1953#issuecomment-1986234415) that #1953 has no benchmarks to run. Therefore, this PR has no benchmarks to run.

jbduncan:
As discussed in https://github.com/gonum/gonum/pull/1956#issuecomment-1988326786, I'm more than happy to double check if new benchmarks are needed first.

kortschak:

jbduncan:

Done. I will pick up the latest changes again if https://github.com/gonum/gonum/pull/1953 is merged in.

internal/sorted/doc.go
```diff
@@ -0,0 +1,6 @@
+// Copyright ©2024 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Package sorted provides common sorting functions.
+package sorted // import "gonum.org/v1/gonum/internal/ordered"
```
kortschak, line 6: Why is the package name changed?

jbduncan:
As per https://github.com/gonum/gonum/pull/1953#issuecomment-2016622004, I'm closing this PR in favour of putting it in #1953.

internal/sorted/doc.go
jbduncan: See https://github.com/gonum/gonum/pull/1953#discussion_r1536703590

Issue #1904, opened 2023-10-02
all: replace min and max helpers with min/max builtins when go1.21 is lowest supported version


kortschak:
There are 18 instances of a `func min(a, b T) T` and 12 instances of a `func max(a, b T) T`. We should replace use of these with calls to the built-ins when they are available in all supported versions of Go.

PR #1965, opened 2024-04-06
all: replace min/max helpers with min/max builtins

Juneezee:
Closes https://github.com/gonum/gonum/issues/1904.

We support minimum Go 1.21 since v0.15.0 (https://github.com/gonum/gonum/pull/1935).

Please take a look.


kortschak (approved):
Thanks

PR #1970, opened 2024-04-17
mat: calculate Q lazily when calling QR.ToQ

### LINKED ISSUE #1968: mat: calling qr.Factorize leads to OOM for matrixes with many rows

After upgrading gonum to 0.15.0 calls to `(VecDense).SolveVec(a,b)` where `a.Dims() == (10k+, 3)`  leads to OOM.
The stack where OOM happens:
```
runtime.systemstack_switch()
        /nix/store/8yw3g52r95h7cv09lcrran92n212997b-go-1.22.0/share/go/src/runtime/asm_amd64.s:474 +0x8 fp=0xc000087230 sp=0xc000087220 pc=0x4788a8
runtime.(*mheap).alloc(0x3667b52000?, 0x1b33da9?, 0x80?)
        /nix/store/8yw3g52r95h7cv09lcrran92n212997b-go-1.22.0/share/go/src/runtime/mheap.go:958 +0x5b fp=0xc000087278 sp=0xc000087230 pc=0x42fc1b
runtime.(*mcache).allocLarge(0x100?, 0x3667b50b88, 0x1?)
        /nix/store/8yw3g52r95h7cv09lcrran92n212997b-go-1.22.0/share/go/src/runtime/mcache.go:234 +0x85 fp=0xc0000872c0 sp=0xc000087278 pc=0x41ce45
runtime.mallocgc(0x3667b50b88, 0xd67320, 0x1)
        /nix/store/8yw3g52r95h7cv09lcrran92n212997b-go-1.22.0/share/go/src/runtime/malloc.go:1165 +0x597 fp=0xc000087348 sp=0xc0000872c0 pc=0x4141d7
runtime.makeslice(0xc0000124b0?, 0x18?, 0xc0000874b0?)
        /nix/store/8yw3g52r95h7cv09lcrran92n212997b-go-1.22.0/share/go/src/runtime/slice.go:107 +0x49 fp=0xc000087370 sp=0xc000087348 pc=0x459989
gonum.org/v1/gonum/mat.NewDense(...)
        /home/ericwenn/go/pkg/mod/gonum.org/v1/gonum@v0.15.1-0.20240413203616-1b7d9ca04ac9/mat/dense.go:60
gonum.org/v1/gonum/mat.(*QR).updateQ(0xc000087630)
        /home/ericwenn/go/pkg/mod/gonum.org/v1/gonum@v0.15.1-0.20240413203616-1b7d9ca04ac9/mat/qr.go:107 +0x78 fp=0xc0000874c0 sp=0xc000087370 pc=0xb5fd38
gonum.org/v1/gonum/mat.(*QR).factorize(0xc000087630, {0x1070598, 0xc0004c6540}, 0x49)
        /home/ericwenn/go/pkg/mod/gonum.org/v1/gonum@v0.15.1-0.20240413203616-1b7d9ca04ac9/mat/qr.go:101 +0x316 fp=0xc0000875e0 sp=0xc0000874c0 pc=0xb5fc36
gonum.org/v1/gonum/mat.(*QR).Factorize(...)
        /home/ericwenn/go/pkg/mod/gonum.org/v1/gonum@v0.15.1-0.20240413203616-1b7d9ca04ac9/mat/qr.go:82
gonum.org/v1/gonum/mat.(*Dense).Solve(0xc0004c6700, {0x1070598, 0xc0004c6540}, {0x1070598, 0xc0004c6740})
        /home/ericwenn/go/pkg/mod/gonum.org/v1/gonum@v0.15.1-0.20240413203616-1b7d9ca04ac9/mat/solve.go:65 +0x245 fp=0xc000087700 sp=0xc0000875e0 pc=0xb63545
gonum.org/v1/gonum/mat.(*VecDense).SolveVec(0xc000110360, {0x1070598, 0xc0004c6540}, {0x10735c0, 0xc0001103c0})
        /home/ericwenn/go/pkg/mod/gonum.org/v1/gonum@v0.15.1-0.20240413203616-1b7d9ca04ac9/mat/solve.go:118 +0x6fa fp=0xc000087930 sp=0xc000087700 pc=0xb63e9a
``` 
I've traced this down to changes in `mat.QR.updateQ` [here](https://github.com/gonum/gonum/commit/45b74210d616dd61bb893f6e8e4af13879b1d6bb#diff-5bf82974a09725d659ece49d56eff0b81eafc1a476a618949ec743544e329aaeR104-R117).

What `updateQ` does is allocate a new `Dense` with dimensions `MxM`, which for our value on M would mean allocating 100M cell matrix. Based on my understanding of the code, the resulting matrix should have dimensions (`MxN`).

```
$ go version
go version go1.22.0 linux/amd64

$ grep gonum.org/v1/gonum go.sum
gonum.org/v1/gonum v0.15.0 h1:2lYxjRbTYyxkJxlhC+LvJIx3SsANPdRybu1tGj9/OrQ=
gonum.org/v1/gonum v0.15.0/go.mod h1:xzZVBJBtS+Mz4q0Yl2LJTk+OxOg4jiXZ7qBoM0uISGo=
```

Yes

kortschak:
When a matrix is very tall, calculating Q will currently allocate a large Q at the end of the factorisation, even if it is not going to be used. The eager calculation was intended to prevent repeated re-calculation of Q when it is used. So move the Q calculation to ToQ, but make it conditional on the stored Q value being empty, and empty Q at the end of the factorisation.

Please take a look.

Fixes #1968

### DIFF
--- a/mat/qr.go
+++ b/mat/qr.go
@@ -98,23 +98,9 @@ func (qr *QR) factorize(a Matrix, norm lapack.MatrixNorm) {
 	lapack64.Geqrf(qr.qr.mat, qr.tau, work, len(work))
 	putFloat64s(work)
 	qr.updateCond(norm)
-	qr.updateQ()
-}
-
-func (qr *QR) updateQ() {
-	m, _ := qr.Dims()
-	if qr.q == nil {
-		qr.q = NewDense(m, m, nil)
-	} else {
-		qr.q.reuseAsNonZeroed(m, m)
+	if qr.q != nil {
+		qr.q.Reset()
 	}
-	// Construct Q from the elementary reflectors.
-	qr.q.Copy(qr.qr)
-	work := []float64{0}
-	lapack64.Orgqr(qr.q.mat, qr.tau, work, -1)
-	work = getFloat64s(int(work[0]), false)
-	lapack64.Orgqr(qr.q.mat, qr.tau, work, len(work))
-	putFloat64s(work)
 }
 
 // isValid returns whether the receiver contains a factorization.
@@ -192,7 +178,26 @@ func (qr *QR) QTo(dst *Dense) {
 			panic(ErrShape)
 		}
 	}
+	if qr.q == nil || qr.q.IsEmpty() {
+		qr.updateQ()
+	}
 	dst.Copy(qr.q)
+}
+
+func (qr *QR) updateQ() {
+	m, _ := qr.Dims()
+	if qr.q == nil {
+		qr.q = NewDense(m, m, nil)
+	} else {
+		qr.q.reuseAsNonZeroed(m, m)
+	}
+	// Construct Q from the elementary reflectors.
+	qr.q.Copy(qr.qr)
+	work := []float64{0}
+	lapack64.Orgqr(qr.q.mat, qr.tau, work, -1)
+	work = getFloat64s(int(work[0]), false)
+	lapack64.Orgqr(qr.q.mat, qr.tau, work, len(work))
+	putFloat64s(work)
 }
 
 // SolveTo finds a minimum-norm solution to a system of linear equations defined
--- a/mat/qr_test.go
+++ b/mat/qr_test.go
@@ -18,9 +18,11 @@ func TestQR(t *testing.T) {
 	rnd := rand.New(rand.NewSource(1))
 	for _, test := range []struct {
 		m, n int
+		big  bool
 	}{
-		{5, 5},
-		{10, 5},
+		{m: 5, n: 5},
+		{m: 10, n: 5},
+		{m: 1e5, n: 3, big: true}, // Test that very tall matrices do not OoM.
 	} {
 		m := test.m
 		n := test.n
@@ -35,7 +37,15 @@ func TestQR(t *testing.T) {
 
 		var qr QR
 		qr.Factorize(a)
-		var q, r Dense
+
+		var r Dense
+		qr.RTo(&r)
+		if test.big {
+			// We cannot proceed past here for big matrices.
+			continue
+		}
+
+		var q Dense
 		qr.QTo(&q)
 
 		if !isOrthonormal(&q, 1e-10) {
@@ -48,8 +58,6 @@ func TestQR(t *testing.T) {
 		if !EqualApprox(a.T(), qr.T(), 1e-14) {
 			t.Errorf("m=%d,n=%d: Aᵀ and (QR)ᵀ are not equal", m, n)
 		}
-
-		qr.RTo(&r)
 
 		var got Dense
 		got.Mul(&q, &r)


tvkn (approved):
Looks good to me!

Do you want to add a dedicated test for the issue it fixes?

kortschak:
> Do you want to add a dedicated test for the issue it fixes?

It's in the table.

mat/qr.go
```diff
@@ -98,23 +98,9 @@ func (qr *QR) factorize(a Matrix, norm lapack.MatrixNorm) {
 	lapack64.Geqrf(qr.qr.mat, qr.tau, work, len(work))
 	putFloat64s(work)
 	qr.updateCond(norm)
-	qr.updateQ()
```
vladimir-ch, line 101: I'm afraid this doesn't work because `q` is currently needed in the `At` method. The test passes just because `QTo` is called in the test before `EqualApprox` (which calls `At`).

We could reconstruct the necessary row of Q in each call to `At` which is terrible but hopefully nobody uses `QR` as `Matrix` except in the call to `Solve`?

kortschak: The other alternative is to lazily calculate Q for `At` the same way it is for `ToQ`, with a warning that it may cause OoM. I think your approach is probably better. A warning in the docs that it will be extremely inefficient should be enough.

tvkn: We could lazily compute `q` in the `At()` method if `q` is nil or empty, in the same way it is done in the `Qto`.

But I guess I'm also a bit confused as to why we need `q` at all in `At()` in the first place. When we factorize a matrix `a`, `qr` stores a copy of that matrix. Why not return directly `qr.At()`?

Isn't https://github.com/gonum/gonum/blame/2ad11cabb395b96efc5b67fa1b64480762d9e703/mat/qr.go#L46 also not faulty?
I'm expected we should return Q*R at element(i,j), but instead it seems we're returning Q*A at element (i,j), with A = Q*R, the matrix we factorize.

vladimir-ch: > The other alternative is to lazily calculate Q for At the same way it is for ToQ.

Agreed. If one element is needed, then most likely all are.

> qr stores a copy of that matrix. Why not return directly qr.At()?

That copy is overwritten by `lapack64.Geqrf` which efficiently stores both Q and R in the same matrix.

vladimir-ch: Another alternative is to indeed store `a` in `Factorize` and compute Q in `QTo` without storing it.

kortschak: I think we would need to copy `a` in this situation.

vladimir-ch: In the end I think it's best to lazily compute Q in At. QR.At should never be called in practice.

tvkn: Any chance this is called inadvertently by using QR as a matrix input to another function (since it now fulfills the `Matrix` interface), e.g. `Trace()`.

ericwenn:
Hey @kortschak, is there anything I can help with to push this PR to the finish line?

kortschak:
Can you increase the number of hours in the day?

mat/qr.go
kortschak: I'm thinking that @tvkn's suggestion of conditionally computing element of QR for `At` via the relevant row of Q if Q is not already calculated, otherwise falling through to the code that is there now. 
```
diff --git a/mat/qr.go b/mat/qr.go
index d4787bc3..7caa2f20 100644
--- a/mat/qr.go
+++ b/mat/qr.go
@@ -41,6 +41,11 @@ func (qr *QR) At(i, j int) float64 {
 		panic(ErrColAccess)
 	}
 
+	if qr.q == nil || qr.q.IsEmpty() {
+		// Calculate Q's row i;
+		// Return Q[i,]×R[,j].
+	}
+
 	var val float64
 	for k := 0; k <= j; k++ {
 		val += qr.q.at(i, k) * qr.qr.at(k, j)
```
This is safer than just lazily calculating Q which has the risk of someone thinking the code is fine until a tall matrix is used down the track. At worst with this approach there will be a perf hit, but not a crash.

I don't have time to do this, so if someone wants to pick this up. I'm happy for that to happen.

tvkn:
@kortschak @vladimir-ch, I took a stab at it. Please look at PR #1988.

PR #1980, opened 2024-06-11
stat/distuv: Correct Gamma Mode() and LogProb(0)/Prob(0) for alpha <= 1

argusdusty:
Please take a look.

The mode of the Gamma distribution should be 0 for alpha < 1, not NaN. There's a singularity in the pdf, but it's still a valid mode.

Similarly, Prob(0) should be +Inf for alpha<1, and beta for alpha=1

### DIFF
--- a/stat/distuv/gamma.go
+++ b/stat/distuv/gamma.go
@@ -47,12 +47,15 @@ func (g Gamma) ExKurtosis() float64 {
 // LogProb computes the natural logarithm of the value of the probability
 // density function at x.
 func (g Gamma) LogProb(x float64) float64 {
-	if x <= 0 {
+	if x < 0 {
 		return math.Inf(-1)
 	}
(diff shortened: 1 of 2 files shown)


stat/distuv/gamma.go
```diff
@@ -63,11 +66,11 @@ func (g Gamma) Mean() float64 {
 
 // Mode returns the mode of the normal distribution.
```
kortschak, line 67: I'm guessing this ("normal") is a result of copy/paste in the original.

kortschak (approved):
LGTM

Do you want to fix up the doc comment while you are here?

kortschak (commented):
nit: Do you mind revising the history so that the the "Correct" is lower-cased?

kortschak (approved):
Thanks

PR #1984, opened 2024-06-28
graph/topo: fix cycle canonicalistion

kortschak:
The cycle was being canonicalised retaining the duplicated bridge node at the end of the slice. Fix this by removing the node, performing the cicular permutation as described, and then adding the new final node.

Please take a look.


kortschak:
Previous me had this under control.

Issue #1992, opened 2024-09-02
[gonum/optimize/convex/lp] Feature Request: Integer Programming


Aurora2500:
I'm proposing to add integer programming to gonum. 

### How does Gonum not allow you to achieve your goal?
[Integer programming](https://en.wikipedia.org/wiki/Integer_programming) is a specific variant of linear programming in which some or all variables are constrained to be whole integer numbers.

Currently the `gonum/optimize/convex/lp` package only features a `Simplex` function, which is only able to optimize for real numbered variables.

It is missing an integer programming optimizer, and these are complex enough to not expect a user of the library to implement for themselves. Thus I'm requesting to add such functionality to the library itself.

### Is this feature absent from the current master?
Yes

### Are you able to help contribute the feature?
I'd be willing to try & implement some parts of it.

Aurora2500:
I've implemented a basic branch & bound algorithm in my own project I'm working on I could add to the library if it's accepted, it might not be the most efficient solution but it's better than nothing.

PR #1994, opened 2024-09-03
[optimize/convex/lp] Implemented basic branch and bound

Aurora2500:
I implemented branch & bound using the existing simplex functions to allow for integer programming, closing #1992. 

Please take a look.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -19,6 +19,7 @@ Alexander Egurnov <alexander.egurnov@gmail.com>
 Andrei Blinnikov <goofinator@mail.ru>
 Andrew Brampton <brampton@gmail.com>
 antichris <chris@u-d13.com>
+Aurora Zuoris <programming.aurora@gmail.com>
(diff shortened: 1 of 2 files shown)


kutlay:
The current simplex implementation in the repository returns 3 variables, you are using a simplex implementation that returns 4 variable. Is that an implementation that you haven't shared yet?

vladimir-ch:
@Aurora2500 Sorry the PR has been sitting here without a response for so long, I don't have much bandwidth lately for the project. If you're still interested in the PR, I'll try to find time to take a look but could you please first add tests that cover the newly added code?

Aurora2500:
Sure! Bit busy these days but I'll try to get one done by the end of the week.

Aurora2500:
Added a test that checks for a known solution and a test that fails specifically for BNB.

There are a bunch of other ways it can fail but I figured they're specific to Simplex, as that's what I'm using under the hood, so if there's an issue with those then the simplex tests would catch that & my code shouldn't be concerened with it imo.

Issue #1865, opened 2023-06-03
Proposal: multivariate & multivariable analysis


petric3:
I'm mechanical engineer migrated from python about 3 years ago. I'm really happy with golang, have speeded up my analyses 80-100x. I really like to promote golang in our community and most of the time I'm asked how's coverage with the libraries for 
1. regression analysis 
2. polynomial fitting and
3. nonlinear analysis. 

One thing myself is really missing is multivariate & multivariable (MISO & MIMO) linear regression analysis. I believe if such packages would be implemented, golang would get much more attention and community would rise quickly. 

Thank you for your efforts and keep up with the good work.

smjure:
+1

xslasd:
+1

PR #1998, opened 2024-10-28
statmat: added multi linear and lasso regression

aouyang1:
Hello! First time contributor here. I've been working on some multi-linear regression work lately and wanted to see if some of that could also be integrated here. This only partially solves some of the features requested in #1865 (MISO) but can extend it to more if we want to in this PR. I tried my best to follow similar conventions that I saw in the existing modules, but let me know what I can change to make it more similar.

The 2 new regression structs are OLSRegression and LassoRegression.
* OLSRegressions uses a QR decomposition to compute ordinary least squares with multiple features. I'm sure there's more efficient algorithms here.
* LassoRegression uses coordinate descent to find the optimal weights.

Each implement the following:
```
Fit(x, y mat.Matrix) error
Predict(x mat.Matrix) ([]float64, error)
Score(x, y mat.Matrix) (float64, error)
Intercept() float64
Coef() []float64
```
Inspired by the Python sklearn interface

### DIFF
--- a/stat/statmat.go
+++ b/stat/statmat.go
@@ -5,6 +5,8 @@
 package stat
 
 import (
+	"errors"
+	"fmt"
 	"math"
 
 	"gonum.org/v1/gonum/floats"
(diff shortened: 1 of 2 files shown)


kortschak (commented):
I've taken a brief look through this. There is more API here that I think is necessary. Also there are error returns that do not conform to the approach that we use in Gonum packages; we use error returns for error conditions that the user could not know before calling a function, but panics for cases where the calling parameters do not conform to the documented invariants for the call. Please match this.

I'll take a deeper look in the next week or so.

aouyang1:
> I've taken a brief look through this. There is more API here that I think is necessary. Also there are error returns that do not conform to the approach that we use in Gonum packages; we use error returns for error conditions that the user could not know before calling a function, but panics for cases where the calling parameters do not conform to the documented invariants for the call. Please match this.
> 
> I'll take a deeper look in the next week or so.

Thanks for explaining the differences! Will get the errors addressed and converted.

kortschak (commented):
The API is too complex here. I'd estimate that we could make this somewhere between 1/3 and 2/3 of the code that's here by removing the

aouyang1:
> The API is too complex here. I'd estimate that we could make this somewhere between 1/3 and 2/3 of the code that's here by removing the extraneous code.

Sounds good! Will take a crack at simplifying it and model something close to the LinearRegression method.

stat/statmat.go
```diff
@@ -140,3 +141,454 @@ func Mahalanobis(x, y mat.Vector, chol *mat.Cholesky) float64 {
...
+}
+
+// SoftThreshold returns 0.0 if the value is less than or equal to the gamma input
+func SoftThreshold(x, gamma float64) float64 {
+	res := math.Max(0, math.Abs(x)-gamma)
+	if math.Signbit(x) {
+		return -res
+	}
+	return res
+}
```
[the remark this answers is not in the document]
aouyang1: good point. will make it private

stat/statmat.go
```diff
@@ -140,3 +141,454 @@ func Mahalanobis(x, y mat.Vector, chol *mat.Cholesky) float64 {
...
+
+	return o
+}
+
+// NewDefaultOLSOptions returns a default set of OLS Regression options
+func NewDefaultOLSOptions() *OLSOptions {
+	return &OLSOptions{
+		FitIntercept: true,
+	}
+}
```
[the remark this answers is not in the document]
aouyang1: Got rid of a lot of the boiler plate. Let me know if this is more of what you were thinking. Took a look at the LinearRegression and the PrincipalComponent methods.

stat/statmat_test.go
```diff
@@ -303,6 +306,235 @@ func TestMahalanobis(t *testing.T) {
...
+	testData := map[string]struct {
+		x         *mat.Dense
+		y         *mat.Dense
+		model     LassoModel
+		err       error
+		tol       float64
+		intercept float64
+		coef      []float64
+	}{
+		"invalid lambda": {
```
kortschak, line 336:
```suggestion
		"invalid_lambda": {
```
(similar throughout; it simplifies finding cases)

stat/statmat_test.go
```diff
@@ -303,6 +306,235 @@ func TestMahalanobis(t *testing.T) {
...
+			panic("bad test: ragged input")
+		}
+		d = append(d, row...)
+	}
+	return r, c, d
+}
+
+func TestLassoRegression(t *testing.T) {
+	// y = 2 + 3*x0 + 4*x1
+	testData := map[string]struct {
```
kortschak, line 327: Please don't do this with a map. Use a struct with a name field. Also put the test cases in a global var, `lassoRegressionTests`, above the `TestLassoRegression` func decl.

stat/statmat_test.go
```diff
@@ -287,7 +289,8 @@ func TestMahalanobis(t *testing.T) {
 				[]float64{
 					0.8, 0.3, 0.1,
 					0.3, 0.7, -0.1,
-					0.1, -0.1, 7}),
+					0.1, -0.1, 7,
+				}),
```
kortschak, line 293: Let's leave these formatting changes alone. Please revert this and the changes above in this file.

stat/statmat_test.go
```diff
@@ -303,6 +306,235 @@ func TestMahalanobis(t *testing.T) {
...
+func TestOLSRegression(t *testing.T) {
+	// y = 2 + 3*x0 + 4*x1
+	testData := map[string]struct {
+		x         *mat.Dense
+		y         *mat.Dense
+		model     OLSModel
+		tol       float64
+		intercept float64
+		coef      []float64
+	}{
```
kortschak, line 471: Similar here.

stat/statmat_test.go
```diff
@@ -321,6 +553,7 @@ func benchmarkCovarianceMatrix(b *testing.B, m mat.Matrix) {
 		CovarianceMatrix(&res, m, nil)
 	}
 }
+
```
kortschak, line 556: Please revert these formatting changes.

stat/statmat_test.go
```diff
@@ -478,3 +718,65 @@ func BenchmarkCorrToCov(b *testing.B) {
 		corrToCov(cc, sigma)
 	}
 }
+
+func BenchmarkLassoRegression(b *testing.B) {
```
kortschak, line 722: If we're adding benchmarks, they will need to have a timer reset after the set-up, but also the set up doesn't need to do all the work that it is; it should work directly into `x`, and the matrix constructions should happen outside the benchmark loop.

stat/statmat_test.go
```diff
@@ -478,3 +718,65 @@ func BenchmarkCorrToCov(b *testing.B) {
...
+	for i := 0; i < b.N; i++ {
+		m, n, d := flatten(data)
+		x := mat.NewDense(m, n, d)
+		y := mat.NewDense(nObs, 1, data2)
+		model := &LassoModel{FitIntercept: false}
+		model.Fit(x, y)
+	}
+}
+
+func BenchmarkOLSRegression(b *testing.B) {
```
kortschak, line 752: Same here.

kortschak (commented):
This looks much more manageable.

Initial review only.

tiloburgey:
Hi, thank you guys for the work you've put in. What's the status of this PR, because I'd really like to see this feature in the main branch.

aouyang1:
> Hi, thank you guys for the work you've put in. What's the status of this PR, because I'd really like to see this feature in the main branch.

It's in early review right now. I should have this PR up to date with the changes needed described in the comments.

PR #1999, opened 2024-10-30
floats: AddScaled: Use f64.AxpyUnitary instead of f64.AxpyUnitaryTo

Jille:


kortschak:
This is reasonable, but it would be nice to have some background. In future, at a minimum, please leave the template in place.

Jille:
I had removed the template (after checking I conformed to it, of course) because it was commented.

Background: I was reading the source code and thought a direct call would be better.

kortschak (approved):

kortschak:
Please send another PR adding yourself to AUTHORS and CONTRIBUTORS with the commit message "A+C: add Jille Timmermans".

PR #2002, opened 2024-11-15
dsp/transform: new package and initial Hilbert transform

robertkleffner:
First time contributor to this project. In a project I'm working on using some DSP-related code in Go, I needed to demodulate a real-valued signal to get its IQ components. The Hilbert transform is a useful way to get this information, and I was already using Gonum's FFTs elsewhere, so the implementation was straightforward and is described in numerous places on the internet.

The signal envelope is also a useful feature, so I included a method to compute that as well.

There is one TODO, regarding the necessity of normalization for the Analytic signal. So far, the results I get for both analytic signal and envelope are visually accurate, so I'm leaning towards necessary.

Justification for new `transform` folder: there are several other related signal transforms that could be eventually included here, e.g. Hilbert-Huang Transform. Originally, I had these files in the `fourier` folder, but it didn't feel right even though the implementation relies heavily on `CmplxFFT`. Open to other folder names or locations.


dsp/transform/hilbert.go
```diff
@@ -0,0 +1,89 @@
...
+
+	"gonum.org/v1/gonum/dsp/fourier"
+)
+
+// Implements an approximate Hilbert transform that allows the user to determine an
+// approximate analytical signal of a real signal, and determine the real 'envelope'
+// of a signal.
+//
+// The underlying implementation uses a DFT and IDFT transform, so usual methods for
+// speeding up DFT/IDFT are likely to apply here as well.
```
kortschak, line 18:
```suggestion
// Hilbert implements an approximate Hilbert transform that allows calculation
// of an approximate analytical signal of a real signal, and determine the
// real envelope of a signal.
//
// The underlying implementation uses a discrete Fourier transform and inverse
// and discrete Fourier transform, so usual methods for speeding up these
// transforms are likely to apply here as well.
```

Is the second paragraph directed at users? If so, a link to a relevant article would be nice. If they are directed at us as maintainers, the comment should not be visible in the godoc.

dsp/transform/hilbert.go
```diff
@@ -0,0 +1,89 @@
...
+
+// Return a new Hilbert transformer for signals of size N.
+func NewHilbert(n int) *Hilbert {
+	return &Hilbert{
+		fft:  fourier.NewCmplxFFT(n),
+		work: make([]complex128, n),
+	}
+}
+
+// Retrieve the length N of signals that are valid input for this Hilbert transform.
```
kortschak, line 32:
```suggestion
// Len returns the length of signals that are valid input for this Hilbert transform.
```

dsp/transform/hilbert.go
```diff
@@ -0,0 +1,89 @@
...
+func (h *Hilbert) Len() int {
+	return len(h.work)
+}
+
+// Compute the analytical signal of a real signal, and stores the result in the dst slice,
+// returning it.
+// If the dst slice is nil, a new slice will be created and returned. The dst slice
+// must be the same length as the input signal.
+func (h *Hilbert) AnalyticSignal(signal []float64, dst []complex128) []complex128 {
```
kortschak, line 41:
```suggestion
func (h *Hilbert) AnalyticSignal(dst []complex128, signal []float64) []complex128 {
```
`dst` is conventionally the first parameter.

dsp/transform/hilbert.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+
+	// forward fft of signal
+	coeff := h.fft.Coefficients(dst, h.work)
+	for i := range h.work {
+		h.work[i] = 0
+	}
+
+	// multiply positive frequencies by 2, zero out negative frequencies
+	// however, leave dc unchanged (and nyquist when n%2 == 0)
```
kortschak, line 57:
```suggestion
	// Multiply positive frequencies by 2, zero out negative frequencies.
	// However, leave dc unchanged (and nyquist when n%2 == 0).
```

dsp/transform/hilbert.go
```diff
@@ -0,0 +1,89 @@
...
+	h.work[0] = coeff[0]
+	for i, d := range coeff[1 : len(coeff)/2+1] {
+		h.work[i+1] = d * 2
+	}
+	if len(coeff)%2 == 0 {
+		h.work[len(coeff)/2] = coeff[len(coeff)/2]
+	}
+
+	// normalize the results for now
+	// TODO: is this important for analytic signal or envelope?
```
kortschak, line 67:
```suggestion
	// Normalize the results for now.
	// TODO: Is this important for analytic signal or envelope?
```
What is the TODO to be done?

dsp/transform/hilbert.go
```diff
@@ -0,0 +1,89 @@
...
+func (h *Hilbert) AnalyticSignal(signal []float64, dst []complex128) []complex128 {
+	if dst == nil {
+		dst = make([]complex128, len(signal))
+	}
+
+	for i, v := range signal {
+		h.work[i] = complex(v, 0)
+	}
+
+	// forward fft of signal
```
kortschak, line 50:
```suggestion
	// Forward FFT of the signal.
```

dsp/transform/hilbert.go
```diff
@@ -0,0 +1,89 @@
...
+
+// Retrieve the length N of signals that are valid input for this Hilbert transform.
+func (h *Hilbert) Len() int {
+	return len(h.work)
+}
+
+// Compute the analytical signal of a real signal, and stores the result in the dst slice,
+// returning it.
+// If the dst slice is nil, a new slice will be created and returned. The dst slice
+// must be the same length as the input signal.
```
kortschak, line 40:
```suggestion
// AnalyticSignal computes the analytical signal of a real signal, and stores
// the result in the dst slice, returning it.
//
// If the dst slice is nil, a new slice will be created and returned. The dst slice
// must be the same length as the input signal.
```

dsp/transform/hilbert.go
```diff
@@ -0,0 +1,89 @@
...
+	if dst == nil {
+		dst = make([]float64, len(signal))
+	}
+
+	analytic := h.AnalyticSignal(signal, nil)
+	for i, a := range analytic {
+		dst[i] = cmplx.Abs(a)
+	}
+	return dst
+}
```
kortschak, line 89: I'm not convinced this is needed. It's just a range loop.

dsp/transform/hilbert_test.go
```diff
@@ -0,0 +1,41 @@
...
+)
+
+func TestHilbertAnalytic(t *testing.T) {
+	testCases := []struct {
+		input []float64
+	}{
+		{[]float64{0, 0, 0, 0}},
+		{[]float64{1, 2, 3, 4}},
+		{[]float64{1, 2, 3, 4, 5}},
+	}
```
kortschak, line 21: Please hoist this out of the function to be a `var hilbertAnalyticSignalTests []struct{ name string; input ...`.

dsp/transform/hilbert_test.go
```diff
@@ -0,0 +1,41 @@
...
+
+	for i, tc := range testCases {
+		t.Run(fmt.Sprintf("tc%d", i), func(t *testing.T) {
```
kortschak, line 24:
```suggestion
	for _, test := range hilbertAnalyticSignalTests {
		t.Run(test.name, func(t *testing.T) {
```
and `s/tc/test/g` elsewhere.

dsp/transform/hilbert_test.go
```diff
@@ -0,0 +1,41 @@
...
+			transformer := NewHilbert(len(tc.input))
+			if transformer.Len() != len(tc.input) {
+				t.Errorf("expected Hilbert transform length to be %d, got %d", len(tc.input), transformer.Len())
+			}
```
kortschak, line 28:
```suggestion
			h := NewHilbert(len(tc.input))
			if h.Len() != len(test.input) {
				t.Errorf("unexpected Hilbert transform length: got:%d want:%d", h.Len(), len(test.input))
			}
```

dsp/transform/hilbert_test.go
```diff
@@ -0,0 +1,41 @@
...
+
+			dst := make([]complex128, len(tc.input))
+			analytic := transformer.AnalyticSignal(tc.input, dst)
+			result := make([]float64, len(tc.input))
+			for i, c := range analytic {
+				result[i] = math.Round(real(c))
```
kortschak, line 34: This makes me queasy. Why is it being done?

dsp/transform/hilbert_test.go
```diff
@@ -0,0 +1,41 @@
...
+			}
+			if !slices.Equal(tc.input, result) {
+				t.Errorf("expected Hilbert transform result %v, got %v", tc.input, result)
+			}
```
kortschak, line 38: `cmplxs.EqualApprox`?

kortschak (commented):
Thanks for the contribution.

dsp/transform/hilbert_test.go
robertkleffner: Originally wrote this test outside the gonum library; it is now using `cmplxs.EqualApprox`.

dsp/transform/hilbert.go
robertkleffner: I think normalized is how most people will expect the analytic signal; removing TODO

dsp/transform/hilbert_test.go
robertkleffner: This feedback is confusing to me, as the FFT test cases are written inside the function. With the update I made for readability, is this still needed?

dsp/transform/hilbert.go
robertkleffner: My only rebuttal is that some users would only be interested in the signal envelope, and possibly have only a fuzzy idea how it relates to the analytic signal (or even that it does, as I did when I first went looking for this). They would then have to look it up in order to recreate this loop, hoping that `AnalyticSignal` leaves the results in a state that all one needs to do is `Abs` each value. (For example, Gonum's FFT does not normalize its results, which is documented but surprising). With this method, it's at least searchable that Gonum can easily be used to compute a signal envelope regardless of how much expertise one has with DSP.

As an alternative, though, perhaps I could add to the doc comment of `AnalyticSignal`: "The envelope of the signal can be straightforwardly computed from the results using `cmplx.Abs` on each element".

dsp/transform/hilbert.go
robertkleffner: Made this comment nicer, and removed the second paragraph. I added another comment to the `NewHilbert` method noting a performance tip. My source for that performance tip is a comment [in the internal/fftpack source](https://github.com/gonum/gonum/blob/1ca563a018b641e805317f1ac9ae0d37b32d162c/dsp/fourier/internal/fftpack/cfft.go#L126)

dsp/transform/hilbert_test.go
kortschak: Yes, I'd like the var outside the function.

dsp/transform/hilbert.go
kortschak: Suggest then having this in an example test. This shows the user how to use what we have and provides the code if they want to just copy/paste.

kortschak: Further to this, I think a more integrated example test would be worthwhile. This would be one that works through a toy example showing how the whole thing would be used.

Also, I have just noticed that this is a new package. It should have a doc.go file with package godoc explaining what the package is intended to cover.

dsp/transform/hilbert_test.go
```diff
@@ -0,0 +1,48 @@
...
+	"gonum.org/v1/gonum/cmplxs"
+)
+
+func TestHilbertAnalytic(t *testing.T) {
+	const tol = 1e-10
+	testCases := []struct {
+		in   []float64
+		want []complex128
+	}{
+		{[]float64{0, 0, 0, 0}, []complex128{0, 0, 0, 0}},
```
ccoVeille, line 20: Please add a test fit a nil slice, and an empty slice

This should be covered

dsp/transform/hilbert.go
```diff
@@ -0,0 +1,87 @@
...
+		unnorm[i] = u / complex(float64(len(unnorm)), 0)
+	}
+	return unnorm
+}
+
+// Compute the positive envelope of a real signal, and stores the result in the dst slice,
+// returning it.
+// If the dst slice is nil, a new slice will be created and returned. The dst slice
+// must be the same length as the input signal.
+func (h *Hilbert) Envelope(dst []float64, signal []float64) []float64 {
```
ccoVeille, line 77: This method is not tested

robertkleffner: Removed based on additional feedback from @kortschak

dsp/transform/hilbert_test.go
robertkleffner: These tests now pass, but I will note that these cases panic for the `CmplxFFT` methods that this transform is based on

dsp/transform/hilbert_test.go
```diff
@@ -0,0 +1,52 @@
...
+var hilbertAnalyticSignalTests = []struct {
+	name string
+	in   []float64
+	want []complex128
+}{
+	{"Hilbert nil slice", nil, nil},
+	{"Hilbert empty slice", []float64{}, []complex128{}},
+	{"Hilbert zeros test", []float64{0, 0, 0, 0}, []complex128{0, 0, 0, 0}},
+	{"Hilbert whole components test", []float64{1, 2, 3, 4}, []complex128{1 + 1i, 2 - 1i, 3 - 1i, 4 + 1i}},
+	{
```
ccoVeille, line 22: Usually, I do a map[string]struct

Where the string is the name, it avoids adding it to the struct and make sure to provide unique name in case of copypasta

dsp/transform/hilbert.go
```diff
@@ -0,0 +1,83 @@
...
+	fft  *fourier.CmplxFFT
+	work []complex128
+}
+
+// NewHilbert returns a new Hilbert transformer for signals of size n.
+// The transform is most efficient when n is a product of small primes.
+func NewHilbert(n int) *Hilbert {
+	if n < 1 {
+		return &Hilbert{nil, nil}
+	}
```
ccoVeille, line 24: Shouldn't it return an error? Or an empty slice still make sense

dsp/transform/hilbert.go
```diff
@@ -0,0 +1,83 @@
...
+
+// AnalyticSignal computes the analytical signal of a real signal, and stores
+// the result in the dst slice, returning it.
+//
+// If the dst slice is nil, a new slice will be created and returned. The dst slice
+// must be the same length as the input signal, otherwise the method will panic.
+func (h *Hilbert) AnalyticSignal(dst []complex128, signal []float64) []complex128 {
+	if len(signal) != h.Len() {
+		panic("transform: input signal length mismatch")
+	}
```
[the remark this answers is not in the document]
robertkleffner: I was trying to keep this struct's API similar to the FFT API, which does not return `error`s for its compute methods, preferring to panic. The constraint is noted in the doc comment, and I can't think of a scenario where someone would reasonably expect to get an analytic signal with different size from the input, so panic felt more appropriate anyway (implementation rather than user error).

But if `error` is the desired API here, I can put it in.

dsp/transform/hilbert.go
robertkleffner: Again, trying to mimic the FFT API, which `panic`s with argument `0`. I think `error` could make more sense here though.

ccoVeille: Let's wait for a maintainer then.

Nice addition, BTW

dsp/transform/hilbert.go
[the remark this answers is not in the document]
kortschak: > I would change signature to return an error and avoid panics

No. Please see https://go-talks.appspot.com/github.com/gonum/talks/2017/gonumtour.slide#14

dsp/transform/hilbert_test.go
kortschak: > I do a map[string]struct

Please don't do this here.

dsp/transform/hilbert.go
kortschak, line 24: Please remove this and allow a panic. A negative would be indicative of failure to hold a known invariant.

dsp/transform/hilbert.go
```diff
@@ -0,0 +1,83 @@
...
+		dst = make([]complex128, len(signal))
+	} else if len(dst) != h.Len() {
+		panic("transform: destination length mismatch")
+	}
+
+	// edge case where the length is zero or less
+	if h.fft == nil {
+		return dst
+	}
+
```
kortschak, line 56: This is not needed.
```suggestion
```

dsp/transform/hilbert_example_test.go
```diff
@@ -0,0 +1,46 @@
+package transform
+
+import (
+	"fmt"
+	"math/cmplx"
+)
+
+func ExampleHilbert_AnalyticSignal() {
+	// Samples is a set of real amplitudes that make up a signal
```
kortschak, line 9:
```suggestion
	// Samples is a set of real amplitudes that make up a signal.
```

dsp/transform/hilbert_example_test.go
```diff
@@ -0,0 +1,46 @@
...
+	samples := []float64{1, 0, 2, 0, 4, 0, 2, 0}
+
+	// Initialize a Hilbert transform and 'demodulate' to get the analytic signal.
+	// The result is the complex I/Q (In-Phase / Quadrature) demodulation of the input signal.
```
kortschak, line 13: These comments should be shorter so that they render nicely in the godoc; the don't get wrapped in the presented code, so we need to do that work here.
```suggestion
	// Initialize a Hilbert transform and 'demodulate' to get the
	// analytic signal.
	// The result is the complex I/Q (In-Phase / Quadrature) demodulation
	// of the input signal.
```

dsp/transform/hilbert_example_test.go
```diff
@@ -0,0 +1,46 @@
...
+	h := NewHilbert(len(samples))
+	iqSamples := h.AnalyticSignal(nil, samples)
+
+	// We can compute the instantaneous amplitude of the signal (or 'envelope') using absolute value.
+	// Analyzing the envelope is an easy way to measure changes in amplitude over time in a signal.
```
kortschak, line 18:
```suggestion
	// We can compute the instantaneous amplitude of the signal
	// (or 'envelope') using absolute value. Analyzing the envelope
	// is an easy way to measure changes in amplitude over time in a
	// signal.
```

dsp/transform/hilbert_example_test.go
```diff
@@ -0,0 +1,46 @@
...
+	envelope := make([]float64, len(samples))
+	for ind, iq := range iqSamples {
+		envelope[ind] = cmplx.Abs(iq)
+	}
+
+	// We can also compute the instantaneous phase of each part of the signal using the 4-quadrant arc-tangent.
+	// With multiple samples, the instantaneous phase can be used to estimate instantaneous frequency of a signal.
```
kortschak, line 25:
```suggestion
	// We can also compute the instantaneous phase of each part of the
	// signal using the 4-quadrant arc-tangent. With multiple samples,
	// the instantaneous phase can be used to estimate instantaneous
	// frequency of a signal.
```

dsp/transform/hilbert_test.go
```diff
@@ -0,0 +1,52 @@
...
+
+	"gonum.org/v1/gonum/cmplxs"
+)
+
+var hilbertAnalyticSignalTests = []struct {
+	name string
+	in   []float64
+	want []complex128
+}{
+	{"Hilbert nil slice", nil, nil},
```
kortschak, line 18:
```suggestion
	{
		name: "nil_slice",
		in:   nil,
		want: nil,
	},
```
We know that this is a Hilbert test, so we don't need that. The field names make things more navigable. Using spaces in the name means that when the test output is emitted, the name can be directly searched for in the source (`go test` does `s/ /_/g` in test names). Similar changes below.

dsp/transform/hilbert_test.go
```diff
@@ -0,0 +1,52 @@
...
+		t.Run(test.name, func(t *testing.T) {
+			h := NewHilbert(len(test.in))
+			if h.Len() != len(test.in) {
+				t.Errorf("unexpected Hilbert transform length: got:%d, want:%d", h.Len(), len(test.in))
+			}
+
+			dst := make([]complex128, len(test.in))
+			analytic := h.AnalyticSignal(dst, test.in)
+			if !cmplxs.EqualApprox(test.want, analytic, tol) {
+				t.Errorf("expected Hilbert transform result %v, got %v", test.want, analytic)
```
kortschak, line 48:
```suggestion
			got := h.AnalyticSignal(dst, test.in)
			if !cmplxs.EqualApprox(got, test.want, tol) {
				t.Errorf("unexpected Hilbert transform result:\ngot: %v\nwant:%v", got, test.want)
```

dsp/transform/doc.go
```diff
@@ -0,0 +1,6 @@
+// Copyright ©2024 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Package transform provides important transforms on signals used in digital signal processing
```
kortschak, line 5:
```suggestion
// Package transform provides important transforms on signals used in digital signal processing.
```

kortschak (commented):
Should be good after this. Thanks.

dsp/transform/hilbert_test.go
ccoVeille: OK.

I'm fine with having no map[string]struct but do you have any reason you could share

dsp/transform/hilbert.go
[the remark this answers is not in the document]
ccoVeille: OK.

dsp/transform/hilbert_test.go
kortschak: Consistent test ordering.

ccoVeille: I have never had a need for that, except an `//Output:` in an `ExampleFoo`

But OK, I'm fine with this

dsp/transform/hilbert_example_test.go
```diff
@@ -0,0 +1,46 @@
...
+	phase := make([]float64, len(samples))
+	for ind, iq := range iqSamples {
+		phase[ind] = cmplx.Phase(iq)
+	}
+
+	for i, iq := range iqSamples {
+		fmt.Printf("ind=%d -> I=%.4g, Q=%.4g, envelope=%.4g, phase=%.4g\n",
```
kortschak, line 32:
```suggestion
		fmt.Printf("ind=%d -> I=%.4f, Q=%.4f, envelope=%.4f, phase=%.4f\n",
```
`%g` offers the opportunity to represent ~0 as the tiny value, which we probably don't want.

dsp/transform/hilbert.go
```diff
@@ -0,0 +1,76 @@
...
+// of an approximate analytical signal of a real signal, and determine the
+// real envelope of a signal.
+type Hilbert struct {
+	fft  *fourier.CmplxFFT
+	work []complex128
+}
+
+// NewHilbert returns a new Hilbert transformer for signals of size n.
+// The transform is most efficient when n is a product of small primes.
+// If n is 0 or negative, the function will panic.
```
kortschak, line 21:
```suggestion
// n must not be less than one.
```
I would prefer to state what a user cannot do than what we guarantee to do.

(though I don't think we make this statement in the cause for the panic; I need to think about how I actually want to handle that)

dsp/transform/hilbert_example_test.go
```diff
@@ -0,0 +1,52 @@
+package transform
```
kortschak, line 1: We need a copyright notice here.

kortschak (approved):
This looks good. Thanks.

Please now send **another PR** adding yourself to AUTHORS and CONTRIBUTORS with the commit message, "A+C: add Robert Kleffner". When that's done and approved, I'll merge both.

kortschak:
No need to update the branch; we'll squash what's here. Don't undo this though.

PR #2004, opened 2024-11-28
graph/path: Improve performance of `YenKShortestPaths`

BarrensZeppelin:
`YenKShortestPaths`'s running time is dominated by calls to Dijkstra's shortest path algorithm, so it makes sense to focus on improvements there.
For Yen's algorithm, we are not interested in shortest path trees, but only _a_ shortest path between some source and target node. We can exploit this in Dijkstra's algorithm by terminating when the target is discovered.
Of course this does not improve the worst-case complexity, but it should be a worthwhile optimization for many practical uses.

### DIFF
--- a/graph/path/dijkstra.go
+++ b/graph/path/dijkstra.go
@@ -20,8 +20,30 @@ import (
 //
 // The time complexity of DijkstraFrom is O(|E|.log|V|).
 func DijkstraFrom(u graph.Node, g traverse.Graph) Shortest {
+	return dijkstraFrom(u, nil, g)
+}
+
+// DijkstraFromTo returns a shortest path from u to v in the graph g. The
+// result is equivalent to DijkstraFrom(u, g).To(v), but DijkstraFromTo can be
+// more efficient, as it can terminate early if v is reached. If the graph does
+// not implement Weighted, UniformCost is used. DijkstraFromTo will panic if g
+// has a u-reachable negative edge weight that is discovered before reaching v.
+//
+// The time complexity of DijkstraFromTo is O(|E|.log|V|).
+func DijkstraFromTo(u graph.Node, v graph.Node, g traverse.Graph) (path []graph.Node, weight float64) {
+	// An implementation of Bidirectional Dijkstra's algorithm could be even
+	// more efficient, but it requires a transposed (or undirected) graph.
+	if v == nil {
+		panic("dijkstra: nil target node")
+	}
+	return dijkstraFrom(u, v, g).To(v.ID())
+}
+
+// Shared implementation of DijkstraFrom and DijkstraFromTo.
+func dijkstraFrom(u, target graph.Node, g traverse.Graph) Shortest {
 	var path Shortest
-	if h, ok := g.(graph.Graph); ok {
+	// Use the incremental version when a target is provided
+	if h, ok := g.(graph.Graph); target == nil && ok {
 		if h.Node(u.ID()) == nil {
 			return Shortest{from: u}
 		}
@@ -58,6 +80,9 @@ func DijkstraFrom(u graph.Node, g traverse.Graph) Shortest {
 			continue
 		}
 		mnid := mid.node.ID()
+		if target != nil && mnid == target.ID() {
+			break
+		}
 		to := g.From(mnid)
 		for to.Next() {
 			v := to.Node()
--- a/graph/path/dijkstra.go
+++ b/graph/path/dijkstra.go
@@ -42,7 +42,7 @@ func DijkstraFromTo(u, t graph.Node, g traverse.Graph) (path []graph.Node, weigh
 
 func dijkstraFrom(u, t graph.Node, g traverse.Graph) Shortest {
 	var path Shortest
-	// Use the incremental version when a target is provided
+	// Use the incremental version when a target is provided.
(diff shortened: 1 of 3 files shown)


kortschak (commented):
Please split this into two PRs.

BarrensZeppelin:
Done. See #2005

kortschak (commented):
I believe this can be done with composition; if `g`  has `From` and `Edge` methods that return `nil` when `u` is the target, this has the same effect. I do not think we need additional API.

BarrensZeppelin:
It does not have the same effect.
Consider the following graph, where we're looking for a path from node 1 to node 2.

gonum

With the implementation in the PR, only nodes 1, 2 and 4 are pushed to the priority queue, and only nodes 1 and 2 are popped.
With you suggestion (as I understand it), we would remove the edge from 2 to 3, but we would still compute the shortest path to nodes 4 and 5, which is wasted work.

I guess it's possible to achieve the same effect with composition by making the `From` function stateful: once the target is reached, remove _all_ edges from the graph, but then correctness will depend on the order that Dijkstra's algorithm calls `From`, which seems brittle.

kortschak:
Yes, making it stateful would be required. In that construction is has exactly the same behaviour as the code here.

The required code is straightforward.
```
type stopGraph struct {
	g      traverse.Graph
	target int64
}

func (g *stopGraph) From(uid int64) graph.Nodes {
	if uid == g.target {
		g.g = empty{}
	}
	return g.g.From(uid)
}

func (g *stopGraph) Edge(uid, vid int64) graph.Edge {
	if uid == g.target {
		g.g = empty{}
	}
	return g.g.Edge(uid, vid)
}

type empty struct{}

func (empty) From(uid int64) graph.Nodes     { return graph.Empty }
func (empty) Edge(uid, vid int64) graph.Edge { return nil }
```

kortschak:
No, you're right.

graph/path/dijkstra.go
```diff
@@ -20,8 +20,30 @@ import (
...
+}
+
+// DijkstraFromTo returns a shortest path from u to v in the graph g. The
+// result is equivalent to DijkstraFrom(u, g).To(v), but DijkstraFromTo can be
+// more efficient, as it can terminate early if v is reached. If the graph does
+// not implement Weighted, UniformCost is used. DijkstraFromTo will panic if g
+// has a u-reachable negative edge weight that is discovered before reaching v.
+//
+// The time complexity of DijkstraFromTo is O(|E|.log|V|).
+func DijkstraFromTo(u graph.Node, v graph.Node, g traverse.Graph) (path []graph.Node, weight float64) {
```
kortschak, line 33:
```suggestion
func DijkstraFromTo(u, t graph.Node, g traverse.Graph) (path []graph.Node, weight float64) {
```
The second param is not a v.

graph/path/dijkstra.go
```diff
@@ -20,8 +20,30 @@ import (
...
+	// An implementation of Bidirectional Dijkstra's algorithm could be even
+	// more efficient, but it requires a transposed (or undirected) graph.
+	if v == nil {
+		panic("dijkstra: nil target node")
+	}
+	return dijkstraFrom(u, v, g).To(v.ID())
+}
+
+// Shared implementation of DijkstraFrom and DijkstraFromTo.
+func dijkstraFrom(u, target graph.Node, g traverse.Graph) Shortest {
```
kortschak, line 43:
```suggestion
func dijkstraFrom(u, t graph.Node, g traverse.Graph) Shortest {
```

graph/path/dijkstra_test.go
```diff
@@ -56,38 +79,41 @@ func TestDijkstraFrom(t *testing.T) {
...
+			p, weight := pt.To(test.Query.To().ID())
+			checkPath(tg.typ, p, weight)
 
 			np, weight := pt.To(test.NoPathFor.To().ID())
 			if pt.From().ID() == test.NoPathFor.From().ID() && (np != nil || !math.IsInf(weight, 1)) {
 				t.Errorf("%q %s: unexpected path:\ngot: path=%v weight=%f\nwant:path=<nil> weight=+Inf",
 					test.Name, tg.typ, np, weight)
 			}
 		}
+
```
kortschak, line 96: I think this should be in the loop, with helper sharing.

BarrensZeppelin: Done.

BarrensZeppelin:
These PRs are ready for re-review.

graph/path/dijkstra.go
```diff
@@ -20,8 +20,30 @@ import (
...
+	if t == nil {
+		panic("dijkstra: nil target node")
+	}
+	return dijkstraFrom(u, t, g).To(t.ID())
+}
+
+func dijkstraFrom(u, t graph.Node, g traverse.Graph) Shortest {
 	var path Shortest
-	if h, ok := g.(graph.Graph); ok {
+	// Use the incremental version when a target is provided
```
kortschak, line 45:
```suggestion
	// Use the incremental version when a target is provided.
```

kortschak (commented):
LGTM after nit.

I'll squash the commits here into a single commit. Please send another PR adding yourself to AUTHORS and CONTRIBUTORS with the commit message "A+C: add Oskar Haarklou Veileborg"

kortschak (approved):

PR #2005, opened 2024-11-28
graph/path: Implement standard improvements to Yen's KSP algorithm

BarrensZeppelin:
I implemented two common optimizations for Yen's algorithm that can reduce the number of calls to the shortest path subroutine: https://en.wikipedia.org/wiki/Yen%27s_algorithm#Improvements


graph/path/yen_ksp_test.go
```diff
@@ -329,6 +329,33 @@ var yenShortestPathTests = []struct {
 			{1, 4, 5, 2, 3, 6},
 		},
 	},
+	{
```
kortschak, line 332: What is this new test for? It passes without the implementation change.

graph/path/yen_ksp.go
kortschak, line 1: I'd like the changes that are not required for the optimisations to be backed out.

graph/path/yen_ksp.go
```diff
@@ -111,7 +111,16 @@ func YenKShortestPaths(g graph.Graph, k int, cost float64, s, t graph.Node) [][]
 		if len(best.path) <= 1 || best.weight > cost {
 			break
 		}
+		if neededPaths := k - i; k >= 0 && len(pot) >= neededPaths && best.weight == pot[neededPaths-1].weight {
```
kortschak, line 114: This is the second optimisation? These should be in separate commits.

kortschak (commented):
Please revise this to be the two optimisations in separate commits with no other changes. We can do the other changes later if they merit it.

Note that the format of commit messages starts the sentence fragment with a lower case.

graph/path/yen_ksp.go
BarrensZeppelin: Done

graph/path/yen_ksp.go
BarrensZeppelin: Yes. Done.

PR #2013, opened 2025-01-18
graph/traverse: add depth parameter to Visit() and Traverse() to breadth-first search

xqqp:
This PR makes the current depth available during breadth-first search when `Visit()` or `Traverse()` is called. The current depth is already available when `until()` is called. This change makes the parameter accessible in the other two functions.

`until()` is only called when the current traversal depth is changed. Having the current depth available during `Visit()` and `Traverse()` opens up more possibilities for collecting edges or nodes during traversal.

While this is an API change, it is a minor one with no change in behavior. Users can simply ignore the new available parameter.

### DIFF
--- a/graph/traverse/traverse.go
+++ b/graph/traverse/traverse.go
@@ -28,15 +28,17 @@ type Graph interface {
 // BreadthFirst implements stateful breadth-first graph traversal.
 type BreadthFirst struct {
 	// Visit is called on all nodes on their first visit.
-	Visit func(graph.Node)
+	// depth is the current traversal depth.
+	Visit func(n graph.Node, depth int)
 
 	// Traverse is called on all edges that may be traversed
 	// during the walk. This includes edges that would hop to
-	// an already visited node.
+	// an already visited node. depth is the current traversal
+	// depth for the from node of the edge.
 	//
 	// The value returned by Traverse determines whether
 	// an edge can be traversed during the walk.
-	Traverse func(graph.Edge) bool
+	Traverse func(e graph.Edge, depth int) bool
 
 	queue   linear.NodeQueue
 	visited set.Ints[int64]
@@ -53,7 +55,7 @@ func (b *BreadthFirst) Walk(g Graph, from graph.Node, until func(n graph.Node, d
 	}
 	b.queue.Enqueue(from)
 	if b.Visit != nil && !b.visited.Has(from.ID()) {
-		b.Visit(from)
+		b.Visit(from, 0)
 	}
 	b.visited.Add(from.ID())
 
@@ -72,14 +74,14 @@ func (b *BreadthFirst) Walk(g Graph, from graph.Node, until func(n graph.Node, d
 		for to.Next() {
 			n := to.Node()
 			nid := n.ID()
-			if b.Traverse != nil && !b.Traverse(g.Edge(tid, nid)) {
+			if b.Traverse != nil && !b.Traverse(g.Edge(tid, nid), depth) {
 				continue
 			}
 			if b.visited.Has(nid) {
 				continue
 			}
 			if b.Visit != nil {
-				b.Visit(n)
+				b.Visit(n, depth)
 			}
 			b.visited.Add(nid)
 			children++
--- a/graph/traverse/traverse_test.go
+++ b/graph/traverse/traverse_test.go
@@ -71,7 +71,7 @@ var (
 var breadthFirstTests = []struct {
 	g     []intset
 	from  graph.Node
-	edge  func(graph.Edge) bool
+	edge  func(graph.Edge, int) bool
 	until func(graph.Node, int) bool
 	final map[graph.Node]bool
 	want  [][]int64
@@ -89,7 +89,7 @@ var breadthFirstTests = []struct {
 	},
 	{
 		g: wpBronKerboschGraph,
-		edge: func(e graph.Edge) bool {
+		edge: func(e graph.Edge, _ int) bool {
 			// Do not traverse an edge between 3 and 5.
 			return (e.From().ID() != 3 || e.To().ID() != 5) && (e.From().ID() != 5 || e.To().ID() != 3)
 		},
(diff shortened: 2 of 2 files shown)


kortschak:
Please show an example of use where this is required. That is, a use which cannot be implemented with the current code. This is a breaking change; breaking is binary, not graduated.

xqqp:
> Please show an example of use where this is required. That is, a use which cannot be implemented with the current code. This is a breaking change; breaking is binary, not graduated.

Example: In a graph of connected web pages, the creation time of each web page is available. 
We start the traversal at a given node, and at each edge we only traverse further if the duration between the creation of the two web pages is smaller than duration_A. If we reach a certain depth of traversal, we want to increase the allowed duration to duration_B.

In the example above, the dynamic change of the maximum duration of creation time between two web pages is only possible if we have access to the current traversal depth. 

I added a test using the traversal depth in `Traverse()`. There, node IDs are compared based on the current traversal depth. 

I agree that a breaking change either happens or does not, but the effort required to react to the breaking change in this case would be small.

kortschak:
> Example: In a graph of connected web pages, the creation time of each web page is available.
We start the traversal at a given node, and at each edge we only traverse further if the duration between the creation of the two web pages is smaller than duration_A. If we reach a certain depth of traversal, we want to increase the allowed duration to duration_B.

This seems like a very bespoke use. Note that the package is very small and "A little copying is better than a little dependency." is a fine proverb for this case. Please provide a code example rather than a natural language example.

> I agree that a breaking change either happens or does not, but the effort required to react to the breaking change in this case would be small.

This gets multiplied by all users that it breaks.

graph/traverse/traverse.go
```diff
@@ -28,15 +28,15 @@ type Graph interface {
 // BreadthFirst implements stateful breadth-first graph traversal.
 type BreadthFirst struct {
 	// Visit is called on all nodes on their first visit.
-	Visit func(graph.Node)
+	Visit func(graph.Node, int)
```
kortschak, line 31: This should have associated documentation changes; the type here is no longer adequate documentation.

graph/traverse/traverse.go
```diff
@@ -28,15 +28,15 @@ type Graph interface {
...
 
 	// Traverse is called on all edges that may be traversed
 	// during the walk. This includes edges that would hop to
 	// an already visited node.
 	//
 	// The value returned by Traverse determines whether
 	// an edge can be traversed during the walk.
-	Traverse func(graph.Edge) bool
+	Traverse func(graph.Edge, int) bool
```
kortschak, line 39: Same here.

graph/traverse/traverse_test.go
```diff
@@ -140,6 +140,24 @@ var breadthFirstTests = []struct {
...
 	},
+	{
+		g:    batageljZaversnikGraph,
+		from: simple.Node(6),
+		edge: func(e graph.Edge, d int) bool {
+			if d < 2 {
+				return true
+			} else {
+				return e.To().ID()%2 == 0
+			}
```
kortschak, line 151:
```suggestion
			return d < 2 || e.To().ID()%2 == 0
```

xqqp:
The depth parameter is useful in every case where edge data is evaluated in combination with the depth, which is in my opinion quite generic. Furthermore, the depth parameter is also already provided in the `until()` function, so why not make it consistent in all the other related breadth-first search functions. 
But you are of course right that for such use-cases copying and modifying the code is easy enough. Feel free to close this PR.

graph/traverse/traverse.go
xqqp: done

graph/traverse/traverse_test.go
xqqp: done

graph/traverse/traverse.go
xqqp: done

kortschak:
> Furthermore, the depth parameter is also already provided in the until() function, so why not make it consistent in all the other related breadth-first search functions.

What I'm saying is that at the moment, it's not compelling enough. If you can show me a more general case that I can provably see that cannot be implemented with the current code, I'm happy to say the balance is in favour, so please provide an example.

xqqp:
> What I'm saying is that at the moment, it's not compelling enough. If you can show me a more general case that I can provably see that cannot be implemented with the current code, I'm happy to say the balance is in favour, so please provide an example.

A more general example I can think of is, that if sombody would want to collect all traversed edges, including at which depth it was traversed, during a BFS search. This is currently not possible because `Traverse()` does not have the depth information available and `until()` does not have the edge information available.

graph/traverse/traverse.go
```diff
@@ -28,15 +28,17 @@ type Graph interface {
 // BreadthFirst implements stateful breadth-first graph traversal.
 type BreadthFirst struct {
 	// Visit is called on all nodes on their first visit.
-	Visit func(graph.Node)
+	// The second parameter is the current traversal depth.
+	Visit func(graph.Node, int)
```
kortschak, line 32:
```suggestion
	// depth is the current traversal depth.
	Visit func(n graph.Node, depth int)
```

graph/traverse/traverse.go
```diff
@@ -28,15 +28,17 @@ type Graph interface {
...
 
 	// Traverse is called on all edges that may be traversed
 	// during the walk. This includes edges that would hop to
-	// an already visited node.
+	// an already visited node. The second parameter is the
+	// current traversal depth.
```
kortschak, line 37:
```suggestion
	// an already visited node. depth is the traversal
	// depth for the from node.
```

graph/traverse/traverse.go
kortschak: This should be like the case for `Visit`, though `depth` needs to be more clearly defined since it refers to node depth, so the documentation needs to be clear about whether this is head or tail depth.

kortschak:
Note that
```
var depth int
bfs := traverse.BreadthFirst{
    Traverse: func(e graph.Edge) bool {
        // use depth in logic.
    },
}
bfs.Walk(g, s, func(_ graph.Node, d int) bool { depth = d; return true })
```
and variations on this, do achieve all that this change does, though it may not be immediately obvious.

kortschak:
@xqqp Please rebase this onto master, and also address my [comment above](https://github.com/gonum/gonum/pull/2013#issuecomment-2600402926).

xqqp:
Rebased. Regarding your comment, while the same behavior can be achieved with a shared variable, I think it is still cleaner to provide the depth as an argument. Let me know what you think.

kortschak:
Thanks. I agree it's clearer. We've just made a significant breaking change (changed x/exp/rand to math/rand/v2), so I'm less concerned about that now. I'll think about this.

kortschak:
@vladimir-ch Do you have a view?

vladimir-ch:
I've gone through the discussion and I'm inclined to _not_ making this change. @kortschak 's snippet shows how to achieve it using `until`, so the API change would be just adding another way. We could convert the snippet into an example to make it more obvious.

xqqp:
Alright, thanks for taking the time to review.

PR #2020, opened 2025-02-13
all: use Go-1.24, drop Go-1.22

sbinet:
Please take a look.


sbinet:
PTAL (also @kortschak could you remove the required flag for `1.22.x` ?)

kortschak (approved):
Thanks

PR #2022, opened 2025-02-14
all: use go tool directive

### LINKED ISSUE #2021: proposal: consider using the new "tool" directive
With Go-1.24 out, we should probably consider using the new "tool" directive (once Go-1.24 becomes the minimal required version, so once Go-1.25 is out, in August 2025).

this would probably be relevant for:

- `github.com/goccmack/gocc`
- `golang.org/x/tools` (well, it's used in `script.d/deps.sh` but also in `stat/distmv/studentst.go` for its `container/intsets`...)
- `gonum.org/v1/tools` (for `v1/tools/cmd/{check-imports,check-copyright}`)

sbinet:
Fixes #2021

Please take a look.


kortschak:
So this needs to wait until go1.25.

sbinet:
Yes. As I said on the issue/proposal (but I guess it's fair to mention it here again), this PR is documentation for our future selves :)

sbinet:
PTAL
(I don't think the flakyness of the tests is coming from this PR)

graph/formats/dot/internal/tools.go
kortschak, line 1: Looks like this file is no longer needed.

go.mod
```diff
@@ -16,10 +15,21 @@ require (
 	git.sr.ht/~sbinet/gg v0.6.0 // indirect
 	github.com/ajstarks/svgo v0.0.0-20211024235047-1546f124cd8b // indirect
 	github.com/campoy/embedmd v1.0.0 // indirect
+	github.com/goccmack/gocc v1.0.2 // indirect
```
kortschak, line 18: This version bump had no impact on the generated code?

sbinet:
```sh
cd ./graph/formats/dot/internal
make clean
make
git diff
```
doesn't show anything.

would that suffice as a quality check ?

kortschak:
#2060

kortschak (approved):
Thanks

PR #2024, opened 2025-02-21
stat/distmv: replace intset.Sparse with linear search

kruskall:
x/tools is only used in one place because of the intsets package to compute the diff between two sets.
Reimplement the logic using the stdlib:

quick benchmarks show some nice perf improvements:

```
            │   old.txt    │               new.txt               │
            │    sec/op    │   sec/op     vs base                │
FindUnob-20   169.80n ± 2%   38.25n ± 2%  -77.47% (p=0.000 n=10)

            │   old.txt   │              new.txt               │
            │    B/op     │    B/op     vs base                │
FindUnob-20   200.00 ± 0%   64.00 ± 0%  -68.00% (p=0.000 n=10)

            │  old.txt   │              new.txt               │
            │ allocs/op  │ allocs/op   vs base                │
FindUnob-20   4.000 ± 0%   1.000 ± 0%  -75.00% (p=0.000 n=10)
```

### DIFF
--- a/stat/distmv/studentst.go
+++ b/stat/distmv/studentst.go
@@ -218,10 +218,10 @@ func findUnob(observed []int, dim int) (unobserved []int) {
 	unobserved = slices.DeleteFunc(all, func(i int) bool {
 		return i == -1
 	})
-	cp := slices.Clone(observed)
-	slices.Sort(cp)
-	for i := 1; i < len(cp); i++ {
-		if cp[i] == cp[i-1] {
+	observed = slices.Clone(observed)
+	slices.Sort(observed)
+	for i, v := range observed[1:] {
+		if observed[i] == v {
 			panic("duplicate values in observed")
 		}
 	}


stat/distmv/studentst.go
```diff
@@ -208,18 +206,18 @@ func studentsTConditional(observed []int, values []float64, nu float64, mu []flo
 // findUnob returns the unobserved variables (the complementary set to observed).
 // findUnob panics if any value repeated in observed.
```
kortschak, line 207: In order to maintain this invariant we need to check for repeats. This suggests an approach of sorting the `observed` slice and walking it, retaining gaps and panicking on repeats. Whether it's better to preallocate or append (or partial preallocate and then append) is unclear and would need measurement.

Can you commit the benchmark please (over a variety of `observed` lengths)?

kortschak (commented):
Suggest commit message: "stat/distmv: replace intset.Sparse with linear search" and then your justification.

stat/distmv/studentst.go
kortschak: The godoc for the user-facing functions does not say the parameter will be mutated, so it is unsafe to sort and what you have is necessary.

stat/distmv/studentst.go
```diff
@@ -208,18 +206,25 @@ func studentsTConditional(observed []int, values []float64, nu float64, mu []flo
...
-		setOb.Insert(v)
+	all := make([]int, 0, dim)
+	for i := range dim {
+		all = append(all, i)
 	}
-	var setAll intsets.Sparse
-	for i := 0; i < dim; i++ {
-		setAll.Insert(i)
+	for _, obs := range observed {
+		if obs >= 0 && obs < dim {
```
kortschak, line 214: nit: we use the convention of left <= right for ranges, so this would be `0 <= obs && obs < dim`.

stat/distmv/studentst.go
```diff
@@ -208,18 +206,25 @@ func studentsTConditional(observed []int, values []float64, nu float64, mu []flo
...
+			all[obs] = -1
+		}
+	}
+	unobserved = slices.DeleteFunc(all, func(i int) bool {
+		return i == -1
+	})
+	cp := slices.Clone(observed)
+	slices.Sort(cp)
+	for i := 1; i < len(cp); i++ {
+		if cp[i] == cp[i-1] {
```
kortschak, line 224:
```suggestion
	observed = slices.Clone(observed)
	slices.Sort(observed)
	for i, v := range observed[1:] {
		if observed[i] == v {
```

kortschak (approved):
I spent a little time yesterday looking to see whether a pre-assessment of missing observeds to guide allocations would improve performance, it did not, at significant complexity cost.

LGTM, but I'd like a second opinion; I'll do the merge when that's given.

/cc @vladimir-ch

PR #2025, opened 2025-02-22
github: add reading check instruction to pr template

kortschak:
Please take a look.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE
+++ b/.github/PULL_REQUEST_TEMPLATE
@@ -13,4 +13,7 @@ Please make sure your commit message summary line and pull request
 title match the Go convention; a one-line summary of the change,
 prefixed by the primary affected package that should complete the
 sentence, "This change modifies Gonum to _____."
+
+Do not delete any of this templated text in the PR description, but
+feel free to add additional context.
 -->


.github/PULL_REQUEST_TEMPLATE
```diff
@@ -13,4 +13,7 @@ Please make sure your commit message summary line and pull request
 title match the Go convention; a one-line summary of the change,
 prefixed by the primary affected package that should complete the
 sentence, "This change modifies Gonum to _____."
+
+Do not delete any of this templated text in the PR description, but
+feel free to add addition context.
```
vladimir-ch, line 18: "additional" or omit?

vladimir-ch: Sorry, I meant there's a typo, should be "additional", not "addition".

kortschak: :facepalm: Thank you. I need get English lessons from you as well as Czech.

vladimir-ch: I see it so often that people write about adding additional something and the redundancy has always bothered me and from now it won't :tada: :v:

Issue #649, opened 2018-11-14
proposal: add hypergeometric functions


buyology:
I recently needed the Gauss hypergeometric function so [did a quick translation](https://github.com/buyology/x/tree/master/hyp2f1) over from the Cephes math lib version.

It would be nice to see parity with [the set of functions available in `scipy`](https://docs.scipy.org/doc/scipy/reference/special.html#hypergeometric-functions).

Maybe it could go into `mathext` or a new package/sub-package.

deasmhumhna:
@buyology  
I also need hypergeometric functions to translate the Cephes implementation of modified Bessel functions of the first kind. Could you please submit a pull request? I think it should go in `mathext` (or at rather, the Cephes translation should go in `mathext/internal/cephes` and should be wrapped at the higher level) since that's where the other "special" functions currently live.

fumin:
> I recently needed the Gauss hypergeometric function so [did a quick translation](https://github.com/buyology/x/tree/master/hyp2f1) over from the Cephes math lib version.

@buyology , the link to your translation seems to have stopped working, can you share an updated link? Thanks.

PR #2026, opened 2025-02-26
mathext: add Hypergeo for computing the Gaussian Hypergeometric function

fumin:
Fixes https://github.com/gonum/gonum/issues/649


mathext/hyp2f1.go
```diff
@@ -0,0 +1,466 @@
...
+	y1 *= math.Gamma(c) / math.Gamma(b) * math.Gamma(b-a) / math.Gamma(c-a) * math.Pow(-z, -a)
+	y2, err := hyp2f1(b, 1-c+b, 1-a+b, 1/z)
+	if err != nil {
+		return y2, err
+	}
+	y2 *= math.Gamma(c) / math.Gamma(a) * math.Gamma(a-b) / math.Gamma(c-b) * math.Pow(-z, -b)
+	return y1 + y2, nil
+}
+
+/*							hyp2f1.c
```
kortschak, line 53: The section below this should live in mathext/internal/cephes where it will be clearly marked as licensed under the Cephes license. Best in its own file with appropriate license specialisation.

This will also allow sharing of precision constants.

It would be nice if fewer fortranish calling conventions were used, but this can be fixed later.

mathext/hyp2f1_scipy_test.go
```diff
@@ -0,0 +1,54 @@
...
+	expected := 1.8202169687521206
+	if d := y - expected; math.Abs(d) > 5e-15 {
+		t.Errorf("%f %f %f", y, expected, d)
+	}
+}
+
+func TestHyp2f1_SomeRealPoints(t *testing.T) {
+	t.Parallel()
+
+	// Taken from test_mpmath.py test_hyp2f1_real_some_points.
```
kortschak, line 23: This needs a link and if it's a translation as opposed to just taking a set of constants, this has license implications.

mathext/hyp2f1_scipy_test.go
```diff
@@ -0,0 +1,54 @@
+package mathext
+
+import (
+	"fmt"
+	"math"
+	"testing"
+)
+
+func TestHyp2f1_Miscellaneous(t *testing.T) {
```
kortschak, line 9: Use subtests rather than underscoring.

mathext/hyp2f1_scipy_test.go
```diff
@@ -0,0 +1,54 @@
...
+func TestHyp2f1_SomeRealPoints(t *testing.T) {
+	t.Parallel()
+
+	// Taken from test_mpmath.py test_hyp2f1_real_some_points.
+	tests := []struct {
+		a        float64
+		b        float64
+		c        float64
+		z        float64
+		expected float64
```
kortschak, line 29:
```suggestion
		want float64
```

mathext/hyp2f1_test.go
```diff
@@ -0,0 +1,230 @@
+// Tests in this file are taken from R's hypergeo package.
```
kortschak, line 1: This has license implications, as above. I would move this down to the tests var though.

mathext/hyp2f1_test.go
```diff
@@ -0,0 +1,230 @@
...
+	}
+	return y
+}
+
+func TestMain(m *testing.M) {
+	flag.Parse()
+	log.SetFlags(log.Lmicroseconds | log.Llongfile | log.LstdFlags)
+
+	m.Run()
+}
```
kortschak, line 230: Please don't do this.

mathext/hyp2f1.go
```diff
@@ -0,0 +1,466 @@
+package mathext
```
kortschak, line 1: We will need license notes for all the files.

mathext/hyp2f1.go
```diff
@@ -0,0 +1,466 @@
...
+var mACHEP float64 = (math.Nextafter(1.0, 2.0) - 1.0) / 2
+
+const eps float64 = 1e-13
+const ethresh float64 = 1e-12
+
+// Hyp2f1 returns the value of the Gaussian Hypergeometric function at z.
+// For |z| < 1, this implementation follows the Cephes library.
+// For |z| > 1, this implementation perform analytic continuation via relevant Hypergeometric identities.
+// See https://en.wikipedia.org/wiki/Hypergeometric_function for more details.
+func Hyp2f1(a float64, b float64, c float64, z float64) (float64, error) {
```
kortschak, line 18: Suggest renaming this.
```suggestion
func Hypergeo(a, b, c, z float64) (float64, error) {
```

Can we express the error conditions in-band? The only tenuous cases seem to me to be overflow (expressible as Inf rather than `math.MaxFloat64` and the partial loss of precision (how important is that?)).

mathext/hyp2f1.go
```diff
@@ -0,0 +1,466 @@
...
+	if math.Abs(z) < 1 {
+		return hyp2f1(a, b, c, z)
+	}
+
+	// Function undefined between the 1 and Inf branch points.
+	if z > 0 {
+		return math.NaN(), fmt.Errorf("analytic continuation failed")
```
kortschak, line 25: If we are using errors, please use `errors.New` and if the sentinels are important (this is an open question), export them as package vars rather than forcing client code to do string comparison if they care.

kortschak (commented):
Thank you for the contribution.

mathext/hyp2f1.go
fumin: Yes, this should really go into internal/cephes, sorry for not being aware of the folder structures.
Putting this into internal/cephes really brings lots of benefits, such as reusing of constants and error messages like you mentioned.

mathext/hyp2f1_scipy_test.go
fumin: Thanks for the suggestion, much more concise.

mathext/hyp2f1_test.go
fumin: Sorry, should have removed this before committing.

mathext/hyp2f1.go
fumin: License notes have been added to all newly introduced files.

mathext/hyp2f1.go
fumin: Good suggestions! `Hypergeo` reads much clearer. Since there're only two tenuous cases, `Inf` and `NaN` represent them nicely.

fumin:
One particular note on the latest refactor:

I have added internal/cephes/psi.go, which really is not needed since we already have `Digamma`.
I had to add psi.go because `mathext` imports `internal/cephes`, and thus if I tried to call `Digamma` in internal/cephes/hyp2f1.go, I get a circular import.
Since psi.go is pretty short, hopefully this isn't too big an issue.

mathext/hypergeo_test.go
```diff
@@ -0,0 +1,296 @@
...
+		{x: 0.28, y: 1.3531156987873853569937},
+		{x: -0.79, y: 0.5773356740314405932679},
+		{x: 0.56, y: 2.1085704049533617876477},
+		{x: -2.13, y: 0.3352446571148822718200},
+		{x: -0.43, y: 0.7150355048137748692483},
+		{x: -1.23, y: 0.4670987707934830535095},
+	}
+	for i, test := range tests {
+		t.Run(fmt.Sprintf("%d", i), func(t *testing.T) {
+			t.Parallel()
```
kortschak, line 30:
```suggestion
```
I don't think we need this here.

mathext/hypergeo_test.go
```diff
@@ -0,0 +1,296 @@
...
+			y := Hypergeo(1.21, 1.443, 1.88, test.x)
+			if d := y - test.y; math.Abs(d) > 1e-12 {
+				t.Errorf("%f %f %f %f", test.x, y, test.y, d)
```
kortschak, line 33:
```suggestion
				t.Errorf("unexpected result from Hypergeo(1.21, 1.443, 1.88, %f): got=%f want=%f", test.x, y, test.y)
```

mathext/hypergeo_test.go
```diff
@@ -0,0 +1,296 @@
...
+		{x: 0.56, y: 2.1085704049533617876477},
+		{x: -2.13, y: 0.3352446571148822718200},
+		{x: -0.43, y: 0.7150355048137748692483},
+		{x: -1.23, y: 0.4670987707934830535095},
+	}
+	for i, test := range tests {
+		t.Run(fmt.Sprintf("%d", i), func(t *testing.T) {
+			t.Parallel()
+			y := Hypergeo(1.21, 1.443, 1.88, test.x)
+			if d := y - test.y; math.Abs(d) > 1e-12 {
```
kortschak, line 32:
```suggestion
			if !scalar.EqualWithinAbs(y, test.y, 1e-12) {
```

mathext/hypergeo_test.go
```diff
@@ -0,0 +1,296 @@
...
+	"testing"
+)
+
+func TestHyp2f1(t *testing.T) {
+	t.Parallel()
+
+	// Constants taken from https://github.com/RobinHankin/hypergeo/blob/master/tests/testthat/test_aaa.R
+	var tests = []struct {
+		x float64
+		y float64
```
kortschak, line 19:
```suggestion
		want float64
```

mathext/hypergeo_test.go
```diff
@@ -0,0 +1,296 @@
...
+		{x: 0.28},
+		{x: -0.79},
+		{x: 0.56},
+		{x: -2.13},
+		{x: -0.43},
+		{x: -1.23},
+	}
+	for i, test := range tests {
+		t.Run(fmt.Sprintf("%d", i), func(t *testing.T) {
+			t.Parallel()
```
kortschak, line 64:
```suggestion
```

mathext/hypergeo_test.go
```diff
@@ -0,0 +1,296 @@
...
+
+			// Ignore z=-2.13, since both R's hypergeo and Maple don't handle this case either.
+			if test.x == -2.13 {
+				return
+			}
+
+			a := eqn15_1_15a(0.2, test.x)
+			b := eqn15_1_15b(0.2, test.x)
+			if d := a - b; math.Abs(d) > 1e-6 {
+				t.Errorf("%f %f %f %f", test.x, a, b, d)
```
kortschak, line 74: Similar changes here and below.

mathext/internal/cephes/.incbeta.go.swp
kortschak, line 1: Please delete this.

mathext/hypergeo_test.go
```diff
@@ -0,0 +1,296 @@
+// Copyright ©2016 The Gonum Authors. All rights reserved.
```
kortschak, line 1:
```suggestion
// Copyright ©2025 The Gonum Authors. All rights reserved.
```

mathext/internal/cephes/hyp2f1.go
```diff
@@ -0,0 +1,422 @@
+// Copyright ©2016 The Gonum Authors. All rights reserved.
```
kortschak, line 1:
```suggestion
// Copyright ©2025 The Gonum Authors. All rights reserved.
```

mathext/internal/cephes/psi.go
```diff
@@ -0,0 +1,147 @@
+// Copyright ©2016 The Gonum Authors. All rights reserved.
```
kortschak, line 1:
```suggestion
// Copyright ©2025 The Gonum Authors. All rights reserved.
```

mathext/internal/cephes/psi.go
```diff
@@ -0,0 +1,147 @@
...
+		z = 1 / (s * s)
+		y = z * polevl(z, A, 6)
+	} else {
+		y = 0
+	}
+
+	y = math.Log(s) - 0.5/s - y - w
+
+done:
+	;
```
kortschak, line 142: ?

kortschak (commented):
I'll try to take another look on the weekend.

mathext/hypergeo_test.go
fumin: Thanks for making me aware of `EqualWithinAbs`, which is a very useful and powerful tool.
Applied here and elsewhere.

fumin:
> I'll try to take another look on the weekend.

Thanks in advance for the coming review.
All comments so far have been addressed, and are ready for review.

mathext/internal/cephes/hyp2f1.go
```diff
@@ -0,0 +1,443 @@
...
+	var ax float64
+	var id float64
+	var d1 float64
+	var d2 float64
+	var e float64
+	var y1 float64
+	var i int32
+	var aid int32
+
+	err = 0
```
kortschak, line 251:
```suggestion
	var (
		p, q, r, s, t, d, err, err1 float64
		ax, id, d1, d2, e, y1       float64
		i, aid                      int32
	)
```

mathext/internal/cephes/hyp2f1.go
```diff
@@ -0,0 +1,443 @@
...
+	s = 1 - x
+	if x < -0.5 {
+		if b > a {
+			y = math.Pow(s, -a) * hys2f1(a, c-b, c, -x/s, &err)
```
kortschak, line 255:
```suggestion
			y, err = hys2f1(a, c-b, c, -x/s)
			y *= math.Pow(s, -a)
```

mathext/internal/cephes/hyp2f1.go
```diff
@@ -0,0 +1,443 @@
...
+		} else {
+			y = math.Pow(s, -b) * hys2f1(c-a, b, c, -x/s, &err)
```
kortschak, line 257:
```suggestion
			y, err = hys2f1(c-a, b, c, -x/s)
			y *= math.Pow(s, -b)
```

mathext/internal/cephes/hyp2f1.go
```diff
@@ -0,0 +1,443 @@
...
+	if x > 0.9 {
+		if math.Abs(d-id) > eps { // test for integer c-a-b
+			// Try the power series first
+			y = hys2f1(a, b, c, x, &err)
+			if err < ethresh {
+				goto done
+			}
+
+			// If power series fails, then apply AMS55 #15.3.6
+			q = hys2f1(a, b, 1-d, s, &err)
```
kortschak, line 274:
```suggestion
			q, err = hys2f1(a, b, 1-d, s)
```

mathext/internal/cephes/hyp2f1.go
```diff
@@ -0,0 +1,443 @@
...
+			q *= gammaADivBDivC(d, c-a, c-b)
+			r = math.Pow(s, d) * hys2f1(c-a, c-b, d+1, s, &err1)
```
kortschak, line 276:
```suggestion
			r, err1 = hys2f1(c-a, c-b, d+1, s)
			r *= math.Pow(s, d)
```

mathext/internal/cephes/hyp2f1.go
```diff
@@ -0,0 +1,443 @@
...
+			}
+
+			y += y1
+		psidon:
+			goto done
+		}
+	}
+
+	// Use defining power series if no special cases
+	y = hys2f1(a, b, c, x, &err)
```
kortschak, line 366:
```suggestion
	y, err = hys2f1(a, b, c, x)
```

mathext/internal/cephes/hyp2f1.go
```diff
@@ -0,0 +1,443 @@
...
+			r = math.Pow(s, d) * hys2f1(c-a, c-b, d+1, s, &err1)
+			r *= gammaADivBDivC(-d, a, b)
+			y = q + r
+
+			q = math.Abs(q) // estimate cancellation error
+			r = math.Abs(r)
+			if q > r {
+				r = q
+			}
+			err += err1 + machEp*r/y
```
kortschak, line 285:
```suggestion
			err += err1 + float64(machEp*r/y)
```

mathext/internal/cephes/hyp2f1.go
```diff
@@ -0,0 +1,443 @@
...
+	var h float64
+	var k float64
+	var m float64
+	var s float64
+	var u float64
+	var umax float64
+	var i int32
+
+	i = 0
+	umax = 0
```
kortschak, line 387:
```suggestion
	var (
		f, g, h, k, m, u, umax float64
		i                      int32
	)
```

mathext/internal/cephes/hyp2f1.go
```diff
@@ -0,0 +1,443 @@
...
+	f = a
+	g = b
+	h = c
+	s = 1
+	u = 1
+	k = 0
```
kortschak, line 393:
```suggestion
```

mathext/internal/cephes/hyp2f1.go
```diff
@@ -0,0 +1,443 @@
...
+	y = hys2f1(a, b, c, x, &err)
+
+done:
+	*loss = err
+	return y
+}
+
+/* Defining power series expansion of Gauss hypergeometric function */
+// loss estimates loss of significance upon return.
+func hys2f1(a, b, c, x float64, loss *float64) float64 {
```
kortschak, line 375:
```suggestion
func hys2f1(a, b, c, x float64) (s, loss float64) {
```

mathext/internal/cephes/hyp2f1.go
```diff
@@ -0,0 +1,443 @@
...
+			goto done
+		}
+	}
+
+	// Use defining power series if no special cases
+	y = hys2f1(a, b, c, x, &err)
+
+done:
+	*loss = err
+	return y
```
kortschak, line 370:
```suggestion
	return y, err
```

mathext/internal/cephes/hyp2f1.go
```diff
@@ -0,0 +1,443 @@
...
+	f = a
+	g = b
+	h = c
+	s = 1
+	u = 1
+	k = 0
+	for {
+		if math.Abs(h) < 1e-13 {
+			*loss = 1
+			return math.MaxFloat64
```
kortschak, line 397:
```suggestion
			return math.MaxFloat64, 1
```

mathext/internal/cephes/hyp2f1.go
```diff
@@ -0,0 +1,443 @@
...
+		k = math.Abs(u) // remember largest term summed
+		if k > umax {
+			umax = k
+		}
+		k = m
+
+		i++
+		if i > 10000 { // should never happen
+			*loss = 1
+			return s
```
kortschak, line 411:
```suggestion
			return s, 1
```

mathext/internal/cephes/hyp2f1.go
```diff
@@ -0,0 +1,443 @@
...
+		}
+
+		if math.Abs(u/s) <= machEp {
+			break
+		}
+	}
+
+	*loss = machEp*umax/math.Abs(s) + machEp*float64(i)
+
+	return s
```
kortschak, line 421:
```suggestion
	return s, machEp*umax/math.Abs(s) + machEp*float64(i)
```

mathext/internal/cephes/psi.go
```diff
@@ -0,0 +1,146 @@
+// Copyright ©2025 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+/*
+Cephes Math Library Release 2.8:  June, 2000
+Copyright 1984, 1987, 1992, 2000 by Stephen L. Moshier
+*/
+
```
kortschak, line 9:
```suggestion

// Code derived from psi.c in https://www.moshier.net/cephes-math-28.tar.gz.
```

mathext/internal/cephes/psi.go
```diff
@@ -0,0 +1,146 @@
...
+package cephes
+
+import "math"
+
+var A []float64 = []float64{
```
kortschak, line 14: Does this need to be exported?

mathext/internal/cephes/psi.go
```diff
@@ -0,0 +1,146 @@
...
+	var p float64
+	var q float64
+	var nz float64
+	var s float64
+	var w float64
+	var y float64
+	var z float64
+	var i int
+	var n int
+	var negative int
```
kortschak, line 86:
```suggestion
	var (
		p, q, nz, s, w, y, z float64
		i, n                 int
		negative             bool
	)
```

mathext/internal/cephes/psi.go
```diff
@@ -0,0 +1,146 @@
...
+
+	negative = 0
+	nz = 0
+
```
kortschak, line 90:
```suggestion
```

mathext/internal/cephes/psi.go
```diff
@@ -0,0 +1,146 @@
...
+	if x <= 0 {
+		negative = 1
```
kortschak, line 92:
```suggestion
		negative = true
```

mathext/internal/cephes/psi.go
```diff
@@ -0,0 +1,146 @@
...
+		q = x
+		p = math.Floor(q)
+		if p == q {
+			panic(errParamFunctionSingularity)
```
kortschak, line 96: When can this happen? Do we need to warn client code, or is this an indication of a programmer error?

mathext/internal/cephes/psi.go
```diff
@@ -0,0 +1,146 @@
...
+		z = 1 / (s * s)
+		y = z * polevl(z, A, 6)
+	} else {
+		y = 0
+	}
+
+	y = math.Log(s) - 0.5/s - y - w
+
+done:
+	if negative != 0 {
```
kortschak, line 142:
```suggestion
	if negative {
```

mathext/hypergeo.go
```diff
@@ -0,0 +1,204 @@
...
+		if xk < 0 {
+			sign *= -1
+		}
+	}
+	return y, sign
+}
+
+func isInt(x float64) bool {
+	return scalar.EqualWithinAbs(math.Round(x), x, 1e-6)
+}
```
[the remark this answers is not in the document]
fumin: This has been fixed to be exact.

mathext/hypergeo.go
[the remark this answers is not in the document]
fumin: Yes, these hypergeometric relations are very complicated indeed.
I'd also like more eyes to scrutinize them, despite taking pains to double check them myself.

mathext/hypergeo.go
```diff
@@ -0,0 +1,204 @@
...
+
+package mathext
+
+import (
+	"math"
+
+	"gonum.org/v1/gonum/floats/scalar"
+	"gonum.org/v1/gonum/mathext/internal/cephes"
+)
+
```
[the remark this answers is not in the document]
fumin: Thanks for this suggestion, I have added said comment.

mathext/hypergeo_test.go
```diff
@@ -0,0 +1,305 @@
...
+		{a: -100, b: -200, c: -300 + 1e-9, z: 0.5 * math.Sqrt(2), want: 2.653635302903707e-31},
+		{a: -100, b: -200, c: 300 + 1e-9, z: -0.5 * math.Sqrt(2), want: -1.073671875630690e-30},
+		{a: 300, b: 10, c: 5, z: 0.5, want: 3.912238919961547e98},
+		{a: 5, b: -300, c: 10, z: 0.5, want: 1.661006238211309e-7},
+		{a: 2.25, b: 3.75, c: -0.5, z: -1, want: -0.631220676949703},
+	}
+	for i, test := range tests {
+		t.Run(fmt.Sprintf("%d", i), func(t *testing.T) {
+			y := Hypergeo(test.a, test.b, test.c, test.z)
+			if !scalar.EqualWithinRel(y, test.want, 1e-10) {
```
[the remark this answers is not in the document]
fumin: Thanks for pointing this important issue out, I have added `tol` to all tests herein.
The criteria for choosing `tol` is:

* For tests from scipy, follow `tol` therein
* For tests from mpmath, follow `tol` therein

* For tests from John Pearson, Computation of Hypergeometric Functions, `tol=1e-15`. 
`tol=1e-15` is what is given as parameter to the methods Taylor(a), Taylor(b), Single Fraction, Analytic Cont., and also roughly the tightest these methods plus Gauss-Jacobi, RK4 actually achieve. 
The exception is for the case (-1000, -2000, -4000.1, -0.5) where `tol=5e-7`, which I suspect is to due the very large |a| and |c|.
However, note that in this case, only Taylor(a) and Taylor achieve 1e-14, whereas Single Fraction gets `tol=1` (i.e. 0 digits correct), RK4 returns `NaN`, and Analytic Cont. doesn't converge.

* For tests that are made up by me, `tol=5e-15`.

mathext/internal/cephes/hyp2f1.go
```diff
@@ -0,0 +1,443 @@
...
+	var y float64
+	var ax float64
+	var ia float64
+	var ib float64
+	var ic float64
+	var id float64
+	var err float64
+	var flag int
+	var i int
+	var aid int
```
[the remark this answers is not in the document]
fumin: Thanks, this reads much better.

mathext/internal/cephes/psi.go
fumin: This is not needed, sorry for accidentally exporting it. This is made private in the latest commit.

mathext/internal/cephes/hyp2f1.go
```diff
@@ -0,0 +1,408 @@
+// Copyright ©2025 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+/*
+Cephes Math Library Release 2.8:  June, 2000
+Copyright 1984, 1987, 1992, 2000 by Stephen L. Moshier
+*/
+
+// Code derived from hyp2f1.c in https://www.moshier.net/cephes-math-28.tar.gz.
```
kortschak, line 10: Oh, sorry, I should have put a blank line below. This comment should be floating, not attached to the package decl.

mathext/internal/cephes/psi.go
```diff
@@ -0,0 +1,140 @@
+// Copyright ©2025 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+/*
+Cephes Math Library Release 2.8:  June, 2000
+Copyright 1984, 1987, 1992, 2000 by Stephen L. Moshier
+*/
+
+// Code derived from psi.c in https://www.moshier.net/cephes-math-28.tar.gz.
```
kortschak, line 10: Same here.

mathext/hypergeo.go
```diff
@@ -0,0 +1,204 @@
...
+	if z >= 1 {
+		return math.NaN()
+	}
+
+	// Special case for z=-1, use equation 15.3.4 in Abramowitz.
+	if z == -1 {
+		return eqn15_3_4(a, b, c, z)
+	}
+
+	// When a or b is a non-positive integer, Hypergeo reduces to the finite polynomial described in equation 15.4.1.
```
kortschak, line 54: For these internal comments, can you wrap them earlier; somewhere around 60-70 would be good. Also in the test files.

kortschak (commented):
Minor comments. This is looking good.

mathext/hypergeo.go
```diff
@@ -0,0 +1,204 @@
...
+		y += math.Log(math.Abs(xk))
+		if xk < 0 {
+			sign *= -1
+		}
+	}
+	return y, sign
+}
+
+func isNonPosInt(x float64) bool {
+	return x <= 0 && (math.Round(x) == x)
```
kortschak, line 203:
```suggestion
	_, frac := math.Modf(x)
	return x <= 0 && frac == 0
```

mathext/internal/cephes/psi.go
fumin: This is can happen when `psi` (which is equivalent to `mathext.Digamma`) is evaluated on negative integers, which is where the function is mathematically undefined. This is often an indication of a programmer error.

In fact, Cephes hit this bug, which we had to workaround here in `isCephesBug`.

For reference, here is a comparison between our behaviour here and those in Cephes and mathext:

* Here: panic
* Cephes: print error message then return MAXFLOAT
* mathext: Digamma returns NaN

I would also like to further note that such discrepancies arise in the Gamma function, too.
Like the Digamma, the Gamma is undefined for negative integers.
In this case, C's math.h returns Inf, whereas Go's math.Gamma returns NaN.
Curiously, Cephes relies on C's Inf weird behaviour to calculate correct values for hyp2f1!
This means when translating to Go, we couldn't use Go's standard library and had to roll our own Gamma.
Funnily, we ended up not needing to do this after we switched to `gammaADivBDivC` which uses math.Lgamma.
Fortuitously, math.Lgamma decided to do the mathematically incorrect thing of returning Inf for negative integers, which is exactly what we need...

For the record, the switch to `gammaADivBDivC` is unrelated to the issue here, and is just to address the case described in https://github.com/scipy/scipy/pull/2734 .

mathext/hypergeo_test.go
```diff
@@ -0,0 +1,305 @@
...
+		{a: -10, b: 900, c: 10.5, z: 0.99, want: 5.5748237303615776e+17},
+		{a: -1, b: 2, c: 1, z: -1, want: 3},
+		{a: 0.5, b: 1 - 270.5, c: 1.5, z: 0.999 * 0.999, want: 0.053963052503373715},
+
+		// Constants based on https://github.com/scipy/scipy/issues/1561
+		{a: 10, b: 5, c: -300.5, z: 0.5, want: -3.85202708152391e32},
+		{a: -290, b: 5, c: -300.5, z: 0.5, want: 26.9076853843542},
+		{a: 10, b: 5, c: -300.5, z: -0.5, want: 1.08800279612752},
+		{a: 10, b: 5, c: 300.5, z: 0.5, want: 1.08796774775660468},
+		{a: 100, b: 5, c: -300.5, z: 0.5, want: -1.07414870340863581e139},
```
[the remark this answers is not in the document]
fumin: Sorry, these are tests made up by me.
However, they are nonetheless inspired by https://github.com/scipy/scipy/issues/1561 .
I have moved them to a separate section, and added comments to clarify their origins.

mathext/hypergeo_test.go
```diff
@@ -0,0 +1,305 @@
...
+		a    float64
+		b    float64
+		c    float64
+		z    float64
+		want float64
+	}{
+		// Constants based on https://github.com/scipy/scipy/blob/main/scipy/special/tests/test_hyp2f1.py
+		{a: 1.3, b: -0.2, c: 0.3, z: -2.1, want: 1.8202169687521206},
+
+		// Constants based on https://github.com/scipy/scipy/blob/main/scipy/special/tests/test_mpmath.py
```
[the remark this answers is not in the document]
fumin: Right, explaining the origins of these wanted values, along with the tolerance is of utmost importance in a numerical library.
I have added comments that fully describe this information in detail.
In summary:

* tests from scipy: copied verbatim
* tests from mpmath: copied verbatim
* tests from J. Pearson paper: copied verbatim
* tests made up by me: Mathematica

mathext/hypergeo.go
```diff
@@ -0,0 +1,222 @@
...
+
+		y += math.Log(math.Abs(xk))
+		if xk < 0 {
+			sign *= -1
+		}
+	}
+	return y, sign
+}
+
+func isCephesBug(a, b, c, z float64) (float64, bool) {
```
kortschak, line 206: I think this needs an explanation. A summary of the [comment](https://github.com/gonum/gonum/pull/2026#discussion_r1976490944) in the PR would be appropriate.

mathext/hypergeo.go
fumin: Right, I have wrapped comments here and in the test files.
When deciding where to wrap, aside from around 60-70, I also tried to wrap when there's a comma.

mathext/hypergeo.go
fumin: Yes, this truly requires a comment. I have added one that explains the bug's origins in Cephes' attempt to evaluate Digamma at negative integers.

mathext/hypergeo.go
```diff
@@ -0,0 +1,231 @@
...
+	return y, sign
+}
+
+// isCephesBug checks if the received arguments hit a Cephes bug.
+// If the bug is hit, isCephesBug returns the remedied value.
+//
+// This bug concerns Cephes' evaluation of mathext.Digamma on negative integers,
+// which is mathematically undefined.
+// For reference, scipy also found this workaround necessary in:
+// https://github.com/scipy/scipy/commit/1dfea85b34e8cbd868f9fc43345898084ba73aab
```
kortschak, line 213:
```suggestion
// isCephesBug checks whether the received arguments hit a Cephes bug,
// occurring when |a| or |b| >> |c|.
// If the bug is hit, isCephesBug returns the remedied value.
//
// This bug concerns Cephes' evaluation of mathext.Digamma on negative integers,
// where it is mathematically undefined.
//
// For reference, scipy also found this workaround necessary in:
// https://github.com/scipy/scipy/commit/1dfea85b34e8cbd868f9fc43345898084ba73aab
```

kortschak (commented):
LGTM after this change, but will wait for @vladimir-ch.

kortschak:
As part of the initial contribution process (only needs to happen once), please send another single-commit PR with the commit message "A+C: add Fumin Wang" adding yourself to AUTHORS and CONTRIBUTORS.

mathext/hypergeo.go
fumin: Thanks for the rewording, reads much better.
I have made a small modification to make the criteria where this bug manifests more accurate.

fumin:
> As part of the initial contribution process (only needs to happen once), please send another single-commit PR with the commit message "A+C: add Fumin Wang" (I hope I have your name correct here) adding yourself to AUTHORS and CONTRIBUTORS.

I have sent https://github.com/gonum/gonum/pull/2028 accordingly.
It has been a terrific experience working on this PR, and I have learnt so much, especially in numerical computing, along the way.
Thanks a lot for the reviews!

kortschak:
Test failure is unrelated. I cannot reproduce this locally, but investigating.

fumin:
> Test failure is unrelated. I cannot reproduce this locally, but investigating.

I haven't identified the cause yet, but I have pinpointed the exact case where this is failing.
It is failing for the case `trans=false na=2 nw=1 extra=1`.

Strangely, running `go version = go version go1.23.2 linux/amd64` on Ubuntu, I get no failures:

```
trans=false na=2 nw=1 extra=1: cb[1]=(0.07786144889974436+0i) mx[1]=(0.07786144889974445+0i)
```

These values are deterministic, since the test fixed the random seed.
The wanted value `cb` is exactly identical as in the error logs in Appendix 1, but for some reason the computed value `mx` is a bit off in the CI build.

I will be posting to here if I have any further findings.


Appendix 1: Error logs from CI build
```
=== NAME  TestDlaln2
    dlaln2.go:148: Case Case trans=false, na=2, nw=1, extra=1: unexpected value of left-hand side at row 1 with scale=1. Want (0.07786144889974436+0i), got (0.07786144886631519+0i)
--- FAIL: TestDlaln2 (0.16s)
```

Appendix 2: Command to run test locally
```
go test gonum.org/v1/gonum/lapack/gonum -tags=safe -tags=noasm -tags=bounds -run="Dlaln2" -v
```

fumin:
The flaky test we encountered here has been fixed in https://github.com/gonum/gonum/pull/2039 .
I think all comments have been addressed, I wonder could anyone give this pull request another look? Thanks.

kortschak:
Sorry about the delay. We will get to this as soon as we can.

vladimir-ch:
Sorry, I missed the mention, I'll try to take a look this week.

fumin:
@vladimir-ch I wonder if you have some time to help take a look at this pull request? Thank you and sorry for the ping.

fumin:
@vladimir-ch appreciate your expertise and time, no pressure.  About the question

> is mathext/hypergeo.go original code?

The code itself is original, but the ideas behind are based on scipy and elsewhere, which are documented in detail.

One reason why mathext/hypergeo.go is not translated from scipy is because scipy's version of [cephes/hyp2f1](https://github.com/scipy/xsf/blob/v0.1.1/include/xsf/cephes/hyp2f1.h) deviates *greatly* from the original version.
The reason why scipy's hyp2f1.h deviates so much from the original version is because scipy discovered over the past decade more than a dozen [cases](https://github.com/scipy/scipy/issues/1561) where the original cephes/hyp2f1.h gives inaccurate results.
In contrast, this code opted for the strategy of retaining the original cephes/hyp2f1.h , but doing the custom enhancements (namely analytic continuation) in mathext/hypergeo.go.
The reason is because the original cephes/hyp2f1.h series expansion is accurate enough when `|a|`, `|b|`, `|c|` are small. What one only needs to do outside of this range and a few other pathological cases, is to use various hypergeometric identities to convert input arguments into this range.

Another reason why this code is not a translation is because even to this day, there remain accuracy [issues](https://github.com/scipy/scipy/issues/1561) with scipy and other libraries out there.
For example, `scipy.special.hyp2f1(10, 5, -300.5, -0.5)` gives `91037876` whereas the correct answer is `1.088`.
However, it seems that there's consensus that the gold standard is Mathematica, according to papers (Pearson et al., more details in the documentation), scipy, [mpmath](https://github.com/mpmath/mpmath), and elsewhere.
Alas, Mathematica is closed source, though I suspect it may be doing `float256` or even `float1024`.
The approach taken here is to at least pass all tests from R's [hypergeo](https://github.com/cran/hypergeo/tree/master), scipy, and mpmath.

kortschak (approved):
Thank you

PR #2027, opened 2025-02-27
graph/iterator: reflect the swiss map iterator shape

kortschak:
This is a safety change to ensure we are allocating the correct size for the iterator.

Please take a look.


vladimir-ch (approved):
Just approving, I trust you on this :-)

kortschak:
I thought I pushed the U1000 comment here, but I sent it to my branch. PTAL

vladimir-ch (approved):

PR #2029, opened 2025-03-06
refactor: use slices.Equal to simplify code

tiaoxizhan:
Please take a look.

In the Go 1.21 standard library, a new function has been introduced that enhances code conciseness and readability. It can be find  [here](https://pkg.go.dev/slices@go1.21.1#Equal).

### DIFF
--- a/mat/eigen_test.go
+++ b/mat/eigen_test.go
@@ -7,6 +7,7 @@ package mat
 import (
 	"math"
 	"math/rand/v2"
+	"slices"
 	"sort"
 	"testing"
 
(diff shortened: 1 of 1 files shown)


kortschak:
I'm not sure we need to do this.

tiaoxizhan:
> I'm not sure we need to do this.

I see that the current version is go1.23, which can use slices.Equal, which can indeed greatly simplify the code.

But I follow your advice.

kortschak:
Thank you for your contribution, but I don't think we will merge this.

PR #2032, opened 2025-03-12
lapack: fix eigenvector normalization residual calculation

kevincatty:
Please take a look.

In the residualEVNormalization function, there was an error in the final return statement where both arguments to math.Max used enrmin . The correct calculation should compare deviations of both minimum and maximum norms from 1.


kortschak (approved):
LGTM, but I'd like to have confirmation of intention from @vladimir-ch.

vladimir-ch:
LGTM. It was an oversight on my side while porting code from the reference. Thanks.

vladimir-ch (approved):

PR #2034, opened 2025-03-12
fix: repeat args from sub-func call

alingse:
```
return math.Max(math.Abs(enrmin-1), math.Abs(enrmin-1))
```

is a copyandpaste bug


sbinet (approved):
LGTM

@kortschak @vladimir-ch ?

alingse:
> I'm not a fan of being a research subject. There was a similar pr by another person. TBH, I'd prefer that these were issues rather than PRs.

🐶 this is a little bug, I change the description or PR

alingse:
@kortschak thx, I forget to search the exists PR I will close this

PR #2036, opened 2025-03-18
stat: add an example to compute a confidence interval

sbinet:
This CL adds an example showing how to compute the confidence interval of the mean for a small sample.

Please take a look.

### DIFF
--- a/stat/stat_example_test.go
+++ b/stat/stat_example_test.go
@@ -46,14 +46,14 @@ func ExampleLinearRegression() {
 	// R^2: 0.999999
 }
 
-// Example_confidenceInterval shows how one can compute a confidence
-// interval to quantify the uncertainty around an estimated parameter,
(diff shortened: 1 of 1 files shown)


stat/stat_example_test.go
```diff
@@ -43,3 +45,54 @@ func ExampleLinearRegression() {
...
+// where:
+//   - x is the sample mean,
+//   - s is the sample standard deviation,
+//   - n is the sample size, and
+//   - t is the critical value from the t-distribution based on the desired
+//     confidence level and degrees of freedom (df=n-1)
+//
+// For more details, see:
+//
+//	https://en.wikipedia.org/wiki/Student's_t-distribution
```
kortschak, line 67: Inside the func?

kortschak (approved):
Thanks

PR #2038, opened 2025-03-25
graph/network: add eccentricity measurement

n-peugnet:
Please take a look.

Closes: #2037

PR #2041, opened 2025-03-26
graph/network: add test for eccentricity with digraphs

n-peugnet:
This is a follow-up of https://github.com/gonum/gonum/pull/2038

Please take a look.


kortschak (approved):
Thanks

PR #2046, opened 2025-04-12
stat: implement Wasserstein distance calculation

Shieldine:
This PR focuses on implementing the Wasserstein Distance (also known as Earth Mover's Distance) for both 1-d and n-d distributions. The implementation was heavily inspired by Scipy - some test cases were taken from there.

The functions include:
- optional weights: if nil, uniform weights are used
- weight normalization
- extensive validity checks
- comments to document and explain the steps taken
- tests

This is my first contribution to Gonum.
I apologize for any rules I may have missed and would welcome suggestions. Some brackets in the stat package were removed as a result to an auto-refactor.

### DIFF
--- a/stat/stat.go
+++ b/stat/stat.go
@@ -30,7 +30,6 @@ const (
 // between two 1D distributions p and q with optional weights. p and q are
 // the support points of each distribution. pWeights and qWeights are the weights
 // for each point.
-// The lengths of p and q don't have to be equal.
 // If a weights slice is nil, then uniform weights are used. If it is not nil,
 // then the length of the weights slice must equal the length of the corresponding points.
 // Otherwise, the function will panic.
@@ -167,10 +166,9 @@ func WassersteinDistance(p, q, pWeights, qWeights []float64) float64 {
 // between two n-dimensional distributions p and q with optional weights. p and q are
 // matrices where each row represents a point. pWeights and qWeights are the weights for each point.
 // The number of columns in p and q must be equal.
-// The number of points (= rows) in p and q doesn't have to be equal.
 // If a weights slice is nil, then uniform weights are used. If it is not nil, then the
 // length of the weights slice must equal the number of rows in the corresponding matrix.
-// The function will panic if the described conditions are not met.
+// WassersteinDistanceND will panic if these conditions are not met.
 //
 // This implementation uses linear programming to solve the optimal transport problem.
 // tol controls the solver's tolerance. See lp.Simplex for more information on this.
--- a/stat/stat.go
+++ b/stat/stat.go
@@ -26,14 +26,14 @@ const (
 	LinInterp CumulantKind = 4
 )
 
-const epsilon = 1e-8
-
 // WassersteinDistance computes the Wasserstein distance (Earth Mover's Distance)
-// between two 1D distributions p and q with optional weights.
-//
-// Parameters:
-// p and q are slices of values representing the support points of each distribution.
-// pWeights and qWeights are optional weights for each point. If nil, uniform weights are used.
+// between two 1D distributions p and q with optional weights. p and q are
+// the support points of each distribution. pWeights and qWeights are the weights
+// for each point.
+// The lengths of p and q don't have to be equal.
+// If a weights slice is nil, then uniform weights are used. If it is not nil,
+// then the length of the weights slice must equal the length of the corresponding points.
+// Otherwise, the function will panic.
 //
 // The function returns the 1-Wasserstein distance (L1 metric).
 // This implementation uses the CDF-based algorithm for the 1D case.
@@ -164,14 +164,17 @@ func WassersteinDistance(p, q, pWeights, qWeights []float64) float64 {
 }
 
 // WassersteinDistanceND computes the Wasserstein distance (Earth Mover's Distance)
-// between two n-dimensional distributions p and q with optional weights.
-//
-// Parameters:
-// p and q are matrices where each row represents a point with coordinates as columns.
-// pWeights and qWeights are optional weights for each point. If nil, uniform weights are used.
+// between two n-dimensional distributions p and q with optional weights. p and q are
+// matrices where each row represents a point. pWeights and qWeights are the weights for each point.
+// The number of columns in p and q must be equal.
+// The number of points (= rows) in p and q doesn't have to be equal.
+// If a weights slice is nil, then uniform weights are used. If it is not nil, then the
+// length of the weights slice must equal the number of rows in the corresponding matrix.
+// The function will panic if the described conditions are not met.
 //
 // This implementation uses linear programming to solve the optimal transport problem.
-func WassersteinDistanceND(p, q mat.Matrix, pWeights, qWeights []float64) (float64, error) {
+// tol controls the solver's tolerance. See lp.Simplex for more information on this.
+func WassersteinDistanceND(p, q mat.Matrix, pWeights, qWeights []float64, tol float64) (float64, error) {
 	pRows, pCols := p.Dims()
 	qRows, qCols := q.Dims()
 
@@ -246,7 +249,7 @@ func WassersteinDistanceND(p, q mat.Matrix, pWeights, qWeights []float64) (float
 		}
 	}
 
-	return solveOptimalTransportLP(cost, pWeights, qWeights)
+	return solveOptimalTransportLP(cost, pWeights, qWeights, tol)
 }
 
 // normalizeWeights checks and normalizes the provided weight array if needed.
@@ -264,7 +267,7 @@ func normalizeWeights(weights []float64) {
 }
 
 // solveOptimalTransportLP solves the optimal transport problem.
-func solveOptimalTransportLP(costMatrix *mat.Dense, supply, demand []float64) (float64, error) {
+func solveOptimalTransportLP(costMatrix *mat.Dense, supply, demand []float64, tol float64) (float64, error) {
 	rows, cols := costMatrix.Dims()
 
 	// Formulate the linear program.
@@ -291,7 +294,7 @@ func solveOptimalTransportLP(costMatrix *mat.Dense, supply, demand []float64) (f
 	copy(b[rows-1:], demand)
 
 	// Solve the linear program.
-	optVal, _, err := lp.Simplex(c, constraints, b, epsilon, nil)
+	optVal, _, err := lp.Simplex(c, constraints, b, tol, nil)
 	return optVal, err
 }
 
--- a/stat/stat.go
+++ b/stat/stat.go
@@ -5,11 +5,11 @@
 package stat
 
 import (
-	"fmt"
-	"gonum.org/v1/gonum/mat"
-	"gonum.org/v1/gonum/optimize/convex/lp"
 	"math"
 	"sort"
+
+	"gonum.org/v1/gonum/mat"
+	"gonum.org/v1/gonum/optimize/convex/lp"
 
 	"gonum.org/v1/gonum/floats"
 )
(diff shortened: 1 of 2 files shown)


stat/stat.go
```diff
@@ -24,6 +27,285 @@ const (
...
+
+	if totalP <= 0 || totalQ <= 0 {
+		panic("stat: weights must sum to positive values")
+	}
+
+	floats.Scale(1.0/totalP, pWeights)
+	floats.Scale(1.0/totalQ, qWeights)
+}
+
+// solveOptimalTransportLP solves the optimal transport problem using Gonum's linear programming solver.
```
kortschak, line 267:
```suggestion
// solveOptimalTransportLP solves the optimal transport problem.
```

stat/stat.go
```diff
@@ -24,6 +27,285 @@ const (
...
+
+	// Formulate the linear program
+	c := make([]float64, rows*cols)
+	for i := 0; i < rows; i++ {
+		for j := 0; j < cols; j++ {
+			c[i*cols+j] = costMatrix.At(i, j)
+		}
+	}
+
+	constraintMatrix := mat.NewDense(rows+cols-1, rows*cols, nil)
```
kortschak, line 279:
```suggestion
	constraints := mat.NewDense(rows+cols-1, rows*cols, nil)
```

stat/stat.go
```diff
@@ -24,6 +27,285 @@ const (
...
+	copy(b[:rows-1], supply[:rows-1])
+	copy(b[rows-1:], demand)
+
+	// Solve the linear program
+	optVal, _, err := lp.Simplex(c, constraintMatrix, b, epsilon, nil)
+	if err != nil {
+		panic(fmt.Sprintf("stat: failed to solve optimal transport problem: %v", err))
+	}
+
+	return optVal
```
kortschak, line 306:
```suggestion
	optVal, _, err := lp.Simplex(c, constraintMatrix, b, epsilon, nil)
	return optVal, err
```

stat/stat.go
```diff
@@ -24,6 +27,285 @@ const (
...
+//   - pWeights, qWeights: Optional weights for each point. If nil, uniform weights are used
+//
+// The function returns the 1-Wasserstein distance (L1 metric).
+// This implementation uses the CDF-based algorithm for the 1D case.
+func WassersteinDistance(p, q, pWeights, qWeights []float64) float64 {
+	if len(p) == 0 || len(q) == 0 {
+		panic("stat: input distributions cannot be empty")
+	}
+
+	// Handle identical distributions case
```
kortschak, line 46: Please make all doc sentences conform to English punctuation rules.

kortschak (commented):
Please revert simplification commit.

stat/stat.go
```diff
@@ -24,6 +27,285 @@ const (
...
+		qCDFAtX1 := 0.0
+		for j := 0; j < len(qPairs); j++ {
+			if qPairs[j].value <= x1 {
+				qCDFAtX1 = qCDF[j]
+			} else {
+				break
+			}
+		}
+
+		distance += math.Abs(pCDFAtX1-qCDFAtX1) * (x2 - x1)
```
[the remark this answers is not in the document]
Shieldine: Is there a reason why this explicit conversion is needed? All the variables used here are of type float64.

stat/stat.go
Shieldine: How would you suggest dealing with the error in the main function? Simply returning NaN, or also returning an error?

stat/stat.go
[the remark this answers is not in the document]
kortschak: This is to ensure that arm64 and amd64 give the same result.

stat/stat.go
kortschak: The error should be propagated.

stat/stat.go
```diff
@@ -24,6 +27,285 @@ const (
...
+
+// WassersteinDistanceND computes the Wasserstein distance (Earth Mover's Distance)
+// between two n-dimensional distributions p and q with optional weights.
+//
+// Parameters:
+//   - p, q: Slices of points where each point is a slice of float64 coordinates
+//   - pWeights, qWeights: Optional weights for each point. If nil, uniform weights are used
+//
+// This implementation uses linear programming to solve the optimal transport problem.
+func WassersteinDistanceND(p, q [][]float64, pWeights, qWeights []float64) float64 {
```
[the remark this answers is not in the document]
Shieldine: To remain consistent with the 1-d implementation, which also uses slices.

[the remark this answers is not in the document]
kortschak: I don't think that is a worthwhile goal.

Shieldine:
> Please revert simplification commit.

Do you mean simplifying own assignment to floats.AddConst?

The new commit should cover all other suggestions provided. The slices were not changed to matrices as I am not sure if that is necessary. If it is explicitly wanted, please say so and I will change it.

kortschak:
> Do you mean simplifying own assignment to floats.AddConst?

No, [all these changes](https://github.com/gonum/gonum/pull/2046/files#diff-d0efc9b269d6e8ead683db4a4992e541ed2f1aeab43f756bb42c45c0e265a0daL335-R1684).

stat/stat.go
```diff
@@ -8,6 +8,9 @@ import (
 	"math"
 	"sort"
 
+	"gonum.org/v1/gonum/mat"
+	"gonum.org/v1/gonum/optimize/convex/lp"
+
 	"gonum.org/v1/gonum/floats"
```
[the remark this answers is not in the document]
Shieldine: I'm sorry, is there some option I've missed? The order I submitted is what goimports gave me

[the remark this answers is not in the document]
kortschak: It was in two stanzas. `goimports` will respect stanzas that are in place, but all gonum imports should be in one stanza.

Shieldine:
The newest commit should cover all requested changes.

The simplification from before has been reverted.

I'd like to take this opportunity to say thank you for the thorough reviews - in just these few functions, I have learned a lot and realized there is a huge room for improvement in the future.

Shieldine:
Interesting, I cant reproduce the failing test locally. I'm testing on arm architecture - I assume it will be the difference there.

stat/stat.go
```diff
@@ -24,6 +26,272 @@ const (
...
+	if pCols != qCols {
+		panic("stat: point dimensions must match between distributions")
+	}
+
+	// Special case for single-point distributions
+	if pRows == 1 && qRows == 1 {
+		var sumSquares float64
+		for j := 0; j < pCols; j++ {
+			diff := p.At(0, j) - q.At(0, j)
+			sumSquares += diff * diff
```
kortschak, line 195:
```suggestion
			sumSquares += float64(diff * diff)
```

stat/stat.go
```diff
@@ -24,6 +26,272 @@ const (
...
+	// Create cost matrix using Euclidean distance.
+	cost := mat.NewDense(pRows, qRows, nil)
+
+	for i := 0; i < pRows; i++ {
+		for j := 0; j < qRows; j++ {
+			// Calculate Euclidean distance between points
+			var sumSquares float64
+			for k := 0; k < pCols; k++ {
+				diff := p.At(i, k) - q.At(j, k)
+				sumSquares += diff * diff
```
kortschak, line 242:
```suggestion
				sumSquares += float64(diff * diff)
```

kortschak (commented):
Try making these changes and adjusting your test expectations. This issue is that on arm64, FMA and related floating point operations are used while on other architectures they are not. So we have to lower precision on any operation that can be optimised to that; this is any float expression that is `a*b + c`, which `c =+ a*b` is.

stat/stat.go
```diff
@@ -24,6 +26,272 @@ const (
...
+	return solveOptimalTransportLP(cost, pWeights, qWeights)
+}
+
+// normalizeWeights checks and normalizes the provided weight array if needed.
+func normalizeWeights(weights []float64) {
+	total := floats.Sum(weights)
+
+	if total <= 0 {
+		panic("stat: weights must sum to positive values")
+	}
```
kortschak, line 258: This check is not correct; it allows negative weights to exist so long as the total is positive.

Shieldine:
Hey, any updates on this?

stat/stat.go
```diff
@@ -24,6 +26,275 @@ const (
 	LinInterp CumulantKind = 4
 )
 
+const epsilon = 1e-8
```
kortschak, line 29: I think this should probably be a user provided parameter `tol` in `WassersteinDistanceND`.

stat/stat.go
```diff
@@ -24,6 +26,275 @@ const (
...
+
+// WassersteinDistance computes the Wasserstein distance (Earth Mover's Distance)
+// between two 1D distributions p and q with optional weights.
+//
+// Parameters:
+// p and q are slices of values representing the support points of each distribution.
+// pWeights and qWeights are optional weights for each point. If nil, uniform weights are used.
```
kortschak, line 36:
```suggestion
// WassersteinDistance computes the Wasserstein distance (Earth Mover's Distance)
// between two 1D distributions, p and q with optional weights. p and q are are
// the support points of each distribution. pWeights and qWeights are the weights
// for each point.
// The lengths of p and q must be equal. If a weights slice is nil then all of
// the weights are 1. If it is not nil, then the length of the weights slice
// must equal the length of the corresponding points.
```
(similar change below)

stat/stat.go
Shieldine: Maybe, but I think it would be quite annoying for the user to set it every time unless they have a specific value in their minds.

Go does not support default values - do you think using optional parameters like this:

`
func WassersteinDistanceND(p, q mat.Matrix, pWeights, qWeights []float64, tol ...float64) (float64, error) {
	epsilon := 1e-8

	if len(tol) > 0 {
		epsilon = tol[0]
	}
`

is a viable option to keep a standard value for users who are happy with a default while allowing its modification?

kortschak: `tol ...float64`? No. This requires an additional check for length. We don't do that.

If you want to use a default (I'd like to know what the value you've chosen was chosen), then a negative `tol` can be a flag, though I don't think we've done that anywhere else, so would be reluctant to allow this.

Shieldine: alright, a required tolerance it is then

Shieldine:
The updates documentation should be clearer

stat/stat.go
```diff
@@ -24,6 +26,278 @@ const (
 	LinInterp CumulantKind = 4
 )
 
+// WassersteinDistance computes the Wasserstein distance (Earth Mover's Distance)
+// between two 1D distributions p and q with optional weights. p and q are
+// the support points of each distribution. pWeights and qWeights are the weights
+// for each point.
+// The lengths of p and q don't have to be equal.
```
kortschak, line 33: Oh, sorry. If there's no constraint, this can be omitted.

stat/stat.go
```diff
@@ -24,6 +26,278 @@ const (
...
+	}
+
+	return distance
+}
+
+// WassersteinDistanceND computes the Wasserstein distance (Earth Mover's Distance)
+// between two n-dimensional distributions p and q with optional weights. p and q are
+// matrices where each row represents a point. pWeights and qWeights are the weights for each point.
+// The number of columns in p and q must be equal.
+// The number of points (= rows) in p and q doesn't have to be equal.
```
kortschak, line 170: Omit.

stat/stat.go
```diff
@@ -24,6 +26,278 @@ const (
...
+// If a weights slice is nil, then uniform weights are used. If it is not nil, then the
+// length of the weights slice must equal the number of rows in the corresponding matrix.
+// The function will panic if the described conditions are not met.
```
kortschak, line 173:
```suggestion
// WassersteinDistanceND will panic if these conditions are not met.
```

kortschak (commented):
LGTM after these changes.

Can you open another PR adding yourself to AUTHORS and CONTRIBUTORS with the commit message "A+C: add \<your name\>" (see existing entries and the file headers for the requirements).

Shieldine:
The failing test stems from a different package - it seems unrelated to the changes presented here.

I've also opened #2047 as requested

kortschak (approved):
Thanks

PR #2048, opened 2025-04-26
spatial: add Umeyama's algorithm for the estimation of transformation parameters between two point patterns

Shieldine:
This PR focuses on adding Umeyama's algorithm for the estimation of transformation parameters between two point patterns as described in this paper: https://doi.org/10.1109/34.88573 

The algorithm has been placed in an own package alongside other spatial packages.

Checklist:

- API changes have been discussed --- Doesn't apply
- code is goformated correctly (goimports) --- Yes
- packages with generated code have had code generation run --- Doesnt' apply
- tests pass locally --- Yes
- linked to relevant issues --- Doesn't apply

### DIFF
--- a/spatial/umeyama/doc.go
+++ b/spatial/umeyama/doc.go
@@ -1,0 +1,8 @@
+// Copyright ©2025 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Package umeyama provides a function to estimate the similarity transformation parameters between two point patterns
+// as described in "Least-Squares Estimation of Transformation Parameters Between Two Point Patterns"
+// by Shinji Umeyama
+package umeyama // import "gonum.org/v1/gonum/spatial/umeyama"
--- a/spatial/umeyama/umeyama.go
+++ b/spatial/umeyama/umeyama.go
@@ -4,7 +4,6 @@ import (
 	"errors"
 
 	"gonum.org/v1/gonum/mat"
-	"gonum.org/v1/gonum/stat"
 )
 
 const (
@@ -28,8 +27,10 @@ const (
 // Y ≈ c * R * X + t
 //
 // The dimensions of X and Y must be equal. The function will panic if they are not.
-// The points require consistent indexing. This means
-// that point i of X needs to correspond to point i of Y.
+// The points require consistent indexing. This means that point i of X needs to correspond
+// to point i of Y.
+//
+// In this implementation, rows represent points and columns represent dimensions.
 //
 // Umeyama returns the scale factor c, the rotation matrix r and the translation vector t.
 // If a computation fails, Umeyama will return an error.
@@ -42,26 +43,32 @@ func Umeyama(X, Y *mat.Dense) (float64, *mat.Dense, *mat.VecDense, error) {
 		panic("umeyama: dimensions of X and Y do not match")
 	}
 
-	rows, cols := rowsX, colsX
+	n := rowsX // number of points
+	m := colsX // number of dimensions
 
 	// Calculate means.
-	muX := mat.NewVecDense(rows, nil)
-	muY := mat.NewVecDense(rows, nil)
+	muX := mat.NewVecDense(m, nil)
+	muY := mat.NewVecDense(m, nil)
 
-	for i := 0; i < rows; i++ {
-		muX.SetVec(i, stat.Mean(mat.Row(nil, i, X), nil))
-		muY.SetVec(i, stat.Mean(mat.Row(nil, i, Y), nil))
+	for i := 0; i < m; i++ {
+		var sumX, sumY float64
+		for j := 0; j < n; j++ {
+			sumX += X.At(j, i)
+			sumY += Y.At(j, i)
+		}
+		muX.SetVec(i, sumX/float64(n))
+		muY.SetVec(i, sumY/float64(n))
 	}
 
 	// Calculate variance of X.
 	var varX float64
-	for j := 0; j < cols; j++ {
-		for i := 0; i < rows; i++ {
-			diff := X.At(i, j) - muX.AtVec(i)
-			varX += float64(diff) * float64(diff)
+	for i := 0; i < n; i++ {
+		for j := 0; j < m; j++ {
+			diff := X.At(i, j) - muX.AtVec(j)
+			varX += diff * diff
 		}
 	}
-	varX /= float64(cols)
+	varX /= float64(n)
 
 	// Check for degenerate case.
 	if varX < 1e-10 {
@@ -69,13 +76,13 @@ func Umeyama(X, Y *mat.Dense) (float64, *mat.Dense, *mat.VecDense, error) {
 	}
 
 	// Calculate covariance matrix.
-	covXY := mat.NewDense(rows, rows, nil)
-	for j := 0; j < cols; j++ {
-		for i := 0; i < rows; i++ {
-			diffY := Y.At(i, j) - muY.AtVec(i)
-			for k := 0; k < rows; k++ {
-				diffX := X.At(k, j) - muX.AtVec(k)
-				covXY.Set(i, k, covXY.At(i, k)+diffY*diffX/float64(cols))
+	covXY := mat.NewDense(m, m, nil)
+	for i := 0; i < n; i++ {
+		for j := 0; j < m; j++ {
+			diffY := Y.At(i, j) - muY.AtVec(j)
+			for k := 0; k < m; k++ {
+				diffX := X.At(i, k) - muX.AtVec(k)
+				covXY.Set(j, k, covXY.At(j, k)+diffY*diffX/float64(n))
 			}
 		}
 	}
@@ -87,42 +94,42 @@ func Umeyama(X, Y *mat.Dense) (float64, *mat.Dense, *mat.VecDense, error) {
 	}
 
 	// Get U and V.
-	u := mat.NewDense(rows, rows, nil)
-	v := mat.NewDense(rows, rows, nil)
+	u := mat.NewDense(m, m, nil)
+	v := mat.NewDense(m, m, nil)
 	svd.UTo(u)
 	svd.VTo(v)
 
 	// Create S matrix (identity matrix).
-	s := mat.NewDiagDense(rows, nil)
-	for i := 0; i < rows; i++ {
+	s := mat.NewDiagDense(m, nil)
+	for i := 0; i < m; i++ {
 		s.SetDiag(i, 1.0)
 	}
 
 	// Check determinants to ensure proper rotation matrix (not reflection).
 	if mat.Det(u)*mat.Det(v) < 0 {
-		s.SetDiag(rows-1, -1.0)
+		s.SetDiag(m-1, -1.0)
 	}
 
 	// Calculate scale factor c.
 	var c float64
 	singularValues := svd.Values(nil)
-	for i := 0; i < rows; i++ {
+	for i := 0; i < m; i++ {
 		c += singularValues[i] * s.At(i, i)
 	}
 	c /= varX
 
 	// Calculate rotation matrix R.
-	r := mat.NewDense(rows, rows, nil)
-	tmp := mat.NewDense(rows, rows, nil)
+	r := mat.NewDense(m, m, nil)
+	tmp := mat.NewDense(m, m, nil)
 	tmp.Mul(u, s)
 	r.Mul(tmp, v.T())
 
 	// Calculate translation vector t.
-	t := mat.NewVecDense(rows, nil)
-	rMuX := mat.NewVecDense(rows, nil)
+	t := mat.NewVecDense(m, nil)
+	rMuX := mat.NewVecDense(m, nil)
 	rMuX.MulVec(r, muX)
 
-	for i := 0; i < rows; i++ {
+	for i := 0; i < m; i++ {
 		t.SetVec(i, muY.AtVec(i)-c*rMuX.AtVec(i))
 	}
 
--- a/spatial/umeyama/umeyama.go
+++ b/spatial/umeyama/umeyama.go
@@ -2,6 +2,7 @@ package umeyama
 
 import (
 	"errors"
+
 	"gonum.org/v1/gonum/mat"
 	"gonum.org/v1/gonum/stat"
 )
--- a/spatial/umeyama/umeyama.go
+++ b/spatial/umeyama/umeyama.go
@@ -9,9 +9,9 @@ import (
 
 const (
 	// ErrMsgSVDFailed is the error message for SVD factorization failure
-	ErrMsgSVDFailed = "SVD factorization failed"
+	ErrMsgSVDFailed = "umeyama: SVD factorization failed"
 	// ErrMsgDegenerateInput is the error message for degenerate input data
-	ErrMsgDegenerateInput = "degenerate input: variance of X is too small"
+	ErrMsgDegenerateInput = "umeyama: variance of X is too small"
 )
 
 // Umeyama estimates the similarity transformation parameters between two matrices X and Y.
@@ -19,7 +19,7 @@ const (
 // This is an implementation of the algorithm presented in:
 // "Least-Squares Estimation of Transformation Parameters Between Two Point Patterns"
 // by Shinji Umeyama, IEEE Transactions on Pattern Analysis and Machine Intelligence,
-// Vol. 13, No. 4, April 1991.
+// Vol. 13, No. 4, April 1991, which can be found here: https://doi.org/10.1109/34.88573
 //
 // The algorithm finds the optimal similarity transformation [c, R, t] ∈ Sim(m)
 // that minimizes the mean squared error between two point patterns.
@@ -31,7 +31,7 @@ const (
 // The points require consistent indexing. This means
 // that point i of X needs to correspond to point i of Y.
 //
-// Umeyama returns the scale factor c, the rotation matrix R and the translation vector t.
+// Umeyama returns the scale factor c, the rotation matrix r and the translation vector t.
 // If a computation fails, Umeyama will return an error.
 func Umeyama(X, Y *mat.Dense) (float64, *mat.Dense, *mat.VecDense, error) {
 	rowsX, colsX := X.Dims()
--- a/spatial/umeyama/umeyama.go
+++ b/spatial/umeyama/umeyama.go
@@ -54,11 +54,11 @@ func Umeyama(X, Y *mat.Dense) (float64, *mat.Dense, *mat.VecDense, error) {
 	}
 
 	// Calculate variance of X.
-	varX := 0.0
+	var varX float64
 	for j := 0; j < cols; j++ {
 		for i := 0; i < rows; i++ {
 			diff := X.At(i, j) - muX.AtVec(i)
-			varX += diff * diff
+			varX += float64(diff) * float64(diff)
 		}
 	}
 	varX /= float64(cols)
@@ -82,20 +82,15 @@ func Umeyama(X, Y *mat.Dense) (float64, *mat.Dense, *mat.VecDense, error) {
 
 	// Singular Value Decomposition
 	var svd mat.SVD
-	ok := svd.Factorize(covXY, mat.SVDFull)
-	if !ok {
+	if !svd.Factorize(covXY, mat.SVDFull) {
 		return 0, nil, nil, errors.New(ErrMsgSVDFailed)
 	}
 
-	// Get U, Σ, and V.
+	// Get U and V.
 	u := mat.NewDense(rows, rows, nil)
-	vt := mat.NewDense(rows, rows, nil)
+	v := mat.NewDense(rows, rows, nil)
 	svd.UTo(u)
-	svd.VTo(vt)
-
-	// Transpose V to get VH.
-	vh := mat.NewDense(rows, rows, nil)
-	vh.Copy(vt.T())
+	svd.VTo(v)
 
 	// Create S matrix (identity matrix).
 	s := mat.NewDiagDense(rows, nil)
@@ -104,15 +99,12 @@ func Umeyama(X, Y *mat.Dense) (float64, *mat.Dense, *mat.VecDense, error) {
 	}
 
 	// Check determinants to ensure proper rotation matrix (not reflection).
-	// This is a key contribution of Umeyama's paper - ensuring a proper rotation.
-	uDet := mat.Det(u)
-	vhDet := mat.Det(vh)
-	if uDet*vhDet < 0 {
+	if mat.Det(u)*mat.Det(v) < 0 {
 		s.SetDiag(rows-1, -1.0)
 	}
 
 	// Calculate scale factor c.
-	c := 0.0
+	var c float64
 	singularValues := svd.Values(nil)
 	for i := 0; i < rows; i++ {
 		c += singularValues[i] * s.At(i, i)
@@ -120,23 +112,15 @@ func Umeyama(X, Y *mat.Dense) (float64, *mat.Dense, *mat.VecDense, error) {
 	c /= varX
 
 	// Calculate rotation matrix R.
-	tmp := mat.NewDense(rows, rows, nil)
 	r := mat.NewDense(rows, rows, nil)
+	tmp := mat.NewDense(rows, rows, nil)
 	tmp.Mul(u, s)
-	r.Mul(tmp, vh)
+	r.Mul(tmp, v.T())
 
 	// Calculate translation vector t.
 	t := mat.NewVecDense(rows, nil)
 	rMuX := mat.NewVecDense(rows, nil)
-	tmp2 := mat.NewDense(rows, 1, nil)
-	for i := 0; i < rows; i++ {
-		for j := 0; j < rows; j++ {
-			tmp2.Set(i, 0, tmp2.At(i, 0)+r.At(i, j)*muX.AtVec(j))
-		}
-	}
-	for i := 0; i < rows; i++ {
-		rMuX.SetVec(i, tmp2.At(i, 0))
-	}
+	rMuX.MulVec(r, muX)
 
 	for i := 0; i < rows; i++ {
 		t.SetVec(i, muY.AtVec(i)-c*rMuX.AtVec(i))
--- a/spatial/umeyama/umeyama.go
+++ b/spatial/umeyama/umeyama.go
@@ -4,12 +4,13 @@ import (
 	"errors"
 
 	"gonum.org/v1/gonum/mat"
+	"gonum.org/v1/gonum/stat"
 )
 
 const (
-	// ErrMsgSVDFailed is the error message for SVD factorization failure
+	// ErrMsgSVDFailed is the error message for SVD factorization failure.
 	ErrMsgSVDFailed = "umeyama: SVD factorization failed"
-	// ErrMsgDegenerateInput is the error message for degenerate input data
+	// ErrMsgDegenerateInput is the error message for degenerate input data.
 	ErrMsgDegenerateInput = "umeyama: variance of X is too small"
 )
 
@@ -50,42 +51,41 @@ func Umeyama(X, Y *mat.Dense) (float64, *mat.Dense, *mat.VecDense, error) {
 	muX := mat.NewVecDense(m, nil)
 	muY := mat.NewVecDense(m, nil)
 
-	for i := 0; i < m; i++ {
-		var sumX, sumY float64
-		for j := 0; j < n; j++ {
-			sumX += X.At(j, i)
-			sumY += Y.At(j, i)
-		}
-		muX.SetVec(i, sumX/float64(n))
-		muY.SetVec(i, sumY/float64(n))
+	colX := make([]float64, n)
+	colY := make([]float64, n)
+
+	for j := 0; j < m; j++ {
+		mat.Col(colX, j, X)
+		mat.Col(colY, j, Y)
+		muX.SetVec(j, stat.Mean(colX, nil))
+		muY.SetVec(j, stat.Mean(colY, nil))
 	}
 
-	// Calculate variance of X.
+	// Center the matrices and calculate variance of X.
 	var varX float64
+	Xc := mat.NewDense(n, m, nil)
+	Yc := mat.NewDense(n, m, nil)
+
 	for i := 0; i < n; i++ {
 		for j := 0; j < m; j++ {
-			diff := X.At(i, j) - muX.AtVec(j)
-			varX += diff * diff
+			Xc.Set(i, j, X.At(i, j)-muX.AtVec(j))
+			Yc.Set(i, j, Y.At(i, j)-muY.AtVec(j))
+
+			varX += Xc.At(i, j) * Xc.At(i, j)
 		}
 	}
 	varX /= float64(n)
 
-	// Check for degenerate case.
+	// Check for degenerate case. This prevents cases of division by zero and mathematical instability due to
+	// very low variance.
 	if varX < 1e-10 {
 		return 0, nil, nil, errors.New(ErrMsgDegenerateInput)
 	}
 
 	// Calculate covariance matrix.
 	covXY := mat.NewDense(m, m, nil)
-	for i := 0; i < n; i++ {
-		for j := 0; j < m; j++ {
-			diffY := Y.At(i, j) - muY.AtVec(j)
-			for k := 0; k < m; k++ {
-				diffX := X.At(i, k) - muX.AtVec(k)
-				covXY.Set(j, k, covXY.At(j, k)+diffY*diffX/float64(n))
-			}
-		}
-	}
+	covXY.Mul(Yc.T(), Xc)
+	covXY.Scale(1/float64(n), covXY)
 
 	// Singular Value Decomposition
 	var svd mat.SVD
@@ -99,15 +99,15 @@ func Umeyama(X, Y *mat.Dense) (float64, *mat.Dense, *mat.VecDense, error) {
 	svd.UTo(u)
 	svd.VTo(v)
 
-	// Create S matrix (identity matrix).
+	// Create identity matrix.
 	s := mat.NewDiagDense(m, nil)
 	for i := 0; i < m; i++ {
-		s.SetDiag(i, 1.0)
+		s.SetDiag(i, 1)
 	}
 
 	// Check determinants to ensure proper rotation matrix (not reflection).
 	if mat.Det(u)*mat.Det(v) < 0 {
-		s.SetDiag(m-1, -1.0)
+		s.SetDiag(m-1, -1)
 	}
 
 	// Calculate scale factor c.
@@ -120,18 +120,15 @@ func Umeyama(X, Y *mat.Dense) (float64, *mat.Dense, *mat.VecDense, error) {
 
 	// Calculate rotation matrix R.
 	r := mat.NewDense(m, m, nil)
-	tmp := mat.NewDense(m, m, nil)
-	tmp.Mul(u, s)
-	r.Mul(tmp, v.T())
+	r.Product(u, s, v.T())
 
 	// Calculate translation vector t.
 	t := mat.NewVecDense(m, nil)
 	rMuX := mat.NewVecDense(m, nil)
 	rMuX.MulVec(r, muX)
 
-	for i := 0; i < m; i++ {
-		t.SetVec(i, muY.AtVec(i)-c*rMuX.AtVec(i))
-	}
+	t.CopyVec(muY)
+	t.AddScaledVec(t, -c, rMuX)
 
 	return c, r, t, nil
 }
--- a/spatial/umeyama/umeyama.go
+++ b/spatial/umeyama/umeyama.go
@@ -1,0 +1,145 @@
+package umeyama
+
+import (
+	"errors"
+	"gonum.org/v1/gonum/mat"
+	"gonum.org/v1/gonum/stat"
+)
+
+const (
+	// ErrMsgSVDFailed is the error message for SVD factorization failure
+	ErrMsgSVDFailed = "SVD factorization failed"
+	// ErrMsgDegenerateInput is the error message for degenerate input data
+	ErrMsgDegenerateInput = "degenerate input: variance of X is too small"
+)
+
+// Umeyama estimates the similarity transformation parameters between two matrices X and Y.
+//
+// This is an implementation of the algorithm presented in:
+// "Least-Squares Estimation of Transformation Parameters Between Two Point Patterns"
+// by Shinji Umeyama, IEEE Transactions on Pattern Analysis and Machine Intelligence,
+// Vol. 13, No. 4, April 1991.
+//
+// The algorithm finds the optimal similarity transformation [c, R, t] ∈ Sim(m)
+// that minimizes the mean squared error between two point patterns.
+//
+// The transformation relates the point sets as:
+// Y ≈ c * R * X + t
+//
+// The dimensions of X and Y must be equal. The function will panic if they are not.
+// The points require consistent indexing. This means
+// that point i of X needs to correspond to point i of Y.
+//
+// Umeyama returns the scale factor c, the rotation matrix R and the translation vector t.
+// If a computation fails, Umeyama will return an error.
+func Umeyama(X, Y *mat.Dense) (float64, *mat.Dense, *mat.VecDense, error) {
+	rowsX, colsX := X.Dims()
+	rowsY, colsY := Y.Dims()
+
+	// Check dimensions.
+	if rowsX != rowsY || colsX != colsY {
+		panic("umeyama: dimensions of X and Y do not match")
+	}
+
+	rows, cols := rowsX, colsX
+
+	// Calculate means.
+	muX := mat.NewVecDense(rows, nil)
+	muY := mat.NewVecDense(rows, nil)
+
+	for i := 0; i < rows; i++ {
+		muX.SetVec(i, stat.Mean(mat.Row(nil, i, X), nil))
+		muY.SetVec(i, stat.Mean(mat.Row(nil, i, Y), nil))
+	}
+
+	// Calculate variance of X.
+	varX := 0.0
+	for j := 0; j < cols; j++ {
+		for i := 0; i < rows; i++ {
+			diff := X.At(i, j) - muX.AtVec(i)
+			varX += diff * diff
+		}
+	}
+	varX /= float64(cols)
+
+	// Check for degenerate case.
+	if varX < 1e-10 {
+		return 0, nil, nil, errors.New(ErrMsgDegenerateInput)
+	}
+
+	// Calculate covariance matrix.
+	covXY := mat.NewDense(rows, rows, nil)
+	for j := 0; j < cols; j++ {
+		for i := 0; i < rows; i++ {
+			diffY := Y.At(i, j) - muY.AtVec(i)
+			for k := 0; k < rows; k++ {
+				diffX := X.At(k, j) - muX.AtVec(k)
+				covXY.Set(i, k, covXY.At(i, k)+diffY*diffX/float64(cols))
+			}
+		}
+	}
+
+	// Singular Value Decomposition
+	var svd mat.SVD
+	ok := svd.Factorize(covXY, mat.SVDFull)
+	if !ok {
+		return 0, nil, nil, errors.New(ErrMsgSVDFailed)
+	}
+
+	// Get U, Σ, and V.
+	u := mat.NewDense(rows, rows, nil)
+	vt := mat.NewDense(rows, rows, nil)
+	svd.UTo(u)
+	svd.VTo(vt)
+
+	// Transpose V to get VH.
+	vh := mat.NewDense(rows, rows, nil)
+	vh.Copy(vt.T())
+
+	// Create S matrix (identity matrix).
+	s := mat.NewDiagDense(rows, nil)
+	for i := 0; i < rows; i++ {
+		s.SetDiag(i, 1.0)
+	}
+
+	// Check determinants to ensure proper rotation matrix (not reflection).
+	// This is a key contribution of Umeyama's paper - ensuring a proper rotation.
+	uDet := mat.Det(u)
+	vhDet := mat.Det(vh)
+	if uDet*vhDet < 0 {
+		s.SetDiag(rows-1, -1.0)
+	}
+
+	// Calculate scale factor c.
+	c := 0.0
+	singularValues := svd.Values(nil)
+	for i := 0; i < rows; i++ {
+		c += singularValues[i] * s.At(i, i)
+	}
+	c /= varX
+
+	// Calculate rotation matrix R.
+	tmp := mat.NewDense(rows, rows, nil)
+	r := mat.NewDense(rows, rows, nil)
+	tmp.Mul(u, s)
+	r.Mul(tmp, vh)
+
+	// Calculate translation vector t.
+	t := mat.NewVecDense(rows, nil)
+	rMuX := mat.NewVecDense(rows, nil)
+	tmp2 := mat.NewDense(rows, 1, nil)
+	for i := 0; i < rows; i++ {
+		for j := 0; j < rows; j++ {
+			tmp2.Set(i, 0, tmp2.At(i, 0)+r.At(i, j)*muX.AtVec(j))
+		}
+	}
+	for i := 0; i < rows; i++ {
+		rMuX.SetVec(i, tmp2.At(i, 0))
+	}
+
+	for i := 0; i < rows; i++ {
+		t.SetVec(i, muY.AtVec(i)-c*rMuX.AtVec(i))
+	}
+
+	return c, r, t, nil
+}
--- a/spatial/umeyama/umeyama_test.go
+++ b/spatial/umeyama/umeyama_test.go
@@ -1,0 +1,108 @@
+package umeyama
+
+import (
+	"math"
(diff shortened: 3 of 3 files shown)


kortschak (commented):
It seems to me that this should be a spatial/transform package with a Umeyama function. This allows other tranformations to be added later.

spatial/transform/umeyama.go
```diff
@@ -0,0 +1,138 @@
...
+//
+// The dimensions of X and Y must be equal. The function will panic if they are not.
+// The points require consistent indexing. This means that point i of X needs to correspond
+// to point i of Y.
+//
+// In this implementation, rows represent points and columns represent dimensions.
+//
+// Umeyama returns the scale factor c, the rotation matrix r and the translation vector t.
+// If a computation fails, Umeyama will return an error.
+func Umeyama(X, Y *mat.Dense) (float64, *mat.Dense, *mat.VecDense, error) {
```
kortschak, line 42:
```suggestion
func Umeyama(x, y *mat.Dense) (c float64, r *mat.Dense, t *mat.VecDense, err error) {
```
with associated changes below

spatial/transform/umeyama.go
```diff
@@ -0,0 +1,138 @@
...
+		mat.Col(colX, j, X)
+		mat.Col(colY, j, Y)
+		muX.SetVec(j, stat.Mean(colX, nil))
+		muY.SetVec(j, stat.Mean(colY, nil))
+	}
+
+	// Center the matrices and calculate variance of X.
+	var varX float64
+	Xc := mat.NewDense(n, m, nil)
+	Yc := mat.NewDense(n, m, nil)
```
kortschak, line 71:
```suggestion
	xc := mat.NewDense(n, m, nil)
	yc := mat.NewDense(n, m, nil)
```

spatial/transform/umeyama.go
```diff
@@ -0,0 +1,138 @@
...
+
+	for i := 0; i < n; i++ {
+		for j := 0; j < m; j++ {
+			Xc.Set(i, j, X.At(i, j)-muX.AtVec(j))
+			Yc.Set(i, j, Y.At(i, j)-muY.AtVec(j))
+
+			varX += Xc.At(i, j) * Xc.At(i, j)
```
kortschak, line 78:
```suggestion
			varX += float64(xc.At(i, j) * xc.At(i, j))
```

spatial/transform/umeyama.go
```diff
@@ -0,0 +1,138 @@
...
+		}
+	}
+	varX /= float64(n)
+
+	// Check for degenerate case. This prevents cases of division by zero and mathematical instability due to
+	// very low variance.
+	if varX < 1e-10 {
```
kortschak, line 85: Do we want this to be a parameter?

spatial/transform/umeyama.go
```diff
@@ -0,0 +1,138 @@
...
+		return 0, nil, nil, errors.New(ErrMsgDegenerateInput)
```
kortschak, line 86: In mat, we have a condition number error. This kind of thing may be useful here.

spatial/transform/umeyama.go
```diff
@@ -0,0 +1,138 @@
...
+	var svd mat.SVD
+	if !svd.Factorize(covXY, mat.SVDFull) {
+		return 0, nil, nil, errors.New(ErrMsgSVDFailed)
+	}
+
+	// Get U and V.
+	u := mat.NewDense(m, m, nil)
+	v := mat.NewDense(m, m, nil)
+	svd.UTo(u)
+	svd.VTo(v)
```
kortschak, line 104:
```suggestion
	var u, v mat.Dense
	svd.UTo(&u)
	svd.VTo(&v)
```
with associated changes below

kortschak (commented):
Initial review

spatial/transform/umeyama.go
Shieldine: I've been thinking about this a lot and I'm really not sure.

On one hand, it would allow the user of the function to adjust this value to their use-case, which may require a higher/lower threshold. I can't come up with one, but that doesn't mean there is none.

On the other hand, the average case probably wouldn't need a specific threshold and the average user wouldn't want to worry about this and/or would be unable to pick a meaningful one, resulting in non-fitting thresholds that cause errors on legitimate data.

In all cases, the most important check would be to keep this > 0. 

Looking at this now, maybe it would be a good compromise to make this a parameter, but let any input <=0 cause a fallback to 1e-10. A user could then simply put a 0. That of course assumes they read the documentation.

kortschak: I'd be happy with a negative/zero (assuming a zero is never a valid threshold) value signifying a sensible default.

What is the basis for the choice of 1e-10?

> That of course assumes they read the documentation.

No assumption is necessary. It's a deemed invariant.

spatial/transform/umeyama.go
kortschak: Yeah, I was thinking more of the pattern rather than the exact semantics; here it would be a degenerate input error which holds the value below the threshold that triggered its return.

spatial/transform/umeyama.go
Shieldine: In general, I couldn't find any kind of "best practice" when it comes to thresholds like these. 

According to my research - and please do correct me if I'm wrong - the machine epsilon when dealing with 64 bit floats sits at 1e-16. Depending on the system architecture and programming language, this could behave slightly differently. Therefore, I believe a somewhat safe threshold sits somewhere above 1e-16 (but not too far), which seems to be further validated by thresholds used in for examply scikit or numpy (the few things Ive seen were at 1e-8 to 1e-12). 

Based on the threshold I saw other libraries use, 1e-10 seems like a reasonable middle to me.
If this reasoning is alright with you, I would keep this value as a default and make the threshold a parameter.

Although now that you say it, the condition could also be changed to varX <= threshold, thus allowing 0 as a threshold as well. We should probably go with it.

spatial/transform/umeyama.go
Shieldine: Yeah, that would fit. Although I do believe that displaying the threshold itself next to the value would be beneficial for debugging. So something like:

```
// DegenerateInputError represents an error due to input data with variance
// below a threshold, which would cause numerical instability.
type DegenerateInputError struct {
	Variance  float64
	Threshold float64
}

func (e DegenerateInputError) Error() string {
	return fmt.Sprintf("variance of X (%.4e) below threshold (%.4e)", e.Variance, e.Threshold)
}
```

Does that look alright?

kortschak: That can be done by client code since the client will have passed in the threshold.

mat/errors.go
```diff
@@ -32,6 +32,16 @@ func (c Condition) Error() string {
 // condition number is above this value, the matrix is considered singular.
 const ConditionTolerance = 1e16
 
+// DegenerateInputError represents an error due to input data with variance
+// below a threshold, which would cause numerical instability.
+type DegenerateInputError struct {
+	Variance float64
+}
```
kortschak:
```suggestion
type DegenerateInputError float64
```

mat/errors.go
```diff
@@ -32,6 +32,16 @@ func (c Condition) Error() string {
...
+
+func (e DegenerateInputError) Error() string {
+	return fmt.Sprintf("variance of X (%.4e) below threshold.", e.Variance)
```
kortschak:
```suggestion
	return fmt.Sprintf("variance too low: %v", float64(e))
```

spatial/transform/umeyama.go
```diff
@@ -0,0 +1,144 @@
...
+
+	"gonum.org/v1/gonum/mat"
+	"gonum.org/v1/gonum/stat"
+)
+
+const (
+	// ErrMsgSVDFailed is the error message for SVD factorization failure.
+	ErrMsgSVDFailed = "transform: SVD factorization failed"
+)
+
```
kortschak:
```suggestion
```

spatial/transform/umeyama.go
```diff
@@ -0,0 +1,144 @@
...
+//
+// In this implementation, rows represent points and columns represent dimensions.
+//
+// Umeyama returns the scale factor c, the rotation matrix r and the translation vector t.
+// If a computation fails, Umeyama will return an error.
+// varThreshold is used for detecting a degenerate input by comparing it with the variance of x. This is necessary
+// because a variance equal or close to zero could cause numerical instability and/or division by zero.
+// In case of variance <= threshold, Umeyama will return an error.
+// The threshold should be >= 0. If a negative value is passed, the default threshold of 1e-10 will be used.
+func Umeyama(x, y *mat.Dense, varThreshold float64) (c float64, r *mat.Dense, t *mat.VecDense, err error) {
```
kortschak: The only error this can return is for when the SVD fails, so this can just be `Umeyama(x, y *mat.Dense, varThreshold float64) (c float64, r *mat.Dense, t *mat.VecDense, ok bool)` with appropriate changes in the body.

kortschak (commented):
I think we are almost there.

spatial/transform/umeyama.go
Shieldine: Two, actually. the function can return:
1) DegenerateInputError (variance too low)
2) Failed SVD

Not sure if replacing both with just an ok is a good idea as it's impossible to tell which error it was.

kortschak: You're right, I missed that. Thanks.

Rather than making the error messages const strings, make the returned errors vars based on the string. Otherwise the errors are not comparable without direct string comparison.

spatial/transform/umeyama.go
```diff
@@ -0,0 +1,140 @@
...
+package transform
+
+import (
+	"errors"
+
+	"gonum.org/v1/gonum/mat"
+	"gonum.org/v1/gonum/stat"
+)
+
+// ErrSVDFailed is returned when the SVD factorization process fails.
```
kortschak:
```suggestion
// ErrSVDFailed is returned when a required SVD factorization process fails.
```
Using the definite article assumes that the reader is aware of the context and code. We can't assume that.

spatial/transform/umeyama.go
```diff
@@ -0,0 +1,140 @@
...
+var ErrSVDFailed = errors.New("transform: SVD factorization failed")
+
+// Umeyama estimates the similarity transformation parameters between two matrices X and Y.
+//
+// This is an implementation of the algorithm presented in:
+// "Least-Squares Estimation of Transformation Parameters Between Two Point Patterns"
+// by Shinji Umeyama, IEEE Transactions on Pattern Analysis and Machine Intelligence,
+// Vol. 13, No. 4, April 1991, which can be found here: https://doi.org/10.1109/34.88573
```
kortschak: Suggest this goes at the end of the comment.
```suggestion
//
// "Least-Squares Estimation of Transformation Parameters Between Two Point Patterns"
// by Shinji Umeyama, IEEE Transactions on Pattern Analysis and Machine Intelligence,
// Vol. 13, No. 4, April 1991, [doi:10.1109/34.88573].
// [doi:10.1109/34.88573]: https://doi.org/10.1109/34.88573
```
Note for others, a free-access version exists [here](https://web.stanford.edu/class/cs273/refs/umeyama.pdf).

spatial/transform/umeyama.go
```diff
@@ -0,0 +1,140 @@
...
+// to point i of Y.
+//
+// In this implementation, rows represent points and columns represent dimensions.
+//
+// Umeyama returns the scale factor c, the rotation matrix r and the translation vector t.
+// If a computation fails, Umeyama will return an error.
+// varThreshold is used for detecting a degenerate input by comparing it with the variance of x. This is necessary
+// because a variance equal or close to zero could cause numerical instability and/or division by zero.
+// In case of variance <= threshold, Umeyama will return a DegenerateInputError.
+// The threshold should be >= 0. If a negative value is passed, the default threshold of 1e-10 will be used.
```
kortschak:
```suggestion
// Umeyama returns the scale factor c, the rotation matrix r and the translation
// vector t.
//
// minVar is used for detecting a degenerate input by comparing it with the
// variance of x. This is necessary because a variance equal or close to zero
// may cause numerical instability and/or division by zero.
// In case of variance ≤ threshold, Umeyama will return a DegenerateInputError.
// If a negative value is provided, the default threshold of 1e-10 will be used.
```
We don't need to say that an error will be returned if the function fails; this is an expectation. Do we want to say when the `ErrSVDFailed` is returned? I would expect this to be vanishingly infrequent, but it may be worth noting.

spatial/transform/umeyama.go
```diff
@@ -0,0 +1,140 @@
...
+func Umeyama(x, y *mat.Dense, varThreshold float64) (c float64, r *mat.Dense, t *mat.VecDense, err error) {
```
kortschak:
```suggestion
func Umeyama(x, y *mat.Dense, minVar float64) (c float64, r *mat.Dense, t *mat.VecDense, err error) {
```

spatial/transform/umeyama.go
```diff
@@ -0,0 +1,140 @@
...
+
+	for i := 0; i < n; i++ {
+		for j := 0; j < m; j++ {
+			xc.Set(i, j, x.At(i, j)-muX.AtVec(j))
+			yc.Set(i, j, y.At(i, j)-muY.AtVec(j))
+
+			varX += float64(xc.At(i, j) * xc.At(i, j))
+		}
+	}
+	varX /= float64(n)
```
kortschak: I'm wondering if we don't want to use [`stat.PopMeanVarian`](https://pkg.go.dev/gonum.org/v1/gonum/stat#PopMeanVariance); we go to a little more effort to avoid numerical errors there.

spatial/transform/umeyama_test.go
```diff
@@ -0,0 +1,160 @@
...
+	"gonum.org/v1/gonum/mat"
+)
+
+func TestUmeyama(t *testing.T) {
+	tol := 1e-10
+
+	testCases := []struct {
+		name      string
+		src       *mat.Dense
+		dst       *mat.Dense
```
kortschak:
```suggestion
		from      *mat.Dense
		to        *mat.Dense
```
`src` and `dst` have Go semantic loading that should be avoided.

spatial/transform/umeyama_test.go
```diff
@@ -0,0 +1,160 @@
...
+				0, 2,
+			}),
+			dst: mat.NewDense(3, 2, []float64{
+				0, 0,
+				-1, 0,
+				0, 2,
+			}),
+			wantScale: 0.7211102550927978,
+			wantRot:   mat.NewDense(2, 2, []float64{0.8320502943378437, 0.554700196225229, -0.554700196225229, 0.8320502943378436}),
+			wantTrans: mat.NewVecDense(2, []float64{-0.7999999999999998, 0.4}),
```
kortschak:
```suggestion
			wantRot:   mat.NewDense(2, 2, []float64{
				0.8320502943378437, 0.554700196225229,
				-0.554700196225229, 0.8320502943378436,
			}),
			wantTrans: mat.NewVecDense(2, []float64{
				-0.7999999999999998,
				0.4,
			}),
```
(similar below)

Can we ask for `[ -0.8; 0.4 ]` for the t with a reasonable tolerance? I would imagine so.

spatial/transform/umeyama_test.go
```diff
@@ -0,0 +1,160 @@
...
+				2, 2,
+			}),
+			dst: mat.NewDense(3, 2, []float64{
+				0, 0,
+				1, 1,
+				2, 2,
+			}),
+			wantScale: 1.0,
+			wantRot:   mat.NewDense(2, 2, []float64{1.0, 1.2116883882008518e-16, 1.2116883882008518e-16, 1.0}),
+			wantTrans: mat.NewVecDense(2, []float64{-2.220446049250313e-16, -2.220446049250313e-16}),
```
kortschak: The values at 1e-16 can be written as 0 for clarity, especially given the tolerance being used.

(similar below)

spatial/transform/umeyama_test.go
```diff
@@ -0,0 +1,160 @@
...
+				0, 0,
+				1, 0,
+				1, 1,
+			}),
+			dst: mat.NewDense(3, 2, []float64{
+				0, 0,
+				0, 1,
+				-1, 1,
+			}),
+			wantScale: 0.9999999999999999,
```
kortschak:
```suggestion
			wantScale: 1,
```
(also for the near unit value below)

spatial/transform/umeyama_test.go
```diff
@@ -0,0 +1,160 @@
...
+				4, 4,
+				5, 5,
+			}),
+			wantScale: 1.0,
+			wantRot:   mat.NewDense(2, 2, []float64{1.0, 1.2116883882008518e-16, 1.2116883882008518e-16, 1.0}),
+			wantTrans: mat.NewVecDense(2, []float64{3.0, 3.0}),
+			tolerance: tol,
+		},
+		{
+			name: "3D_case",
```
kortschak: Where does this come from?

spatial/transform/umeyama_test.go
```diff
@@ -0,0 +1,160 @@
...
+	"math"
+	"testing"
+
+	"gonum.org/v1/gonum/mat"
+)
+
+func TestUmeyama(t *testing.T) {
+	tol := 1e-10
+
+	testCases := []struct {
```
kortschak: Please hoist this out of the function as `umeyamaTests`.

spatial/transform/umeyama_test.go
```diff
@@ -0,0 +1,160 @@
...
+				1, 7, 11,
+			}),
+			wantScale: 1.0205423989219404,
+			wantRot:   mat.NewDense(3, 3, []float64{0.5699453289954445, 0.5900767342443888, -0.5718144538744644, -0.5030534073108366, 0.8008235178014148, 0.324990711758234, 0.6496919203355019, 0.10242627123762431, 0.7532657350571071}),
+			wantTrans: mat.NewVecDense(3, []float64{1.4155929948174535, 1.1579295387121973, 3.0877861136679647}),
+			tolerance: tol,
+		},
+	}
+
+	for _, tc := range testCases {
```
kortschak:
```suggestion
	for _, test := range umeyamaTests {
```

spatial/transform/umeyama_test.go
```diff
@@ -0,0 +1,160 @@
...
+		t.Run(tc.name, func(t *testing.T) {
+			scale, rotation, translation, err := Umeyama(tc.src, tc.dst, tol)
+			if err != nil {
+				t.Fatalf("Umeyama returned error: %v", err)
+			}
+
+			// Check scale
+			if math.Abs(scale-tc.wantScale) > tc.tolerance {
```
kortschak:
```suggestion
			if scalar.EqualWithinAbs(scale, test.wantScale, test.tolerance) {
```
(similar below)

spatial/transform/umeyama_test.go
Shieldine: The 3D case? I made it up just to have a test case with more than 2 dimensions.

spatial/transform/umeyama_test.go
Shieldine: We can. Ill adjust values like these for clarity

spatial/transform/umeyama_test.go
kortschak: Thanks. Have you verified it either manually or with another implementation?

Shieldine: Of course. All numbers you see in the test cases have been verified with 2 different implementations.

spatial/transform/umeyama.go
Shieldine: We could. Is this what you had in mind? (I also moved the degenerate case check up since we can spare ourselves the matrix centering in that case)

```
	// Calculate means and variance.
	muX := mat.NewVecDense(m, nil)
	muY := mat.NewVecDense(m, nil)

	colX := make([]float64, n)
	colY := make([]float64, n)

	var varX float64

	for j := 0; j < m; j++ {
		mat.Col(colX, j, x)
		mat.Col(colY, j, y)
		
		meanX, varXj := stat.PopMeanVariance(colX, nil)
		meanY, _ := stat.PopMeanVariance(colY, nil)
		
		muX.SetVec(j, meanX)
		muY.SetVec(j, meanY)
		
		varX += varXj
	}

	// Check for degenerate case. This prevents cases of division by zero and mathematical instability due to
	// very low variance.
	if varX <= minVar {
		return 0, nil, nil, mat.DegenerateInputError(varX)
	}

	// Center the matrices.
	xc := mat.NewDense(n, m, nil)
	yc := mat.NewDense(n, m, nil)

	for i := 0; i < n; i++ {
		for j := 0; j < m; j++ {
			xc.Set(i, j, x.At(i, j)-muX.AtVec(j))
			yc.Set(i, j, y.At(i, j)-muY.AtVec(j))
		}
	}
```

kortschak: `meanY, _ := stat.PopMeanVariance(colY, nil)` can stay as `stat.Mean(colY, nil)` to avoid the useless work calculating the pop variance, but yes, essentially that.

Shieldine:
Sorry! I used the resolving to track which things I've already changed in my version.

kortschak:
Yeah, no worries. I figured it was something like that. I use the emoji's for that. That may work for you.

spatial/transform/umeyama.go
```diff
@@ -0,0 +1,147 @@
...
+	"errors"
+
+	"gonum.org/v1/gonum/mat"
+	"gonum.org/v1/gonum/stat"
+)
+
+// ErrSVDFailed is returned when a required SVD factorization process fails.
+var ErrSVDFailed = errors.New("transform: SVD factorization failed")
+
+// Umeyama estimates the similarity transformation parameters between two matrices X and Y.
```
vladimir-ch: I'd replace "estimates" with "finds", there is no approximation, and I'd rewrite it a bit:
```
// Umeyama finds the similarity transformation between two sets of points
// that minimizes the mean squared error between them.
//
// The transformation relates two sets of n corresponding points {x_i}
// and {y_i} as:
//  y_i ≈ c * R * x_i + t,  i=1,...,n
// where c is the scale factor, R is the rotation matrix and t is
// the translation vector.
// 
// The point sets are represented as two n×m matrices X and Y, where
// m is the the number of dimensions and x_i and y_i are stored in the i-th
// row of X and Y, respectively. Typically, m is equal to 2 or 3.
// If the dimensions of X and Y are not equal, Umeyama will panic.
```

spatial/transform/umeyama.go
```diff
@@ -0,0 +1,147 @@
...
+// "Least-Squares Estimation of Transformation Parameters Between Two Point Patterns"
+// by Shinji Umeyama, IEEE Transactions on Pattern Analysis and Machine Intelligence,
+// Vol. 13, No. 4, April 1991, [doi:10.1109/34.88573].
+// [doi:10.1109/34.88573]: https://doi.org/10.1109/34.88573
+func Umeyama(x, y *mat.Dense, minVar float64) (c float64, r *mat.Dense, t *mat.VecDense, err error) {
+	if minVar < 0 {
+		minVar = 1e-10
+	}
+
+	rowsX, colsX := x.Dims()
```
vladimir-ch: Avoid the re-assignment down below?
```suggestion
	n, m := x.Dims()
```

spatial/transform/umeyama_test.go
```diff
@@ -0,0 +1,191 @@
...
+		}),
+		to: mat.NewDense(3, 2, []float64{
+			0, 0,
+			1, 1,
+			2, 2,
+		}),
+		wantScale: 1.0,
+		wantRot: mat.NewDense(2, 2, []float64{
+			1.0, 0.0,
+			0.0, 1.0,
```
vladimir-ch:
```suggestion
			1, 0,
			0, 1,
```

spatial/transform/umeyama_test.go
```diff
@@ -0,0 +1,191 @@
...
+		}),
+		wantTrans: mat.NewVecDense(2, []float64{
+			-2.220446049250313e-16,
+			-2.220446049250313e-16,
```
vladimir-ch:
```suggestion
			0,
			0,
```

spatial/transform/umeyama_test.go
```diff
@@ -0,0 +1,191 @@
...
+		}),
+		to: mat.NewDense(3, 2, []float64{
+			0, 0,
+			0, 1,
+			-1, 1,
+		}),
+		wantScale: 1.0,
+		wantRot: mat.NewDense(2, 2, []float64{
+			-5.613347976343136e-17, -1.0,
+			1.0, -2.9040269150165053e-16,
```
vladimir-ch:
```suggestion
			0, -1,
			1, 0,
```

spatial/transform/umeyama_test.go
```diff
@@ -0,0 +1,191 @@
...
+		}),
+		wantTrans: mat.NewVecDense(2, []float64{
+			-5.551115123125783e-17,
+			3.3306690738754696e-16,
```
vladimir-ch:
```suggestion
			0,
			0,
```

spatial/transform/umeyama_test.go
```diff
@@ -0,0 +1,191 @@
...
+			0, 0,
+			1, 0,
+			1, 1,
+		}),
+		to: mat.NewDense(3, 2, []float64{
+			0, 0,
+			0, 1,
+			-1, 1,
+		}),
+		wantScale: 1.0,
```
vladimir-ch:
```suggestion
		wantScale: 1,
```

spatial/transform/umeyama_test.go
```diff
@@ -0,0 +1,191 @@
...
+			0, 0,
+			1, 1,
+			2, 2,
+		}),
+		to: mat.NewDense(3, 2, []float64{
+			0, 0,
+			1, 1,
+			2, 2,
+		}),
+		wantScale: 1.0,
```
vladimir-ch:
```suggestion
		wantScale: 1,
```

spatial/transform/umeyama_test.go
```diff
@@ -0,0 +1,191 @@
...
+			0, 0,
+			1, 1,
+			2, 2,
+		}),
+		to: mat.NewDense(3, 2, []float64{
+			0, 0,
+			2, 2,
+			4, 4,
+		}),
+		wantScale: 2.0,
```
vladimir-ch:
```suggestion
		wantScale: 2,
```

spatial/transform/umeyama_test.go
```diff
@@ -0,0 +1,191 @@
...
+		wantRot: mat.NewDense(2, 2, []float64{
+			1.0, 0.0,
+			0.0, 1.0,
```
vladimir-ch:
```suggestion
			1, 0,
			0, 1,
```

spatial/transform/umeyama_test.go
```diff
@@ -0,0 +1,191 @@
...
+		}),
+		wantTrans: mat.NewVecDense(2, []float64{
+			-4.440892098500626e-16,
+			-4.440892098500626e-16,
```
vladimir-ch:
```suggestion
			0,
			0,
```

spatial/transform/umeyama_test.go
```diff
@@ -0,0 +1,191 @@
...
+			0, 0,
+			1, 1,
+			2, 2,
+		}),
+		to: mat.NewDense(3, 2, []float64{
+			3, 3,
+			4, 4,
+			5, 5,
+		}),
+		wantScale: 1.0,
```
vladimir-ch:
```suggestion
		wantScale: 1,
```

spatial/transform/umeyama_test.go
```diff
@@ -0,0 +1,191 @@
...
+		wantRot: mat.NewDense(2, 2, []float64{
+			1.0, 0.0,
+			0.0, 1.0,
```
vladimir-ch:
```suggestion
			1, 0,
			0, 1,
```

spatial/transform/umeyama_test.go
```diff
@@ -0,0 +1,191 @@
...
+		}),
+		wantTrans: mat.NewVecDense(2, []float64{
+			3.0,
+			3.0,
```
vladimir-ch:
```suggestion
			3,
			3,
```

spatial/transform/umeyama_test.go
```diff
@@ -0,0 +1,191 @@
...
+		}),
+	},
+}
+
+func TestUmeyama(t *testing.T) {
+	tol := 1e-10
+
+	for _, test := range umeyamaTests {
+		t.Run(test.name, func(t *testing.T) {
+			scale, rotation, translation, err := Umeyama(test.from, test.to, tol)
```
vladimir-ch: Using `tol` here might give the wrong impression to the reader that somehow the number is related to the accuracy of the output. I'd pass -1 here to use the default value and use `tol` (with a more stringent value of 1e-14, for example) only for checking the output due to floating point errors.

spatial/transform/umeyama_test.go
```diff
@@ -0,0 +1,191 @@
...
+
+			// Check rotation
+			_, d := test.wantRot.Dims()
+			for i := 0; i < d; i++ {
+				for j := 0; j < d; j++ {
+					if !scalar.EqualWithinAbs(rotation.At(i, j), test.wantRot.At(i, j), tol) {
+						t.Errorf("Rotation[%d,%d] = %v, want %v", i, j, rotation.At(i, j), test.wantRot.At(i, j))
+					}
+				}
+			}
```
vladimir-ch: A better, succinct way of doing this check would be with `mat.EqualApprox`:
```
			if !mat.EqualApprox(rotation, test.wantRot, tol) {
				t.Errorf("Rotation matrix mismatch:\nGot:\n%v\nWant:\n%v",
					mat.Formatted(rotation), mat.Formatted(test.wantRot))
			}
```
However, it's a pain to construct the error message and then to visually compare the two matrices in case of an error. Plus, if the implementation is wrong (rare), the individual entries are not very relevant, while if the test tolerance is too strict (common), the difference is harder to spot.

My preferred way would be to compute a _norm_ of the difference between the computed and expected rotation and compare that against the test tolerance:
```suggestion
			// Check rotation
			var rDiff mat.Dense
			rDiff.Sub(rotation, test.wantRot)
			diff := rDiff.Norm(1)
			if diff > tol {
				t.Errorf("unexpected rotation matrix, |R_got-R_want| = %v", diff)
			}
```
If the test tolerance is too strict, it will print a tiny number such as:
```
unexpected rotation matrix, |R_got-R_want| = 3.552713678800501e-15
```
If the implementation is wrong, it will print a wild number.

spatial/transform/umeyama_test.go
```diff
@@ -0,0 +1,191 @@
...
+
+			// Check translation
+			for i := 0; i < d; i++ {
+				if !scalar.EqualWithinAbs(translation.AtVec(i), test.wantTrans.AtVec(i), tol) {
+					t.Errorf("Translation[%d] = %v, want %v", i, translation.AtVec(i), test.wantTrans.AtVec(i))
+				}
+			}
```
vladimir-ch, line 190: Same here:
```suggestion
			var tDiff mat.VecDense
			tDiff.SubVec(translation, test.wantTrans)
			diff = tDiff.Norm(1)
			if diff > tol {
				t.Errorf("unexpected translation vector, |t_got-t_want| = %v", diff)
			}
```

spatial/transform/umeyama.go
```diff
@@ -0,0 +1,146 @@
...
+// ErrSVDFailed is returned when a required SVD factorization process fails.
+var ErrSVDFailed = errors.New("transform: SVD factorization failed")
+
+// Umeyama finds the similarity transformation between two sets of points
+// that minimizes the mean squared error between them.
+//
+// The transformation relates two sets of n corresponding points {x_i}
+// and {y_i} as:
+//
+//	y_i ≈ c * R * x_i + t,  i=1,...,n
```
vladimir-ch: Let's use 0-based indexing here so that the notation here matches when we talk about i-th row of X and Y below:
```suggestion
//	y_i ≈ c * R * x_i + t,  i=0,...,n-1
```

spatial/transform/umeyama.go
```diff
@@ -0,0 +1,146 @@
...
+//
+// where c is the scale factor, R is the rotation matrix and t is
+// the translation vector.
+//
+// The point sets are represented as two n×m matrices X and Y, where
+// m is the number of dimensions and x_i and y_i are stored in the i-th
+// row of X and Y, respectively. Typically, m is equal to 2 or 3.
+// If the dimensions of X and Y are not equal, Umeyama will panic.
+//
+// Umeyama returns the scale factor c, the rotation matrix r and the translation
```
vladimir-ch: Just a nitpick, I'd refer to the mathematical entity R here:
```suggestion
// Umeyama returns the scale factor c, the rotation matrix R and the translation
```

spatial/transform/doc.go
```diff
@@ -0,0 +1,6 @@
+// Copyright ©2025 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Package transform provides functions for spatial transformations.
+package transform // import "gonum.org/v1/gonum/spatial/transform"
```
vladimir-ch, line 6: Is the canonical import path comment necessary? I wasn't even aware this existed.

vladimir-ch (commented):
Last few comments and then LGTM.

spatial/transform/doc.go
Shieldine: I actually copied that from other doc.go files in this repository for consistency. 
Example here:
https://github.com/gonum/gonum/blob/master/spatial/r2/doc.go

Shieldine:
CI fail seems unrelated to the changes here

kortschak:
@fumin would you be able to take a look at this?

mat/errors.go
```diff
@@ -32,6 +32,14 @@ func (c Condition) Error() string {
 // condition number is above this value, the matrix is considered singular.
 const ConditionTolerance = 1e16
 
+// DegenerateInputError represents an error due to input data with variance
+// below a threshold, which would cause numerical instability.
+type DegenerateInputError float64
```
fumin: I think it is inappropriate to introduce `DegenerateInputError` to the general `mat` package.
Errors defined in `mat` relate to central concepts in linear algebra such as strides and condition numbers, whereas DegenerateInput is not.

spatial/transform/umeyama.go
```diff
@@ -0,0 +1,146 @@
...
+
+import (
+	"errors"
+
+	"gonum.org/v1/gonum/mat"
+	"gonum.org/v1/gonum/stat"
+)
+
+// ErrSVDFailed is returned when a required SVD factorization process fails.
+var ErrSVDFailed = errors.New("transform: SVD factorization failed")
```
fumin: I see that there's a `ErrFailedEigen` in the `mat` package. Perhaps it's better to put `ErrSVDFailed` next to `ErrFailedEigen` and rename it as `ErrFailedSVD`?

spatial/transform/umeyama.go
```diff
@@ -0,0 +1,146 @@
...
+// variance of x. This is necessary because a variance equal or close to zero
+// may cause numerical instability and/or division by zero.
+// In case of variance ≤ minVar, Umeyama will return a DegenerateInputError.
+// If a negative value is provided, the default threshold of 1e-10 will be used.
+//
+// "Least-Squares Estimation of Transformation Parameters Between Two Point Patterns"
+// by Shinji Umeyama, IEEE Transactions on Pattern Analysis and Machine Intelligence,
+// Vol. 13, No. 4, April 1991, [doi:10.1109/34.88573].
+// [doi:10.1109/34.88573]: https://doi.org/10.1109/34.88573
+func Umeyama(x, y *mat.Dense, minVar float64) (c float64, r *mat.Dense, t *mat.VecDense, err error) {
```
fumin: I prefer removing the `minVar` argument. It sort of clouds the main purpose of this function.
The original paper does not mention about this either.
Nonetheless, we could mention this caveat in the function doc.

fumin (commented):
Looks good in general, just some comments about the `minVar` argument and error handling.

spatial/transform/umeyama.go
Shieldine: Correct, it's not in the original paper. It is used for defining a threshold on detecting a degenerate input, thus allowing an early exit in order to avoid numerical instability and thus rubbish results.

Why do you think it clouds the main purpose?

Having this as a parameter allows the user to set this threshold as he pleases - if not sure, a negative value will cause a default. See other comments below on this.

fumin: I am thinking in terms of the analogy with respect to solutions of linear equations.
The purpose of a linear solver is to produce the solutions, whether or not the solutions are *good* depends on the condition number.
In our case, the purpose is to produce the transformation, whether or not the transformations are good depends on `varX`.
In both cases, the definition of good enough is subjective.
Linear solvers rarely accept a `conditionNumber` argument, but users that are wary of the goodness of the returned solution can inspect the condition number separately should they wish to.
Likewise, I imagine the way Umeyama is used is something like compute the transform, then optionally inspect `varX` should users wish to.

Shieldine: I see where you are coming from, but I'm not sure I agree. The variance we are checking here can cause division by zero which needs to be avoided - so we do need some sort of check anyway. Furthermore, we do unnecessary work if the solution will be garbage anyway. We can spare this work by skipping further calculation and telling the user that the number will be, in fact, garbage. It is more user-friendly to say exactly what happened (variance low, the output will be useless) than to have them inspect the numbers afterwards and come up with what happened.

Passing a negative value gives a reasonable variance check that can be further changed as the user pleases, in case their use-case already produces useless results way earlier. We give the user a chance to define what is "good enough" and what can be skipped right away. While it may not be the most standard approach, it gives the user more flexibility, which is a worthwile goal in my opinion

mat/errors.go
Shieldine: Where would you suggest putting it instead?

kortschak: It could go in spatial/transform.

spatial/transform/umeyama.go
kortschak: I've taken a look at a bunch of implementations (not only in python) and most seem to not even consider the possibility of an insufficient variance. I'm beginning to think that maybe we could have a hard threshold at the machine eps and return the variance as an error when below that, but always complete the work, and also return the variance so that the caller can determine whether they are within their threshold. This would change the signature to `func Umeyama(x, y *mat.Dense) (c, varX float64, r *mat.Dense, t *mat.VecDense, err error)`.

mat/errors.go
fumin: We can dispense with this variable if we don't need the `minVar` argument

spatial/transform/umeyama.go
fumin: I think the above function signature makes sense, except we could order the returns as `{c, r, t, varX}` according to their importance?
The performance impact from the potential duplicate `varX` computation is negligible, since the complexity of `SVD` is an order of magnitude above that of `varX`.

mat/errors.go
kortschak: No, I don't think so. We would still need to be able to signal a below machine eps condition. That would be an error.

spatial/transform/umeyama.go
kortschak: We can't drop the `error` since we have two possible failure modes; the variance being below machine eps, and the SVD failure. The reason I put `varX` second was for two reasons: `varX` and `c` are related; and they have the same type, simplifying the signature.

I'm not sure what you're getting at with "The performance impact from the potential duplicate `varX` computation is negligible". What is the significance of that?

mat/errors.go
fumin: Sorry, I meant perhaps we refrain from checking `varX` altogether.
Users who care about `varX` can check for it using their custom method before running the function.

spatial/transform/umeyama.go
fumin: Sorry, typo, yes we should keep `err` to signal SVD errors.

About the "duplicate varX computations" phrase, I was referring to Shieldine's observation that users typically check `varX` *before* calling Umeyama. In this case, Umeyama's second computation of `varX` is a duplicate one. I was suggesting that this is a minor concern since the main bottleneck is SVD.

kortschak: Ah yeah. I was thinking about something similar that when I was writing the comment above. Another option is for the process be a struct constructor where the struct has two methods, `Var() float64` and `Transform() (c float64, r *mat.Dense, t *mat.VecDense, err error)`. The constructor could be either the first half of the current implementation (up to completion of `varX` calculation), with `Var` returning that value, and then `Transform` doing the rest. Alternatively, if the X and Y are exported fields in the struct, then a struct literal could be the constructor and `Var`does the work as well, rather than just being an accessor.

fumin: Right, a `struct` would work. Another way to use a struct is to include `minVar` as a field and implement Umeyama as is in the current version.

Shieldine: Interesting ideas! Reading through that, in terms of simplicity of usage, having the current single function but always returning the variance and only returning the error below machine eps is probably best. Admittedly, I really like this idea. Of course, if the results are garbage, we do unnecessary work in this scenario. It may be worth considering to still *not* do the work here.

On the other hand, having a struct with two steps (Var() and Transform()) sounds interesting and allows the user to check the variance before heavy computation begins. If we opt for a struct, I think I'd prefer this way. I don't think creating a struct with `minVar` as a field brings any advantages.
Then again - isn't a struct for a simple calculation like this slightly over-engineered? We would be sacrificing simplicity of usage, which is one of Go's main goals, for a function a user probably just wants to call.

For a decision we should probably think what our main goal is: do we want to give an option to prevent unnecessary work, unlike classical solvers that don't? Do we expect our users to check the variance themselves beforehand?
If we answer yes to first and no to second, I think the current state is perfectly fine.
If we answer no to first and no to second, we could go with the modified function always returning the variance and an error below machine eps, but always completing the work.
If we answer no and yes, then we could remove the check altogether (I don't think this is wise).
If we answer yes and yes, then perhaps a struct is best. This way, we'd give them a simple standard flow for a variance check and then computation.

Personally, I believe a user wants to simply call the function and be done - speaking against a struct.
They probably expect proper error handling in the function, also in case they didn't think of something - speaking for a variance check
They probably don't want to deal with garbage data - speaking for an early exit if we detect garbage. Not wanting to wait for garbage data also speaks for this.

I know many implementations don't consider this possibility at all and assume the user made all necessary precautions, but that doesn't mean we can't be more user-friendly. When opting for libraries, the ease of use is a huge factor among all developers I know, and giving extra options here doesn't cost us anything, but could be a big relief to anybody using this.
What do you guys make of all that?

fumin: Thinking deeper, I think all cases of `varX` can be handled:

case varX == 0:
This means X collapses to a point. In this case, if Y is not collapsed, then `c` is truly Inf. Else, the problem reduces to the translation between two points.

case 0 < varX < epsilon:
We can scale X with `math.Pow(radix, n)` so that `varX ~= O(1)`. This is actually how LAPACK typically operates.

Overall, no errors need to be raised provided we enumerate all special cases, and scale X with `radix^n` properly.

Shieldine: I don't think artificially scaling it up makes sense. With the variance being 0, we can't properly calculate. However, due to machine imprecision, it is possible that something that should be zero ends up being slightly above that - like 1e-9 or something. This is why we are catching this case and also exiting - it probably should be 0.

fumin: I suggest you consult Section 3.5.2, Golub & Van Loan to understanding how scaling works in numerical computations. A concrete example is the LAPACK `dgeequb` routine. The key thing to note is that the factor is chosen to powers of the floating point radix in the above scheme.

Shieldine: I may be misunderstanding, but I don't think it applies here. If we artificially scale up, we become numerically stable, yes, but the output we are producing is still garbage. Our issue is that the input is degenerate - meaning the points in the set are essentially clustered together in basically the same place, maybe due to previous computations that left an error, maybe because they are essentially the same but slightly off due to measurement error. This means the point set has no meaningful orientation or extent. 

If I understand correctly, LAPACK's equilibration helps when you have a well-posed problem with badly scaled coefficients. Here, we have a fundamentally ill-defined problem: you can't find a meaningful similarity transformation when one point set has essentially collapsed to a point. 
Scaling up by powers of 2 doesn't change the fact that the original points had no spread - it just artificially inflates numerically meaningless differences into seemingly valid results. I believe it's better to fail with a clear error than to return mathematically well-formed but physically meaningless transformations.

kortschak: @Shieldine I think the answer [above](https://github.com/gonum/gonum/pull/2048#discussion_r2596189259) is yes and yes, but we can provide a convenience helper that does the "just one call gives you the answer" approach. This grows the API slightly, but I think the growth is paid for by the flexibility. My suggestion for the convenience helper's behaviour would be to not do fruitless work in the case of an insufficient variance; this means the default case favours efficiency, but the complete work case _can_ be achieved.

fumin: I'd to refrain from hijacking this pull request, but to offer another point of view:

Imagine a concrete use case of this function: a researcher doing both archeology and astronomy trying to map markings in Thebes to stars in a remote galaxy. If she chooses to use units from astronomy, then `varX` would be smaller than `epsilon`. Nonetheless, with proper scaling, the answer of this function remains to be absolutely "numerically correct" in the sense that the answer is identical to the one computed under float1024 truncated to the same digits. Most importantly, it is also "scientifically correct" with respect to this research topic.

While in most engineering fields such as robotics, a small `varX` commonly indicates an input error, such inputs may be considered legitimate in other scientific fields such as astronomy and quantum physics.

I think the goal of gonum, like many numerical libraries, is to be just "numerically correct". Whether or not the calculation being done makes scientific sense is best left to the user to decide.

kortschak: The researcher can either pre-condition the values to be in a reasonable range, or if they are brave, use the construction that's available with the struct's methods to do whatever they want with the values.

fumin: Right, I agree a struct based design offers the flexibility required for these diverse use cases.

Shieldine:
The latest commits implement the discussed changes:
Umeyama is now a struct, with its constructor performing everything up to calculation of varX.
Var() returns the calculated variance. The actual transformation calculation can be triggered using Transform().

Additionally, a convenience wrapper is provided. The wrapper checks for machine epsilon before triggering further computation.

The tests have been updated to use the convenience function.

The errors have been moved as requested.

spatial/transform/umeyama.go
```diff
@@ -0,0 +1,228 @@
...
+
+	// Calculate covariance matrix.
+	covXY := mat.NewDense(u.m, u.m, nil)
+	covXY.Mul(yc.T(), xc)
+	covXY.Scale(1/float64(u.n), covXY)
+
+	// Singular Value Decomposition
+	var svd mat.SVD
+	if !svd.Factorize(covXY, mat.SVDFull) {
+		return 0, nil, nil, mat.ErrFailedSVD
```
kortschak: With the new design, the only failure that can happen in this method is the SVD factorisation. So we don't need an error at all; this can just return an `ok bool` like the actual factorisation does.

spatial/transform/umeyama.go
```diff
@@ -0,0 +1,228 @@
...
+// Vol. 13, No. 4, April 1991, [doi:10.1109/34.88573].
+// [doi:10.1109/34.88573]: https://doi.org/10.1109/34.88573
+func UmeyamaTransform(x, y *mat.Dense) (c float64, r *mat.Dense, t *mat.VecDense, err error) {
+	u := NewUmeyama(x, y)
+
+	if u.varX <= math.Nextafter(1.0, 2.0)-1.0 {
+		return 0, nil, nil, DegenerateInputError(u.varX)
+	}
+
+	return u.Transform()
```
kortschak: We do need the error here, but I think it should be local to the package. I don't think we need to say anything about SVD; ` var ErrFactorizationFailed = errors.New("transform: factorization failed")`.

mat/errors.go
```diff
@@ -137,6 +137,7 @@ var (
 	ErrSliceLengthMismatch = Error{"mat: input slice length mismatch"}
 	ErrNotPSD              = Error{"mat: input not positive symmetric definite"}
 	ErrFailedEigen         = Error{"mat: eigendecomposition not successful"}
+	ErrFailedSVD           = Error{"mat: SVD factorization failed"}
```
kortschak: See previous comment.

spatial/transform/umeyama.go
```diff
@@ -224,5 +228,12 @@ func UmeyamaTransform(x, y *mat.Dense) (c float64, r *mat.Dense, t *mat.VecDense
 		return 0, nil, nil, DegenerateInputError(u.varX)
 	}
 
-	return u.Transform()
+	var ok bool
+	c, r, t, ok = u.Transform()
+
```
kortschak:
```suggestion
```

spatial/transform/umeyama.go
```diff
@@ -224,5 +228,12 @@ func UmeyamaTransform(x, y *mat.Dense) (c float64, r *mat.Dense, t *mat.VecDense
...
+	if !ok {
+		return 0, nil, nil, ErrFactorizationFailed
+	}
+
```
kortschak:
```suggestion
```

kortschak (commented):
nits only, then LGTM

fumin:
Thanks, looks good to me.

PR #2050, opened 2025-06-18
Lapack: Changing Dpstrf and Dpstf2 return values

dgelody:
Currently, the functions `Dpstrf` and `Dpstf2` do not attempt to check that the input matrix `A` is positive semi-definite, so if the return variable `ok` is false, the matrix A is either rank deficient or is not positive semidefinite. 

However, it should be necessary to distinguish between these two possibilities simply by looking at the value of `ok`. Hence, in the change these functions will return false if the input matrix `A` is not positive semi-definite and return true if the matrix `A` is rank deficient, with the return variable `rank` indicating the rank of the matrix. Note, the Cholesky factorization with complete pivoting of an nxn symmetric positive semidefinite matrix A is expected to be rank deficient.

These are my thoughts so I would love any feedback!

### DIFF
--- a/lapack/gonum/dpstf2.go
+++ b/lapack/gonum/dpstf2.go
@@ -30,9 +30,7 @@ import (
 // piv contains P stored such that P[piv[k],k] = 1.
 //
 // Dpstf2 returns the computed rank of A and whether the factorization can be
-// used to solve a system. Dpstf2 does not attempt to check that A is positive
-// semi-definite, so if ok is false, the matrix A is either rank deficient or is
-// not positive semidefinite.
+// used to solve a system. If 'ok' is false, matrix A is not positive semidefinite.
 //
 // The length of piv must be n and the length of work must be at least 2*n,
 // otherwise Dpstf2 will panic.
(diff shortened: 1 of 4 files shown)


kortschak (commented):
The implementation here should be consistent with the LAPACKE implementation. This would break that consistency since it does not report this.

vladimir-ch:
I will close the PR with this comment because apart from @kortschak 's objection there is no theoretical justification for this algorithm and it doesn't work, it fails to identify the matrix [1 0 0; 0 0 1; 0 1 0] as indefinite.

The follow-up comment is false in a way that smells of A"I".

kortschak:
> The matrix results are the same.

Yes, because the Go implementation is a port of the [Fortran](https://netlib.org/lapack/explore-html-3.6.1/dd/dad/dpstrf_8f_a31cdc13a7f4ad687f4aefebff870e1cc.html).

> The "ok" flag in Gonum is an irritant but we will work around it.

The return value occupies the same role as the info out parameter in the Fortran. I don't understand how this is an irritant.

As noted above, the Go implementation reflects the Fortran implementation as close to exactly as it can (being constrained by the fact that we are row major and zero indexed while Fortran is no). We are not free to move the behaviour of our implementation. I would suggest that if you have issues with how these routines operate, you file an issue with netlib.

PR #2051, opened 2025-07-29
Add Dinic maximum flow algorithm 

Schwarf:
## Summary
This pull request introduces the [Dinic maximum flow algorithm](https://en.wikipedia.org/wiki/Dinic%27s_algorithm) into the `network` package. It provides:

* **Public API**

  ```go
  func MaxFlowDinic(g graph.WeightedDirected, source, target graph.Node) (float64, error)
  ```

  Computes the maximum flow from `source` to `target` on any weighted, directed graph.

* **Internal helpers**

  * `initializeResidualGraph` — builds the residual graph
  * `canReachTargetInLevelGraph` — constructs level graphs and records parent relationships
  * `computeBlockingPath` — finds and augment blocking flows

* **Comprehensive tests** (`network/maximum_flow_test.go`):

  * Error on identical source and target
  * Three disjoint unit‑capacity paths
  * Cycle plus tail graph
  * Four‑layered DAG scenarios
  * Diamond shape with cross edges
  * Disconnected graphs

This is my first open source contribution to a Go project, and I would appreciate any feedback and guidance!

---

Please take a look.


kortschak:
Thanks. I'll take a look on the weekend.

graph/network/maximum_flow.go
```diff
@@ -0,0 +1,168 @@
+package network
```
kortschak, line 5: Please add copyright headers to the new files.

graph/network/maximum_flow.go
```diff
@@ -0,0 +1,168 @@
...
+import (
+	"container/list"
+	"fmt"
+	"math"
+
+	"gonum.org/v1/gonum/graph"
+	"gonum.org/v1/gonum/graph/simple"
+)
+
+type edgeKey struct{ from, to int64 }
```
kortschak: Structure code so that caller is placed before callee (and the equivalent in type use). This gives context for the reader to understand where/how a defined function or type will be used, and we are not writing C, so we are not constrained by its source ordering.

graph/network/maximum_flow.go
```diff
@@ -0,0 +1,168 @@
...
+	flow := make(map[edgeKey]float64)
+	// Add all nodes
+	for nodes := originalGraph.Nodes(); nodes.Next(); {
+		graphCopy.AddNode(nodes.Node())
+	}
+
+	// For each node u :
+	for nodes := originalGraph.Nodes(); nodes.Next(); {
+		u := nodes.Node()
+		// Iterate over all children of u
```
kortschak: Comments should be complete punctuated sentences (full stop here and capitalisation in the sentences below), but I'm not convinced that this comment is needed.

graph/network/maximum_flow.go
```diff
@@ -0,0 +1,168 @@
...
+func initializeResidualGraph(originalGraph graph.WeightedDirected) (*residualGraph, error) {
+	graphCopy := simple.NewWeightedDirectedGraph(0, 0)
+	flow := make(map[edgeKey]float64)
+	// Add all nodes
+	for nodes := originalGraph.Nodes(); nodes.Next(); {
+		graphCopy.AddNode(nodes.Node())
+	}
+
+	// For each node u :
+	for nodes := originalGraph.Nodes(); nodes.Next(); {
```
kortschak:
```suggestion
	nodes := g.Nodes()
	for nodes.Next() {
```

graph/network/maximum_flow.go
```diff
@@ -0,0 +1,168 @@
...
+		u := nodes.Node()
+		// Iterate over all children of u
+		for it := originalGraph.From(u.ID()); it.Next(); {
```
kortschak:
```suggestion
		 it := g.From(u.ID())
		 for it.Next() {
```

graph/network/maximum_flow.go
```diff
@@ -0,0 +1,168 @@
...
+			v := it.Node()
+			// get the weight/capacity
+			capacity, ok := originalGraph.Weight(u.ID(), v.ID())
+			if !ok {
+				panic("expected a weight for existing edge")
+			}
+			if capacity < 0.0 {
```
kortschak:
```suggestion
			if capacity < 0 {
```

graph/network/maximum_flow.go
```diff
@@ -0,0 +1,168 @@
...
+				return nil, fmt.Errorf("edge weights (capacities) can not be negative")
+			}
+			// add forward edge to residualGraph (capacity)
+			forward := graphCopy.NewWeightedEdge(u, v, capacity)
+			graphCopy.SetWeightedEdge(forward)
+			// add reverse edge v->u with zero weight (flow)
+			flow[edgeKey{from: u.ID(), to: v.ID()}] = 0.0
```
kortschak:
```suggestion
			r.flow[edgeKey{from: u.ID(), to: v.ID()}] = 0
```

graph/network/maximum_flow.go
```diff
@@ -0,0 +1,168 @@
...
+
+	"gonum.org/v1/gonum/graph"
+	"gonum.org/v1/gonum/graph/simple"
+)
+
+type edgeKey struct{ from, to int64 }
+
+type residualGraph struct {
+	Graph *simple.WeightedDirectedGraph
+	Flow  map[edgeKey]float64
```
kortschak: These don't need to be exported, suggest `g` and `flow`.

graph/network/maximum_flow.go
```diff
@@ -0,0 +1,168 @@
...
+}
+
+func initializeResidualGraph(originalGraph graph.WeightedDirected) (*residualGraph, error) {
+	graphCopy := simple.NewWeightedDirectedGraph(0, 0)
+	flow := make(map[edgeKey]float64)
+	// Add all nodes
+	for nodes := originalGraph.Nodes(); nodes.Next(); {
+		graphCopy.AddNode(nodes.Node())
+	}
```
kortschak:
```suggestion
	r := &residualGraph{
		g:    simple.NewWeightedDirectedGraph(0, 0)},
		flow: make(map[edgeKey]float64),
	}
	nodes := originalGraph.Nodes()
	for nodes.Next() {
		r.g.AddNode(nodes.Node())
	}
```
similar changes below

graph/network/maximum_flow.go
```diff
@@ -0,0 +1,168 @@
...
+		Flow:  flow,
+	}, nil
+}
+
+func computeBlockingPath(residualGraph *residualGraph, source, target graph.Node, parents [][]int64) float64 {
+	var totalFlow = 0.0
+	var path []int64
+	targetID := target.ID()
+	path = append(path, targetID)
+	var uID = targetID
```
kortschak:
```suggestion
	totalFlow := 0
	tid := t.ID()
	path := []int64{tid}
	uid := tid
```

graph/network/maximum_flow.go
```diff
@@ -0,0 +1,168 @@
...
+	for {
+		var vID int64
```
kortschak:
```suggestion
		var vid int64
```

graph/network/maximum_flow.go
```diff
@@ -0,0 +1,168 @@
...
+			flow[edgeKey{from: u.ID(), to: v.ID()}] = 0.0
+		}
+	}
+	return &residualGraph{
+		Graph: graphCopy,
+		Flow:  flow,
+	}, nil
+}
+
+func computeBlockingPath(residualGraph *residualGraph, source, target graph.Node, parents [][]int64) float64 {
```
kortschak:
```suggestion
func computeBlockingPath(r *residualGraph, s, t graph.Node, parents [][]int64) float64 {
```

graph/network/maximum_flow.go
```diff
@@ -0,0 +1,168 @@
...
+			forward := graphCopy.NewWeightedEdge(u, v, capacity)
+			graphCopy.SetWeightedEdge(forward)
+			// add reverse edge v->u with zero weight (flow)
+			flow[edgeKey{from: u.ID(), to: v.ID()}] = 0.0
+		}
+	}
+	return &residualGraph{
+		Graph: graphCopy,
+		Flow:  flow,
+	}, nil
```
kortschak:
```suggestion
	return r, nil
```

graph/network/maximum_flow.go
```diff
@@ -0,0 +1,168 @@
...
+			path = append(path, vID)
+		} else {
+			path = path[:len(path)-1]
+			if len(path) == 0 {
+				break
+			}
+			vID = path[len(path)-1]
+		}
+		// path has been build from target to source, so the parent is on i+1 position of the ith child
+		if vID == source.ID() {
```
kortschak:
```suggestion
		if vid == s.ID() {
```

graph/network/maximum_flow.go
```diff
@@ -0,0 +1,168 @@
...
+			bottleNeckOnPath := math.MaxFloat64
+			// determine minimal flow on path
+			for i := 0; i+1 < len(path); i++ {
+				parentID := path[i+1]
```
kortschak:
```suggestion
				pid := path[i+1]
```

graph/network/maximum_flow.go
```diff
@@ -0,0 +1,168 @@
...
+				childID := path[i]
```
kortschak:
```suggestion
				cid := path[i]
```

graph/network/maximum_flow.go
```diff
@@ -0,0 +1,168 @@
...
+				weight, ok := residualGraph.Graph.Weight(parentID, childID)
```
kortschak:
```suggestion
				w, ok := r.Graph.Weight(pid, cid)
```

graph/network/maximum_flow.go
```diff
@@ -0,0 +1,168 @@
...
+func MaxFlowDinic(graph graph.WeightedDirected, source, target graph.Node) (float64, error) {
+	if source.ID() == target.ID() {
+		return 0, fmt.Errorf("source and target must be different")
+	}
+	parents := make([][]int64, graph.Nodes().Len())
+	residualGraph, err := initializeResidualGraph(graph)
+	if err != nil {
+		return 0, fmt.Errorf("could not build residual graph: %v", err)
+	}
+	epsilon := 1.e-12
```
kortschak: Why is this chosen?
```suggestion
	const epsilon = 1e-12
```

graph/network/maximum_flow.go
```diff
@@ -0,0 +1,168 @@
...
+	var maxFlow = 0.0
```
kortschak:
```suggestion
	var max float64
```

graph/network/maximum_flow_test.go
```diff
@@ -0,0 +1,246 @@
...
+	"testing"
+
+	"gonum.org/v1/gonum/graph/simple"
+)
+
+// almostEqual tests float64 equality within a small epsilon.
+func almostEqual(a, b float64) bool {
+	const eps = 1e-9
+	return math.Abs(a-b) <= eps
+}
```
kortschak: We have functions for this is floats/scalar.

graph/network/maximum_flow_test.go
kortschak, line 1: I'll take a look at this tomorrow, but I think we could revise it to use table driven tests.

Schwarf:
Hi @kortschak,

thank you for your review. Regarding some of your suggestions:

I understand the suggestion to use shorter names — and in Go it’s idiomatic to use one‑letter variables for small, localized scopes (e.g. `i` in a loop). However, in this context, with several different graph‑related concepts in play, I find that names like `parentID`, `childID`, `source`, `target`, and `residualGraph` make the code much more self‑explanatory.

Using abbreviations like `pid`, `cid`, `s`, `t`, and `r` adds cognitive overhead for readers, since they have to constantly recall what each abbreviation means. With the longer names, the code reads itself and there’s no ambiguity about what each variable represents.

If you don’t mind, I’d prefer to keep the more descriptive names, and perhaps improve them where needed — for example, using `currentID` instead of `uID` for better clarity and consistency.

kortschak:
The names that I suggested here are conventional within the graph packages. For this reason, using them reduces cognitive load for maintainers.

Schwarf:
Got it.
I’ll update the variable names to match the existing conventions.

graph/network/maximum_flow.go
Schwarf: I chose 1e-12 because it exceeds accumulated rounding-errors (~1e-13) yet remains well below typical capacity scales. But its not perfect (e.g. when capacities of the order of epsilon come into play).

graph/network/maximum_flow.go
```diff
@@ -0,0 +1,209 @@
...
+	"math"
+
+	"gonum.org/v1/gonum/floats/scalar"
+	"gonum.org/v1/gonum/graph"
+	"gonum.org/v1/gonum/graph/simple"
+)
+
+// MaxFlowDinic computes the maximum flow from source to target in a directed,
+// weighted graph using Dinic’s algorithm. It repeatedly builds level graphs
+// and augments blocking flows until no more augmenting paths exist.
```
kortschak:
```suggestion
// MaxFlowDinic computes the maximum flow from source to target in a directed,
// weighted graph using [Dinic's algorithm]. It repeatedly builds level graphs
// and augments blocking flows until no more augmenting paths exist.
//
// [Dinic's algorithm]: https://en.wikipedia.org/wiki/Dinic%27s_algorithm
```

graph/network/maximum_flow.go
```diff
@@ -0,0 +1,209 @@
...
+func MaxFlowDinic(graph graph.WeightedDirected, s, t graph.Node) (float64, error) {
```
kortschak:
```suggestion
func MaxFlowDinic(g graph.WeightedDirected, s, t graph.Node) float64 {
```
and a documented panic for the invariants that are expected.

graph/network/maximum_flow.go
```diff
@@ -0,0 +1,209 @@
...
+	if s.ID() == t.ID() {
+		return 0, fmt.Errorf("s and t must be different")
+	}
+	parents := make([][]int64, graph.Nodes().Len())
+	r, err := initializeResidualGraph(graph)
+	if err != nil {
+		return 0, fmt.Errorf("could not build residual graph: %v", err)
+	}
+	const epsilon = 1.e-12
+	var maxFlow = 0.0
```
kortschak:
```suggestion
	const epsilon = 1e-12
	var maxFlow float64
```

graph/network/maximum_flow.go
```diff
@@ -0,0 +1,209 @@
...
+		// flow tracks the current flow for each forward edge (initially zero).
+		flow: make(map[edgeKey]float64),
+	}
+
+	nodes := g.Nodes()
+	for nodes.Next() {
+		r.g.AddNode(nodes.Node())
+	}
+
+	nodes = g.Nodes()
```
kortschak:
```suggestion
	nodes.Reset()
```

graph/network/maximum_flow.go
```diff
@@ -0,0 +1,209 @@
...
+	// For each edge current->neighbor in the original graph:
+	// 1) add a forward edge with its capacity
+	// 2) initialize its flow to 0.0
```
kortschak:
```suggestion
	// 2) initialize its flow to 0
```

graph/network/maximum_flow.go
```diff
@@ -0,0 +1,209 @@
...
+			if capacity < 0 {
+				return nil, fmt.Errorf("edge weights (capacities) can not be negative")
+			}
+
+			// Add forward edge with full capacity.
+			forward := r.g.NewWeightedEdge(u, v, capacity)
+			r.g.SetWeightedEdge(forward)
+
+			// Initialize flow for this edge to zero.
+			r.flow[edgeKey{from: u.ID(), to: v.ID()}] = 0.0
```
kortschak:
```suggestion
			r.flow[edgeKey{from: u.ID(), to: v.ID()}] = 0
```

graph/network/maximum_flow.go
```diff
@@ -0,0 +1,209 @@
...
+		}
+	}
+	return r, nil
+}
+
+// canReachTargetInLevelGraph builds a level graph using BFS on residualGraph.
+// It records, for each reachable node, the list of parents at the previous level.
+// Returns true iff target is reachable from source via positive-capacity edges.
+func canReachTargetInLevelGraph(residualGraph *residualGraph, s, t graph.Node, parents [][]int64) bool {
```
kortschak:
```suggestion
func canReachTargetInLevelGraph(r *residualGraph, s, t graph.Node, parents [][]int64) bool {
```

graph/network/maximum_flow.go
```diff
@@ -0,0 +1,209 @@
...
+			// Initialize flow for this edge to zero.
+			r.flow[edgeKey{from: u.ID(), to: v.ID()}] = 0.0
+		}
+	}
+	return r, nil
+}
+
+// canReachTargetInLevelGraph builds a level graph using BFS on residualGraph.
+// It records, for each reachable node, the list of parents at the previous level.
+// Returns true iff target is reachable from source via positive-capacity edges.
```
kortschak:
```suggestion
// It returns whether target is reachable from source via positive-capacity edges.
```

graph/network/maximum_flow.go
```diff
@@ -0,0 +1,209 @@
...
+	// levels[i] holds the BFS level of node i, or -1 if unvisited.
+	levels := make([]int32, residualGraph.g.Nodes().Len())
+	for i := range levels {
+		levels[i] = -1
+	}
+
+	sID := s.ID()
+	levels[sID] = 0
+
+	queue := list.New()
```
kortschak: Please don't use container/list. There is a node queue in graph/internal/linear.

graph/network/maximum_flow.go
```diff
@@ -0,0 +1,209 @@
...
+				parents[cid] = append(parents[cid], pid)
+				queue.PushBack(cid)
+				// If we reach cid again at the same level, record an additional p.
+			} else if levels[cid] == levels[pid]+1 {
+				parents[cid] = append(parents[cid], pid)
+			}
+		}
+	}
+	// The t node is reachable iff it was assigned a level ≥ 0.
+	return levels[t.ID()] > -1
```
kortschak:
```suggestion
	return levels[t.ID()] >= 0
```
Let's make the syntax match the comment.

graph/network/maximum_flow.go
```diff
@@ -0,0 +1,209 @@
...
+}
+
+// computeBlockingPath finds and augments all blocking‐flow paths in the current
+// level graph of Dinic’s algorithm. It backtracks from target to source using
+// the parents slices, computes each path’s bottleneck capacity, updates both
+// the residual capacities and the flow map, and returns the total flow added.
+func computeBlockingPath(r *residualGraph, s, t graph.Node, parents [][]int64) float64 {
+	var totalFlow = 0.0
```
kortschak:
```suggestion
	var totalFlow float64
```
and move this down to L143 so that it's introduced at the last possible time.

graph/network/maximum_flow.go
```diff
@@ -0,0 +1,209 @@
...
+
+	// path holds node IDs from t back to (eventually) s.
+	path := []int64{t.ID()}
+	var uid = t.ID()
```
kortschak:
```suggestion
	uid := t.ID()
```

graph/network/maximum_flow_test.go
```diff
@@ -0,0 +1,247 @@
...
+// license that can be found in the LICENSE file.
+
+package network
+
+import (
+	"strings"
+	"testing"
+
+	"gonum.org/v1/gonum/floats/scalar"
+
```
kortschak:
```suggestion
```

graph/network/maximum_flow_test.go
```diff
@@ -0,0 +1,247 @@
...
+	"gonum.org/v1/gonum/graph/simple"
+)
+
+const numericalTolerance = 1.e-10
```
kortschak:
```suggestion
const tol = 1e-10
```

graph/network/maximum_flow.go
```diff
@@ -0,0 +1,209 @@
...
+	"gonum.org/v1/gonum/graph"
+	"gonum.org/v1/gonum/graph/simple"
+)
+
+// MaxFlowDinic computes the maximum flow from source to target in a directed,
+// weighted graph using Dinic’s algorithm. It repeatedly builds level graphs
+// and augments blocking flows until no more augmenting paths exist.
+func MaxFlowDinic(graph graph.WeightedDirected, s, t graph.Node) (float64, error) {
+	if s.ID() == t.ID() {
+		return 0, fmt.Errorf("s and t must be different")
```
kortschak: This should be a panic. The message should be something like "no cut between s and t".

graph/network/maximum_flow.go
```diff
@@ -0,0 +1,209 @@
...
+		it := g.From(u.ID())
+		for it.Next() {
+			v := it.Node()
+
+			capacity, ok := g.Weight(u.ID(), v.ID())
+			if !ok {
+				panic("expected a weight for existing edge")
+			}
+			if capacity < 0 {
+				return nil, fmt.Errorf("edge weights (capacities) can not be negative")
```
kortschak: This should be a panic, "negative edge weight".

graph/network/maximum_flow_test.go
kortschak:
```
// Copyright ©2025 The Gonum Authors. All rights reserved.

package network

import (
	"fmt"
	"reflect"
	"testing"

	"gonum.org/v1/gonum/floats/scalar"
	"gonum.org/v1/gonum/graph"
	"gonum.org/v1/gonum/graph/simple"
)

const tol = 1e-10

var dinicTests = []struct {
	name      string
	g         graph.WeightedDirected
	testFlows []testFlow
}{
	{
		name: "source_is_target",
		g: func() graph.WeightedDirected {
			g := simple.NewWeightedDirectedGraph(0, 0)
			g.AddNode(simple.Node(0))
			return g
		}(),
		testFlows: []testFlow{
			{s: simple.Node(0), t: simple.Node(0), wantPanic: "no cut between s and t"},
		},
	},
	{
		name: "negative_capacity",
		g: func() graph.WeightedDirected {
			g := simple.NewWeightedDirectedGraph(0, 0)
			for _, e := range []simple.WeightedEdge{
				{F: simple.Node(0), T: simple.Node(1), W: 0.3},
				{F: simple.Node(1), T: simple.Node(2), W: -0.6},
			} {
				g.SetWeightedEdge(e)
			}
			return g
		}(),
		testFlows: []testFlow{
			{s: simple.Node(0), t: simple.Node(1), wantPanic: "negative edge weight"},
		},
	},
	{
		name: "three_disjoint_paths",
		g: func() graph.WeightedDirected {
			g := simple.NewWeightedDirectedGraph(0, 0)
			for _, e := range []struct{ u, v int64 }{
				{0, 1},
				{1, 4},
				{0, 2},
				{2, 4},
				{0, 3},
				{3, 4},
			} {
				g.SetWeightedEdge(simple.WeightedEdge{F: simple.Node(e.u), T: simple.Node(e.v), W: 1})
			}
			return g
		}(),
		testFlows: []testFlow{
			{s: simple.Node(0), t: simple.Node(4), want: 3},
		},
	},
	{
		name: "three_disjoint_paths_parallel",
		g: func() graph.WeightedDirected {
			g := simple.NewWeightedDirectedGraph(0, 0)
			for _, e := range []struct{ u, v int64 }{
				{0, 1},
				{1, 0},
				{1, 4},
				{0, 2},
				{2, 0},
				{2, 4},
				{0, 3},
				{3, 0},
				{3, 4},
			} {
				g.SetWeightedEdge(simple.WeightedEdge{F: simple.Node(e.u), T: simple.Node(e.v), W: 1})
			}
			return g
		}(),
		testFlows: []testFlow{
			{s: simple.Node(0), t: simple.Node(4), want: 3},
		},
	},
	{
		name: "cycle_with_tail",
		g: func() graph.WeightedDirected {
			g := simple.NewWeightedDirectedGraph(0, 0)
			// Cycle: 0->1->2->0 and tail 2->3
			for _, e := range []simple.WeightedEdge{
				{F: simple.Node(0), T: simple.Node(1), W: 0.3},
				{F: simple.Node(1), T: simple.Node(2), W: 0.6},
				{F: simple.Node(2), T: simple.Node(0), W: 0.9},
				{F: simple.Node(2), T: simple.Node(3), W: 0.7},
			} {
				g.SetWeightedEdge(e)
			}
			return g
		}(),
		testFlows: []testFlow{
			{s: simple.Node(0), t: simple.Node(3), want: 0.3},
			{s: simple.Node(1), t: simple.Node(3), want: 0.6},
		},
	},
	{
		name: "cycle_with_tail_parallel",
		g: func() graph.WeightedDirected {
			g := simple.NewWeightedDirectedGraph(0, 0)
			// Layers: 0->{1,2,3}, {1,2}->{4,5}, {3}->{5,6}, {4,5,6}->{7}
			for _, e := range []simple.WeightedEdge{
				{F: simple.Node(0), T: simple.Node(1), W: 0.3},
				{F: simple.Node(1), T: simple.Node(2), W: 0.6},
				{F: simple.Node(2), T: simple.Node(0), W: 0.9},
				{F: simple.Node(2), T: simple.Node(3), W: 0.7},
				{F: simple.Node(1), T: simple.Node(0), W: 1.3},
				{F: simple.Node(2), T: simple.Node(1), W: 1.6},
				{F: simple.Node(0), T: simple.Node(2), W: 1.9},
			} {
				g.SetWeightedEdge(e)
			}
			return g
		}(),
		testFlows: []testFlow{
			{s: simple.Node(0), t: simple.Node(3), want: 0.7},
			{s: simple.Node(1), t: simple.Node(3), want: 0.7},
		},
	},
	{
		name: "four_layer_dag",
		g: func() graph.WeightedDirected {
			g := simple.NewWeightedDirectedGraph(0, 0)
			for _, e := range []struct{ u, v int64 }{
				{0, 1},
				{0, 2},
				{0, 3},
				{1, 4},
				{2, 4},
				{2, 5},
				{3, 5},
				{3, 6},
				{4, 7},
				{5, 7},
				{6, 7},
			} {
				g.SetWeightedEdge(simple.WeightedEdge{F: simple.Node(e.u), T: simple.Node(e.v), W: 1})
			}
			return g
		}(),
		testFlows: []testFlow{
			{s: simple.Node(0), t: simple.Node(7), want: 3},
			{s: simple.Node(3), t: simple.Node(7), want: 2},
			{s: simple.Node(0), t: simple.Node(5), want: 2},
			{s: simple.Node(2), t: simple.Node(4), want: 1},
		},
	},
	{
		name: "diamond_with_cross",
		g: func() graph.WeightedDirected {
			g := simple.NewWeightedDirectedGraph(0, 0)
			for _, e := range []simple.WeightedEdge{
				{F: simple.Node(0), T: simple.Node(1), W: 10},
				{F: simple.Node(0), T: simple.Node(2), W: 10},
				{F: simple.Node(1), T: simple.Node(2), W: 5},
				{F: simple.Node(1), T: simple.Node(3), W: 10},
				{F: simple.Node(2), T: simple.Node(3), W: 10},
			} {
				g.SetWeightedEdge(e)
			}
			return g
		}(),
		testFlows: []testFlow{
			{s: simple.Node(0), t: simple.Node(3), want: 20},
			{s: simple.Node(0), t: simple.Node(2), want: 15},
		},
	},
	{
		name: "disconnected",
		g: func() graph.WeightedDirected {
			g := simple.NewWeightedDirectedGraph(0, 0)
			for _, e := range []simple.WeightedEdge{
				{F: simple.Node(0), T: simple.Node(1), W: 10},
				{F: simple.Node(1), T: simple.Node(2), W: 5},
				{F: simple.Node(2), T: simple.Node(3), W: 7},
				{F: simple.Node(4), T: simple.Node(5), W: 11},
				{F: simple.Node(5), T: simple.Node(6), W: 10},
			} {
				g.SetWeightedEdge(e)
			}
			return g
		}(),
		testFlows: []testFlow{
			{s: simple.Node(0), t: simple.Node(5), want: 0},
		},
	},
}

type testFlow struct {
	s, t      graph.Node
	want      float64
	wantPanic any
}

func TestMaxFlowDinic(t *testing.T) {
	const tol = 1e-10

	for _, test := range dinicTests {
			for _, flow := range test.testFlows {
				t.Run(fmt.Sprintf("%d_to_%d", flow.s, flow.t), func(t *testing.T) {
					defer func() {
						r := recover()
						if !reflect.DeepEqual(r, flow.wantPanic) {
							t.Errorf("unexpected panic: got:%v want:%v", r, flow.wantPanic)
						}
					}()
					got := MaxFlowDinic(test.g, flow.s, flow.t)
					if !scalar.EqualWithinAbs(got, flow.want, tol) {
						t.Errorf("unexpected maximum flow: got = %v, want = %v", got, flow.want)
					}
				})
			}
		})
	}
}
```

graph/network/maximum_flow.go
```diff
@@ -0,0 +1,211 @@
...
+
+	"gonum.org/v1/gonum/floats/scalar"
+	"gonum.org/v1/gonum/graph"
+	"gonum.org/v1/gonum/graph/simple"
+)
+
+// MaxFlowDinic computes the maximum flow from source to target in a directed,
+// weighted graph using [Dinic's algorithm]. It repeatedly builds level graphs
+// and augments blocking flows until no more augmenting paths exist.
+//
```
kortschak, line 19:
```suggestion
//
// MaxFlowDinic will panic if s and t are the same node or g has any
// reachable negative edge weight.
// 
```

graph/network/maximum_flow.go
```diff
@@ -0,0 +1,211 @@
...
+// [Dinic's algorithm]: https://en.wikipedia.org/wiki/Dinic%27s_algorithm
+func MaxFlowDinic(g graph.WeightedDirected, s, t graph.Node) float64 {
+	if s.ID() == t.ID() {
+		panic("no cut between s and t")
+	}
+	parents := make([][]int64, g.Nodes().Len())
+	r, err := initializeResidualGraph(g)
+	if err != nil {
+		panic("negative edge weight")
+	}
```
kortschak: Do this in `initializeResidualGraph`.

graph/network/maximum_flow.go
```diff
@@ -0,0 +1,211 @@
...
+		}
+		maxFlow += flow
+	}
+	return maxFlow
+}
+
+// initializeResidualGraph builds the residual graph for Dinic’s algorithm.
+// It copies all nodes, adds directed edges with their original capacities,
+// and initializes a zero flow map for each forward edge.
+func initializeResidualGraph(g graph.WeightedDirected) (*residualGraph, error) {
```
kortschak:
```suggestion
func initializeResidualGraph(g graph.WeightedDirected) *residualGraph {
```

Schwarf: I undo this, since your comment above requires an error to cause the panic. Or was this not intended?

graph/network/maximum_flow.go
```diff
@@ -0,0 +1,210 @@
+// Copyright ©2025 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package network
+
+import (
+	"gonum.org/v1/gonum/graph/internal/linear"
```
kortschak: Import decls are stanza'd "std; them; us". Please run `goimports`.

Schwarf: Thanks.

graph/network/maximum_flow.go
```diff
@@ -0,0 +1,211 @@
...
+
+import (
+	"math"
+
+	"gonum.org/v1/gonum/graph/internal/linear"
+
+	"gonum.org/v1/gonum/floats/scalar"
+	"gonum.org/v1/gonum/graph"
+	"gonum.org/v1/gonum/graph/simple"
+)
```
kortschak:
```suggestion
import (
	"math"

	"gonum.org/v1/gonum/floats/scalar"
	"gonum.org/v1/gonum/graph"
	"gonum.org/v1/gonum/graph/internal/linear"
	"gonum.org/v1/gonum/graph/simple"
)
```

kortschak:
The implementation here (maybe the algorithm, I've not yet looked) is not robust to iteration order.
```
 go test -run TestMaxFlowDinic/four_layer_dag/0_to_7 -count 10             
--- FAIL: TestMaxFlowDinic (0.00s)
    --- FAIL: TestMaxFlowDinic/four_layer_dag (0.00s)
        --- FAIL: TestMaxFlowDinic/four_layer_dag/0_to_7 (0.00s)
            maximum_flow_test.go:227: unexpected maximum flow: got = 2, want = 3
--- FAIL: TestMaxFlowDinic (0.00s)
    --- FAIL: TestMaxFlowDinic/four_layer_dag (0.00s)
        --- FAIL: TestMaxFlowDinic/four_layer_dag/0_to_7 (0.00s)
            maximum_flow_test.go:227: unexpected maximum flow: got = 2, want = 3
--- FAIL: TestMaxFlowDinic (0.00s)
    --- FAIL: TestMaxFlowDinic/four_layer_dag (0.00s)
        --- FAIL: TestMaxFlowDinic/four_layer_dag/0_to_7 (0.00s)
            maximum_flow_test.go:227: unexpected maximum flow: got = 2, want = 3
--- FAIL: TestMaxFlowDinic (0.00s)
    --- FAIL: TestMaxFlowDinic/four_layer_dag (0.00s)
        --- FAIL: TestMaxFlowDinic/four_layer_dag/0_to_7 (0.00s)
            maximum_flow_test.go:227: unexpected maximum flow: got = 2, want = 3
FAIL
```

Schwarf:
Hi @kortschak 
Thank you for the feedback.
I will have a look. It took me some time to reproduce the test failure because the test fails "only" sometimes on my machine. This might be subtle bug or me misunderstanding parts of the algorithm.

graph/network/maximum_flow_test.go
```diff
@@ -231,3 +232,35 @@ func TestMaxFlowDinic(t *testing.T) {
...
+	type Edge struct{ u, v int64 }
+	edges := []Edge{
+		{0, 1}, {0, 2}, {0, 3},
+		{1, 4},
+		{2, 4}, {2, 5},
+		{3, 5}, {3, 6},
+		{4, 7}, {5, 7}, {6, 7},
+	}
+
+	r := rand.New(rand.NewSource(42)) // deterministic RNG
```
kortschak: This doesn't do what you think it does here. Iteration order is random for maps and this cannot be controlled except by removing code from the runtime.

All that is needed is to run the tests about ten times. This will uncover the issue in the code in this case. I don't think this test addition is necessary.

graph/network/maximum_flow.go
```diff
@@ -203,8 +200,5 @@ func computeBlockingPath(r *residualGraph, s, t graph.Node, parents [][]int64) f
 }
 
 type residualGraph struct {
-	g    *simple.WeightedDirectedGraph
-	flow map[edgeKey]float64
+	g *simple.WeightedDirectedGraph
 }
```
kortschak: So this type doesn't need to exist anymore. True?

On the basis of how the graph is being used, it seems to me that `r.g` above should be renamed `flow`.

Schwarf:
> Hi @kortschak, I fixed the bug and added a test to check the iteration order robustness.

kortschak:
@Schwarf was this closed intentionally?

Schwarf:
> Just me my own clarity, the change in the fix is that you are now doing all the accounting in the residual graph. In that change you are also apparently changing the overt semantics of the accounting in a way that does seem sensible, but I'd like to understand what the intention was with the original code. Can you give me a brief summary.
> 
> I think we are almost there. There are some style things that I'll go over in the next round, and we need you to open another PR adding yourself to AUTHORS and CONTRIBUTORS with the commit message "A+C: add Andreas Scharf". We merge that PR unaltered to retain your metadata on the commit which is why it's a separate PR.

https://github.com/gonum/gonum/pull/2051#pullrequestreview-3148913199

Originally I thought it would be clearer to track “flow” directly rather than residuals, so I reduced forward capacities and kept a separate flow map as a stand-in for reverse residual capacity. The mistake was that I attached this “flow” to the forward edges, leaving no true reverse arcs in the residual graph, which broke correctness under different iteration orders.
The fix now does all accounting in the residual graph itself: forward residuals decrease, reverse residuals increase. This matches the standard Dinic formulation and avoids the naming confusion, which is why I dropped the explicit flow field.

graph/network/maximum_flow.go
Schwarf: The conventional name in the Dinic literature is residual graph. That’s why I kept the type and field named residual, rather than flow.

graph/network/maximum_flow_test.go
Schwarf: So I just remove the function ```TestDinicIterationOrderRobustness ```?

kortschak: Yes, please.

graph/network/maximum_flow.go
kortschak: OK, in that case, call it that, but if can be a `*simple.WeightedDirectedGraph` instead of a new type.

graph/network/maximum_flow.go
```diff
@@ -0,0 +1,198 @@
...
+
+// MaxFlowDinic computes the maximum flow from source to target in a directed,
+// weighted graph using [Dinic's algorithm]. It repeatedly builds level graphs
+// and augments blocking flows until no more augmenting paths exist.
+//
+// MaxFlowDinic will panic if s and t are the same node or g has any
+// reachable negative edge weight.
+//
+// [Dinic's algorithm]: https://en.wikipedia.org/wiki/Dinic%27s_algorithm
+func MaxFlowDinic(g graph.WeightedDirected, s, t graph.Node) float64 {
```
kortschak: Are we certain that there won't be a case where a user will want to have control over the epsilon?

graph/network/maximum_flow.go
```diff
@@ -0,0 +1,198 @@
...
+
+	sID := s.ID()
+	levels[sID] = 0
+
+	queue := linear.NodeQueue{}
+	queue.Enqueue(s)
+
+	for queue.Len() > 0 {
+		p := queue.Dequeue()
+		pid := p.ID()
```
kortschak:
```suggestion
	levels[s.ID()] = 0

	var queue linear.NodeQueue
	queue.Enqueue(s)

	for queue.Len() > 0 {
		pid := queue.Dequeue().ID()
```

graph/network/maximum_flow.go
Schwarf: You’re right that a fixed `epsilon` is arbitrary. If we make it configurable, we should also let users pick the comparison mode (absolute, relative, mixed), which Gonum already provides via `scalar.EqualWithinAbs`, `scalar.EqualWithinRel`, and `scalar.EqualWithinAbsOrRel`.
For this PR, I’ll switch the internal check to the mixed strategy using `EqualWithinAbsOrRel`. I’ll compute a graph-scale once (max capacity) to set a sensible absolute tolerance, and keep a small, dimensionless relative tolerance. If there’s demand, we can later add a `MaxFlowDinicTol(..., absTol, relTol)` variant to expose these knobs to users.

What do you think?

kortschak: I don't think it needs to be that complex since we're just test for distance from zero (abs is the only thing that needs testing). If it's negative, a reasonable default, otherwise use the value.

Schwarf: Do I understand your suggestion correctly?
```
func MaxFlowDinic(g graph.WeightedDirected, s, t graph.Node, epsilon float64) float64 {
	if s.ID() == t.ID() {
		panic("no cut between s and t")
	}
	parents := make([][]int64, g.Nodes().Len())
	r := initializeResidualGraph(g)

	if epsilon <= 0 {
		epsilon = 1e-12
	}
	var maxFlow float64
	for canReachTargetInLevelGraph(r, s, t, parents) {
		flow := computeBlockingPath(r, s, t, parents)
		if scalar.EqualWithinAbs(flow, 0, epsilon) {
			break
		}
		maxFlow += flow
	}
	return maxFlow
}
```

kortschak: `s/epsilon/eps/g` and `if eps < 0`, but yes. Should also have a comment in the doc explaining this.

graph/network/maximum_flow.go
```diff
@@ -0,0 +1,200 @@
...
+
+// MaxFlowDinic computes the maximum flow from source to target in a directed,
+// weighted graph using [Dinic's algorithm]. It repeatedly builds level graphs
+// and augments blocking flows until no more augmenting paths exist.
+//
+// MaxFlowDinic will panic if s and t are the same node or g has any
+// reachable negative edge weight.
+//
+// The eps parameter specifies an absolute tolerance for treating tiny flow
+// updates as zero. If eps <= 0, a default of 1e-12 is used.
```
kortschak:
```suggestion
// updates as zero. If eps is negative a default of 1e-12 is used.
```

graph/network/maximum_flow.go
```diff
@@ -0,0 +1,200 @@
...
+func MaxFlowDinic(g graph.WeightedDirected, s, t graph.Node, eps float64) float64 {
+	if s.ID() == t.ID() {
+		panic("no cut between s and t")
+	}
+	parents := make([][]int64, g.Nodes().Len())
+	r := initializeResidualGraph(g)
+
+	if eps <= 0 {
+		eps = 1e-12
+	}
```
kortschak:
```suggestion
	if eps < 0 {
		eps = 1e-12
	}
```

Schwarf:
> @Schwarf, please apply the logic change that goes with the comment change.

Thanks for the hint ... I was convinced I applied both.

graph/network/maximum_flow.go
```diff
@@ -0,0 +1,200 @@
...
+
+	nodes := g.Nodes()
+	for nodes.Next() {
+		r.AddNode(nodes.Node())
+	}
+
+	nodes.Reset()
+	// For each edge current->neighbor in the original graph:
+	// 1) add a forward edge with its capacity
+	// 2) initialize its flow to 0
```
kortschak: This is implicit now, correct?

graph/network/maximum_flow.go
```diff
@@ -0,0 +1,200 @@
...
+			if len(path) == 0 {
+				break
+			}
+			vid = path[len(path)-1]
+		}
+
+		// When we’ve backtracked all the way to s:
+		if vid == s.ID() {
+			// 1) Find the minimum residual capacity (bottleneck) along this path.
+			bottleNeckOnPath := math.MaxFloat64
```
kortschak:
```suggestion
			bottleNeckOnPath := math.Inf(0)
```

graph/network/maximum_flow.go
Schwarf: Good point. Since I dropped the flow data structure I suggest to drop this comment section and instead rewrite the method description. 
```
// initializeResidualGraph builds the residual graph for Dinic’s algorithm.
// It copies all nodes and, for each original edge u→v:
//   - adds a forward edge u→v with capacity equal to the original capacity
//   - adds a reverse edge v→u with capacity 0 if one doesn’t already exist
func initializeResidualGraph(g graph.WeightedDirected) *simple.WeightedDirectedGraph {
	r := simple.NewWeightedDirectedGraph(0, 0)

	nodes := g.Nodes()
	for nodes.Next() {
		r.AddNode(nodes.Node())
	}

	nodes.Reset()
	for nodes.Next() {
		u := nodes.Node()
		it := g.From(u.ID())
		for it.Next() {
			v := it.Node()

			capacity, ok := g.Weight(u.ID(), v.ID())
			if !ok {
				panic("expected a weight for existing edge")
			}
			if capacity < 0 {
				panic("negative edge weight")
			}

			// Add forward edge with full capacity.
			forward := r.NewWeightedEdge(u, v, capacity)
			r.SetWeightedEdge(forward)
			// Add reverse edge if it does not exist
			if _, ok := r.Weight(v.ID(), u.ID()); !ok {
				r.SetWeightedEdge(r.NewWeightedEdge(v, u, 0))
			}
		}
	}
	return r
}

```

kortschak: SGTM

Schwarf:
> Thanks
Hi @kortschak 
Thank you for your support, guidance, and patience. I learned a lot during this review and hope to apply it in the future — maybe even in another gonum contribution.

PR #2052, opened 2025-07-31
internal/asm/f64: add AVX2/FMA optimizations for key BLAS operations

LynnColeArt:
This commit adds AVX2/FMA optimized implementations for several Level 1 BLAS operations and GEMM micro-kernels. The optimizations use runtime CPU feature detection to automatically select the best implementation while maintaining compatibility with older processors.

New AVX2/FMA implementations:
- DotUnitary: Dot product of vectors (up to 2.66x faster than SSE2)
- ScalUnitary: Scalar-vector multiplication (up to 1.48x faster than SSE2)
- GemvN: General matrix-vector multiplication
- AxpyUnitary and AxpyUnitaryTo: y += alpha * x operations
- 4x4 and 8x8 GEMM micro-kernels for matrix multiplication

All implementations correctly handle:
- Negative strides
- Non-unit increments
- NaN propagation
- Zero-length vectors

The AVX2 implementations process 4 doubles per vector instruction and use FMA instructions to reduce operation count and improve accuracy.

Please take a look.

### DIFF
--- a/blas/gonum/dgemm.go
+++ b/blas/gonum/dgemm.go
@@ -235,6 +235,13 @@ func dgemmSerial(aTrans, bTrans bool, m, n, k int, a []float64, lda int, b []flo
 
 // dgemmSerial where neither a nor b are transposed
 func dgemmSerialNotNot(m, n, k int, a []float64, lda int, b []float64, ldb int, c []float64, ldc int, alpha float64) {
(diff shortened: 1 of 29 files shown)


kortschak:
Please address the test failures and goformat the code.

Issue #2054, opened 2025-08-23
stat.Quantile does not work correctly for stat.LinInterp


michealroberts:
### What are you trying to do?

I'm trying to calculate the median of an even sample, using the stat.Quantile method on a sorted list of values ... 

### What did you do?

A simple reproduction of the issue:

```go
func TestQuantileReproduction(t *testing.T) {
	values := []float64{1, 2, 3, 4, 5, 6, 7, 8, 9, 10}

	sort.Float64s(values)

	empirical := stat.Quantile(0.5, stat.Empirical, values, nil) // lower middle
	linear := stat.Quantile(0.5, stat.LinInterp, values, nil) // linear interp

	t.Logf("Empirical=%v  LinInterp=%v", empirical, linear)

	if linear != 5.5 {
		t.Fatalf("expected LinInterp median 5.5, got %v", linear)
	}
}
```

### What did you expect to happen?

The value for empirical should be 5, which it is. The value for linear should be 5.5, it is 5.

### What version of Go and Gonum are you using?

```
go version go1.24.5 linux/arm64
```

```
gonum.org/v1/gonum v0.16.0 h1:5+ul4Swaf3ESvrOnidPp4GZbzf0mxVQpDCYUQE7OJfk=
gonum.org/v1/gonum v0.16.0/go.mod h1:fef3am4MQ93R2HHpKnLk4/Tbh/s0+wqD5nfa6Pnwy4E=
```

### Does this issue reproduce with the current master?

Yes.

michealroberts:
Ok sure, in that case are you happy for me to implement the CumulantKind that we need for this ... or perhaps implement the Median method we need without using quantiles. Just a standard even / odd determination? Happy to do both ...

Having a stats module without a simple, user-friendly Median method is somewhat of a mis-step. The measure of central tendency is fundamental to statistics.

kortschak:
It's not a misstep, it's an absence of time to do the correct implementation.

We won't accept a `Median(…)` since it is just `Quantile(0.5, …)`, but if you want to send the change needed to fix #1836, that would be appreciated. Note that the implementation must correctly handle weighted data. This was the sticking point in #1836. The support for this would need to be tested and suitably match a reference implementation's behaviour (this would be R-7, and would need to be simulated in R since R's quantile does not support weights — also the sticking point).

michealroberts:
@kortschak I think you should accept an alias for Median as that's probably what will be the least amount of friction for users. It's just an alias for simpler comprehension.

[...]

PR #2056, opened 2025-08-24
stat: add Linear CumulantKind

### LINKED ISSUE #1836: `stat.Quantile` returns wrong value when dataset is even size

I'm trying to get median value when I request even size of dataset to `stat.Quantile`

### What did you do?

```
func main() {
	xs := []float64{2, 3, 5, 9}
	median := stat.Quantile(0.5, stat.Empirical, xs, nil)

	fmt.Println(median)
}
```

https://go.dev/play/p/8b6vDMunna8

I expect return value `4`

### What actually happened?

its returns `3`

1.19

Yes

michealroberts:
### Description

This PR adds support for a new CumulantKind (Linear) implementing Hyndman & Fan’s Type 7 quantile estimator (matching with the R default). 🚀 

### Summary

✅ Introduces Linear as a new CumulantKind.
✅ Implements linearQuantile, using R7 behaviour e.g., (h = 1 + p*(n-1)) for the unweighted case.
✅ Provides a weighted branch that interpolates linearly over the weighted ECDF (without averaging on exact step boundaries).
✅ Updates TestQuantile to include Linear, with expected values aligned to R7 for unweighted inputs and consistent weighted behavior.
✅ Keeps existing Empirical and LinInterp unchanged, ensuring no breaking changes.

### Hygiene

✅ API changes have been discussed (#2054)
✅ Code is formatted correctly
✅ Tests pass locally

### Misc

This makes gonum’s quantile functionality consistent with widely used defaults (R, NumPy, Excel), while still supporting existing gonum methods.

Also, I wasn't really sure if I should add myself to the `AUTHORS` or `CONTRIBUTORS` lists, if deemed irrelevant I am happy to remove these changes.

---

Closes #1836

Related to #2054

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -93,7 +93,6 @@ Matthew Connelly <matthew.b.connelly@gmail.com>
 Matthieu Di Mercurio <matthieu.dimercurio@gmail.com>
 Max Halford <maxhalford25@gmail.com>
 Maxim Sergeev <gudvinr@gmail.com>
-Michael Roberts <michael@observerly.com>
 Microsoft Corporation
 MinJae Kwon <k239507@gmail.com>
 Nathan Edwards <etaoinshrdluwho@gmail.com>
--- a/AUTHORS
+++ b/AUTHORS
@@ -93,6 +93,7 @@ Matthew Connelly <matthew.b.connelly@gmail.com>
 Matthieu Di Mercurio <matthieu.dimercurio@gmail.com>
 Max Halford <maxhalford25@gmail.com>
 Maxim Sergeev <gudvinr@gmail.com>
+Michael Roberts <michael@observerly.com>
 Microsoft Corporation
 MinJae Kwon <k239507@gmail.com>
 Nathan Edwards <etaoinshrdluwho@gmail.com>
--- a/stat/stat.go
+++ b/stat/stat.go
@@ -1437,12 +1437,12 @@ func linearQuantile(p float64, x, weights []float64, sumWeights float64) float64
 		if p >= 1 {
 			return x[n-1]
 		}
-		h, frac := math.Modf(1 + p*float64(n-1))
+		h, frac := math.Modf(1 + float64(p*float64(n-1)))
 		i := int(h) - 1
 		if frac == 0 {
 			return x[i]
 		}
-		return (1-frac)*x[i] + frac*x[i+1]
+		return float64((1-frac)*x[i]) + float64(frac*x[i+1])
 	}
 
 	// Weighted: linear interpolation over the weighted ECDF.
@@ -1464,7 +1464,7 @@ func linearQuantile(p float64, x, weights []float64, sumWeights float64) float64
 			if alpha >= 1 {
 				return x[i]
 			}
-			return (1-alpha)*x[i-1] + alpha*x[i]
+			return float64((1-alpha)*x[i-1]) + float64(alpha*x[i])
 		}
 	}
 	panic("impossible")
--- a/stat/stat.go
+++ b/stat/stat.go
@@ -24,6 +24,8 @@ const (
 	Empirical CumulantKind = 1
 	// LinInterp linearly interpolates the empirical distribution between sample values, with a flat extrapolation.
 	LinInterp CumulantKind = 4
+	// Linear applies a linear transformation to the empirical distribution.
+	Linear CumulantKind = 7
 )
 
 // WassersteinDistance computes the Wasserstein distance (Earth Mover's Distance)
@@ -1378,6 +1380,8 @@ func Quantile(p float64, c CumulantKind, x, weights []float64) float64 {
 		return empiricalQuantile(p, x, weights, sumWeights)
 	case LinInterp:
 		return linInterpQuantile(p, x, weights, sumWeights)
+	case Linear:
+		return linearQuantile(p, x, weights, sumWeights)
 	default:
 		panic("stat: bad cumulant kind")
 	}
@@ -1417,6 +1421,50 @@ func linInterpQuantile(p float64, x, weights []float64, sumWeights float64) floa
 				t /= weights[i]
 			}
 			return t*x[i-1] + (1-t)*x[i]
+		}
+	}
+	panic("impossible")
+}
+
+func linearQuantile(p float64, x, weights []float64, sumWeights float64) float64 {
+	n := len(x)
+
+	// Unweighted: Hyndman & Fan Type 7 (R default).
+	if weights == nil {
+		if p <= 0 {
+			return x[0]
+		}
+		if p >= 1 {
+			return x[n-1]
+		}
+		h, frac := math.Modf(1 + p*float64(n-1))
+		i := int(h) - 1
+		if frac == 0 {
+			return x[i]
+		}
+		return (1-frac)*x[i] + frac*x[i+1]
+	}
+
+	// Weighted: linear interpolation over the weighted ECDF.
+	// On exact boundaries, return the current x[i] (no averaging).
+	fidx := p * sumWeights
+	var cumsum float64
+	for i := range x {
+		w := weights[i]
+		prev := cumsum
+		cumsum += w
+		if cumsum >= fidx {
+			if i == 0 {
+				return x[0]
+			}
+			alpha := (fidx - prev) / w // in [0,1]
+			if alpha <= 0 {
+				return x[i-1]
+			}
+			if alpha >= 1 {
+				return x[i]
+			}
+			return (1-alpha)*x[i-1] + alpha*x[i]
 		}
 	}
 	panic("impossible")
--- a/stat/stat_test.go
+++ b/stat/stat_test.go
@@ -1534,7 +1534,7 @@ func TestQuantile(t *testing.T) {
 					continue
 				}
 				want := test.ans[k][j]
-				if !scalar.EqualWithinAbs(v, want, tol) && !scalar.Same(v, want) && !(math.IsNaN(v) && math.IsNaN(want)) && math.Abs(v-want) > tol {
+				if !scalar.EqualWithinAbs(v, want, tol) && !scalar.Same(v, want) {
 					t.Errorf("mismatch case %d kind %d percentile %v. Expected: %v, found: %v", i, kind, p, want, v)
 				}
 			}
--- a/stat/stat_test.go
+++ b/stat/stat_test.go
@@ -1434,12 +1434,13 @@ func TestCDF(t *testing.T) {
 }
 
 func TestQuantile(t *testing.T) {
+	const tol = 1e-12
+
 	cumulantKinds := []CumulantKind{
 		Empirical,
 		LinInterp,
 		Linear,
 	}
-	const tol = 1e-12
 
 	for i, test := range []struct {
 		p      []float64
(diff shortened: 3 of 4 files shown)


stat/stat_test.go
```diff
@@ -1524,8 +1532,9 @@ func TestQuantile(t *testing.T) {
 				if test.panics {
 					continue
 				}
-				if v != test.ans[k][j] && !(math.IsNaN(v) && math.IsNaN(test.ans[k][j])) {
-					t.Errorf("mismatch case %d kind %d percentile %v. Expected: %v, found: %v", i, k, p, test.ans[k][j], v)
+				expected := test.ans[k][j]
```
michealroberts: Just easier to handle floats imprecision  if we use this. e.g., `1.90000000001 != 1.9` failing

stat/stat_test.go
```diff
@@ -1490,7 +1496,8 @@ func TestQuantile(t *testing.T) {
 			w: []float64{0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9, 1.0},
 			ans: [][]float64{
 				{1, 2, 3, 4, 7, 7, 8, 10, 10, 10, 10},
-				{1, 1.875, 2.833333333333333, 3.5625, 6.535714285714286, 6.928571428571429, 7.281250000000001, 9.175, 9.45, 9.725, 10},
```
michealroberts, line 1493: Irrelevant change here, but it's covered by the tolerance check addition.

AUTHORS
kortschak, line 1: Please do not do this in this PR. Also for CONTRIBUTORS. This will be done in a separate PR.

stat/stat_test.go
kortschak: We write `expected` as `want` (or in this file, `ans`).

stat/stat_test.go
```diff
@@ -1524,8 +1532,9 @@ func TestQuantile(t *testing.T) {
 				if test.panics {
 					continue
 				}
-				if v != test.ans[k][j] && !(math.IsNaN(v) && math.IsNaN(test.ans[k][j])) {
-					t.Errorf("mismatch case %d kind %d percentile %v. Expected: %v, found: %v", i, k, p, test.ans[k][j], v)
+				expected := test.ans[k][j]
+				if !(math.IsNaN(v) && math.IsNaN(expected)) && math.Abs(v-expected) > tol {
```
kortschak: `!scalar.EqualWithinAbs(v, test.ans[k][j], tol) && !scalar.Same(v, test.ans[k][j])` and leave the rest the same.

stat/stat_test.go
kortschak: Please explain why this is being removed.

stat/stat_test.go
```diff
@@ -1437,7 +1437,10 @@ func TestQuantile(t *testing.T) {
 	cumulantKinds := []CumulantKind{
 		Empirical,
 		LinInterp,
+		Linear,
 	}
+	const tol = 1e-12
```
kortschak: If there is a `tol`, it should be on the first line of the function, with a blank line between it and the following code.

stat/stat.go
```diff
@@ -1422,6 +1426,50 @@ func linInterpQuantile(p float64, x, weights []float64, sumWeights float64) floa
...
+				return x[0]
+			}
+			alpha := (fidx - prev) / w // in [0,1]
+			if alpha <= 0 {
+				return x[i-1]
+			}
+			if alpha >= 1 {
+				return x[i]
+			}
+			return (1-alpha)*x[i-1] + alpha*x[i]
```
kortschak:
```suggestion
			return float64((1-alpha)*x[i-1]) + float64(alpha*x[i])
```
to avoid FMA which can cause differences in behaviour between arm64 and amd64.

stat/stat.go
```diff
@@ -1422,6 +1426,50 @@ func linInterpQuantile(p float64, x, weights []float64, sumWeights float64) floa
...
+		}
+		if p >= 1 {
+			return x[n-1]
+		}
+		h, frac := math.Modf(1 + p*float64(n-1))
+		i := int(h) - 1
+		if frac == 0 {
+			return x[i]
+		}
+		return (1-frac)*x[i] + frac*x[i+1]
```
kortschak:
```suggestion
		return float64((1-frac)*x[i]) + float64(frac*x[i+1])
```

stat/stat.go
```diff
@@ -1422,6 +1426,50 @@ func linInterpQuantile(p float64, x, weights []float64, sumWeights float64) floa
...
+
+	// Unweighted: Hyndman & Fan Type 7 (R default).
+	if weights == nil {
+		if p <= 0 {
+			return x[0]
+		}
+		if p >= 1 {
+			return x[n-1]
+		}
+		h, frac := math.Modf(1 + p*float64(n-1))
```
kortschak:
```suggestion
		h, frac := math.Modf(1 + float64(p*float64(n-1)))
```

AUTHORS
michealroberts: Amended ✅

stat/stat_test.go
michealroberts: It's not being removed, just amending the floating point precision used as it is covered with the new tolerance changes.

stat/stat.go
michealroberts: Amended ✅

stat/stat.go
michealroberts: Amended ✅

stat/stat.go
michealroberts: Amended ✅

stat/stat_test.go
michealroberts: Amended ✅

stat/stat_test.go
michealroberts: Amended ✅

stat/stat_test.go
kortschak: Removal+addition is identical to mutation under a model of immutability.

stat/stat_test.go
```diff
@@ -1434,10 +1434,14 @@ func TestCDF(t *testing.T) {
...
 
 func TestQuantile(t *testing.T) {
+	const tol = 1e-12
+
 	cumulantKinds := []CumulantKind{
 		Empirical,
 		LinInterp,
+		Linear,
 	}
+
```
kortschak:
```suggestion
```

stat/stat_test.go
```diff
@@ -1524,8 +1533,9 @@ func TestQuantile(t *testing.T) {
 				if test.panics {
 					continue
 				}
-				if v != test.ans[k][j] && !(math.IsNaN(v) && math.IsNaN(test.ans[k][j])) {
-					t.Errorf("mismatch case %d kind %d percentile %v. Expected: %v, found: %v", i, k, p, test.ans[k][j], v)
+				want := test.ans[k][j]
+				if !scalar.EqualWithinAbs(v, want, tol) && !scalar.Same(v, want) && !(math.IsNaN(v) && math.IsNaN(want)) && math.Abs(v-want) > tol {
```
kortschak:
```suggestion
				if !scalar.EqualWithinAbs(v, want, tol) && !scalar.Same(v, want) {
```

stat/stat_test.go
michealroberts: Semantics. I wouldn't worry about it. The tests are identical.

stat/stat_test.go
michealroberts: Does this suggested amendment improve either the readability or compatibility of the code?

stat/stat_test.go
michealroberts: Amended ✅

stat/stat.go
michealroberts: I'd also push back on the idea that we need to do this altogether as alpha and x[i] are both float64.

stat/stat.go
michealroberts: You could probably make an argument that we could make use of:

```go
math.FMA(frac, float64(x[i+1]-x[i]), float64(x[i]))
```
 
Which was added in Go 1.14 and is more idiomatic ...

stat/stat_test.go
kortschak: Everything is semantics. The tests are not identical, but your explanation is satisfactory.

stat/stat_test.go
kortschak: You have introduced an unnecessary change. Please remove it.

stat/stat.go
kortschak: Please read the language spec [regarding intermediate floating point values](https://go.dev/ref/spec#Floating_point_operators) in the context of my comment above.

> An explicit [floating-point type](https://go.dev/ref/spec#Numeric_types) [conversion](https://go.dev/ref/spec#Conversions) rounds to the precision of the target type, preventing fusion that would discard that rounding.

stat/stat_test.go
michealroberts: Should I remove also line 1438?

stat/stat.go
kortschak: You are welcome to send a change that updates all the uses of operations that are currently written to suppress FMA/FMS on arm64 in the code base and we can discuss whether that makes the code base more readable.

stat/stat.go
michealroberts: But they’re already a float64 type …  we don’t need to do any further type conversion. It looks to me that this is redundant in this situation and perhaps a misunderstanding of the above specification you linked. The tests passing both before and after perhaps suggest the change is not necessary.

stat/stat.go
michealroberts: A change for another PR.

stat/stat.go
kortschak: OK. So take look [here](https://github.com/golang/go/blob/25c2d4109f5061425ca2633fb12450d43839bb79/src/cmd/compile/internal/ssa/rewriteARM64.go#L4640-L4729). These rewrites do not happen in amd64. This means that the floating point behaviour of arm64 and amd64 GOARCHs are different (even more so than otherwise). To mitigate this, we force all architectures to downgrade from FMA/FMS. This is unfortunate, but it makes testing and maintenance of the code base given our limited time resources possible. The link that I provided explains exactly how this works; we force the precision to be `float64` not the FPU's internal precision (likely 80 bits).

I don't really feel like arguing about this.

stat/stat_test.go
kortschak: No, that is stanza'd as asked for [here](https://github.com/gonum/gonum/pull/2056#discussion_r2296791950).

stat/stat.go
michealroberts: There are a few confusions going on here, the 80-bits in anything other than x86 architectures is not correct. Go on amd64 uses SSE2 (XMM) registers for floating point, not the old x87 80-bit stack. There is no hidden 80-bit “internal precision” on Go/amd64; values live in 64-bit lanes. Arm64 likewise uses 64-bit IEEE-754 in NEON/FP registers. The 80-bit extended-precision issue applied to the 386 port when using x87, and the Go team explicitly moved away from x87 and requires SSE2 for 386 now.

However, yes there will be tiny, tiny differences that are all but negligible for most if not all practical cases. Just use tolerances in tests and the maintenance burden will not be needed.

I also don’t want to argue, it’s very late here. I think we’re probably circling around two differing mindsets. Happy for this to be resolved.

kortschak: > However, yes there will be tiny, tiny differences that are all but negligible for most if not all practical cases. Just use tolerances in tests and the maintenance burden will not be needed.

The impact is more wide reaching that this in areas where there are unstable primitives. It's not just a matter of increasing tolerances.

michealroberts: Yeh but I certainly wouldn’t advocate for increasing tolerances. Happy to close this out and leave as is to ensure perfect precision across multiple architectures.

michealroberts:
@kortschak Unless there’s anything else, all amends have been applied.

kortschak (commented):
I would expect `Quantile(p, <cumulant kind>, []float64{x₁, x₂, … xₙ}, nil)` to equal `Quantile(p, <cumulant kind>, []float64{x₁, x₂, …, x}, []float64{1, 1, …, 1)`. This is not the case for this addition.

michealroberts:
> I would expect `Quantile(p, <cumulant kind>, []float64{x₁, x₂, … xₙ}, nil)` to equal `Quantile(p, <cumulant kind>, []float64{x₁, x₂, …, x}, []float64{1, 1, …, 1)`. This is not the case for this addition.

Agree. Will amend the test cases for this and will find the issue with my implementation.

kortschak:
This was implemented in #2076.

PR #2074, opened 2026-03-09
Yukai Yang patch

yukai-yang:
Please take a look.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -142,6 +142,7 @@ Will Tekulve <tekulve.will@gmail.com>
 Yasuhiro Matsumoto <mattn.jp@gmail.com>
 Yevgeniy Vahlis <evahlis@gmail.com>
 Yucheng Zhu <zyctc000@gmail.com>
+Yukai Yang <yukai.yang@gmail.com>
 Yunomi <ynmtywn@gmail.com>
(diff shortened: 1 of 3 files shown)


kortschak (commented):
I agree with the `SymRankOne` change.

I'm less convinced with the `IsEmpty` change, maybe; I don't see a sane code path that lets client code read the `ch.chol.IsEmpty()` call with `ch.chol` being nil. It is possible if `Factorize()` was never called on the receiver, but why would client code do that? I see two fixes for this; either what you have, because people will do things that are not sane if they are not advised to avoid them, or document that it will panic if it is called before `Factorize()` has been called on the receiver.

Please note the documented form for commit messages.

yukai-yang:
Thanks for the review! I have updated the commit messages to follow the documented format.

kortschak:
Sorry for the delay, can you amend and push the last commit to trigger actions again.

kortschak:
@yukai-yang Can add another commit adding yourself to AUTHORS and CONTRIBUTORS with the commit message "A+C: add Yukai Yang".

yukai-yang:
Done. CI is running. Thanks!

kortschak (approved):
Thanks

PR #2075, opened 2026-03-10
Fix/dstarlite fixes

schummar:
Fixes two issues that are kind of connected, so I combined them in one PR - let me know if you want to have it differently.

- When initializing DStarLite with the start node being the same as the target node, that leaves the instance in an inconsisten state (the target's `g` value is `+Inf`). After moving and trying to calculate a path, it would either return an invalid path or even go into an infinte loop, osciallting between two or more nodes next to the target.

- When using `MoveTo` to go to a node that has not previously been expanded, `Path` would return an invalid path because a new shortest route was never calculated.

### DIFF
--- a/graph/path/dynamic/dstarlite.go
+++ b/graph/path/dynamic/dstarlite.go
@@ -281,6 +281,7 @@ func (d *DStarLite) MoveTo(n graph.Node) {
 	d.last = d.s
 	d.s = d.model.Node(n.ID()).(*dStarLiteNode)
 	d.keyModifier += d.heuristic(d.last, d.s)
+	d.findShortestPath()
 }
 
 // UpdateWorld updates or adds edges in the world graph. UpdateWorld will
--- a/graph/path/dynamic/dstarlite.go
+++ b/graph/path/dynamic/dstarlite.go
@@ -63,6 +63,9 @@ func NewDStarLite(s, t graph.Node, g graph.Graph, h path.Heuristic, m WorldModel
 		heuristic: h,
 	}
 	d.t.rhs = 0
+	if d.s.ID() == d.t.ID() {
+		d.s.rhs = 0
+	}
 
 	/*
 		procedure Main()
--- a/graph/path/dynamic/dstarlite_test.go
+++ b/graph/path/dynamic/dstarlite_test.go
@@ -632,6 +632,29 @@ func TestDStarLiteDynamic(t *testing.T) {
 	}
 }
 
+func TestMoveToAway(t *testing.T) {
+	t.Parallel()
+
+	g := simple.NewWeightedDirectedGraph(0, math.Inf(1))
+	p0 := simple.Node(0)
+	p1 := simple.Node(1)
+	p2 := simple.Node(2)
+	g.SetWeightedEdge(simple.WeightedEdge{F: p0, T: p1, W: 1})
+	g.SetWeightedEdge(simple.WeightedEdge{F: p1, T: p2, W: 1})
+
+	d := NewDStarLite(p1, p2, g, path.NullHeuristic, simple.NewWeightedDirectedGraph(0, math.Inf(1)))
+	_, _ = d.Path()
+
+	d.MoveTo(p0)
+	p, w := d.Path()
+	wantP := []graph.Node{p0, p1, p2}
+	wantW := 2.0
+
+	if !samePath(p, wantP) || w != wantW {
+		t.Errorf("unexpected path after move. got %v (weight %v), want %v (weight %v)", p, w, wantP, wantW)
+	}
+}
+
 type memory bool
 
 func (m memory) String() string {
--- a/graph/path/dynamic/dstarlite_test.go
+++ b/graph/path/dynamic/dstarlite_test.go
@@ -636,79 +636,43 @@ func TestDStarLiteDynamic(t *testing.T) {
 func TestMoveToAway(t *testing.T) {
 	t.Parallel()
 
-	g := simple.NewWeightedDirectedGraph(0, math.Inf(1))
-	p0 := simple.Node(0)
-	p1 := simple.Node(1)
-	p2 := simple.Node(2)
-	g.SetWeightedEdge(simple.WeightedEdge{F: p0, T: p1, W: 1})
-	g.SetWeightedEdge(simple.WeightedEdge{F: p1, T: p2, W: 1})
+	g := simple.NewUndirectedGraph()
+	for i := range 2 {
+		g.SetEdge(simple.Edge{F: simple.Node(i), T: simple.Node(i + 1)})
+	}
 
-	d := NewDStarLite(p1, p2, g, path.NullHeuristic, simple.NewWeightedDirectedGraph(0, math.Inf(1)))
-	_, _ = d.Path()
+	d := NewDStarLite(simple.Node(1), simple.Node(2), g, path.NullHeuristic, simple.NewWeightedDirectedGraph(0, math.Inf(1)))
 
-	d.MoveTo(p0)
+	// Shortest path so far would be 1->2. Now move off the shortest path to force a replan.
+	d.MoveTo(simple.Node(0))
+
+	wantP := []graph.Node{simple.Node(0), simple.Node(1), simple.Node(2)}
 	p, w := d.Path()
-	wantP := []graph.Node{p0, p1, p2}
-	wantW := 2.0
-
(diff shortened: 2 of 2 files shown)


graph/path/dynamic/dstarlite.go
```diff
@@ -63,6 +63,9 @@ func NewDStarLite(s, t graph.Node, g graph.Graph, h path.Heuristic, m WorldModel
 		heuristic: h,
 	}
 	d.t.rhs = 0
+	if d.s.ID() == d.t.ID() {
+		d.s.rhs = 0
+	}
```
kortschak, line 68: For this to be true, t == s and we should make d be in a state that that it returns a nil path. This should happen at the head of the function. But I do wonder about why this would be called when t == s. Sure, we should handle it, but it's a little odd.

graph/path/dynamic/dstarlite.go
```diff
@@ -281,6 +284,7 @@ func (d *DStarLite) MoveTo(n graph.Node) {
 	d.last = d.s
 	d.s = d.model.Node(n.ID()).(*dStarLiteNode)
 	d.keyModifier += d.heuristic(d.last, d.s)
+	d.findShortestPath()
```
kortschak, line 287: No, this should not happen here. This is a movement of the cursor. Pathfinding should only happen in the world update. If you want a path after you have moved, you need to update the world.

graph/path/dynamic/dstarlite.go
kortschak: I've had a look at the test case now; it's more complex than I suggested, but I need to know why you would do that. Can you explain how this came up in reality?

This is a simpler repro that gets to the heart of the issue and shows why it's unclear why you'd want to be in this situation
```
func TestStartEqualsTarget(t *testing.T) {
	ok := make(chan struct{})
	go func() {
		select {
		case <-time.After(time.Second):
			panic("timed out")
		case <-ok:
		}
	}()

	g := simple.NewUndirectedGraph()
	for uid := range 3 {
		for vid := uid + 1; vid < 3; vid++ {
			g.SetEdge(simple.Edge{F: simple.Node(uid), T: simple.Node(vid)})
		}
	}
	d := NewDStarLite(simple.Node(0), simple.Node(0), g, path.NullHeuristic, simple.NewWeightedDirectedGraph(0, math.Inf(1)))
	d.MoveTo(simple.Node(1))
	d.Path()

	close(ok)
}
```
Specifically, you start at the target and then move away from it. Why? Maybe there's a good reason for this, but I'd like to understand what it is first. ISTM that a sensible approach for client code would be to not create a planner unless s ≠ t since there is no work to be done when you are already at your destination.

I have a similar question about why the other case would arise, the test asks the question, "If I have the path graph 0--1--2 and I want to get from 1 to 2, and ask for that solution, but ignore it and move to 0, how do I get to 2?" This does not seem sensible, but again, I'd like to understand how it arose. Maybe there _is_ a good reason for it. This case does not need a fix though, even if there is a sensible reason for it; the API is not being used correctly. This is a test that shows how it should be used, based off the one added in this PR.
```
func TestMoveToAway(t *testing.T) {
	t.Parallel()

	g := simple.NewUndirectedGraph()
	var wantP []graph.Node
	for i := range 3 {
		wantP = append(wantP, simple.Node(i))
		if i == 0 {
			continue
		}
		g.SetEdge(simple.Edge{F: simple.Node(i - 1), T: simple.Node(i)})
	}

	d := NewDStarLite(simple.Node(1), simple.Node(2), g, path.NullHeuristic, simple.NewWeightedDirectedGraph(0, math.Inf(1)))
	d.Path() // Why? This has no effect.

	d.MoveTo(simple.Node(0))

	// This is a missing action in the original test. With appeal to real
	// world physics, we can say that for the cursor to have moved from
	// 1 to 0, there must be an 0--1, so we need to update the world since
	// it has not previously seen that edge.
	d.UpdateWorld([]graph.Edge{simple.Edge{F: simple.Node(0), T: simple.Node(1)}})

	p, w := d.Path()
	if !samePath(p, wantP) || w != float64(len(wantP)-1) {
		t.Errorf("unexpected path after move. got %v (weight %v), want %v (weight %v)", p, w, wantP, len(wantP)-1)
	}
}
```

schummar: I have a use case where I want to display a shortest route to the destination that the user has picked. But the user might decide at any point to not strictly follow the ideal path but to move in a completely different direction for a time. The displayed route should then switch to the new shortest path - from the user's new location. The world does not change in our case.

The case where s == t is really only coincidental. It happended because in our code the user's and the target's real positions are first discritized into a grid. So even though they are not at exactly the same place, the algorithm might have nothing to do. I can easily catch this cases myself. But still I think the algorithm should not go into a state that could crash the entire system in case of such an oversight. 😉

The case where `MoveTo` is called not along the calculated path, but potentially in the opposite direction, is hopefully explained by the use case I described. My understanding of the D* Lite algorithm is, that it can efficiently calculate new paths if the agent moves off path by repeating the `findShortestPath` step and therefore expanding the traversed part of the graph. But it's the maintainers' decision of course, whether this possibility should be supported.

kortschak: OK. I think you have convinced me.

I agree that the ergonomics of not hanging a process are worthwhile, and the change to support re-finding the path will not negatively impact on the normal operation of the routine. The tests need to be improved though since the tests that were in the change didn't properly describe the situation you were working against.

schummar: Great! I will work on your feedback for the tests then.

schummar: Done. Let me know if you need further changes or clarifications.

graph/path/dynamic/dstarlite_test.go
```diff
@@ -632,6 +633,48 @@ func TestDStarLiteDynamic(t *testing.T) {
...
+		t.Errorf("unexpected path after move. got %v (weight %v), want %v (weight %v)", p, w, wantP, len(wantP)-1)
+	}
+}
+
+func TestStartEqualsTarget(t *testing.T) {
+	ok := make(chan struct{})
+	go func() {
+		select {
+		case <-time.After(time.Second):
+			panic("timed out")
```
kortschak:
```suggestion
			t.Error("timed out")
```

kortschak (commented):
nit only, then LGTM

graph/path/dynamic/dstarlite_test.go
schummar: Not sure if it matters, since the bug has been fixed. But if you e.g. comment out my changes to verify that the bug existed before the fix, the test process will not cleanly shut down with this, but continue to eat up all the memory and cpu it can get, potentially crashing the system.

So I'd suggest either keeping the `panic` (rather stop the whole test process than the alternative), or maybe somehow try to isolate that part in it's own processe to be able to safely cancel it. The latter seems overkill to me...

kortschak: You're right. There is no good way to test this without the panic.

kortschak (dismissed):
Thanks.

Please send another single commit PR adding yourself to AUTHORS and CONTRIBUTORS with the commit message "A+C: add Marco Schumacher". Once that's done, I'll merge both.

graph/path/dynamic/dstarlite_test.go
kortschak: This works and is not terrible.

```
func TestStartEqualsTarget(t *testing.T) {
	ok := make(chan struct{})
	hang := make(chan struct{})
	go func() {
		select {
		case <-time.After(time.Second):
			close(hang)
		case <-ok:
		}
	}()

	g := simple.NewUndirectedGraph()
	for i := range 2 {
		g.SetEdge(simple.Edge{F: simple.Node(i), T: simple.Node(i + 1)})
	}

	go func() {
		defer close(ok)
		d := NewDStarLite(simple.Node(0), simple.Node(0), g, path.NullHeuristic, simple.NewWeightedDirectedGraph(0, math.Inf(1)))
		d.MoveTo(simple.Node(1))
		d.Path()
	}()

	select {
	case <-ok:
	case <-hang:
		t.Error("timed out")
	}
}
```

schummar:
> Thanks.
> 
> Please send another single commit PR adding yourself to AUTHORS and CONTRIBUTORS with the commit message "A+C: add Marco Schumacher". Once that's done, I'll merge both.

Done

kortschak (approved):

PR #2076, opened 2026-03-21
stat: add type 7 linear cumulant kind

### LINKED ISSUE #1836: `stat.Quantile` returns wrong value when dataset is even size

kortschak:
Please take a look.

Closes #1836

### DIFF
--- a/stat/stat.go
+++ b/stat/stat.go
@@ -24,6 +24,9 @@ const (
 	Empirical CumulantKind = 1
 	// LinInterp linearly interpolates the empirical distribution between sample values, with a flat extrapolation.
 	LinInterp CumulantKind = 4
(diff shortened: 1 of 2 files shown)


vladimir-ch (approved):

fumin (approved):
I think the code matches the algorithm on Wikipedia, looks good to me.

PR #2078, opened 2026-04-06
graph/encoding/{di,}graph6: don't validate graph for all calls

kortschak:
Improves performance of d6 and g6 graph iteration. Identified in #2071.

Please take a look.


graph/encoding/digraph6/digraph6.go
```diff
@@ -269,13 +270,20 @@ func (i *d6ReverseIterator) Reset() { i.from = -1 }
 func (i *d6ReverseIterator) Node() graph.Node { return simple.Node(i.from) }
 
 // numberOf returns the digraph6-encoded number corresponding to g.
-func numberOf(g Graph) int64 {
+func numberOf(g Graph) (n int64) {
```
hamzajer: Here we return a copy of n so at the end The caller has a copy but no one holds the (strong) reference for it to live

 as to why this may be working 
Note that because [Pointer.Value](https://pkg.go.dev/weak#Pointer.Value) is not guaranteed to eventually return nil, even after an object is no longer referenced, the runtime is allowed to perform a space-saving optimization that batches objects together in a single allocation slot. The weak pointer for an unreferenced object in such an allocation may never become nil if it always exists in the same batch as a referenced object. Typically, this batching only happens for tiny (on the order of 16 bytes or less) and pointer-free objects.
found here : https://pkg.go.dev/weak

graph/encoding/graph6/graph6.go
```diff
@@ -220,10 +223,17 @@ func (i *g6Iterator) Reset() { i.to = -1 }
 func (i *g6Iterator) Node() graph.Node { return simple.Node(i.to) }
 
 // numberOf returns the graph6-encoded number corresponding to g.
-func numberOf(g Graph) int64 {
+func numberOf(g Graph) (n int64) {
```
hamzajer: same here

hamzajer:
i think we can use something like this 
`
type g6Iterator struct {
	g    Graph
	meta *graphMetaData // strong reference → weak.Pointer in cache stays valid
	from int64
	to   int64
}
`
`type graphMetaData struct {
	n      int64
	offset int
}`

and have the weak ptr point to this graphMetaData
so that when the iterator is no longer in use the GC clears it

kortschak:
I don't think that fixes the issue; we still have a reference that is not rooted.

vladimir-ch:
Should the code in `graph/encoding/graph6/graph6_example_test.go` call `IsValid`?

graph/encoding/graph6/graph6_test.go
```diff
@@ -261,3 +265,49 @@ func (i implicitCycle) Edge(xid, yid int64) graph.Edge     { return nil }
...
 func (n node) ID() int64 { return int64(n) }
+
+func BenchmarkCopy(b *testing.B) {
+	g6, err := loadGraph6("arxiv_5000.graph6.gz")
+	if err != nil {
+		b.Fatalf("failed to load graph6 data: %v", err)
+	}
+	b.ResetTimer()
+	g := nodeIgnorer{simple.NewUndirectedGraph()}
+	for range b.N {
```
vladimir-ch, line 276: Use `b.Loop`? Then maybe the `if` below can be removed? It was introduced in 1.24, so should be ok to use.

graph/coloring/coloring_test.go
```diff
@@ -748,6 +779,9 @@ func undirectedGraphFrom(g []intset) graph.Undirected {
 func BenchmarkColoring(b *testing.B) {
 	for _, bench := range coloringTests {
 		b.Run(bench.name, func(b *testing.B) {
+			if !graph6.IsValid(bench.g) {
+				b.Fatalf("graph is not valid g6")
+			}
 			b.Run("Dsatur", func(b *testing.B) {
 				for i := 0; i < b.N; i++ {
```
vladimir-ch, line 786: `b.Loop` here and below?

graph/encoding/graph6/graph6_test.go
kortschak: `b.Loop` doesn't just prevent optimising away the call, it prevents optimisation of any of the code within the call. I don't think that's helpful.

vladimir-ch: Ah, ok, thanks.

vladimir-ch (approved):

PR #2082, opened 2026-04-22
all: bump gonum/plot

sbinet:
Please take a look.


sbinet:
I wanted to get rid of the weekly alert from dependabot vis a vis the stale x/image dependency.
but there is a mismatch coming from Go-1.25/Go-1.26 _wrt_ how formatting should be applied.
I guess we can live with the dependabot reminder until we drop Go-1.25 (in ~4months) ?

kortschak:
The issue is how the `rdf.Kind` `String` method is being generated. We're going to need to understand this at some point I suspect. I'll try to take a look this weekend.

kortschak:
Some dependency of goccmack has changed its behaviour, so the rdf generated code needs to be regenerated; `go generate ./graph/formats/rdf`.

kortschak (approved):

PR #2099, opened 2026-09-03
stat: add example for BivariateMoment

### LINKED ISSUE #247: stat: add an example for BivariateMoment
A good idea would be to have the error in the covariance. 
https://stats.stackexchange.com/questions/48366/standard-errors-for-covariance-estimate-in-r

hashirmuzaffar:
Please take a look.

Fixes #247.

The issue suggested that a good example would show the error in the covariance, linking https://stats.stackexchange.com/questions/48366. The example does that: it uses the (2,2) moment to attach a standard error to a covariance estimate, and notes that `BivariateMoment` applies no degrees of freedom correction where `Covariance` does.

The data is generated with a fixed linear dependence so the true covariance is 0.5 by construction, which makes the printed values checkable against theory. For that bivariate normal the analytic μ_22 is `Var(x)Var(y) + 2Cov² = 1.75`, so the estimate lands about 0.23 standard errors from the true value and the reported standard error is demonstrably the right size.

Seeding follows the existing style in `stat_example_test.go`.

`go test ./stat/...`, `go vet ./stat/` and `gofmt -l` are clean.

Apologies for the missing template text on the first push — restored above, and the review comments are addressed in the follow-up commit.

### DIFF
--- a/stat/stat_example_test.go
+++ b/stat/stat_example_test.go
@@ -13,6 +13,52 @@ import (
 	"gonum.org/v1/gonum/stat/distuv"
 )
 
+func ExampleBivariateMoment() {
+	// The bivariate central moment
+	//
+	//	μ_rs = E[(x-μ_x)^r*(y-μ_y)^s]
+	//
+	// generalises the covariance, which is the (1,1) moment. The (2,2)
+	// moment gives the leading term of the sampling variance of a
+	// covariance estimate,
+	//
+	//	Var(cov) ≈ (μ_22 - μ_11^2)/n
+	//
+	// so BivariateMoment can be used to attach a standard error to a
+	// covariance. See https://stats.stackexchange.com/q/48366.
+	const n = 1000
+
+	xs := make([]float64, n)
+	ys := make([]float64, n)
+
+	// Draw ys with a fixed linear dependence on xs so that the covariance
+	// is known to be non-zero.
+	rnd := rand.New(rand.NewPCG(1, 1))
+	for i := range xs {
+		xs[i] = rnd.NormFloat64()
+		ys[i] = 0.5*xs[i] + rnd.NormFloat64()
+	}
+
+	// BivariateMoment applies no degrees of freedom correction, so the
+	// (1,1) moment divides by n where Covariance divides by n-1.
+	m11 := stat.BivariateMoment(1, 1, xs, ys, nil)
+	m22 := stat.BivariateMoment(2, 2, xs, ys, nil)
+	cov := stat.Covariance(xs, ys, nil)
+
+	stdErr := math.Sqrt((m22 - m11*m11) / n)
+
+	fmt.Printf("μ_11     = %.6f\n", m11)
+	fmt.Printf("cov      = %.6f\n", cov)
+	fmt.Printf("μ_22     = %.6f\n", m22)
+	fmt.Printf("std err  = %.6f\n", stdErr)
+
+	// Output:
+	// μ_11     = 0.490614
+	// cov      = 0.491105
+	// μ_22     = 1.685326
+	// std err  = 0.038008
+}
+
 func ExampleLinearRegression() {
 	var (
 		xs      = make([]float64, 100)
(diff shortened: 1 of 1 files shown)


stat/stat_example_test.go
```diff
@@ -13,6 +13,52 @@ import (
 	"gonum.org/v1/gonum/stat/distuv"
 )
 
+func ExampleBivariateMoment() {
+	// The bivariate central moment
+	//
+	//	μ_rs = E[(x-μ_x)^r*(y-μ_y)^s]
```
kortschak, line 19: This will be visible in the godoc for the function, so we don't need to repeat it here (the example will be placed immediately below the function, so the locality is sufficient).

stat/stat_example_test.go
```diff
@@ -13,6 +13,52 @@ import (
...
+	// Draw ys with a fixed linear dependence on xs so that the covariance
+	// is known to be non-zero.
+	rnd := rand.New(rand.NewPCG(1, 1))
+	for i := range xs {
+		xs[i] = rnd.NormFloat64()
+		ys[i] = 0.5*xs[i] + rnd.NormFloat64()
+	}
+
+	// BivariateMoment applies no degrees of freedom correction, so the
+	// (1,1) moment divides by n where Covariance divides by n-1.
```
kortschak, line 43: I think this comment could be placed better; ISTM that it should be on the stanza that calculates `stdErr`. But maybe the clearest would be
```
	// BivariateMoment applies no degrees of freedom correction, so the
	// (1,1) moment divides by n where Covariance divides by n-1.
	m11 := stat.BivariateMoment(1, 1, xs, ys, nil)
	m22 := stat.BivariateMoment(2, 2, xs, ys, nil)
	stdErr := math.Sqrt((m22 - m11*m11) / n)

	cov := stat.Covariance(xs, ys, nil)
```

stat/stat_example_test.go
```diff
@@ -13,6 +13,52 @@ import (
 	"gonum.org/v1/gonum/stat/distuv"
 )
 
+func ExampleBivariateMoment() {
+	// The bivariate central moment
+	//
+	//	μ_rs = E[(x-μ_x)^r*(y-μ_y)^s]
+	//
+	// generalises the covariance, which is the (1,1) moment. The (2,2)
+	// moment gives the leading term of the sampling variance of a
```
kortschak, line 22: "gives" is too strong here.

stat/stat_example_test.go
```diff
@@ -13,6 +13,52 @@ import (
...
+	fmt.Printf("μ_11     = %.6f\n", m11)
+	fmt.Printf("cov      = %.6f\n", cov)
+	fmt.Printf("μ_22     = %.6f\n", m22)
+	fmt.Printf("std err  = %.6f\n", stdErr)
+
+	// Output:
+	// μ_11     = 0.490614
+	// cov      = 0.491105
+	// μ_22     = 1.685326
+	// std err  = 0.038008
```
kortschak, line 59:
```suggestion
	fmt.Printf("μ_11    = %.6f\n", m11)
	fmt.Printf("cov     = %.6f\n", cov)
	fmt.Printf("μ_22    = %.6f\n", m22)
	fmt.Printf("std err = %.6f\n", stdErr)

	// Output:
	// μ_11    = 0.490614
	// cov     = 0.491105
	// μ_22    = 1.685326
	// std err = 0.038008
```

kortschak (commented):
Please note the [Brown M&Ms](https://www.smithsonianmag.com/arts-culture/why-did-van-halen-demand-concert-venues-remove-brown-mms-from-the-menu-180982570/) [in the PR template](https://github.com/gonum/gonum/blob/2b52c28fc756441b75b3af99cddc1649f4439ded/.github/PULL_REQUEST_TEMPLATE#L17-L18).

hashirmuzaffar:
PTAL. The review comments are addressed in fbec41b: the restatement of the moment definition is gone since it is already in the function's documentation just above, the degrees of freedom note moved onto the stanza it describes, the claim about the (2,2) moment is softened, and the printed labels use your alignment.

PR #2100, opened 2026-09-03
stat: document ordering of PCA direction vectors

### LINKED ISSUE #698: stat: PCA - Document that vectors are returned in descending order corresponding to variances
See https://godoc.org/github.com/gonum/stat#PC.Vectors and https://godoc.org/github.com/gonum/stat#PC.Vars

It is documented that the Vars are returned in descending order but not mention of the order is made for the Vectors. One can only assume that they are in the same order as otherwise the PCA process would be meaningless but it would be nice to be explicit in the docs.

hashirmuzaffar:
Please take a look.

Fixes #698.

`VarsTo` documents that the variances are returned in descending order, but `VectorsTo` did not state that its columns are in the corresponding order. As the issue notes, a reader can only infer it, since otherwise the analysis would be meaningless.

Documentation only; no behaviour change.

Apologies for the missing template text on the first push — restored above, and the review comment is addressed in the follow-up commit.

### DIFF
--- a/stat/pca_cca.go
+++ b/stat/pca_cca.go
@@ -48,6 +48,9 @@ func (c *PC) PrincipalComponents(a mat.Matrix, weights []float64) (ok bool) {
 
 // VectorsTo returns the component direction vectors of a principal components
 // analysis. The vectors are returned in the columns of a d×min(n, d) matrix.
+// The columns are ordered to correspond to the variances returned by VarsTo,
+// so the first column is the direction of greatest variance and subsequent
+// columns are in order of decreasing variance.
 //
 // If dst is empty, VectorsTo will resize dst to be d×min(n, d). When dst is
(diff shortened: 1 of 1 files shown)


stat/pca_cca.go
```diff
@@ -48,6 +48,9 @@ func (c *PC) PrincipalComponents(a mat.Matrix, weights []float64) (ok bool) {
 
 // VectorsTo returns the component direction vectors of a principal components
 // analysis. The vectors are returned in the columns of a d×min(n, d) matrix.
+// The columns are ordered to correspond to the variances returned by VarsTo,
+// so the first column is the direction of greatest variance and subsequent
+// columns are in order of decreasing variance.
```
kortschak, line 53:
```suggestion
// The columns are ordered to correspond to the variances returned by VarsTo.
```

hashirmuzaffar:
PTAL. Shortened to the single line you suggested in e18df40.

PR #2102, opened 2026-09-06
dsp/window: report ENBW, fix leakage measurement and check convergence

### LINKED ISSUE #1457: dsp/window: fix spectral leakage parameter documentation

hashirmuzaffar:
Please take a look.

Fixes #1457.

Following the discussion on the issue, this is one change rather than three.

**Measurement.** `parameters` took `spectrum[0]` as the reference, assuming the peak is at DC. A flat top window is built to have a flat passband and peaks a little away from zero frequency, so for `FlatTop` the main lobe width, `K` and the maximum sidelobe were not measuring anything — it reported `ΔF_0 = 0.005`, `K = 0.00` and a positive `ɣ_max = +0.002`. The peak is now located. The half power width is also interpolated between the samples that straddle the −3 dB crossing rather than snapping to the last one above it, which biased it low by up to a bin.

**Convergence.** The parameters are asymptotic in the window length and the default of 20 was too short for them to settle:

| n | ΔF_0 | ɣ_max | β |
|---:|---:|---:|---:|
| 20 | 9.56 | −85.08 | −9.23 |
| 100 | 8.19 | −96.00 | −8.88 |
| 400 | 8.04 | −98.08 | −8.81 |
| converged | 8 | −98.2 | −8.79 |

The default is now 1024 with a matching `-m`, and `-eps` recomputes at twice the window length and warns about any parameter that has not converged to that relative tolerance. It defaults to 1e-3 and can be disabled with `-eps 0`. This seemed better than only raising the default, since it tells the next person when the length they have chosen is not enough.

**ENBW.** Reported alongside `ΔF_0.5` and documented in `doc.go` as the distinct parameter it is. This is what made the existing figures hard to reconcile: ten of the twelve documented `ΔF_0.5` values were equivalent noise bandwidths rather than half power widths. Hann documented 1.5, whose ENBW is 1.500 and whose ΔF_0.5 is 1.44; `Rectangular` documented 0.89, which is its ΔF_0.5, its ENBW being exactly 1. Both are now given for every window so there is nothing left to infer.

All values in `window.go` and `window_complex.go` were measured at `-n 4096 -m 262144` with the convergence check silent. `ΔF_0` and `K` for `Lanczos` and `FlatTop` move as a result of the peak fix; the rest of `ΔF_0`, `K`, `ɣ_max` and `β` were already correct and are unchanged apart from precision.

`go test ./dsp/window/...`, `go vet` and `gofmt -l` are clean.

### DIFF
--- a/dsp/window/cmd/leakage/leakage.go
+++ b/dsp/window/cmd/leakage/leakage.go
@@ -143,13 +143,14 @@ func main() {
 	name := flag.String("window", "", "specify a built-in window name (required if csv not given)")
 	param := flag.Float64("param", math.NaN(), "specify parameter for parametric windows")
 	symm := flag.Bool("symm", true, "specify whether the window is symmetric")
-	n := flag.Int("n", 20, "specify window length")
-	m := flag.Int("m", 1000, "specify sample length (must be greater than n)")
+	n := flag.Int("n", 1024, "specify window length")
+	m := flag.Int("m", 1<<16, "specify sample length (must be greater than n)")
 	csv := flag.String("csv", "", "specify an input file of dB transformed frequency amplitudes (required if window not given)")
 	out := flag.String("o", "", "specify output file for plot (required, formats eps, jpg, jpeg, pdf, png, svg, tex or tif)")
 	width := flag.Float64("width", 16, "specify plot width (cm)")
 	height := flag.Float64("height", 8, "specify plot height (cm)")
 	comment := flag.Bool("comment", false, "output a comment line for the window (only for window function)")
+	eps := flag.Float64("eps", 1e-3, "warn if parameters have not converged to this relative tolerance at -n (zero disables the check)")
 	flag.Parse()
 
 	win := windows[strings.ToLower(*name)]
@@ -190,9 +191,12 @@ func main() {
 		if err != nil {
 			log.Fatal(err)
 		}
+		if *eps > 0 {
+			warnUnconverged(os.Stderr, win.fn(*param), *n, *m, *symm, c, *eps)
+		}
 		if *comment {
-			fmt.Printf("// Spectral leakage parameters: ΔF_0 = %2f, ΔF_0.5 = %.2f, K = %.2f, ɣ_max = %2f, β = %2f.\n",
-				c.deltaF0, c.deltaFhalf, c.k(), c.gammaMax, c.beta)
+			fmt.Printf("// Spectral leakage parameters: ΔF_0 = %2f, ΔF_0.5 = %.2f, ENBW = %.3f, K = %.2f, ɣ_max = %2f, β = %2f.\n",
+				c.deltaF0, c.deltaFhalf, c.enbw, c.k(), c.gammaMax, c.beta)
 		}
 	} else {
 		f, err := os.Open(*csv)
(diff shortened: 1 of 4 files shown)


dsp/window/cmd/leakage/leakage.go
```diff
@@ -266,6 +270,11 @@ type characteristics struct {
 	deltaFhalf float64
 	gammaMax   float64
 	beta       float64
+
+	// enbw is the equivalent noise bandwidth, which is a distinct
+	// parameter from deltaFhalf. It is NaN when the window itself is not
+	// available, as when reading a spectrum from csv.
+	enbw float64
```
kortschak: Do we need a comment on this any more than on the others?

dsp/window/doc.go
```diff
@@ -30,7 +30,16 @@
...
+// the frequency spectrum at -3 dB, that is at half the power of the peak.
+//
+// The ENBW parameter is the equivalent noise bandwidth: the width of the
+// ideal rectangular filter that would pass the same noise power as the
+// window, in DFT bins,
+//
+//	ENBW = N sum(w[i]^2) / (sum(w[i]))^2
+//
+// ENBW and ΔF_0.5 are distinct and should not be substituted for one
+// another; for the Hann window ENBW is 1.5 bins while ΔF_0.5 is 1.44.
```
kortschak: Why do we have this comment?

PR #2111, opened 2026-09-11
Add `cluster` package and kmeans

jacobsorme:
Didn't find a `cluster` package, added here with kmeans. Thoughts? 

Generic implementation that for anything that has a way to calculate a centroid from a cluster and a way to measure distance between centroids/data points (funcs `dist` and `centroid` has to be provided). 

Exposes the generic `KMeans` as well as instantiated `Kmeans2D` and `Kmeans3D`.

Some tests + a visual test (starts localhost server, run with `VISUAL_TEST=1 go test gonum.org/v1/gonum/cluster`)


cluster/cluster.go
```diff
@@ -0,0 +1,116 @@
...
+	return centroids, clusters
+}
+
+type Config struct {
+	Seed1, Seed2 uint64
+	Iter         *int
+	Eps          float64
+}
+
+// KMeans for [x,y] float64 data points. Default config run up to 5 iterations.
```
kortschak, line 69: Conventionally, we would provide examples for instantiation rather than making helpers in the package.

cluster/cluster_test.go
```diff
@@ -0,0 +1,141 @@
...
+				found = true
+			}
+		}
+		if !found {
+			t.Errorf("center %+v not correct", c1)
+		}
+	}
+}
+
+func TestKmeans2DVisual(t *testing.T) {
```
kortschak, line 66: Please don't do this. It's not test code, it's an example.

cluster/cluster_test.go
```diff
@@ -0,0 +1,141 @@
...
+	if os.Getenv("VISUAL_TEST") == "1" {
```
kortschak, line 67: If we were doing this, this would be done with a flag rather than an env var.

cluster/cluster.go
```diff
@@ -0,0 +1,116 @@
+package cluster
```
kortschak, line 5: I suspect that if we were going to do this it would fit into spatial rather than making a new package.

cluster/cluster.go
```diff
@@ -0,0 +1,116 @@
...
+		if allEqual {
+			break
+		}
+		centroids = newCentroids
+	}
+	return centroids, clusters
+}
+
+type Config struct {
+	Seed1, Seed2 uint64
```
kortschak: Please take a look at other packages in the repo to understand how we allow client code to configure random source behaviour.

cluster/cluster.go
```diff
@@ -0,0 +1,116 @@
+package cluster
+
+import (
+	"math"
+	"math/rand/v2"
+)
+
+// Generic implementation of KMeans. Use KMeans2D or KMeans3D for float64 2D or 3D float64 data
+func Kmeans[T1 any](k int, data []T1, config Config, centroid func([]T1) T1, dist func(T1, T1) float64) (centroids []T1, clusters [][]T1) {
```
kortschak: I don't think this needs a config struct.

cluster/cluster.go
```diff
@@ -0,0 +1,116 @@
...
+	Eps          float64
+}
+
+// KMeans for [x,y] float64 data points. Default config run up to 5 iterations.
+// Tolerance for centroid equality is default 0.01, with option epsilon in config
+// Centroid initialization defaults to rand.Shuffle, with seed options in config
+// Setting 0 iterations returns the initialized centroids and their clusters
+func Kmeans2D(k int, data [][2]float64, config Config) (centroids [][2]float64, clusters [][][2]float64) {
+	dist := func(p1, p2 [2]float64) float64 {
+		return math.Sqrt(math.Pow(p1[0]-p2[0], 2) + math.Pow(p1[1]-p2[1], 2))
```
kortschak: `math.Hypot`

cluster/cluster.go
```diff
@@ -0,0 +1,116 @@
...
+	return Kmeans(k, data, config, centroid, dist)
+}
+
+// KMeans for [x,y,z] float64 data points. Default config run up to 5 iterations.
+// Tolerance for centroid equality is default 0.01, with option epsilon in config
+// Centroid initialization defaults to rand.Shuffle, with seed options in config
+// Setting 0 iterations returns the initialized centroids and their clusters
+func Kmeans3D(k int, data [][3]float64, config Config) (centroids [][3]float64, clusters [][][3]float64) {
+	dist := func(p1, p2 [3]float64) float64 {
+		return math.Sqrt(math.Pow(p1[0]-p2[0], 2) + math.Pow(p1[1]-p2[1], 2) + +math.Pow(p1[2]-p2[2], 2))
```
kortschak: `math.Hypot`

kortschak (commented):
There are issues with the initialisation that could be improved (see k-means++). License headers are required on the files.

cluster/cluster.go
jacobsorme: I see. I wouldn't consider the 2D float KMeans truly a helper here tho - feels like 99% will use 2D float KMeans. Having them define the `centroid` and `dist` func could be bad UX. 

I did consider making the generic one private, but could also be bad UX to only provide 2D & 3D float64 variants.

cluster/cluster.go
jacobsorme: Took inspo from there https://docs.scipy.org/doc/scipy/reference/generated/scipy.cluster.vq.kmeans2.html

jacobsorme:
Thanks for the feedback, I'll look into kmeans++ and possibly other methods of init.

Added license headers, `math.Hypot` (only for 2D). Removed config struct input to generic (also removed `Seed2`, now only one `Seed` for both inputs to PCG)

cluster/cluster_test.go
jacobsorme: Changed to `var visualTest *bool = flag.Bool("visual", false, "run visual tests")`

jacobsorme:
@kortschak made some changes, see above

